PR #397, opened 2020-12-18
Classes updates in templates

Ilya-Krylov:
- [x] add dataset requirements for object detection model templates
- [x] add custom_classes as a custom parameter to custom object detection template, that implies calling object detection tools with additional 'classes' parameter.

### DIFF
--- a/pytorch_toolkit/object_detection/model_templates/custom-detection/mobilenet_v2-2s_ssd-256x256/template.yaml
+++ b/pytorch_toolkit/object_detection/model_templates/custom-detection/mobilenet_v2-2s_ssd-256x256/template.yaml
@@ -4,7 +4,6 @@ problem: Custom Object Detection
 framework: OTEDetection v2.1.1
 summary: Custom Object Detection based on MobileNetV2 (SSD).
 annotation_format: COCO
-classes: []
 dependencies:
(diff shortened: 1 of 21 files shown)


pytorch_toolkit/object_detection/model_templates/custom-detection/mobilenet_v2-2s_ssd-256x256/template.yaml
```diff
@@ -35,6 +34,12 @@ hyper_parameters:
     batch_size: 64
     base_learning_rate: 0.05
     epochs: 15
+  extra:
+    custom_classes:
+      name: "Custom classes"
+      param: "--classes"
+      type: "string"
+      default: ""
```
druzhkov-paul, line 42: Classes here are not the part of `dataset_requirements->classes` because those cannot be configured by the user, right? Or what is the reason?

Ilya-Krylov: short: yes

dataset_requirements/classes is used to check dataset, whether it is ok for training or not.

druzhkov-paul (approved):

PR #403, opened 2020-12-28
Fix issue with nncf config reading from checkpoint

LeonidBeynenson:
Update external/mmdetection submodule

Fix issue with bufferisation during loading nncf part of config from a
snapshot's metainfo.


LeonidBeynenson:
run tests

Ilya-Krylov (approved):

PR #411, opened 2021-01-12
templates fixes

Ilya-Krylov:


pytorch_toolkit/tests/run_model_templates_tests.py
```diff
@@ -94,6 +94,39 @@ def test_existence_of_mandatory_files_in_template_dir(self):
...
+            with open(template) as read_file:
+                template_fields = list(yaml.load(read_file, yaml.SafeLoader).keys())
+            max_pos = -1
+            for field in mandatory_fields:
+                try:
+                    pos = template_fields.index(field)
+                except ValueError as e:
+                    logging.error(f'Failed to find "{field}" in {template}')
+                    raise e
+                assert pos > max_pos, f'Unsorted fields in template file: {template}'
```
LeonidBeynenson, line 127: 2 @Ilya-Krylov :  
Ilya, are you sure that the order of keys in the yaml file should be checked?

1. The first point: I have not seen any requirement in any specification on the fields order of `template.yaml` files!
Are you sure that it is required?

2. The second point is using `list(yaml.load(...).keys())`

Please, note that ordering of the keys is **NOT** the default feature of YAML.
YAML specification does **NOT** guarantee that the order of keys after loading is the same as the order of keys in the original file.
It is just a detail of PyYAML implementation (see detailed explanation below).

Moreover, when PyYAML makes `yaml.dump`, it **sorts the keys of the yaml file**!
(So, it will be impossible to create proper yaml config file by `yaml.dump`!)

There were discussions on that in pyyaml github: see the issues https://github.com/yaml/pyyaml/issues/110  and https://github.com/yaml/pyyaml/issues/403

So, if we make this check using PyYAML, we rely on implementation details.
Although I do not think that this implementation detail will be changed in the nearest future, I'm not sure that this is a good idea.


_DETAILS ON PyYAML implementation_
_When PyYAML detects mapping, it just reads subnodes one by one, and stores them as pairs `(key, value)` (see [this line of code](https://github.com/yaml/pyyaml/blob/d0d660d035905d9c49fc0f8dafb579d2cc68c0c8/lib3/yaml/composer.py#L135)). Later when it constructs a mapping node it creates a new dict and inserts keys one by one into the dict (see [this line of code](https://github.com/yaml/pyyaml/blob/d0d660d035905d9c49fc0f8dafb579d2cc68c0c8/lib3/yaml/constructor.py#L144)). Since python  >=3.7 keeps the insertion order of keys in dict-s, the keys are stored in the same order as in the file_.

Ilya-Krylov: 1. There is no such requirement, but I've decided that it is better to have the same order. I think this kind of alignment is useful for visual comparison for example. I would like to see the order across templates.
2. We are using pyyaml only, order of fields does not matter for external consumers. yaml.dump has the option to not sort the keys.

pytorch_toolkit/action_recognition_2/model_templates/gesture-recognition/s3d-rgb-mobilenet-v3-stream-jester/template.yaml
```diff
@@ -38,7 +39,7 @@ output_format:
   openvino:
     default: true
     input_format: RGB
-compression: null
+optimisations: ~
```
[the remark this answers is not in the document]
Ilya-Krylov: both of them is defined in yaml standard. I like `~` because it looks like a special symbol rather than string/text. But I have no strong argument to use `~` and not to use `null`

pytorch_toolkit/tests/run_model_templates_tests.py
LeonidBeynenson: Ok, let it be!

pytorch_toolkit/action_recognition_2/model_templates/gesture-recognition/s3d-rgb-mobilenet-v3-stream-jester/template.yaml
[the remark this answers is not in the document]
LeonidBeynenson: Ok! But IMHO the choice should be consistent -- so, if we use `~`, we should not use `null`.

pytorch_toolkit/tests/run_model_templates_tests.py
```diff
@@ -94,6 +94,39 @@ def test_existence_of_mandatory_files_in_template_dir(self):
...
+            'optimisations',
+            'metrics',
+            'tensorboard',
+        ]
+        template_files = self._get_template_files()
+        for template in template_files:
+            with open(template) as read_file:
+                template_fields = list(yaml.load(read_file, yaml.SafeLoader).keys())
+            max_pos = -1
+            for field in mandatory_fields:
```
LeonidBeynenson, line 121: @Ilya-Krylov :  
Ilya, what do you think on the more pythonic way?
It is not mandatory, but I could suggest something like
```python
if not set(template_fields) >= set(mandatory_fields):
    absent_fields = set(mandatory_fields) - set(template_fields)
    raise RuntimeError(f'Template {template} does not contain mandatory fields {absent_fields}')

template_mandatory_fields = [s for s in template_fields if s in mandatory_fields]
if template_mandatory_fields != mandatory_fields:
    diff_set = set(enumerate(template_mandatory_fields)) - set(enumerate(mandatory_fields))
    first_diff_field = sorted(diff_set)[0][1]
    raise RuntimeError(f'Wrong order of mandatory fields in template {template} from the field {first_diff_field}')
```

Ilya-Krylov:
run tests

pytorch_toolkit/tests/run_model_templates_tests.py
Ilya-Krylov: Thank you for your suggestion, but I agree that this is not mandatory, and my code is simpler I guess though it is no so pythonic as yours 😉

Ilya-Krylov:
@LeonidBeynenson If you don't mind I would merge the PR since tests are passed, linter issues mostly should be resolved by your PR.

pytorch_toolkit/tests/run_model_templates_tests.py
LeonidBeynenson: Ok!

PR #429, opened 2021-02-04
Feature/ik/extract classes from dataset

Ilya-Krylov:
* custom detection models automatically determine classes from annotation if --classes is not specified
* removed --classes from other scripts such as 'eval.py', 'export.py'
* removed export tests, added e2e tests


pytorch_toolkit/object_detection/tests/train_tests_custom_detection.py
```diff
@@ -12,23 +12,63 @@
 # See the License for the specific language governing permissions
```
evgeny-izutov, line 12: Change the name of the file to reflect the e2e nature of tests?

Ilya-Krylov: Understand your concern, but I would not change a name because it is used in tests discoverer. And this system is in redesign stage now. I would not like to change current behavior. May I leave it as is?

evgeny-izutov: Ok

evgeny-izutov (approved):

PR #434, opened 2021-02-09
Fix/ik/revisited custom detection

Ilya-Krylov:


evgeny-izutov (approved):
Generally OK, but with next PR (which uses `--classes` too) we need to back propagate the `--classes` processing functionality to the base class (to exclude the imposed code duplication).

Ilya-Krylov:
@evgeny-izutov I am going to create a PR that updates your and @sovrasov 's templates with CustomClassesArgParser. is it what you mean?

evgeny-izutov:
@Ilya-Krylov I hope so)

pytorch_toolkit/ote/ote/modules/arg_converters/mmdetection.py
```diff
@@ -63,8 +64,83 @@ class MMDetectionArgsConverter(BaseArgConverter):
...
+class MMDetectionCustomClassesArgsConverter(MMDetectionArgsConverter):
+
+    @staticmethod
+    def _get_classes_from_annotation(annotation_file):
+        with open(annotation_file) as read_file:
+            categories = sorted(json.load(read_file)['categories'], key=lambda x: x['id'])
+        classes_from_annotation = [category_dict['name'] for category_dict in categories]
+        return classes_from_annotation
+
     def _get_extra_train_args(self, args):
```
LeonidBeynenson, line 100: 2 @Ilya-Krylov 
Ilya, unfortunately I have to say that this function `_get_extra_train_args` is not sufficient for compression.
At the moment the base arg converter uses the function `_get_extra_train_args` both for train and compression -- see the method `convert_compress_args` in `arg_converters/base.py`
As you can see there is `TODO(LeonidBeynenson): think on _get_extra_compress_args` over there.
IMHO now it should be implemented. 
(Unfortunately for me, since it suspends my last PR that I hoped to merge soon)
Could you, implement it in `BaseArgConverter` (by default the same as `_get_extra_train_args` and `_get_extra_test_args`).
I think for `compress.py` the function `_get_extra_compress_args` should return `_get_extra_test_args`
It will also receive class info from `.extra_params` , but IMHO it is OK.

pytorch_toolkit/ote/ote/modules/arg_converters/mmdetection.py
```diff
@@ -63,8 +64,83 @@ class MMDetectionArgsConverter(BaseArgConverter):
...
+        classes_from_args = None
+        if 'classes' in args and args['classes']:
+            classes_from_args = args['classes'].split(',')
+
+        classes_from_snapshot = None
+        if args['load_weights'].endswith('.pth'):
+            classes_from_snapshot = load_classes_from_snapshot(args['load_weights'])
+        else:
+            import yaml
+            with open(os.path.splitext(args['load_weights'])[0] + '.extra_params') as read_file:
```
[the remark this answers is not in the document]
LeonidBeynenson: Also could you make this file have name `***.extra_params.yml` ?

pytorch_toolkit/ote/ote/modules/exporters/mmdetection.py
```diff
@@ -65,9 +87,48 @@ def _export_to_openvino(self, args, tools_dir):
...
+        update_config = '--update_config ' + ' '.join(f'{k}={v}' for k,v in update_config_dict.items())
+        update_config = update_config.replace('"', '\\"')
+
+        return update_config
+
+    def _dump_classes(self, args):
+        classes = load_classes_from_snapshot(args['load_weights'])
+        out_folder = args['save_model_to']
+        out_basename = os.path.splitext(args['config'])[0] + '.extra_params'
+        import yaml
```
LeonidBeynenson, line 124: Could you move this import yaml to the header of the file? (it is safe)

pytorch_toolkit/ote/ote/modules/arg_converters/mmdetection.py
```diff
@@ -63,8 +64,83 @@ class MMDetectionArgsConverter(BaseArgConverter):
...
+
+        classes_from_annotation = self._get_classes_from_annotation(args['train_ann_files'].split(',')[0])
+
+        if classes_from_args:
+            if not set(classes_from_args).issubset(set(classes_from_annotation)):
+                raise RuntimeError('Set of classes passed through CLI is not subset of classes in training dataset: '
+                                   f'{classes_from_args} vs {classes_from_annotation}')
+            classes = classes_from_args
+        else:
+            classes = classes_from_annotation
```
LeonidBeynenson, line 113: Shouldn't we also print a warning if `classes_from_snapshot != classes`?
Something about "Attention, the number of classes are different, start finetuning on another dataset"

pytorch_toolkit/ote/ote/modules/arg_parsers/custom_classes.py
```diff
@@ -26,13 +26,32 @@
...
     def get_train_parser(self, config_path):
         parser = argparse.ArgumentParser(parents=[train_args_parser(config_path)], add_help=False)
         parser.add_argument('--classes', required=False,
                             help='Comma-separated list of classes (e.g. "cat,dog,mouse").')
 
         return parser
+
+    def get_test_parser(self, config_path):
+        parser = argparse.ArgumentParser(parents=[test_args_parser(config_path)], add_help=False)
+        parser.add_argument('--classes', required=False,
```
LeonidBeynenson, line 40: 2 @Ilya-Krylov     
Ilya, I think, `required=True` should be here (and below).
It is need to be sure that external app calls the script with the required number of classes.

As we considered, one of "fragile use cases" is when somebody instantiates a template, does not make any operations with it, and then copies IR to the folder of the template just to make evaluation.
In this case if the file `*.extra_params` (or as I proposed `*.extra_params.yml` ) is absent, the script `eval.py` will give a wrong result without any clarification. (This is what I mean by "fragility")

Obviously, at the moment you use another decision when the file `*.extra_params` reads **_always_**  if the parameter `--classes` is not passed, so it is impossible to run `eval.py` on an IR without `export.py` before.
But it seems to me that it would be better to make the file `*.extra_params`  optional,  but make the parameter `--classes` required.
Note that (AFAIK) the parameter `classes` should be always present for model templates of custom detection (if it is not present it is a bug, isn't it?), so it is not an issue to pass `--classes` to the train/eval/export/compress scripts.

pytorch_toolkit/ote/ote/modules/arg_converters/mmdetection.py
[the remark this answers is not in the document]
LeonidBeynenson: And one more thing:
I think that the case when **both** param `--classes` is passed and file `.extra_params` exists, the function should check that they are equal -- otherwise we have a strange bug here!
(I think we _should_ catch such bugs!)

LeonidBeynenson (changes requested):

pytorch_toolkit/object_detection/tests/train_tests_custom_object_detection.py
```diff
@@ -68,29 +68,12 @@ def setUp(self):
...
-                value = [metrics['value'] for metrics in content['metrics'] if metrics['key'] == metric_key][0]
-                self.assertGreaterEqual(value, 0.0)
-
-        def do_finetuning(self, on_gpu):
+        def do_finetuning(self, with_classes, on_gpu):
             log_file = os.path.join(self.output_folder, 'test_finetuning.log')
             initial_command = 'export CUDA_VISIBLE_DEVICES=;' if not on_gpu else ''
+            classes = ''
+            if with_classes:
+                classes = 'person,vehicle,non-vehicle'
```
LeonidBeynenson, line 76: Ilya, why do you test only cases "all classes" and "3 classes"?
As far as I remember, the test uses airport dataset that contains 3 classes -- is it correct?
If yes, both of the use cases check similar functionality.

Could you make several tests with different number of classes? (`num_classes` from 1 to 3 with special value `None`) 

But I think that in this case the test itself should not change the original folder where the model template is instantiated.
Instead of that each of the tests should make a copy of the model template folder (with special name `= original_name + test.id()`) and run tests inside -- I made it for NNCF tests, see `create_object_detection_nncf_test_case`
(Note that for this approach each tests class should have its own different number of classes, whereas now each test function has its number of classes)

pytorch_toolkit/ote/ote/modules/arg_converters/mmdetection.py
```diff
@@ -63,8 +64,83 @@ class MMDetectionArgsConverter(BaseArgConverter):
...
     }
 
+
+def load_classes_from_snapshot(snapshot):
+    import torch
+    return torch.load(snapshot)['meta'].get('CLASSES', [])
+
+def classes_list_to_update_config_dict(cfg, classes):
+    if isinstance(cfg, str):
+        from mmcv import Config
```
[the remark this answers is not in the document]
Ilya-Krylov: fixed

pytorch_toolkit/ote/ote/modules/arg_converters/mmdetection.py
```diff
@@ -63,8 +64,83 @@ class MMDetectionArgsConverter(BaseArgConverter):
...
+    def _get_extra_test_args(self, args):
+        classes_from_args = None
+        if 'classes' in args and args['classes']:
+            classes_from_args = args['classes'].split(',')
+
+        classes_from_snapshot = None
+        if args['load_weights'].endswith('.pth'):
+            classes_from_snapshot = load_classes_from_snapshot(args['load_weights'])
+        else:
+            import yaml
```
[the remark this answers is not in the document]
Ilya-Krylov: fixed

pytorch_toolkit/ote/ote/modules/exporters/mmdetection.py
Ilya-Krylov: fixed

pytorch_toolkit/ote/ote/modules/arg_converters/mmdetection.py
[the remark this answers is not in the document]
Ilya-Krylov: 1. I think it is better to store classes for custom object detection always. So that it is assumed that this file exists.
2. Renamed
3. Already checked.

pytorch_toolkit/ote/ote/modules/exporters/mmdetection.py
Ilya-Krylov: fixed

pytorch_toolkit/ote/ote/modules/arg_converters/mmdetection.py
Ilya-Krylov: what is 'classes' here?

pytorch_toolkit/ote/ote/modules/arg_parsers/custom_classes.py
Ilya-Krylov: not relevant anymore, I guess

pytorch_toolkit/ote/ote/modules/arg_converters/mmdetection.py
Ilya-Krylov: @LeonidBeynenson As far as I understand, not it also works as well, let's avoid complicating this PR.

pytorch_toolkit/object_detection/tests/train_tests_custom_object_detection.py
Ilya-Krylov: what is `num_classes`?

what are you talking about? please don't mix questions/concerns in a single message.
not enough?

def setUp(self):
            self.output_folder = os.path.join(self.template_folder, f'output_{self.id()}')
            os.makedirs(self.output_folder, exist_ok=True)

pytorch_toolkit/ote/ote/modules/arg_converters/mmdetection.py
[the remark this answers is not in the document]
LeonidBeynenson: Ok!

pytorch_toolkit/object_detection/tests/train_tests_custom_object_detection.py
LeonidBeynenson: Ilya, it is one question/concern.
I think you should make several tests (trest classes) for different number of classes.
But to implement it you will have to make copying of instantiated template folder to new folders, like I do in NNCF tests...

pytorch_toolkit/ote/ote/modules/arg_converters/mmdetection.py
LeonidBeynenson: The classes from args/annotation, see the lines above...

pytorch_toolkit/ote/ote/modules/arg_converters/mmdetection.py
LeonidBeynenson: It works, but it won't raise `RuntimeError` if we pass another numer of classes through args

pytorch_toolkit/object_detection/tests/train_tests_custom_object_detection.py
LeonidBeynenson: Ok, after consideration -- there is no need to copy the whole model template folder in the test

pytorch_toolkit/ote/ote/modules/arg_converters/mmdetection.py
LeonidBeynenson: Fixed!

pytorch_toolkit/ote/ote/modules/arg_converters/mmdetection.py
Ilya-Krylov: added

pytorch_toolkit/object_detection/tests/train_tests_custom_object_detection.py
Ilya-Krylov: added more variants of classes

pytorch_toolkit/ote/ote/modules/arg_converters/mmdetection.py
LeonidBeynenson: Ok!

pytorch_toolkit/object_detection/tests/train_tests_custom_object_detection.py
LeonidBeynenson: Ilya, I re-checked: yes, you are right, the file `data/airport/annotation_example_train.json` contains only one class `person`.
But I found the annotation file `data/airport/annotation_faces_train.json` -- it contains faces! (I re-checked that -- drawed the faces, they are correct)
I think these two annotation files may be merges: we can use as the base the file `annotation_example_train.json`  and merge into it `data/airport/annotation_faces_train.json` 
I sent you my script to merge them and the result of merging

So, now we can use in tests two classes (faces and persons) and one class (persons)

Ilya-Krylov: I think its enough to test what is presented now. You can modify test data by yourself if you really want.

PR #443, opened 2021-02-18
modify pytorch version and remove nnfc for object detection

FengYen-Chang:
As the mmdetection is using pytorch 1.5.1, I patched it to make this readme correct.
And I also found the nnfc part is already including in mmdetection, so the it should be no need to get this sub-module. Moreover, this sub-module seems incorrect, it links to OpenCV account, but this package already move to openvinotoolkit account, and it will make error when get the sub-module, so I removed it from `init_venv.sh`.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -17,9 +17,9 @@
 [submodule "external/bert"]
 	path = external/bert
 	url = https://github.com/google-research/bert
-[submodule "external/nncf_pytorch"]
-	path = external/nncf_pytorch
-	url = ../nncf_pytorch
 [submodule "external/cvat"]
 	path = external/cvat
 	url = ../cvat
+[submodule "external/nncf"]
+	path = external/nncf
+	url = ../nncf
--- a/pytorch_toolkit/object_detection/init_venv.sh
+++ b/pytorch_toolkit/object_detection/init_venv.sh
@@ -18,7 +18,6 @@ fi
 
 # Download mmdetection and nncf submodules
 git submodule update --init ../../external/mmdetection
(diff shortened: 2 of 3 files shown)


LeonidBeynenson:
2 @FengYen-Chang:
Hi John Feng,

>  I also found the nncf part is already including in mmdetection, so the it should be no need to get this sub-module. 
At the moment we would prefer to keep `nncf_pytorch` submodule

Yes, at the moment it is not used, since we install nncf package using pip as mmdetection dependency.
But the idea is to use the external submodule `nncf_pytorch` if we are going to use some features of NNCF that are being developed right now -- it is useful if a feature is not present in the current NNCF release yet, but we do not want to wait for the next release (usually NNCF has ~1 release per quarter, it is quite rare).

Indeed, most probably we will use such approach during the next 2-3 weeks.
So, please, do not remove it from `init_venv.sh` script.

> Moreover, this sub-module seems incorrect, it links to OpenCV account, but this package already move to openvinotoolkit account, and it will make error when get the sub-module
There is redirection from the previous address of `nncf_pytorch` repo to the new one, so git clone (and git submodule update) works well.
But I agree that using the new address should be better.
I will appreciate if you fix this submodule address.

And thank you for the fix of the documentation! we missed this inconsistency.

FengYen-Chang:
Hi @LeonidBeynenson 

Thanks for your information, I changed the submodule from nncf_pytorch to nncf and initial this nncf at `init_env.sh`.

LeonidBeynenson (approved):

Ilya-Krylov:
run tests

Ilya-Krylov:
run tests

PR #445, opened 2021-02-20
Some changes to ArgConverter

Ilya-Krylov:


evgeny-izutov (approved):

pytorch_toolkit/ote/ote/modules/builder.py
```diff
@@ -40,8 +41,8 @@ def build_arg_parser(obj_type):
     return build(obj_type, ARG_PARSERS)
 
 
-def build_arg_converter_map(obj_type):
-    return build(obj_type, ARG_CONVERTER_MAPS)
+def build_arg_converter(obj_type):
```
LeonidBeynenson, line 44: Please, Ilya, rename this function to `create_arg_converter`

The reason is uniformity of the code: it is unconvenient, when all `build_...` functions make just "taking from Registry", whereas one build function makes another operation.
I suppose that this break of uniformity should be reflected in the name of the function.

2 @evgeny-izutov:   Please, write your opinion?

pytorch_toolkit/ote/ote/modules/arg_converters/base.py
```diff
@@ -202,28 +200,25 @@ def _convert_args_by_hooks_for_action(hooks_for_action, model_template_path, arg
         extra_args = hooks_for_action.get_extra_args_hook(args)
         update_args.update(extra_args)
 
-        template_folder = os.path.dirname(model_template_path)
         converted_args = {
-            'config': os.path.join(template_folder, args['config']),
+            'config': args['config'],
```
LeonidBeynenson, line 204: Ilya, I'm sorry, but I do not understand how it is possible to skip the `template_folder` here
It is great if you managed to avoid this dependency, but I reviewed the changes and did not find where do you move 
`os.path.join(template_folder, ...)`

Or you force absolute paths here?

LeonidBeynenson (changes requested):

pytorch_toolkit/ote/ote/modules/builder.py
evgeny-izutov: Actually the rule `all build_... functions make just "taking from Registry"` is not defined anywhere, so there is no reason to think in such way in the future. Moreover renaming breaks logical structure of such peace of code: some config info in the input and a functor in the output.

pytorch_toolkit/ote/ote/modules/arg_converters/base.py
Ilya-Krylov: I did not found when it appeared (usage of template folder), but it is assumed that all scripts are run from template folder, so that there is no need to pass template folder.

pytorch_toolkit/ote/ote/modules/builder.py
Ilya-Krylov: So, it will not be renamed, I resolves the comment.

Ilya-Krylov:
run tests

pytorch_toolkit/ote/ote/modules/arg_converters/base.py
LeonidBeynenson: I see. OK.
(Although I like absolute paths a bit more -- this is more reliable. But in this concrete case it should be ok.)

pytorch_toolkit/ote/ote/modules/builder.py
LeonidBeynenson: Ok, as you wish. But my "minority report": I think that this change will decrease the quality of the code a bit.

LeonidBeynenson (approved):
"Disagree and commit" 🙂

PR #447, opened 2021-02-23
Update misc.py

doctorai-in:
argument error

### DIFF
--- a/pytorch_toolkit/ote/ote/utils/misc.py
+++ b/pytorch_toolkit/ote/ote/utils/misc.py
@@ -158,9 +158,9 @@ def run_through_shell(cmd, verbose=True, check=True):
     assert isinstance(cmd, str)
     log_shell_cmd(cmd)
     return subprocess.run(cmd,
+                          verbose,
                           shell=True,
                           check=check,
-                          capture_output=not verbose,
                           executable="/bin/bash")
 
 def get_cuda_device_count():


LeonidBeynenson:
2 @doctorai-in :  I'm sorry, but as far as I can see the issue was in the version of python.
Please, use python 3.7.* or 3.8.*.
The argument `capture_output` appeared in python 3.7 (see details [here](https://docs.python.org/3.7/library/subprocess.html#using-the-subprocess-module))

Ilya-Krylov:
@doctorai-in @LeonidBeynenson If is the only one thing that does not let us work in older versions of python3 I would ask to add an `if` block checking version of python.
@LeonidBeynenson what do you think? will it work correctly if we agree on proposed change with added `if`. Otherwise we have to write everywhere inside markdown pages that python3.7+ is supported only.

LeonidBeynenson:
2 @Ilya-Krylov : I'm not sure that it is the only thing that does not let us work on python 3.6.
As far as I remember may be other issues on python 3.6 -- it seems to me that we had some of them when we tried different options in CI until we moved CI to python 3.7
I use python 3.7 and did not check this.

pytorch_toolkit/ote/ote/utils/misc.py
```diff
@@ -158,9 +158,9 @@ def run_through_shell(cmd, verbose=True, check=True):
     assert isinstance(cmd, str)
     log_shell_cmd(cmd)
     return subprocess.run(cmd,
+                          verbose,
```
Ilya-Krylov, line 161: please add a condition checking minor version of python interpreter and use `capture_output` if minor is greater than 6. Otherwise ignore `verbose` param since `subprocess.run` does not have `verbose` as an argument.

See: 
https://docs.python.org/3.6/library/subprocess.html#module-subprocess
https://docs.python.org/3.7/library/subprocess.html#module-subprocess

Ilya-Krylov:
Closing because of https://github.com/openvinotoolkit/training_extensions/pull/474

PR #450, opened 2021-02-26
Refactor

Ilya-Krylov:


tests/run_model_templates_tests2.py
```diff
@@ -254,7 +252,7 @@ def instantiate_work_dir(root_path, work_dir, all_tests, verbose):
 
     verbose_param = ' --verbose' if verbose else ''
 
-    run_with_log(f'cd {root_path}; python3 ./tools/instantiate.py'
+    run_with_log(f'python3 ./tools/instantiate.py'
```
evgeny-izutov, line 255: It creates the implicit dependency to run the script from the root dir only. Maybe it's better to infer abs path from `models_path`?

Ilya-Krylov: @evgeny-izutov I think that tests are run in controlled environment, so that is does not look as a problem

Ilya-Krylov: do you insist?

evgeny-izutov: Ok

PR #461, opened 2021-03-04
Develop new

Ilya-Krylov:
* removed useless run* scripts that were part of test system
* hid web ote since it is not supported
* hid tf models inside Misc folder and mark them as unsupported
* hid pt models (non-template) inside Misc folder and mark some of them as unsupported
* divided ote tests and models tests


models/text_spotting/model_templates/alphanumeric-text-spotting/text-spotting-0003/template.yaml
```diff
@@ -31,7 +31,7 @@ hyper_parameters:
   basic:
     batch_size: 4
     base_learning_rate: 0.01
-    epochs: 25
+    epochs: 24
```
evgeny-izutov, line 34: the easy fix) :1st_place_medal:

evgeny-izutov (approved):
Previously reviewed too, so no objection.

LeonidBeynenson (changes requested):
2 @Ilya-Krylov :
Ilya, please, remove the function `run_tests_by_pattern` from the files `ote/ote/tests/utils.py` and `models/action_recognition_2/tests/common/utils.py`.
"This is not required any more"
Most probably, the entire file `models/action_recognition_2/tests/common/utils.py` should be removed too.

Ilya-Krylov:
@LeonidBeynenson https://github.com/openvinotoolkit/training_extensions/pull/463

PR #488, opened 2021-03-18
add CONTRIBUTING.md

morkovka1337:
Add contribution guide.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,12 +29,11 @@ If you have an idea how to enhance code of the existing models, you are welcome 
 
 If you want to add training code of your model to this repository, make sure:
 1. Your model code is licensed under the Apache-2.0 or MIT license (GPL-like licenses are not acceptable).
-2. Your model is trained on a dataset which is acceptable for commercial use.
-3. Source framework of the model is PyTorch\*.
-4. You provide the code to:
+2. Source framework of the model is PyTorch\*.
+3. You provide the code to:
    1. Successfully convert your model to the ONNX\* and OpenVINO\* IR format. See details [here](https://docs.openvinotoolkit.org/latest/openvino_docs_MO_DG_Deep_Learning_Model_Optimizer_DevGuide.html)
    2. Check the converted model's accuracy.
-5. Your model code is structured like this:
+4. Your model code is structured like this:
    1. Option A. Your model supports model-templates interface. See examples [here](https://github.com/openvinotoolkit/training_extensions/tree/develop/models). In this case, place your model's code to the `models` folder, make changes in `ote` folder (if needed).
    2. Option B. Your model has classical interface:
       1. Train, Evaluate, Export scripts for training model, evaluating and exporting to the ONNX\*  and OpenVINO\* IR format trained checkpoint.
@@ -42,8 +41,8 @@ If you want to add training code of your model to this repository, make sure:
       In this case, your model should be in a corresponding subfolder of the [misc/pytorch_toolkit](https://github.com/openvinotoolkit/training_extensions/tree/develop/misc/pytorch_toolkit) folder.
    3. All models should provide a detailed documentation (README.md file), containing information about how to train, evaluate and export to ONNX\* and OpenVINO\* IR your model, which datasets are supported, etc.
    4. All models should have unittests for main functionality (train, test, export).
-6. You create a PR to the `develop` branch.
-7. You provide `requirements.txt` file and a bash script for creating virtual environment.
+5. You create a PR to the `develop` branch.
+6. You provide `requirements.txt` file and a bash script for creating virtual environment.
 
 ## Legal Information
 


CONTRIBUTING.md
```diff
@@ -0,0 +1,59 @@
+# Contributing to the OpenVINO\* Training Extensions
+
+Welcome to the Training Extensions and thank you for your interest to it! Training Extensions is licensed under the Apache* License, Version 2.0. By contributing to the project, you agree to the license and copyright terms therein and release your contribution under these terms.
```
Ilya-Krylov, line 3: interest **in** I guess

CONTRIBUTING.md
```diff
@@ -0,0 +1,59 @@
...
+
+If you want to help developing this project, you can do it several ways:
```
Ilya-Krylov, line 5: **in** several ways

Ilya-Krylov (changes requested):

Ilya-Krylov:
Probably some things can be taken from opencv https://github.com/opencv/opencv/wiki/How_to_contribute

CONTRIBUTING.md
morkovka1337: fixed

CONTRIBUTING.md
morkovka1337: fixed

CONTRIBUTING.md
```diff
@@ -0,0 +1,59 @@
...
+If you have an idea how to enhance code of the existing models, you are welcome to create a pull request. You can also add a new model, which does not exist in the repository yet.
+
+### Adding training code of the model
+
+If you want to add training code of your model to this repository, make sure:
+1. Your model code is licensed under the Apache-2.0 or MIT license (GPL-like licenses are not acceptable).
+2. Your model is trained on a dataset which is acceptable for commercial use.
+3. Source framework of the model is PyTorch\*.
+4. You provide the code to:
+   1. Successfully convert your model to the ONNX\* and OpenVINO\* IR format
```
sovrasov, line 35: A link to openvino model optimizer docs page would be useful for to understand what is IR and why do we need onnx.

CONTRIBUTING.md
```diff
@@ -0,0 +1,59 @@
...
+   2. Check the converted model's accuracy.
+5. Your model code is structured like this:
+   1. Option A. Your model supports model-templates interface. See examples [here](https://github.com/openvinotoolkit/training_extensions/tree/develop/models). In this case, place your model's code to the `models` folder, make changes in `ote` folder (if needed).
+   2. Option B. Your model has classical interface:
+      1. Train, Evaluate, Export scripts for training model, evaluating and exporting to the ONNX\*  and OpenVINO\* IR format trained checkpoint.
+      2. Separable components: dataset, model, etc.
+      In this case, your model should be in a corresponding subfolder of the [misc/pytorch_toolkit](https://github.com/openvinotoolkit/training_extensions/tree/develop/misc/pytorch_toolkit) folder.
+   3. All models should provide a detailed documentation (README.md file), containing information about how to train, evaluate and export to ONNX\* and OpenVINO\* IR your model, which datasets are supported, etc.
+   4. All models should have unittests for main functionality (train, test, export).
+6. You create a PR to the `develop` branch.
```
sovrasov, line 45: May be we should also require providing a script to create a virtualenv.

CONTRIBUTING.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Creating pull requests
+
+If you have an idea how to enhance code of the existing models, you are welcome to create a pull request. You can also add a new model, which does not exist in the repository yet.
+
+### Adding training code of the model
+
+If you want to add training code of your model to this repository, make sure:
+1. Your model code is licensed under the Apache-2.0 or MIT license (GPL-like licenses are not acceptable).
+2. Your model is trained on a dataset which is acceptable for commercial use.
```
sovrasov, line 32: I think only our legal can actually decide what is appropriate for commercial use and for a 3rd party contributor this definition is unclear. This repo doesn't store any results of training or training data, so why do we have to require data clearance?

CONTRIBUTING.md
morkovka1337: Agree. Added link to MO.

CONTRIBUTING.md
morkovka1337: Added mention about venv script

CONTRIBUTING.md
evgeny-izutov: Actually it depends on the place to store user models. If we don't give some kind of access to upload a model to our storage we should declare that model (weights) should be shared by GDrive link. In case of GDrive link we should not limit the commercial usage only.

morkovka1337: Ok, so the second point should be removed. What about the first, is it acceptable?

sovrasov: I think yes. At least for code we have clear license requirements.

morkovka1337:
jenkins please retry a build

Ilya-Krylov (approved):

PR #521, opened 2021-04-13
Hide NNCF

druzhkov-paul:


druzhkov-paul:
run tests

druzhkov-paul:
run tests

druzhkov-paul:
run tests

druzhkov-paul:
@Ilya-Krylov @LeonidBeynenson @aleksei-kashapov Please take a look at the last two test runs for this branch. The difference in the code base is solely inside OTEDetection (external/mmdetection submodule), and essentially in the way NNCF is imported. This small change has a great influence of the time needed for OTE tests to complete: 7 hours vs 12 hours. According to the logs, it looks like **every** mmdetection script run takes ~1 minute of extra time if NNCF is imported inside. NNCF import is expensive indeed, but in a normal situation it affects only the first import inside a new clean environment (before caching comes into play). This is somehow does not hold true for OTE testing infrastructure. I'd suggest to rather make changes to the way tests handle this, or hide NNCF imports to be done only if really needed. I'd prefer the former solution. Looking forward to hearing your thoughts about this.

aleksei-kashapov:
The problem with import is a bug of torch. We fixed it for torch < 1.8 and now I am working to add torch > 1.8

druzhkov-paul:
Respective changes have been included in #530 as a part of submodule update.

PR #526, opened 2021-04-15
support lmdb text rec datasets and

morkovka1337:
Add support of text recognition datasets in LMDB format.

Also do not concatenate validation datasets, compute separate metric over every dataset and return weighted average.


morkovka1337:
run tests

morkovka1337:
run tests

Ilya-Krylov (approved):

PR #538, opened 2021-05-11
add time_series training pipeline

bes-dev:
Add training pipeline for time series forecasting in pytorch


Ilya-Krylov:
Could you add `init_venv.sh` and some tests (not long in time) in 'tests' folder? 
See here how tests will be run: https://github.com/openvinotoolkit/training_extensions/blob/develop/tests/run_misc_tests.py#L11

Tests are run manually if you type 'run tests' message here, in conversation of this PR. The test system is not very smart for now, so all existing tests in repo are run.

Please also add link to your algo to the readme page https://github.com/bes-dev/training_extensions/blob/sb/time_series_pipeline/misc/pytorch_toolkit/README.md

bes-dev:
@Ilya-Krylov omg, which type of tests do you want to see for the training pipeline? Tests for the workability of the pipeline? Conformance tests for target metrics? etc.?

Ilya-Krylov:
@bes-dev I do not insists. If you don't care that some new version of some package listed in requirements.txt will break your code you can ignore my comment, if not - please add tests that will check somehow that your code can be run and produce some results at least.

Ilya-Krylov (approved):

bes-dev:
@Ilya-Krylov tests are ok for me, but I did not find any developer guidelines that determine required test sets and maybe some code style for tests. Previously, we did not add tests for training pipelines, so it would be great to clarify this process now.

In my opinion, there are two main ways to testing of training pipelines:
* Unit testing all modules of the pipeline with whole code coverage, but it's expensive by development.
* Conformance testing the whole pipeline, but it's expensive by testing time.

Which way is recomended for openvino/training_extensions?

Ilya-Krylov:
@bes-dev For a recently added algorithm to 'Misc' folder it is desirable to have free-form tests that run your scripts like train.py/eval.py/export.py/etc and check whether they produce some reasonable output.

For example, tests for training procedure should check that the procedure gives us some valid loss value after several training iterations on "toy" dataset (here we check that the code can be run without being crashed, that all python modules are imported successfully , etc), for sure we are not going to test that training will converge using "real" training set. The "toy" dataset should be generated on the fly if it is possible or stored in the repo and must be legally-clean.

During evaluation and exporting we would like to have a check that the original and exported to OpenVINO IR model give quite similar output on small portion of data (the "toy" dataset) , e.g. value of quality metrics should be quite close to each other and to some predefined expected values, that should be stored in the repo.

All this stuff is optional, but desirable in order to have some evidence that the code can be run at least, and new version of pytorch of whatever does not totally break the training/evaluation/export.

An example of tests https://github.com/openvinotoolkit/training_extensions/tree/develop/misc/pytorch_toolkit/text_recognition/tests

Ilya-Krylov:
@bes-dev feel free to add tests as a separate PR

bes-dev:
@Ilya-Krylov omg, sorry, I delayed for other PRs. Ok, I'll add tests separately.

PR #540, opened 2021-05-13
Added sci-kit pre-install

pfinashx:
Freeze version scikit-image to 0.18.1, to avoid AC install error.
Changed venv for AC to Python 3.7 (because of Ubuntu 18 in docker)

### DIFF
--- a/tests/accuracy_validation/test_accuracy_validation.py
+++ b/tests/accuracy_validation/test_accuracy_validation.py
@@ -35,7 +35,7 @@ ids_list = []
 DIFF_TARGET_MIN_GLOBAL = -0.1
 DIFF_TARGET_MAX_GLOBAL = 0.1
 
-TEST_ROOT = Path(__file__).parent
+TEST_ROOT = Path(__file__).parent.resolve()
 PROJECT_ROOT = TEST_ROOT.parent.parent
 env = os.environ.copy()
 if 'INTEL_OPENVINO_DIR' in env:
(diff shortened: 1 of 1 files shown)


pfinashx:
Please see build 291 in OTE validation tests

tests/accuracy_validation/test_accuracy_validation.py
```diff
@@ -59,11 +59,17 @@
 
 @pytest.fixture(autouse=True, scope="module")
 def setup_ac():
-    subprocess.run('virtualenv -ppython3.8 acc_check', cwd=PROJECT_ROOT, check=True, shell=True)
-    subprocess.run(f'{acc_check_activate_string} && {ACC_CHECK_VENV_DIR}/bin/python setup.py install',
+    subprocess.run('virtualenv -ppython3.7 acc_check', cwd=PROJECT_ROOT, check=True, shell=True)
```
LeonidBeynenson, line 62: Pavel, probably it would be better:
```suggestion
    subprocess.run(f'virtualenv -ppython3.7 {ACC_CHECK_VENV_DIR}', cwd=PROJECT_ROOT, check=True, shell=True)
```
-- it should work, shouldn't it?

LeonidBeynenson: Probably it would be better to use
```diff
- TEST_ROOT = Path(__file__).parent
+ TEST_ROOT = Path(__file__).parent.resolve()
```
-- I like using absolute paths everywhere, even if the`__file__` is a relative path

pfinashx: > Pavel, probably it would be better:
> 
> -- it should work, shouldn't it?

Done - see Build 295

LeonidBeynenson (approved):

LeonidBeynenson:
2 @Ilya-Krylov : Ilya, could you merge this PR?
I think it is OK now.

PR #574, opened 2021-07-09
check whether process was killed by OOM killer

Ilya-Krylov:


Ilya-Krylov:
run tests

Ilya-Krylov:
run tests

ote/ote/utils/runners.py
```diff
@@ -79,3 +79,10 @@ def run_with_termination(cmd):
         stderr = stderr.decode('utf-8')
         print(stderr, end='')
         sys.stdout.flush()
+
+    time.sleep(1)
```
sovrasov, line 83: May be something like ` dmesg -T | egrep -i 'killed process'` would allow to avoid waiting and accessing to the file system, but it needs to be tested.

ote/ote/utils/runners.py
```diff
@@ -79,3 +79,11 @@ def run_with_termination(cmd):
...
+    err = f'Out of memory: Killed process {process.pid}'
```
sovrasov, line 84: Seems like sleep is not required in since dmesg reads directly from the kernel's message buffer.

sovrasov (approved):

Ilya-Krylov:
run tests

PR #581, opened 2021-07-26
Ai/fix ov 2021.4 compatipility

aiarkinx:
Changing package dependencies for compatibility with OpenVINO 2021.4

### DIFF
--- a/models/instance_segmentation/init_venv.sh
+++ b/models/instance_segmentation/init_venv.sh
@@ -65,7 +65,7 @@ export MMCV_VERSION=1.3.0
 if [[ $CUDA_VERSION_CODE == "102" ]]; then
   pip install torch==${TORCH_VERSION} torchvision==${TORCHVISION_VERSION} -c constraints.txt
 else
- pip install torch==${TORCH_VERSION}+cu${CUDA_VERSION_CODE} torchvision==${TORCHVISION_VERSION}+cu${CUDA_VERSION_CODE} -f https://download.pytorch.org/whl/torch_stable.html -c constraints.txt
+  pip install torch==${TORCH_VERSION}+cu${CUDA_VERSION_CODE} torchvision==${TORCHVISION_VERSION}+cu${CUDA_VERSION_CODE} -f https://download.pytorch.org/whl/torch_stable.html -c constraints.txt
 fi
 
 pip uninstall -y mmcv
--- a/models/object_detection/init_venv.sh
+++ b/models/object_detection/init_venv.sh
@@ -63,12 +63,12 @@ export TORCHVISION_VERSION=0.9.1
 export MMCV_VERSION=1.3.0
 
 if [[ $CUDA_VERSION_CODE == "102" ]]; then
-  pip install  torch==${TORCH_VERSION} torchvision==${TORCHVISION_VERSION} -c constraints.txt
+  pip install torch==${TORCH_VERSION} torchvision==${TORCHVISION_VERSION} -c constraints.txt
 else
   pip install torch==${TORCH_VERSION}+cu${CUDA_VERSION_CODE} torchvision==${TORCHVISION_VERSION}+cu${CUDA_VERSION_CODE} -f https://download.pytorch.org/whl/torch_stable.html -c constraints.txt
 fi
 
-pip uninstall -y mmcv 
+pip uninstall -y mmcv
 pip install --no-cache-dir mmcv-full==${MMCV_VERSION} -f https://download.openmmlab.com/mmcv/dist/cu${CUDA_VERSION_CODE}/torch${TORCH_VERSION}/index.html -c constraints.txt
 
 # Install other requirements.
--- a/models/object_detection/requirements.txt
+++ b/models/object_detection/requirements.txt
@@ -1,3 +1,4 @@
+--no-cache-dir --no-binary=Polygon3 Polygon3==3.0.8
 -r ../../external/mmdetection/requirements/build.txt
 --no-cache-dir --no-binary=mmpycocotools mmpycocotools==12.0.3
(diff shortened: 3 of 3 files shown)


aiarkinx:
run tests

aiarkinx:
run tests

LeonidBeynenson (approved):

aiarkinx:
Test job with OpenVINO 2021.4 is passed http://213.221.44.203/job/ote/job/experimental/job/precommit/job/trigger/170/

models/object_detection/requirements.txt
```diff
@@ -1,4 +1,5 @@
 -r ../../external/mmdetection/requirements/build.txt
```
Ilya-Krylov, line 1: what about Polygon3?

models/instance_segmentation/init_venv.sh
```diff
@@ -52,33 +55,33 @@ CUDA_VERSION_CODE=$(echo ${CUDA_VERSION} | sed -e "s/\.//" -e "s/\(...\).*/\1/")
...
 export TORCH_VERSION=1.8.1
 export TORCHVISION_VERSION=0.9.1
 export MMCV_VERSION=1.3.0
 
 if [[ $CUDA_VERSION_CODE == "102" ]]; then
-  pip install torch==${TORCH_VERSION} torchvision==${TORCHVISION_VERSION}
+  pip install torch==${TORCH_VERSION} torchvision==${TORCHVISION_VERSION} -c constraints.txt
 else
-  pip install torch==${TORCH_VERSION}+cu${CUDA_VERSION_CODE} torchvision==${TORCHVISION_VERSION}+cu${CUDA_VERSION_CODE} -f https://download.pytorch.org/whl/torch_stable.html
+ pip install torch==${TORCH_VERSION}+cu${CUDA_VERSION_CODE} torchvision==${TORCHVISION_VERSION}+cu${CUDA_VERSION_CODE} -f https://download.pytorch.org/whl/torch_stable.html -c constraints.txt
```
Ilya-Krylov, line 68: indent

models/object_detection/init_venv.sh
```diff
@@ -52,33 +55,33 @@ CUDA_VERSION_CODE=$(echo ${CUDA_VERSION} | sed -e "s/\.//" -e "s/\(...\).*/\1/")
...
-  pip install torch==${TORCH_VERSION} torchvision==${TORCHVISION_VERSION}
+  pip install  torch==${TORCH_VERSION} torchvision==${TORCHVISION_VERSION} -c constraints.txt
 else
-  pip install torch==${TORCH_VERSION}+cu${CUDA_VERSION_CODE} torchvision==${TORCHVISION_VERSION}+cu${CUDA_VERSION_CODE} -f https://download.pytorch.org/whl/torch_stable.html
+  pip install torch==${TORCH_VERSION}+cu${CUDA_VERSION_CODE} torchvision==${TORCHVISION_VERSION}+cu${CUDA_VERSION_CODE} -f https://download.pytorch.org/whl/torch_stable.html -c constraints.txt
 fi
 
-pip uninstall -y mmcv
-pip install --no-cache-dir mmcv-full==${MMCV_VERSION} -f https://download.openmmlab.com/mmcv/dist/cu${CUDA_VERSION_CODE}/torch${TORCH_VERSION}/index.html
+pip uninstall -y mmcv
```
Ilya-Krylov, line 71: trailing white space

models/object_detection/init_venv.sh
```diff
@@ -52,33 +55,33 @@ CUDA_VERSION_CODE=$(echo ${CUDA_VERSION} | sed -e "s/\.//" -e "s/\(...\).*/\1/")
...
+pip install -e ../../ote/ -c constraints.txt
 
 # install PyTorch and MMCV.
 export TORCH_VERSION=1.8.1
 export TORCHVISION_VERSION=0.9.1
 export MMCV_VERSION=1.3.0
 
 if [[ $CUDA_VERSION_CODE == "102" ]]; then
-  pip install torch==${TORCH_VERSION} torchvision==${TORCHVISION_VERSION}
+  pip install  torch==${TORCH_VERSION} torchvision==${TORCHVISION_VERSION} -c constraints.txt
```
Ilya-Krylov, line 66: extra white space

Ilya-Krylov (changes requested):

models/object_detection/requirements.txt
aiarkinx: Polygon3 is missing on object detection dependencies. It isn't needed here.

Ilya-Krylov: object detection, instance segmentation, text spotting is based on the same code base

LeonidBeynenson: I think Polygon3 is not used in objct detection domain, since masks are not used.
Ilya, do you think that we need to add building Polygon3 here for uniformity?

Ilya-Krylov: Polygon3 is not used anywhere except text_spotting, but it is exported everywhere 
https://github.com/openvinotoolkit/mmdetection/blob/ote/mmdet/datasets/coco.py#L19 -> https://github.com/openvinotoolkit/mmdetection/blob/ote/mmdet/core/evaluation/text_evaluation.py#L19

If you face with some problem in instance_segmentation case, then the same problem should be fixed in object_detection, should not it?

Is it possible that cache is not cleaned? that is probably why the problem might be hidden

aiarkinx: >> If you face with some problem in instance_segmentation case, then the same problem should be fixed in object_detection, should not it?
Ok
>> Is it possible that cache is not cleaned? that is probably why the problem might be hidden
No, only instance segmentation and text spotting had problems with Polygon3. Object detection was passed

PR #661, opened 2021-11-02
Add template for stfpm

ashwinvaidya17:
# Changes

- Fix anomalib path for tests
- Add stfpm template
- Fix divide by 0 error
- Add stfpm to test

# Known Issues

- The progress bar still shows current epoch progress rather than global.


external/anomaly/ote/task.py
```diff
@@ -131,6 +131,9 @@ def train(
         model_monitor = ModelMonitorCallback(output_model, self.save_model)
         callbacks = [progress, model_monitor]
 
+        if hasattr(self.model, "callbacks"):
+            callbacks.append(self.model.callbacks)
+
```
djdameln, line 136: Is this really necessary? I think PL should automatically append the model specific callbacks to the callbacks passed to the trainer.

ashwinvaidya17: This is just to make it similar to the original task in anomalib. This is in the train.py 
image

external/anomaly/ote/configuration.yaml
```diff
@@ -17,6 +17,7 @@ model:
     enum_name: AnomalyModel
     options:
       PADIM: padim
+      STFPM: stfpm
```
samet-akcay, line 20: Not sure if we need these here. As recently discussed, multiple model support is currently not available, and is not clear how backend will sort this out

external/anomaly/ote/configs/ote_anomaly_classification_stfpm.yaml
```diff
@@ -0,0 +1,37 @@
+name: Anomaly Classification STFPM
```
samet-akcay, line 1: I think, instead of naming the template filename we need this naming inside the `template.yaml` file. See [this](https://gitlab.devtools.intel.com/vmc-eip/IMPT/impt/-/blob/develop/src/ote_sdk/ote_sdk/entities/model_template.py#L484) model template parser, where it checks `model_template_id` in the yaml file. If it doesn't exist, it'll check `name` in the yaml file. It then replaces " " with "_". 

In the backend, the padim model is registered as `ote_anomaly_classification_padim`. Therefore, we need this either in `model_template_id` in the yaml file, or replace the name in the yaml file

samet-akcay: so the template paths could be like:

- external/anomalib/anomaly_classification/configs/padim/template.yaml
- external/anomalib/anomaly_classification/configs/stfpm/template.yaml

in padim/template.yaml,

```
name: ote anomaly classification padim
```

external/anomaly/ote/task.py
djdameln: Then maybe both should be changed. This check seems unnecessary

samet-akcay: @ashwinvaidya17 noticed that early stopping doesnt work as expected. If this is the case, that means that model callbacks are not detected, and this check could be needed. In stfpm early stopping callback is added [here](https://gitlab-icv.inn.intel.com/algo_rnd_team/anomaly/-/blob/development/anomalib/models/stfpm/model.py#L238).

if model callbacks are not appended to `callbacks` list, then the trainer will only take into account `callbacks = [progress, model_monitor]`, which explains why early stopping doesnt work for stfpm

djdameln: According to [the documentation](https://pytorch-lightning.readthedocs.io/en/latest/common/trainer.html) the model-specific callbacks should extend the list of callbacks passed to the trainer. I'm guessing this only works if the model-specific callbacks are added by the `configure_callbacks` method of the PL module. We're not using this method to add our model-specific callbacks which is probably why it didn't work earlier. I think it would be better to use the `configure_callbacks` method in the way it was designed so that we keep the task implementation as clean as possible.

samet-akcay: This is great! Yes, our Lightning implementations do not implement `configure_callbacks` method, but instead assign `self.callbacks`.  Seems like the above implementation then re-invents `configure_callbacks`. Maybe we could prioritise this and update anomalib algos

djdameln: We'll need to make a separate PR in the anomalib repo that updates the PL modules. I'll create a ticket for this. Maybe we can keep the check here for now and remove it once the PL modules are updated.

samet-akcay: yes agree. i've already created this https://jira.devtools.intel.com/browse/CVS-69866

djdameln: perfect, thanks

external/anomaly/ote/configuration.yaml
ashwinvaidya17: I'll remove this.

samet-akcay (approved):
lgtm

PR #664, opened 2021-11-03
Restructure anomalib folder + add setup.py

ashwinvaidya17:
# Changes

- Restructure folder structure
- Add setup.py
- Fix init_venv.sh

# Coverage

image

# Commands used
```
rm -rf /tmp/ote*
export SC_SDK_REPO=~/impt
export ANOMALIB_REPO=~/anomalib
cd external/anomalib
bash ./init_venv.sh /tmp/ote-anomaly
. /tmp/ote-anomaly/bin/activate
pip install pytest
pytest tests/ -v

### CODE REVIEW

**Comment by ashwinvaidya17** [comment]
@samet-akcay @djdameln I have changed the folder structure as well based on what we discussed in the chat. Might not be the best approach. Feel free to suggest alternatives.

**Review by samet-akcay — CHANGES_REQUESTED** [review]
I think the structrure should be something like the following:

```
external
└── anomalib
    ├── anomaly_classification
    │   ├── __init__.py
    │   ├── configs
    │   │   ├── padim
    │   │   │   └── template.yaml
    │   │   └── stfpm
    │   │       └── template.yaml
    │   ├── configuration.yaml
    │   ├── openvino.py
    │   └── task.py
    ├── init_venv.sh
    ├── setup.py
    ├── src
    │   ├── callbacks
    │   ├── config
    │   │   ├── __init__.py
    │   │   └── config.py
    │   └── data
    │       ├── __init__.py
    │       └── data.py
    └── tests
        ├── __init__.py
        ├── helpers
        └── test_ote_anomaly_classification.py
```

The only part I'm not so sure is the name of `src`. If we cannot find any alternative, it could stay as is. It would be nice to rename it to something more meaningful, though. Something like `ote_anomalib` or similar. This package could also store the classes and wrappers for the ote_cli stuff that are requested from the team. Any suggestions?

ashwinvaidya17:
@samet-akcay I agree with renaming it from `src`. Once we remove it from `anomalib_classification`, we will have to add it to `setup.py`. How about `ote_anomalib_utils` or is it too verbose?

samet-akcay:
yeah could be a bit verbose. In terms of revealing the intention of the package regarding what it's doing, `ote_anomalib` would work, I think. @djdameln what is your thought on this?

samet-akcay:
because, the package could extend more than being utils, covering ote_sdk, ote_cli and so on

djdameln:
> yeah could be a bit verbose. In terms of revealing the intention of the package regarding what it's doing, `ote_anomalib` would work, I think. @djdameln what is your thought on this?

I would personally prefer to have the contents of `src` under `anomaly_classification, like we had initially. But this is because I feel that the difference between the segmentation task and the classification task will be only in the configuration files, and the tasks will use a shared task implementation.

If we do end up with separate task implementations, then it would make sense to have the `src` folder at the same level of `anomaly_classification` (and `anomaly_segmentation`) like you suggested. In that case maybe the folder could be called `shared` or `core` to indicate that it contains the core modules that are needed by both the anomalib task implementations?

Also I'm not sure if we should have the ote_sdk stuff in the same folder as the ote_cli stuff, because they serve quite different purposes and probably won't rely much on the same files.

ashwinvaidya17:
@samet-akcay I have restructured the repo. But I have renamed `src` to `shared` as @djdameln suggested. Let me know if you guys are okay with this?

djdameln (approved):

samet-akcay:
> I would personally prefer to have the contents of src under anomaly_classification, like we had initially. But this is because I feel that the difference between the segmentation task and the classification task will be only in the configuration files, and the tasks will use a shared task implementation.

I think the difference would be in the dataset and the inference. We might need the ground-truth segmentation annotation to compute the performance on the validation set to find the adaptive threshold (we may not need at all, not sure). If yes, we might need a segmentation dataset class.  Of course, we could also use some checks such as `if task==classification; do this else that` in a single dataset class. 

For the inference, we will need to create segmentation mask and annotation using the ote-sdk. Therefore, we'll need to modify the inference callback. either a `SegmentationInferenceCallback` or some if checks in the existing callback.

> In that case maybe the folder could be called shared or core to indicate that it contains the core modules that are needed by both the anomalib task implementations?

not sure about `shared` but I could vote for `core`. I agree they are shared,  and `shared` sounds like it contains material that is shared between classification and segmentation. However, this is mainly because they are the source code wrappers for ote. So I'd prefer a specific name, `src` or `core`. 

> Also I'm not sure if we should have the ote_sdk stuff in the same folder as the ote_cli stuff, because they serve quite different purposes and probably won't rely much on the same files.

Yeah, their purpose is different, but our source code could provide the sufficient wrappers for both `ote_cli` and `ote_sdk`. For instance, with a couple of minor tweaks, I think our test helpers could work nicely with `ote_cli` to support ote training.

samet-akcay (approved):
thanks a lot for the effort

PR #665, opened 2021-11-03
add anomaly heatmap to dataset item as metadata

djdameln:
Computes the anomaly heatmaps in the inference callback and adds it to the dataset item as a metadata.

### DIFF
--- a/external/anomaly/ote/callbacks/inference.py
+++ b/external/anomaly/ote/callbacks/inference.py
@@ -63,6 +63,6 @@ class InferenceCallback(Callback):
 
             heatmap = anomaly_map_to_color_map(anomaly_map.squeeze())
             heatmap_media = ResultMediaEntity(
-                name="Anomaly map", type="anomaly_map", annotation_scene=dataset_item.annotation_scene, numpy=heatmap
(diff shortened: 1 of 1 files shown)


external/anomaly/ote/callbacks/inference.py
```diff
@@ -56,3 +60,9 @@ def on_predict_epoch_end(self, _trainer: pl.Trainer, _pl_module: AnomalyModule,
             )
 
             dataset_item.append_annotations([shape])
+
+            heatmap = anomaly_map_to_color_map(anomaly_map.squeeze())
+            heatmap_media = ResultMediaEntity(
+                name="Anomaly map", type="anomaly_map", annotation_scene=dataset_item.annotation_scene, numpy=heatmap
```
samet-akcay, line 66: maybe we could have the name as `name="Anomaly Map"` ?

samet-akcay (commented):
Only a minor comment. Looks good otherwise.

ashwinvaidya17:
Looks good to me. Thanks!

ashwinvaidya17 (approved):

samet-akcay (approved):
Thanks @djdameln, looks good!

PR #674, opened 2021-11-09
Minor Refactor

ashwinvaidya17:
# Changes

- isort imports
- Fix OpenVINO naming
- Remove inference parameters from `template.yaml`
# Coverage

image

### DIFF
--- a/external/anomalib/anomaly_classification/openvino.py
+++ b/external/anomalib/anomaly_classification/openvino.py
@@ -22,7 +22,6 @@ import tempfile
 from typing import Optional, Union
 
 from addict import Dict as ADDict
-from anomalib.core.model.inference import OpenVINOInferencer
 from compression.api import DataLoader
 from compression.engines.ie_engine import IEEngine
 from compression.graph import load_model, save_model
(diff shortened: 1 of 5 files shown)


external/anomalib/anomaly_classification/openvino.py
```diff
@@ -22,6 +22,7 @@
 from typing import Optional, Union
 
 from addict import Dict as ADDict
+from anomalib.core.model.inference import OpenVinoInferencer
```
samet-akcay, line 25: Should we rename this in `anomalib` to match the new naming convention? Maybe  I could quickly add this to my existing config [MR](https://gitlab-icv.inn.intel.com/algo_rnd_team/anomaly/-/merge_requests/143)

ashwinvaidya17: That would be nice. Plus I guess this convention might also be something to do with the trademark.

samet-akcay: in terms of trademark, I agree. It is just against our camel case naming convention, but should be alright

ashwinvaidya17: Yeah I agree. It breaks our coding convention.

djdameln (approved):
Looks good to me!

samet-akcay (approved):

PR #675, opened 2021-11-09
using and updating model status, add model monitor tests

djdameln:
Adds handling of `model.model_status` in the model monitor callback. Also adds a test case to verify that the model monitor works as expected.


ashwinvaidya17 (approved):

external/anomalib/core/callbacks/model_monitor.py
```diff
@@ -59,7 +59,7 @@ def on_fit_end(self, trainer: pl.Trainer, pl_module: pl.LightningModule) -> None
         """
         new_performance = pl_module.results.performance["image_f1_score"]
         is_improved = self.old_performance < new_performance
-        is_first_training = self.output_model.performance == NullPerformance()
+        is_first_training = self.output_model.model_status == ModelStatus.NOT_READY
```
druzhkov-paul, line 62: I'm still not sure that this is a safe assumption. Are you sure that this is fine?

external/anomalib/core/callbacks/model_monitor.py
```diff
@@ -68,6 +68,9 @@ def on_fit_end(self, trainer: pl.Trainer, pl_module: pl.LightningModule) -> None
             performance = Performance(score=ScoreMetric(name="F1 Score", value=new_performance))
             self.save_model_fn(self.output_model)
             self.output_model.performance = performance
+            self.output_model.precision = ModelPrecision.FP32
```
druzhkov-paul, line 71: This should be a list.

druzhkov-paul (changes requested):

external/anomalib/core/callbacks/model_monitor.py
samet-akcay: @druzhkov-paul do you have any suggestions here?

I'm checking out `ModelEntitity` in [`ote_sdk`](https://gitlab.devtools.intel.com/vmc-eip/IMPT/impt/-/blob/develop/src/ote_sdk/ote_sdk/entities/model.py#L120), where the default value of `model_status` is set to `ModelStatus.SUCCESS`. `performance` is also set to `None`. When we create our tests to train ote models we set `model_status` in `output_model` as `ModelStatus.NOT_READY`. This was @djdameln 's motivation.

There are also `previous_revision` and `previous_trained_revision` for this purpose, but maybe you could hint your approach in detection task.

djdameln: You're right. Using `model_status` here won't work, because the status is initialized as `SUCCESS`. Checking for `NullPerformance` however should work. When the `ModelEntity` is instantiated, a NullPerformance object is assigned to the `performance` attribute. I reverted this line to use the `NullPerformance` check.

druzhkov-paul: AFAIK, there are no explicit restrictions on the state of the `output_model` as it is passed to `Task.train`. So, for example, I'd expect that `model_status` is always `NOT_READY` in the begging of every training round and set to `SUCCESS` inside the task (if it is really so, for sure) as it is finished.

Good news here is that we've just agreed that the decision, whether the model is improved or not, should *not* be the OTE responsibility. This being said, OTE Task should now always put some weights to the `output_model` during training (that will make it suitable for new Task initialization) and set status to `SUCCESS`.

samet-akcay: yeah, good to hear that. I also think that the default value of `model_status` should be `NOT_READY`. Setting the default value to `SUCCESS` doesn't make sense and would be similar to setting `performance` to a high score.

samet-akcay: @druzhkov-paul, are you happy with this? If yes, we'll merge this change to the actual PR

druzhkov-paul: Not sure that it resolves the comment. My point, is that we should not rely on the content of the `output_model` at all -- it's just a placeholder for the output.

djdameln: It was my understanding that the `train` method of the task receives the output model of the previous training iteration, provided that the task has been trained before, and an empty model otherwise. If that is the case, then this check should work. 

If instead the `train` method will always receive an empty output model, then this check won't work and we need to use the model from the task environment. Though I would not see the point of passing an output model to the `train` method if it will always be an empty model entity. In that case we might just as well instantiate it within the task.

I would be happy to do the check based on the model from the task environment, but the problem is that we can't directly access the task environment from this callback unless we pass it to the constructor. That's why I would prefer to do the check based on the output model, if possible.

druzhkov-paul: > It was my understanding that the train method of the task receives the output model of the previous training iteration.

It may be like this. May not. It's just outside of the OTE's responsibility and not well-specified. This being said, the task itself already has an up-to-date model, so no need to receive it once again.

> Though I would not see the point of passing an output model to the train method if it will always be an empty model entity.

Output model is a placeholder for the actual model weights and attributes. And there a couple of benefits when it's created outside of the task. First, it may be something more than `ModelEntity` (some class derived from it). Second, it makes it possible to store intermediate checkpoints to restore the training after some external failure. To do that you just put data inside the output model and call the `save_model` callback from training parameters.

> That's why I would prefer to do the check based on the output model, if possible.

I can suggest the same thing as in the comments above --- just don't care about model improvement here, return the best you have all the time and that's it.

druzhkov-paul (approved):

samet-akcay (approved):

PR #691, opened 2021-11-11
addressing the comments left from the previous commits

samet-akcay:


samet-akcay:
could merge it myself since it's a super simple one.

PR #703, opened 2021-11-16
Address threshold pr comment

ashwinvaidya17:
# Changes
- Add check for `task_environment.model`

# Note
This is the simplest solution I could come up with.  One option can be to add the model's default threshold defined in the config to the task_environment's hyperparameters (in `get_config` method). But this would lead to the same issue of it not being always defined in the task environment.


samet-akcay:
Not having a model and running an inference on openvino infer method would be an edge-case. If there is no model and we try to infer, there are more fundamental issues than threshold we would need to consider.

samet-akcay (approved):

PR #705, opened 2021-11-17
Updates on chest_xray_screening model

Rakshith2597:
Made following updates on the [chest_xray_screening](https://github.com/openvinotoolkit/training_extensions/tree/master/misc/pytorch_toolkit/chest_xray_screening) model.

1. Updated ReadMe file
2. Replaced Google drive links (for model weights, sample data, etc).
3. Modified config files

### DIFF
--- a/misc/pytorch_toolkit/chest_xray_screening/chest_xray_screening/utils/download_weights.py
+++ b/misc/pytorch_toolkit/chest_xray_screening/chest_xray_screening/utils/download_weights.py
@@ -1,16 +1,12 @@
 from .get_config import get_config
 import os
 import zipfile
-import pycurl
+import subprocess
 
 
 def download_and_extract(path, url, expath):
-    with open(path, 'wb') as f:
-        cl = pycurl.Curl()
-        cl.setopt(cl.URL, url)
-        cl.setopt(cl.WRITEDATA, f)
-        cl.perform()
-        cl.close()
+    export_command = f'curl {url} --output {path}'
+    subprocess.run(export_command, shell = True, check = True)
     with zipfile.ZipFile(path, 'r') as zip_ref:
         zip_ref.extractall(expath)
 
--- a/misc/pytorch_toolkit/chest_xray_screening/init_venv.sh
+++ b/misc/pytorch_toolkit/chest_xray_screening/init_venv.sh
@@ -25,7 +25,7 @@ fi
 
 . ${venv_dir}/bin/activate
 
-apt install libcurl4-openssl-dev libssl-dev
+
 cat requirements.txt | xargs -n 1 -L 1 pip3 install
 
 pip install -e .
--- a/misc/pytorch_toolkit/chest_xray_screening/requirements.txt
+++ b/misc/pytorch_toolkit/chest_xray_screening/requirements.txt
@@ -3,7 +3,6 @@ torchvision==0.6.1
 pydicom
 scikit-learn
 tqdm
-pycurl
 pandas
 requests==2.26.0
 openvino-dev[onnx]


misc/pytorch_toolkit/chest_xray_screening/chest_xray_screening/utils/download_weights.py
```diff
@@ -1,17 +1,29 @@
...
-                                    dest_path=config['densenet_eff']['dest_path'],
-                                    unzip=False)
+    config = get_config(action = 'download')
+    if not os.path.exists('model_weights'):
+        os.makedirs('model_weights')
+    densenet_url = config['densenet']['url']
+    densenet_path = config['densenet']['dest_path']
+    denseneteff_url = config['densenet_eff']['url']
+    denseneteff_path = config['densenet_eff']['dest_path']
+    os.system(f'wget -mnv {densenet_url} -O {densenet_path}')
```
morkovka1337, line 13: Please, do not use os.system, this is not safe. Use special python libraries instead - wget (which is already in your requirements.txt, requests or urllib.

morkovka1337 (changes requested):

misc/pytorch_toolkit/chest_xray_screening/chest_xray_screening/utils/download_weights.py
Rakshith2597: Okay, I Will make the change. I migrated from python wget to os.system as the latter was used [here](https://github.com/openvinotoolkit/training_extensions/blob/962b324d55595fa6be1ac285468933c706a58a0c/misc/pytorch_toolkit/text_recognition/text_recognition/utils/common.py#L30) in the text_recognition model.

morkovka1337: Well, thank you for this, I will need to change it also.

Rakshith2597: Replaced `os.system` with pycurl. Python wget sometimes downloads an HTML file instead of the actual file.

misc/pytorch_toolkit/chest_xray_screening/chest_xray_screening/utils/download_weights.py
```diff
@@ -1,17 +1,47 @@
...
+    denseneteff_path = config['densenet_eff']['dest_path']
+
+    with open(densenet_path, 'wb') as f:
+        cl = pycurl.Curl()
+        cl.setopt(cl.URL, densenet_url)
+        cl.setopt(cl.WRITEDATA, f)
+        cl.perform()
+        cl.close()
+    with zipfile.ZipFile(config['densenet']['dest_path'], 'r') as zip_ref:
+        zip_ref.extractall('model_weights/')
```
morkovka1337, line 22: For here and on, this action could be extracted in the separate function that downloads file from a given url to a given path and then extracts it. Something like `download_and_extract(url, path)`.

morkovka1337 (changes requested):

morkovka1337:
run tests

morkovka1337:
Seems like in order to use pycurl on ubuntu, some libs need to be installed additionally. https://stackoverflow.com/questions/23937933/could-not-run-curl-config-errno-2-no-such-file-or-directory-when-installing

So you can try to add `apt install libcurl4-openssl-dev libssl-dev` into init_venv.sh in front of [line 29](https://github.com/openvinotoolkit/training_extensions/blob/master/misc/pytorch_toolkit/chest_xray_screening/init_venv.sh#L29), rerun tests and see if it works.

The other option is to use `urllib.request.urlretrieve(url, path)`. This package is built-in python package, i.e., it does not require any further installation.

morkovka1337:
run tests

morkovka1337:
> E: Unable to locate package libcurl4-openssl-dev

Okay, it will be easier and more stable to use urllib. Please, do required changes. Sorry for inconvenience.

Rakshith2597:
@morkovka1337 
wget and urllib.requests are downloading an HTML file instead of the actual file. The [link ](http://miriad.digital-health.one/models/densenet.zip) works perfectly fine when used via a browser (tested on Brave, Chrome and Edge), `curl` and `pycurl`. I'm not exactly sure about the reason but this seems to be a common problem. I have tried most of the solutions and hacks suggested in StackOverflow and StackExchange, but none seems to work. Is it possible to continue with pycurl itself? or else can I use `os.system(curl)` or `subprocess.run(curl)` (both works). If not, please suggest an alternative.

morkovka1337:
> @morkovka1337 wget and urllib.requests are downloading an HTML file instead of the actual file. The [link ](http://miriad.digital-health.one/models/densenet.zip) works perfectly fine when used via a browser (tested on Brave, Chrome and Edge), `curl` and `pycurl`. I'm not exactly sure about the reason but this seems to be a common problem. I have tried most of the solutions and hacks suggested in StackOverflow and StackExchange, but none seems to work. Is it possible to continue with pycurl itself? or else can I use `os.system(curl)` or `subprocess.run(curl)` (both works). If not, please suggest an alternative.

Understood. Although I have tried to use such code to download the file and it works perfectly for me:
```
import urllib.request
link = 'http://miriad.digital-health.one/models/densenet.zip'
urllib.request.urlretrieve(link,'archive.zip')
```

An alternative way is to debug the existing problem with the installation of the pycurl. In the end, if we fail to make this work, we can always use a workaround with the os.system. From my side I can ask CI guys to help with this problem.

morkovka1337:
> > ```
> > urllib.request.urlretrieve(link,'archive.zip')
> > ```
> 
> The command runs without any error but it stores an HTML file with the filename as 'archive.zip'.
> 
> > In the end, if we fail to make this work, we can always use a workaround with the os.system.
> 
> Both os.system(curl) and subprocess.run(curl) works. Since we are already using subprocess.run() for onnx to IR conversion [here](https://github.com/openvinotoolkit/training_extensions/blob/ff5a69f78d1dbfdf43af52200fac0b792ecf72da/misc/pytorch_toolkit/text_recognition/text_recognition/utils/exporter.py#L108) and it is considered to be relatively [safer](https://stackoverflow.com/questions/44730935/advantages-of-subprocess-over-os-system) maybe we can use that. Even the [documentation](https://docs.python.org/3/library/os.html#os.system) of os.system states _using subprocess is preferable to using this function_.

Ok, agree, using of `subprocess.run()` seems reasonable.

Rakshith2597:
> An alternative way is to debug the existing problem with the installation of the pycurl. In the end, if we fail to make this work, we can always use a workaround with the os.system. From my side I can ask CI guys to help with this problem.

@morkovka1337 Any updates on this? Or shall I proceed with `subprocess.run()`?

morkovka1337:
> > An alternative way is to debug the existing problem with the installation of the pycurl. In the end, if we fail to make this work, we can always use a workaround with the os.system. From my side I can ask CI guys to help with this problem.
> 
> @morkovka1337 Any updates on this? Or shall I proceed with `subprocess.run()`?

Yes, please, use `subprocess.run()`.

Rakshith2597:
run tests

morkovka1337 (approved):

morkovka1337:
@Ilya-Krylov, please, merge

PR #728, opened 2021-11-23
OTE_SDK. Increase code coverage

saltykox:
Test list:
ote_sdk/ote_sdk/tests/entities/shapes/test_rectangle.py
ote_sdk/ote_sdk/tests/entities/shapes/test_shape.py

Task: CVS-71119

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ env
 .env
 
 data/*
+.coverage
 
 *egg-info
 
--- a/ote_sdk/ote_sdk/tests/entities/test_annotation.py
+++ b/ote_sdk/ote_sdk/tests/entities/test_annotation.py
@@ -1,0 +1,612 @@
+# Copyright (C) 2021 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions
+# and limitations under the License.
+
+import datetime
+from typing import List
+
+import pytest
+
+from ote_sdk.entities.annotation import (
+    Annotation,
+    AnnotationSceneEntity,
+    AnnotationSceneKind,
+    NullAnnotationSceneEntity,
+)
+from ote_sdk.entities.color import Color
+from ote_sdk.entities.id import ID
+from ote_sdk.entities.label import Domain, LabelEntity
+from ote_sdk.entities.scored_label import ScoredLabel
+from ote_sdk.entities.shapes.ellipse import Ellipse
+from ote_sdk.entities.shapes.polygon import Point, Polygon
+from ote_sdk.entities.shapes.rectangle import Rectangle
+from ote_sdk.tests.constants.ote_sdk_components import OteSdkComponent
+from ote_sdk.tests.constants.requirements import Requirements
+from ote_sdk.utils.time_utils import now
+
+
+@pytest.mark.components(OteSdkComponent.OTE_SDK)
+class TestAnnotation:
+
+    rectangle = Rectangle(x1=0.5, x2=1.0, y1=0.0, y2=0.5)
+    labels: List[ScoredLabel] = []
+    annotation = Annotation(shape=rectangle, labels=labels)
+
+    car = LabelEntity(
+        id=ID(123456789),
+        name="car",
+        domain=Domain.DETECTION,
+        color=Color(red=16, green=15, blue=56, alpha=255),
+        is_empty=True,
+    )
+    person = LabelEntity(
+        id=ID(987654321),
+        name="person",
+        domain=Domain.DETECTION,
+        color=Color(red=11, green=18, blue=38, alpha=200),
+        is_empty=False,
+    )
+    car_label = ScoredLabel(car)
+    person_label = ScoredLabel(person)
+    labels2 = [car_label, person_label]
+
+    @pytest.mark.priority_medium
+    @pytest.mark.component
+    @pytest.mark.reqids(Requirements.REQ_1)
+    def test_annotation_default_property(self):
+        """
+        <b>Description:</b>
+        Check that Annotation can correctly return default property value
+
+        <b>Input data:</b>
+        Annotation class
+
+        <b>Expected results:</b>
+        Test passes if the Annotation return correct values
+
+        <b>Steps</b>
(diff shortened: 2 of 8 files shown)


saltykox:
run ote_sdk tests

saltykox:
run ote_sdk tests

saltykox:
run ote_sdk tests

saltykox:
run ote_sdk tests

ote_sdk/ote_sdk/tests/entities/shapes/test_rectangle.py
```diff
@@ -0,0 +1,667 @@
...
+        return Rectangle(**self.horizontal_rectangle_params())
+
+    @staticmethod
+    def vertical_rectangle_params() -> dict:
+        return {
+            "x1": 0.1,
+            "y1": 0.1,
+            "x2": 0.3,
+            "y2": 0.4,
+            "labels": ["label_1", "label_2"],
```
Ilya-Krylov: labels here are ScoredLabel

ote_sdk/ote_sdk/tests/entities/shapes/test_rectangle.py
```diff
@@ -0,0 +1,667 @@
...
+
+        <b>Steps</b>
+        1. Compare default label Rectangle instance attribute with expected value
+        2. Compare default modification_date Rectangle instance attribute with expected value
+        3. Compare specified label Rectangle instance attribute with expected value
+        4. Compare specified modification_date Rectangle instance attribute with expected value
+        """
+        # check for default values of optional parameters
+        default_params_rectangle = self.horizontal_rectangle()
+        expected_default_params_rectangle_modification_date = now().replace(
```
Ilya-Krylov: the same issue with `now`

Ilya-Krylov (changes requested):

ote_sdk/ote_sdk/tests/entities/shapes/test_rectangle.py
Ilya-Krylov: I guess that it can be false in some rare cases even if you replace microseconds by 0. 
Imagine that SECOND (but not MICROSECOND) was changed between two calls of `now`. One here and another inside ` self.horizontal_rectangle()`

I would call `now` only once, store it as class member and re-use where it is needed, without replacing microseconds.

saltykox:
run ote_sdk tests

ote_sdk/ote_sdk/tests/entities/shapes/test_rectangle.py
```diff
@@ -0,0 +1,673 @@
...
+        <b>Steps</b>
+        1. Compare default label Rectangle instance attribute with expected value
+        2. Compare default modification_date Rectangle instance attribute with expected value
+        3. Compare specified label Rectangle instance attribute with expected value
+        4. Compare specified modification_date Rectangle instance attribute with expected value
+        """
+        # Checking default values of optional parameters
+        default_params_rectangle = self.horizontal_rectangle()
+        assert default_params_rectangle._labels == []
+        assert default_params_rectangle.modification_date == self.now
```
Ilya-Krylov: it still might fail. call `now()` once, and use it for object initialization and then check equality

ote_sdk/ote_sdk/tests/entities/shapes/test_rectangle.py
saltykox: Thanks, replaced str-type labels to ScoredLabel objects

ote_sdk/ote_sdk/tests/entities/shapes/test_rectangle.py
saltykox: Thanks. In tests with default values for optional parameters, expected that if parameter modification_date equal “None” (default value) during Rectangle object initialization, then now() will be used to set date_time class attribute, based on the condition on line 63 of ote_sdk / ote_sdk / entities / shapes / rectangle.py module. In default_parameters tests it is expected that modification_date parameter will not be set manually, there are performance issues on Jenkins, so I used replace microseconds method for date_time object

saltykox: Could you advice if there is a better way than replacing microseconds and seconds to tests with not specified modification_date parameter for Rectangle instances? Perhaps, in such scenarios, it would be enough to check that modification_date attribute changed its type from None to datetime?

Ilya-Krylov: > Thanks. In tests with default values for optional parameters, expected that if parameter modification_date equal “None” (default value) during Rectangle object initialization, then now() will be used to set date_time class attribute, based on the condition on line 63 of ote_sdk / ote_sdk / entities / shapes / rectangle.py module. In default_parameters tests it is expected that modification_date parameter will not be set manually, there are performance issues on Jenkins, so I used replace microseconds method for date_time object

The problem is that fist call of `now()` can give us 10h59m59s999999ms, and second call can give us 11h00m00s000001ms, that's why setting microseconds to 0 does not work always.

Ilya-Krylov: > Could you advice if there is a better way than replacing microseconds and seconds to tests with not specified modification_date parameter for Rectangle instances? Perhaps, in such scenarios, it would be enough to check that modification_date attribute changed its type from None to datetime?

Yes, I think it would be enough to check `type` of `modification_date`

saltykox: Thanks, changed to checking default modification_date attribute “type” instead of “value” for Rectangle instances with non-specified modification_date parameter

ote_sdk/ote_sdk/tests/entities/shapes/test_rectangle.py
```diff
@@ -0,0 +1,672 @@
...
+                    "y2": 1.0,
+                    "labels": self.rectangle_labels(),
+                },
+                "params_expected": {"x1": 0.0, "y1": 0.0, "x2": 1.0, "y2": 1.0},
+            },
+        ]
+        for scenario in positive_scenarios:
+            rectangle_actual = Rectangle(**scenario.get("input_params"))
+            rectangle_expected = Rectangle(**scenario.get("params_expected"))
+            rectangle_actual.modification_date = self.modification_date
```
Ilya-Krylov, line 348: the same issue

saltykox: modification_date attribute belongs to class "TestRectangle" and is evaluated once, should I rename it?

Ilya-Krylov: sorry, I what thinking that there is `assert rectangle_actual.modification_date == self.modification_date `
Misread.
Nevermind.

saltykox:
run ote_sdk tests

Ilya-Krylov:
run ote_sdk tests

PR #737, opened 2021-11-25
Ik/label schema in snapshot

Ilya-Krylov:


ote_sdk/ote_sdk/serialization/id_mapper.py
```diff
@@ -0,0 +1,49 @@
...
+from ote_sdk.entities.id import ID
+
+
+class IDMapper:
+    """
+    This class maps an `ID` entity to a string, and vice versa
+    """
+
+    def forward(self, instance: ID) -> Union[ObjectId, str]:
+        instance = ID(instance)
```
druzhkov-paul, line 33: Input/output types look weird.

ote_sdk/ote_sdk/serialization/id_mapper.py
```diff
@@ -0,0 +1,49 @@
...
+        # if len(instance) in (12, 24):
+        #     return ObjectId(str(instance))
+        # if str(instance).isdigit():
+        #     # ID exists of digits only. Pad with zero's to a string of length 24 (e.g.19 -> "..0019")
+        #     id_int = int(str(instance))
+        #     return ObjectId("{:024d}".format(id_int))
+
+        if len(instance) != 0:
+            logger = logging.getLogger(__name__)
```
druzhkov-paul, line 42: I believe this is not the best way to get the logger. Each call will add another handler to the list and logs will be duplicated once and once again.

ote_sdk/ote_sdk/serialization/label_mapper.py
```diff
@@ -0,0 +1,209 @@
...
+        )
+
+
+class LabelMapper:
+    """
+    This class maps a `Label` entity to a serialized dictionary, and vice versa
+    """
+
+    def __init__(self) -> None:
+        pass
```
druzhkov-paul, line 60: Hm...

ote_sdk/ote_sdk/serialization/label_mapper.py
```diff
@@ -0,0 +1,209 @@
...
+        }
+        for label in label_map.values():
+            output.add_node(label)
+        for edge in instance["edges"]:
+            output.add_edge(label_map[edge[0]], label_map[edge[1]])
+
+        return output
+
+
+class LabelSchemaMapper:
```
druzhkov-paul, line 161: Is it possible to directly pickle a `LabelSchema`?

druzhkov-paul (commented):
In general it looks good.

I like having model's weights and the label schema being packed together. But, at the same time this design puts some constraints on the format and content of a model snapshot, which is otherwise only defined at a particular task level.

On the other hand, we can also consider an option of storing label schema separately. It will resemble the way it is defined in the SDK. And once the label schema is created and saved to file it may be used for all CLI tools. But it will increase the probability that inconsistent label schema and model weights are passed to a script.

What do you think?

ote_sdk/ote_sdk/serialization/label_mapper.py
Ilya-Krylov: 🤦

Ilya-Krylov:
> In general it looks good.
> 
> I like having model's weights and the label schema being packed together. But, at the same time this design puts some constraints on the format and content of a model snapshot, which is otherwise only defined at a particular task level.
> 
> On the other hand, we can also consider an option of storing label schema separately. It will resemble the way it is defined in > the SDK. And once the label schema is created and saved to file it may be used for all CLI tools. But it will increase the > probability that inconsistent label schema and model weights are passed to a script.
> 
> What do you think?

I don't like idea to store labels schema separately from checkpoint. I prefer either :
* put constraints on task save_model implementation so that resulting snapshot contains following keys: "label_schema", "model_data" (arbitrary, task-specific data), ["version"], 
* do this higher level snapshot on SDK side, e.g. `model.get_data("weights.pth")` does this creation of final (torch independent) serialization by combining task serialized data and serialization of ModelConfiguration (contains LabelSchema, or serialize just LabelSchema). The `get_data` is not proper method for sure for doing this, that was said to demonstrate the idea. Design of a Model should be changed in order to do this in natural way.

As a (probably) temporal solution I would put these (not so strict) constraints to make OTE CLI work without explicitly passed labels and then discuss with folks changing SDK and its use on another side.

ote_cli/ote_cli/utils/loading.py
```diff
@@ -27,3 +32,30 @@ def load_model_weights(path):
...
+    """
+    Reads serialized representation from binary snapshot and returns deserialized LabelSchema.
+    """
+    
+    import torch
+    return torch.load(io.BytesIO(model_bytes))["label_schema"]
+    
+
+def generate_label_schema(dataset, task_type):
+    if task_type == TaskType.CLASSIFICATION and dataset.is_multilabel():
```
druzhkov-paul, line 47: Is data here of type `ImageClassificationDataset`?

ote_cli/ote_cli/utils/loading.py
```diff
@@ -27,3 +32,30 @@ def load_model_weights(path):
...
+        not_empty_labels = dataset.get_labels()
+        assert len(not_empty_labels) > 1
+        label_schema = LabelSchemaEntity()
+        emptylabel = LabelEntity(name="Empty label", is_empty=True, domain=Domain.CLASSIFICATION)
+        empty_group = LabelGroup(name="empty", labels=[emptylabel], group_type=LabelGroupType.EMPTY_LABEL)
```
druzhkov-paul, line 52: `emptylabel` `empty_group`

druzhkov-paul (approved):

Ilya-Krylov:
run ote_sdk tests

ote_sdk/ote_sdk/serialization/id_mapper.py
Ilya-Krylov: not relevant anymore

ote_sdk/ote_sdk/serialization/id_mapper.py
Ilya-Krylov: not relevant anymore

ote_sdk/ote_sdk/serialization/label_mapper.py
Ilya-Krylov: not relevant anymore

ote_sdk/ote_sdk/serialization/label_mapper.py
Ilya-Krylov: not relevant anymore

Ilya-Krylov:
run ote_sdk tests

Ilya-Krylov:
run ote_sdk tests

Ilya-Krylov:
BMM 20077 🕐

druzhkov-paul (approved):

PR #747, opened 2021-12-01
Add anomaly dataset

ashwinvaidya17:
# Changes
- Add dummy anomaly dataset to repo


data/anomaly/train.json
```diff
@@ -0,0 +1,70 @@
...
+        "16": "train\/good\/012.png",
+        "17": "train\/good\/005.png",
+        "18": "train\/good\/018.png",
+        "19": "train\/good\/010.png",
+        "20": "train\/good\/003.png",
+        "21": "train\/good\/009.png",
+        "22": "train\/good\/000.png",
+        "23": "train\/good\/016.png",
+        "24": "train\/good\/001.png",
+        "25": "test\/star\/006.png",
```
djdameln, line 27: Not sure I understand the format of these annotation files. Why is there `star` category in the training set, instead of only images from the `train/good` folder?

ashwinvaidya17: Yeah we don't need it. Since we have `val.json` as well, I can distribute these between `test` and `val`.

PR #748, opened 2021-12-01
Add anomaly dataset

ashwinvaidya17:
# Changes
- Add anomaly dataset sample
- Fix json files from previous PR


samet-akcay (approved):

djdameln (dismissed):

data/anomaly/test.json
```diff
@@ -0,0 +1,42 @@
+{
```
Ilya-Krylov, line 1: it is required to store information about masks images.

data/anomaly/val.json
```diff
@@ -0,0 +1,40 @@
+{
```
Ilya-Krylov, line 1: is there any reason to keep both val.json and test.json if they are the same?

Ilya-Krylov (approved):

PR #772, opened 2021-12-13
Fix segmentation results processing

druzhkov-paul:
Build 21388 ✔️ 
Tests 21452 ✔️

### DIFF
--- a/ote_sdk/ote_sdk/entities/shapes/rectangle.py
+++ b/ote_sdk/ote_sdk/entities/shapes/rectangle.py
@@ -73,7 +73,7 @@ class Rectangle(Shape):
             is_valid = is_valid and self._validate_coordinates(x, y)
         if not is_valid:
             warnings.warn(
-                f"Invalid {type(self).__name__} : x1={x1}, y1={y1}, x2={x2}, y2={y2}",
+                f"{type(self).__name__} coordinates are invalid : x1={x1}, y1={y1}, x2={x2}, y2={y2}",
                 UserWarning,
             )
 
--- a/ote_sdk/ote_sdk/entities/shapes/rectangle.py
+++ b/ote_sdk/ote_sdk/entities/shapes/rectangle.py
@@ -18,6 +18,7 @@
 
 import datetime
 import math
+import warnings
 from typing import List, Optional
 
 import numpy as np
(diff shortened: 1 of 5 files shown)


Ilya-Krylov:
run ote_sdk tests

ote_sdk/ote_sdk/entities/shapes/rectangle.py
```diff
@@ -67,8 +68,14 @@ def __init__(
             modification_date=modification_date,
         )
 
+        is_valid = True
         for (x, y) in [(x1, y1), (x2, y2)]:
-            self._validate_coordinates(x, y)
+            is_valid = is_valid and self._validate_coordinates(x, y)
+        if not is_valid:
+            warnings.warn(
+                f"Invalid {type(self).__name__} : x1={x1}, y1={y1}, x2={x2}, y2={y2}",
```
Ilya-Krylov: Please suppress warnings in tests OR change warning message to start from `f"{type(self).__name__} coordinates`

druzhkov-paul:
run ote_sdk tests

ote_sdk/ote_sdk/utils/segmentation_utils.py
```diff
@@ -184,27 +230,34 @@ def create_annotation_from_segmentation_map(
         # Contour retrieval mode CCOMP (Connected components) creates a two-level
         # hierarchy of contours
         contours, hierarchies = cv2.findContours(
-            label_index_map, cv2.RETR_CCOMP, cv2.CHAIN_APPROX_SIMPLE
+            label_index_map, cv2.RETR_CCOMP, cv2.CHAIN_APPROX_NONE
```
evgeny-izutov, line 233: Do we really need to disable the approximation? Seems it doesn't break the logic of subcontour selections but speeds up inference.

ote_sdk/ote_sdk/utils/segmentation_utils.py
```diff
@@ -142,6 +143,51 @@ def create_hard_prediction_from_soft_prediction(
...
 
+Contour = List[Tuple[float, float]]
+
+
+def get_subcontours(contour: Contour) -> List[Contour]:
+    """
+    Splits contour into subcontours that do not have self intersections.
+    """
+
+    ContourInternal = List[Optional[Tuple[float, float]]]
```
evgeny-izutov, line 154: Why do you need to convert it to float and then compare instead of comparing int values?

Ilya-Krylov:
@druzhkov-paul,  Will not it be too difficult for you to add several tests for `get_subcontours` with maybe typical invalid contours and a test that would fail without these changes if the test existed before.

ote_sdk/ote_sdk/utils/segmentation_utils.py
druzhkov-paul: This is just type annotation. It has no effect on the commands that are executed at runtime, i.e. no conversion is happening. It just says mypy then float and its subclasses are acceptable here. And `int` is a subclass of `float`. [PEP-484](https://www.python.org/dev/peps/pep-0484/#the-numeric-tower)

evgeny-izutov (approved):

PR #792, opened 2021-12-17
[ANOMALY CLASSIFICATION] Set threshold for output model

ljcornel:
In the anomaly classification task.py and openvino.py, the 'threshold' value was not set correctly for the output models generated after optimization. This caused inference for optimized models to fail. This PR makes sure that the threshold is set correctly on all output models.


djdameln (approved):

ljcornel:
run ote_sdk tests

samet-akcay (approved):
Thanks @ljcornel

PR #802, opened 2021-12-21
[OTE_SDK] Update visualizing probability for labels

akorobeinikov:
These changes is needed for anomaly models, because these return anomaly map with probability for each pixel in range [0, 100].


Ilya-Krylov:
I won't accept this change.

What if you have one prediction with label.probability = 0.99 and another 1.01 ? 
Will it be drawn as 99% and 1% correspondingly?

The fact that label.probability for anomaly is not in range of 0..1 it is not problem of OTE SDK. This is a bug of anomaly.

samet-akcay:
This may not be needed as there is already a [PR](https://github.com/openvinotoolkit/anomalib/pull/35) for this confidence score fix?

PR #828, opened 2021-12-29
[ANOMALY_CLASSIFICATION] Modify Exportable Code

ashwinvaidya17:
# Changes

- Add normalization support to exportable code.

### DIFF
--- a/external/anomaly/ote_anomalib/exportable_code/anomaly_classification.py
+++ b/external/anomaly/ote_anomalib/exportable_code/anomaly_classification.py
@@ -35,10 +35,10 @@ class AnomalyClassification(SegmentationModel):
         parameters["resize_type"].update_default_value("standard")
         parameters.update(
             {
-                "image_threshold": NumericalValue(default_value=0.5, description="Threshold value to locate anomaly"),
+                "image_threshold": NumericalValue(description="Threshold value to locate anomaly"),
                 "pixel_threshold": NumericalValue(description="Pixel Threshold value to locate anomaly"),
                 "pixel_mean": ListValue(description="Pixel mean used to normalize"),
-                "pixel_std": ListValue(description="Pixel standard deviation value to locate anomaly"),
+                "pixel_std": ListValue(description="Pixel standard deviation value used to normalize"),
                 "image_mean": NumericalValue(description="Image mean value used to normalize"),
                 "image_std": NumericalValue(description="Image standard deviation used to normalize"),
             }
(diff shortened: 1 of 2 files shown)


external/anomaly/ote_anomalib/exportable_code/anomaly_classification.py
```diff
@@ -35,6 +36,11 @@ def parameters(cls):
         parameters.update(
             {
                 "image_threshold": NumericalValue(default_value=0.5, description="Threshold value to locate anomaly"),
```
djdameln: Can we remove the `default_value` here, like for the other stats?

external/anomaly/ote_anomalib/exportable_code/anomaly_classification.py
```diff
@@ -50,10 +56,31 @@ def postprocess(self, outputs: Dict[str, np.ndarray], meta: Dict[str, Any]) -> n
...
+        meta["pixel_threshold"] = self.pixel_threshold  # pylint: disable=no-member
+        meta["image_mean"] = self.image_mean  # pylint: disable=no-member
+        meta["image_std"] = self.image_std  # pylint: disable=no-member
+
+        # standardize pixel scores
+        if "pixel_mean" in meta.keys() and "pixel_std" in meta.keys():
+            anomaly_map = np.log(anomaly_map)
+            anomaly_map = (anomaly_map - meta["pixel_mean"]) / meta["pixel_std"]
+            anomaly_map -= (meta["image_mean"] - meta["pixel_mean"]) / meta["pixel_std"]
+            anomaly_map = norm.cdf(anomaly_map - meta["pixel_threshold"])
```
samet-akcay, line 74: It would also be an idea to create a function in `anomalib` to perform this operation. This function could then be used both here and in ScoreNormalizationCallback in `anomalib`. I guess it's ok for now, but would be good practice in general since we need to modify things multiple times nowadays.

ashwinvaidya17: I agree. My only concern is the torch and numpy inputs. Maybe we can look into a design which adds normalization to forward method during onnx conversion. This way we will directly get normalized anomaly maps without needing to export all the thresholds as well.

djdameln: We'll need separate functions for torch and numpy. I don't think including it in the forward method is a good idea. With the current design we can easily disable the normalization and switch between different normalization methods. Also the idea is to keep the forward method as clean as possible and only use it for the model's core functionality. Including the post-processing steps in the forward method will step away from this design and could lead to problems down the road, like we've seen with tiling.

samet-akcay: yeah. Plus, we already have this in a normalization callback, which limits the use of this in a forward method. 

What I meant though is we could still wrap this in a function to use it in both the callback and here. I'm currently looking at the implementation. Looks like the only difference is `torch.log` vs `np.log`. We could have an argument to properly check and assign this.

PR #832, opened 2021-12-31
[ANOMALY CLASSIFICATION] Fix exportable code

ashwinvaidya17:
# Changes

- Add anomaly masks to OpenVINO inferencer
- Correct predicted score label.

### DIFF
--- a/ote_sdk/ote_sdk/usecases/exportable_code/prediction_to_annotation_converter.py
+++ b/ote_sdk/ote_sdk/usecases/exportable_code/prediction_to_annotation_converter.py
@@ -274,13 +274,17 @@ class AnomalyClassificationToAnnotationConverter(IPredictionToAnnotationConverte
     def convert_to_annotation(
         self, predictions: float, metadata: Dict[str, Any]
     ) -> AnnotationSceneEntity:
-        pred_label = predictions >= 0.5
+        pred_label = metadata["pred_score"] >= 0.5
         assigned_label = self.anomalous_label if pred_label else self.normal_label
 
         annotations = [
             Annotation(
                 Rectangle.generate_full_box(),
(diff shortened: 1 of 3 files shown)


ote_sdk/ote_sdk/usecases/exportable_code/prediction_to_annotation_converter.py
```diff
@@ -274,13 +274,17 @@ def __init__(self, label_schema: LabelSchemaEntity):
     def convert_to_annotation(
         self, predictions: float, metadata: Dict[str, Any]
```
ashwinvaidya17, line 275: predictions is not used but is kept to comply with the design

djdameln (approved):

ote_sdk/ote_sdk/usecases/exportable_code/prediction_to_annotation_converter.py
```diff
@@ -274,13 +274,17 @@ def __init__(self, label_schema: LabelSchemaEntity):
...
     ) -> AnnotationSceneEntity:
-        pred_label = predictions >= 0.5
+        pred_label = metadata["pred_score"] >= 0.5
```
samet-akcay, line 277: Is this always the case, including min-max and cdf normalisations, maybe other implementations we might add in the future? We could always do the following without modifying this function:

```
predictions = metadata["pred"]
annotation = convert_to_annotation(predictions, metadata)
```

Am I missing anything here?

ashwinvaidya17: We can remove the anomaly masks from the `postprocess` function or move that to metadata and return the scores as predictions. But we can't do anything in the between as it is called from SyncDemo 
image

PR #835, opened 2022-01-10
[ANOMALY CLASSIFICATION] Fix progress bar

ashwinvaidya17:
# Changes

- Remove val step from progress bar calculation as it resets the progress bar after each epoch. Since SC has a separate val step, the progress for val step is not affected

# Screenshots
image

image

# Builds

BMM - https://10.211.120.61/
VM - https://10.55.252.81/


samet-akcay:
run ote_sdk tests

samet-akcay (approved):

PR #842, opened 2022-01-12
[OTE_SDK_TESTS] Increase stability test_ellipse.py

saltykox:
Using pytest.approx method instead of rounding or direct comparison for calculated values of Ellipse-class object

Task: CVS-74513


saltykox:
run ote_sdk tests

PR #849, opened 2022-01-13
[OTE CLI] Add `if __name__ == "__main__"` to `ote-cli` tools

samet-akcay:
## Description
This PR adds the following to the ote-cli tools to be able to attach debuggers in IDEs

## Known Issue
When committing, mypy throw an error in `ote_cli/ote_cli/utils/hpo.py`, which is the reason why I added ignore to bypass it.


samet-akcay:
run ote_sdk tests

samet-akcay:
run ote_sdk tests

druzhkov-paul (approved):

PR #882, opened 2022-01-27
[OTE_SDK_TESTS] Added input parameters validation and tests

saltykox:
Added input parameters validations for OTE_SDK modules:
•	configuration/helper/create.py: “create” function
•	entities/annotation.py: initialization methods of Annotation and AnnotationSceneEntity classes
•	entities/dataset_item.py: initialization method of DatasetItemEntity class
•	entities/datasets.py: initialization method of DatasetEntity class
•	entities/image.py: initialization method of Image class
•	entities/label.py: initialization method of LabelEntity class
•	entities/label_schema.py: initialization and “from_labels” methods of LabelSchemaEntity class
•	entities/model.py: initialization method of ModelEntity class
•	entities/model_template.py: “parse_model_template” function
•	entities/resultset.py: initialization method of ResultSetEntity class
•	entities/scored_label.py: initialization method of ScoredLabel class
•	entities/shapes/rectangle.py: initialization method of Rectangle class
•	entities/task_environment.py: initialization method of TaskEnvironment class
Changed OTE SDK tests according to changes:
•	tests/entities/test_datasets.py: test_dataset_entity_append: removed scenario when appending DatasetItemEntity object with None-type “media” attribute
•	tests/entities/test_label_schema.py: test_label_schema_from_labels: added check that ValueError exception is raised when unexpected type object is specified as “label_groups” parameter for “from_labels” method
•	tests/entities/test_model_template.py: test_parse_model_template: added check that ValueError exception is raised when unexpected object is specified as “model_template_path” parameter for “parse_model_template” function
•	tests/entities/test_resultset.py: test_resultset_entity: changed input dataset and expected results
Added tests:
•	ote_sdk/ote_sdk/tests/parameters_validation/test_input_parameters_validation.py

Task: CVS-76584

### DIFF
--- a/ote_sdk/ote_sdk/entities/model.py
+++ b/ote_sdk/ote_sdk/entities/model.py
@@ -141,7 +141,6 @@ class ModelEntity:
                 (previous_revision, "previous_revision", ModelEntity),
                 (version, "version", int),
                 (tags, "tags", list),
-                (model_status, " model_status", ModelStatus),
                 (model_format, "model_format", ModelFormat),
                 (training_duration, "training_duration", (int, float)),
                 (model_adapters, "model_adapters", dict),
--- a/ote_sdk/ote_sdk/entities/model.py
+++ b/ote_sdk/ote_sdk/entities/model.py
@@ -126,11 +126,7 @@ class ModelEntity:
         check_parameter_str_class_name(
             parameter=train_dataset,
             parameter_name="train_dataset",
-            expected_class_names=[
-                "DatasetEntity",
-                "HpoDataset",
-                "ObjectDetectionDataset",
-            ],
+            expected_class_names=["DatasetEntity"],
         )
         check_parameter_type(
             parameter=configuration,
--- a/ote_sdk/ote_sdk/entities/model.py
+++ b/ote_sdk/ote_sdk/entities/model.py
@@ -126,7 +126,11 @@ class ModelEntity:
         check_parameter_str_class_name(
             parameter=train_dataset,
             parameter_name="train_dataset",
-            expected_class_name="DatasetEntity",
+            expected_class_names=[
+                "DatasetEntity",
+                "HpoDataset",
+                "ObjectDetectionDataset",
+            ],
         )
         check_parameter_type(
             parameter=configuration,
--- a/ote_sdk/ote_sdk/entities/model.py
+++ b/ote_sdk/ote_sdk/entities/model.py
@@ -19,6 +19,13 @@ from ote_sdk.usecases.adapters.model_adapter import (
     IDataSource,
     ModelAdapter,
 )
+from ote_sdk.utils.argument_checks import (
+    check_optional_parameters_type,
+    check_parameter_str_class_name,
+    check_parameter_type,
+    check_several_dictionaries_keys_values_type,
+    check_several_lists_elements_type,
+)
 from ote_sdk.utils.time_utils import now
 
 if TYPE_CHECKING:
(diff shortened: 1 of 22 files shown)


saltykox:
run ote_sdk tests

saltykox:
run ote_sdk tests

saltykox:
run ote_sdk tests

ote_sdk/ote_sdk/utils/argument_checks.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+def check_optional_parameters_type(parameter_name_expected_type: list):
+    """
+    Function checks if optional parameters exist and raises ValueError exception if one of them has unexpected type
+    :param parameter_name_expected_type: list with tuples that contain optional parameter, name for exception message
+    and expected type
+    """
+    for parameter, name, expected_type in parameter_name_expected_type:
+        if parameter:
```
LeonidBeynenson: Please, do not use `if parameter` here, since `parameter` may be a valid value that is converted to `False` boolean -- e.g. `0`.
Since `Optional[smthng]` in typing is an alias for `Union[smthng, None]`, it is better to use 
`if parameter is not None` 
instead of `if parameter`.
```suggestion
        if parameter is not None:
```

ote_sdk/ote_sdk/utils/argument_checks.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+def check_several_lists_elements_type(parameter_name_expected_type: list):
+    """
+    Function checks if parameters lists exist and raises ValueError exception if lists elements have unexpected type
+    :param parameter_name_expected_type: list with tuples that contain parameter with nested elements, name for
+    exception message and expected type
+    """
+    for parameter, name, expected_type in parameter_name_expected_type:
+        if parameter:
```
LeonidBeynenson: Also I suggest
```suggestion
        if parameter is not None:
```

ote_sdk/ote_sdk/utils/argument_checks.py
```diff
@@ -0,0 +1,206 @@
...
+            check_nested_elements_type(
+                iterable=parameter, parameter_name=name, expected_type=expected_type
+            )
+
+
+def check_parameter_str_class_name(parameter, parameter_name, expected_class_names):
+    """Function raises ValueError exception if string class name is not equal to expected"""
+    parameter_class_name = type(parameter).__name__
+    if (parameter_class_name in expected_class_names) is False:
```
LeonidBeynenson:
```suggestion
    if parameter_class_name not in expected_class_names:
```

ote_sdk/ote_sdk/utils/argument_checks.py
```diff
@@ -0,0 +1,206 @@
...
+    :param parameter_name_expected_type: list with tuples that contain dictionary parameter, name for exception message
+    and expected type
+    """
+    for (
+        parameter,
+        name,
+        expected_key_class,
+        expected_value_class,
+    ) in parameter_name_expected_type:
+        if parameter:
```
LeonidBeynenson:
```suggestion
        if parameter is not None:
```

ote_sdk/ote_sdk/utils/argument_checks.py
```diff
@@ -0,0 +1,206 @@
...
+    file_extension = file_path.split(".")[-1].lower()
+    if file_extension not in expected_extensions:
+        raise ValueError(
+            f"Unexpected extension of {file_path_name} file. expected: {expected_extensions} actual: {file_extension}"
+        )
+
+
+def check_that_null_character_absents_in_path(file_path: str, file_path_name: str):
+    """Function raises ValueError exception if null character: '\0' is specified in path to file"""
+    if "\\0" in file_path:
```
LeonidBeynenson: **Attention!** I think that there should be `"\0" in file_path` instead of `"\\0" in file_path`,
since `"\0"` is a string that consists of one char with code zero, whereas `"\\0"` is just a string that consists of two chars: `"\\"` and `"0"`
```suggestion
    if "\0" in file_path:
```

ote_sdk/ote_sdk/utils/argument_checks.py
LeonidBeynenson: BTW, why `if parameter` or `if parameter is not None` should be here?
I thought that the restriction should be fulfilled for all elements of the dicts.

ote_sdk/ote_sdk/utils/argument_checks.py
LeonidBeynenson: BTW, why if parameter or if parameter is not None should be here?
I thought that the restriction should be fulfilled for all elements of the list.

LeonidBeynenson (changes requested):
Please, make changes or write if some places should be kept without changes.

ote_sdk/ote_sdk/entities/model.py
```diff
@@ -115,6 +122,64 @@ def __init__(
         model_size_reduction: float = 0.0,
         _id: Optional[ID] = None,
     ):
+        # Initialization parameters validation
+        if TYPE_CHECKING:
+            check_parameter_str_class_name(
+                parameter=train_dataset,
+                parameter_name="train_dataset",
+                expected_class_names=["DatasetEntity", "HpoDataset", "ObjectDetectionDataset"],
+            )
```
LeonidBeynenson: 1. Why this check is inside branch `if TYPE_CHECKING`?

   Please, note that `typing.TYPE_CHECKING` is `True` if and only if the code is checked by mypy linter, it means that the code is NOT working indeed. So, this check will never work in real working code.

2. Also please note that this check compares the class name w.r.t. the strings that should not appear in the ote_sdk -- otherwise we will have to fix the checks in the base library (ote_sdk) when there are changes in names in some library that uses ote_sdk.

   I thought it could be sufficient to make `isinstance` with some base class -- e.g. `DatasetEntity` or some other class.
But `HpoDataset` uses its own trick -- it uses `__getattr__` method to forward call of attributes to its inner dataset field.
In this case it may require to use "duck typing" (see [article 1](https://en.wikipedia.org/wiki/Duck_typing), [article 2](https://ru.wikipedia.org/wiki/%D0%A3%D1%82%D0%B8%D0%BD%D0%B0%D1%8F_%D1%82%D0%B8%D0%BF%D0%B8%D0%B7%D0%B0%D1%86%D0%B8%D1%8F) )
So, you can create a function `is_like_dataset_entity` that checks an object using `hasattr` that it has such attributes as `__len__`, `__getitem__`, and `get_subset`.

ote_sdk/ote_sdk/utils/argument_checks.py
saltykox: Would it be more appropriate to check the type of the optional parameters as a tuple containing None and an suggested type? For example for "id" optional parameter it will be: isinstance(id, (None, ID))

saltykox: I added this check because the "parameter" variable is the optional and may be None-type object (ex. "tags" list-type parameter for ModelEntity class), nested list elements are checked in the "check_nested_elements_type" function

ote_sdk/ote_sdk/utils/argument_checks.py
saltykox: I added this check because the "parameter" variable is the optional and may be None-type object (ex. "model_adapters" dict-type parameter for ModelEntity class), nested dictionary keys and values are checked in the "check_dictionary_keys_values_type" function

ote_sdk/ote_sdk/entities/model.py
saltykox: Thanks, checking "train_dataset" parameter attributes instead of class name

ote_sdk/ote_sdk/utils/argument_checks.py
saltykox: Applied suggestion, thanks

ote_sdk/ote_sdk/utils/argument_checks.py
saltykox: Applied suggestion, thanks

ote_sdk/ote_sdk/utils/argument_checks.py
saltykox: This function is replaced by "check_is_parameter_like_dataset"

saltykox:
run ote_sdk tests

saltykox:
Due to add less commits in git submodules opened Pull Request https://github.com/openvinotoolkit/training_extensions/pull/890

PR #890, opened 2022-02-01
[OTE_SDK_TESTS] Added input parameters validation and tests. Part 1

saltykox:
Added input parameters validations for OTE_SDK modules:
• configuration/helper/create.py: “create” function
• entities/annotation.py: initialization methods of Annotation and AnnotationSceneEntity classes
• entities/dataset_item.py: initialization method of DatasetItemEntity class
• entities/datasets.py: initialization method of DatasetEntity class
• entities/image.py: initialization method of Image class
• entities/label.py: initialization method of LabelEntity class
• entities/label_schema.py: initialization and “from_labels” methods of LabelSchemaEntity class
• entities/model.py: initialization method of ModelEntity class
• entities/model_template.py: “parse_model_template” function
• entities/resultset.py: initialization method of ResultSetEntity class
• entities/scored_label.py: initialization method of ScoredLabel class
• entities/shapes/rectangle.py: initialization method of Rectangle class
• entities/task_environment.py: initialization method of TaskEnvironment class
Changed OTE SDK tests according to changes:
• tests/entities/test_datasets.py: test_dataset_entity_append: removed scenario when appending DatasetItemEntity object with None-type “media” attribute
• tests/entities/test_label_schema.py: test_label_schema_from_labels: added check that ValueError exception is raised when unexpected type object is specified as “label_groups” parameter for “from_labels” method
• tests/entities/test_model_template.py: test_parse_model_template: added check that ValueError exception is raised when unexpected object is specified as “model_template_path” parameter for “parse_model_template” function
• tests/entities/test_resultset.py: test_resultset_entity: changed input dataset and expected results
Added tests:
• ote_sdk/ote_sdk/tests/parameters_validation/test_input_parameters_validation.py
Updated mmdetection and mmsegmentation submodules to ote branches on March 2

Task: CVS-76584


saltykox:
run ote_sdk tests

saltykox:
run ote_sdk tests

ote_sdk/ote_sdk/utils/argument_checks.py
```diff
@@ -0,0 +1,217 @@
...
+        parameter_type = type(parameter)
+        raise ValueError(
+            f"Unexpected type of '{parameter_name}' parameter, expected: {expected_type}, actual: {parameter_type}"
+        )
+
+
+def check_parameter_type(parameter, parameter_name, expected_type):
+    """Function extracts nested expected types and raises ValueError exception if parameter has unexpected type"""
+    # pylint: disable=W0212
+    if isinstance(expected_type, typing._GenericAlias):  # type: ignore
```
LeonidBeynenson: 1. First of all, we have there a very big `if`, and one of branch of the `if` is much shorter than the other.
   In this case I really like use approach "quick handle simple branch and return", it allows to decrease indent level for most of the function, it forms the code as follows:
   ```python
   if not isinstance(expected_type, typing._GenericAlias):
         quick_handle_short_branch(...)
         return
   ```

2. As we spoke, we need special checks for paths (no null chars, no empty str, no non-printable characters). Formally, they are just `str`, but we could make special type of checks just here. We could make a special base class and special derived class for paths:
   ```python
   class BaseInputArgumentChecker(ABC):
       @abstractmethod
       def check(self, var):
           raise NotImplementedError('The check is not implemented')
   class PathChecker(BaseArgumentChecker):
       def check(self, var, var_name):
           raise_value_error_if_parameter_has_unexpected_type(var, ...., str)
           check_path_no_null_char(var, var_name)
           check_path_no_nonprintable_chars(var, var_name)
   ```
   After creating these classes in the top of this file you will be able to add branch in this function
   ```python
       if isinstance(expected_type, BaseInputArgumentChecker):
           expected_type.check(parameter, parameter_name)
           return
   ```
   After that you will have to use PathChecker everywhere where it is required. 🙂

ote_sdk/ote_sdk/utils/argument_checks.py
```diff
@@ -0,0 +1,217 @@
...
+                file_path=input_config, file_path_name=parameter_name
+            )
+    if isinstance(input_config, dict):
+        if input_config == {}:
+            raise ValueError(
+                "Empty dictionary is specified as 'input_config' parameter"
+            )
+
+
+def check_is_parameter_like_dataset(parameter, parameter_name):
```
LeonidBeynenson, line 304: Please, see the comment above on `BaseInputArgumentChecker` and path checks.
I think it could be better to make special `class DatasetParameterChecker(BaseInputArgumentChecker)` and use it.

LeonidBeynenson (changes requested):

ote_sdk/ote_sdk/utils/argument_checks.py
```diff
@@ -0,0 +1,217 @@
...
+    if isinstance(expected_type, typing._GenericAlias):  # type: ignore
+        origin_class = expected_type.__dict__.get("__origin__")
+        # Checking origin class
+        raise_value_error_if_parameter_has_unexpected_type(
+            parameter=parameter,
+            parameter_name=parameter_name,
+            expected_type=origin_class,
+        )
+        # Checking nested elements
+        if issubclass(origin_class, typing.Sequence):
```
LeonidBeynenson: The following change will allow to use `List` without parameters -- formally it is not error!
```suggestion
        args = expected_type.__dict__.get("__args__")
        if issubclass(origin_class, typing.Sequence) and args:
```
Also I would check that `len(args) == 1` inside the branch and raise `TypeError` if it is not true (to avoid `List[int, int]`)

ote_sdk/ote_sdk/utils/argument_checks.py
```diff
@@ -0,0 +1,217 @@
...
+            check_nested_elements_type(
+                iterable=parameter,
+                parameter_name=parameter_name,
+                expected_type=expected_type.__dict__.get("__args__")[0],
+            )
+        elif origin_class == dict:
```
LeonidBeynenson: The following change will allow to use Dict without parameters -- formally it is not error!
```suggestion
        elif origin_class == dict and args:
```
Also I would check that `len(args) == 2` inside the branch and raise TypeError if it is not true (to avoid `Dict[int]`)

ote_sdk/ote_sdk/utils/argument_checks.py
```diff
@@ -0,0 +1,217 @@
...
+
+from numpy import floating
+from omegaconf import DictConfig
+from yaml import safe_load
+
+
+def raise_value_error_if_parameter_has_unexpected_type(
+    parameter, parameter_name, expected_type
+):
+    """Function raises ValueError exception if parameter has unexpected type"""
```
LeonidBeynenson: I would insert here
```python
            if expected_type == float:
                value = (int, float, floating)
```

ote_sdk/ote_sdk/utils/argument_checks.py
```diff
@@ -0,0 +1,217 @@
...
+        raise ValueError(
+            f"Unexpected type of '{parameter_name}' parameter, expected: {expected_type}, actual: {parameter_type}"
+        )
+
+
+def check_parameter_type(parameter, parameter_name, expected_type):
+    """Function extracts nested expected types and raises ValueError exception if parameter has unexpected type"""
+    # pylint: disable=W0212
+    if isinstance(expected_type, typing._GenericAlias):  # type: ignore
+        origin_class = expected_type.__dict__.get("__origin__")
```
LeonidBeynenson: Please, add before this line special check  `if expected_type == typing.Any`
In this case just return. (Note that `Any.__dict__ = {}`)

ote_sdk/ote_sdk/utils/argument_checks.py
```diff
@@ -0,0 +1,217 @@
...
+):
+    """Function raises ValueError exception if required or optional parameter has unexpected type"""
+    check_required_parameters_type(required_parameters)
+    check_optional_parameters_type(optional_parameters)
+
+
+def check_nested_elements_type(iterable, parameter_name, expected_type):
+    """Function raises ValueError exception if one of elements in collection has unexpected type"""
+    for element in iterable:
+        raise_value_error_if_parameter_has_unexpected_type(
```
LeonidBeynenson: I would use here `check_parameter_type` -- it will be a sort of recursion.
And `parameter_name` in the call I would set to `f"element in {parameter_name}"` -- it will allow to work with `List[List[int]]`.

ote_sdk/ote_sdk/utils/argument_checks.py
```diff
@@ -0,0 +1,217 @@
...
+            parameter_name=f"nested {parameter_name}",
+            expected_type=expected_type,
+        )
+
+
+def check_dictionary_keys_values_type(
+    parameter, parameter_name, expected_key_class, expected_value_class
+):
+    """Function raises ValueError exception if dictionary keys or values have unexpected type"""
+    for key, value in parameter.items():
```
LeonidBeynenson, line 108: I would use here `check_parameter_type` -- it will be a sort of recursion.
Also it will allow to work with `Dict[Any, str]`
And `parameter_name` in the call I would set to f"key in {parameter_name}" .

ote_sdk/ote_sdk/utils/argument_checks.py
```diff
@@ -0,0 +1,217 @@
...
+        parameter_type = type(key)
+        if not isinstance(key, expected_key_class):
+            raise ValueError(
+                f"Unexpected type of nested '{parameter_name}' dictionary key, expected: {expected_key_class}, "
+                f"actual: {parameter_type}"
+            )
+        parameter_type = type(value)
```
LeonidBeynenson: I would use here `check_parameter_type` instead of `if` -- it will be a sort of recursion.
Also it will allow to work with `Dict[str, Any]`
And `parameter_name` in the call I would set to `f"value in {parameter_name}" `.

LeonidBeynenson (changes requested):

ote_sdk/ote_sdk/utils/argument_checks.py
```diff
@@ -0,0 +1,280 @@
...
+            expected_key_class=key,
+            expected_value_class=value,
+        )
+
+
+def check_input_param_type(checks: list):
+    """Function to apply methods on checks according to their type"""
+    for param_check in checks:
+        if isinstance(param_check, BaseInputArgumentChecker):
+            param_check.check()
```
LeonidBeynenson:
```suggestion
def check_input_param(*checks: BaseInputArgumentChecker):
    """Function to apply methods on checks according to their type"""
    for param_check in checks:
        if not isinstance(param_check, BaseInputArgumentChecker):
            raise TypeError(f"Wrong parameter of check_input_param: {param_check}")
        param_check.check()
```
Please, note that now the function receives `*checks` -- it means, that you can just point checks as the list of arguments, no need to enclose into `[...]`

ote_sdk/ote_sdk/utils/argument_checks.py
```diff
@@ -0,0 +1,280 @@
...
+            expected_type=(str, DictConfig, dict),
+        )
+        check_that_parameter_is_not_empty(
+            parameter=self.parameter, parameter_name=parameter_name
+        )
+        if isinstance(self.parameter, str):
+            check_that_null_character_absents_in_string(
+                parameter=self.parameter, parameter_name=parameter_name
+            )
+            if isinstance(safe_load(self.parameter), str):
```
LeonidBeynenson: Please, use `yaml.safe_load` instead of just `safe_load`. Otherwise it is not obvious if it is your function or yaml.

ote_sdk/ote_sdk/utils/argument_checks.py
```diff
@@ -0,0 +1,280 @@
...
+            parameter_name=parameter_name,
+            expected_type=(str, DictConfig, dict),
+        )
+        check_that_parameter_is_not_empty(
+            parameter=self.parameter, parameter_name=parameter_name
+        )
+        if isinstance(self.parameter, str):
+            check_that_null_character_absents_in_string(
+                parameter=self.parameter, parameter_name=parameter_name
+            )
```
LeonidBeynenson, line 369: Please, check non-printable characters here too.

ote_sdk/ote_sdk/utils/argument_checks.py
LeonidBeynenson: **Attention**
And why do you make `yaml.safe_load` and expects just one string inside it? Any arbitrary structure may be in YAML, if a YAML string is used to keep only one string it is really strange.

LeonidBeynenson (changes requested):

ote_sdk/ote_sdk/utils/argument_checks.py
saltykox: If DatasetManagerConfig object will be converted into string and passed to “create” function (as for “test_config_reconstruction” unit test), there will be new line characters in the end of “description” “dummy_float_selectable”, etc sections

LeonidBeynenson: Oh, agree!
Ok, then, please, add to the function `check_that_all_characters_printable` parameter `allow_crlf`

ote_sdk/ote_sdk/utils/argument_checks.py
```diff
@@ -0,0 +1,280 @@
...
+
+def check_that_parameter_is_not_empty(parameter, parameter_name):
+    """Function raises ValueError if parameter is empty"""
+    if not parameter:
+        raise ValueError(f"parameter {parameter_name} is empty")
+
+
+def check_that_all_characters_printable(parameter, parameter_name):
+    """Function raises ValueError if one of string-parameter characters is not printable"""
+    if not all(c.isprintable() for c in parameter):
```
LeonidBeynenson: Please, do something like that:
```suggestion
def check_that_all_characters_printable(parameter, parameter_name, allow_crlf=False):
    """Function raises ValueError if one of string-parameter characters is not printable"""
    if not allow_crlf:
        is_ok = all(c.isprintable() for c in parameter)
    else:
        is_ok = all((c.isprintable() or c=="\n" or c == "\r") for c in parameter)
    if not is_ok:
```

LeonidBeynenson (changes requested):

ote_sdk/ote_sdk/utils/argument_checks.py
saltykox: thanks, changed input arguments for "check_input_param" function

ote_sdk/ote_sdk/utils/argument_checks.py
saltykox: thanks, reworked "check" method of InputConfigCheck class

ote_sdk/ote_sdk/utils/argument_checks.py
saltykox: thanks, added "allow_crlf" flag in "check_that_all_characters_printable" function

nervana-ff:

Ilya-Krylov:
draft?

saltykox:
This pull request is in the “Draft” state due to unmerged changes to mmdetection and mmsegmentation repositories related to fixing unexpected type values, but the approach to check input parameters is ready for review

saltykox:
ote-test/1166

saltykox:
@LeonidBeynenson @Ilya-Krylov @samet-akcay please review

LeonidBeynenson (dismissed):
Approve the changes

LeonidBeynenson:
@Ilya-Krylov , could you review this PR? it seems to me that it is OK.

The build ote-test/1166 contains errors, but the errors are not related to the Valry's changes.

Ilya-Krylov:
run ote_sdk tests

ote_sdk/ote_sdk/entities/annotation.py
```diff
@@ -26,7 +31,15 @@ class Annotation(metaclass=abc.ABCMeta):
...
     ):
+
         self.__id_ = ID(ObjectId()) if id is None else id
+        check_input_param_type(
+            RequiredParamTypeCheck(shape, "shape", ShapeEntity),
+            RequiredParamTypeCheck(labels, "labels", List[ScoredLabel]),
+            OptionalParamTypeCheck(id, "id", ID),
+        )
+
+        self.__id = ID(ObjectId()) if id is None else id
```
Ilya-Krylov: where does it come from?

ote_sdk/ote_sdk/entities/annotation.py
```diff
@@ -26,7 +31,15 @@ class Annotation(metaclass=abc.ABCMeta):
     def __init__(
         self, shape: ShapeEntity, labels: List[ScoredLabel], id: Optional[ID] = None
     ):
+
```
Ilya-Krylov: rm extra line

ote_sdk/ote_sdk/entities/annotation.py
```diff
@@ -26,7 +31,15 @@ class Annotation(metaclass=abc.ABCMeta):
...
         self.__id_ = ID(ObjectId()) if id is None else id
+        check_input_param_type(
+            RequiredParamTypeCheck(shape, "shape", ShapeEntity),
```
Ilya-Krylov: Why do you need a second str parameter?

ote_sdk/ote_sdk/entities/dataset_item.py
```diff
@@ -96,6 +101,16 @@ def __init__(
...
         ] = None,
     ):
+        check_input_param_type(
+            RequiredParamTypeCheck(media, "media", IMedia2DEntity),
+            RequiredParamTypeCheck(
+                annotation_scene, "annotation_scene", AnnotationSceneEntity
+            ),
+            OptionalParamTypeCheck(roi, "roi", Annotation),
+            OptionalParamTypeCheck(metadata, "metadata", Sequence[MetadataItemEntity]),
+            RequiredParamTypeCheck(subset, "subset", Subset),
```
Ilya-Krylov: ignored_labels ?

ote_sdk/ote_sdk/entities/datasets.py
```diff
@@ -127,6 +132,11 @@ def __init__(
         items: Optional[List[DatasetItemEntity]] = None,
         purpose: DatasetPurpose = DatasetPurpose.INFERENCE,
     ):
+        check_input_param_type(
+            OptionalParamTypeCheck(items, "items", List[DatasetItemEntity]),
```
Ilya-Krylov: won't it work if  RequiredParamTypeCheck(items, "items",  Optional[List[DatasetItemEntity]]) ?

I guess there is might be a confusion, what 'optional' means

ote_sdk/ote_sdk/entities/annotation.py
Ilya-Krylov: Is it possible not to list parameters typings using \_\_annotations__ for example?

ote_sdk/ote_sdk/entities/model.py
```diff
@@ -116,6 +124,49 @@ def __init__(
...
+                optimization_objectives, "optimization_objectives", Dict[str, str]
+            ),
+            OptionalParamTypeCheck(
+                performance_improvement, "performance_improvement", Dict[str, float]
+            ),
+            RequiredParamTypeCheck(model_size_reduction, "model_size_reduction", float),
+            OptionalParamTypeCheck(_id, "_id", (ID, ObjectId)),
+        )
+        if train_dataset:
+            DatasetParamTypeCheck(train_dataset, "train_dataset").check()
```
Ilya-Krylov: why is  `RequiredParamTypeCheck(train_dataset, "train_dataset", DatasetEntity)` not here?

ote_sdk/ote_sdk/tests/entities/test_datasets.py
```diff
@@ -562,10 +560,6 @@ def test_dataset_entity_append(self):
         dataset.append(item_to_add)
         expected_items.append(item_to_add)
         assert dataset._items == expected_items
-        # Checking that ValueError exception is raised when appending DatasetEntity with "media" is "None" attribute
```
Ilya-Krylov, line 565: why?

saltykox:
ote-test/1268

ote_sdk/ote_sdk/entities/annotation.py
saltykox: I need second str parameter for ValueError exception message, to give information which parameter has unexpected type

ote_sdk/ote_sdk/entities/datasets.py
saltykox: Subscripted generics cannot be used with class and instance checks, I can try to add new branch to the check_parameter_type function

ote_sdk/ote_sdk/entities/annotation.py
Ilya-Krylov: Can we avoid this duplicating somehow? Get str by variable or get variable by str?

ote_sdk/ote_sdk/entities/model.py
saltykox: Because as the dataset-like parameter may be used non-child class of DatasetEntity, for example HpoDataset, which has same DatasetEntity attributes and methods, but not defined as child class

ote_sdk/ote_sdk/tests/entities/test_datasets.py
saltykox: Because type annotation for DatasetItemEntity object requires “media” initialization parameter equal to IMedia2DEntity, not Optional[IMedia2DEntity]. Because of this, I can’t append DatasetItemEntity object with “media” attribute equal to “None”

ote_sdk/ote_sdk/entities/model.py
Ilya-Krylov: It may mean that HpoDataset is wrong. @LeonidBeynenson what exact requirements on input parameters validation? I would not like to see in the SDK such a mess. In one place DatasetEntity is checked in one way, in another place it is checked in another way.

ote_sdk/ote_sdk/entities/annotation.py
saltykox: thanks, I will try to do it

ote_sdk/ote_sdk/entities/model.py
saltykox: I can add DatasetParamTypeCheck in all methods and functions, where dataset-like parameter is used

ote_sdk/ote_sdk/entities/annotation.py
saltykox: If I try to get parameter_name inside the ote_sdk/ote_sdk/utils/argument_checks.py classes, it will be already changed to the “parameter”, so I have to pass str values directly in the ote_sdk classes and functions. Ex., if I change RequiredParamTypeCheck __init__ method to:
self.parameter = parameter
self.parameter_name = f"{parameter=}".split("=")[0]
self.expected_type = expected_type
Then parameter_name attribute will be equal to the “parameter”
Is there better way to get parameter name? global() and local() not return correct parameter names inside argument_checks.py too

ote_sdk/ote_sdk/entities/annotation.py
saltykox: Thanks, I will remove it

saltykox: Removed 42nd line appeared due to incorrect resolve of merge conflicts

ote_sdk/ote_sdk/entities/annotation.py
saltykox: Thanks, removed extra line

ote_sdk/ote_sdk/entities/annotation.py
saltykox: Thanks, added InputParamTypeCheck class, removed RequiredParamTypeCheck and OptionalParamTypeCheck classes

ote_sdk/ote_sdk/entities/dataset_item.py
saltykox: Thanks, added check of "ignored_labels" input parameter, added related unit tests

ote_sdk/ote_sdk/entities/datasets.py
saltykox: Thanks, added InputParamTypeCheck class, removed RequiredParamTypeCheck and OptionalParamTypeCheck classes

ote_sdk/ote_sdk/entities/model.py
saltykox: Added DatasetParamTypeCheck in all methods and functions, where dataset-like parameter is used

Ilya-Krylov: Thanks, but it is better to have InputParamTypeCheck for everything.

saltykox: Ok, lets wait for exact requirements from Leonid, or I can add DatasetItemEntity branch in the check_parameter_type function

Ilya-Krylov: Could you investigate whether method can be decorated so that there is no need to list parameters, their names, their types, for example https://stackoverflow.com/questions/15299878/how-to-use-python-decorators-to-check-function-arguments

saltykox: Thanks, added draft version of check_input_parameters_type decorator and updated check_parameter_type function

saltykox: Thanks, added “check_input_parameters_type” decorator, since there are custom checks required, added checks_types parameter dictionary with custom checks

ote_sdk/ote_sdk/entities/annotation.py
saltykox: Thanks, added “check_input_parameters_type” decorator currently works with __annotations__ attribute to get expected types of input parameters

saltykox:
ote-test/1443

ote_sdk/ote_sdk/entities/dataset_item.py
```diff
@@ -93,7 +95,7 @@ def __init__(
         metadata: Optional[Sequence[MetadataItemEntity]] = None,
         subset: Subset = Subset.NONE,
         ignored_labels: Optional[
-            Union[List[LabelEntity], Tuple[LabelEntity, ...], Set[LabelEntity]]
```
Ilya-Krylov: Why?

Ilya-Krylov: Why?

ote_sdk/ote_sdk/entities/model.py
```diff
@@ -110,11 +124,11 @@ def __init__(
...
         optimization_type: ModelOptimizationType = ModelOptimizationType.NONE,
-        optimization_methods: List[OptimizationMethod] = None,
-        optimization_objectives: Dict[str, str] = None,
-        performance_improvement: Dict[str, float] = None,
+        optimization_methods: Optional[List[OptimizationMethod]] = None,
+        optimization_objectives: Optional[Dict[str, str]] = None,
+        performance_improvement: Optional[Dict[str, float]] = None,
         model_size_reduction: float = 0.0,
-        _id: Optional[ID] = None,
+        _id: Optional[Union[ID, ObjectId]] = None,
```
Ilya-Krylov: Why?

Ilya-Krylov: Why?

ote_sdk/ote_sdk/utils/argument_checks.py
```diff
@@ -0,0 +1,407 @@
...
+
+
+class OptionalImageFilePathCheck(OptionalFilePathCheck):
+    """Class to check optional image file path parameters"""
+
+    def __init__(self, parameter, parameter_name):
+        super().__init__(
+            parameter=parameter,
+            parameter_name=parameter_name,
+            expected_file_extension=["jpg", "png"],
```
Ilya-Krylov: https://docs.opencv.org/3.4/d4/da8/group__imgcodecs.html#ga288b8b3da0892bd651fce07b3bbd3a56

Ilya-Krylov: https://docs.opencv.org/3.4/d4/da8/group__imgcodecs.html#ga288b8b3da0892bd651fce07b3bbd3a56

ote_sdk/ote_sdk/entities/image.py
```diff
@@ -27,6 +31,7 @@ class Image(IMedia2DEntity):
     """
 
     # pylint: disable=too-many-arguments, redefined-builtin
+    @check_input_parameters_type({"file_path": OptionalImageFilePathCheck})
```
Ilya-Krylov, line 34: is it really required that this is a file path to an image?

Ilya-Krylov, line 34: is it really required that this is a file path to an image?

ote_sdk/ote_sdk/entities/dataset_item.py
saltykox: If I use debugger, “Tuple[LabelEntity, ...].__dict__.get("__args__")” (to get list of expected types ) will return (<class 'ote_sdk.entities.label.LabelEntity'>, Ellipsis). This “Ellipsis” (second element of tuple) even not type-object. Does it acceptable as ignored_labels expected type?

ote_sdk/ote_sdk/entities/model.py
saltykox: Thanks I will roll back changes in expected types and verify them on Jenkins, I see that several updates related to ObjectId are commited after open of Pull Request

ote_sdk/ote_sdk/entities/image.py
saltykox: As I see in Image class description - lines 23-30
`class Image(IMedia2DEntity):
    """
    Represents a 2D image.

    The image must be instantiated with either a NumPy array containing the image data
    or a path to an image file.

    :param data: NumPy data.
    **:param file_path: Path to image file**.
    """`
What checks should I add for this input parameter?

saltykox: I used Image class description as expected type of "file_path" parameter lines 24-30:

Represents a 2D image.

The image must be instantiated with either a NumPy array containing the image data
or a path to an image file.
   :param data: NumPy data.
   :**param file_path: Path to image file**.

Should I use other check for file_path parameter?

ote_sdk/ote_sdk/utils/argument_checks.py
saltykox: Thanks, I will update the list of expected files extensions according to the imread method supported formats

saltykox: Updated expected types, but in specified link there are no extensions included for “Raster and Vector geospatial data supported by GDAL”, should I add all extensions from https://svn.osgeo.org/gdal/tags/gdal_1_2_5/frmts/formats_list.html for Raster and for https://gdal.org/drivers/vector/index.html Vector?

ote_sdk/ote_sdk/entities/dataset_item.py
Ilya-Krylov: Tuple[LabelEntity, ...] is not the same as Tuple[LabelEntity] , see https://docs.python.org/3/library/typing.html#typing.Tuple

ote_sdk/ote_sdk/entities/image.py
Ilya-Krylov: sorry, this question is more for @LeonidBeynenson , I forgot to mention him

@LeonidBeynenson 
Is it required to check whether it is path to an image, folder, etc ? Is not it enough to check whether it is str or not.
I would like to see this `check_input_parameters_type` without any parameter, if it is possible.

ote_sdk/ote_sdk/entities/dataset_item.py
saltykox: Thanks, I will roll back changes in ignored_labels expected type, with actual version of input parameters validation recursion, this case is included and works as expected

saltykox: Thanks, now I understand.
@LeonidBeynenson for that case the check (argument_checks.py lines 126-129)

if len(nested_elements_class) != 1:
            raise TypeError(
                "length of nested expected types for Sequence should be equal to 1"
            )

does not work, I will add separate branch for tuple objects, with TypeError if len(nested_elements_class) > 2 check

saltykox: Rolled back changes in ignored_labels expected type for DatasetItemEntity, added branch for tuples length equal 2 in check_parameter_type function

ote_sdk/ote_sdk/entities/image.py
LeonidBeynenson: I think it is required. We need to validate input parameters where it is possible. It is SDL requirement.
IMHO, addition of one parameter into the decorator is not an evil. 🙂

ote_sdk/ote_sdk/tests/parameters_validation/validation_helper.py
```diff
@@ -0,0 +1,23 @@
...
+
+# Copyright (C) 2021-2022 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
+
+import pytest
+
+
+def check_value_error_exception_raised(
+    correct_parameters: dict, unexpected_values: list, class_or_function
```
druzhkov-paul: Callable?

druzhkov-paul (dismissed):

ote_sdk/ote_sdk/entities/image.py
Ilya-Krylov: @LeonidBeynenson How can you imagine the process when developer will add new functionality and need to understand what is going on with this `check_input_parameters_type` and which parameters should be passed here.

ote_sdk/ote_sdk/tests/parameters_validation/validation_helper.py
saltykox: thanks, added "Callable" annotation for "class_or_function" parameter

saltykox:
ote-test/1581

Ilya-Krylov:
run ote_sdk tests

Ilya-Krylov:
run ote-test

saltykox:
run ote_sdk tests

saltykox:
run ote-test

saltykox:
run ote-test

saltykox:
run ote_sdk tests

saltykox:
run ote-test

saltykox:
run ote_sdk tests

PR #892, opened 2022-02-02
Fix QUICK_START_GUIDE.md

nmanovic:
It is necessary to install ote_sdk to run OTE CLI. I'm not sure that the fix is correct, but it works.


Ilya-Krylov:
Thanks @nmanovic  for contribution, but we have already fix for this issue in develop branch.
https://github.com/openvinotoolkit/training_extensions/blob/develop/QUICK_START_GUIDE.md#setup-openvino-training-extensions
user has to export OTE_SDK_PATH variable and it is enough

PR #900, opened 2022-02-07
[Anomaly Classification] Add check to logger

ashwinvaidya17:
Fixes: https://jira.devtools.intel.com/browse/CVS-76094
BMM: https://10.211.120.58/projects/6200f29b61caa56c8b6b3bf3/models

### DIFF
--- a/external/anomaly/anomaly_classification/task.py
+++ b/external/anomaly/anomaly_classification/task.py
@@ -226,6 +226,8 @@ class AnomalyClassificationTask(ITrainingTask, IInferenceTask, IEvaluationTask, 
 
         # NOTE: This is for debugging purpose.
         for i, _ in enumerate(output_resultset.ground_truth_dataset):
+            if i >= len(output_resultset.prediction_dataset):
+                break
             logger.info(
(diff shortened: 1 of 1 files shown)


ashwinvaidya17:
run ote_sdk tests

djdameln (commented):
Thanks for addressing this! It is still a bit unclear to me how it would be possible how the number of items in the GT and prediction dataset could differ. Also, I imagine this difference could cause problems elsewhere in the task?

samet-akcay:
yeah, it's still a bit unclear how/why they are not the same. This would maybe avoid the error; however, i'm not sure what would be the behaviour when the metrics are computed with GT and predicted sets don't have the same lengths

atwinand:
Same comment, better not to hide the issue like this. But you guys can confirm that the code is not playing around with the subsets?

samet-akcay:
@atwinand, can confirm that subsets are not modified

samet-akcay:
Closing this since this doesn't fully solve the issue, but silence it instead. The actual issue is thought to be on the backend

PR #903, opened 2022-02-08
[CLASSIFICATION] Fixes

sovrasov:
- Prevent early stopping if LR warmup is not finished
- Fix usage of removed exclusive_with flag in LabelSchema

BLD 877 🏃‍♂️
BMM 24812 🏃‍♂️


sovrasov:
run ote_sdk tests

PR #914, opened 2022-02-11
Add PR labeler

morkovka1337:
This PR adds PR labeler like in the misc branch and adds automatic settings of 3 other labels (CLI, SDK, tests)


.github/workflows/labaler.yml
```diff
@@ -0,0 +1,11 @@
+name: "Pull Request Labeler"
```
Ilya-Krylov: typo in filename

.github/labeler.yml
```diff
@@ -0,0 +1,11 @@
+# See help here: https://github.com/marketplace/actions/labeler
+
```
Ilya-Krylov, line 2: tasks:
  - 'external/*'

PR #931, opened 2022-02-17
add is_anomalous flag to OTE label

Harmjandevries:
As discussed a few weeks ago, we will replace checking the name of the label with checking for a flag is_anomalous. We need to do it in multiple steps to avoid causing regression between merges:

1. OTE: Add is_anomalous attribute to LabelEntity

2. Backend: Add is_anomalous in SC and add it as a non-required field in the project creation contract. Automatically set it from the “name” field on project creation. Replace all checks for the label name with checks for is_anomalous.

3. OTE: Replace all checks for the label name with checks for the value of is_anomalous.

4. Backend: Use the is_anomalous field in the API to set the value of the attribute during project creation instead of determining it based on the label name. Remove checks that block the API from changing normal and anomalous label names.

### DIFF
--- a/ote_sdk/ote_sdk/entities/label.py
+++ b/ote_sdk/ote_sdk/entities/label.py
@@ -78,7 +78,7 @@ class LabelEntity:
         will be assigned upon saving. If the argument is None, it will be set to ID()
     """
 
-    # pylint: disable=redefined-builtin, too-many-arguments; Requires refactor
+    # pylint: disable=redefined-builtin, too-many-instance-attributes, too-many-arguments; Requires refactor
     def __init__(
         self,
         name: str,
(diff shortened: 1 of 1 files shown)


nervana-ff:

Harmjandevries:
run ote_sdk tests

ote_sdk/ote_sdk/entities/label.py
```diff
@@ -88,6 +88,7 @@ def __init__(
         creation_date: Optional[datetime.datetime] = None,
         is_empty: bool = False,
         id: Optional[ID] = None,
+        is_anomalous: bool = False,
```
Ilya-Krylov, line 93: please update docstring

Harmjandevries: Done, thanks.

Harmjandevries:
run ote_sdk tests

PR #936, opened 2022-02-18
added rotated rectangle

Ilya-Krylov:


nervana-ff:

Ilya-Krylov:
run ote_sdk tests

Ilya-Krylov:
run ote_sdk tests

ote_sdk/ote_sdk/entities/shapes/rotated_rectangle.py
```diff
@@ -0,0 +1,187 @@
...
+    ):
+        labels = [] if labels is None else labels
+        modification_date = now() if modification_date is None else modification_date
+        super().__init__(
+            type=ShapeType.ROTATED_RECTANGLE,
+            labels=labels,
+            modification_date=modification_date,
+        )
+
+        if len(points) != 4:
```
AlbertvanHouten, line 48: I think it would be good to add further validation to ensure it is a rectangle.

Ilya-Krylov: since these points are normalised, we can check whether it is a parallelogram, will it be enough for you?

AlbertvanHouten: Ah right. I guess you can't check for 90 degree corners in normalised space? Otherwise disregard my comment.

leoll2: Imho it would be good to reject also degenerate rectangles (actually parallelograms) e.g. with zero-length sides or all the points lined up.

ote_sdk/ote_sdk/entities/shapes/rotated_rectangle.py
```diff
@@ -0,0 +1,187 @@
...
+            )
+        return False
+
+    def __hash__(self):
+        return hash(str(self))
+
+    def normalize_wrt_roi_shape(self, roi_shape: Rectangle) -> "RotatedRectangle":
+        """
+        Transforms from the `roi` coordinate system to the normalized coordinate system.
+        This function is the inverse of ``denormalize_wrt_roi_shape``.
```
leoll2, line 119: How true is this statement? I mean, is it true that `denormalize(normalize(x)) == x`? I suspect that floating point rounding errors and ill-conditioned inputs can break the property. If true, it should be tested with challenging inputs; if false, it should be documented.

ote_sdk/ote_sdk/entities/shapes/rotated_rectangle.py
```diff
@@ -0,0 +1,187 @@
...
+
+        points = [p.denormalize_wrt_roi_shape(roi_shape) for p in self.points]
+        return RotatedRectangle(points=points)
+
+    def _as_shapely_polygon(self) -> shapely_polygon:
+        """
+        Returns the RotatedRectangle object as a shapely polygon
+        which is used for calculating intersection between shapes.
+        """
+        return shapely_polygon([(point.x, point.y) for point in self.points])
```
leoll2, line 192: shapely considers the counter-clockwise (ccw) orientation to be the "normal" one, but it allows to build clockwise (cw) polygons anyway. Just double-checking, the fact that our points are cw is not going to create issues right?

Ilya-Krylov: I think it depends on Y-axis orientation, if it points down, a polygon is cw, otherwise is ccw

I assume that our normalized coordinates are comes from image plane where Y-axis is directed downward. If my assumption is wrong, please let me know.

ote_sdk/ote_sdk/entities/shapes/rotated_rectangle.py
Ilya-Krylov: added checks related to sides and area

ote_sdk/ote_sdk/entities/shapes/rotated_rectangle.py
Ilya-Krylov: you are right, added a note

ote_sdk/ote_sdk/entities/shapes/rotated_rectangle.py
leoll2: Thanks for adding the checks! A more efficient way to check the parallelogram condition would be exploiting the property that its diagonals bisect each other, instead of comparing the side lengths in pairs (0 vs 4 sqrt operations). Basically you can simply compute the midpoints of the two diagonals and verify their closeness.  
But I'm ok with the current implementation, my comment is perhaps overkill.

leoll2 (approved):
LGTM but please wait for approval from Albert too

AlbertvanHouten (approved):

PR #941, opened 2022-02-21
[OTE][OTE Val][SDL] Validation ote cli input parameters common part

0de554K:
Validation ote cli input parameters common part


nervana-ff:

ote_cli/ote_cli/utils/validate_path.py
```diff
@@ -0,0 +1,54 @@
...
+    try:
+        if not isinstance(test_string, str) or not test_string:
+            return False
+        _, test_string = os.path.splitdrive(test_string)
+
+        root = os.path.sep
+        assert os.path.isdir(root)
+
+        root = root.rstrip(os.path.sep) + os.path.sep
+        for pathname_part in test_string.split(os.path.sep):
```
LeonidBeynenson: Why do not you use the standard `os.path.isdir`, `os.path.isfile`, etc?

ote_cli/ote_cli/utils/validate_path.py
```diff
@@ -0,0 +1,54 @@
...
+            try:
+                os.lstat(root + pathname_part)
```
LeonidBeynenson: I think that if I use `test_string=/home/lbeynens/somefolder/somefile.txt`, then this cycle will check paths `/home`, `/lbeynens`, `/somefolder` -- and obviously will fail for all paths except the first one.
Is this correct?

ote_cli/ote_cli/utils/validate_path.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def is_path_valid(test_string: str) -> bool:
+    try:
+        if not isinstance(test_string, str) or not test_string:
+            return False
+        _, test_string = os.path.splitdrive(test_string)
+
+        root = os.path.sep
+        assert os.path.isdir(root)
```
LeonidBeynenson: At the moment OTE works on Linux only -- we do not suppose that it can work on Windows or other OS.
In Linux `os.path.sep` is always `/`, and the path `/` always exists.
Why do you want to check this?

ote_cli/ote_cli/utils/validate_path.py
```diff
@@ -0,0 +1,54 @@
...
+                return False
+    except TypeError:
+        return False
+    else:
+        return True
+
+
+def validate_single_path(path: str) -> None:
+    if not is_path_valid(path):
+        raise Exception(f"Path is not valid: {path}")
```
LeonidBeynenson, line 49: Why do you raise the root class `Exception`? 
You could use `ValueError`
>[exception ValueError](https://docs.python.org/3/library/exceptions.html#ValueError)
>Raised when an operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as [IndexError](https://docs.python.org/3/library/exceptions.html#IndexError).

LeonidBeynenson (changes requested):

ote_cli/ote_cli/utils/validate_path.py
0de554K: It was a try to catch this error 
>>> a = "aaaa\0bbb"
>>> all([s.isprintable() for s in a])
False
>>> os.system(a)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: embedded null byte

But it should be removed due its helpless.

PR #946, opened 2022-02-24
CVS-80322 Update requirements.txt to loosen attrs version

younesintel:


nervana-ff:

PR #950, opened 2022-02-24
[Anomaly Segmentation] Reduce requests version

ashwinvaidya17:
Reduce requests version to match tests.


nervana-ff:

PR #957, opened 2022-03-03
[OTE_SDK] Add attribute ignored_labels to DatasetItemEntity

leoll2:
It is needed to implement 'dynamic labels'-aware training routines.

### DIFF
--- a/ote_sdk/ote_sdk/entities/dataset_item.py
+++ b/ote_sdk/ote_sdk/entities/dataset_item.py
@@ -9,7 +9,7 @@ import copy
 import itertools
 import logging
 from threading import Lock
-from typing import List, Optional, Sequence
+from typing import List, Optional, Sequence, Set, Tuple, Union
 
 import numpy as np
 
@@ -31,13 +31,14 @@ class DatasetItemEntity(metaclass=abc.ABCMeta):
     DatasetItemEntity represents an item in the DatasetEntity. It holds a media item, annotation and an ROI.
     The ROI determines the region of interest for the dataset item, and is described by a shape entity.
 
-    Dataset items hold five fundamental properties:
+    The fundamental properties of a dataset item are:
 
     - A 2d media entity (e.g. Image)
     - A 2d annotation entity for the full resolution media entity
     - An ROI, describing the region of interest.
     - The subset it belongs to
     - Metadata for the media entity (e.g. saliency map or active score)
+    - A list of labels to ignore
 
     .. rubric:: Getting data from dataset item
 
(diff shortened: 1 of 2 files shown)


leoll2:
run ote_sdk tests

leoll2:
run ote_sdk tests

leoll2:
e2e: https://ci.iotg.sclab.intel.com/job/IMPT/job/IMPTOps/25991/

ote_sdk/ote_sdk/entities/dataset_item.py
```diff
@@ -107,11 +114,28 @@ def __init__(
         if metadata is not None:
             self.__metadata = list(metadata)
 
+        self.__ignored_labels: Set[LabelEntity] = (
```
druzhkov-paul, line 117: Nit: I guess it could be a `FrozenSet` as it's not supposed to be modified.

ote_sdk/ote_sdk/entities/dataset_item.py
```diff
@@ -88,6 +92,9 @@ def __init__(
         roi: Optional[Annotation] = None,
         metadata: Optional[Sequence[MetadataItemEntity]] = None,
         subset: Subset = Subset.NONE,
+        ignored_labels: Optional[
```
druzhkov-paul, line 95: Why ignored labels are directly stored in the dataset item, while other labels are stored in the annotation scene?

leoll2: Good question, it is part of the SC backend design. Long story short, we do not create new AnnotationScenes upon label schema change due to performance and consistency reasons, but instead we store the revisit/ignore information in a parallel entity (AnnotationSceneState). When it comes to build the datasets, it's much easier and efficient for us to push this information in the DatasetItem rather than in the AnnotationScene.

ote_sdk/ote_sdk/entities/dataset_item.py
leoll2: Nice suggestion, but I prefer not to change it now (unless needed) because it could conflict with some type hints that are already in SC. Unfortunately `FrozenSet` is not a subtype of `Set`...

ote_sdk/ote_sdk/entities/dataset_item.py
druzhkov-paul: I see, thanks for explanation. I'm not familiar with the design proposal for this, but at a high level it seems that attaching ignored labels to a dataset item ROI, rather than a dataset item itself, could be a better choice in terms of consistency.

sovrasov (approved):

PR #975, opened 2022-03-16
[ANOMALY SEGMENTATION] Use only polygon annotations for evaluation

djdameln:
This PR fixes an issue in anomaly segmentation where the evaluation was done on both global (image-level) and local (polygon) annotations. With these changes, the resultset is now split into a subset that has local annotations available and a subset that has only global annotations available. If the local subset contains anomalous images, it will be used to compute performance. Otherwise the global dataset will be used (in this case the reported performance will be image-level F1). This would allow users to train the task and see the predictions before making any polygon annotations.

Additionally, a similar split is done for the validation set before sending the data to Anomalib. This makes sure that Anomalib does not try to compute segmentation performance without ground truth masks available. If the global dataset is used for validation, Anomalib will use the image-level threshold to generate segmentation predictions.

[BMM](https://ci.iotg.sclab.intel.com/job/IMPT/job/IMPTOps/27001/)

### DIFF
--- a/external/anomaly/ote_anomalib/callbacks/inference.py
+++ b/external/anomaly/ote_anomalib/callbacks/inference.py
@@ -58,16 +58,13 @@ class AnomalyInferenceCallback(Callback):
             self.ote_dataset, pred_scores, pred_labels, anomaly_maps, pred_masks
         ):
             label = self.anomalous_label if pred_label else self.normal_label
-            if self.task_type == TaskType.ANOMALY_CLASSIFICATION:
-                probability = (1 - pred_score) if pred_score < 0.5 else pred_score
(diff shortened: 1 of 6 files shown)


djdameln:
run ote_sdk tests

Ilya-Krylov:
run ote_sdk tests

Ilya-Krylov:
run ote_sdk tests

Ilya-Krylov:
run ote_sdk tests

Ilya-Krylov:
http://213.221.44.203/job/ote/job/precommit/job/ote_sdk_tests_runner/753/console 
so can think that tests passed, there was no problem with anomaly

djdameln:
Verified that the changes achieve the desired effect in the UI. I think this can be merged @Ilya-Krylov

PR #979, opened 2022-03-17
[ANOMALY DETECTION] - Add Anomaly Detection Task.

samet-akcay:
I've recently pulled the development, which is the reason why so many changes appear here.


nervana-ff:

external/anomaly/ote_anomalib/openvino.py
```diff
@@ -212,7 +220,7 @@ def evaluate(self, output_resultset: ResultSetEntity, evaluation_metric: Optiona
             output_resultset (ResultSetEntity): Result set storing ground truth and predicted dataset.
             evaluation_metric (Optional[str], optional): Evaluation metric. Defaults to None.
         """
-        if self.task_type == TaskType.ANOMALY_CLASSIFICATION:
+        if self.task_type in (TaskType.ANOMALY_CLASSIFICATION, TaskType.ANOMALY_DETECTION):
```
djdameln: Not sure if F-measure is a sensible metric for bbox-level performance. Does the other detection task use this as well?

djdameln (approved):

external/anomaly/ote_anomalib/openvino.py
samet-akcay: Yes, f1 had also been used in NOUS tasks. You could check it out [here](https://github.com/openvinotoolkit/mmdetection/blob/6024bb76624c84dfbc9c44c402a97b29ee9b270c/mmdet/apis/ote/apis/detection/inference_task.py#L329).

PR #981, opened 2022-03-17
move mmseg templates

Ilya-Krylov:

PR #972, opened 2022-03-16
Move templates and related code to OTE

Ilya-Krylov:

PR #983, opened 2022-03-18
[OTE_SDK]  updated TaskType enum

AlbertvanHouten:
Added several booleans to tasktype as well as domain, so you can determine just from the task type if it supports training, is anomaly or is global. This prevents us from having to update multiple lists whenever a new tasktype is added.

After this is merged, usages in the impt repo can be updated and the TRAINABLE TASKS / ANOMALY TASKS lists and domain helper functions can be removed in the OTE_SDK

CVS-81165


ote_sdk/ote_sdk/entities/model_template.py
```diff
@@ -75,22 +75,49 @@ def __str__(self) -> str:
...
+        :param value: Unique integer for .value property of Enum (auto() does not work)
+        :param domain: Label domain of the task type
+        :param is_trainable: Boolean indicating if the task type can be trained
+        :param is_anomaly: Boolean indicating if the task is an anomaly task
+        :param is_global: Boolean indicating if the task support global labels
+        """
+        self.domain = domain
+        self.is_trainable = is_trainable
+        self.is_anomaly = is_anomaly
+        self.is_global = is_global
```
leoll2: I wonder about adding an `is_local` attribute as well. Note that `is_local` is not the opposite of `is_global`, because some tasks like `DATASET` are neither local nor global. Just want to make sure that people don't mistakenly use `not task_type.is_global` to check if a task is local.

ote_sdk/ote_sdk/entities/model_template.py
```diff
@@ -75,22 +75,49 @@ def __str__(self) -> str:
...
+    SEGMENTATION = 4, Domain.SEGMENTATION, True, False, False
+    DETECTION = 5, Domain.DETECTION, True, False, False
+    ANOMALY_DETECTION = 6, Domain.ANOMALY_DETECTION, True, True, False
+    CROP = 7, Domain.NULL, False, False, False
+    TILE = 8, Domain.NULL, False, False, False
+    INSTANCE_SEGMENTATION = 9, Domain.INSTANCE_SEGMENTATION, True, False, False
+    ACTIVELEARNING = 10, Domain.NULL, False, False, False
+    ANOMALY_SEGMENTATION = 11, Domain.ANOMALY_SEGMENTATION, True, True, False
+    ANOMALY_CLASSIFICATION = 12, Domain.ANOMALY_CLASSIFICATION, True, True, True
+    ROTATED_DETECTION = 13, Domain.ROTATED_DETECTION, True, False, False
```
leoll2: I think this is not the cleanest approach because:

1. the order of these hardcoded values depends implicitly on the order of the constructor parameters and it is not evident.
2. The lack of keys makes it difficult to read and spot potential mistakes (e.g. two bools swapped by position)

Explicit is better than implicit. My proposal is something like this:
```python
from enum import Enum
from collections import namedtuple
from typing import NamedTuple

class T(NamedTuple):
    value: int
    is_trainable: bool
    is_global: bool

class TaskType(T, Enum):
    NULL = T(value=1, is_trainable=False, is_global=False)
    CLASSIFICATION = T(value=2, is_trainable=True, is_global=True)

    def __repr__(self) -> str:
        return self.name
```

ote_sdk/ote_sdk/entities/model_template.py
```diff
@@ -452,6 +479,7 @@ def __init__(self) -> None:
 TRAINABLE_TASK_TYPES: Sequence[TaskType] = (
     TaskType.CLASSIFICATION,
     TaskType.DETECTION,
+    TaskType.ROTATED_DETECTION,
```
leoll2: All these global lists can be generated procedurally: 
```python
ANOMALY_TASK_TYPES = {t for t in TaskType if t.is_anomaly}
TRAINABLE_TASK_TYPES = {t for t in TaskType if t.is_trainable}
```

ote_sdk/ote_sdk/entities/model_template.py
AlbertvanHouten: Thanks for this suggestion. I agree that this is cleaner and will update it

ote_sdk/ote_sdk/entities/model_template.py
AlbertvanHouten: I agree that it is a bit strange having these values for all TaskTypes with domain “Domain.NULL”, however for trainable tasks  if it's global it can't be local. I believe in the IMPT repo we also first check if a task is trainable before checking if it is global or local which would work with this approach too. For that reason, I think it is a bit redundant to have is_local.

AlbertvanHouten:
run ote_sdk tests

ote_sdk/ote_sdk/entities/model_template.py
leoll2: > for trainable tasks if it's global it can't be local

Yeah that's true, but this property exists for both trainable and non-trainable tasks. Since this is an SDK, consumed by SC, OTE and other projects, I think it's especially important to make the interface as clear as possible. It's easy to forget (or even ignore) that `not is_global` in practice does not mean "local" but "local OR non-trainable". With `is_local` we bypass the problem because one will never need to write `not is_global` in his code.

leoll2: Also the concepts of trainability and scope (i.e. locality) are orthogonal. In theory, there could exist tasks implementing non-parametric (or with non-trainable params) algorithms that predict shapes (e.g. Selective Search) or labels. As we are at the lowest possible level, the SDK, we must make as few assumptions as possible.

AlbertvanHouten: Good point. Added

ote_sdk/ote_sdk/entities/model_template.py
```diff
@@ -73,24 +73,132 @@ def __str__(self) -> str:
...
+    ROTATED_DETECTION = 13, TaskInfo(
+        domain=Domain.ROTATED_DETECTION,
+        is_trainable=True,
+        is_anomaly=False,
+        is_global=False,
+        is_local=True,
+    )
 
     def __str__(self) -> str:
         return str(self.name)
```
leoll2: Can you replace it with 
```python
def __repr__(self) -> str:
    return self.name
```
It looks cleaner when you print a collection of `TaskType`, e.g. `print(list(TaskType))` or `[x for x in TaskType]`. The common usecase is when logging the list of supported types for a certain feature.

AlbertvanHouten: Sure, done.

ote_sdk/ote_sdk/tests/entities/test_model_template.py
```diff
@@ -252,19 +252,19 @@ def test_tasktype(self):
...
+        assert str(TaskType.SEGMENTATION) == "TaskType.SEGMENTATION"
+        assert str(TaskType.DETECTION) == "TaskType.DETECTION"
+        assert str(TaskType.ANOMALY_DETECTION) == "TaskType.ANOMALY_DETECTION"
+        assert str(TaskType.CROP) == "TaskType.CROP"
+        assert str(TaskType.TILE) == "TaskType.TILE"
+        assert str(TaskType.INSTANCE_SEGMENTATION) == "TaskType.INSTANCE_SEGMENTATION"
+        assert str(TaskType.ACTIVELEARNING) == "TaskType.ACTIVELEARNING"
+        assert str(TaskType.ANOMALY_SEGMENTATION) == "TaskType.ANOMALY_SEGMENTATION"
+        assert str(TaskType.ANOMALY_CLASSIFICATION) == "TaskType.ANOMALY_CLASSIFICATION"
+        assert str(TaskType.ROTATED_DETECTION) == "TaskType.ROTATED_DETECTION"
```
leoll2: My bad, sorry, I didn't foresee this side-effect. Instead of *replacing* `__str__` with `__repr__`, can we keep both? The old `__str__` and the new `__repr__`. I tried locally and it looks smooth.

AlbertvanHouten: Oh yea, I changed the test without thinking too... This will be breaking indeed. I'll fix it again :)

leoll2 (dismissed):

leoll2:
run ote_sdk tests

druzhkov-paul:
Please resolve conflicts and run the tests.

AlbertvanHouten:
run ote_sdk tests

leoll2:
@AlbertvanHouten I think you accidentally pushed modifications to the submodules

AlbertvanHouten:
I'm a bit confused here. Does every developer use a personal .git ignore? I pull from develop and then the changes in the external folder are added

druzhkov-paul:
No, .gitignore is the same for everyone. Some stuff from submodules has been recently moved to OTE itself (#972, #981, #985 )

AlbertvanHouten:
Yea but how come the changes show up in my branch?

leoll2:
> I'm a bit confused here. Does every developer use a personal .git ignore? I pull from develop and then the changes in the external folder are added

Maybe your local submodules are not up-to-date and they conflict when you merge develop. 
An easy (but be careful) way to recover is to reset --mixed your changes, remove the unwanted ones, commit, rebase to develop and force push.

AlbertvanHouten:
Ah right. That makes sense. I'll update my local submodules first then

druzhkov-paul:
I can only guess that it is the issue with a local git client/settings, because changes show file mode switch for an executable script. It may also happed if you are on a Windows machine. Tweaking `core.filemode` git setting might save the day.

AlbertvanHouten:
run ote_sdk tests

leoll2 (dismissed):

ote_sdk/ote_sdk/entities/label.py
```diff
@@ -18,14 +18,15 @@ class Domain(Enum):
...
 
-    CLASSIFICATION = 0
-    DETECTION = 1
-    SEGMENTATION = 2
-    ANOMALY_CLASSIFICATION = 3
-    ANOMALY_DETECTION = 4
-    ANOMALY_SEGMENTATION = 5
-    INSTANCE_SEGMENTATION = 6
-    ROTATED_DETECTION = 7
+    NULL = auto()
```
druzhkov-paul, line 21: What is NULL domain?

ote_sdk/ote_sdk/entities/model_template.py
```diff
@@ -75,28 +75,139 @@ def __str__(self) -> str:
...
 
+    def __repr__(self) -> str:
+        return self.name
```
druzhkov-paul, line 214: What's the difference to the `__str__` here?

leoll2: If only `__str__` is defined and you print a list of TaskType, then it will display like `[<TaskType.NULL: 1>, <TaskType.CLASSIFICATION: 2>]` which is ugly. This is because the containers use `__repr__` of the contained objects to represent them. Conversely, if you define only `__repr__`, then `print(TaskType.DETECTION)` will output `TaskType.DETECTION` because it uses the `__str__` inherited from the `Enum` class.
Redefining both methods will produce the most human-readable output in any case.

ote_sdk/ote_sdk/entities/label.py
AlbertvanHouten: NULL domain is used for all TaskTypes that do not have an associated label domain.

AlbertvanHouten:
run ote_sdk tests

ote_sdk/ote_sdk/entities/model_template.py
druzhkov-paul: Yeah, thanks. 😃 

I was trying to figure out, what's the difference in the implementation of `__str__` and `__repr__` in this particular case:

```python
    def __str__(self) -> str:
        return str(self.name)

    def __repr__(self) -> str:
        return self.name
```

That is, when `str(self.name)` differs from `self.name`?

leoll2: I think it's just redundant, `self.name` is already a string, the `str(...)` can be removed

druzhkov-paul:
@Ilya-Krylov Seems that these changes break OTE CLI registry, could you please take a look?

aiarkinx:
run ote-test

AlbertvanHouten:
run ote_sdk tests

leoll2 (dismissed):

ote_sdk/ote_sdk/entities/model_template.py
```diff
@@ -75,27 +75,138 @@ def __str__(self) -> str:
...
-    DETECTION = auto()
-    ANOMALY_DETECTION = auto()
-    CROP = auto()
-    TILE = auto()
-    INSTANCE_SEGMENTATION = auto()
-    ACTIVELEARNING = auto()
-    ANOMALY_SEGMENTATION = auto()
-    ANOMALY_CLASSIFICATION = auto()
-    ROTATED_DETECTION = auto()
+    def __init__(
```
Ilya-Krylov, line 104: https://docs.python.org/3/library/enum.html#when-to-use-new-vs-init

AlbertvanHouten: Ah, thanks. Fixed.

AlbertvanHouten:
run ote_sdk tests

druzhkov-paul (approved):

PR #985, opened 2022-03-18
move mmdet templates

Ilya-Krylov:


Ilya-Krylov:
run ote_sdk tests

Ilya-Krylov:
ote-test/1451/

Ilya-Krylov:
ote-test/1452/

Ilya-Krylov:
run ote_sdk tests

Ilya-Krylov:
ote-test/1453/

external/mmdetection/configs/custom-object-detection/readme.md
```diff
@@ -0,0 +1,12 @@
+# Custom object detector
```
druzhkov-paul: Not sure if these readme files have any relevance now. Ok to leave them as is, for sure.

ote_cli/ote_cli/tools/find.py
```diff
@@ -31,6 +31,7 @@ def parse_args():
         "--root", help="A root dir where templates should be searched.", default="."
     )
     parser.add_argument("--task_type")
+    parser.add_argument("--experimental", action="store_true")
```
druzhkov-paul, line 34: Maybe turn it on by default?

aiarkinx:
run ote-test

external/mmdetection/detection_tasks/apis/detection/__init__.py
```diff
@@ -0,0 +1,42 @@
+# Copyright (C) 2021 Intel Corporation
+#
```
druzhkov-paul, line 2: Just as a discussion point.

1. As we now have common set of tasks/utils/whatever for all mmdetection-powered templates, it looks safe to remove extra level of nesting (`detection`) here. But probably not in this PR.
2. Having `mmdetection` directory containing `detection_tasks` makes the story a bit complicated. Maybe we should not reference any particular algo backend in the folders' tree of OTE at all and leave it being hidden under the neutral `submodule`? I foresee difficulties if we end up having multiple detection backends though. Any thoughts on this?

druzhkov-paul (dismissed):

external/mmdetection/detection_tasks/apis/detection/__init__.py
Ilya-Krylov: 1. makes sense
2. not sure that I fully understand, but "external" and "mmdetection" (and other) names are kept here because of CI. I made this transition so that all paths such as "external/mmdetection/init_venv.sh", "external/mmdetection/tests", etc are not changed. I guess these paths are hardcoded, @LeonidBeynenson , right? Or are you, @druzhkov-paul , telling about different things?

ote_cli/ote_cli/tools/find.py
Ilya-Krylov: maybe, but in this case I would set it to False while checking ote_cli features like train, eval, export and etc. Otherwise testing will take much more longer. However experimental templates have to be tested as well, I don't know how it can be done during reasonable time.

external/mmdetection/configs/custom-object-detection/readme.md
Ilya-Krylov: yeah, these files as well as some quality evaluation tools placed near (experimental) templates should be revisited.

Ilya-Krylov:
run ote_sdk tests

Ilya-Krylov:
run ote_sdk tests

Ilya-Krylov:
ote-test/1468/ 🟡

ote_cli/ote_cli/tools/find.py
druzhkov-paul: This is testing code, this is a user-facing API. And IMO having this feature disabled here means that we don't want users to use "experimental" templates in OTE. So, I think about switching the default.

This being said, I agree that non-experimental templates should be tested in the first place, and experimental ones may be skipped to decrease CI load.

external/mmdetection/detection_tasks/apis/detection/__init__.py
druzhkov-paul: You captured my point right.

I see, that we have limitation at the CI side. @LeonidBeynenson could you please estimate how difficult and time-consuming would that be to support modified directories structure? I mean 100% transition, not "almost completed" state.

At the same time do we have an agreement that such renaming is a reasonable thing to do? Or it does not pay off taking into account all of the required changes in the infrastructure?

PR #993, opened 2022-03-21
[ANOMALY SEGMENTATION] Add multi-metric support to OTE SDK for anomaly tasks

djdameln:
## Description
This PR adds multi-metric support to OTE SDK for anomaly localization tasks (segmentation and detection).

**Changes on OTE SDK:**
- Performance.score is now a property instead of an attribute
- Add `AnomalyLocalizationPerformance` as new subclass of `Performance`
- Add `AnomalyLocalizationScores` class which computes the localization performance from a resultset.
- Add helper functions in `ote_sdk/ote_sdk/utils/dataset_utils` to split dataset into local and globally annotated items (moved from `ote_anomalib`).
- Add `compute_anomaly_localization_scores` to `MetricsHelper`.

**Changes to anomaly task implementation:**
- Remove resultset splitting logic from `evaluate` method and use `compute_anomaly_localization_scores` instead.

> NOTE to the OTE Maintainers
Just a bit of context to @Ilya-Krylov  and @druzhkov-paul, this PR has been created in line with the iAi and UI teams. This is the reason why @atwinand has been added as the reviewer.

## Build
- [BMM](https://ci.iotg.sclab.intel.com/job/IMPT/job/IMPTOps/28142/)


ote_sdk/ote_sdk/entities/metrics.py
```diff
@@ -702,3 +707,29 @@ def __repr__(self):
...
         return isinstance(other, NullPerformance)
+
+
+class AnomalyLocalizationPerformance(Performance):
+    """
+    This class is used to report multiple metrics in anomaly tasks that perform anomaly localization.
+    Local score takes priority as the primary score.
+    """
+
+    def __init__(self, global_score, local_score=None, dashboard_metrics=None):
```
atwinand: Type hints all function signatures please.

ote_sdk/ote_sdk/utils/dataset_utils.py
```diff
@@ -32,8 +32,16 @@ def split_local_global_dataset(dataset) -> Tuple[DatasetEntity, DatasetEntity]:
     for gt_item in dataset:
 
         annotations = gt_item.get_annotations()
-        global_annotations = [annotation for annotation in annotations if Rectangle.is_full_box(annotation.shape)]
-        local_annotations = [annotation for annotation in annotations if not Rectangle.is_full_box(annotation.shape)]
+        global_annotations = [
```
atwinand: When you do this split, do you change somewhat the original content of the resultset.ground_truth_dataset and resultset.prediction_dataset?

djdameln: We're splitting the resultset based on what is available in the ground truth dataset. Both the ground truth dataset and the prediction dataset in the original resultset contain a mixture of global and local labels. This causes problems in the metric computation, which always includes all available annotations in the computation. For example, a global (full bbox) annotation will effectively count as a segmentation polygon that masks the entire image when computing the dice score. This is why we need to split the resultset between partially (only global GT annotations) and fully (both local and global GT annotations) annotated items.

The new global dataset will contain all dataset items for which global annotations are available, and the local dataset will contain all dataset items for which local annotations are available. Moreover, the global dataset will only contain the global annotations, and the local dataset will only contain the local annotations. This holds for both the ground truth datasets and the prediction datasets.

To summarize, this is the structure of the original and new resultsets:

**Original resultset**
ground_truth_dataset:
	- global annotations
	- optional local annotations
	
prediction_dataset:
	- global predictions
	- local predictions
	
**New global resultset**
ground_truth_dataset:
	- global annotations
	
prediction_dataset:
	- global predictions
	
**New local resultset**
ground_truth_dataset:
	- local annotations

prediction_dataset:
	- local predictions
	
	
The original resultset will not be modified, because we are creating new dataset items when filling the new datasets. So the original resultset will still have the original dataset items which may contain local annotations in addition to global annotations. This makes sure that we can still use both global and local annotations and predictions when visualizing the results in the UI.

Hope that answers your question?

djdameln:
run ote_sdk tests

ote_sdk/ote_sdk/usecases/evaluation/anomaly_metrics.py
```diff
@@ -0,0 +1,60 @@
...
+        global_resultset, local_resultset = split_local_global_resultset(resultset)
+
+        global_metric = FMeasure(resultset=global_resultset)
+        global_performance = global_metric.get_performance()
+        self.global_score = global_performance.score
+
+        if contains_anomalous_images(local_resultset.ground_truth_dataset):
+            local_metric = DiceAverage(
+                resultset=local_resultset, average=MetricAverageMethod.MICRO
+            )
```
samet-akcay: I guess, we'll need another refactor here after merging the detection task. This is because the detection task uses FMeasure. To accommodate this, we would either need another class for detection or `task_type` argument in the constructor to assign the corresponding performance metric.

ote_sdk/ote_sdk/utils/dataset_utils.py
atwinand: Thanks Dick. Quite clear. And also, the AnomalyLocalizationPerformance will then be assigned to the original resultset, correct?

djdameln: correct

ote_sdk/ote_sdk/usecases/evaluation/anomaly_metrics.py
```diff
@@ -0,0 +1,60 @@
...
+        self.local_score: Optional[ScoreMetric] = None
+        self.dashboard_metrics: Optional[List[MetricsGroup]] = None
+
+        global_resultset, local_resultset = split_local_global_resultset(resultset)
+
+        global_metric = FMeasure(resultset=global_resultset)
+        global_performance = global_metric.get_performance()
+        self.global_score = global_performance.score
+
+        if contains_anomalous_images(local_resultset.ground_truth_dataset):
```
Ilya-Krylov, line 85: Is it allowed not to have anomalous images in dataset?

Ilya-Krylov: It it is allowed, then primary metric can have different kind different from run to run (if the dataset is changed between runs).
If it is not allowed, then it always has local_metric different from None.

ote_sdk/ote_sdk/entities/metrics.py
```diff
@@ -702,3 +707,34 @@ def __repr__(self):
 
     def __eq__(self, other):
         return isinstance(other, NullPerformance)
+
+
+class AnomalyLocalizationPerformance(Performance):
```
Ilya-Krylov: Why does it have a name related to Anomaly?
Is it actually needed to have such class instead of using score + dashboard metrics ?

ote_sdk/ote_sdk/usecases/evaluation/anomaly_metrics.py
djdameln: In an anomaly segmentation project, the user can upload some images but not make any segmentation annotations. In this case the global (image-level label) ground truth dataset will contain anomalous images, but the local (segmentation polygon) ground truth dataset will not contain any anomalous images. When we train the model, we still want to report some performance metric to the user, so we take global score as the primary metric. The user may then annotate some images, after which we can compute the local metric. This will then be the primary metric.

We do this because we always need to report a primary metric to show in the UI. In the future, the UI will show both global and local performance based on the availability. We could then change the AnomalyLocalizationPerformance to always return the local score if needed.

ote_sdk/ote_sdk/entities/metrics.py
djdameln: We need this class to support multi-metric projects. We initially looked at generic MultiScorePerformance implementations, but this is tricky because it would be hard to distinguish the different metrics from each other, especially if not all metrics can be reported at all times (like in the example above with only global, no local annotations). Since anomaly tasks are currently the only tasks that require multi-metric support, we decided to make this class specific to anomaly.

ote_sdk/ote_sdk/entities/metrics.py
```diff
@@ -702,3 +707,34 @@ def __repr__(self):
...
+    """
+    This class is used to report multiple metrics in anomaly tasks that perform anomaly localization.
+    Local score takes priority as the primary score.
+    """
+
+    def __init__(
```
druzhkov-paul, line 722: It would be good to implement a proper version of `__repr__` for this class.

ote_sdk/ote_sdk/usecases/evaluation/anomaly_metrics.py
```diff
@@ -0,0 +1,60 @@
...
+            local_metric = DiceAverage(
+                resultset=local_resultset, average=MetricAverageMethod.MICRO
+            )
+            local_performance = local_metric.get_performance()
+            self.local_score = local_performance.score
+            self.dashboard_metrics = local_performance.dashboard_metrics
```
druzhkov-paul: As an option you can combine both global and local dashboard metrics here to provide more information when local score is None.

ote_sdk/ote_sdk/usecases/evaluation/anomaly_metrics.py
samet-akcay: And a little bit motivation why we do this: We don't want to block the training if there is no detection/segmentation annotation available. We want to train the model as soon as the user uploads normal and abnormal images. If we forced the user to annotate to start the training, the user would be confused as we claim that anomaly tasks are unsupervised tasks.

ote_sdk/ote_sdk/utils/dataset_utils.py
```diff
@@ -137,7 +181,9 @@ def split_local_global_resultset(resultset) -> Tuple[ResultSetEntity, ResultSetE
                 locally_predicted.append(
                     DatasetItemEntity(
                         media=pred_item.media,
-                        annotation_scene=AnnotationSceneEntity(local_predictions, kind=AnnotationSceneKind.PREDICTION),
+                        annotation_scene=AnnotationSceneEntity(
+                            local_predictions, kind=AnnotationSceneKind.PREDICTION
+                        ),
```
druzhkov-paul: Code duplication level is pretty high here. Is it possible to provide some utility functions to get rid of it?

ote_sdk/ote_sdk/utils/dataset_utils.py
```diff
@@ -54,7 +66,9 @@ def split_local_global_dataset(dataset) -> Tuple[DatasetEntity, DatasetEntity]:
                 locally_annotated.append(
                     DatasetItemEntity(
                         media=gt_item.media,
-                        annotation_scene=AnnotationSceneEntity(local_annotations, kind=AnnotationSceneKind.ANNOTATION),
+                        annotation_scene=AnnotationSceneEntity(
+                            local_annotations, kind=AnnotationSceneKind.ANNOTATION
+                        ),
                         metadata=gt_item.metadata,
                         subset=gt_item.subset,
                         ignored_labels=gt_item.ignored_labels,
```
druzhkov-paul: What about ROIs?

druzhkov-paul (commented):
What about tests for the new SDK stuff?

Ilya-Krylov:
run ote_sdk tests

ote_sdk/ote_sdk/utils/dataset_utils.py
```diff
@@ -0,0 +1,191 @@
...
+        global_annotations = [
+            annotation
+            for annotation in item.get_annotations()
+            if Rectangle.is_full_box(annotation.shape)
+        ]
+        global_items.append(
+            DatasetItemEntity(
+                media=item.media,
+                annotation_scene=AnnotationSceneEntity(
+                    global_annotations, kind=AnnotationSceneKind.ANNOTATION
```
druzhkov-paul: What about predictions (not ground truth)? I guess, it does not matter for metric evaluation code and these datasets are only created for these purposes. Is it Ok to have `AnnotationSceneKind.ANNOTATION` even for predictions?

ote_sdk/ote_sdk/utils/dataset_utils.py
```diff
@@ -0,0 +1,191 @@
...
+        model=resultset.model,
+        ground_truth_dataset=local_gt_dataset,
+        prediction_dataset=local_pred_dataset,
+        purpose=resultset.purpose,
+    )
+    return global_resultset, local_resultset
+
+
+def contains_anomalous_images(dataset: DatasetEntity) -> bool:
+    """Find the number of local annotations in a resultset."""
```
druzhkov-paul: Maybe not the number of annotation?

Ilya-Krylov:
@djdameln could you please resolve conflicts and address remaining comments

samet-akcay:
@djdameln you could maybe point out the issue you're having here.

ote_sdk/ote_sdk/entities/metrics.py
djdameln: I refactored the class so that we now have a generic MultiScorePerformance class that supports adding one or more additional metrics, and a subclass AnomalyLocalizationPerformance which contains the global and local scores specific for anomaly tasks.

djdameln:
run ote_sdk tests

djdameln:
@Ilya-Krylov @druzhkov-paul I addressed your comments. Most importantly, I refactored the new class into a generic `MultiScorePerformance` and an anomaly-specific `AnomalyLocalizationPerformance`, and reduced code duplication in the dataset splitting utils.

One problem that we encountered is the fact that we can't use the resultset purpose property in OTE CLI. Ideally we would like to assign slightly different labels between inference and evaluation. During inference we want to provide a global normal label when no anomalous regions are found. During evaluation we don't want to provide this label, because the segmentation performance is based on a single-class metric. Currently the OTE CLI does not properly assign the purpose in the dataset and resultset classes, so we could not implement this for now. I fixed it temporarily by always adding the normal label, but this will lead to slightly inaccurate local performance metric. I suggest we merge this PR first and then see if we can address the resultset purpose issue. I created a ticket for this: https://jira.devtools.intel.com/browse/CVS-83052

druzhkov-paul:
run ote-test

aiarkinx:
run ote-test

djdameln:
run ote-test

Ilya-Krylov:
run ote_sdk tests

external/anomaly/ote_anomalib/callbacks/inference.py
```diff
@@ -63,17 +65,36 @@ def on_predict_epoch_end(self, _trainer: pl.Trainer, pl_module: AnomalyModule, o
...
-                    annotations=create_detection_annotation_from_anomaly_heatmap(
+                    create_detection_annotation_from_anomaly_heatmap(
                         hard_prediction=pred_mask,
                         soft_prediction=anomaly_map,
                         label_map=self.label_map,
                     )
                 )
+                # TODO: only add full normal label if dataset purpose is inference
+                if len(dataset_item.get_annotations()) == 1:
+                    dataset_item.append_annotations(
```
druzhkov-paul: IMO, `append_labels` is a better (right one?) way to do this.

external/anomaly/ote_anomalib/callbacks/inference.py
```diff
@@ -63,17 +65,36 @@ def on_predict_epoch_end(self, _trainer: pl.Trainer, pl_module: AnomalyModule, o
...
             if self.task_type == TaskType.ANOMALY_DETECTION:
                 dataset_item.append_annotations(
-                    annotations=create_detection_annotation_from_anomaly_heatmap(
+                    create_detection_annotation_from_anomaly_heatmap(
                         hard_prediction=pred_mask,
                         soft_prediction=anomaly_map,
                         label_map=self.label_map,
                     )
                 )
+                # TODO: only add full normal label if dataset purpose is inference
```
druzhkov-paul: Why can't you just filter that one later during metric calculation?

external/anomaly/ote_anomalib/callbacks/inference.py
```diff
@@ -63,17 +65,36 @@ def on_predict_epoch_end(self, _trainer: pl.Trainer, pl_module: AnomalyModule, o
...
+                if len(dataset_item.get_annotations()) == 1:
```
druzhkov-paul: What does that condition mean? Does it assume that annotation is empty before inference? Why?

external/anomaly/ote_anomalib/callbacks/inference.py
```diff
@@ -63,17 +65,36 @@ def on_predict_epoch_end(self, _trainer: pl.Trainer, pl_module: AnomalyModule, o
...
+                    dataset_item.append_annotations(
+                        [
+                            Annotation(
+                                Rectangle.generate_full_box(),
+                                labels=[ScoredLabel(label=self.normal_label, probability=0.5)],
```
druzhkov-paul: Why is it always 0.5?

external/anomaly/ote_anomalib/callbacks/inference.py
djdameln: That would also be possible, but since we are using generic helper functions for splitting the dataset in the `evaluate` method, we would still need to check the dataset purpose there.

druzhkov-paul: Can't say that I understood the point. Could you please elaborate a bit, in what cases DatasetItem RoI level normal label makes a problem?

djdameln: > Ideally we would like to assign slightly different labels between inference and evaluation. During inference we want to provide a global normal label when no anomalous regions are found. During evaluation we don't want to provide this label, because the segmentation performance is based on a single-class metric. Currently the OTE CLI does not properly assign the purpose in the dataset and resultset classes, so we could not implement this for now.

As I tried to explain here, having the normal label is problematic when computing the performance metric. The localization performance should be based on the anomalous class only, so any predictions with the `normal` label should be ignored. In the other processing stages (training, inference) however, we do need the normal label. This is why we need to be able to check the dataset/resultset purpose when assigning labels and/or when splitting the dataset into local and global components.

druzhkov-paul: > The localization performance should be based on the anomalous class only, so any predictions with the normal label should be ignored.

This is clear.

> In the other processing stages (training, inference) however, we do need the normal label. 

This is not.

djdameln: During training, we need the normal label so that the model can distinguish between normal and abnormal images. This is mainly because the model computes the adaptive threshold value during validation, which happens on the anomalib side right after training. To compute the adaptive threshold, the model needs to know which images are normal and which are abnormal.

During inference, we need to show the normal label to the user if no anomalous regions were found.

druzhkov-paul: I still miss understanding why can't you remove all normal labels from the predicted dataset right before the call to the metric evaluation function when needed.

djdameln: I see your point. It required some refactoring, but I am now filtering out the normal labels before computing the metrics.

djdameln:
run ote_sdk tests

druzhkov-paul:
run ote_sdk tests

druzhkov-paul:
run ote-test

djdameln:
Did something change in the mypy configuration? Since this morning I'm getting a whole list of mypy errors unrelated to my changes
```
external/anomaly/ote_anomalib/callbacks/progress.py:38: error: Invalid self argument "InferenceParameters" to attribute function "update_progress" with type "Callable[[int], None]"  [misc]
external/anomaly/ote_anomalib/callbacks/progress.py:118: error: Too many arguments  [call-arg]
external/anomaly/ote_anomalib/inference_task.py:97: error: Need type annotation for 'hyper_parameters'  [var-annotated]
external/anomaly/ote_anomalib/inference_task.py:189: error: Incompatible types in assignment (expression has type "AnomalyDetectionScores", variable has type "FMeasure")  [assignment]
external/anomaly/ote_anomalib/inference_task.py:191: error: Incompatible types in assignment (expression has type "AnomalySegmentationScores", variable has type "FMeasure")  [assignment]
external/anomaly/ote_anomalib/openvino.py:134: error: Need type annotation for 'ote_config'  [var-annotated]
external/anomaly/ote_anomalib/openvino.py:154: error: Invalid self argument "InferenceParameters" to attribute function "update_progress" with type "Callable[[int], None]"  [misc]
external/anomaly/ote_anomalib/openvino.py:154: error: Incompatible types in assignment (expression has type "Callable[[], None]", variable has type "Callable[[int], Any]")  [assignment]
external/anomaly/ote_anomalib/openvino.py:231: error: Incompatible types in assignment (expression has type "AnomalyDetectionScores", variable has type "FMeasure")  [assignment]
external/anomaly/ote_anomalib/openvino.py:233: error: Incompatible types in assignment (expression has type "AnomalySegmentationScores", variable has type "FMeasure")  [assignment]
external/anomaly/ote_anomalib/openvino.py:241: error: Need type annotation for 'hparams'  [var-annotated]
external/anomaly/ote_anomalib/openvino.py:287: error: Item "None" of "Optional[ModelEntity]" has no attribute "get_data"  [union-attr]
external/anomaly/ote_anomalib/openvino.py:288: error: Item "None" of "Optional[ModelEntity]" has no attribute "get_data"  [union-attr]
external/anomaly/ote_anomalib/openvino.py:301: error: Invalid self argument "OptimizationParameters" to attribute function "update_progress" with type "Callable[[int], None]"  [misc]
external/anomaly/ote_anomalib/openvino.py:301: error: Too many arguments  [call-arg]
external/anomaly/ote_anomalib/openvino.py:309: error: Invalid self argument "OptimizationParameters" to attribute function "update_progress" with type "Callable[[int], None]"  [misc]
external/anomaly/ote_anomalib/openvino.py:309: error: Too many arguments  [call-arg]
external/anomaly/ote_anomalib/openvino.py:317: error: Item "None" of "Optional[ModelEntity]" has no attribute "get_data"  [union-attr]
external/anomaly/ote_anomalib/openvino.py:318: error: Item "None" of "Optional[ModelEntity]" has no attribute "get_data"  [union-attr]
external/anomaly/ote_anomalib/openvino.py:319: error: Item "None" of "Optional[ModelEntity]" has no attribute "get_data"  [union-attr]
external/anomaly/ote_anomalib/openvino.py:320: error: Item "None" of "Optional[ModelEntity]" has no attribute "get_data"  [union-attr]
external/anomaly/ote_anomalib/openvino.py:330: error: Invalid self argument "OptimizationParameters" to attribute function "update_progress" with type "Callable[[int], None]"  [misc]
external/anomaly/ote_anomalib/openvino.py:330: error: Too many arguments  [call-arg]
external/anomaly/ote_anomalib/nncf_task.py:178: error: Argument "parameters" to "ProgressCallback" has incompatible type "Optional[OptimizationParameters]"; expected "Union[TrainParameters, InferenceParameters, None]"  [arg-type]
Found 24 errors in 4 files (checked 1 source file)
```

djdameln:
run ote_sdk tests

external/anomaly/ote_anomalib/openvino.py
```diff
@@ -172,16 +159,38 @@ def infer(self, dataset: DatasetEntity, inference_parameters: InferenceParameter
...
+                annotations = create_annotation_from_segmentation_map(
+                    pred_mask, anomaly_map.squeeze(), {0: self.normal_label, 1: self.anomalous_label}
+                )
+                dataset_item.append_annotations(annotations)
+                if len(annotations) == 0:
+                    dataset_item.append_labels([ScoredLabel(label=self.normal_label, probability=float(probability))])
+                else:
+                    dataset_item.append_labels(
+                        [ScoredLabel(label=self.anomalous_label, probability=float(probability))]
+                    )
```
druzhkov-paul:
```suggestion
                label = self.normal_label if len(annotations) == 0 else self.anomalous_label 
                dataset_item.append_labels([ScoredLabel(label=label, probability=float(probability))])
```
and `dataset_item.append_labels` can be moved to the outside of the if-else clause, because it's the same for all branches.

Ilya-Krylov:
run ote_sdk tests

ote_sdk/ote_sdk/utils/dataset_utils.py
```diff
@@ -0,0 +1,199 @@
...
+            or len(local_annotations) > 0
+        ):
+            local_idx.append(idx)
+    return local_idx
+
+
+def get_local_subset(
+    dataset: DatasetEntity,
+    fully_annotated_idx: Optional[List[int]] = None,
+    include_normal: bool = True,
```
druzhkov-paul, line 54: This argument is not documented.

druzhkov-paul (dismissed):

Ilya-Krylov:
run ote_sdk tests

external/anomaly/ote_anomalib/openvino.py
djdameln: Good catch. I've made the changes

druzhkov-paul (dismissed):

druzhkov-paul:
run ote_sdk tests

druzhkov-paul:
run ote-test

djdameln:
run ote_sdk tests

djdameln:
run ote-test

druzhkov-paul (approved):

PR #1023, opened 2022-03-29
[OTE_SDK] Fix calculation of auto hpo support in ModelTemplate

ljcornel:
This PR fixes a mistake in the calculation of the `supports_auto_hpo` function in the ModelTemplate. It now correctly returns True for algorithms that support auto HPO.


ljcornel:
run ote_sdk tests

druzhkov-paul:
run ote-test

druzhkov-paul (approved):

PR #1027, opened 2022-03-30
Aa/text to speech

a-a-egorovich:
draft


AlexeyKruglov:
@a-a-egorovich  Now code is more-or-less ready from my side.  Please, check what everything still works and I didn't break anything. And I changed `--model` parameter in `demo.py` from optional to required -- because it will fail without it at `os.path.join()` with `None`.

a-a-egorovich:
run ote_sdk tests

goodsong81:
Closing this PR as no one is in charge.

PR #1054, opened 2022-04-12
OTE Docker container

0de554K:
Docker container for OTE


ekaakurniawan:
I added [PR-1115](https://github.com/openvinotoolkit/training_extensions/pull/1115) for Docker image that uses Intel optimized PyTorch. It is under `docker/cpu` directory. Please take a look if we can sync the structure and arguments.

PR #1076, opened 2022-04-25
[CVS-84226] Apply constraints for pip install

goodsong81:
Remedy for https://jira.devtools.intel.com/browse/CVS-84226 was applied to constraint.txt
but it is not being applied to actual init_venv.sh.
Updating other backends, too.


goodsong81:
run ote-test

PR #1099, opened 2022-05-11
[HPO] Enabling ote_anomalib report score for hpopt

SangdaeNam:
This PR contains adding validation_epoch_end function in ProgressCallback in ote_anomalib.
In hpo flow, score in every validation epoch end report to hpopt with order below
1. In ProgressCallback, update_progress_callback is HPOCallback delivered from train_task in ote_anomalib
2. In every validation epoch end, the score produced by metric deliver to update_progress_callback
3. When update_progress_callback(which is HPOCallback) is called, it reports the score to hpopt
4. With the score delivered, hpopt find new hyperparameters and report them

Furthermore, we are wondering if this modified ProgressCallback will work well in anomaly_detection and anomaly_segmentation

CI result: https://ci-ote.iotg.sclab.intel.com/job/ote/job/pr-ote-test/344/

### DIFF
--- a/external/anomaly/ote_anomalib/callbacks/progress.py
+++ b/external/anomaly/ote_anomalib/callbacks/progress.py
@@ -112,6 +112,15 @@ class ProgressCallback(ProgressBar):
             raise ValueError(f"Unknown stage {stage}. Available: train, predict and test")
 
         return self._progress
+    
+    def on_validation_epoch_end(self, trainer, pl_module):
+        super().on_validation_epoch_end(trainer, pl_module)
+        score = None
+        metric = self.update_progress_callback.metric if hasattr(self.update_progress_callback, 'metric') else None
+        if metric in list(trainer.logged_metrics.keys()):
+            score = float(trainer.logged_metrics[metric])
+        progress = int(self._get_progress('predict'))
+        self.update_progress_callback(progress=progress, score=score)
 
     def _update_progress(self, stage: str):
         progress = self._get_progress(stage)
--- a/external/anomaly/ote_anomalib/callbacks/progress.py
+++ b/external/anomaly/ote_anomalib/callbacks/progress.py
@@ -113,13 +113,13 @@ class ProgressCallback(ProgressBar):
 
         return self._progress
     
-    def on_validation_epoch_end(self, trainer, pl_module):
-        super().on_validation_epoch_end(trainer, pl_module)
+    def on_train_epoch_end(self, trainer, pl_module):
+        super().on_train_epoch_end(trainer, pl_module)
         score = None
         metric = getattr(self.update_progress_callback, 'metric', None)
         if metric in trainer.logged_metrics:
             score = float(trainer.logged_metrics[metric])
-        progress = int(self._get_progress('predict'))
+        progress = int(self._get_progress('train'))
         self.update_progress_callback(progress=progress, score=score)
 
     def _update_progress(self, stage: str):
--- a/external/anomaly/ote_anomalib/callbacks/progress.py
+++ b/external/anomaly/ote_anomalib/callbacks/progress.py
@@ -112,15 +112,6 @@ class ProgressCallback(ProgressBar):
             raise ValueError(f"Unknown stage {stage}. Available: train, predict and test")
 
         return self._progress
-    
-    def on_train_epoch_end(self, trainer, pl_module):
-        super().on_train_epoch_end(trainer, pl_module)
-        score = None
-        metric = getattr(self.update_progress_callback, 'metric', None)
-        if metric in trainer.logged_metrics:
-            score = float(trainer.logged_metrics[metric])
-        progress = int(self._get_progress('train'))
-        self.update_progress_callback(progress=progress, score=score)
 
     def _update_progress(self, stage: str):
         progress = self._get_progress(stage)
--- a/external/anomaly/ote_anomalib/callbacks/progress.py
+++ b/external/anomaly/ote_anomalib/callbacks/progress.py
@@ -116,7 +116,7 @@ class ProgressCallback(ProgressBar):
     def on_validation_epoch_end(self, trainer, pl_module):
         super().on_validation_epoch_end(trainer, pl_module)
         score = None
-        metric = self.update_progress_callback.metric if hasattr(self.update_progress_callback, 'metric') else None
+        metric = getattr(self.update_progress_callback, 'metric', None)
         if metric in list(trainer.logged_metrics.keys()):
             score = float(trainer.logged_metrics[metric])
         progress = int(self._get_progress('predict'))
--- a/external/anomaly/ote_anomalib/callbacks/progress.py
+++ b/external/anomaly/ote_anomalib/callbacks/progress.py
@@ -117,7 +117,7 @@ class ProgressCallback(ProgressBar):
         super().on_validation_epoch_end(trainer, pl_module)
         score = None
         metric = getattr(self.update_progress_callback, 'metric', None)
-        if metric in list(trainer.logged_metrics.keys()):
+        if metric in trainer.logged_metrics:
             score = float(trainer.logged_metrics[metric])
         progress = int(self._get_progress('predict'))
         self.update_progress_callback(progress=progress, score=score)
--- a/external/anomaly/ote_anomalib/callbacks/score_report.py
+++ b/external/anomaly/ote_anomalib/callbacks/score_report.py
@@ -1,0 +1,43 @@
+"""Score reporting callback"""
+
+# Copyright (C) 2020 Intel Corporation
+#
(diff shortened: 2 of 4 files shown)


external/anomaly/ote_anomalib/callbacks/progress.py
```diff
@@ -112,6 +112,15 @@ def _get_progress(self, stage: str = "train") -> float:
...
         return self._progress
+    
+    def on_validation_epoch_end(self, trainer, pl_module):
+        super().on_validation_epoch_end(trainer, pl_module)
+        score = None
+        metric = getattr(self.update_progress_callback, 'metric', None)
+        if metric in trainer.logged_metrics:
+            score = float(trainer.logged_metrics[metric])
+        progress = int(self._get_progress('predict'))
+        self.update_progress_callback(progress=progress, score=score)
```
ashwinvaidya17: Not sure why `update_progress_callback` is called from here. We only update progress for train, test or predict. I feel this will have unintended consequences such as progress jumping between x%-100%. Has this been tested in the UI? We removed val step just to avoid this issue. https://github.com/openvinotoolkit/training_extensions/commit/95126f6fd6731f89921731ba901d1e150c72aca0
In case this works in the UI then I am fine with it.

yunchu: I agree with Ashwin and 'predict' progress should not be used at the validation event as well. how about modifying like this:
1. grep the metric score from the trainer.logged_metrics on this event and keep it to the instance (self)
2. and then pass it through on_train_batch_end() event and clear the saved one from the instance to prevent redundant reporting the score.

anyway, @ashwinvaidya17 @samet-akcay I think the progress reporting at every 'batch' end seems too frequent, isn't it? AFAIK other task aligned with 'epoch' event.

SangdaeNam: Yes I understood your comment. We checked again and we found that if we call "update_progress_callback" in the function "on_train_epoch_end", the score comes after validation step end.
So how about replacing "on_validation_epoch_end" to "on_train_epoch_end" and pass "train" as argument of "_get_progress"?
We think it will work well in both hpopt and UI.

samet-akcay: @SangdaeNam, thank you for the PR. 

I have a question: can you tell me why you want to add hpo update to the `ProgressCallback`? Unless I miss something here, I feel like HPO and ProgressBar serve for different purposes. 

Would it be an idea to create a separate callback for HPO itself? If we have a HPO callback, we could add it to the list of callbacks in the training [here](https://github.com/openvinotoolkit/training_extensions/blob/develop/external/anomaly/ote_anomalib/train_task.py#L53)

And our callbacks, would become
```
callbacks = [
    ProgressCallback(parameters=train_parameters), 
    MinMaxNormalizationCallback(), 
    HPOCallback(),
]
```

Am I missing something? Can you please clarify?

eunwoosh: Hi Samet, thank you for the comment!
HPO actually utilizes ote_sdk as much as possible for using common interface for all tasks.
HPOCallback is inherited from UpdateProgressCallback class as one of this direction.
For this reason, HPO uses ProgressCallback by being passed as attribute of "train_parameters" argument for reporting.
This usage is also commented in UpdateProgressCallback class.
If you have any concerns to use ProgressCallback, please let me know about that.
We'll try our best to resolve that concerns.

eunwoosh: We pushed commit as we said above.
Could you comment for this modification, @samet-akcay @ashwinvaidya17?

ashwinvaidya17: Can you point me to the code for HPOCallback? I still feel that HPOCallback should be a separate thing. Now, if we use HPOCallback on metrics from train_epoch_end, we will not be using val set for hyperparameter optimization. And I feel that we might overfit the params on the train set. If we decouple HPOCallback and Progress callback then we can safely add val_epoch_end to the model without messing with the progress on the UI.

yunchu: @ashwinvaidya17, you can find that from here: https://github.com/openvinotoolkit/training_extensions/blob/db0570244bebd3effc664a31be0b303692e8f447/ote_cli/ote_cli/utils/hpo.py#L225

external/anomaly/ote_anomalib/callbacks/score_report.py
```diff
@@ -0,0 +1,43 @@
...
+    Callback for reporting score.
+    """
+
+    def __init__(self, parameters: Optional[TrainParameters] = None) -> None:
+        if parameters is not None:
+            self.score_reporting_callback = parameters.update_progress
+        else:
+            self.score_reporting_callback = None
+
+    def on_train_epoch_end(self, trainer, pl_module):
```
ashwinvaidya17: I think this can be safely changed back to `on_validation_epoch_end`

ashwinvaidya17:
Probably out of scope for this PR but it would be nice to have a separate field in `TrainParameters` for hpo instead of replacing the update_progress object. I have pushed the proposal here https://github.com/openvinotoolkit/training_extensions/compare/develop...av/refactor_proposal the changes here will have to be propagated to other tasks such as detection and segmentation. But the idea behind this is to start a discussion.

The diff looks big but the key lines are
1. https://github.com/openvinotoolkit/training_extensions/compare/develop...av/refactor_proposal#diff-5ebeab25c618a314b6c5aae1533f78bf52952aa8162d9a605b2722c6d26ab13eR38
2. https://github.com/openvinotoolkit/training_extensions/compare/develop...av/refactor_proposal#diff-b92b2e145585581d8620bb4a723a53fdfd3147e6f5315f9ecce049c153a9533aR1-R26
3. https://github.com/openvinotoolkit/training_extensions/compare/develop...av/refactor_proposal#diff-b92b2e145585581d8620bb4a723a53fdfd3147e6f5315f9ecce049c153a9533aR1-R26
4. https://github.com/openvinotoolkit/training_extensions/compare/develop...av/refactor_proposal#diff-29ea76f6a531718a62189344389ade91312b587b765314c2fb7af1be396abdf8R55
5. https://github.com/openvinotoolkit/training_extensions/compare/develop...av/refactor_proposal#diff-57f84e87ecf6bedddb537d000409e2d170ec5e1b513e96b5c554aca8e609cd1eR1-R42

Another idea is to have a list of callbacks in `ote_sdk/ote_sdk/entities/train_parameters.py` so that we can support more callbacks from ote side in the future.

```python
@dataclass
class TrainParameters:

    resume: bool = False
    callbacks: List[Callback] = [] # from ote_sdk.usecases.reporting.callback import Callback
    save_model: Callable[[], None] = default_save_model_callback
```

There is no need to inherit anything in HPOCallback on the ote side.

```python
class HpoCallback():
    """Callback class to report score to hpopt"""

    def __init__(self, hp_config, metric, hpo_task):
        super().__init__()
        self.hp_config = hp_config
        self.metric = metric
        self.hpo_task = hpo_task

    def __call__(self, score: Optional[float] = None):
        if score is not None:
            if hpopt.report(config=self.hp_config, score=score) == hpopt.Status.STOP:
                self.hpo_task.cancel_training()
    
```

external/anomaly/ote_anomalib/callbacks/score_report.py
eunwoosh: It's right. I'll change it as you said.

yunchu:
@ashwinvaidya17, @samet-akcay, if latest update looks good for you guys, please leave a comment. i think i can approve this PR.

eunwoosh:
@ashwinvaidya17  I agree to your opinion. or maybe it's also better to be able to apply new callback as necessary. But you know, modification of ote_sdk is quite a big task. It would be nice to have conversation when planning next step of ote_sdk.

ashwinvaidya17 (dismissed):
Looks fine to me but I'll wait for @samet-akcay's feedback

samet-akcay (dismissed):
Looks good, thanks a lot for the effort!

samet-akcay:
@ashwinvaidya17's proposal would make it clearer on the ote side. of course, this would require some discussions and agreements before the implementation. @goodsong81, FYI

samet-akcay:
Would be good to schedule a BMM build to ensure everything is fine on the UI side.

yunchu:
run ote_sdk tests

yunchu:
@samet-akcay if you mean the SC UI, it requires other changes to enable it for SC. will be addressed by other PR later.

yunchu:
run ote-test

yunchu (dismissed):

eunwoosh:
run ote-test

eunwoosh:
run ote-test

goodsong81 (approved):
LGTM.
Failures are irrelevant to this change.

PR #1101, opened 2022-05-11
[OTE_SDK] Removed labels from shape and update append labels method

AlbertvanHouten:
- Removed labels from shape since they are never used, labels are added to Annotation instead. 
- Fixed a bug in dataset item.append_labels where the full annotation was compared instead of just the shape. In taskchains, this would lead to two shapes being produced, one for the first task with the first task label and another almost identical shape with the addition of the second tasks label. https://jira.devtools.intel.com/browse/CVS-83880


ote_sdk/ote_sdk/entities/annotation.py
```diff
@@ -23,42 +21,17 @@ class Annotation(metaclass=abc.ABCMeta):
...
         self.__shape = shape
         self.__labels = labels
 
     def __repr__(self):
         return (
             f"{self.__class__.__name__}("
             f"shape={self.shape}, "
-            f"labels={self.get_labels(True)}, "
-            f"id={self.id_})"
+            f"labels={self.get_labels(True)}"
```
JortBergfeld: I think for this call and the other calls to get_labels it is nice to include keyword arguments, for example "get_labels(include_empty=True)", because I do not know what the True stands for here.

ote_sdk/ote_sdk/tests/entities/shapes/test_rectangle.py
```diff
@@ -464,29 +453,17 @@ def test_rectangle_generate_full_box(self):
...
 
-        <b>Input data:</b>
-        Labels specified for full_box instance of Rectangle class
-
         <b>Expected results:</b>
         Test passes if generate_full_box method returns instance of Rectangle class with coordinates
         (x1=0.0, y1=0.0, x2=1.0, y2=1.0)
 
         <b>Steps</b>
         1. Check generate_full_box method for Rectangle instance with no labels specified
```
JortBergfeld: Probably should update this line as well (remove with no labels specified)

JortBergfeld (dismissed):
Looks good to me, just two small comments.

ote_sdk/ote_sdk/entities/annotation.py
AlbertvanHouten: Agreed, didn't really touch it because I was just deleting functionality, but I've addressed the issues in the shapes/annotation/dataset now.

JortBergfeld (dismissed):

AlbertvanHouten:
After discussion with UI team, I have reverted removal of id in annotation. They require it when merging together predictions/annotations. I've opted to compare the shape in dataset_item.append_label instead of the full annotation

ljcornel (approved):
Looks good to me!

ote_sdk/ote_sdk/entities/annotation.py
```diff
@@ -188,14 +188,14 @@ def __repr__(self):
...
         Returns the ID of the AnnotationSceneEntity.
         """
         return self.__id_
 
     @id_.setter
-    def id_(self, value):
+    def id_(self, value) -> None:
         self.__id_ = value
 
     @property
```
JortBergfeld, line 201: Line 201-208 is deprecated, did you really want to revert the removal of these lines?

JortBergfeld (approved):
Looks good to me, but check if the revert of removing the id went well, see my comment.

ote_sdk/ote_sdk/entities/annotation.py
AlbertvanHouten: It has been deprecated, but I am unsure if all usages in SC have been updated to use the new id_ property.

ljcornel: Just did a quick search and unfortunately I still found 4 usages in SC

goodsong81:
run ote-test

goodsong81:
> run ote-test
Build: https://ci-ote.iotg.sclab.intel.com/job/ote/job/pr-ote-test/295/
Report: http://validationreports.sclab.intel.com:8006/reports/build_number_report?test_session_build_number=pr-ote-test-295&environment=iotg-dev-workstation-08

goodsong81 (approved):
LGTM

PR #1128, opened 2022-06-07
[OTE][MPA] Apply ignored_labels to C-IL for detection

harimkang:
### This PR includes :
- Add 'ignored_labels' for each item in the OTEDataset used for the detection task
- MPA task modifies 'meta_keys' (in 'Collect') so that 'ignored_labels' are passed through 'img_metas'


harimkang:
run ote-test

chuneuny-emily (commented):
Is matching by id appropriate? How about getting the id by name?

harimkang:
> Is matching by id appropriate? How about getting the id by name?

@chuneuny-emily The ignored_label is converted based on the id-specific index of the label, which is the same as the index of the classes used for training (index delivered to the gt_label is converted in the same way)

chuneuny-emily:
> he ignored_label is converted based on the id-specific index of the label,

I see. If ids and names are always matched, it's okay. I just thought even if the id changes, I thought that I should get the matching indx in the label by name(car, person..).

chuneuny-emily (dismissed):
It looks good for me.

external/mmdetection/detection_tasks/extension/datasets/mmdataset.py
```diff
@@ -121,17 +121,19 @@ def __len__(self):
...
             Prepare a dict 'data_info' that is expected by the mmdet pipeline to handle images and annotations
-            :return data_info: dictionary that contains the image and image metadata, as well as the labels of the objects
-                in the image
+            :return data_info: dictionary that contains the image and image metadata, as well as the labels of
+            the objects in the image
             """
 
             dataset = self.ote_dataset
             item = dataset[index]
+            label_idx = {label.id: i for i, label in enumerate(self.labels)}
```
goodsong81: This line does not need to be run for each get_item() call.
Please move it to init() or somewhere and then reuse label_idx.

external/model-preparation-algorithm/mpa_tasks/apis/detection/task.py
```diff
@@ -270,6 +270,17 @@ def patch_color_conversion(pipeline):
...
+                    if pipeline_step.get('meta_keys', False):
+                        meta_keys = list(pipeline_step['meta_keys'])
+                        meta_keys.append('ignored_labels')
+                    else:
+                        meta_keys = (
+                            'filename', 'ori_filename', 'ori_shape',
+                            'img_shape', 'pad_shape', 'scale_factor', 'flip',
+                            'flip_direction', 'img_norm_cfg', 'ignored_labels'
+                        )
+                    pipeline_step['meta_keys'] = set(meta_keys)
```
goodsong81: Does this need to be verbose like this? Is there some cases where 'meta_keys' is absent?

goodsong81 (changes requested):
Please look into the comments; and update MPA commit to the latest one.

external/mmdetection/detection_tasks/extension/datasets/mmdataset.py
harimkang: Move label_idx to init and set self.label_idx

external/model-preparation-algorithm/mpa_tasks/apis/detection/task.py
harimkang: @goodsong81 @chuneuny-emily 
It is added to 'img_metas' based on the key in 'meta_keys'.
Currently, this parameter (Collect meta_keys) has a default value (when 'meta_keys' is absent) of 
('filename', 'ori_filename', 'ori_shape', 'img_shape', 'pad_shape', 'scale_factor', 'flip', 'flip_direction', 'img_norm_cfg') for the mmdet.datasets.pipelines.formatting.Collect criteria. 

Here, we need to put 'ignored_labels' in the parameter 'meta_keys', and I chose this method to keep the existing keys and put 'ignored_labels'. 

In fact, the best way is to modify mmdet.datasets.pipelines.formatting.Collect default 'meta_keys' directly. Do you have any other ideas?

harimkang: Since detection and segmentation use the same code, I changed it to call with a function in BaseTask, and modified the code as discussed.

chuneuny-emily: Looks good. Once it's merged, then I'll rebase it and call the function.

goodsong81 (approved):
LGTM

chuneuny-emily (approved):

PR #1137, opened 2022-06-15
[Anomaly Task] Add extra log statement

ashwinvaidya17:
# Description

Should be merged along with PR: https://github.com/intel-innersource/frameworks.ai.interactive-ai-workflow.sonoma-creek/pull/1742
BMM: 10.91.242.163

# Changes

- Adds an extra log statement to anomaly task
- Targets a new tag in anomalib

### DIFF
--- a/external/anomaly/ote_anomalib/inference_task.py
+++ b/external/anomaly/ote_anomalib/inference_task.py
@@ -193,19 +193,23 @@ class AnomalyInferenceTask(IInferenceTask, IEvaluationTask, IExportTask, IUnload
             raise ValueError(f"Unknown task type: {self.task_type}")
         output_resultset.performance = metric.get_performance()
 
+        logger.info(output_resultset)
+
         # print gt labels
         for gt_item in output_resultset.ground_truth_dataset:
-            logger.info(
-                f"GT: [Name: {gt_item.get_shapes_labels()[0].name}, "
-                f"is_anomalous: {gt_item.get_shapes_labels()[0].is_anomalous}], "
-            )
+            if len(gt_item.get_shapes_labels()) > 0:
+                logger.info(
+                    f"GT: [Name: {gt_item.get_shapes_labels()[0].name}, "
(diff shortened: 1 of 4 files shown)


external/anomaly/ote_anomalib/inference_task.py
```diff
@@ -193,19 +193,23 @@ def evaluate(self, output_resultset: ResultSetEntity, evaluation_metric: Optiona
...
+                    f"is_anomalous: {gt_item.get_shapes_labels()[0].is_anomalous}], "
+                )
 
         # print pred labels
         for pred_item in output_resultset.prediction_dataset:
-            logger.info(
-                f"Pred: [Name: {pred_item.get_shapes_labels()[0].name}, "
-                f"is_anomalous: {pred_item.get_shapes_labels()[0].is_anomalous}]"
-            )
+            if len(pred_item.get_shapes_labels()) > 0:
```
djdameln: This would prevent us from comparing the GT and predictions. We wouldn't know which GT item corresponds to which pred item.

ashwinvaidya17: That's there but right now I am trying to just see the label outputs. Maybe a better option would be to add try statement within zip. I'll also try experimenting with `Debugging in Kube Pod` section of the `Build-Debug-SC.md`

djdameln: Since we've already confirmed that the `is_anomalous` attribute is not the issue, maybe we can use the zip statement, and just print the full list of labels instead of name and `is_anomalous`. This way we'll be able to compare pred and GT, and also see which items don't have any labels.

ashwinvaidya17: Makes sense

djdameln: I'll push an update

ashwinvaidya17: Thanks

PR #1138, opened 2022-06-15
Add label source to scored label

JortBergfeld:
Add label source to scored label and update tests. The label source indicates either who made the annotation/label, or which model created the prediction/label.

### DIFF
--- a/ote_sdk/ote_sdk/entities/scored_label.py
+++ b/ote_sdk/ote_sdk/entities/scored_label.py
@@ -6,10 +6,31 @@
 
 import datetime
 import math
+from dataclasses import dataclass
+from typing import Optional
 
 from ote_sdk.entities.color import Color
 from ote_sdk.entities.id import ID
 from ote_sdk.entities.label import Domain, LabelEntity
+
+
+@dataclass
+class LabelSource:
+    """
+    This dataclass contains information about the source of a scored label.
+
+    For annotations, the id of the user who created the label and for predictions, the
+    id and model storage id of the model that created the prediction. When a user has
+    accepted a predictions as is, both the user id of the user who accepted and the
+    model/model storage id of the model that predicted should be filled in.
+
+    Note that usually the LabelSource is created with default values and the source is
+    set at a later stage.
+    """
+
+    user_id: str = ""
+    model_id: ID = ID()
+    model_storage_id: ID = ID()
 
 
 class ScoredLabel:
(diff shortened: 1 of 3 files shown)


leoll2:
run ote-test

ote_sdk/ote_sdk/entities/scored_label.py
```diff
@@ -6,25 +6,55 @@
...
+    """
+    This dataclass contains information about the source of a scored label.
+
+    For annotations, the id of the user who created the label and for predictions, the
+    id and model storage id of the model that created the prediction. When a user has
+    accepted a predictions as is, both the user id of the user who accepted and the
+    model/model storage id of the model that predicted should be filled in.
+
+    Note that usually the LabelSource is created with default values and the source is
+    set at a later stage.
```
leoll2: This comment doesn't fit here imho, the low-level SDK entities should be unaware of the application code that uses them.

leoll2 (commented):
LGTM, waiting for tests before approving.
Some questions:
- Who is responsible for updating the `label_source` value? OTE, SC or both?
- Have you considered the possibility of implementing this information in SC only, not OTE? Afaik, it's not strictly needed by OTE, and fields like "user" are a bit out of context in OTE. On the other hand, if OTE creates the predictions, it's more efficient to set the right value there rather than later in SC.

JortBergfeld:
> LGTM, waiting for tests before approving. Some questions:
> 
>     * Who is responsible for updating the `label_source` value? OTE, SC or both?
> 
>     * Have you considered the possibility of implementing this information in SC only, not OTE? Afaik, it's not strictly needed by OTE, and fields like "user" are a bit out of context in OTE. On the other hand, if OTE creates the predictions, it's more efficient to set the right value there rather than later in SC.

@leoll2 SC is indeed responsible for setting this information and we have considered putting the LabelSource in SC, but due to how we set up the code of the SDK and SC, it is no longer possible to properly add it to the SC code without a redesign of the SDK's. AnnotationScene (SC) inherits from AnnotationSceneEntity (OTE), which has a list of Annotations (OTE), which has a list of ScoredLabels (OTE), therefore adding LabelSource info to/about the ScoredLabel and making sure all methods properly deal with it is hard without adding it to the OTE_SDK. We have discussed the possible solutions for this problem, but all of them require a huge refactoring. We still want to do this, more generally move all information that OTE does not need to SC, but we will make a separate ticket/epic for that.

leoll2 (dismissed):
Thanks for the detailed explanation, now I see the problem. Hopefully in the future we will decouple the two SDKs in a better fashion.

AlbertvanHouten (dismissed):

leoll2 (dismissed):

JortBergfeld:
run ote-test

atwinand (dismissed):
LGTM!

goodsong81 (changes requested):
There seems to be some test failures regarding this change.
Could you resolve them or confirm they are not related to this PR?
Build: https://ci-ote.iotg.sclab.intel.com/job/ote/job/pr-ote-test/393/
Failures:
- http://validationreports.sclab.intel.com:8006/reports/test_result/62b1a18c83a7dafed5ece8ee
- http://validationreports.sclab.intel.com:8006/reports/test_result/62b1a19283a7dafed5ecea0a

goodsong81:
run ote-test

leoll2 (approved):

AlbertvanHouten (approved):

goodsong81 (approved):
LGTM.

PR #1148, opened 2022-07-02
[CVS-86879] Assertion Error: While training the models

JihwanEom:
### Summary
- Disable get_leaf_labels which is used for hierarchical classification in exportable code


harimkang:
This PR seems to be related to https://github.com/openvinotoolkit/deep-object-reid/pull/222
so is there no need to replace commit?

JihwanEom:
Yes, we need but I postponed commit update due to the hash is being different if torchreid pr is merged.
I'll update torchreid commit after merge torchreid PR.

harimkang:
I think it would be appropriate to request approval again after deep-reid's PR is merged :)

PR #1149, opened 2022-07-04
Bugfix for on_epoch_end function of TrainingProgressCallback.

eunwoosh:
This PR fixes a bug which occurs when MPA classification task runs.
The trick code is executed only if HPO callback is used.
Code checks that "hp_config" attribute exists in callback class because HPO callback function is not in HPOpt but in hpo.py.


JihwanEom (approved):
LGTM. Both torchreid and mpa classification work well.

goodsong81 (approved):
LGTM.

PR #1159, opened 2022-07-08
Enable hierarchical classification in MPA

JihwanEom:
### Summary
- Disable torchreid jumping in MPA when the task runs hierarchical classification
- Add hierarchical model templates for MPA
- Add hierarchical training capacity in MPA 
- Revised classification inference for stable operation
- Integrated separated load_annotation method to one method in mpa_cls_dataset.py
- Set drop_last=True when run hierarchical classification
- Support hierarchical classification CLI
- Relocation classification sample datasets in one folder


external/model-preparation-algorithm/mpa_tasks/apis/classification/task.py
```diff
@@ -58,38 +58,29 @@ def __init__(self, task_environment: TaskEnvironment):
...
-
-        for key, val in rename_dict.items():
-            model_template.model_template_path = model_template.model_template_path.replace(key, val)
-        model_template.entrypoints.base = 'torchreid_tasks.train_task.OTEClassificationTrainingTask'
-        model_template.framework = 'OTEClassification v1.2.3'
-        return task_env
+            self._hierarchical_info = get_hierarchical_info(task_environment.label_schema)
+        
+        if not self._multilabel and not self._hierarchical:
+            self._multiclass = True
```
harimkang: I don't think we need this variable, what do you think? self._multilabel and self._hierarchical is False, the remaining cases are It's true that it's a multi-class, but it doesn't seem very useful.

external/model-preparation-algorithm/mpa_tasks/apis/classification/task.py
```diff
@@ -101,29 +92,69 @@ def infer(self,
...
+                            label_str = self._hierarchical_info['all_groups'][self._hierarchical_info['num_multiclass_heads'] + i][0]
+                            ote_label = next(x for x in self._labels if x.name == label_str)
+                            label.append(ScoredLabel(label=ote_label, probability=float(head_logits[i])))
             else:
                 label_idx = prediction_item.argmax()
                 cls_label = ScoredLabel(self._labels[label_idx], probability=float(prediction_item[label_idx]))
                 label.append(cls_label)
             dataset_item.append_labels(label)
             update_progress_callback(int(i / dataset_size * 100))
+        # for idx, item in enumerate(dataset):
```
harimkang: Do you need this annotation?

external/model-preparation-algorithm/mpa_tasks/apis/classification/task.py
```diff
@@ -101,29 +92,69 @@ def infer(self,
...
 
     def evaluate(self,
                  output_result_set: ResultSetEntity,
                  evaluation_metric: Optional[str] = None):
         logger.info('called evaluate()')
         metric = MetricsHelper.compute_accuracy(output_result_set)
+
+        gt_dataset = output_result_set.ground_truth_dataset
+        pred_dataset = output_result_set.prediction_dataset
+        # for idx, gt_item in enumerate(gt_dataset):
```
harimkang: It's the same here.

external/model-preparation-algorithm/mpa_tasks/apis/classification/task.py
```diff
@@ -201,10 +228,15 @@ def _init_model_cfg(self):
         base_dir = os.path.abspath(os.path.dirname(self.template_file_path))
         if self._multilabel:
             cfg_path = os.path.join(base_dir, 'model_multilabel.py')
-        else:
+        elif self._hierarchical:
+            cfg_path = os.path.join(base_dir, 'model_hierarchical.py')
+        elif self._multiclass:
```
harimkang: I think we can just use this as an else.

harimkang (changes requested):
I didn't look at all the codes because they didn't seem to be organized. 
It seems good to request a review after code organization.

JihwanEom:
> I didn't look at all the codes because they didn't seem to be organized. It seems good to request a review after code organization.

This PR was in draft status, I'll change the status to open when these code is organized.

external/model-preparation-algorithm/mpa_tasks/extensions/datasets/mpa_cls_dataset.py
```diff
@@ -18,10 +18,16 @@
 @DATASETS.register_module()
 class MPAClsDataset(BaseDataset):
 
-    def __init__(self, ote_dataset=None, labels=None, **kwargs):
+    def __init__(self, ote_dataset=None, labels=None, empty_label=None, multilabel=False,
+                 hierarchical=False, hierarchical_info=None, **kwargs):
```
harimkang: I think the parameter of _init_ seems complicated, what do you think?
Multilabel, hierarchical, and multiclass can be combined into one, and the method of subtracting from kwargs looks better for hierarchical_info. 
Also, it seems good to get a separate parameter for the __init__ of the MPAMultilabelClsDataset.

external/model-preparation-algorithm/mpa_tasks/extensions/datasets/mpa_cls_dataset.py
```diff
@@ -49,10 +55,49 @@ def get_indices(self, *args):
...
-        for dataset_item in self.ote_dataset:
-            roi_label = dataset_item.get_roi_labels(self.labels)
-            label = [self.label_idx[lbl.id] for lbl in roi_label] if roi_label else [None]
-            self.gt_labels.append(label)
+        include_empty = self.empty_label in self.labels
+        for i, _ in enumerate(self.ote_dataset):
+            class_indices = []
+            item_labels = self.ote_dataset[i].get_roi_labels(self.labels, include_empty=include_empty)
+            ignored_labels = self.ote_dataset[i].ignored_labels
+            if item_labels:
```
harimkang, line 60: Can't we separate the codes inside as a function?

external/model-preparation-algorithm/mpa_tasks/apis/classification/task.py
```diff
@@ -104,19 +93,51 @@ def infer(self,
...
                     if pred_item > pos_thr:
                         cls_label = ScoredLabel(self.labels[cls_idx], probability=float(pred_item))
-                        label.append(cls_label)
+                        item_labels.append(cls_label)
+                    else:
+                        logger.info('item_labels is empty.')
+                        item_labels.append(ScoredLabel(self._empty_label, probability=1.))
+
+            elif self._hierarchical:
+                for i in range(self._hierarchical_info['num_multiclass_heads']):
```
harimkang: Can't we take out the codes in this part as a function? (If possible?)

harimkang (changes requested):
I left a few comments.

harimkang:
When you're ready, it seems good to turn the CI test around and look at the report. Of course, there may be few related tests, but it would be nice to check the multi-class & multi-label classification.

external/model-preparation-algorithm/mpa_tasks/extensions/datasets/mpa_cls_dataset.py
JihwanEom: Yes, I agree with your opinion that init seems too complicated. But we if we separate init method, we should handle in task.py.
I think this can be same result compared with current status, just for where we locate complicated statements.
How do you think about that?

harimkang: I still think that even if the task handles it, MPAClsDataset's init should not accept the variables (multilabel, hierarchical, and hierarchical_info...etc) and receive them separately. in the case of multi-class, you can modify it a little so that the variables don't come in the __init__ place.

external/model-preparation-algorithm/mpa_tasks/apis/classification/task.py
JihwanEom: Surely, I prefer to make a function to parse hierarchical information. I'll think about this carefully.

external/model-preparation-algorithm/mpa_tasks/extensions/datasets/mpa_cls_dataset.py
JihwanEom: Okay, I'll separate this function on each dataset.

external/model-preparation-algorithm/mpa_tasks/extensions/datasets/mpa_cls_dataset.py
JihwanEom: Same with below.

external/model-preparation-algorithm/mpa_tasks/apis/classification/task.py
JihwanEom: Hi, hierarchical information is indexing on infer and mpa_cls_dataset.py. But the actual work is a little different. Of course, there will be nice ways to parse the information, but the way should cover HierarchicalClsHead in MPA. So, I'll leave the code as-is and I'll refactor later.

external/deep-object-reid/torchreid_tasks/utils.py
```diff
@@ -54,6 +54,7 @@
 
 from torchreid.utils import set_model_attr, get_model_attr
 
+
```
harimkang, line 57: Should we use this class even if we use MPATemplate?

external/model-preparation-algorithm/mpa_tasks/apis/classification/task.py
harimkang: Is there any problem that 'i' overlaps with the index of dataset_item? 
Is there a reason why you have to use it as an "i"?

And it's still hard to understand why it's hard to separate it into functions.

external/model-preparation-algorithm/mpa_tasks/apis/classification/task.py
```diff
@@ -109,19 +98,52 @@ def infer(self,
...
+                item_labels = self._task_environment.label_schema.resolve_labels_probabilistic(item_labels)
+                if not item_labels:
+                    logger.info('item_labels is empty.')
+                    item_labels.append(ScoredLabel(self._empty_label, probability=1.))
             else:
-                label_idx = prediction_item.argmax()
-                cls_label = ScoredLabel(self._labels[label_idx], probability=float(prediction_item[label_idx]))
-                label.append(cls_label)
-            dataset_item.append_labels(label)
+                if any(np.isnan(prediction_item)):
```
harimkang: Isn't this supposed to be an 'elif' in this code-flow?

external/model-preparation-algorithm/mpa_tasks/extensions/datasets/mpa_cls_dataset.py
```diff
@@ -238,3 +262,124 @@ def evaluate(self,
...
+                class_indices = [0]*(self.hierarchical_info['num_multiclass_heads'] +
+                                     self.hierarchical_info['num_multilabel_classes'])
+                for j in range(num_cls_heads):
+                    class_indices[j] = -1
+                for ote_lbl in item_labels:
+                    group_idx, in_group_idx = self.hierarchical_info['class_to_group_idx'][ote_lbl.name]
+                    if group_idx < num_cls_heads:
+                        class_indices[group_idx] = in_group_idx
+                    else:
+                        if ote_lbl not in ignored_labels:
```
harimkang: This part also seems to need to be modified to 'elif'.

harimkang (changes requested):
I left a few comments.
Additionally,
1. When I looked at the added sample data, I could look at the data sets that already exist, such as 'data/car_tree_bug'. Is there a reason to add an image directly? I think you just need to specify the path in the annotation file, what do you think? (The capacity may increase due to overlapping image data.)
2. There seems to be a conflict in the submodule of the MPA. Please check.

external/deep-object-reid/torchreid_tasks/utils.py
JihwanEom: Yes, we need utils.py for MPA classification, but also we can move these classes into MPA space. I'll remove torchreid dependency in future PR.

external/model-preparation-algorithm/mpa_tasks/apis/classification/task.py
JihwanEom: Okay, I'll clear index overlapping issues. The hierarchical information indexing looks like "Similar" with the thing in MPAClsDataset, but it actually works differently.

external/model-preparation-algorithm/mpa_tasks/apis/classification/task.py
JihwanEom: Revised. Thanks.

external/model-preparation-algorithm/mpa_tasks/extensions/datasets/mpa_cls_dataset.py
JihwanEom: Revised. Thanks.

external/model-preparation-algorithm/mpa_tasks/apis/classification/task.py
harimkang: I didn't mean to combine the same function as a function, but I wanted to have a separate referral function for multi-label or h-label.

harimkang (dismissed):
Some code looks complicated. It would be good to divide it into functions, but it would be good to consider refactoring with SAMClassifier later PR.

jaegukhyun (dismissed):
LGTM

harimkang (approved):

jaegukhyun (approved):
LGTM

goodsong81 (approved):
LGTM.
But please note that we need refactoring in future PRs.
There are too many / long if-else blocks that should be extracted as method or subclassed.

PR #1162, opened 2022-07-13
[Detection] Fix OpenVINO backward incompatibility

eugene123tw:
Feature map and saliency map auxiliary output were added to OpenVINO model trace, however, the older OpenVINO model doesn't have the trace and causes the error.


eugene123tw:
run ote-test

goodsong81 (approved):
LGTM!

harimkang (approved):
LGTM

PR #1170, opened 2022-07-15
[MPA Detection] Change model optimization hyperparameters and final validation

jaegukhyun:
This PR includes

- Change YOLOX model's optimization hyperparameters
- Anchors update when the model, whose anchors are re-clustered, is evaluated
- Change final validation checkpoint to the best epoch's weights


goodsong81 (changes requested):
All the newly added code blocks could be refactored into _xxxx_anchors() methods. Anchor clustering is only for current SSD and I think it should be removed later.
In addition, anchors -> _anchors would be better unless it is intended as public variable.

jaegukhyun:
@goodsong81, thank you for your review, I changed variable name, and I made _update_anchors function to refactor added code block

harimkang (approved):
LGTM

goodsong81 (approved):
LGTM.

PR #1174, opened 2022-07-19
[OTE] Comment out time.sleep(2) in EpochRunnerWithCancel

JihwanEom:
### Summary
- Comment out time.sleep(2) in EpochRunnerWithCancel for all tasks. (Deadlock is not worried for single-gpu)

### DIFF
--- a/QUICK_START_GUIDE.md
+++ b/QUICK_START_GUIDE.md
@@ -3,20 +3,29 @@
 ## Prerequisites
 * Ubuntu 18.04 / 20.04
 * Python 3.8+
-* [CUDA Toolkit 11.1](https://developer.nvidia.com/cuda-11.1.1-download-archive) - for training on GPU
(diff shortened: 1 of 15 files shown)


goodsong81 (changes requested):
Could you add comments?
ex) # TODO: uncomment this line or resolve root cause of deadlock issue if multi-GPUs need to be supported.

JihwanEom:
> Could you add comments? ex) # TODO: uncomment this line or resolve root cause of deadlock issue if multi-GPUs need to be supported.

Revised. Thank you.

harimkang (approved):
LGTM,
I think there is no problem because the multi-gpu environment is not currently used, but I would like to proceed with the SC test. I would like you to check and comment on this before it is merged with SC.

goodsong81 (approved):
LGTM

PR #1177, opened 2022-07-21
[Deployment] Improve post-process speed by prefiltering score

eugene123tw:
Total inference time is reduced from 62 sec to 37 sec for 200 images.


jaegukhyun (approved):
So the key is filter out before making output. Good job!

harimkang (approved):
LGTM

PR #1182, opened 2022-07-27
[CVS-88321] Classification training fails in detection -> classification projects

JihwanEom:
### Summary
- Resolve CVS-88321; Check whether ote_dataset label is in task labels. (for chain task)

### DIFF
--- a/external/model-preparation-algorithm/mpa_tasks/utils/data_utils.py
+++ b/external/model-preparation-algorithm/mpa_tasks/utils/data_utils.py
@@ -17,7 +17,8 @@ def get_cls_img_indices(labels, dataset):
     for i, item in enumerate(dataset):
         item_labels = item.annotation_scene.get_labels()
         for i_l in item_labels:
-            img_indices[i_l.name].append(i)
+            if i_l in labels:
+                img_indices[i_l.name].append(i)
 
     return img_indices
 


jaegukhyun (approved):

supersoob (approved):

external/model-preparation-algorithm/mpa_tasks/utils/data_utils.py
```diff
@@ -17,7 +17,8 @@ def get_cls_img_indices(labels, dataset):
     for i, item in enumerate(dataset):
         item_labels = item.annotation_scene.get_labels()
         for i_l in item_labels:
-            img_indices[i_l.name].append(i)
+            if i_l in labels:
+                img_indices[i_l.name].append(i)
```
goodsong81, line 21: Why don't we make img_indices as defaultdict(list)?
Could you explain why there are some labels in annotations are missing in 'labels'?

goodsong81 (approved):
Approving, but still defaultdict looks better to me.

external/model-preparation-algorithm/mpa_tasks/utils/data_utils.py
```diff
@@ -17,7 +17,8 @@ def get_cls_img_indices(labels, dataset):
     for i, item in enumerate(dataset):
         item_labels = item.annotation_scene.get_labels()
         for i_l in item_labels:
-            img_indices[i_l.name].append(i)
+            if i_l in labels:
```
harimkang, line 20: I'm actually still wondering if this is the right way to go. is the structure of label_schema a problem? is this the problem to be addressed here?

JihwanEom: Label schema is incoming properly, but previous code does not check the domain of dataset_item. So, if we train detection -> classification task, detection label is coming as item_labels on above code, this caused KeyError. 
To resolve this, I adopted simple if statement to check the domain of item_labels based on label schema (because it comes properly)

external/model-preparation-algorithm/mpa_tasks/utils/data_utils.py
supersoob: This error occurred when det -> cls task transition. `get_labels()` brings all labels the item contains no matter what the domain is. The label for the previous det task is brought to item_labels. But the keys(classes) in img_indices has only classification domain(It means label schema is working well). To check if item_label is cls or else, that `if i_l in labels` is needed. If we use defaultdict(list) and let img_indices make keys from item_labels without checking the domain, we cannot extract the labels of the classification domain.

PR #1184, opened 2022-07-27
fix bug about override_configs in MPA task

eunwoosh:
This PR fixes bug about override_configs in MPA task.
if custom_hooks is in override_configs, then custom hooks is not appended to existing custom_hooks but override it.
This PR modifies code to make custom_hooks to append to existing custom hooks.


yunchu (approved):
LGTM

eunwoosh:
run ote-test

harimkang (approved):

PR #1193, opened 2022-08-05
Add common setup.py

ashwinvaidya17:
# Description

- Sample setup.py which installs both ote_sdk and ote_cli from the root folder.
- The issue with this is that pylint will fail if you do an editable install. This is because setup installs in `ote_sdk, ote_cli` whereas editable points to `ote_sdk/ote_sdk, ote_cli/ote_cli`. This can serve as a temporary fix till everything is moved to the `ote` folder.
- I haven't updated all the shell scripts

image


ashwinvaidya17:
Closing as this will be addressed in a different PR

PR #1194, opened 2022-08-05
[OTE Enhancements] - Add `tox` to the repo.

samet-akcay:
# Description
- Remove rc files argumetns from pre-commit configuration.
- Add `pyproject.toml` file
- Create `requirements/dev.txt` file to be able to install dev packages.
- Add tox.ini file to the repo

> **NOTE** tox is currently limited such that we need to install ote-sdk and other packages into the venv. This will be addressed once we combine packages into a single ote package.


pyproject.toml
```diff
@@ -0,0 +1,97 @@
...
+    "deprecated-pragma",
+    "use-symbolic-message-instead",
+    "import-outside-toplevel",
+    "arguments-differ",
+]
+
+[tool.pylint.typecheck]
+generated-members = ["numpy.*", "torch.*"]
+ignored-modules = [
+    "ote_sdk",
```
ashwinvaidya17, line 50: I would be in favour of treating ote_sdk as first party and fixing the isort issues first as we might forget about these settings in the future. Plus it would make it easier to move cli and sdk into a single package.

ashwinvaidya17 (commented):
I feel we should have setup.py first as currently tox requires a base env with everything installed. But that might be a lot more effort.

pyproject.toml
samet-akcay: > I feel we should have setup.py first as currently tox requires a base env with everything installed. But that might be a lot more effort.

This draft PR adds https://github.com/openvinotoolkit/training_extensions/pull/1199 `setup.py`

samet-akcay: > I would be in favour of treating ote_sdk as first party and fixing the isort issues first as we might forget about these settings in the future. Plus it would make it easier to move cli and sdk into a single package.

I also thought about that, but gave moving `ote.cli` a try, which seemed to sort imports ok so far.

ashwinvaidya17 (approved):

djdameln (approved):
Thanks

PR #1198, opened 2022-08-08
Enable mixed precision for all tasks and model templates

supersoob:
This PR includes
- Set mixed precision mode as default
- Tuned hyperparameter for fp16 training
- Fix export errors by setting exported model to fp32 precision mode


supersoob:
run ote-test

supersoob:
run ote-test

supersoob:
run ote-test

supersoob:
run ote-test

supersoob:
I'm testing this PR by mpa/tests because there's some problems in ote ci. Until that, please review this PR. I hope you review the [PR in  mmdetection](https://github.com/openvinotoolkit/mmdetection/pull/355) along with this PR. It should be changed to enable FP16 in detection task. 
And I put fp16 cfg in config/model.py to make it easily disable or enable. Or It maybe put in somewhere in mpa submodule. I don't know what's the right choice. If you have other idea, please share it. Thank you.

external/model-preparation-algorithm/configs/instance-segmentation/resnet50_maskrcnn/template.yaml
```diff
@@ -30,7 +30,7 @@ hyper_parameters:
         default_value: 4
         auto_hpo_state: POSSIBLE
       learning_rate:
-        default_value: 0.001
+        default_value: 0.002
```
harimkang: Is this the best param setting in the experiment?

supersoob:
run ote-test

alexsu52 (commented):
Hello @supersoob, 

Could you provide evidence that these changes will not break the optimization flow. I mean optimization flow is based on training pipeline and depends on training precision.

external/model-preparation-algorithm/configs/segmentation/ocr-lite-hrnet-18-mod2/model.py
```diff
@@ -3,3 +3,11 @@
 ]
 
 load_from = 'https://storage.openvinotoolkit.org/repositories/openvino_training_extensions/models/custom_semantic_segmentation/litehrnet18_imagenet1k_rsc.pth'
+optimizer_config = dict(
+    type='Fp16OptimizerHook',
```
JihwanEom: Could you explain why only segmentation task requires whole optimizer_config?

supersoob: mmsegmentation does not handle fp16 configuration. It just build optimizer from optimizer_config. Please check mmsegmentation submodule.

supersoob: Do I need to make it fp16 config and then handle it inside MPA submodule? What do you think? I checked this again and I think it's better to deal it inside the mpa code because it's a bit long and the grad_norm can be changed for each training scheme. Can you give me some idea?

supersoob:
run ote-test

supersoob:
run ote-test

external/model-preparation-algorithm/mpa_tasks/apis/task.py
```diff
@@ -180,7 +176,7 @@ def _initialize(self, dataset=None, output_model=None):
                 del self._model_cfg._cfg_dict['fp16']
             elif isinstance(self._model_cfg, ConfigDict):
                 del self._model_cfg['fp16']
-        self._precision = self._precision_from_config
+        self._precision = self._model_cfg.get('precision', [ModelPrecision.FP32])
```
harimkang: Does model_cfg always have a precision? It seems that the key 'fp16' is used in the setting (model.py) above, then where do this put the precision to model_cfg?

supersoob: Both in formal code and current code, _precision only used for what type model has to be exported, orthogonal with training mode(which will be enabled by fp16) Since we adopt to export model to fp32, I made to get precision cfg from cfg file and set default mode to fp32.

harimkang: Yes, I understand, but I wonder which code allows the key 'precision' to exist in model_cfg.

supersoob: Could you give me some suggestion? I don't think hard-code like [ModelPrecision.FP32] is not good idea, so want to let it get from cfg.

goodsong81: This is task implementation in OTE repo... not MPA recipe. So we don't need to put the export options to config. We may make it as hyper-parameter with FP32 as default. Anyway, if you want to put it in config, it should not be in the form of "enum ModelPrecision" type; maybe string, boolean or number?

goodsong81 (changes requested):

supersoob:
run ote-test

external/model-preparation-algorithm/mpa_tasks/apis/task.py
supersoob: I fixed it to get FP32 export option.

goodsong81 (dismissed):

JihwanEom (dismissed):
LGTM

chuneuny-emily (dismissed):

supersoob:
> Hello @supersoob,
> 
> Could you provide evidence that these changes will not break the optimization flow. I mean optimization flow is based on training pipeline and depends on training precision.



> Hello @supersoob,
> 
> Could you provide evidence that these changes will not break the optimization flow. I mean optimization flow is based on training pipeline and depends on training precision.

Hi, @alexsu52.
I ran this branch in local test, because NNCF test cannot be run in OTE CI because there were issues in running mmseg test causing the rest of CI tests to be stopped.
I checked that there is no issue in NNCF optimization for MPA templates. These are test result files. Please take a look at them.

[local_test_log.zip](https://github.com/openvinotoolkit/training_extensions/files/9313992/local_test_log.zip)

alexsu52:
> > Hello @supersoob,
> > Could you provide evidence that these changes will not break the optimization flow. I mean optimization flow is based on training pipeline and depends on training precision.
> 
> > Hello @supersoob,
> > Could you provide evidence that these changes will not break the optimization flow. I mean optimization flow is based on training pipeline and depends on training precision.
> 
> Hi, @alexsu52. I ran this branch in local test, because NNCF test cannot be run in OTE CI because there were issues in running mmseg test causing the rest of CI tests to be stopped. I checked that there is no issue in NNCF optimization for MPA templates. These are test result files. Please take a look at them.
> 
> [local_test_log.zip](https://github.com/openvinotoolkit/training_extensions/files/9313992/local_test_log.zip)

Thanks for the provided logs!

alexsu52 (dismissed):
LGTM

JihwanEom (approved):
LGTM

goodsong81 (approved):
LGTM!

PR #1205, opened 2022-08-10
[HPO] fix bug occurred when running HPO without batch size in search space

eunwoosh:
This PR fixes a bug which is occurred when HPO runs without batch size in search space.

### DIFF
--- a/ote_cli/ote_cli/utils/hpo.py
+++ b/ote_cli/ote_cli/utils/hpo.py
@@ -232,7 +232,9 @@ def run_hpo_trainer(
         else:
             hyper_parameters.learning_parameters.num_iters = hp_config["iterations"]
 
-    # add batch size value if batch size isn't in search space
+    # current HPO requires batch_size information.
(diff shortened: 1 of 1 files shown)


eunwoosh:
run ote-test

ote_cli/ote_cli/utils/hpo.py
```diff
@@ -232,6 +232,13 @@ def run_hpo_trainer(
         else:
             hyper_parameters.learning_parameters.num_iters = hp_config["iterations"]
 
+    # add batch size value if batch size isn't in search space
```
yunchu: this comment should be changed like
"current HPO requires batch_size information so if batch_size is not in the HPO's target hparams, need to set batch_size config from the instance of ConfigurableParameters"

eunwoosh: I see. I modified as you said.

goodsong81 (approved):

supersoob (approved):
LGTM

yunchu (approved):

PR #1209, opened 2022-08-16
[ANOMALY] Add PRO Metric

djdameln:
This PR adds the performance provider for the Per-Region Overlap (PRO) score.

The metric is not yet used by any task, but we are planning to use it as local performance metric for the anomaly segmentation and detection task. The metric currently does not report any dashboard metrics.


ashwinvaidya17 (approved):

harimkang (changes requested):
I think the unit test code on the ote_sdk side for the metric is missing.

PR #1212, opened 2022-08-18
Enhance training schedule for multi-label classification

JihwanEom:
### Summary:
- Adopt early-stop for multi-label classification
- Update multi-label classification model templates

### DIFF
--- a/external/model-preparation-algorithm/configs/classification/efficientnet_b0_cls_incr/model_multilabel.py
+++ b/external/model-preparation-algorithm/configs/classification/efficientnet_b0_cls_incr/model_multilabel.py
@@ -2,7 +2,6 @@ _base_ = [
   '../../../submodule/models/classification/ote_efficientnet_b0_multilabel.yaml',
 ]
 
-runner = dict(max_epochs=90)
 fp16 = dict(loss_scale=512.)
 
 custom_hooks = [
--- a/external/model-preparation-algorithm/configs/classification/efficientnet_b0_cls_incr/model_multilabel.py
+++ b/external/model-preparation-algorithm/configs/classification/efficientnet_b0_cls_incr/model_multilabel.py
@@ -2,7 +2,6 @@ _base_ = [
   '../../../submodule/models/classification/ote_efficientnet_b0_multilabel.yaml',
 ]
 
-optimizer = dict(lr=0.025)
 runner = dict(max_epochs=90)
 fp16 = dict(loss_scale=512.)
 
@@ -14,6 +13,6 @@ custom_hooks = [
         iteration_patience=0,
         metric='mAP',
         interval=1,
-        priority='ABOVE_NORMAL',
+        priority=75,
     ),
 ]
--- a/external/model-preparation-algorithm/configs/classification/efficientnet_b0_cls_incr/model_multilabel.py
+++ b/external/model-preparation-algorithm/configs/classification/efficientnet_b0_cls_incr/model_multilabel.py
@@ -1,4 +1,19 @@
 _base_ = [
   '../../../submodule/models/classification/ote_efficientnet_b0_multilabel.yaml',
 ]
+
+optimizer = dict(lr=0.025)
+runner = dict(max_epochs=90)
 fp16 = dict(loss_scale=512.)
+
+custom_hooks = [
+    dict(
+        type='LazyEarlyStoppingHook',
+        start=0,
+        patience=10,
+        iteration_patience=0,
+        metric='mAP',
+        interval=1,
+        priority='ABOVE_NORMAL',
+    ),
+]
--- a/external/model-preparation-algorithm/mpa_tasks/apis/classification/task.py
+++ b/external/model-preparation-algorithm/mpa_tasks/apis/classification/task.py
@@ -300,6 +300,16 @@ class ClassificationInferenceTask(BaseTask, IInferenceTask, IExportTask, IEvalua
         else:
             cfg_path = os.path.join(base_dir, 'model.py')
         cfg = MPAConfig.fromfile(cfg_path)
+
+        # To initialize different HP for multi-label classification / Support HP change via CLI & UI
+        if self._multilabel:
+            template = MPAConfig.fromfile(self.template_file_path)
+            template_params = template.hyper_parameters.parameter_overrides.learning_parameters
+            if template_params.learning_rate.default_value != self._hyperparams.learning_parameters.learning_rate:
+                cfg.pop('optimizer', False)
+            if template_params.num_iters.default_value != self._hyperparams.learning_parameters.num_iters:
+                cfg.pop('runner', False)
+
         cfg.model.multilabel = self._multilabel
         cfg.model.hierarchical = self._hierarchical
         if self._hierarchical:
--- a/external/model-preparation-algorithm/mpa_tasks/apis/classification/task.py
+++ b/external/model-preparation-algorithm/mpa_tasks/apis/classification/task.py
@@ -301,8 +301,8 @@ class ClassificationInferenceTask(BaseTask, IInferenceTask, IExportTask, IEvalua
             cfg_path = os.path.join(base_dir, 'model.py')
         cfg = MPAConfig.fromfile(cfg_path)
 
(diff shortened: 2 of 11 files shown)


JihwanEom:
run ote-test

JihwanEom:
CI is not working properly, tested CLI tests on local environment with below command:
• pytest -rs external/model-preparation-algorithm/tests/ote_cli/test_multilabel_cls_il.py
image
All test cases are passed. (also include multiclass, hierarchical classification tests.)

external/model-preparation-algorithm/configs/classification/efficientnet_b0_cls_incr/model_multilabel.py
```diff
@@ -1,4 +1,19 @@
 _base_ = [
   '../../../submodule/models/classification/ote_efficientnet_b0_multilabel.yaml',
 ]
+
+optimizer = dict(lr=0.025)
```
sovrasov: Note that changing the LR here normally shouldn't affect the training via OTE CLI or SC, because the value from a model template is suppose to be default in a ote_sdk.entities.train_parameters.TrainParameters instance (so 0.007 will be substituted).

Because each classification template is tied to 3 tasks (multiclass, multilabel, hierarchical), we can't set the optimal learning rate for each task. That's a restriction caused by the current SC design.

external/model-preparation-algorithm/mpa_tasks/apis/classification/task.py
```diff
@@ -300,6 +300,16 @@ def _init_model_cfg(self):
         else:
             cfg_path = os.path.join(base_dir, 'model.py')
         cfg = MPAConfig.fromfile(cfg_path)
+
+        # To initialize different HP for multi-label classification / Support HP change via CLI & UI
+        if self._multilabel:
+            template = MPAConfig.fromfile(self.template_file_path)
+            template_params = template.hyper_parameters.parameter_overrides.learning_parameters
+            if template_params.learning_rate.default_value != self._hyperparams.learning_parameters.learning_rate:
```
sovrasov: This hack won't update the value on UI unfortunately: user will see 0.007 by default instead of actual 0.025 for some models.

supersoob (dismissed):

external/model-preparation-algorithm/mpa_tasks/apis/classification/task.py
```diff
@@ -300,6 +300,14 @@ def _init_model_cfg(self):
         else:
             cfg_path = os.path.join(base_dir, 'model.py')
         cfg = MPAConfig.fromfile(cfg_path)
+
+        # To initialize different HP for multi-label classification / Support HP change via CLI & UI
+        if self._multilabel:
+            template = MPAConfig.fromfile(self.template_file_path)
+            template_params = template.hyper_parameters.parameter_overrides.learning_parameters
+            if template_params.num_iters.default_value != self._hyperparams.learning_parameters.num_iters:
+                cfg.pop('runner', False)
```
goodsong81: What does this mean?

supersoob (commented):
Do you have any results of updated CI test or local test?

external/model-preparation-algorithm/mpa_tasks/apis/classification/task.py
```diff
@@ -300,6 +300,15 @@ def _init_model_cfg(self):
...
             cfg_path = os.path.join(base_dir, 'model.py')
         cfg = MPAConfig.fromfile(cfg_path)
+
+        # To initialize different HP for multi-label classification / Support HP change via CLI & UI
+        if self._multilabel:
+            template = MPAConfig.fromfile(self.template_file_path)
+            template_params = template.hyper_parameters.parameter_overrides.learning_parameters
+            incoming_params = self._hyperparams.learning_parameters
+            if cfg.get('runner', False) and (template_params.num_iters.default_value != incoming_params.num_iters):
+                cfg.runner.max_epochs = incoming_params.num_iters
```
goodsong81, line 310: Do we update # epochs only?

JihwanEom: Yes, another configs (lr, warmup_iters) are discarded by performance drop on class-incremental learning.

harimkang: Actually, it's hard for me to understand this code. If it is different from the default value, change max_epochs, and in the case of the default value, do you mean using the epoch of the recipe? 

In this case, isn't it possible to actually train with a different value from the value shown in the template?

JihwanEom: No, 
If incoming parameter (with CLI, that might be `params --learning_parameters.num_iters`) is **same** with default value of template, replace max epoch from 20 -> 90
But if incoming parameter is **different** with default value of template, **max_epoch should be changed into incoming parameters**. So, this code does not change max epoch to 90 to support control num_iters via OTE-CLI or SC UI.

JihwanEom:
CI is not working properly, tested CLI tests on local environment with below command:
• pytest -rs external/model-preparation-algorithm/tests/ote_cli/test_multilabel_cls_il.py
image

external/model-preparation-algorithm/mpa_tasks/apis/classification/task.py
harimkang: I think the epochs will be 90 in training, but the value displayed in the SC UI will be 20 so wouldn't this be a problem?

supersoob: Seems like no other choice currently. Since we did not separate the template of multi-class and multi-label(which requires some huge code change), we should have shared the parameters even if it has a best results at the different configuration. According to @JihwanEom , that features(early-stop and more epochs) influence hugely in performance for multi-label, while they do not in multi-class (I also tried to match to those parameters what he founds, but ended up needing additional hp search for early-stop/epochs that fits for multi-class classification. That was why I shared the early-stop exp. suddenly today.)

supersoob (dismissed):

external/model-preparation-algorithm/mpa_tasks/apis/classification/task.py
harimkang: This would be considered a bug in the eyes of the user.  Also, I actually felt that early-stop look good for me at the experimental results.  Of course, it would be nice to proceed when there are more experimental results, but this code feels like a bug to me.
Is there any other way not to look like a bug?

@goodsong81 What do you think?

JihwanEom: > I think the epochs will be 90 in training, but the value displayed in the SC UI will be 20 so wouldn't this be a problem?

Yeap sure, that's a limitation of this PR. But multi-label classification performance significantly drops without change of max_epoch up to 40%. As vlad said, I believe this is the limitation of SC design. I expect this issue will be resolved on next version of OTE/SC.

supersoob: I also think it look like a bug. But changing default setup to early-stop and larger epoch is quite not good choice for multi-class classification hastily without additional exp. In one template with cifar10, early stop was helpful. But generally, I think it's not good idea to use it in order to earn 1% accuracy taking 2x/3x more epochs.(some are even better both in accuracy and time with no_early_stop setting) It needs to run more experiments definitely with different lr or other param, which takes long period to search and analyze the performance of it.

jaegukhyun (dismissed):
LGTM

external/model-preparation-algorithm/mpa_tasks/apis/classification/task.py
harimkang: I agree with soobee.  But I don't think it's appropriate to leave it as it is, and some other way to make it not look like a bug should be added in this PR..

harimkang: Actually, I don't even have any ideas too.. Let's think about it.

goodsong81 (approved):
LGTM

harimkang (approved):

PR #1219, opened 2022-08-23
[ANOMALY] Fix local anomaly segmentation performance bug

djdameln:
This PR fixes a bug in the anomaly segmentation performance computation. The full-image bounding boxes for the normal images of the ground truth dataset were included in the Dice score computation, but not the normal predictions of the prediction dataset. This resulted in underestimation of the performance. 

Fixed by excluding the normal annotations from the computation.


djdameln:
@atwinand

ashwinvaidya17 (approved):
Thanks for discovering this

samet-akcay (approved):
Nice catch!

goodsong81 (approved):
LGTM!

PR #1225, opened 2022-08-25
[CVS-88098] Temporary workaround

goodsong81:
* Not raising exceptions for wrong hyperparam format
* Set defaults to Class-IL recipe


harimkang:
It was solved together in https://github.com/openvinotoolkit/training_extensions/pull/1226

PR #1228, opened 2022-08-25
[CVS-89912] Fixed config path computation

nikita-savelyevv:
**Description** 
Fixed how a path to NNCF config file is computed so that optimization of anomaly classification task would start.

Ticket: https://jira.devtools.intel.com/browse/CVS-89912
BMM: https://10.211.120.60/

PR #1230, opened 2022-08-26
[Anomaly Task] Revert hpo template

ashwinvaidya17:
# Description

- Reverts anomaly configuration location for anomaly task.

# Linting

image

# Tests

BMM - https://10.211.120.64/projects/630df620d6855fe642a8c33e/models/630df65b43a94424795576e0/630dfc0243a9442479557e13

Manual testing
- HPO works
- NNCF POT and Training time optimization work

image

OTX - http://validationreports.sclab.intel.com:8006/reports/build_number_report?test_session_build_number=pr-ote-test-552&environment=iotg-dev-workstation-10

Only these tests are failing
image

# Merge along with

https://github.com/intel-innersource/frameworks.ai.interactive-ai-workflow.sonoma-creek/pull/2720


ashwinvaidya17:
run ote-test

goodsong81:
Do you revert the config paths due to https://jira.devtools.intel.com/browse/CVS-89912?
It will be resolved by this PR: https://github.com/openvinotoolkit/training_extensions/pull/1228

ashwinvaidya17:
> Do you revert the config paths due to https://jira.devtools.intel.com/browse/CVS-89912? It will be resolved by this PR: #1228

Yeah. That's great! Fewer lines of code 🙂. I'll close this PR

samet-akcay:
Wait, I think [1228](https://github.com/openvinotoolkit/training_extensions/pull/1228) addresses the NNCF issue. This PR is for the HPO stuff right? @ashwinvaidya17

samet-akcay:
Am I missing anything here?

goodsong81:
Wanted to check if this is a duplicate. We concluded that this PR is needed. In addition, we may or may not revert the change in #1228 as we are changing the path itself.

nikita-savelyevv:
I've also found https://jira.devtools.intel.com/browse/CVS-90401 . This seems like a relevant PR to me. @ashwinvaidya17, is it by any chance fixed here?

Edit: after looking into it in more detail, I see that the issue above comes up because pruning is not supported for anomaly classification task, but due to wrong configuration file it is possible to run it nevertheless. As this PR merely changes the location of config files, it's not that relevant. Will create a separate one.

ashwinvaidya17:
@nikita-savelyevv https://jira.devtools.intel.com/browse/CVS-90401 this hasn't been addressed here but I can have a look as I am looking into why the nncf optimization is failing.

nikita-savelyevv:
@ashwinvaidya17, thanks, if you find anything that will be great. 

I suspect the issue is somewhere [here](https://github.com/openvinotoolkit/training_extensions/blob/develop/external/anomaly/configs/stfpm/configuration.yaml#L247)

ashwinvaidya17:
@nikita-savelyevv pruning is now disabled in the UI
image

nikita-savelyevv:
> @nikita-savelyevv pruning is now disabled in the UI image

Good. Have you tried uploading more than 1000 data samples? This is usually the requirement for the pruning to get enabled.

ashwinvaidya17:
@nikita-savelyevv It shows up with 1000> images 😕
image

nikita-savelyevv:
@ashwinvaidya17 Hmm, maybe also set `visible_in_ui: false` for `pruning_supported`. 

Edit: On the second thought, this probably won't help. This parameter however should be set to false anyway.

nikita-savelyevv:
All in all, it was decided to leave https://jira.devtools.intel.com/browse/CVS-90401 for further investigation and possibly another PR, if the issue is on OTE side.

external/anomaly/configs/draem/configuration.py
```diff
@@ -1,6 +1,4 @@
-"""
```
djdameln, line 1: Maybe we could move these files to `external/configs/base/draem`. That way the `configs` directory would be organized in a more consistent task -> algo structure.

external/anomaly/tasks/inference.py
```diff
@@ -283,8 +283,11 @@ def save_model(self, output_model: ModelEntity) -> None:
         output_model.set_data("label_schema.json", label_schema_to_bytes(self.task_environment.label_schema))
         self._set_metadata(output_model)
 
-        f1_score = self.model.image_metrics.F1Score.compute().item()
-        output_model.performance = Performance(score=ScoreMetric(name="F1 Score", value=f1_score))
+        if hasattr(self.model, "image_metrics"):
+            f1_score = self.model.image_metrics.F1Score.compute().item()
+            output_model.performance = Performance(score=ScoreMetric(name="F1 Score", value=f1_score))
```
djdameln, line 288: I'm not sure exactly what this `output_model.performance` is used for, but shouldn't we report pixel-level performance when the task type is segmentation/detection?

djdameln (commented):
In `external/anomaly/configs/classification/draem` folder we now have `compression_config.json` and `compression_config copy.json`, with different contents. I guess the copy is there by mistake?

external/anomaly/configs/draem/configuration.py
ashwinvaidya17: I tried that earlier but since all the model configurations depend on `base/configuration.py` I got error due to cyclic dependencies. Ideally I feel configuration.py and templates should be in a separate directory.

djdameln: This structure works fine when I try it:
Screenshot from 2022-08-31 14-08-03

ashwinvaidya17: What do you have in `__init__.py`?

djdameln: I have the following in `configs/base/__init__.py`:
```
from .padim import PadimAnomalyBaseConfig
from .stfpm import STFPMAnomalyBaseConfig
from .draem import DraemAnomalyBaseConfig

__all__ = ["PadimAnomalyBaseConfig", "STFPMAnomalyBaseConfig", "DraemAnomalyBaseConfig"]
```
We don't need to import `BaseAnomalyConfig` in the `__init__.py`, because it's not used outside the `configs.base` module.

ashwinvaidya17: Good point

djdameln (approved):
Thanks!

ashwinvaidya17:
run ote-test

samet-akcay (approved):
thanks for reverting this, and sorry for the continous change :(

goodsong81 (approved):
What a huge PR, but LGTM!

ashwinvaidya17:
Latest build - https://10.211.120.64/projects/631057083ab7bc8ba784abd3/models
image

PR #1238, opened 2022-08-31
[OTX Refactor] 📄 Documentation

ashwinvaidya17:
# Description

- Initial docstrings refactor.


otx/api/configuration/configurable_parameters.py
```diff
@@ -1,6 +1,4 @@
-"""
-This module contains the base class to define ConfigurableParameters within OTX
-"""
+"""This module contains the base class to define ConfigurableParameters within OTX."""
```
samet-akcay, line 1: Can't really tell. Is there a line break between the docstring and the license header

samet-akcay (approved):
Thanks, much better!

PR #1215, opened 2022-08-19
Support hpo to multilabel and hierarchy classification

eunwoosh:
This PR includes supporting HPO to multilabel and hierarchy classification.

PR #1241, opened 2022-09-01
Expose early stopping hyper-parameters for all tasks

supersoob:
This PR includes

- Add early stopping parameters in config.py 
- Overwrite early stopping hook
- hyper-parameter change for multi-class classification to enable early stopping


supersoob:
run ote-test

external/model-preparation-algorithm/configs/classification/configuration.yaml
```diff
@@ -123,15 +123,32 @@ learning_parameters:
...
       operator: AND
       rules: []
       type: UI_RULES
-    visible_in_ui: false
+    visible_in_ui: true
     warning: null
+  patience:
+    affects_outcome_of: TRAINING
+    default_value: 10
+    description: Number of epochs without improvement before the system will automatically stop training.
```
JihwanEom: How about "Training will be stopped if the model is not improved during patience."

external/model-preparation-algorithm/configs/detection/configuration.yaml
```diff
@@ -119,6 +119,37 @@ learning_parameters:
     value: 0
     visible_in_ui: true
     warning: null
+  enable_early_stopping:
+    affects_outcome_of: TRAINING
+    default_value: true
+    description: Adaptive early exit from training when accuracy isn't changed or decreased for several epochs.
```
JihwanEom: I'm confused with "Adaptive".. Does that mean early stopping by adaptive interval?

external/model-preparation-algorithm/mpa_tasks/apis/classification/task.py
```diff
@@ -381,10 +375,13 @@ def _patch_evaluation(self, config: MPAConfig):
         cfg = config.evaluation
         if self._multilabel:
             cfg.metric = ['accuracy-mlc', 'mAP', 'CP', 'OP', 'CR', 'OR', 'CF1', 'OF1']
+            config.early_stop_metric = 'mAP'
         elif self._hierarchical:
             cfg.metric = ['MHAcc', 'avgClsAcc', 'mAP']
+            config.early_stop_metric = 'mAP'
```
JihwanEom: Could you change into `MHAcc` or `accuracy`? mAP can be always 0.0 if hierarchical project is consisted only with single-label classification groups.
For 'accuracy', please refer this PR : https://github.com/openvinotoolkit/training_extensions/pull/1215

external/model-preparation-algorithm/mpa_tasks/apis/detection/task.py
```diff
@@ -330,9 +332,7 @@ def _patch_evaluation(config: MPAConfig):
         cfg.metric = 'mAP'
         cfg.save_best = 'mAP'
         # EarlyStoppingHook
-        for cfg in config.get('custom_hooks', []):
-            if 'EarlyStoppingHook' in cfg.type:
-                cfg.metric = 'mAP'
+        config.early_stop_metric = 'mAP'
```
JihwanEom, line 343: @jaegukhyun , I can't remember exactly, does detection not use 'bbox_mAP'?

jaegukhyun: 'mAP' is right

external/model-preparation-algorithm/mpa_tasks/apis/task.py
```diff
@@ -155,10 +156,48 @@ def _initialize(self, dataset=None, output_model=None, export=False):
...
         logger.info('initializing....')
         self._init_recipe()
+
         if not export:
             recipe_hparams = self._init_recipe_hparam()
             if len(recipe_hparams) > 0:
                 self._recipe_cfg.merge_from_dict(recipe_hparams)
+
+            # Put early stop hook
+            if self._recipe_cfg.early_stop is True:
```
harimkang: I think it would be good to subtract the code related to early stop separately as a function.

external/model-preparation-algorithm/mpa_tasks/apis/task.py
```diff
@@ -155,10 +156,48 @@ def _initialize(self, dataset=None, output_model=None, export=False):
...
+
+                for hook in custom_hooks:
+                    if hook.type in ['EarlyStoppingHook','LazyEarlyStoppingHook']:
+                        early_stop_hook = hook
+                        break
+                
+                early_stop_hook['patience'] = patience
+                early_stop_hook['metric'] = metric
+
+                self.update_override_configurations({"custom_hooks" : [early_stop_hook]})
```
harimkang: we can use `update_or_add_custom_hook`

external/model-preparation-algorithm/mpa_tasks/apis/task.py
```diff
@@ -155,10 +156,48 @@ def _initialize(self, dataset=None, output_model=None, export=False):
...
+            else:
+                custom_hooks = self._recipe_cfg.get('custom_hooks',[])
```
harimkang: I think this can be solved simply by increasing the start value of early-stop to more than max_epoch. What do you think?

harimkang (changes requested):

external/model-preparation-algorithm/mpa_tasks/apis/task.py
supersoob: Thank you for suggestion, but it kind of looks like a hack. Completely removing the early stopping hook can also prevent from redundant part such as compare between current epoch and start epoch for every epoch in early stopping hook.

supersoob:
Hi, I refactored the codes reflecting your reviews. I changed the descriptions for early stopping hp and added new hp for adaptive interval. And I also wanted to make a function for early stopping update/removal, but it seems like there's no where to put related functions. Instead, I downsized the codes for early stopping hp. Please take a look again.

external/model-preparation-algorithm/mpa_tasks/apis/config.py
```diff
@@ -77,6 +77,44 @@ class BaseLearningParameters(ParameterGroup):
...
+
+        enable_early_stopping = configurable_boolean(
+            default_value=True,
+            header="Enable early stopping of the training",
+            description="Early exit from training when validation accuracy isn't \
+                         changed or decreased for several epochs.",
+            affects_outcome_of=ModelLifecycle.TRAINING
+        )
+        
+        patience = configurable_integer(
```
harimkang: I think this is too general a variable name. This is the patience of early stopping, but it seems to be used together elsewhere.

external/model-preparation-algorithm/mpa_tasks/apis/config.py
```diff
@@ -77,6 +77,44 @@ class BaseLearningParameters(ParameterGroup):
...
+            default_value=5,
+            min_value=0,
+            max_value=50,
+            header="Patience for early stopping",
+            description="Training will stop if the model does not improve within the number of epochs of patience.",
+            warning="This is applied exclusively when early stopping is enabled.",
+            affects_outcome_of=ModelLifecycle.TRAINING
+        )
+
+        iteration_patience = configurable_integer(
```
harimkang: This variable also looks too general.

external/model-preparation-algorithm/mpa_tasks/apis/detection/task.py
```diff
@@ -330,9 +339,7 @@ def _patch_evaluation(config: MPAConfig):
...
```
harimkang, line 343: To always be equal to the metric of evaluation, looks better to use early_stop_metric = cfg.metric. (The same goes for other tasks)

external/model-preparation-algorithm/mpa_tasks/apis/segmentation/task.py
```diff
@@ -149,9 +149,17 @@ def _init_recipe_hparam(self) -> dict:
         warmup_iters = int(self._hyperparams.learning_parameters.learning_rate_warmup_iters)
         lr_config = ConfigDict(warmup_iters=warmup_iters) if warmup_iters > 0 \
             else ConfigDict(warmup_iters=warmup_iters, warmup=None)
+        
+        if self._hyperparams.learning_parameters.enable_early_stopping is True:
```
harimkang: It looks good to remove `is True`.

external/model-preparation-algorithm/mpa_tasks/apis/task.py
```diff
@@ -181,6 +184,34 @@ def _initialize(self, dataset=None, output_model=None, export=False):
...
+            if self._recipe_cfg.use_adaptive_interval is True:
+                ada_interval_cfg = self._recipe_cfg.get('adaptive_validation_interval', None)
+                if ada_interval_cfg is None:
+                    self._recipe_cfg.adaptive_validation_interval = dict(max_interval=5)
+            else:
+                self._recipe_cfg.pop('adaptive_validation_interval')
+
+        # Add/remove early stop hook
+        if 'early_stop' in self._recipe_cfg:
+            remove_custom_hook(self._recipe_cfg, 'EarlyStoppingHook')
```
harimkang, line 195: Does this mean that this will not use the regular 'EarlyStoppingHook' but only the 'LazyEarlyStoppingHook'?

external/model-preparation-algorithm/mpa_tasks/apis/task.py
```diff
@@ -181,6 +184,34 @@ def _initialize(self, dataset=None, output_model=None, export=False):
...
+            early_stop = self._recipe_cfg.get('early_stop', False)
+            logger.info(f"early_stop : {early_stop}")
+            if early_stop is not False:
```
harimkang: `if early_stop` looks appropriate

external/model-preparation-algorithm/mpa_tasks/apis/task.py
```diff
@@ -181,6 +184,34 @@ def _initialize(self, dataset=None, output_model=None, export=False):
                 del self._model_cfg['fp16']
         self._precision = [ModelPrecision.FP32]
 
+        # Add/remove adaptive interval hook
+        if 'use_adaptive_interval' in self._recipe_cfg:
+            if self._recipe_cfg.use_adaptive_interval is True:
```
harimkang: I don't think we need this if statement.

external/model-preparation-algorithm/mpa_tasks/apis/task.py
```diff
@@ -181,6 +184,34 @@ def _initialize(self, dataset=None, output_model=None, export=False):
                 del self._model_cfg['fp16']
         self._precision = [ModelPrecision.FP32]
 
+        # Add/remove adaptive interval hook
+        if 'use_adaptive_interval' in self._recipe_cfg:
```
harimkang: I think `self._recipe_cfg.get('use_adaptive_interval', False)` is better

external/model-preparation-algorithm/mpa_tasks/apis/task.py
```diff
@@ -181,6 +184,34 @@ def _initialize(self, dataset=None, output_model=None, export=False):
...
+            if self._recipe_cfg.use_adaptive_interval is True:
+                ada_interval_cfg = self._recipe_cfg.get('adaptive_validation_interval', None)
+                if ada_interval_cfg is None:
+                    self._recipe_cfg.adaptive_validation_interval = dict(max_interval=5)
```
harimkang: Doesn't this cause an error if `adaptive_validation_interval `does not exist?
and If it's not an issue that occurs, it seems that `self._recipe_cfg.adaptive_validation_interval = self._recipe_cfg.get('adaptive_validation_interval', dict(max_interval=5))` is more appropriate.

harimkang (changes requested):

external/model-preparation-algorithm/mpa_tasks/apis/detection/task.py
supersoob: I put the early stopping check code in task.py but cls uses mutiple metrics, so early_stop_metric = cfg.metric will bring the lists of metrics in task.py in apis folder, uncertain what to choose for early stopping metric.

external/model-preparation-algorithm/mpa_tasks/apis/config.py
supersoob: Do you suggest to change the variable name? Like early_stop_patience/early_stop_iteration_patience?

harimkang: Yes, I think that is appropriate.

external/model-preparation-algorithm/mpa_tasks/apis/task.py
supersoob: Yes, you are right.

external/model-preparation-algorithm/mpa_tasks/apis/task.py
supersoob: Doesn't it just overwrite to its own value? Can you explain why it's more appropriate? And why do you think it causes error? I'm just curious.

external/model-preparation-algorithm/mpa_tasks/apis/task.py
supersoob: What statement are you talking? You reviewed the same part as below.

supersoob: I can't remove that if statement. Otherwise, it will cause error in the export stage. initialize function where init_recipe_hparam is initiated is called only when it's not export.

harimkang: What I mean is that if statement right above and this if statement can be combined into one, and I don't think it's going to cause an error because of this during export, so what do you think?

external/model-preparation-algorithm/mpa_tasks/apis/task.py
harimkang: I was just thinking about maybe we could reduce the code to three lines -> 1 line code. Also I thought there would be an issue if `self._recipe_cfg` didn't have `adaptive_validation_interval`.

external/model-preparation-algorithm/mpa_tasks/apis/task.py
supersoob: Your suggestion is really good, but it raises error in else for cls/seg tasks -> self._recipe_cfg.pop('adaptive_validation_interval')

use_adaptive_interval is the parameter only for detection.(because it did not work in other tasks.. must test at other time) So it needs to check first if use_adaptive_interval is actually in the recipe_cfg. and then need to pop and remove adaptive_validation_interval config only for det task when use_adaptive_interval is set to False.

harimkang: I think we can prevent errors by using `self._recipe_cfg.pop('adaptive_validation_interval', None)`.

supersoob: I modified this part. Please check it again!

external/model-preparation-algorithm/mpa_tasks/apis/task.py
supersoob: I modified this part. Please check it again!

external/model-preparation-algorithm/mpa_tasks/apis/segmentation/task.py
```diff
@@ -149,9 +149,17 @@ def _init_recipe_hparam(self) -> dict:
...
+        if self._hyperparams.learning_parameters.enable_early_stopping:
+            early_stop = ConfigDict(patience=int(self._hyperparams.learning_parameters.early_stop_patience),
+                                    iteration_patience=int(self._hyperparams.learning_parameters.early_stop_iteration_patience))
+        else:
+            early_stop = False
+
         return ConfigDict(
             optimizer=ConfigDict(lr=self._hyperparams.learning_parameters.learning_rate),
             lr_config=lr_config,
+            early_stop=early_stop,
```
harimkang, line 163: Of course, the adaptive_interval of the segmentation, the default is still set to False, but can't we enable this function?

external/model-preparation-algorithm/mpa_tasks/apis/task.py
```diff
@@ -181,6 +184,30 @@ def _initialize(self, dataset=None, output_model=None, export=False):
                 del self._model_cfg['fp16']
         self._precision = [ModelPrecision.FP32]
 
+        # Add/remove adaptive interval hook
+        if self._recipe_cfg.get('use_adaptive_interval', False):
+            self._recipe_cfg.adaptive_validation_interval = self._recipe_cfg.get('adaptive_validation_interval', dict(max_interval=5))
```
harimkang, line 189: `self._recipe_cfg` is `ConfigDict`. Therefore, `self._recipe_cfg.adaptive_validation_interval` will raise the `AttributeError` if the key `adaptive_validation_interval` does not exist.

supersoob: Sorry, I don't understand. As far as I know, ConfigDict can insert keys and values directly by using `self._recipe_cfg.adaptive_validation_interval`. I tested it works.

harimkang: I guess I got it wrong. I'm sorry.

harimkang (dismissed):
If the adaptive interval of detection works well in SC, I think it can be merged.

harimkang:
+Currently, the minimum value of patience is 0 (editable in UI), so I think we need to check if it works well.

supersoob:
Test results
image

supersoob:
I checked all tasks are working well with early stopping hp. Also working well when patience=0. I will update the hyperparameters resulted from early stop exp. for classification soon. Please review the related [MPA PR](https://github.com/openvinotoolkit/model_preparation_algorithm/pull/56), too

JihwanEom (commented):
Could you resolve below issue?
image

https://app.codacy.com/gh/openvinotoolkit/training_extensions/pullRequest?prid=10125167

supersoob:
> Could you resolve below issue? image
> 
> https://app.codacy.com/gh/openvinotoolkit/training_extensions/pullRequest?prid=10125167

I think it doesn't matter if it fails that check. There are longer lines in QUICK_START_GUIDE.md. Old PRs are merged even though it fails some codacy code analysis.

harimkang (commented):
There are no other comments on the implementation of the code. 
There are some modified model template values and hpo_config values, what kind of background does this have? (Maybe I missed some of the experimental results...)

(I missed it briefly, but hpo value is modified by Eun-Woo, Forget about hpo_config.)

jaegukhyun (approved):

JihwanEom (approved):

PR #1247, opened 2022-09-05
[Feature/OTX] Code Quality Checks

ashwinvaidya17:
# Description

- Add workflow for quality checks

# Note

Currently configured to run only for PRs to `feature/otx`


.github/workflows/pre_commit.yml
```diff
@@ -0,0 +1,22 @@
+name: Pre-Merge Checks
```
samet-akcay: Should this file be named as `pre_merge.yaml`? I guess we will add pre-merge checks here as well?

samet-akcay: Or do you want to plan these to be separate files/jobs?

ashwinvaidya17: No my bad. I think we can have it in the same workflow file

.github/workflows/pre_merge.yml
```diff
@@ -17,6 +17,6 @@ jobs:
         with:
           python-version: 3.8
       - name: Install dependencies
-        run: python -m pip install tox
+        run: python -m pip install tox pylint
```
samet-akcay: we could perhaps add `pylint` into `requirements/dev.txt`, and install `dev.txt`. This would ensure that we always install the same thing?

samet-akcay:
@goodsong81, FYI, this is the first trial of the CI, which works for the pre-commit checks! Great news!

samet-akcay (approved):
This is super cool, thanks!

PR #1249, opened 2022-09-06
🔨 Refactor some tests

ashwinvaidya17:
# Description

Removes create venv


otx/cli/utils/tests.py
```diff
@@ -99,13 +64,13 @@ def otx_train_testing(template, root, otx_dir, args):
...
     command_line.extend(args["train_params"])
-    assert run(command_line, env=collect_env_vars(work_dir)).returncode == 0
+    assert run(command_line).returncode == 0
     assert os.path.exists(f"{template_work_dir}/trained_{template.model_template_id}/weights.pth")
     assert os.path.exists(f"{template_work_dir}/trained_{template.model_template_id}/label_schema.json")
 
 
 def otx_hpo_testing(template, root, otx_dir, args):
-    work_dir, template_work_dir, _ = get_some_vars(template, root)
+    _, template_work_dir, _ = get_some_vars(template, root)
```
samet-akcay: Should we rename `get_some_vars` as `get_template_dir` or something?

samet-akcay: And perhaps only return `template_work_dir`?

ashwinvaidya17: Yeah makes sense!

tests/integration/cli/anomaly/test_anomaly_detection.py
```diff
@@ -15,13 +15,13 @@
 # and limitations under the License.
```
samet-akcay, line 15: Should we make these changes to classification and segmentation as well?

ashwinvaidya17: Oh right I forgot about those 😅 I was just testing detection

PR #1250, opened 2022-09-07
[Feature/OTX] 🔨 Fix openvino tests

ashwinvaidya17:
# Description

Adds requirements.txt, license and readme to otx package as this is needed to deploy model tests.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -241,6 +241,7 @@ setup(
     name="otx",
     version=get_otx_version(),
     packages=find_packages(exclude=("tests",)),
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -241,6 +241,7 @@ def get_requirements(requirement_files: Union[str, List[str]]) -> List[str]:
     name="otx",
     version=get_otx_version(),
     packages=find_packages(exclude=("tests",)),
+    package_data={"": ["requirements.txt", "README.md", "LICENSE"]},  # Needed for exportable code
```
samet-akcay, line 244: Nice catch!

PR #1252, opened 2022-09-07
[Feature/OTX] 📄 Add docs build stage to CI

ashwinvaidya17:


.github/workflows/docs.yml
```diff
@@ -20,10 +20,10 @@ jobs:
...
       - name: Install dependencies
         run: |
-          python -m pip install -r{toxinidir}/requirements/base.txt
-          python -m pip install -r{toxinidir}/requirements/anomaly.txt
-          python -m pip install -r{toxinidir}/requirements/openvino.txt
-          python -m pip install -r{toxinidir}/requirements/docs.txt
+          python -m pip install -r requirements/base.txt
+          python -m pip install -r requirements/anomaly.txt
+          python -m pip install -r requirements/openvino.txt
+          python -m pip install -r requirements/docs.txt
```
samet-akcay, line 23: Rather than doing this, would it be an idea to do `python -m pip install -e .[full]` to install all of these requirements? Of course, we need to ensure `[full]` installs all these packages.

ashwinvaidya17: Yeah that would be a good idea.

PR #1255, opened 2022-09-08
[Feature/OTX] Add Dockerfile + Containerize CI

ashwinvaidya17:


.dockerignore
```diff
@@ -0,0 +1,20 @@
+.tox
```
ashwinvaidya17, line 6: Need to clean this file a bit

PR #1263, opened 2022-09-14
[CVS-91472] Add pruning_supported value

harimkang:
### This PR includes :
- Add `pruning_supported` value in all tasks configurations
- Add `nncf_optimization` value per model template
- Align all model template with old configurations

### DIFF
--- a/external/model-preparation-algorithm/configs/classification/configuration.yaml
+++ b/external/model-preparation-algorithm/configs/classification/configuration.yaml
@@ -208,6 +208,21 @@ nncf_optimization:
     value: false
     visible_in_ui: true
     warning: null
+  pruning_supported:
+    affects_outcome_of: TRAINING
+    default_value: false
+    description: Whether filter pruning is supported
+    editable: false
+    header: Whether filter pruning is supported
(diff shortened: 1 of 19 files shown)


chuneuny-emily (approved):

AlbertvanHouten:
What is the use now fo the files in each individual template? can they be removed? e.g. https://github.com/openvinotoolkit/training_extensions/blob/bcf33fe0976ba062b68dc25db537bd71e798763c/external/deep-object-reid/configs/ote_custom_classification/mobilenet_v3_large_1/template_experimental.yaml

AlbertvanHouten (approved):

harimkang:
> What is the use now fo the files in each individual template? can they be removed? e.g. https://github.com/openvinotoolkit/training_extensions/blob/bcf33fe0976ba062b68dc25db537bd71e798763c/external/deep-object-reid/configs/ote_custom_classification/mobilenet_v3_large_1/template_experimental.yaml

The template location has been modified by adding the incremental training function. Models with experimental will be removed from the next version.

PR #1275, opened 2022-09-20
[Feature/OTX] Upgrade anomalib version in otx

ashwinvaidya17:
Matches commit https://github.com/openvinotoolkit/training_extensions/commit/0ef82adbd425ae1fde779587dbde00625a003612 in OTX


samet-akcay (approved):
Thanks!

samet-akcay (approved):
Thanks a lot for the effort!

PR #1272, opened 2022-09-16
Resolve pre-commit issues

harimkang:
### This PR includes :
- Update external/README.md algorithm tables (Resolve test_docs issues)
- Resolve flake8 issues & enable isort &black format for MPA
- Resolve some issues in ote_sdk & ote_cli
- Temporal disable mypy & pylint for MPA -> This will be added in the refactoring phase
- Resolve H-label cls issues in MPA (MPA commit update)
- Add TT_STABILITY_TESTS for stability test (Default model training test 100 times) [anomaly & MPA]

PR #1277, opened 2022-09-21
[OTE-TEST] Add TT_STABILITY_TEST for MPA test

harimkang:


harimkang:
will be checked with https://github.com/openvinotoolkit/training_extensions/pull/1272

PR #1282, opened 2022-09-23
Replace to IB loss for incremental learning in multi-class classification task

supersoob:
This PR includes

- Replace focal loss to IB loss for incremental learning in multi-class classification task [(Related MPA PR)](https://github.com/openvinotoolkit/model_preparation_algorithm/pull/64)
- Add get_num_data function in OTE Dataset in order to get num-per-class-data for IB loss weighting


supersoob:
IB loss requires num-per-cls-data for reweighting at the point of initialization. Since loss is defined in stage.py, I couldn't find the when/where to get num-per-cls-data but the class method in OTE dataset in ote_sdk/datasets.py. If have better idea, please suggest.

PR #1287, opened 2022-09-29
[Feature/OTX] Common & Detection Task Refactoring to otx.algorithms

harimkang:
## Summary
Refactor external backends to otx algorithms of Base (common) & Detection

### This PR includes :
- Add otx.algorithms.common for common base code of classification / detection / segmentation tasks
- Add ote.algorithms.detection for Detection task (mmdetection + model-preparation-algorithm)
- Detection NNCF & POT Task refactoring
- Add common & detection adapters
- (TODO) Installation for training -> Next PR
- (TODO) Test code Refactoring for detection -> Next PR

### Test command (Sample Code for detection training)
```
bash otx/algorithms/detection/init_venv.sh ./venv python3.8
source venv/bin/activate
python -m otx.algorithms.detection.tools.detection_sample otx/algorithms/detection/configs/detection/mobilenetv2_atss/template.yaml --export
```


otx/algorithms/detection/tasks/dataset/__init__.py
```diff
@@ -0,0 +1,9 @@
+"""OTX Algorithms - Detection Dataset."""
```
supersoob: I don't know if it's proper to put dataset-related codes in tasks folder. In anomaly, task folder is for such type as train/inference/nncf. some modules like model, datasets were in adapter folder. In my case, I made new subfolder called extension like `otx/algorithms/classification/extension`.

harimkang: I agree that it is not the right path. It can probably be moved to the adapters folder. (adapter is not ready now) 

and I think a new module called extension is not good idea. i think we can use adapters submodule.

supersoob: Okay. Since we are just in initial refactoring phase, let's take dataset in auxiliary path and move on to adapter folder.

harimkang: '+ Maybe the codes included in utils can be modified to belong to adapters. (If the adapters are implemented, I will consider it.)

tox.ini
```diff
@@ -23,7 +23,10 @@ deps =
     -r{toxinidir}/requirements/base.txt
     -r{toxinidir}/requirements/anomaly.txt
     -r{toxinidir}/requirements/openvino.txt
-commands = pre-commit run --all-files
+    -r{toxinidir}/requirements/common.txt
+commands = 
+    pip install --no-cache-dir --no-binary=mmpycocotools mmpycocotools
```
ashwinvaidya17: What's the advantage of `--no-cache-dir`? Wouldn't it slow the CI if we download and install it each time?

otx/algorithms/common/adapters/__init__.py
```diff
@@ -0,0 +1,15 @@
+"""Adapters for OTX Common Algorithms."""
+
+# Copyright (C) 2022 Intel Corporation
```
ashwinvaidya17: Minor issue but for consistency we can stick to spdx license header

requirements/common.txt
```diff
@@ -0,0 +1,3 @@
+# # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
+# Common Tasks Requirements.                                                       #
+-e otx/algorithms/common/adapters/mpa
```
ashwinvaidya17: Why is this a separate requirement and why does it point to an internal folder? I think a similar thing can be achieved from setup.py https://docs.python.org/3/distutils/examples.html?highlight=package_dir#pure-python-distribution-by-package. But I am not clear with this design. If we want the stuff in mpa to be accessable from `otx` root then we should probably restructure our repository.

otx/algorithms/common/configs/base.py
```diff
@@ -0,0 +1,253 @@
+"""Base Configuration of OTX Common Algorithms."""
```
ashwinvaidya17: Is this a base configuration for all the models? What if there are models which do not require training (like anomaly)? Should we move it to `base/configs/training` or something similar that implies that models that require training should import this config.

otx/algorithms/common/utils/hooks.py
```diff
@@ -0,0 +1,569 @@
+"""Collections of hooks for common OTX algorithms."""
+
```
ashwinvaidya17: These hooks seem mmx specific. Should we move it to `common/utils/mmx` or something similar?

otx/algorithms/common/utils/runner.py
```diff
@@ -0,0 +1,153 @@
+"""Runner with cancel for common OTX algorithms."""
```
ashwinvaidya17: Same here

otx/algorithms/detection/tasks/dataset/__init__.py
ashwinvaidya17: I agree. Adapters will be a better location

otx/algorithms/detection/tasks/nncf.py
```diff
@@ -0,0 +1,436 @@
...
+            task_type_to_label_domain(self._task_type),
+            random_seed=42,
+        )
+        set_hyperparams(self._config, self._hyperparams)
+        self.confidence_threshold: float = self._hyperparams.postprocessing.confidence_threshold
+
+        # Set default model attributes.
+        self._optimization_methods = []  # type: List
+        self._precision = self._precision_from_config
+        self._optimization_type = ModelOptimizationType.MO
```
ashwinvaidya17: Isn't `optimization_type` being overwritter on line 157?

otx/algorithms/detection/tasks/nncf.py
```diff
@@ -0,0 +1,436 @@
...
+                        get_fake_input_func=get_fake_input,
+                    )
+                    logger.info("Loaded model weights from Task Environment and wrapped by NNCF")
+                else:
+                    try:
+                        load_state_dict(model, model_data["model"])
+
+                        # It prevent model from being overwritten
+                        # if "load_from" in self._config:
+                        #    self._config.load_from = None
```
ashwinvaidya17: Minor issue but if these lines are not needed then we can remove them.

ashwinvaidya17 (commented):
Thanks. I have a few comments.

harimkang:
> Thanks. I have a few comments.

Hi @ashwinvaidya17 ,
Thanks for the review!
I'm still thinking about the mpa (& each algo) installation and this PR still has work to do. I will continue to work with reference to the comments. Also, when this PR is ready, I will request a review again. :)

otx/algorithms/detection/tasks/dataset/__init__.py
harimkang: @supersoob Could you please check this one more time?

tox.ini
harimkang: Removed this.

otx/algorithms/common/adapters/__init__.py
harimkang: The header was constructed by referring to anomaly as much as possible. Could you check it one more time?

requirements/common.txt
harimkang: There are still a lot of problems with installation (related to lib version). In this PR, only the necessary modifications are made in tox.ini (Not requirements/) so i removed this file

otx/algorithms/detection/adapters/mmdet/dataset/mmdataset.py
```diff
@@ -0,0 +1,238 @@
+"""Base MMDataset for Detection Task."""
```
supersoob: I'd like to suggest to change this file name to otx_det_dataset.py. What do you think?

otx/algorithms/common/configs/base.py
harimkang: I have reviewed your suggestions. Currently, everything in `otx/algorithms/common` is for the training backend (These are files commonly used in classification/detection/segmentation training. not anomaly). 

The name `common` doesn't seem appropriate, do you have any other ideas?

otx/algorithms/common/utils/hooks.py
harimkang: Thanks for the nice comment. I modified it in a similar way, please check it again.

otx/algorithms/detection/tasks/nncf.py
harimkang: Good point. thanks! This has been corrected.

otx/algorithms/detection/tasks/nncf.py
harimkang: Yes I deleted this.

otx/algorithms/detection/adapters/mmdet/config_utils.py
```diff
@@ -0,0 +1,394 @@
+"""Collection of utils for task implementation in Detection Task."""
```
supersoob: Why don't we put this file into adapters/mmdet/utils folder? I think it's better to put all files in subfolders of mmdet. So make it remain no files in mmdet except for` __init__.py` like anomaly.

otx/algorithms/detection/adapters/mmdet/pipelines.py
```diff
@@ -0,0 +1,139 @@
+"""Collection Pipeline for detection task."""
```
supersoob: Similar comments with above. I'd like to suggest put this in `data` folder together with dataset files.

otx/algorithms/detection/adapters/mmdet/dataset/mmdataset.py
harimkang: I took a look at `anomalib `to see along with other comments below.
This has been modified to `mmdet/data/dataset.py`.

otx/algorithms/detection/adapters/mmdet/pipelines.py
harimkang: I moved it inside `mmdet/data`.

otx/algorithms/detection/adapters/mmdet/config_utils.py
harimkang: I moved it inside `mmdet/utils`. Thanks for the nice comment.

supersoob: It looks great!

otx/algorithms/common/configs/base.py
sungmanc: I think it is not mandatory that all configurations defined under this file are consumed by all models even if it's named 'common'. If some configurations are frequently used for different tasks, then 'common' is the proper location.

So, how about using `common/configs/training_base.py` for these configurations?

harimkang:
Hi code reviewers (with @ashwinvaidya17)
Currently, with the script that creates the `venv` in `otx/algorithms/detection/init_venv.sh`, it was confirmed that the otx cli & sample (with nncf, pot) run well locally, but with other installation methods (in tox), some version conflicts still occur.

so i would like to take a PR that adds a test code **separately**. (next PR)
in the next PR, I will add the unit-test and integration test and work on making it work in the otx environment.

Please review with reference to the above.

otx/algorithms/common/utils/data_utils.py
```diff
@@ -0,0 +1,55 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions
+# and limitations under the License.
+
+from mpa.utils.logger import get_logger
+
+logger = get_logger()
```
eunwoosh: isn't it better to remove logger and "import get_logger", if not used?

otx/algorithms/common/utils/data_utils.py
```diff
@@ -0,0 +1,55 @@
...
+    for i, item in enumerate(dataset):
+        item_labels = item.annotation_scene.get_labels()
+        for i_l in item_labels:
+            if i_l in labels:
+                img_indices[i_l.name].append(i)
+
+    return img_indices
+
+
+def get_old_new_img_indices(labels, new_classes, dataset):
```
sungmanc: The reason why this method is located in `common/utils/data_utils.py` is our default learning method is `class_incremental`? I'm not sure that methods related to `class_incremental` are exposed on `common`.

Even if we selected `class_incremental` as a base, it is unfamiliar to developers who used a conventional framework.

otx/algorithms/detection/tasks/train.py
```diff
@@ -0,0 +1,265 @@
...
+            "labels": labels,
+            "confidence_threshold": self.confidence_threshold,
+            "VERSION": 1,
+        }
+        if (
+            self._model_cfg
+            and hasattr(self._model_cfg.model, "bbox_head")
+            and hasattr(self._model_cfg.model.bbox_head, "anchor_generator")
+        ):
+            if getattr(
```
eunwoosh: how about modifying getattr to hasattr and the line with the line right above it?

otx/algorithms/detection/tasks/train.py
```diff
@@ -0,0 +1,265 @@
...
+        # update checkpoint to the newly trained model
+        self._model_ckpt = model_ckpt
+
+        # Update anchors
+        if (
+            self._model_cfg
+            and hasattr(self._model_cfg.model, "bbox_head")
+            and hasattr(self._model_cfg.model.bbox_head, "anchor_generator")
+        ):
+            if getattr(
```
eunwoosh: how about modifying getattr to hasattr and the line with the line right above it?

otx/algorithms/detection/adapters/mmdet/config_utils.py
harimkang: Please resolve comments that have been resolved. :)

otx/algorithms/common/utils/data_utils.py
harimkang: This is something I agree with to some extent. 
This needs to be cleaned up, but with the current structure, it is most efficient to proceed in the training task. 

I think this will also be included in the backlog of future work. Thanks for the good comments.

otx/algorithms/common/utils/data_utils.py
harimkang: Thanks for the nice comment. Removed.

otx/algorithms/detection/tasks/train.py
harimkang: Edited.

otx/algorithms/detection/tasks/train.py
harimkang: Edited as above.

otx/algorithms/common/utils/data_utils.py
sungmanc: I agree, let's remain some refactorings related to architecture.

otx/algorithms/detection/tasks/train.py
eunwoosh: I'm sorry I forgot the word combine. isn't it better to combine all hasattr lines together?

otx/algorithms/detection/tasks/train.py
eunwoosh: it's same as above

otx/algorithms/detection/utils/otx_utils.py
```diff
@@ -0,0 +1,220 @@
+"""Utils for OTX Detection."""
```
ashwinvaidya17: Should we keep this just to `data.py` and the `otx_utils.py` to just `utils.py`? As they are in the utils folder the postfix can be dropped.

otx/algorithms/common/configs/base.py
ashwinvaidya17: That's true @sungmanc. Plus some anomaly models also require training parameters such as early stopping. I am fine with `common/configs/training_base.py`. Maybe others might also have some suggestions on the naming.

ashwinvaidya17 (approved):
Thanks I am fine with these changes. We can look into the tests in a different PR.

otx/algorithms/detection/tasks/train.py
harimkang: reflected.

otx/algorithms/detection/tasks/train.py
harimkang: reflected.

otx/algorithms/detection/utils/otx_utils.py
harimkang: I agree with the comments and renamed this.

otx/algorithms/common/configs/base.py
harimkang: Renamed `base.py` in `common/configs` and `tasks` to `training_base.py`. Thanks for the nice comment.

PR #1292, opened 2022-10-12
[NNCF] Early GPU model move

daniil-lyakhov:
Fix to enable FP16 NNCF training.
Motivation is to move the model to GPU before ` register_default_init_args` call. If this function is called with a model on CPU, NNCF uses CPU for initialization stage. When model use half precision it leads to runtime error.


external/mmsegmentation/segmentation_tasks/apis/segmentation/nncf_task.py
```diff
@@ -149,6 +149,9 @@ def _load_model(self, model: ModelEntity):
             raise ValueError(f"No trained model in project. NNCF require pretrained weights to compress the model")
 
         self._compression_ctrl = compression_ctrl
+
+        if torch.cuda.is_available():
+            model.cuda()
```
alexsu52, line 154: I would recommend to do it in the `optimize` function.

daniil-lyakhov:
Moved to https://github.com/openvinotoolkit/mmsegmentation/pull/95

PR #1293, opened 2022-10-12
[NNCF] Enable FP16 in mmdetection

daniil-lyakhov:
With improvement introduced in https://github.com/openvinotoolkit/nncf/pull/1279 models could be optimized in FP16 precision. This change allows to use FP16 during optimization stage

### DIFF
--- a/external/mmdetection/detection_tasks/apis/detection/nncf_task.py
+++ b/external/mmdetection/detection_tasks/apis/detection/nncf_task.py
@@ -230,12 +230,14 @@ class OTEDetectionNNCFTask(OTEDetectionInferenceTask, IOptimizationTask):
         self._is_training = True
         self._model.train()
 
+        meta = dict()
         train_detector(model=self._model,
                        dataset=mm_train_dataset,
                        cfg=training_config,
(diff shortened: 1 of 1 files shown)


external/mmdetection/detection_tasks/apis/detection/nncf_task.py
```diff
@@ -230,12 +230,6 @@ def optimize(
         self._is_training = True
         self._model.train()
 
-        fp16 = training_config.get("fp16", None)
```
alexsu52, line 233: Should NNCF version be updated for this changes?

alexsu52 (changes requested):

external/mmdetection/detection_tasks/apis/detection/nncf_task.py
daniil-lyakhov: Yes, it should

alexsu52: Let's update it here

sovrasov:
Closed as outdated

PR #1294, opened 2022-10-12
[NNCF] FP16 NNCF support mmsegmentation

daniil-lyakhov:
Purpose of this PR is to check NNCF is working in FP16 precision. This is basic POC which could possibly broke other scenarios.

FP16 NNCF configs for https://github.com/openvinotoolkit/mmsegmentation/pull/95


daniil-lyakhov:
Purpose of this PR is to check NNCF is working in FP16 precision. This is basic POC which could possibly broke other scenarios.

sovrasov:
Closed as outdated

PR #1298, opened 2022-10-18
[OTX] Upgrade openvino requirements & Add detection integration test codes

harimkang:
## Summary
Modify requirements/openvino.txt with upgrading openvino & nncf version for numpy version
Add detection training installation (requirements)
Add detection test codes (Integration)

### This PR includes :
- Check requirements/openvino.txt for upgrading openvino & nncf version
  - openvino & openvino-dev 2022.1.0 -> 2022.2.0
  - nncf 2.2.0 -> 2.3.0
- Fix wrong tests/ configration codes
- Fix init_venv for Detection training
- Detection Requirements & setup
- Add Detection test code (Integration) - CLI & API

### Test Codes
```
# Detection Sample Code
bash otx/algorithms/detection/init_venv.sh ./venv python3.8
source venv/bin/activate
python -m otx.algorithms.detection.tools.detection_sample otx/algorithms/detection/configs/detection/mobilenetv2_atss/template.yaml --export

# CLI Tests
pytest -s -v tests/integration/cli/detection/test_detection.py
# API Tests (E2E)
pytest -s -v tests/integration/api/detection/test_api_detection.py
```


requirements/openvino.txt
```diff
@@ -1,7 +1,7 @@
 # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
```
harimkang, line 1: Hi @ashwinvaidya17,
A numpy conflict is occurring in the existing requirements (numpy 1.19) for OTX detection training
So I want to upgrade the versions of openvino and nncf..
Could you please confirm that there are no other issues with this change on Anomaly?
(This PR is not ready for review, so there is no need for a separate review at the moment. I will make a request later.)

+ c.c @goodsong81

ashwinvaidya17: @harimkang Yeah it shouldn't be a problem. Please go ahead 👍

otx/api/usecases/exportable_code/demo/requirements.txt
```diff
@@ -1,3 +1,2 @@
-openvino==2022.1.0
```
harimkang, line 1: The otx demo currently uses that requirement. Since nncf, openvino, and otx are referencing out-of-date versions, problems may occur in the demo after the version update.

I've modified this to a working version, but it seems need code to automatically update it to always have the latest requirements.
(+ I have commented out the test command to check this requirement by creating an additional venv // in tests.py)

harimkang:
Hi all reviewers, currently, not all tests pass (I'm checking it out.), but I would appreciate it if you could review the contents so far while I am fixing test cases. So I updated it to 'Ready for review' status.

tests/integration/api/detection/test_detection.py
```diff
@@ -0,0 +1,296 @@
...
+
+
+class TestDetectionTaskAPI:
+    """
+    Collection of tests for OTE API and OTE Model Templates
+    """
+
+    @e2e_pytest_api
+    def test_reading_detection_cls_incr_model_template(self):
+        detection_template = ["mobilenetv2_atss_cls_incr"]
```
jaegukhyun: Is atss the only supported model?

harimkang: I'll edit that file and tell you again. Thanks for the comment.

setup.py
```diff
@@ -221,16 +231,19 @@ def get_requirements(requirement_files: Union[str, List[str]]) -> List[str]:
...
+                    requirement = Requirement.parse(package)
+                    if requirement.name in ("torch", "torchvision"):
+                        package = update_torch_requirement(requirement)
                     requirements.append(package)
 
     return requirements
 
 
-REQUIRED_PACKAGES = get_requirements(requirement_files=["base", "dev", "openvino"])
+REQUIRED_PACKAGES = get_requirements(requirement_files=["base", "dev", "openvino", "detection"])
```
harimkang: Unfortunately, mmdetection requires installation in an environment with torch installed. So, I want to guide users to install torch for each environment and install otx to make it work.
(torch install -> otx install,  it working now)

harimkang: It seems to work well locally, but it seems that the environment is not properly installed in the tox test that works in github action. 

@ashwinvaidya17 Could you give me some advice? I'm not familiar with the tox. It looks like mmcv is installed without being properly built.. 

test works fine with `otx/algorithms/detection/init_venv.sh` env

otx/cli/utils/tests.py
```diff
@@ -210,43 +211,49 @@ def otx_deploy_openvino_testing(template, root, otx_dir, args):
...
-                "install",
-                "-r",
-                os.path.join(deployment_dir, "python", "requirements.txt"),
-            ],
-            cwd=os.path.join(deployment_dir, "python"),
-        ).returncode
-        == 0
-    )
-
+    # TODO: Need to check Requirements.txt & new environment is working
```
harimkang, line 214: The commands below are currently messing up venv due to an invalid requirement file.
I want to comment this out first and work it out with the requirements in the demo later.

setup.py
ashwinvaidya17: I'll have a look

ashwinvaidya17: @harimkang is there a reason that torch version is limited 1.9.0 in mmdet https://github.com/openvinotoolkit/mmdetection/blob/ote/requirements/build.txt#L4? This ends up overwriting the torch version on ote and leads to mmcv-full installation removing the old torch version using which mmcv is build. It leads to `libtorch_cuda_cu.so: cannot open shared object file: No such file or directory`

harimkang: Thanks for the detailed analysis! 
Perhaps there is no other reason and the our mmdetection version is a bit lower than expected..  We plan to change it to public mmdet later and update the version, but not right now.  Tomorrow I will check if I can control the mmcv version and the torch version of mmdet.

However, if this is difficult, first leave only pre-commit, and in the github action, the detection test should be omitted and put again next time..

ashwinvaidya17: But I don't think this is the only issue here. I set the version to 1.9.0 in base.txt and I am still getting some errors. I am still investigating the issue.

tests/integration/api/detection/test_detection.py
harimkang: Fixed!

setup.py
harimkang: First of all, it has been modified so that the detection test is not run in tox. I'll add it later when the installation is properly organized..

sungmanc:
@harimkang , I got below error although I followed your instruction to install. Do you know about this issue?
image


CLI: python -m otx.algorithms.detection.tools.detection_sample otx/algorithms/detection/configs/detection/mobilenetv2_atss/template.yaml --export

harimkang:
> 
@sungmanc 
Yes, this is an issue that appears when running nncf, but it was not always an issue. Perhaps it will be resolved if additional work related to nncf is done on the mpa side.. So, for now, I think you can ignore it.

alexsu52:
> > 
> 
> @sungmanc Yes, this is an issue that appears when running nncf, but it was not always an issue. Perhaps it will be resolved if additional work related to nncf is done on the mpa side.. So, for now, I think you can ignore it.

@harimkang , @goodsong81  Do you a ticket on fixing it? As I remember, the detection sample worked in the legacy code.

harimkang:
> > > 
> > 
> > 
> > @sungmanc Yes, this is an issue that appears when running nncf, but it was not always an issue. Perhaps it will be resolved if additional work related to nncf is done on the mpa side.. So, for now, I think you can ignore it.
> 
> @harimkang , @goodsong81 Do you a ticket on fixing it? As I remember, the detection sample worked in the legacy code.

There are no tickets for this yet. This issue is also not reproduced in the develop or release branch. so we'll be looking into it a bit more next week.

harimkang:
@ashwinvaidya17 The detection test works!  Thanks for the help.  However, most of the results are failing due to OOM issues.

harimkang:
@ashwinvaidya17 I found that some tests were grabbing the resource and I dealt with it. It was confirmed that the CI test, including the detection CLI test, works well and passes.

goodsong81 (approved):

PR #1304, opened 2022-10-19
Fix anomaly hpo save model bug

chuneuny-emily:
This PR solves bug that error raises when run HPO on anomaly tasks.


chuneuny-emily:
test PR

PR #1307, opened 2022-10-21
Minor changes to setup

ashwinvaidya17:


ashwinvaidya17:
@harimkang I tried these changes and they seem to work. Can you try running `tox -re detection` locally on this branch. If it works we can add it to your PR.

PR #1308, opened 2022-10-23
[NNCF] Update test nncf_graph 

AlexanderDokuchaev:
* Add feature to generate reference graph by set the global variable `NNCF_TEST_REGEN_DOT`
* Add reference graphs for anomaly models
* Task: 85980, 85990, 85991. 85992


AlexanderDokuchaev:
run ote-test

external/anomaly/tests/anomaly_common.py
```diff
@@ -107,3 +107,19 @@ def get_anomaly_domain_test_action_classes(
         OTETestNNCFExportEvaluationAction,
         OTETestNNCFGraphAction,
     ]
+
+
+def get_dummy_compressed_model(task):
```
samet-akcay: Would it be possible to annotate the args with type?

ote_sdk/ote_sdk/test_suite/training_tests_actions.py
```diff
@@ -562,15 +562,23 @@ def __call__(self, data_collector: DataCollector, results_prev_stages: OrderedDi
 
 
 # TODO: think about move to special file
-def check_nncf_model_graph(model, path_to_dot):
+def check_nncf_model_graph(model, path_to_dot, force_regen_dot=False):
```
samet-akcay: Here as well. Would be good to know the arg and return types, which improves the readability.

samet-akcay (commented):
Thanks @AlexanderDokuchaev. I have just few comments regarding the PR.

I've another concern regarding the integration with the otx branch, though. We need to think about how/where we merge these changes in the `feature/otx` [branch](https://github.com/openvinotoolkit/training_extensions/tree/feature/otx). 

@ashwinvaidya17, maybe we could discuss this?

external/anomaly/tests/anomaly_common.py
AlexanderDokuchaev: done

ote_sdk/ote_sdk/test_suite/training_tests_actions.py
AlexanderDokuchaev: done

AlexanderDokuchaev:
Changed target branch

external/anomaly/tests/reference/ote_anomaly_classification_padim/nncf/nncf_quantization.dot
```diff
@@ -0,0 +1,240 @@
+strict digraph  {
```
ashwinvaidya17, line 1: Why are these dot files added? Can't we regenerate these from the tests?

external/anomaly/tasks/inference.py
```diff
@@ -291,8 +291,11 @@ def save_model(self, output_model: ModelEntity) -> None:
     def _set_metadata(self, output_model: ModelEntity):
         output_model.set_data("image_threshold", self.model.image_threshold.value.cpu().numpy().tobytes())
         output_model.set_data("pixel_threshold", self.model.pixel_threshold.value.cpu().numpy().tobytes())
-        output_model.set_data("min", self.model.min_max.min.cpu().numpy().tobytes())
-        output_model.set_data("max", self.model.min_max.max.cpu().numpy().tobytes())
+        if hasattr(self.model, "min_max"):
+            output_model.set_data("min", self.model.min_max.min.cpu().numpy().tobytes())
+            output_model.set_data("max", self.model.min_max.max.cpu().numpy().tobytes())
+        else:
+            logger.warning("Model saved without min_max")
```
ashwinvaidya17: Maybe we should mention that model wasn't initialized properly.

ote_sdk/ote_sdk/test_suite/training_tests_actions.py
```diff
@@ -562,15 +565,25 @@ def __call__(self, data_collector: DataCollector, results_prev_stages: OrderedDi
 
 
 # TODO: think about move to special file
-def check_nncf_model_graph(model, path_to_dot):
+def check_nncf_model_graph(
+    model: "NNCFNetwork", path_to_dot: str, force_regen_dot: bool = False
```
ashwinvaidya17, line 569: Why is `force_regen_dot` set to `False`? Personally, I am in the favour of avoiding committing `dot` files to the repo. The tests might fail if the model architecture changes or if a new model is added with missing dot files.

ashwinvaidya17 (commented):
Thanks. I have a few comments.

external/anomaly/tests/reference/ote_anomaly_classification_padim/nncf/nncf_quantization.dot
AlexanderDokuchaev: No. The idea of test to compare graph of model with reference that generated when it works correctly. If graph changes without update nncf configs, it will be unwanted behavior. It's like reference evaluation metric, you want to compare current state with reference.

ashwinvaidya17: Okay that makes sense. But it is something for us to keep in mind when we make any changes to the model.

ote_sdk/ote_sdk/test_suite/training_tests_actions.py
AlexanderDokuchaev: > The tests might fail if the model architecture changes or if a new model is added with missing dot files.

Yes, it was the main reason to add this test.


The test added to indicate that model was changed. It's matter because optimization configs was added for specific model and if model will be changes optimization can work incorrectly.


`force_regen_dot` added to simplify method to update reference graph, it does not change the test. It's works before but to generate the reference graph we need to add some code. Now it can be done by set environment variable.

external/anomaly/tasks/inference.py
AlexanderDokuchaev: I think no, because a model was initialized properly. It happens if save a model without running training pipeline. 
Can you suggest more correctly warning message?

ashwinvaidya17: Min-Max is assigned when `setup` is called. It is initialized when the training pipeline is run. If we just create a model and do not call the train loop, it will in consequence not initialize the required classes used for normalization. The output produced in this case will not be good.

ashwinvaidya17: Maybe we can say something like `the model was not trained before saving. This will lead to incorrect normalization of the heatmaps.`

AlexanderDokuchaev: Updated

ashwinvaidya17: Thanks

ashwinvaidya17 (commented):
Looks good to me. I'll wait for @samet-akcay 's suggestions before approving this PR

samet-akcay:
Just to give @AlexanderDokuchaev a little background, we have been working on `OTE v2`, of which the first phase will soon be merged.

The tricky part in this situation is that the changes need to appear on both the `develop` and `feature/otx` branches. In the `feature/otx` branch, the `external` directory has been deprecated and will be removed in favor of the `algorithms` directory. In this case, task related changes should be made to `algorithms/anomaly`. The tests should be located in the `tests` directory in the root folder, not in the task implementation directory. (I'm not sure where exactly to place the tests as there is no test design for this yet.)

Do you have any ideas regarding how to merge this PR in the simplest way, @ashwinvaidya17 and @AlexanderDokuchaev?

AlexanderDokuchaev:
@samet-akcay @ashwinvaidya17 
I created https://github.com/openvinotoolkit/training_extensions/pull/1330 to develop branch.
In this PR i added changes to `training_extensions/otx/*` and `training_extensions/tests/*'.

external/anomaly/tasks/inference.py
```diff
@@ -291,8 +291,13 @@ def save_model(self, output_model: ModelEntity) -> None:
...
         output_model.set_data("pixel_threshold", self.model.pixel_threshold.value.cpu().numpy().tobytes())
-        output_model.set_data("min", self.model.min_max.min.cpu().numpy().tobytes())
-        output_model.set_data("max", self.model.min_max.max.cpu().numpy().tobytes())
+        if hasattr(self.model, "min_max"):
+            output_model.set_data("min", self.model.min_max.min.cpu().numpy().tobytes())
+            output_model.set_data("max", self.model.min_max.max.cpu().numpy().tobytes())
+        else:
+            logger.warning(
+                "The model was not trained before saving. This will lead to incorrect normalization of the heatmaps."
+            )
```
samet-akcay, line 300: @AlexanderDokuchaev, can you make these changes in [otx/algorithms/anomaly/tasks/inference](https://github.com/openvinotoolkit/training_extensions/blob/feature/otx/otx/algorithms/anomaly/tasks/inference.py#L306-L307)?

Once `feature/otx` is merged, `external` directory will be removed.

external/anomaly/tests/anomaly_common.py
```diff
@@ -18,6 +18,10 @@
 from collections import namedtuple
 from copy import deepcopy
 from typing import List, Type
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from nncf.torch.nncf_network import NNCFNetwork
```
samet-akcay, line 24: And will you be able to move all these tests to somewhere in`tests` directory in the root `~/training_extensions/tests`?

samet-akcay: The idea is that we would like to combine all the tests to a single location, rather than scattering them around different places in the repo

external/anomaly/tasks/inference.py
samet-akcay: Never mind, they are there. Thanks

external/anomaly/tests/anomaly_common.py
samet-akcay: Same as above, thanks for addressing this.

samet-akcay (approved):
lgtm, thanks!

PR #1312, opened 2022-10-25
Disable NNCF optimization for FP16 models

daniil-lyakhov:
FP16 model optimization by NNCF will be supported only in OTX2.0. Until then NNCF optimization for that models should be disabled.

### DIFF
--- a/external/model-preparation-algorithm/configs/instance-segmentation/efficientnetb2b_maskrcnn/template.yaml
+++ b/external/model-preparation-algorithm/configs/instance-segmentation/efficientnetb2b_maskrcnn/template.yaml
@@ -40,8 +40,9 @@ hyper_parameters:
       stat_requests_number:
         default_value: 2
     nncf_optimization:
+      visible_in_ui: false
       enable_quantization:
-        default_value: true
+        default_value: false
       enable_pruning:
         default_value: false
       pruning_supported:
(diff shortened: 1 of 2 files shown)


daniil-lyakhov:
CC: @goodsong81

external/model-preparation-algorithm/configs/instance-segmentation/efficientnetb2b_maskrcnn/template.yaml
```diff
@@ -40,8 +40,9 @@ hyper_parameters:
       stat_requests_number:
         default_value: 2
     nncf_optimization:
+      visible_in_ui: false
       enable_quantization:
-        default_value: true
+        default_value: false
```
alexsu52, line 43: I'm not sure if default_value: false will disable NNCF task. Please, check it on the Geti side. As far as I remember, we removed nncf entrypoint: `mpa_tasks.apis.detection.DetectionNNCFTask` to disable the nncf task.

daniil-lyakhov: Done, I'll check results on a SC instance

daniil-lyakhov:
Release branch PR: #1319

alexsu52 (approved):

goodsong81 (approved):

PR #1316, opened 2022-10-27
Monkey patch mmdet and add trainable yolox, ssd, mask r-cnn

cih9088:
- Most of the monkey patches are not to add functional features but just to block the trace of its methods/functions in NNCF.
- Add trainable and NNCF optimisable YOLOX
- Add trainable and NNCF optimisable SSD
- Rearange nncf adapter
- Add instance-segmentation test code
- Add trainable and NNCF optimisable Mask R-CNN


otx/algorithms/detection/adapters/mmdet/config_utils.py
```diff
@@ -272,7 +274,7 @@ def set_data_classes(config: Config, labels: List[LabelEntity]):
     for subset in ("train", "val", "test"):
         cfg = get_data_cfg(config, subset)
         cfg.labels = labels
-        config.data[subset].labels = labels
+        #  config.data[subset].labels = labels
```
goodsong81, line 275: Moved to somewhere? Or not needed actually?

otx/algorithms/detection/adapters/mmdet/nncf/compression.py
```diff
@@ -255,12 +220,14 @@ def dummy_forward(model):
...
                 img_metas = img_metas[0]
-            ctx = model.forward_export_context(img_metas)
+            ctx = model.nncf_trace_context(img_metas)
             logger.debug(f"NNCF will compress a postprocessing part of the model")
         else:
             ctx = model.forward_dummy_context(img_metas)
             logger.debug(f"NNCF will NOT compress a postprocessing part of the model")
         with ctx:
+            # The device where model is could be changed under this context
+            img = [i.to(next(model.parameters()).device) for i in img]
```
goodsong81: forward_dummy_context() also should be moved or fetched to somewhere else.

otx/algorithms/detection/configs/detection/mobilenetv2_ssd/model.py
```diff
@@ -23,16 +23,27 @@
...
         anchor_generator=dict(
             type="SSDAnchorGeneratorClustered",
             strides=(16, 32),
-            reclustering_anchors=True,
+            # FIXME: Add this feature to mpa.det.inferrer
+            # This feature is implemented in mpa.det.trainer but not in mpa.det.inferrer
+            # To fix this, we need DetectionInferenceTask.infer in otx
+            # to take whole dataset, not validation subset only.
+            # This fix could impact other tasks infer method
+            reclustering_anchors=False,
```
goodsong81: We don't need to re-cluster anchors in infer/eval stage as we just need to adjust anchor settings from the training set before train start. Could you explain why you disabled this?

goodsong81:
In addition, could you recover the LR scheduler, etc that I marked as TODO? Thank you!

otx/algorithms/detection/adapters/mmdet/config_utils.py
cih9088: This should be deleted but I left it commented for a reference purpose.
If a dataset is a wrapper dataset such as `RepeatDataset`, `MultiImageMixDataset`, etc., not `OTXDataset`, assigning `labels` config in them does not make sense.
For `RepeatDataset` and others, it is okay to set it safely, since they are not used when initializing but it is not the case for `MultiImageMixDataset` because all the config will be used to initialize it.
Please refer this line in `build_dataset`: https://github.com/open-mmlab/mmdetection/blob/master/mmdet/datasets/builder.py#L78
Note that only `MultiImageMixDataset` uses all the config presented but not other wrapper dataests.

Should I delete it?

otx/algorithms/detection/adapters/mmdet/nncf/compression.py
cih9088: This was resolved in a branch where I am working on `mmdeploy`. Let me cherry-pick relevant commits into this branch.

otx/algorithms/detection/configs/detection/mobilenetv2_ssd/model.py
cih9088: Since this feature is not implemented in `mpa.det.inferrer`, when doing inference, anchors are not reclustered resulting hard-to-debug error.
Inference performance is a disaster since anchors are not the same when training and inference. Further more, it is hard to debug because there is no error at all.

cih9088:
> In addition, could you recover the LR scheduler, etc that I marked as TODO? Thank you!

Will do!

otx/algorithms/detection/configs/detection/mobilenetv2_ssd/model.py
goodsong81: I thought that clustered anchor setting is used in inference time as is.
@jaegukhyun Could you give some advice here?

otx/algorithms/detection/adapters/mmdet/config_utils.py
goodsong81: @harimkang Could you give some advice on this?

harimkang: I think it won't matter. As long as this doesn't affect other training, we can delete it. Alignment is required for Dataset Wrapping configuration. This can also be true for segmentation.
@jaegukhyun What do you think?

harimkang: Let's see the test results later. Especially on the YOLOX side.

otx/algorithms/detection/configs/detection/mobilenetv2_ssd/model.py
cih9088: @goodsong81 @jaegukhyun This has been fixed by https://github.com/openvinotoolkit/training_extensions/pull/1316/commits/92c551f63a0f31dcbb8e6d4326a7192091e8a87f

otx/algorithms/detection/adapters/mmdet/nncf/compression.py
```diff
@@ -236,6 +205,8 @@ def model_eval_fn(model):
         nncf_compress_postprocessing = cfg.get('nncf_compress_postprocessing')
         logger.debug('set should_compress_postprocessing='f'{nncf_compress_postprocessing}')
     else:
+        # TODO: Do we have to keep this configuration?
+        # This configuration is not enabled in forked mmdetection library in the first place
         nncf_compress_postprocessing = True
```
goodsong81, line 211: Seems that there is no setting value in config files so that it means it is always True.
Do you mean that we can remove this config and treat it as True always?

otx/algorithms/detection/configs/detection/mobilenetv2_ssd/model.py
```diff
@@ -23,12 +23,18 @@
 
 __width_mult = 1.0
 
+runner = dict(max_epochs=20)
```
goodsong81: Where did this come from?

otx/algorithms/detection/tasks/nncf.py
```diff
@@ -323,8 +327,7 @@ def optimize(
             loading_stage_progress_percentage=5,
             initialization_stage_progress_percentage=5,
         )
-        learning_curves = DefaultDict(OTXLoggerHook.Curve)  # type: DefaultDict
```
goodsong81, line 326: Not used?

otx/algorithms/detection/tasks/train.py
```diff
@@ -255,3 +245,18 @@ def _generate_training_metrics(learning_curves, scores) -> Iterable[MetricsGroup
...
+        # by this method or by loading trained model
+        if should_cluster_anchors(model_cfg) and len(self._anchors) == 0:
+            cluster_anchors(
+                model_cfg,
+                recipe_cfg,
+                self._data_cfg.data.train.otx_dataset,
+            )
+            self._update_anchors(
+                self._anchors, self._model_cfg.model.bbox_head.anchor_generator
+            )
```
goodsong81, line 262: @jaegukhyun Please review this mechanism.

goodsong81 (commented):
Finally, I'm not sure of this part. Could you explain?
super() may not be the BaseDetector.
image

otx/algorithms/detection/adapters/mmdet/nncf/compression.py
cih9088: Apart from that, in forked mmdet repository, when building NNCF graph, it always does not build postprocessing part.
Please refer: [this](https://github.com/openvinotoolkit/mmdetection/blob/ote/mmdet/models/detectors/base.py#L231)
In `forward_export` method, it always calls `simple_test` method with `postprocess=False` and this is true for other detectors as well.

Whatever `nncf_compress_postprcessing` is, it always do not compress postprocessing part in mmdet.
So [forward_dummy](https://github.com/openvinotoolkit/mmdetection/blob/ote/mmdet/models/detectors/single_stage.py#L68-L75) and [forward_export](https://github.com/openvinotoolkit/mmdetection/blob/ote/mmdet/models/detectors/single_stage.py#L107-L124) do the same thing when tracing graph.

otx/algorithms/detection/tasks/nncf.py
cih9088: This log_config is added in `_initialized` method in [here](https://github.com/openvinotoolkit/training_extensions/blob/feature/otx/otx/algorithms/common/tasks/training_base.py#L231).
It is a duplicate addition.

otx/algorithms/detection/configs/detection/mobilenetv2_ssd/model.py
cih9088: It came from nowhere. This is caused by the change `CustomSSDHead` in [`this commit`](https://github.com/openvinotoolkit/model_preparation_algorithm/commit/3d128fe720a576cec229eb94b9c1c13b625fdd9a).
Although the architectures of old SSDHead and new SSDHead are the same, there are subtle differences when initializing layers which make [`load_sate_dict_pre_hook()`](https://github.com/openvinotoolkit/model_preparation_algorithm/blob/develop/mpa/modules/models/detectors/custom_single_stage_detector.py#L68-L126) go wrong.
I brought back the old initialization in SSDHead in https://github.com/openvinotoolkit/training_extensions/pull/1316/commits/f8c7cbade82d6967021af47c666d5c1abd5b68e4

Can we rename `CustomSingleStageDetector` to `CustomSSD`? That `load_state_dict_pre_hook` method seems like  pretty much bound to `SSD` model only. It could confuse people just like me.

cih9088:
> Finally, I'm not sure of this part. Could you explain? super() may not be the BaseDetector. image

@goodsong81 
Thanks for pointing this out. This is only true for NNCF task.
When doing NNCF, a original model, such as mmdet model, is wrapped by `NNCFNetwork` to do NNCF stuff. 

In MM series, a model is trained with `train_step()` method and `NNCFNetwork` is able to call `train_step()` method, even though it is not defined in it, because `NNCFNetwork` [overrides](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/torch/nncf_network.py#L347-L376) `__getattr__()` to find attributes or methods of a underline model.
The problem arises when a underline model wants to call its parent's method because `self` is bound to `NNCFNetwork`

Note that the `forward` method is fine since it is [properly wrapped](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/torch/nncf_network.py#L237-L249) in `NNCFNetwork`. Please pay attention to [this line](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/torch/nncf_network.py#L246) where `NNCFNetwork` brings a underline model with `get_nncf_wrapped_model()` and calls `__call__()` method which in turn calls `forward()` method with a properly bound `self` which is a underline model (mmdet model).

I managed to bypass it by defining proper wrapper of `train_step()` [by monkey patch](https://github.com/openvinotoolkit/training_extensions/pull/1316/commits/521035b21ad3c2204928bbdcbc12309116e8efa0) but I have a doubt to keep adding extra features by monkey patch. In a way, it seems inevitable that we merely bridge the gap between them but anyway, if you have a better idea, please let me know.

PR #1339, opened 2022-11-07
[Hot-fix] Move drop_last into mpa.cls

harimkang:


harimkang:
Another [solution](https://github.com/openvinotoolkit/training_extensions/pull/1340) has been merged.

PR #1348, opened 2022-11-09
Revert num_workers on classification task.

JihwanEom:
To prevent unexpected dataloader issue, revert num_workers on classification task as (2->0).


JihwanEom:
run ote-test

sungmanc:
AFAIK, Do we plan to num_workers=0 on `multi-label` and `h-label` cases? `multi-class` is not target for the revert

PR #1351, opened 2022-11-10
Update if statement for drop_last in BaseTask

JihwanEom:
previous:
variable dataset contained all of the subsets, `len(dataset)` = `len(train_dataset)+len(val_dataset)`


JihwanEom:
run ote-test

PR #1352, opened 2022-11-10
[RELEASE] Update if statement for drop_last in BaseTask

JihwanEom:
The if statement is checking the entire dataset(train+val), modified to check only the train set.


JihwanEom:
run ote-test

PR #1364, opened 2022-11-14
[OTX] Copy model_api from OMZ repo for pip install

goodsong81:
- Temporary code copy to support standalone pypi pip install otx
- To be redirected to the original module once pypi pip install is supported from OMZ side
- (Also applied to ote_sdk & external directories for reference & completeness)

### DIFF
--- a/external/anomaly/adapters/anomalib/exportable_code/base.py
+++ b/external/anomaly/adapters/anomalib/exportable_code/base.py
@@ -7,8 +7,8 @@
 from typing import Union
 
 import numpy as np
-from openvino.model_zoo.model_api.models import SegmentationModel
-from openvino.model_zoo.model_api.models.types import NumericalValue
+from ote_sdk.usecases.model_api.models import SegmentationModel
+from ote_sdk.usecases.model_api.models.types import NumericalValue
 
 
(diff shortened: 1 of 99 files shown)


goodsong81:
image
(CI is not working due to storage issue)
Verified with detection CLI tests including OV tasks. Should work for other tasks.

ashwinvaidya17 (approved):
Looks good to me but do you know the timeline of the omz team to get it on pypi? I am concerned about having the same code in both otx and ote_sdk. Syncing changes might be a bit tedious.

goodsong81:
> Looks good to me but do you know the timeline of the omz team to get it on pypi? I am concerned about having the same code in both otx and ote_sdk. Syncing changes might be a bit tedious.

No one knows the timeline as the OMZ project was de-prioritized. So, it would take some time and OTE is referring a specific branch (SCv1.1) which has been infrequently updated.
I think we'll have not that much trouble.

harimkang:
From what I've looked at, it doesn't seem to pass the pre-commit test. (Not sure because CI doesn't work.)
Could you please execute this command? `tox -re pre-commit`

ashwinvaidya17:
From the logs it looks like the system on which the CI is running is out of memory

PR #1367, opened 2022-11-14
[Develop CI-Fix] Add xfail mark for deprecated model template

harimkang:
## Summary
Add xfail mark for deprecated model template & Enable black format to model.py

## Resolves
- https://jira.devtools.intel.com/browse/CVS-96084

PR #1368, opened 2022-11-14
Disable num_workers>0 at OTE CI test

sungmanc:
Reflect latest commits

### DIFF
--- a/external/model-preparation-algorithm/configs/classification/configuration.yaml
+++ b/external/model-preparation-algorithm/configs/classification/configuration.yaml
@@ -82,7 +82,7 @@ learning_parameters:
     visible_in_ui: true
     warning: null
   num_workers:
-    affects_outcome_of: NONE
+    affects_outcome_of: TRAINING
     default_value: 2
     description:
       Increasing this value might improve training speed however it might
--- a/external/model-preparation-algorithm/tests/ote_cli/test_classification.py
+++ b/external/model-preparation-algorithm/tests/ote_cli/test_classification.py
@@ -46,6 +46,8 @@ args0 = {
         "2",
         "--learning_parameters.batch_size",
         "4",
+        "--learning_parameters.num_workers",
+        "0",
     ],
 }
 
@@ -64,6 +66,8 @@ args = {
         "2",
         "--learning_parameters.batch_size",
         "4",
+        "--learning_parameters.num_workers",
+        "0",
     ],
 }
 
@@ -220,6 +224,8 @@ args0_m = {
         "2",
         "--learning_parameters.batch_size",
         "4",
+        "--learning_parameters.num_workers",
+        "0",
     ],
 }
 
@@ -238,6 +244,8 @@ args_m = {
         "2",
         "--learning_parameters.batch_size",
         "4",
+        "--learning_parameters.num_workers",
+        "0",
     ],
 }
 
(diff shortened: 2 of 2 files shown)


sungmanc:
ote run-test

external/model-preparation-algorithm/configs/classification/configuration.yaml
```diff
@@ -82,7 +82,7 @@ learning_parameters:
     visible_in_ui: true
     warning: null
   num_workers:
-    affects_outcome_of: NONE
+    affects_outcome_of: TRAINING
```
goodsong81, line 85: Speed changes but the output does not.
@harimkang What do you think?

harimkang: What does this variable do? Does that mean it's only used for training?

sungmanc: > Speed changes but the output does not. @harimkang What do you think?

Hmm, I can't understand this sentence. what's the meaning of the output?

sungmanc: > What does this variable do? Does that mean it's only used for training?

That field is required to `hyperparmaeter_override`. If `affects_outcome_of` is set to `NONE`, that variable is not overridden.

image

goodsong81: Output of training = model (accuracy)

goodsong81: It seems that 'learning rate' like h-params are marke as 'affects_outcome_of: TRAINING'.

sungmanc: Yeah, if we want to override a specific hyper-parameter defined on `configuration` (i.e. `learning rate`), then that variable should be set as `affects_outcome_of = TRAINING or INFERENCE`. 

So, in this case, because I want to override `num_workers` variable, I changed the `affects_outcome_of = TRAINING`. And add additional argument `learning_parameters.num_workers=0` for passing the OTE CI test.

However, as @harimkang said in PR #1367, if there is no problem on `num_workers`, then let's drop this PR

sungmanc:
@goodsong81 @harimkang , I have tried 3 trials CI tests, and there is no `num_workers` related FAILED cases.
So, I closed this PR. 

http://validationreports.sclab.intel.com:8006/reports/build_number_report?test_session_build_number=pr-ote-test-740&environment=iotg-dev-workstation-09
http://validationreports.sclab.intel.com:8006/reports/build_number_report?test_session_build_number=pr-ote-test-738&environment=iotg-dev-workstation-12
The result of one CI test case was weird because the number of tests was quite small. However, even in this case, there is no `num_workers` related FAIL case
http://validationreports.sclab.intel.com:8006/reports/build_number_report?test_session_build_number=pr-ote-test-739&environment=iotg-dev-workstation-10

sungmanc:
run ote-test

sungmanc:
run ote-test

sungmanc:
run ote-test

sungmanc:
run ote-test

sungmanc:
run ote-test

sungmanc:
run ote-test

sungmanc:
run ote-test

sungmanc:
We'll use num_workers > 0 for > OTEv0.4.0 and also will handle the stability issue by using a kind of auto-config.
So, I'm going to close this PR

PR #1371, opened 2022-11-15
[OTX/Semi-SL] Support unlabeled data via OTX CLI

supersoob:
**This PR includes:**

- Add cli arguments for unlabeled data
- Convert to otx api UNLABELED dataset items
- Enable semi-sl train type & load semi-sl recipe file

What we discussed about refactoring (e.g. unifying recipe filenames, model cfg change, etc.) will be handled in other PRs.

Please review this PR, too -> https://github.com/openvinotoolkit/model_preparation_algorithm/pull/86/files


jaegukhyun (commented):
LGTM, Let's wait test results and other members' review!

otx/algorithms/classification/tasks/inference.py
```diff
@@ -283,25 +283,41 @@ def _init_recipe(self):
...
-            if train_type == TrainType.SEMISUPERVISED:
-                raise NotImplementedError(f"train type {train_type} is not implemented yet.")
-            if train_type == TrainType.SELFSUPERVISED:
-                raise NotImplementedError(f"train type {train_type} is not implemented yet.")
-            # raise NotImplementedError(f'train type {train_type} is not implemented yet.')
-            # FIXME: Temporary remedy for CVS-88098
             recipe = os.path.join(recipe_root, "class_incr.yaml")
-            logger.warning(f"train type {train_type} is not implemented yet. Running incremental training.")
+
+        if train_type != TrainType.INCREMENTAL:
```
eunwoosh: How about removing this line and change else statement (301 line) to "elif train_type != TrainType.INCREMENTAL:"?
then, branch depth can be decreased.

otx/algorithms/classification/tasks/inference.py
```diff
@@ -347,22 +363,23 @@ def patch_color_conversion(pipeline):
...
-            if self._multilabel:
-                cfg.type = "MPAMultilabelClsDataset"
-            elif self._hierarchical:
-                cfg.type = "MPAHierarchicalClsDataset"
-                cfg.hierarchical_info = self._hierarchical_info
-                if subset == "train":
-                    cfg.drop_last = True  # For stable hierarchical information indexing
             else:
-                cfg.type = "MPAClsDataset"
+                if self._multilabel:
```
eunwoosh, line 372: how about change if statement to elif statement and decreasing indent level of below elif and else statement?

otx/algorithms/classification/tasks/train.py
```diff
@@ -157,6 +157,11 @@ def _init_train_data_cfg(self, dataset: DatasetEntity):
                 ),
             )
         )
+        if len(dataset.get_subset(Subset.UNLABELED)):
```
eunwoosh: how about assign dataset.get_subset(Subset.UNLABELED) to variable and reuse it?

otx/algorithms/classification/utils/data.py
```diff
@@ -60,6 +61,8 @@ class ClassificationDatasetAdapter(DatasetEntity):
             "val_data_root": OptionalDirectoryPathCheck,
             "test_ann_file": OptionalDirectoryPathCheck,
             "test_data_root": OptionalDirectoryPathCheck,
+            "ul_data_root": OptionalDirectoryPathCheck,
```
eunwoosh: I think it's better to use unlabeled_data_loot than jsut ul_data_root to increase readability.
how do you think about?

otx/algorithms/classification/utils/data.py
```diff
@@ -70,8 +73,10 @@ def __init__(
         val_data_root=None,
         test_ann_file=None,
         test_data_root=None,
+        ul_data_root=None,
```
eunwoosh: it's same as above comment. how about change this name to unlabeled_data_root?

otx/algorithms/classification/tasks/inference.py
harimkang: The order of these if statements doesn't seem appropriate. My suggestion is to check the semi first, and then the incremental check separately. 
And in the case of multi_label (or h-label), What recipe does semi-Multilabel use? if it is not currently supported, we need to raise NotImplemented or change to INCREMENTAL

otx/algorithms/classification/tasks/inference.py
```diff
@@ -347,22 +363,23 @@ def patch_color_conversion(pipeline):
                     patch_color_conversion(pipeline_step.transforms)
 
         assert "data" in config
-        for subset in ("train", "val", "test"):
+        for subset in ("train", "val", "test", "unlabeled"):
             cfg = config.data.get(subset, None)
             if not cfg:
                 continue
             if cfg.type == "RepeatDataset":
```
harimkang, line 368: Is this `RepeatDataset` still used in classification? If it's not being used, it's good to remove it.

otx/algorithms/classification/utils/data.py
harimkang: It looks better to use `unlabeled` instead of `ul` for easier recognition.

otx/algorithms/classification/utils/data.py
```diff
@@ -115,6 +120,15 @@ def __init__(
                 dataset_item = DatasetItemEntity(image, annotation_scene, subset=subset)
                 dataset_items.append(dataset_item)
 
+        if ul_data_root is not None:
```
harimkang: You can combine two if statements.

otx/algorithms/detection/tasks/inference.py
```diff
@@ -223,25 +223,35 @@ def _init_recipe(self):
...
-            # recipe = os.path.join(recipe_root, 'pretrain.yaml')
-            raise NotImplementedError(f"train type {train_type} is not implemented yet.")
-        elif train_type == TrainType.INCREMENTAL:
-            recipe = os.path.join(recipe_root, "imbalance.py")
-        else:
-            # raise NotImplementedError(f'train type {train_type} is not implemented yet.')
-            # FIXME: Temporary remedy for CVS-88098
-            logger.warning(f"train type {train_type} is not implemented yet.")
+
+        if train_type != TrainType.INCREMENTAL:
```
harimkang: Same with cls part comments

otx/algorithms/segmentation/tasks/inference.py
```diff
@@ -184,18 +184,26 @@ def _init_recipe(self):
...
-            # recipe = os.path.join(recipe_root, 'pretrain.yaml')
-            raise NotImplementedError(f"train type {train_type} is not implemented yet.")
-        elif train_type == TrainType.INCREMENTAL:
-            recipe = os.path.join(recipe_root, "class_incr.py")
-        else:
-            # raise NotImplementedError(f'train type {train_type} is not implemented yet.')
-            # FIXME: Temporary remedy for CVS-88098
-            logger.warning(f"train type {train_type} is not implemented yet.")
+
+        if train_type != TrainType.INCREMENTAL:
```
harimkang: Same here

otx/cli/datasets/image_classification/dataset.py
```diff
@@ -25,6 +25,7 @@ def __init__(
         train_subset=None,
         val_subset=None,
         test_subset=None,
+        ul_subset=None,
```
harimkang: Same with otx/algorithms/classification/utils/data.py's comment

otx/cli/datasets/semantic_segmentation/dataset.py
```diff
@@ -27,6 +28,7 @@ def __init__(
         train_subset=None,
         val_subset=None,
         test_subset=None,
+        ul_subset=None,
```
harimkang: Same with otx/algorithms/classification/utils/data.py's comment

otx/algorithms/classification/tasks/inference.py
```diff
@@ -347,22 +363,23 @@ def patch_color_conversion(pipeline):
...
-                    cfg.drop_last = True  # For stable hierarchical information indexing
             else:
-                cfg.type = "MPAClsDataset"
+                if self._multilabel:
+                    cfg.type = "MPAMultilabelClsDataset"
+                elif self._hierarchical:
+                    cfg.type = "MPAHierarchicalClsDataset"
+                    cfg.hierarchical_info = self._hierarchical_info
+                    if subset == "train":
+                        cfg.drop_last = True  # For stable hierarchical information indexing
```
JihwanEom, line 378: Oh drop last is here :( it should be removed on another PR, I'll create PR for this soon.

otx/algorithms/classification/utils/data.py
```diff
@@ -115,6 +120,15 @@ def __init__(
...
+            if not self.data_type == ClassificationType.MULTIHEAD:
```
JihwanEom, line 124: Do we have specific issue for multihead type?

otx/algorithms/classification/utils/data.py
JihwanEom: How about `if ul_data_root:`

harimkang (changes requested):
I left a few comments. This is a concern that there is currently no test code, but at least some sample code would be nice.
Also, because CI of feature/otx does not work, it would be nice to run the pre-commit test and the API & CLI test of all tasks locally and attach the results in this PR.

otx/algorithms/common/utils/data.py
```diff
@@ -14,6 +14,86 @@
...
+
+    def is_valid(file_path):
+        return file_path.lower().endswith(tuple(IMAGE_FILE_EXTENSIONS))
+
+    with open(file_list_path, "r", encoding="UTF-8") as f:
+        file_names = f.read().splitlines()
+    unlabeled_files = []
+    for fn in file_names:
+        file_path = os.path.join(base_root, fn.strip())
+        if is_valid(file_path) and os.path.isfile(file_path):
```
JihwanEom, line 58: Can't we replace this for statement with `filter` and list comprehension?

otx/algorithms/classification/tasks/inference.py
supersoob: It's out of scope of this PR. I hope it can be dealt in other.

otx/algorithms/classification/tasks/inference.py
supersoob: Great

otx/algorithms/classification/utils/data.py
eunwoosh: "is not None" is better in readability. and it's also faster than just "if ul_data_roots"

otx/algorithms/classification/utils/data.py
supersoob: As far as I know, we don't support semi for hierarchical, so disabled to get unlabeled cli arg.

JihwanEom:
I think we need to update document about supporting semi-supervised learning...!!

supersoob:
> I left a few comments. This is a concern that there is currently no test code, but at least some sample code would be nice. Also, because CI of feature/otx does not work, it would be nice to run the pre-commit test and the API & CLI test of all tasks locally and attach the results in this PR.

Thanks for the comments. I think making api/cli test codes for semi is not good idea because it's not complete version of semi, which eventually fails anyway with this implementation. I will try to make a sample test code so that other members can debug easily. And I will attach the image file after running API CLI tasks for all tasks.

otx/algorithms/common/utils/data.py
```diff
@@ -14,6 +14,86 @@
...
+
+    Returns:
+        List[str]: a list of existing image file paths which will be unlabeled data items.
+    """
+
+    def is_valid(file_path):
+        return file_path.lower().endswith(tuple(IMAGE_FILE_EXTENSIONS))
+
+    with open(file_list_path, "r", encoding="UTF-8") as f:
+        file_names = f.read().splitlines()
```
eunwoosh, line 54: how about just using readlilnes() instead of read().splitlines() ?

otx/algorithms/detection/tasks/train.py
```diff
@@ -210,12 +210,13 @@ def _init_train_data_cfg(self, dataset: DatasetEntity):
                     otx_dataset=dataset.get_subset(Subset.VALIDATION),
                     labels=self._labels,
                 ),
-                unlabeled=ConfigDict(
-                    otx_dataset=dataset.get_subset(Subset.UNLABELED),
-                    labels=self._labels,
-                ),
             )
         )
+        if len(dataset.get_subset(Subset.UNLABELED)):
```
eunwoosh: same as above comment. how about reuse dataset.get_subset(Subset.UNLABELED) by using variable?

otx/algorithms/segmentation/tasks/train.py
```diff
@@ -152,6 +152,11 @@ def _init_train_data_cfg(self, dataset: DatasetEntity):
                 ),
             )
         )
+        if len(dataset.get_subset(Subset.UNLABELED)):
```
eunwoosh: I think other tasks have same part of this. how about making this part as function and using it to reduce duplication?

otx/algorithms/common/utils/data.py
```diff
@@ -14,6 +14,86 @@
...
+        if is_valid(file_path) and os.path.isfile(file_path):
+            unlabeled_files.append(file_path)
+    return unlabeled_files
+
+
+@check_input_parameters_type({"data_root_dir": DirectoryPathCheck})
+def load_unlabeled_dataset_items(
+    data_root_dir: str,
+    file_list_path: Optional[str] = None,
+    subset: Subset = Subset.UNLABELED,
```
eunwoosh: why does this function need to get subset argument?
should subset value always be Subset.UNLABELED according to it's name?

otx/algorithms/classification/tasks/train.py
supersoob: I wonder if it's better in code implementation. I'm curious. I asked because I don't know.

otx/algorithms/classification/utils/data.py
supersoob: Fixed. Please check.

otx/algorithms/classification/utils/data.py
supersoob: Fixed. Please check.

otx/algorithms/classification/utils/data.py
supersoob: Fixed. Please check.

otx/algorithms/classification/utils/data.py
supersoob: Fixed. Please check.

otx/cli/datasets/image_classification/dataset.py
supersoob: Fixed. Please check.

otx/algorithms/classification/tasks/train.py
eunwoosh: ah, I just think that if using variable then, we don't have to call dataset.get_subset(Subset.UNLABELED) twice.

otx/cli/datasets/semantic_segmentation/dataset.py
supersoob: Fixed. Please check.

otx/algorithms/classification/tasks/inference.py
harimkang: We just need to delete this line  This is something we missed in our previous cls task refactoring.

supersoob: But still I don't think it needs to be handled in semi supervised learning related PR. This line still remains in develop branch. I can make comments TODO sign for it.

otx/algorithms/classification/tasks/train.py
supersoob: Yeah, makes sense. I made a function to get unlabeled dataset. Please check and resolve related conversations if no objections.

otx/algorithms/common/utils/data.py
supersoob: I don't know if it's good in point of readability. Could you explain why you prefer this and examples?

otx/algorithms/common/utils/data.py
supersoob: Good point, so removed this argument. Please check.

otx/algorithms/segmentation/tasks/train.py
eunwoosh: Thanks for applying! Actually, my intention is making as function not only dataset.get_subset code but also the whole of if statement.
as you know, it is duplicated in all three tasks. so, if whole part is made as function, then remove more duplication.

supersoob: Hi, but it is my view that making this whole if-statement as a function is not good. Because this is not a final implementation for semi-dataset so we don't know what more args are needed for unlabeled mmdataset for each tasks at the last. So I prefer to leave it as is to make it modify easily by this train_cfg in future implementation. And just making a function just for unlabeled cfg is not appropriate and looks better when we declare them inside init_train_data_cfg.

eunwoosh: I got it. Maybe moving "_init_train_data_cfg" to parent class can be better choice. but it's out of this score, I think.

otx/algorithms/common/utils/data.py
```diff
@@ -14,6 +14,92 @@
...
+        List[DatasetItemEntity]: a list of unlabeled dataset item entity.
+    """
+    if file_list_path is not None:
+        data_list = get_unlabeled_filename(data_root_dir, file_list_path)
+
+    else:
+        data_list = []
+
+        for ext in IMAGE_FILE_EXTENSIONS:
+            data_list.extend(glob.glob(f"{data_root_dir}/**/*{ext}", recursive=True))
```
sungmanc, line 84: It seems that the current implementation only can read images under single-level `ImageDirectory` architecture. Is it intended? 

I'm not sure the word  "single-level, multi-level `ImageDirectory`" is appropriate.. So, I added a description for single-level `ImageDirectory` for clarity

single-level `ImageDirectory`
image

multi-level `ImageDirectory`
image

supersoob: Are you asking that current implementation seems like only support single-level dir image search? If I get your question right, No, it does not. By setting recursive=True. It supports multi-level dir image search.

sungmanc: Ah, okay. I was mistaken.

otx/algorithms/classification/utils/data.py
sungmanc: Then, I prefer to leave a log message that can notify the user. i.e.) "In hierarchical classification, semi-SL training is automatically converted to supervised training"

harimkang (approved):
LGTM, Agree that sample code or test code seems hard to add as of now. Each person in charge of Semi-SL Implementation should add the test code later. As long as the PR doesn't affect the existing test, I think it's okay to merge.

otx/algorithms/classification/utils/data.py
supersoob: Oh that error log in _init_recipe when it tries to load semi recipe but it's hierarchical. It would be better we log that warning msg in that code.

otx/algorithms/classification/tasks/inference.py
supersoob: Revised.

JihwanEom (approved):

otx/algorithms/common/utils/data.py
supersoob: I tried at the first time but got weird results so changed to this.

supersoob:
Hi, test is passed except for 1 anomaly test. 
image
I hope you change the state if no objection.

PR #1372, opened 2022-11-15
[OTX][Action] Enable action detection training and inference in OTX

jaegukhyun:
This PR includes enabling training and inference action detection models in otx.
This is draft implementation. After merge this PR, refactoring for action classification and action detection will be followed.
I tried to reduce modification in action classification classes and functions, but some evitable changes is included in this PR

[Changes]
- Add action detection annotations to data/custom_action_recognition folder
- Add action detection utils and models to mmaction adapters
- Add ava dataset as a default action detection dataset to mmaction adapters
- Add model templates and configs of two stage two backbone structure
- Expand ActionClsTask(inference, train) to support action detection, so the name is changed
- Add test code and sample code of action detection


otx/algorithms/action/adapters/mmaction/core/evaluation/ava_utils.py
```diff
@@ -0,0 +1,125 @@
...
+import numpy as np
+from mmaction.core.evaluation.ava_evaluation import (
+    object_detection_evaluation as det_eval,
+)
+from mmaction.core.evaluation.ava_evaluation import standard_fields
+from mmaction.core.evaluation.ava_utils import print_time, read_exclusions
+
+
+# pylint: disable=too-many-locals, too-many-branches
+def ava_eval(predictions, result_type, labels, video_infos, exclude_file, verbose=True, custom_classes=None):
```
harimkang: Is `ava` just a dataset format? As a newcomer, the word `ava` is not familiar. Why is eval here separately? Can't it be attached to the Dataset class?
Also, it would be easier to understand if change the name of the function and the file to `ava_dataset` (eg, `ava_utils `-> `ava_dataset_utils`) instead of `ava`.

otx/algorithms/action/configs/detection/base/ava.py
```diff
@@ -0,0 +1,89 @@
+"""Data configuration for AVA dataset."""
```
harimkang: The name `ava_dataset.py` seems more appropriate.

otx/algorithms/action/adapters/mmaction/core/evaluation/ava_utils.py
jaegukhyun: ava_eval function is in exact same path in latest mmaction(mmaction/core/evaluation).
They name this function as ava_eval because mmaction only support AVA dataset for action detection.
However otx support other dataset format too(e.g. UCF, JHMDB), therefore it is needed to change function's name, and add description as "Adapt from ava_eval from original mmaction"
What do you think the path of this file? I put it in this directory to sync with mmaction. I wonder general strategy of otx for adaptation of inner function of mmfamily

otx/algorithms/action/configs/detection/base/ava.py
jaegukhyun: Actually ava_dataset.py is named from mmaction. As mmaction only supports AVA dataset, so they name dataset code for action detection as ava_dataset.py. It is needed to change ava_dataset.py to detection_dataset.py. 
This config is for AVA dataset. It shares dataset class with other dataset format, however some config info is different.
Therefore the description looks right

otx/algorithms/action/adapters/mmaction/core/evaluation/ava_utils.py
harimkang: Hmm... I agree with this location, but others may have ideas. It would be nice to explain at the OTX development meeting and listen to others opinion.

otx/algorithms/action/configs/detection/base/ava.py
harimkang: You mean the file name must be different from `ava_dataset.py`? If so, I agree with this name. Or `ava_data_pipeline.py` seems fine too. It looks good as it is.

jaegukhyun: ava_data_pipeline.py looks great thanks

otx/algorithms/action/configs/detection/base/ava.py
```diff
@@ -0,0 +1,89 @@
...
+    # Rename is needed to use mmdet detectors
+    dict(type="Rename", mapping=dict(imgs="img")),
+    dict(type="ToTensor", keys=["img", "proposals"]),
+    dict(type="ToDataContainer", fields=[dict(key="proposals", stack=False)]),
+    dict(type="Collect", keys=["img", "proposals"], meta_keys=["scores", "img_shape"], nested=True),
+]
+
+# Dataset structure
+# TODO Sync with latest mmaction2
+# pylint: disable=unhashable-member, no-member
```
harimkang: image

could you check this?

jaegukhyun:
@harimkang @sungmanc @JihwanEom @supersoob Please check this PR

harimkang (approved):
LGTM, I mainly looked at OTX structures. There is no disagreement. Thank you for hard work!
I wish someone would give you another review :)

otx/algorithms/action/configs/detection/two_stage_two_backbone/__init__.py
```diff
@@ -0,0 +1,15 @@
+"""Initialization of Two Stage Two Backbone structure for Action Detection Task."""
```
sungmanc: We usually set the folder name with the name of `model` or `method` like `x3d`.
However, `two_stage_two_backbone` can include many methods. How about changing the name of this folder?

jaegukhyun: Actually, two_stage_two_backbone is like a structure of action detector such as YOLO or Faster-RCNN, and it can take any action classification model such as SlowFast, X3D. What name could be proper for this situation?

sungmanc: How about using `faster_rcnn_with_x3d`? For the detection case, although we can use many kinds of backbones for `SSD / ATSS / YOLOX.`, their name aren't `Onestage Detector`.

jaegukhyun: Change to x3d_fast_rcnn followed from object detection(backbone_structure).

sungmanc (approved):
LGTM

PR #1382, opened 2022-11-21
[OTX-MMCV] Public mmdetection

goodsong81:
- Train / eval / NNCF optimization
- TODO: OV export (to be handled in another PR regarding mmdeploy support)


goodsong81 (commented):
Pickling issue will be resolved in mmseg PR.

PR #1388, opened 2022-11-24
Add log saving feature for training & task.unload() for resolving /tmp/ deletion issue

harimkang:
## Summary
- Add '--save-logs-to' into otx train
- Add task.unload() for OTX CLI
- Add task.unload() for all tests


harimkang:
I'm closing it because it didn't work out this way.

PR #1391, opened 2022-11-25
[OTX][CLI] Public Backbone Support (otx find & build)

harimkang:
## Summary
Public Backbone Support (otx find & build commands)

### This PR includes
- Update otx find command
  - [X] Support available backbone list (mmX + torchvision)
  - [X] Update Required Argument List of Backbone
image

- New otx build command
  - [X] Create OTX user workspace from task_type    
image
  - [X] Create Backbone configuration file (Backbone to support)
  - [X] Update Model configuration file with new backbone
  - [X] Check otx commands with user-workspace
- Backbone Replacement with ATSS & MaskRCNN
  - [X] Automatically changes input channels of the model according to the backbone
- Etc
  - [X] Support new data configuration (data.yaml) in user-workspace
  - [X] Support saving logs commands with `otx train`
  - [X] Update mmdetection branch to `otx`
  - [X] Add CLI test codes

### PR Coverage
|TASK|Arch|Build Support|
|:---|:---|:---:|
|DETECTION|ATSS|O|
|DETECTION|YOLOX|WIP(checking...)|
|INSTANCE_SEGMENTATION|Mask-RCNN|O|
|SEGMENTATION|ClsIncrSegmentor|O|
|CLASSIFICATION|CustomClassifier|X|
|DETECTION|SSD|X|

### Quick Start Guide for this new feature
```
# Show available templates
otx find --template
# Show available backbones per backend
otx find --backbone {mmdet, mmseg, torchvision, pytorchcv, otx}
# Create OTX-workspace
otx build --task detection
cd otx-workspace-detection
# Need to edit data.yaml in workspace
# Traininig in workspace -> no need any inputs just 'otx train'
otx train
# Get Backbone configuration file (Example)
otx build --backbone torchvision.resnet18 {--save-backbone-to} {path/to/backbone-file}
# Edit required args [if needed] of backbone
# Update model.py with new backbone config file
otx build --model model.py --backbone backbone.yaml
# If Backbone doesn't have required args, 
# you can do the above two builds at once.
otx build --model model.py --backbone torchvision.resnet18
# And train them
otx train
```


otx/algorithms/common/adapters/mmcv/models/__init__.py
```diff
@@ -1,4 +1,4 @@
-"""Adapters for OTX Common Algorithm. - mmcls.model."""
+"""Adapters for OTX Common Algorithm. - mmcv.model."""
```
jaegukhyun, line 1: Is this for only mmcv model? I can see the torchvision models

otx/algorithms/common/adapters/mmcv/models/backbones/torchvision_backbones.py
```diff
@@ -0,0 +1,241 @@
+"""Model templates builder."""
```
jaegukhyun: You need more detail description for this file.
I can't understand the purpose of this file through this description

otx/algorithms/common/adapters/mmcv/models/__init__.py
harimkang: Thanks for the review!
Register the torchvision model in the mmcv registry and use it. In addition, OTX's custom backbones are registered and used in the mmcv registry, so these are mmcv models.

otx/algorithms/common/adapters/mmcv/models/backbones/torchvision_backbones.py
harimkang: Yes, I will supplement the description of this and write it. Thanks for the advice.

otx/cli/builder/builder.py
```diff
@@ -0,0 +1,276 @@
...
+}
+
+
+# pylint: disable=too-many-locals, too-many-statements, too-many-branches
+
+
+def get_backbone_out_channels(backbone):
+    """Get output channels of backbone using fake data."""
+    out_channels = []
+    input_size = backbone.input_size if hasattr(backbone, "input_size") else 64
```
jaegukhyun, line 48: I think 64 is too small for default input resolution.
Did you check this value for multiple models?

otx/cli/builder/builder.py
```diff
@@ -0,0 +1,276 @@
...
+
+        task_type: The type of task want to get (str)
+        model_type: Specifies the template of a model (str)
+        workspace_path: This is the folder path of the workspace want to create (os.path)
+        """
+
+        # Create OTX-workspace
+        if workspace_path is None:
+            workspace_path = f"./otx-workspace-{task_type}"
+        os.makedirs(workspace_path, exist_ok=True)
```
jaegukhyun: Is it safe re-using existed workspace? Many templates and configs will be replaced

otx/cli/builder/builder.py
```diff
@@ -0,0 +1,276 @@
...
+                    f"[otx build] {model_type} is not a type supported by OTX {task_type}."
+                    f"\n[otx build] Please refer to 'otx find --template --task_type {task_type}'"
+                )
+            template = template[0]
+        else:
+            template = otx_registry.get(DEFAULT_MODEL_TEMPLATE_ID[task_type])
+        template_dir = os.path.dirname(template.model_template_path)
+
+        # Copy task configuration file
+        task_configuration_path = os.path.join("/".join(template_dir.split("/")[:-1]), "configuration.yaml")
```
jaegukhyun: Is it safe root for semisl configs? I think it's better to use template.hyper_parameters.base_path for finding configuration.yaml

otx/cli/tools/build.py
```diff
@@ -0,0 +1,60 @@
+"""Model templates searching tool."""
```
jaegukhyun: Need to be update. It is description for find.py

otx/cli/tools/build.py
```diff
@@ -0,0 +1,60 @@
...
+    parser.add_argument("--task", help=f"The currently supported options: {SUPPORTED_TASKS}.")
+    parser.add_argument("--model", help="Input OTX model config file (e.g model.py).", default=None)
+    parser.add_argument("--backbone", help="Enter the backbone configuration file path or available backbone type.")
+    parser.add_argument("--save-backbone-to", default="./backbone.yaml")
+
+    return parser.parse_args()
+
+
+def main():
+    """Main function for model templates searching."""
```
jaegukhyun: Same here, this description needs to be update

otx/cli/tools/eval.py
```diff
@@ -111,7 +123,18 @@ def main():
...
+        with open(args.data, "r", encoding="UTF-8") as stream:
+            data_config = yaml.safe_load(stream)
+        stream.close()
+
+        test_ann_files, test_data_roots = (
+            data_config["data"]["test"]["ann-files"],
+            data_config["data"]["test"]["data-roots"],
+        )
+
+    dataset = dataset_class(test_subset={"ann_file": test_ann_files, "data_root": test_data_roots})
```
jaegukhyun: I have a doubt for this logic. It seems that if a workspace contains data.yaml then otx eval will evaluate the trained algorithms through dataset in data.yaml. However what if a user want to evaluate the trained algorithm to another dataset? In this situation user will give the path through --test-data-roots, but this will be replaced to the default data.yaml's information. Could you check this scenario?

otx/cli/tools/explain.py
```diff
@@ -44,16 +44,21 @@ def parse_args():
     """Parses command line arguments."""
 
     pre_parser = argparse.ArgumentParser(add_help=False)
-    pre_parser.add_argument("template")
-    parsed, _ = pre_parser.parse_known_args()
+    if os.path.exists("./template.yaml"):
```
jaegukhyun, line 47: I think we need a unified function for finding template.yaml and that function could be placed to otx/cli/utils

otx/cli/tools/find.py
```diff
@@ -15,17 +15,41 @@
...
+from otx.cli.utils.importing import get_backbone_registry, get_required_args
+
+# pylint: disable=too-many-locals
+
+SUPPORTED_TASKS = (
+    "CLASSIFICATION",
+    "DETECTION",
+    "INSTANCE_SEGMENTATION",
+    "SEGMENTATION",
+    "ACTION_CLASSIFICATION",
```
jaegukhyun, line 38: I didn't know otx find can support action classification

otx/cli/tools/optimize.py
```diff
@@ -118,12 +131,28 @@ def main():
     dataset_class = get_dataset_class(template.task_type)
 
     # Create instances of Task, ConfigurableParameters and Dataset.
+    train_ann_files, train_data_roots = args.train_ann_files, args.train_data_roots
```
jaegukhyun: Same here with eval.py. If someone wants to export another dataset, not in data.yaml?

otx/cli/tools/train.py
```diff
@@ -134,17 +153,40 @@ def main():
     dataset_class = get_dataset_class(template.task_type)
 
     # Create instances of Task, ConfigurableParameters and Dataset.
+    train_ann_files, train_data_roots = args.train_ann_files, args.train_data_roots
```
jaegukhyun: Same here with eval and export. If a user wants to reuse training algorithm(template) with another dataset(not data.yaml)?

otx/cli/tools/train.py
```diff
@@ -199,6 +241,12 @@ def main():
     assert resultset.performance is not None
     print(resultset.performance)
 
+    if args.save_logs_to:
```
jaegukhyun, line 225: save_logs_to is good. Is there any documentation updates for introducing this new args?

jaegukhyun (changes requested):
I left a few comments. Generally it looks good to me, but I have a doubt for loading data from data.yaml and replace the cli data roots. It cause unexpected behavior those who wants to reuse model template to another dataset. Let's discuss about it. In addition I want you to modify some description of file and function

otx/algorithms/common/adapters/mmcv/models/backbones/__init__.py
```diff
@@ -17,5 +17,6 @@
 from .efficientnet import OTXEfficientNet
 from .efficientnetv2 import OTXEfficientNetV2
 from .mobilenetv3 import OTXMobileNetV3
+from .torchvision_backbones import *
```
eunwoosh: If you add this code for executing torchvision_backbones.py, then you can just add code like bellow

`from . import torchvision_backbones`

otx/algorithms/classification/tasks/__init__.py
```diff
@@ -26,7 +26,7 @@
 
 __all__ = [
     "MPAClassification",
-    "MPABackbones",
+    "OTXBackbones",
```
eunwoosh, line 29: I can't find using `MPABackbones`. Should `OTXBackbones` still need to be in `__all__`?

otx/algorithms/common/adapters/mmcv/models/backbones/torchvision_backbones.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+def init_weights(self, pretrained=True):
+    """Init weights function for new model (copy from mmdet)."""
+    if pretrained and self.model_urls:
+        state_dict = load_state_dict_from_url(self.model_urls)
+        self.load_state_dict(state_dict)
+
+
+def generate_torchvision_backbones():
```
eunwoosh, line 168: it's just a question. why don't you just use inheritance? then, you just need to override some of attributes.

otx/algorithms/detection/tasks/__init__.py
```diff
@@ -27,4 +29,5 @@
     "DetectionNNCFTask",
     "OpenVINODetectionTask",
     "MPADetection",
+    "OTXModels",
```
eunwoosh: is it used in other code by `import *`?

otx/cli/tools/train.py
eunwoosh: how about make codes to make data_roots (or dataset?) as separate function?

otx/algorithms/common/adapters/mmcv/models/backbones/torchvision_backbones.py
harimkang: Added additional description to main file. Could you please check it?

otx/cli/builder/builder.py
harimkang: I tested ATSS and Mask-RCNN by fixing the current value (64). Do you think this will have a big impact on other models?

otx/cli/builder/builder.py
harimkang: Changed the value of exist_ok to False so that it is not replaced in a situation where it already exists.

otx/cli/builder/builder.py
harimkang: Edited as per your suggestion.
`task_configuration_path = os.path.join(template_dir, template.hyper_parameters.base_path)`

otx/cli/tools/build.py
harimkang: Added some clarification.

otx/cli/tools/eval.py
harimkang: In my opinion, just like your opinion, it's good to give commands the most importance. Added a function called `configure_dataset` to `otx/cli/utils/config.py` and modified it to import and use it from `train`, `eval`, and `optimize`. could you please check it?

otx/cli/tools/explain.py
harimkang: I've considered this method, but it looks like it needs some more thought. 
I'm not sure if it's appropriate to put a common function in `utils` for just lines 1-3 and then call the function inside the function. [main() -> parse_arg() -> get_template_arg..?]

otx/cli/tools/find.py
harimkang: `otx find` it works fine for all template in `otx`! :)

otx/cli/tools/optimize.py
harimkang: Corrected and explained the same in the comments above.

otx/cli/tools/train.py
harimkang: Corrected and explained the same in the comments above.

otx/cli/tools/train.py
harimkang: Not sure where to add this? Do you mean doc documentation? like start_guide.md..?

otx/cli/tools/build.py
harimkang: fixed!

otx/algorithms/common/adapters/mmcv/models/backbones/__init__.py
harimkang: thanks for the nice comment! Edited.

otx/algorithms/classification/tasks/__init__.py
harimkang: the role of registering the backbone in the registry by importing it. These are required imports that are not used directly in our code, but only serve to register in the registry.

otx/algorithms/common/adapters/mmcv/models/backbones/torchvision_backbones.py
harimkang: This is the code copied from `mmdetection` to register the `pytorchcv` backbone in the registry. (I modified the existing code for torchvision.) I tried to clean it up as much as possible, but the inheritance still remains. 
If mmdetection changes to public later and pytorchcv comes to the otx side, I think it's good to fix it together. What do you think?

otx/algorithms/detection/tasks/__init__.py
harimkang: This is as described in the comments above. import to register the model and backbone to the registry.

otx/cli/tools/train.py
harimkang: As per your advice, I moved it into a separate function (`otx/cli/utils/config.py::configure_dataset`) and modified it to be used when configuring the `train`, `eval`, and `optimize` datasets in otx cli.

otx/algorithms/classification/tasks/__init__.py
eunwoosh: but does it need to be in `__all__`, not just import it in `__init__.py`?

otx/algorithms/common/adapters/mmcv/models/backbones/torchvision_backbones.py
eunwoosh: I understand. let's refactor together later then.

otx/algorithms/classification/tasks/__init__.py
harimkang: I put it in with the thought that it might be used someday.. Is it neat to erase it?

eunwoosh: it won't make problem. I just think that useless functions don't have to be imported when this file is imported.
but if you think they'll be used later, then I can understand, of course.

otx/cli/tools/train.py
supersoob: quick start guide.md needed to be updated I think,, I missed to add unlabeled arg there too. Let's update it together later.

otx/cli/utils/config.py
```diff
@@ -30,3 +34,38 @@ def override_parameters(overrides, parameters):
...
+    # Create instances of Task, ConfigurableParameters and Dataset.
+    data_subset_format = {"ann-files": None, "data-roots": None}
+    data_config = {"data": {subset: data_subset_format.copy() for subset in ("train", "val", "test")}}
+    data_config["data"]["unlabeled"] = {"file-list": None, "data-roots": None}
+    if os.path.exists(args.data):
+        with open(args.data, "r", encoding="UTF-8") as stream:
+            data_config = yaml.safe_load(stream)
+        stream.close()
+        if train:
+            args.save_model_to = "./models"
```
supersoob: Can't we get the custom path from save_model_to arg? It looks like it always saves train ckpt in './models'. Maybe I can miss something so could you check? And I'm curious which path models and logs are if data.yaml is not loaded and just initiated 'otx train' with default path and custom args. In that case -> I guess it's gonna be in 'training_extension/models/' and 'training_extension/logs' ?

otx/algorithms/common/adapters/mmcv/models/backbones/efficientnet.py
```diff
@@ -15,14 +15,15 @@
...
 import torch.nn.functional as F
-from mmcls.models.builder import BACKBONES
 from mmcv.cnn import build_activation_layer
 from mmcv.cnn.bricks import ConvModule
 from mmcv.runner import load_checkpoint
 from mpa.utils.logger import get_logger
 from torch import nn
 from torch.nn import init
 
+from ..builder import BACKBONES
```
supersoob, line 25: I saw this BACKBONES is from registry MMCV_MODELS. Is it compatible with mmcls backbone? I mean did it find well as classification backbones? or these models are used globally across the tasks?

supersoob (commented):
I'm impressed about hard work you did !! :)

otx/algorithms/common/adapters/mmcv/models/backbones/torchvision_backbones.py
```diff
@@ -0,0 +1,247 @@
...
+            torchvision_models[model_key] = model_value
+        elif inspect.ismodule(model_value) and hasattr(model_value, "model_urls"):
+            torchvision_urls.update(model_value.model_urls)
+    return torchvision_models, torchvision_urls
+
+
+TORCHVISION_MODELS, TORCHVISION_MODEL_URLS = get_torchvision_models()
+
+
+def replace_activation(model, activation_cfg):
```
sovrasov, line 52: Is this functionality really used?

otx/cli/tools/train.py
```diff
@@ -90,16 +101,22 @@ def parse_args():
...
         required=False,
         help="Load only weights from previously saved checkpoint",
     )
     parser.add_argument(
         "--save-model-to",
-        required="True",
+        required=False,
         help="Location where trained model will be stored.",
     )
+    parser.add_argument(
```
sungmanc, line 115: Do we need to separate the `--save-model-to` and `--save-logs-to` ?
we might combine together like as `--output_path` then user don't need to add an additional argument.

otx/algorithms/common/adapters/mmcv/models/backbones/torchvision_backbones.py
```diff
@@ -0,0 +1,247 @@
+"""Torchvision Model Backbone Class Generation.
```
sungmanc, line 1: I'm not sure whether the location of `torchvision_backbones.py`  is right or not. 
Because it is a kind of `torch` series, I think `common/adapter/pytorch/` is more suitable for me.

otx/algorithms/common/adapters/mmcv/models/builder.py
```diff
@@ -0,0 +1,21 @@
...
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions
+# and limitations under the License.
+
+from mmcv.cnn import MODELS as MMCV_MODELS
+from mmcv.utils import Registry
+
+BACKBONES = Registry("models", parent=MMCV_MODELS, scope="otx")
+TORCHVISION_BACKBONES = Registry("models", parent=MMCV_MODELS, scope="torchvision")
```
sungmanc, line 21: Same with this part, maybe `Torchvision backbones` should be divided from `MMCV` family.

otx/cli/builder/builder.py
sungmanc: This function just calculates the output channel, so I think the input resolution doesn't matter.

otx/cli/utils/config.py
harimkang: This looks like my mistake. thanks for the nice comment! By default, it targets `otx-workspace/models` and `workspace/logs`, but it seems most appropriate to use those paths if there are commands. I will correct this.

otx/algorithms/common/adapters/mmcv/models/backbones/efficientnet.py
harimkang: The backbone of mmcls also uses the backbone of mmcv. It is registered as one registry and configured to be used at once.
It (mmcls.backbone) need some more checking, but first my goal is to register all backbones in the registry under mmcv.

otx/algorithms/common/adapters/mmcv/models/backbones/torchvision_backbones.py
harimkang: This can be used on the configuration of the backbone. Depending on the value of `activation_cfg`, it goes through this function.

otx/cli/tools/train.py
harimkang: It looks like this will be a bigger task than I thought. (Several test code needs to be fixed.) After discussion with group, it seems appropriate to work on another PR. What do you think?

otx/algorithms/common/adapters/mmcv/models/backbones/torchvision_backbones.py
harimkang: This is a code that converts the backbone class provided by `torchvision` to the `mmcv` backbone class and registers it in the `mmcv` registry. I think it's appropriate here because it's the code that makes it usable in `mmcv`.

otx/algorithms/common/adapters/mmcv/models/builder.py
harimkang: If this is separated, there may be a problem with public backbone support.
This is configured to register all backbones such as `otx`, `mmcls`, `mmdet`, `mmseg`, `torchvision` backbone, etc. in the backbone registry of `mmcv` and use them at once.
Let me know if you have any other ideas :)

otx/cli/utils/config.py
harimkang: Fixed it!

otx/cli/tools/build.py
```diff
@@ -0,0 +1,69 @@
...
+
+def parse_args():
+    """Parses command line arguments."""
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--task", help=f"The currently supported options: {SUPPORTED_TASKS}.")
+    parser.add_argument("--workspace-root", help="The path to use as the workspace.")
+    parser.add_argument("--model", help="Input OTX model config file (e.g model.py).", default=None)
+    parser.add_argument("--backbone", help="Enter the backbone configuration file path or available backbone type.")
+    parser.add_argument("--save-backbone-to", help="Enter where to save the backbone configuration file.", default=None)
+    parser.add_argument("--root", help="A root dir where templates should be searched.", default=".")
```
eunwoosh: I think root could be confusing because it can mean other things in CS. how about to use other words? (For example, template-dir or template-root?)

otx/cli/tools/find.py
```diff
@@ -15,17 +17,50 @@
...
 
 
 def parse_args():
     """Parses command line arguments."""
 
     parser = argparse.ArgumentParser()
-    parser.add_argument("--root", help="A root dir where templates should be searched.", default=".")
-    parser.add_argument("--task_type")
-    parser.add_argument("--experimental", action="store_true")
+    parser.add_argument("--root", help="A root dir where templates should be searched.", default="otx")
```
eunwoosh: same as above comment

otx/cli/tools/build.py
harimkang: This means the root path of otx. Would otx-root be appropriate? I'm not sure.

otx/cli/tools/train.py
jaegukhyun: I meant start guide, it's okay to update later as Soob said

otx/cli/tools/eval.py
jaegukhyun: LGTM

otx/cli/tools/build.py
eunwoosh: I see. then why user needs to set this argument? As far as I know, it can get own package path by itself. does this code consider finding template in other OTX directory?

otx/cli/utils/tests.py
```diff
@@ -517,3 +520,90 @@ def otx_explain_testing(template, root, otx_dir, args):
...
+            "--model",
+            os.path.join(task_workspace, "model.py"),
+            "--backbone",
+            backbone,
+            "--workspace-root",
+            task_workspace,
+            "--root",
+            otx_dir,
+        ]
+        assert run(command_line).returncode == 0
```
jaegukhyun, line 599: I have two questions for test. 
1. Other tests such as train and evaluation will be run in the workspace built by this build test
If not, should we test other cli tools in the workspace?
2. Are there codes for deleting workspace generated by this build test? 
If not, is there no danger about volume in test server?

jaegukhyun (commented):
Generally it looks good to me, but I think others reviews make sense. In addition I have questions about tests

otx/cli/utils/tests.py
harimkang: Thanks for the great question. 
1. This could be added, but the role of `otx build` is to update otx model.py and check till train. So it doesn't check other otx commands. It might be more appropriate to add test code just for the workspace. (Would it be appropriate to be included in this PR..?)
2. I confirmed that this is created in the `/tmp/` folder and deleted after the test in my local area.

otx/cli/tools/build.py
harimkang: This `root` argment is because we need to know the otx root to build the template (when create workspace).
Of course, most users think that they will create a workspace in the otx root, but even in a non-root place, if the root is given as a separate input, it helps to create it there.

otx/algorithms/common/adapters/mmcv/models/backbones/torchvision_backbones.py
sovrasov: I see the code, but does any of existing configuration files use that functionality? Is it possible to update this config via CLI and will users do that?
This function looks like a workaround to avoid using Hardswish activations, because we used to have problems with exporting swish to onnx in older opsets. Overall, it's better to avoid comparisons like `name == "activ"` even in such hacky wrappers, because they are easy to brake.

harimkang: Thanks for the nice comment! This was a code copied from the method used when registering the pytorchcv backbone in `mmdetection`. I didn't know the details...
Does this mean that this function is now that can be removed? If not sure, I'll edit the comparison as per your comment. :)

harimkang: Thanks for the nice comment! This was a code copied from the method used when registering the pytorchcv backbone in `mmdetection`. I didn't know the details...
I tried to change the comparisons as you suggested, but I have no particular idea. Is there any way you can suggest?

otx/cli/tools/build.py
harimkang: Edited as per your suggestion. could you please check it?

otx/cli/tools/find.py
harimkang: same here!

otx/algorithms/common/adapters/mmcv/models/backbones/torchvision_backbones.py
sovrasov: I guess so, we can safely remove that workaround. It silently replaces swish activations, which basically is not expected by users.

harimkang: Yes i understood!
but, could I create a separate PR to remove and test it?
(There are things that need to be manually tested locally. need to check various backbone configs..)

harimkang:
@sungmanc @eunwoosh @jaegukhyun Could you update your status or add a comment for this review?

otx/cli/tools/build.py
eunwoosh: thanks for reflecting!

eunwoosh (approved):

jaegukhyun (approved):
LGTM

otx/cli/tools/train.py
sungmanc: Agreed

sungmanc (approved):

otx/algorithms/common/adapters/mmcv/models/backbones/torchvision_backbones.py
harimkang: I made a jira [ticket](https://jira.devtools.intel.com/browse/CVS-98625) to check this. I'll take the time later to make a related PR and request a review!

kprokofi:
@harimkang 
First of all, thank you for the hard work. 
I have few questions regarding this feature.
image
1) What is the section "Confirmed model". Why do we need it?
2) Why do all the backbones from `pytorchcv` have `mmdet` prefix?
3) Why the command `otx find` do nothing if nothing passed as arguments? I expect all the available templates showing or raising the error. `otx find --task classification` also does nothing. From the user perspective it is not what he expects to see, I think.
4) When I use the command `otx find --backbone pytorchcv --task classification` I expect to see which backbones are appropriate for the classification task, but I see the same output as without the ` --task` parameter. I think it is not intuitive and `deeplabv3_resnetd101b_cityscapes`, for example, isn't appropriate for the classification task.

harimkang:
Thanks for the comment!
> 1. What is the section "Confirmed model". Why do we need it?

This is based on the current architecture I've tested. Based on which model template, a test was conducted to see if it works when the backbone is changed, and this is based on what works well until training. This word is subject to change and may disappear from this table. 
> 2. Why do all the backbones from pytorchcv have mmdet prefix?

Currently `pytorchcv` backbones are registered in the registry within `mmdetection`. This can be fixed in the future if it changes to `public mmdetection` and the `pytorchcv` part comes over to `otx`.

> 3. Why the command otx find do nothing if nothing passed as arguments? I expect all the available templates showing or raising the error. otx find --task classification also does nothing. From the user perspective it is not what he expects to see, I think.

That's a good idea. I will be able to update this in a future PR. 

> 4. When I use the command otx find --backbone pytorchcv --task classification I expect to see which backbones are appropriate for the classification task, but I see the same output as without the  --task parameter. I think it is not intuitive and deeplabv3_resnetd101b_cityscapes, for example, isn't appropriate for the classification task.

Backbone find is designed to work regardless of the task. So when --backbone comes in, --task seems to be ignored. This part seems to require a bit more thought. If you have any good ideas, please share. Currently, it is difficult to test available for each task architecture for all backbones.

Thanks for the detailed review comments. Please tell us your suggestions anytime.
The PR that follows this is the PR below. please refer to this. :)
https://github.com/openvinotoolkit/training_extensions/pull/1460

PR #1392, opened 2022-11-25
[OTX] Enable multi-GPU training

eunwoosh:
### Summary
Enable multi-GPU training for classification, detection, segmentation tasks.

### This PR includes

- Enable multi-GPU training
- make it available to set output_path in each task class.
-  add test code for multi-GPU training

### DIFF
--- a/otx/algorithms/anomaly/tasks/inference.py
+++ b/otx/algorithms/anomaly/tasks/inference.py
@@ -70,7 +70,7 @@ logger = get_logger(__name__)
 class InferenceTask(IInferenceTask, IEvaluationTask, IExportTask, IUnload):
     """Base Anomaly Task."""
 
-    def __init__(self, task_environment: TaskEnvironment) -> None:
+    def __init__(self, task_environment: TaskEnvironment, output_path: Optional[str] = None) -> None:
         """Train, Infer, Export, Optimize and Deploy an Anomaly Classification Task.
 
         Args:
@@ -87,7 +87,7 @@ class InferenceTask(IInferenceTask, IEvaluationTask, IExportTask, IUnload):
         self.base_dir = os.path.abspath(os.path.dirname(template_file_path))
 
         # Hyperparameters.
-        self.project_path: str = tempfile.mkdtemp(prefix="otx-anomalib")
+        self.project_path: str = output_path if output_path is not None else tempfile.mkdtemp(prefix="otx-anomalib")
         self.config = self.get_config()
 
         # Set default model attributes.
--- a/otx/algorithms/anomaly/tasks/inference.py
+++ b/otx/algorithms/anomaly/tasks/inference.py
@@ -75,6 +75,7 @@ class InferenceTask(IInferenceTask, IEvaluationTask, IExportTask, IUnload):
 
         Args:
             task_environment (TaskEnvironment): OTX Task environment.
+            output_path (Optional[str]): output path where task output are saved.
         """
         torch.backends.cudnn.enabled = True
         logger.info("Initializing the task environment.")
--- a/otx/algorithms/common/tasks/training_base.py
+++ b/otx/algorithms/common/tasks/training_base.py
@@ -62,7 +62,8 @@ class BaseTask(IInferenceTask, IExportTask, IEvaluationTask, IUnload):
         self._model_name = task_environment.model_template.name
         self._task_type = task_environment.model_template.task_type
         self._labels = task_environment.get_labels(include_empty=False)
-        self._output_path = tempfile.mkdtemp(prefix="OTX-task-")
+        output_path = getattr(task_environment, "work_dir", None)
+        self._output_path = tempfile.mkdtemp(prefix="OTX-task-") if output_path is None else output_path
         logger.info(f"created output path at {self._output_path}")
         self.confidence_threshold = self._get_confidence_threshold(self._hyperparams)
         # Set default model attributes.
--- a/otx/algorithms/common/tasks/training_base.py
+++ b/otx/algorithms/common/tasks/training_base.py
@@ -62,9 +62,6 @@ class BaseTask(IInferenceTask, IExportTask, IEvaluationTask, IUnload):
         self._model_name = task_environment.model_template.name
         self._task_type = task_environment.model_template.task_type
         self._labels = task_environment.get_labels(include_empty=False)
-        output_path = getattr(task_environment, "work_dir", None)
-        self._output_path = tempfile.mkdtemp(prefix="OTX-task-") if output_path is None else output_path
-        logger.info(f"created output path at {self._output_path}")
         self.confidence_threshold = self._get_confidence_threshold(self._hyperparams)
         # Set default model attributes.
         self._model_label_schema = []  # type: List[LabelEntity]
@@ -72,6 +69,11 @@ class BaseTask(IInferenceTask, IExportTask, IEvaluationTask, IUnload):
         self._model_ckpt = None
         self._data_pipeline_path = None
         self._anchors = {}  # type: Dict[str, int]
+        output_path = os.environ.get("OTX_TASK_OUTPUT_PATH")
+        if output_path is None:
+            output_path = tempfile.mkdtemp(prefix="OTX-task-")
+        self._output_path = output_path
+        logger.info(f"created output path at {self._output_path}")
         if task_environment.model is not None:
             logger.info("loading the model from the task env.")
             state_dict = self._load_model_state_dict(self._task_environment.model)
@@ -99,6 +101,10 @@ class BaseTask(IInferenceTask, IExportTask, IEvaluationTask, IUnload):
 
         # to override configuration at runtime
         self.override_configs = {}  # type: Dict[str, str]
+
+    @property
+    def output_path(self):
+        return self._output_path
 
     def _run_task(self, stage_module, mode=None, dataset=None, **kwargs):
         # FIXME: Temporary remedy for CVS-88098
--- a/otx/algorithms/common/tasks/training_base.py
+++ b/otx/algorithms/common/tasks/training_base.py
@@ -55,7 +55,7 @@ class BaseTask(IInferenceTask, IExportTask, IEvaluationTask, IUnload):
     _task_environment: TaskEnvironment
 
     @check_input_parameters_type()
-    def __init__(self, task_config, task_environment: TaskEnvironment):
+    def __init__(self, task_config, task_environment: TaskEnvironment, output_path: Optional[str] = None):
         self._task_config = task_config
         self._task_environment = task_environment
         self._hyperparams = task_environment.get_hyper_parameters(self._task_config)  # type: ConfigDict
@@ -69,7 +69,6 @@ class BaseTask(IInferenceTask, IExportTask, IEvaluationTask, IUnload):
         self._model_ckpt = None
         self._data_pipeline_path = None
         self._anchors = {}  # type: Dict[str, int]
-        output_path = os.environ.get("OTX_TASK_OUTPUT_PATH")
         if output_path is None:
             output_path = tempfile.mkdtemp(prefix="OTX-task-")
         self._output_path = output_path
@@ -101,10 +100,6 @@ class BaseTask(IInferenceTask, IExportTask, IEvaluationTask, IUnload):
 
         # to override configuration at runtime
         self.override_configs = {}  # type: Dict[str, str]
-
-    @property
-    def output_path(self):
-        return self._output_path
 
     def _run_task(self, stage_module, mode=None, dataset=None, **kwargs):
         # FIXME: Temporary remedy for CVS-88098
--- a/otx/cli/tools/train.py
+++ b/otx/cli/tools/train.py
@@ -113,9 +113,9 @@ def parse_args():
         help="Location where trained model will be stored.",
     )
     parser.add_argument(
-        "--save-logs-to",
+        "--output-path",
         required=False,
-        help="Location where logs will be stored.",
+        help="Location where product of the training will be stored.",
     )
     parser.add_argument(
         "--enable-hpo",
@@ -206,9 +206,9 @@ def main():
         task = run_hpo(args, environment, dataset, template.task_type)
         if task is None:
             print("cannot run HPO for this task. will train a model without HPO.")
-            task = task_class(task_environment=environment, output_path=args.save_logs_to)
+            task = task_class(task_environment=environment, output_path=args.output_path)
     else:
-        task = task_class(task_environment=environment, output_path=args.save_logs_to)
+        task = task_class(task_environment=environment, output_path=args.output_path)
 
     if args.gpus:
         multigpu_manager = MultiGPUManager(main, args.gpus, str(args.multi_gpu_port))
--- a/otx/cli/tools/train.py
+++ b/otx/cli/tools/train.py
@@ -16,11 +16,11 @@
 
 import argparse
 import os
-import sys
 import signal
+import sys
 import threading
-from typing import List, Optional
 import time
+from typing import List, Optional
 
 import torch
 import torch.distributed as dist
@@ -221,10 +221,7 @@ def main():
     if args.gpus:
         multigpu_manager = MultiGPUManager(args.gpus, str(args.multi_gpu_port))
         if multigpu_manager.is_available(template):
-            multigpu_manager.setup_multi_gpu_train(
-                task.project_path,
-                hyper_parameters if args.enable_hpo else None
-            )
+            multigpu_manager.setup_multi_gpu_train(task.project_path, hyper_parameters if args.enable_hpo else None)
 
     output_model = ModelEntity(dataset, environment.get_model_configuration())
 
@@ -253,6 +250,7 @@ def main():
     if args.gpus:
         multigpu_manager.finalize()
 
+
 class MultiGPUManager:
     def __init__(self, gpu_ids: str, multi_gpu_port: str):
         self._gpu_ids = self._get_gpu_ids(gpu_ids)
@@ -263,7 +261,7 @@ class MultiGPUManager:
     def _get_gpu_ids(self, gpus: str) -> List[int]:
         num_available_gpu = torch.cuda.device_count()
         gpu_ids = []
-        for gpu_id in gpus.split(','):
+        for gpu_id in gpus.split(","):
             if not gpu_id.isnumeric():
                 raise RuntimeError("--gpus argument should be numbers separated by ','.")
             gpu_ids.append(int(gpu_id))
@@ -285,9 +283,7 @@ class MultiGPUManager:
         return len(self._gpu_ids) > 1 and not template.task_type.is_anomaly
 
     def setup_multi_gpu_train(
-        self,
-        output_path: str,
-        optimized_hyper_parameters: Optional[ConfigurableParameters] = None
+        self, output_path: str, optimized_hyper_parameters: Optional[ConfigurableParameters] = None
     ):
         if optimized_hyper_parameters is not None:
             self._set_optimized_hp_for_child_process(optimized_hyper_parameters)
@@ -309,19 +305,19 @@ class MultiGPUManager:
 
     @staticmethod
     def initialize_multigpu_train(rank: int, gpu_ids: List[int], multi_gpu_port: str):
-        os.environ['MASTER_ADDR'] = 'localhost'
-        os.environ['MASTER_PORT'] = multi_gpu_port
+        os.environ["MASTER_ADDR"] = "localhost"
+        os.environ["MASTER_PORT"] = multi_gpu_port
         torch.cuda.set_device(gpu_ids[rank])
-        dist.init_process_group(backend='nccl', world_size=len(gpu_ids), rank=rank)
-        print(f'dist info world_size = {dist.get_world_size()}, rank = {dist.get_rank()}')
+        dist.init_process_group(backend="nccl", world_size=len(gpu_ids), rank=rank)
+        print(f"dist info world_size = {dist.get_world_size()}, rank = {dist.get_rank()}")
 
     @staticmethod
     def run_child_process(rank: int, gpu_ids: List[int], output_path: str, multi_gpu_port: str):
-        gpus_arg_idx = sys.argv.index('--gpus')
+        gpus_arg_idx = sys.argv.index("--gpus")
         for _ in range(2):
             sys.argv.pop(gpus_arg_idx)
         if "--enable-hpo" in sys.argv:
-            sys.argv.remove('--enable-hpo')
+            sys.argv.remove("--enable-hpo")
         MultiGPUManager.set_arguments_to_argv("--save-logs-to", output_path)
 
         MultiGPUManager.initialize_multigpu_train(rank, gpu_ids, multi_gpu_port)
@@ -338,16 +334,15 @@ class MultiGPUManager:
                 sys.argv.insert(sys.argv.index("params"), value)
             else:
                 if after_params and "params" not in sys.argv:
-                    sys.argv.append('params')
+                    sys.argv.append("params")
                 sys.argv.extend([key, value])
 
     def _spawn_multi_gpu_processes(self, output_path: str) -> List[mp.Process]:
-        processes= []
+        processes = []
         spawned_mp = mp.get_context("spawn")
         for rank in range(1, len(self._gpu_ids)):
             task_p = spawned_mp.Process(
-                target=MultiGPUManager.run_child_process,
-                args=(rank, self._gpu_ids, output_path, self._multi_gpu_port)
+                target=MultiGPUManager.run_child_process, args=(rank, self._gpu_ids, output_path, self._multi_gpu_port)
             )
             task_p.start()
             processes.append(task_p)
@@ -379,14 +374,10 @@ class MultiGPUManager:
 
     def _set_optimized_hp_for_child_process(self, hyper_parameters: ConfigurableParameters):
         self.set_arguments_to_argv(
-            "--learning_parameters.learning_rate",
-            str(hyper_parameters.learning_parameters.learning_rate),
-            True
+            "--learning_parameters.learning_rate", str(hyper_parameters.learning_parameters.learning_rate), True
         )
         self.set_arguments_to_argv(
-            "--learning_parameters.batch_size",
-            str(hyper_parameters.learning_parameters.batch_size),
-            True
+            "--learning_parameters.batch_size", str(hyper_parameters.learning_parameters.batch_size), True
         )
 
     def _check_child_processes_alive(self):
--- a/otx/cli/tools/train.py
+++ b/otx/cli/tools/train.py
@@ -113,9 +113,9 @@ def parse_args():
         help="Location where trained model will be stored.",
     )
     parser.add_argument(
-        "--output-path",
+        "--work-dir",
         required=False,
-        help="Location where product of the training will be stored.",
+        help="Location where the intermediate output of the training will be stored.",
     )
     parser.add_argument(
         "--enable-hpo",
@@ -206,9 +206,9 @@ def main():
         task = run_hpo(args, environment, dataset, template.task_type)
         if task is None:
             print("cannot run HPO for this task. will train a model without HPO.")
-            task = task_class(task_environment=environment, output_path=args.output_path)
+            task = task_class(task_environment=environment, output_path=args.work_dir)
     else:
-        task = task_class(task_environment=environment, output_path=args.output_path)
+        task = task_class(task_environment=environment, output_path=args.work_dir)
 
     if args.gpus:
         multigpu_manager = MultiGPUManager(main, args.gpus, str(args.multi_gpu_port))
--- a/otx/cli/tools/train.py
+++ b/otx/cli/tools/train.py
@@ -17,6 +17,11 @@
 import argparse
 import os
 import shutil
+import os.path as osp
+
+import torch
+import torch.distributed as dist
+import torch.multiprocessing as mp
 
 from otx.api.configuration.helper import create
 from otx.api.entities.inference_parameters import InferenceParameters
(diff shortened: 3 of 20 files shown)


otx/algorithms/common/tasks/training_base.py
```diff
@@ -54,21 +54,23 @@ class BaseTask(IInferenceTask, IExportTask, IEvaluationTask, IUnload):
...
         self._labels = task_environment.get_labels(include_empty=False)
-        self._output_path = tempfile.mkdtemp(prefix="OTX-task-")
-        logger.info(f"created output path at {self._output_path}")
         self.confidence_threshold = self._get_confidence_threshold(self._hyperparams)
         # Set default model attributes.
         self._model_label_schema = []  # type: List[LabelEntity]
         self._optimization_methods = []  # type: List[OptimizationMethod]
         self._model_ckpt = None
         self._anchors = {}  # type: Dict[str, int]
+        if output_path is None:
```
jaegukhyun, line 72: Is Changing output path related with this PR?

otx/cli/tools/train.py
```diff
@@ -128,6 +138,18 @@ def parse_args():
         type=float,
         help="Expected ratio of total time to run HPO to time taken for full fine-tuning.",
     )
+    parser.add_argument(
```
jaegukhyun, line 132: Is there no default parameter for gpus?

otx/cli/tools/train.py
```diff
@@ -195,9 +216,14 @@ def main():
         task = run_hpo(args, environment, dataset, template.task_type)
         if task is None:
             print("cannot run HPO for this task. will train a model without HPO.")
-            task = task_class(task_environment=environment)
+            task = task_class(task_environment=environment, output_path=args.save_logs_to)
```
jaegukhyun: So, we have logs and checkpoints in the args.save_logs_to, is it right?

otx/cli/tools/train.py
```diff
@@ -222,11 +248,215 @@ def main():
...
-    if args.save_logs_to:
-        tmp_path = task.project_path
-        logs_path = os.path.join(args.save_logs_to, tmp_path.split("/")[-1])
-        shutil.copytree(tmp_path, logs_path)
-        print(f"Save logs: {logs_path}")
+    if args.gpus:
+        multigpu_manager.finalize()
+
+
+class MultiGPUManager:
```
jaegukhyun: I'm not sure this is right place for MultiGPUManager.
I prefer short codes for cli

otx/cli/tools/train.py
```diff
@@ -222,11 +248,215 @@ def main():
...
+        wrong_gpus = []
+        for gpu_idx in gpu_ids:
+            if gpu_idx >= num_available_gpu:
+                wrong_gpus.append(gpu_idx)
+
+        for wrong_gpu in wrong_gpus:
+            gpu_ids.remove(wrong_gpu)
+
+        if wrong_gpus:
+            print(f"Wrong gpu indeces are excluded. {','.join([str(val) for val in gpu_ids])} GPU will be used.")
```
jaegukhyun: How about using logger's warning?

otx/algorithms/common/tasks/training_base.py
eunwoosh: yes. All multi GPU processes should use same output_path. This modification is for that.

tests/integration/cli/classification/test_classification.py
```diff
@@ -105,6 +105,14 @@ def test_otx_train(self, template, tmp_dir_path):
         args1["--load-weights"] = f"{template_work_dir}/trained_{template.model_template_id}/weights.pth"
         otx_train_testing(template, tmp_dir_path, otx_dir, args1)
 
+    @e2e_pytest_component
```
jaegukhyun: Is it okay, setting gpus like this? If ci machine run multiple tests, then the ci machine could give a lot of burden for specific gpus

otx/cli/tools/train.py
eunwoosh: yes. it will be set only if user try to run multi gpu training.

otx/cli/tools/train.py
eunwoosh: yes. you're right.

otx/cli/tools/train.py
eunwoosh: but it's only used in train.py. Do you think that it's better to write this class in separate file?

tests/integration/cli/classification/test_classification.py
eunwoosh: you're right. but current CI doesn't run multiple job at the same time.

otx/cli/tools/train.py
eunwoosh: yeah... actually logger isn't used previously, so I used also print. is it better to use logger instead print? how do you think about?

jaegukhyun (changes requested):
Generally it looks good to me. Frankly of speaking I'm not expert of multiprocessing, so let's check the test results. I left some comment, actually most of them are questions. Also I have two questions for overall PR.
1. Does using multiprocessing give more benefits than using nn.parallel.DistributedDataParallel? I have read [this article](https://pytorch.org/docs/stable/notes/cuda.html#cuda-nn-ddp-instead)
2. Don't we need stressed test for multiprocessing? Multiprocessing works always looks good if they run small amount(training schedule, # of jobs), but when the volume of works grow, they give error out of sudden.

otx/cli/tools/train.py
jaegukhyun: Maybe I missed something, is there any argument checking use of multi gpu training? What happens when user don't give gpus argument(he wants single gpu training)

otx/cli/tools/train.py
jaegukhyun: If so, args.save_logs_to should change name such like args.output_path. In addition what if someone want to separate output log path and weight path, can this code handle it?

otx/cli/tools/train.py
jaegukhyun: I'm not sure, but otx/cli/utils looks fine

eunwoosh:
> 



> Generally it looks good to me. Frankly of speaking I'm not expert of multiprocessing, so let's check the test results. I left some comment, actually most of them are questions. Also I have two questions for overall PR.
> 
> 1. Does using multiprocessing give more benefits than using nn.parallel.DistributedDataParallel? I have read [this article](https://pytorch.org/docs/stable/notes/cuda.html#cuda-nn-ddp-instead)
> 2. Don't we need stressed test for multiprocessing? Multiprocessing works always looks good if they run small amount(training schedule, # of jobs), but when the volume of works grow, they give error out of sudden.

Thanks for comment! I think I can answer your questions.

Answer1 : You're right. So I implement to use  nn.parallel.DistributedDataParallel.
Answer 2 : Actually, I can't understand your exact point. So Do you mean that we need to make test for multiple multi GPU training?

otx/cli/tools/train.py
jaegukhyun: If this is warning, then how about add [Warning] flag to the stdout

otx/cli/tools/train.py
eunwoosh: then gpus argument is set as None, and single gpu training will start.

otx/cli/tools/train.py
eunwoosh: Although I just used already made argument `args.save_logs_to`, but I agree with your opinion. I'll change its' name to `args.output_path`.
For second opinion, do you mean that weight file is final model weight? then, user can set by using `--save-model-to` argument.

jaegukhyun:
> > 
> 
> > Generally it looks good to me. Frankly of speaking I'm not expert of multiprocessing, so let's check the test results. I left some comment, actually most of them are questions. Also I have two questions for overall PR.
> > 
> > 1. Does using multiprocessing give more benefits than using nn.parallel.DistributedDataParallel? I have read [this article](https://pytorch.org/docs/stable/notes/cuda.html#cuda-nn-ddp-instead)
> > 2. Don't we need stressed test for multiprocessing? Multiprocessing works always looks good if they run small amount(training schedule, # of jobs), but when the volume of works grow, they give error out of sudden.
> 
> Thanks for comment! I think I can answer your questions.
> 
> Answer1 : You're right. So I implement to use nn.parallel.DistributedDataParallel. Answer 2 : Actually, I can't understand your exact point. So Do you mean that we need to make test for multiple multi GPU training?

To the answer2, I'm not sure, but I think we should check whether multi-gpu training can run multiple training jobs
For example, run below training jobs simultaneously
 1. one training job for gpu1, 2
 2. one training job for gpu 1,2,3
 3. one training job for 2,3 
 4. one training job for 2,3,4
Did you check this situation in your environment? Test code can test simple situation, but we should check complex situation in local environment

otx/cli/tools/train.py
jaegukhyun: Then a user gives both save_logs_to and save-model-to, then are log and model weight saved in separate directory?

eunwoosh: Yes. but check point weight, which is saved by mmXX framework, is saved in `save_logs_to`, and final model weight saved by OTX will be in `--save-model-to`. So, the reason why there is `args.save_logs_to` is to make it available to save training history in specific path. If `args.save_logs_to` isn't set by user, then training history is saved in /tmp as previous.

otx/cli/tools/train.py
eunwoosh: I see. I'll modify code as you said.

eunwoosh:
> > > 
> > 
> > 
> > > Generally it looks good to me. Frankly of speaking I'm not expert of multiprocessing, so let's check the test results. I left some comment, actually most of them are questions. Also I have two questions for overall PR.
> > > 
> > > 1. Does using multiprocessing give more benefits than using nn.parallel.DistributedDataParallel? I have read [this article](https://pytorch.org/docs/stable/notes/cuda.html#cuda-nn-ddp-instead)
> > > 2. Don't we need stressed test for multiprocessing? Multiprocessing works always looks good if they run small amount(training schedule, # of jobs), but when the volume of works grow, they give error out of sudden.
> > 
> > 
> > Thanks for comment! I think I can answer your questions.
> > Answer1 : You're right. So I implement to use nn.parallel.DistributedDataParallel. Answer 2 : Actually, I can't understand your exact point. So Do you mean that we need to make test for multiple multi GPU training?
> 
> To the answer2, I'm not sure, but I think we should check whether multi-gpu training can run multiple training jobs For example, run below training jobs simultaneously
> 
> 1. one training job for gpu1, 2
> 2. one training job for gpu 1,2,3
> 3. one training job for 2,3
> 4. one training job for 2,3,4
>    Did you check this situation in your environment? Test code can test simple situation, but we should check complex situation in local environment

I understand. I'll check the situation in the local environment, and I'll update result.

eunwoosh:
@jaegukhyun I checked that 4 multi GPU training worked well parallelly.

otx/algorithms/anomaly/tasks/inference.py
```diff
@@ -70,11 +70,12 @@
 class InferenceTask(IInferenceTask, IEvaluationTask, IExportTask, IUnload):
     """Base Anomaly Task."""
 
-    def __init__(self, task_environment: TaskEnvironment) -> None:
+    def __init__(self, task_environment: TaskEnvironment, output_path: Optional[str] = None) -> None:
```
harimkang, line 73: Doesn't it matter if this task is done in common with other tasks? (nncf, etc.)
It seems strange that the init parameters are different for each task.

otx/cli/tools/train.py
```diff
@@ -113,9 +113,9 @@ def parse_args():
         help="Location where trained model will be stored.",
     )
     parser.add_argument(
-        "--save-logs-to",
+        "--output-path",
```
harimkang: Is this an argment that spits out both model weights and log at the same path here?

otx/cli/utils/hpo.py
```diff
@@ -136,7 +136,7 @@ def run_hpo(args: Namespace, environment: TaskEnvironment, dataset: DatasetEntit
     task_class = get_impl_class(environment.model_template.entrypoints.base)
     task_class = get_train_wrapper_task(task_class, task_type)
 
-    task = task_class(task_environment=environment)
+    task = task_class(task_environment=environment, output_path=args.save_logs_to)
```
harimkang: Why does this use `save_logs_to` as is?

otx/cli/tools/train.py
```diff
@@ -128,6 +128,18 @@ def parse_args():
...
         help="Expected ratio of total time to run HPO to time taken for full fine-tuning.",
     )
+    parser.add_argument(
+        "--gpus",
+        type=str,
+        help="Comma-separated indcies of GPU. \
+              If there are more than one available GPU, then model is trained with multi GPUs.",
+    )
+    parser.add_argument(
+        "--multi-gpu-port",
```
harimkang, line 139: Does the user need to set this?

harimkang (changes requested):
I left a few comments.

otx/cli/tools/train.py
eunwoosh: In mmXX family, yes. mmXX task saves model weight and log file while model is training.

otx/cli/utils/hpo.py
eunwoosh: Thanks to find it. I'll update it.

otx/cli/tools/train.py
eunwoosh: yes, if user already uses those port. and if user want to run multi multi-GPU training parallelly, then user should set different port to each training.

otx/cli/tools/train.py
harimkang: I'm sorry, but I didn't understand that it was a mmXX task. Do you mean the inference task of each task? And why should this be separated from `--model-weight-to`? 
I think `output-path` is a confusing word from the user's point of view.

otx/cli/tools/train.py
harimkang: Any guide on this? If the port is the same, then we have two processes on the same gpu, and if they are given differently, we can run them on different gpu?

And what does default mean? (I don't really know much about this part. I hope you understand.)

otx/algorithms/anomaly/tasks/inference.py
eunwoosh: In the perspective of chance to make some error, it's ok. because it's an optional argument. I think there won't any error.
In the perspective of design, I think that it's rather natural to have different arguments to different task.
For example, user may need to save some output in inference task or training task but doesn't need to save some output from training in nncf.
maybe I miss some point of view. If you have, then please let me know.

otx/cli/tools/train.py
eunwoosh: Maybe some guides need to be added. I'll add some guides.
multi processes in a same multi-GPU training need to use same port.
multi processes, each belonging to different multi-GPU training, should use different port.
default means literally default value of argument. So, if user doesn't set this argument but set `--gups` to run multi-GPU training, then 25000 port number will be used.

otx/cli/tools/train.py
eunwoosh: mmXX task means any task using mmcv based training. As f as I know, inference and train tasks of mmcls, mmdet and mmseg save the model weight and log file while task is working.
I agree that user can be confused. but `--model-weight-to` is used for path where OTX final model weight is saved and `--output-path` is used for mmXX output file. So, if user doesn't want to save an intermediate result of training, then user can save it in /tmp , by not setting `--output-path`.
But I think it's also good design choice to make an intermediate result from mmXX framework saved. how do you think about?

otx/algorithms/anomaly/tasks/inference.py
harimkang: As far as I know, pot or nncf also creates a log folder and new model weights that there is an output.
Also, doesn't the addition of this parameter mean to replace the existing `/tmp/`?

otx/cli/tools/train.py
harimkang: Yes I understand your explanation. But I still think `output-path` is not appropriate. Rather confusing.

eunwoosh: ok. then I'll modify it to `work-dir`.

otx/algorithms/anomaly/tasks/inference.py
eunwoosh: I see. I'll also add new argument to other tasks not updated yet.

otx/cli/tools/train.py
sungmanc: I also agree with Harim's opinion. Although there is a difference between OTX tasks and MM***, why should we divide individually? How about using same path for OTX tasks and MM*** tasks?

AFAIK, Users don't need to check the output of MM***.

otx/cli/tools/train.py
```diff
@@ -222,11 +238,8 @@ def main():
     assert resultset.performance is not None
     print(resultset.performance)
 
-    if args.save_logs_to:
-        tmp_path = task.project_path
-        logs_path = os.path.join(args.save_logs_to, tmp_path.split("/")[-1])
-        shutil.copytree(tmp_path, logs_path)
-        print(f"Save logs: {logs_path}")
+    if args.gpus:
```
sungmanc, line 248: Can't we move this line to the #217?

otx/cli/utils/multi_gpu.py
```diff
@@ -0,0 +1,245 @@
...
+    """
+
+    def __init__(self, train_func: Callable, gpu_ids: str, multi_gpu_port: str):
+        self._train_func = train_func
+        self._gpu_ids = get_gpu_ids(gpu_ids)
+        self._multi_gpu_port = multi_gpu_port
+        self._main_pid = os.getpid()
+        self._processes: Optional[List[mp.Process]] = None
+
+    def is_available(self, template: ModelTemplate) -> bool:
```
sungmanc: `is_available` function assumes that this function only counts the number of GPUs.
So, conditional return regarding task_type is misleading to me. What do you think about it?

Maybe we need additional function or `if` sentence to filter out the `anomaly task`.

otx/cli/utils/multi_gpu.py
```diff
@@ -0,0 +1,245 @@
...
+        Args:
+            rank (int): index of multi GPU processes.
+            gpu_ids (List[int]): list including which GPU indeces will be used.
+            multi_gpu_port (str): port for communication between multi GPU processes.
+        """
+        os.environ["MASTER_ADDR"] = "localhost"
+        os.environ["MASTER_PORT"] = multi_gpu_port
+        torch.cuda.set_device(gpu_ids[rank])
+        dist.init_process_group(backend="nccl", world_size=len(gpu_ids), rank=rank)
+        print(f"dist info world_size = {dist.get_world_size()}, rank = {dist.get_rank()}")
```
sungmanc: Using `Logger` instead of print

otx/cli/utils/multi_gpu.py
```diff
@@ -0,0 +1,245 @@
...
+
+        Args:
+            train_func (Callable): model training function.
+            rank (int): index of multi GPU processes.
+            gpu_ids (List[int]): list including which GPU indeces will be used.
+            output_path (str): output path where task output are saved.
+            multi_gpu_port (str): port for communication between multi GPU processes.
+        """
+        gpus_arg_idx = sys.argv.index("--gpus")
+        for _ in range(2):
```
sungmanc, line 173: is there any better way to change this line? I can't easily understand the meaning of 2.
Or adding a description? ``` for in range(2) # meaning of 2```

otx/cli/utils/multi_gpu.py
```diff
@@ -0,0 +1,245 @@
...
+
+        sys.exit(1)
+
+    def _kill_child_process(self):
+        if self._processes is None:
+            return
+
+        for process in self._processes:
+            if process.is_alive():
+                print(f"Kill child process {process.pid}")
```
sungmanc: Need to replace `print` with `logger`

sungmanc (changes requested):
Please replace `print` sentence with `logger` and also left some comments

otx/cli/utils/multi_gpu.py
eunwoosh: I see. I'll modify it.

otx/cli/utils/multi_gpu.py
eunwoosh: I see. I'll modify it.

otx/cli/tools/train.py
eunwoosh: It can't. main process also needs to execute remained job. So, child processes should be killed after all jobs are done.

otx/cli/utils/multi_gpu.py
eunwoosh: I agree with your opinion. I'll update it.

otx/cli/utils/multi_gpu.py
eunwoosh: It means that it removes both argument key and value.
I think I can write that code as bellow. what's your opinion about this? is it better?
```
sys.argv.pop(gpus_arg_idx+1)
sys.argv.pop(gpus_arg_idx)
```

otx/cli/tools/train.py
eunwoosh: Actually, I think it's out of scope to decide whether to combine them or not.
how about to discuss about it and make separate PR, if needed?

jaegukhyun (dismissed):

tests/integration/cli/classification/test_classification.py
```diff
@@ -105,6 +105,14 @@ def test_otx_train(self, template, tmp_dir_path):
...
+    @pytest.mark.skipif(TT_STABILITY_TESTS, reason="This is TT_STABILITY_TESTS")
+    @pytest.mark.parametrize("template", templates, ids=templates_ids)
+    def test_otx_multi_gpu_train(self, template, tmp_dir_path):
```
harimkang: We should skip this test in a non-multi-gpu (single gpu or cpu) environment.

eunwoosh: Thanks for comment. I'll update it.

otx/cli/utils/multi_gpu.py
sungmanc: > It means that it removes both argument key and value. I think I can write that code as bellow. what's your opinion about this? is it better?
> 
> ```
> sys.argv.pop(gpus_arg_idx+1)
> sys.argv.pop(gpus_arg_idx)
> ```

Ah, I see. let keep the current code :)

sungmanc (dismissed):

eunwoosh:
mutl gpu test result
this is multi gpu test result.

harimkang (dismissed):
LGTM,
As long as the test passes, I think we can merge.
(Currently, the multi-gpu test won't run in CI, right? - Looks like this needs to be checked.)
thank you for your hard work!

eunwoosh:
> LGTM, As long as the test passes, I think we can merge. (Currently, the multi-gpu test won't run in CI, right? - Looks like this needs to be checked.) thank you for your hard work!

Thank! Yes, current CI doesn't run multi GPU test. So I attached the result of them above.

eunwoosh:
There are two fail cases in Pre-Merge Check as bellow.
```
TestToolsMPAMultilabelClassification.test_otx_eval_openvino[Custom_Image_Classification_EfficientNet-V2-S]
AssertionError: trained_performance[k]=1.0, exported_performance[k]=0.9607843137254902
TestToolsMPAMultilabelClassification.test_otx_eval_openvino[Custom_Image_Classification_EfficientNet-B0]
AssertionError: trained_performance[k]=0.9803921568627451, exported_performance[k]=0.9607843137254902
```
These are due to gap between exported model performance and trained model performance.
I think that it's acceptable difference. So, could you approve the PR?

sungmanc (dismissed):

otx/cli/tools/train.py
```diff
@@ -128,6 +129,18 @@ def parse_args():
         type=float,
         help="Expected ratio of total time to run HPO to time taken for full fine-tuning.",
     )
+    parser.add_argument(
+        "--gpus",
+        type=str,
+        help="Comma-separated indices of GPU. \
+              If there are more than one available GPU, then model is trained with multi GPUs.",
```
JihwanEom, line 136: `If input 1+ GPUs, the model will be trained with multiple GPUs.`?

otx/cli/tools/train.py
```diff
@@ -128,6 +129,18 @@ def parse_args():
...
+    )
+    parser.add_argument(
+        "--multi-gpu-port",
+        default=25000,
+        type=int,
+        help="port for communication beteween multi GPU processes.",
```
JihwanEom, line 142: How about `port for multi GPU training`

otx/cli/tools/train.py
JihwanEom: Or... if you mean gpu indices, we need to consider `gpu_ids` instead of `gpus`.

otx/cli/utils/multi_gpu.py
```diff
@@ -0,0 +1,240 @@
+"""Multi GPU training utility."""
+
+# Copyright (C) 2021 Intel Corporation
```
JihwanEom: 2022?

otx/cli/utils/multi_gpu.py
```diff
@@ -0,0 +1,240 @@
...
+    wrong_gpus = []
+    for gpu_idx in gpu_ids:
+        if gpu_idx >= num_available_gpu:
+            wrong_gpus.append(gpu_idx)
+
+    for wrong_gpu in wrong_gpus:
+        gpu_ids.remove(wrong_gpu)
+
+    if wrong_gpus:
+        logger.warning(f"Wrong gpu indeces are excluded. {','.join([str(val) for val in gpu_ids])} GPU will be used.")
```
JihwanEom: `indeces` -> `indices`

otx/cli/utils/multi_gpu.py
```diff
@@ -0,0 +1,240 @@
...
+            else:
+                sys.argv.append(key)
+
+
+class MultiGPUManager:
+    """Class to manage multi GPU training.
+
+    Args:
+        train_func (Callable): model training function.
+        gpu_ids (str): GPU indices to use. Format should be Comma-separated indices.
```
JihwanEom, line 97: Collision with `gpus`?

otx/cli/utils/multi_gpu.py
```diff
@@ -0,0 +1,240 @@
...
+        multi_gpu_port (str): port for communication between multi GPU processes.
+    """
+
+    def __init__(self, train_func: Callable, gpu_ids: str, multi_gpu_port: str):
+        self._train_func = train_func
+        self._gpu_ids = get_gpu_ids(gpu_ids)
+        self._multi_gpu_port = multi_gpu_port
```
JihwanEom: `self._port` is also good for me.

otx/cli/utils/multi_gpu.py
```diff
@@ -0,0 +1,240 @@
...
+    def is_available(self) -> bool:
+        """Check multi GPU training is available.
+
+        Returns:
+            bool:
+                whether multi GPU training is available.
+        """
+        return len(self._gpu_ids) > 1
+
+    def setup_multi_gpu_train(
```
JihwanEom, line 117: Similar with above, `setup_train` is okay for me.

otx/cli/utils/multi_gpu.py
```diff
@@ -0,0 +1,240 @@
...
+        threading.Thread(target=self._check_child_processes_alive, daemon=True).start()
+
+    def finalize(self):
+        """Join all child processes."""
+        if self._processes is not None:
+            for p in self._processes:
+                p.join()
+
+    @staticmethod
+    def initialize_multigpu_train(rank: int, gpu_ids: List[int], multi_gpu_port: str):
```
JihwanEom, line 147: How about `initialize`?

tests/integration/cli/classification/test_classification.py
```diff
@@ -81,6 +82,7 @@ def tmp_dir_path():
         yield Path(tmp_dir)
 
 
+MULTI_GPU_UNAVAILABLE = torch.cuda.device_count() <= 1
```
JihwanEom, line 85: How about `MULTI_GPU_SUPPORTED` or `MULTI_GPU_SUPPORT`

tests/integration/cli/classification/test_classification.py
```diff
@@ -222,6 +224,15 @@ def test_pot_optimize(self, template, tmp_dir_path):
     def test_pot_eval(self, template, tmp_dir_path):
         pot_eval_testing(template, tmp_dir_path, otx_dir, args)
 
+    @e2e_pytest_component
+    @pytest.mark.skipif(TT_STABILITY_TESTS, reason="This is TT_STABILITY_TESTS")
+    @pytest.mark.skipif(MULTI_GPU_UNAVAILABLE, reason="The number of gpu is insufficient")
```
JihwanEom, line 229: How about `The number of gpu is scarce to run multi-gpu training.`

JihwanEom (commented):
Can we use also 3 GPUs or 4 GPUs?

otx/cli/utils/multi_gpu.py
JihwanEom: `wrong` -> `invalid`?

otx/cli/utils/multi_gpu.py
eunwoosh: I'll update it

otx/cli/utils/multi_gpu.py
eunwoosh: I'll update it

tests/integration/cli/classification/test_classification.py
eunwoosh: It means multi GPU isn't available. MULTI_GPU_SUPPORTED = True, when multi GPU isn't available is weird

tests/integration/cli/classification/test_classification.py
eunwoosh: I think this sentence is also ok

otx/cli/utils/multi_gpu.py
eunwoosh: I'll update it

otx/cli/utils/multi_gpu.py
eunwoosh: I think this function name can convey meaning better.

otx/cli/tools/train.py
eunwoosh: I think `gpus` also can convey meaning

otx/cli/tools/train.py
eunwoosh: This sentence can convey more detailed meaning

otx/cli/utils/multi_gpu.py
eunwoosh: I think this function name can convey meaning better.

otx/cli/utils/multi_gpu.py
eunwoosh: > Can we use also 3 GPUs or 4 GPUs?

yes

eunwoosh:
> Can we use also 3 GPUs or 4 GPUs?

yes

JihwanEom (approved):

sungmanc (approved):

PR #1398, opened 2022-11-28
[OTX/Semi-SL-Det] Enable training and evaluation for unbiased teacher structure

jaegukhyun:
This PR is for enabling training and evaluation for unbiased teacher structure.

'semisl' folder will be placed inside of each detection model config folder. 'semisl' folder would have template_semi.yaml, model.py, data_pipeline.py for semi-supervised learning.
Detection task determine selects its training algorithm below logic:
if template.TrainType == "SEMISUPERVISED" and 'unlabeled' in cfg.data_cfg:
    train_type = "SEMISUPERVISED"
else:
    train_type = "INCREMENTAL"
In addition detection task transfer its training algorithm through setting self._recipe_cfg.train_type


otx/algorithms/common/tasks/training_base.py
```diff
@@ -101,6 +101,7 @@ def _run_task(self, stage_module, mode=None, dataset=None, **kwargs):
         # FIXME: Temporary remedy for CVS-88098
         export = kwargs.get("export", False)
         self._initialize(export=export)
+        stage_module = self._update_stage_module(stage_module)
```
supersoob, line 105: If it is called for train, does it use the method inside inference? And I saw the code 'inference.py', that it uses always 'DetectionInferrer' when calling infer_detector. Was it meant to be? Could you explain the workflow of this stage_module?

jaegukhyun: _run_task function in the traininig_base.py is used both in inference and training task in detection, and both task do not modify this function. Inference and training task call _run_task function to link each task and mpa stage. Therefore this new _update_stage_module function change stage_module if the task is not normal incremental learning(e.g. semi-sl)

otx/algorithms/common/tasks/training_base.py
```diff
@@ -279,6 +280,9 @@ def _init_recipe_hparam(self) -> dict:
             runner=ConfigDict(max_epochs=int(params.num_iters)),
         )
 
+    def _update_stage_module(self, stage_module: str):
+        return stage_module
```
harimkang: what is this..? just return input?

otx/algorithms/common/tasks/training_base.py
supersoob: I really get why you call update_stage_module inside training_base. So this is like changer for trainer class or inference class for each train type(recipe). Thanks for explanation.

otx/algorithms/detection/tasks/inference.py
```diff
@@ -226,18 +226,22 @@ def _init_recipe(self):
...
-                train_type = TrainType.INCREMENTAL
+                recipe = os.path.join(recipe_root, "semisl.py")
 
         if train_type == TrainType.INCREMENTAL:
             recipe = os.path.join(recipe_root, "incremental.py")
 
         logger.info(f"train type = {train_type} - loading {recipe}")
 
         self._recipe_cfg = MPAConfig.fromfile(recipe)
+        self._recipe_cfg.train_type = train_type.name
```
supersoob: I'd like to suggest to change it just `self.train_type=train_type`. I guess we don't need to use or distinguish semisl or incr-l inside mpa trainer or inferencer since already differentiate them here.

otx/algorithms/detection/tasks/inference.py
```diff
@@ -267,6 +271,15 @@ def _init_test_data_cfg(self, dataset: DatasetEntity):
         )
         return data_cfg
 
+    def _update_stage_module(self, stage_module):
+        if hasattr(self._recipe_cfg, "train_type"):
```
supersoob: Same comments above. I think we don't need to check train_type inside recipe_cfg. Just checking it by if `self.train_type = TrainType.SEMISUPERVISED`

otx/algorithms/common/tasks/training_base.py
jaegukhyun: That's right!! Thanks

otx/algorithms/common/tasks/training_base.py
jaegukhyun: Yes, This is base function, if there are no certain stage change rule for each task(cls, det, seg) than it do not change it's stage

otx/algorithms/detection/tasks/inference.py
```diff
@@ -226,18 +226,22 @@ def _init_recipe(self):
...
             raise NotImplementedError(f"Train type {train_type} is not implemented yet.")
         if train_type == TrainType.SEMISUPERVISED:
-            if self._data_cfg.get("data", None) and self._data_cfg.data.get("unlabeled", None):
-                recipe = os.path.join(recipe_root, "semisl.py")
+            if self._is_training:
+                if self._data_cfg.get("data", None) and self._data_cfg.data.get("unlabeled", None):
+                    recipe = os.path.join(recipe_root, "semisl.py")
+                else:
+                    logger.warning("Cannot find unlabeled data.. convert to INCREMENTAL.")
+                    train_type = TrainType.INCREMENTAL
```
supersoob, line 279: I found some possible bug here. Even if we loaded it with template_semi.yaml, it can be load and run without passing unlabeled data. In this case, we need to convert to INCREMENTAL mode but since model is also hard-coded to semi (mean teacher / unbiased teacher), it would give error or run with (semi-model + incremental-mode). I dealt with this problem by changing model.py path double-checking the train type inside `_init_model_cfg`. Please refer to my [draft PR](https://github.com/openvinotoolkit/training_extensions/pull/1425/files#diff-38970adf7b9d362c1a674ca03fa4e639def08cbe36c4da5daaeea4f2f5bc6b54L212). If you have a better solution, let me know.

otx/algorithms/detection/tasks/inference.py
```diff
@@ -226,18 +226,22 @@ def _init_recipe(self):
...
             else:
-                logger.warning("Cannot find unlabeled data.. convert to INCREMENTAL.")
-                train_type = TrainType.INCREMENTAL
+                recipe = os.path.join(recipe_root, "semisl.py")
```
supersoob, line 281: This also can cause issue if the train was not actually run in supervised mode. For example, even if we set train type as SEMISUPERVISED but can be run through incremental mode by if-statement above(unlabeled data actually does not exist). In that case, the right way is that running the inference is in incremental mode as well. But I guess it would run anyway even though we load whatever because it's just a inference. We can leave it as is but hope to resolve wisely. If you got some solution, please share. This is not serious but it would be nice if we can match the train type load between train and inference.

otx/algorithms/detection/tasks/inference.py
goodsong81: More flexible way is to treat unlabeled data as optional input. No unlabeled? -> supervised loss only. Not mandatory at this point, but it will be more desirable.

otx/cli/registry/registry.py
```diff
@@ -26,15 +26,20 @@
...
             if templates_dir is None:
                 raise RuntimeError("The templates_dir is not set.")
 
-            template_filenames = glob.glob(os.path.join(templates_dir, "**", "template.yaml"), recursive=True)
+            if semisl:
+                template_filenames = glob.glob(
+                    os.path.join(templates_dir, "**", "template_semisl.yaml"), recursive=True
+                )
+            else:
+                template_filenames = glob.glob(os.path.join(templates_dir, "**", "template.yaml"), recursive=True)
```
goodsong81: experimental appends templates, but semisl is disjoint. Is it intended?

otx/algorithms/detection/tasks/nncf.py
```diff
@@ -246,6 +246,11 @@ def _create_model(config: Config, from_scratch: bool = False):
         """
         model_cfg = copy.deepcopy(config.model)
 
+        super_type = model_cfg.pop("super_type", None)
+        if super_type:
+            model_cfg.arch_type = model_cfg.type
+            model_cfg.type = super_type
```
goodsong81, line 252: It was my idea but feel free to change the mechanism.

otx/cli/registry/registry.py
```diff
@@ -26,15 +26,20 @@
 class Registry:
     """Class that implements a model templates registry."""
 
-    def __init__(self, templates_dir=None, templates=None, experimental=False):
+    def __init__(self, templates_dir=None, templates=None, experimental=False, semisl=False):
```
goodsong81: Might be generalized using the task type.

goodsong81 (commented):
Thank you for the PR. Looks good to me in general.
Giving some comments.

otx/algorithms/detection/tasks/inference.py
supersoob: Thanks for advices. We will make it run them by semi recipe but in a supervised learning way.

otx/algorithms/detection/tasks/inference.py
supersoob: This may be resolved by songki's suggestion above.

otx/algorithms/detection/tasks/inference.py
jaegukhyun: As we agreed on one template for one model, we can solve this problem. @supersoob Could you check the changes

otx/algorithms/detection/tasks/inference.py
jaegukhyun: Modified

jaegukhyun:
There are two major changes
1. Change template directory structure to one template for one model policy
2. Change test method using configurable training type

otx/algorithms/detection/configs/detection/mobilenetv2_ssd/semisl/data_pipeline.py
```diff
@@ -0,0 +1,183 @@
...
+# This is from mpa/recipes/stages/_base_/data/pipelines/ubt.py
+# This could be needed sync with incr-learning's data pipeline
+__img_scale = (864, 864)
+__img_norm_cfg = dict(mean=[0, 0, 0], std=[255, 255, 255], to_rgb=True)
+
+common_pipeline = [
+    dict(
+        type="Resize",
+        img_scale=__img_scale,
+        # img_scale=[
```
harimkang: Any reason to leave a comment?

otx/algorithms/detection/tasks/inference.py
```diff
@@ -226,28 +229,33 @@ def _init_recipe(self):
         if train_type not in (TrainType.SEMISUPERVISED, TrainType.INCREMENTAL):
```
harimkang: How about declaring it at the top of the file and using it?
`SUPPORTED_TRAIN_TASK = [TrainType.SEMISUPERVISED, TrainType.INCREMENTAL]`
&
`if train_type not in SUPPORTED_TRAIN_TASK:`

otx/algorithms/detection/tasks/inference.py
```diff
@@ -226,28 +229,33 @@ def _init_recipe(self):
...
         if train_type == TrainType.SEMISUPERVISED:
-            if self._data_cfg.get("data", None) and self._data_cfg.data.get("unlabeled", None):
-                recipe = os.path.join(recipe_root, "semisl.py")
+            if self._is_training:
+                if self._data_cfg.get("data", None) and self._data_cfg.data.get("unlabeled", None):
+                    recipe = os.path.join(recipe_root, "semisl.py")
+                    self.base_dir = os.path.join(self.template_dir, "semisl")
+                    self.data_pipeline_path = os.path.join(self.base_dir, "data_pipeline.py")
+                else:
+                    logger.warning("Cannot find unlabeled data.. convert to INCREMENTAL.")
```
harimkang, line 278: Shouldn't [SEMISUPERVISED + no unlabeled data] raise an error?

otx/algorithms/detection/tasks/inference.py
```diff
@@ -226,28 +229,33 @@ def _init_recipe(self):
         if train_type not in (TrainType.SEMISUPERVISED, TrainType.INCREMENTAL):
             raise NotImplementedError(f"Train type {train_type} is not implemented yet.")
         if train_type == TrainType.SEMISUPERVISED:
-            if self._data_cfg.get("data", None) and self._data_cfg.data.get("unlabeled", None):
-                recipe = os.path.join(recipe_root, "semisl.py")
+            if self._is_training:
+                if self._data_cfg.get("data", None) and self._data_cfg.data.get("unlabeled", None):
+                    recipe = os.path.join(recipe_root, "semisl.py")
+                    self.base_dir = os.path.join(self.template_dir, "semisl")
```
harimkang, line 275: This seems to be a problem with templates generated by `otx build`. Even if the folder created by otx build is semi, it will consist of the same file structure as supervised.

harimkang (changes requested):
It looks good. I left some concerned comments.

otx/algorithms/detection/configs/detection/mobilenetv2_ssd/semisl/data_pipeline.py
jaegukhyun: No, I'll remove it. Thank you for notice

otx/algorithms/detection/tasks/inference.py
jaegukhyun: It would be better, let me modify it. Thanks

otx/algorithms/detection/tasks/inference.py
jaegukhyun: Our current policy is when train type is semisl and there is no unlabeled data, we should train model through incremental settings. Do you have any concern about this policy?

harimkang: In fact, there is no problem to see, but if the user misses that log, the user may still think that SEMI is running well.. It was just such a small worry.

otx/algorithms/detection/tasks/inference.py
jaegukhyun: This will be resolved using builder class

harimkang: We put this as a temporary measure, and agreed to have a subdirectory structure even after the build. If anyone sees this and has ideas, please leave a comment.

harimkang (dismissed):
LGTM. Let's wait for the CI to be fixed and check the tests.

otx/algorithms/common/tasks/training_base.py
```diff
@@ -170,10 +172,16 @@ def template_file_path(self):
     @property
     def data_pipeline_path(self):
         """Base Data Pipeline file path."""
-        return os.path.join(
-            os.path.dirname(os.path.abspath(self.template_file_path)),
-            self._task_environment.model_template.data_pipeline_path,
-        )
+        if self._data_pipeline_path is None:
+            self._data_pipeline_path = os.path.join(
+                os.path.dirname(os.path.abspath(self.template_file_path)),
```
sungmanc: Don't we need to check `self.template_file_path` and `self._task_environment.model_template.data_pipeline_path`?

otx/algorithms/detection/tasks/inference.py
```diff
@@ -68,14 +68,19 @@
...
 
     @check_input_parameters_type()
     def __init__(self, task_environment: TaskEnvironment):
         # self._should_stop = False
+        self.train_type = None
         super().__init__(DetectionConfig, task_environment)
+        self.template_dir = os.path.abspath(os.path.dirname(self.template_file_path))
+        self.base_dir = self.template_dir
+        # TODO Move this to the common
+        self.supported_task = [TrainType.INCREMENTAL, TrainType.SEMISUPERVISED]
```
sungmanc: It would be better to use `self.supported_train_type`.

otx/algorithms/detection/tasks/inference.py
```diff
@@ -223,31 +228,36 @@ def _init_recipe(self):
         train_type = self._hyperparams.algo_backend.train_type
         logger.info(f"train type = {train_type}")
 
-        if train_type not in (TrainType.SEMISUPERVISED, TrainType.INCREMENTAL):
+        if train_type not in self.supported_task:
```
sungmanc: Same as above

sungmanc (commented):
I left some comments please check it

otx/algorithms/detection/tasks/inference.py
harimkang: https://github.com/openvinotoolkit/training_extensions/blob/9870328c036e6e0abce666a9779849ca0d0ab24e/otx/cli/utils/importing.py#L25

How about making it a global variable in this file like this?

tests/integration/cli/detection/test_detection.py
```diff
@@ -56,6 +56,26 @@
     "train_params": ["params", "--learning_parameters.num_iters", "2", "--learning_parameters.batch_size", "4"],
 }
 
+semisl_args = {
```
supersoob: I didn't define new args for test, just appended keys for semi in copied args dict.

supersoob (dismissed):
LGTM, thanks for the hard work. I left one command, but just suggestion.

supersoob (dismissed):

jaegukhyun:
https://github.com/openvinotoolkit/training_extensions/actions/runs/3754946842/jobs/6380632103
@harimkang, @supersoob, @sungmanc Tests are passed except 3 openvino eval tests. The failed cases are from performance difference after openvino export, and they are not related with this PR. I think we can finish this PR

sungmanc (approved):
LGTM

harimkang (approved):
LGTM!

PR #1400, opened 2022-11-29
Enable model training and NNCF in mmseg

cih9088:
- enable training and NNCF for mmseg
- refactor NNCF task for mmseg and mmdet
- add common utilities


otx/algorithms/common/adapters/mmcv/nncf/utils.py
```diff
@@ -0,0 +1,37 @@
...
+# SPDX-License-Identifier: Apache-2.0
+#
+
+import torch
+from mmcv.parallel import MMDataParallel, MMDistributedDataParallel
+
+from otx.algorithms.common.adapters.mmcv.data_cpu import MMDataCPU
+
+
+def prepare_model_for_execution(model, cfg, distributed=False):
```
goodsong81, line 11: Is this function for NNCF only? It looks rather general.

otx/algorithms/common/adapters/nncf/patches.py
```diff
@@ -25,24 +29,17 @@ def nncf_trace_context(self, img_metas, nncf_compress_postprocessing=True):
...
+        self.forward = partial(self.forward, img_metas=img_metas, return_loss=False)
     else:
         self.forward = partial(self.forward_dummy)
 
     yield
 
     # make everything normal
-    self.forward = forward_backup
-    torch.onnx.utils.__IN_ONNX_EXPORT = onnx_backup
+    self.__dict__.pop("forward")
```
goodsong81, line 41: I didn't know this trick. We don't need the forward_backup, then?

otx/algorithms/common/tasks/training_base.py
```diff
@@ -100,7 +100,8 @@ def __init__(self, task_config, task_environment: TaskEnvironment):
     def _run_task(self, stage_module, mode=None, dataset=None, **kwargs):
         # FIXME: Temporary remedy for CVS-88098
         export = kwargs.get("export", False)
-        self._initialize(export=export)
+        force = kwargs.get("force", False)
+        self._initialize(export=export, force=force)
```
goodsong81, line 104: Not clear what is being forced here.

otx/algorithms/common/tasks/training_base.py
```diff
@@ -230,6 +233,16 @@ def _initialize(self, export=False):
...
 
+        # if num_workers is 0, persistent_workers must be False
+        data_cfg = self._recipe_cfg.data
+        if data_cfg.get("workers_per_gpu", 0) == 0:
+            for subset in ["train", "val", "test"]:
+                dataloader_cfg = data_cfg.get(
+                    f"{subset}_dataloader", ConfigDict()
+                )
+                dataloader_cfg["persistent_workers"] = False
+                data_cfg[f"{subset}_dataloader"] = dataloader_cfg
```
goodsong81, line 244: @JihwanEom do you have any comment?

otx/algorithms/detection/adapters/mmdet/utils/config_utils.py
```diff
@@ -177,20 +190,55 @@ def patch_adaptive_repeat_dataset(
...
+        for key in list(subset_config.keys()):
+            found_config = get_configs_by_keys(
+                config.get(subset),
+                key,
+                return_path=True
+            )
+            assert len(found_config) == 1
+            value = subset_config.pop(key)
+            path = list(found_config.keys())[0]
+            update_config(subset_config, {path: value})
```
goodsong81, line 211: Could you shortly explain what this function is doing?

otx/algorithms/detection/init_venv.sh
```diff
@@ -108,6 +108,10 @@ echo torch=="${TORCH_VERSION}" >> "${CONSTRAINTS_FILE}"
 echo torchvision=="${TORCHVISION_VERSION}" >> "${CONSTRAINTS_FILE}"
 pip install torch=="${TORCH_VERSION}" torchvision=="${TORCHVISION_VERSION}" -f https://download.pytorch.org/whl/lts/1.8/torch_lts.html || exit 1
 
+# Install mmcv
+pip install --no-cache-dir mmcv-full==${MMCV_VERSION} || exit 1
+sed -i "s/force=False/force=True/g" "${venv_dir}"/lib/python"${PYTHON_VERSION}"/site-packages/mmcv/utils/registry.py  # Patch: remedy for MMCV registry collision from mmdet/mmseg
+
```
goodsong81, line 114: Is this still needed?

otx/algorithms/detection/tasks/nncf.py
```diff
@@ -248,46 +226,157 @@ def _create_model(config: Config, from_scratch: bool = False):
...
+        # merge initial model config
+        self._model_cfg.merge_from_dict(initial_model_cfg)
+
+        # Disable task adaptation in NNCF task
+        self._recipe_cfg.pop("task_adapt", None)
+        self._model_cfg.pop("task_adapt", None)
+        if hasattr(self._recipe_cfg.model, "is_task_adapt"):
+            self._recipe_cfg.model.is_task_adapt = False
+        if hasattr(self._model_cfg.model, "is_task_adapt"):
+            self._model_cfg.model.is_task_adapt = False
```
goodsong81, line 293: Could you let me know the reason?

otx/algorithms/detection/tasks/nncf.py
```diff
@@ -248,46 +226,157 @@ def _create_model(config: Config, from_scratch: bool = False):
...
+
+        patch_model_config(self._model_cfg, self._labels)
+
+        self._recipe_cfg.data.pop("super_type", None)
```
goodsong81, line 297: @jaegukhyun is this OK?

otx/algorithms/segmentation/configs/ocr_lite_hrnet_18/compression_config.json
```diff
@@ -22,10 +22,14 @@
...
     },
     "lr_config": {
+      "_delete_": true,
+      "policy": "customstep",
       "fixed": null,
       "fixed_iters": 0,
+      "fixed_ratio": 10.0,
       "warmup": null,
       "warmup_iters": 0,
+      "warmup_ratio": 0.3333,
```
goodsong81: Where were the updates from?

goodsong81 (commented):
Thank you @cih9088, quite a work!
I've left some comments & questions.

otx/algorithms/common/adapters/mmcv/nncf/utils.py
cih9088: Yes, it does but I didn't want to change other parts as they are being actively refactored.
I would move this function under mmcv not mmcv/nncf if you want.

otx/algorithms/common/adapters/nncf/patches.py
cih9088: Yes, we don't.
If one assigns something to an instance, internally, `__dict__` of an instance is updated.
Here, because of line 35, `__dict__` of `self` has `partial` function with a key of `forwrad`.
Since methods are non-data descriptors, which means no `__set__` method, updated `__dict__` has a precedence over a method. It means if one accesses `forward` attribute of `self`, it always gets to one defined in `__dict__`, not a method defined in a `class`.

If one pops `forward` from `__dict__`, everything will be normal.
This is indeed crucial when patching class methods (I mean when employing `mmdeploy`) because as I said, if `__dict__` have a key `forward`, even if we change class method `forward` by monkey patch, patched method will never be called.

I removed backup in https://github.com/openvinotoolkit/training_extensions/pull/1400/commits/f5177f6b55c24ea0d5f4cfeaea28b67c254e33ca for clarity.

otx/algorithms/detection/init_venv.sh
cih9088: I just wanted to align with other scripts, for example, [segmentation](https://github.com/cih9088/training_extensions/blob/feature/otx-new-mmcv/mmseg/otx/algorithms/segmentation/init_venv.sh#L111-L113), [classification](https://github.com/cih9088/training_extensions/blob/feature/otx-new-mmcv/mmseg/otx/algorithms/classification/init_venv.sh#L111-L113).
If this is not used, we should remove it from others as well.

otx/algorithms/detection/tasks/nncf.py
cih9088: This key [causes](https://github.com/openvinotoolkit/model_preparation_algorithm/blob/develop/mpa/det/stage.py#L58-L59) to run `configure_task` method in MPA which is not needed for NNCF as NNCF compressed model is initialized in OTX.

otx/algorithms/common/tasks/training_base.py
cih9088: [Here](https://github.com/cih9088/training_extensions/blob/feature/otx-new-mmcv/mmseg/otx/algorithms/common/tasks/training_base.py#L176-L177).
If `force` is `True`, it initializes the recipe again even if recipe has been initialized before.

otx/algorithms/detection/adapters/mmdet/utils/config_utils.py
cih9088: `self._data_cfg` is always initialized in [this](https://github.com/cih9088/training_extensions/blob/feature/otx-new-mmcv/mmseg/otx/algorithms/detection/tasks/nncf.py#L360-L373) structure. If data configuration of recipe has a wrapper dataset, such as `RepeatDataset`, `self._data_cfg` structure must be changed accordingly to [update recipe configuration](https://github.com/cih9088/training_extensions/blob/feature/otx-new-mmcv/mmseg/otx/algorithms/detection/adapters/mmdet/utils/config_utils.py#L228-L237).

otx/algorithms/segmentation/configs/ocr_lite_hrnet_18/compression_config.json
cih9088: Current OTX segmentation recipe does not use `CustomstepLrUpdateHook` but `ReduceLROnPlateauLrUpdaterHook`. The configuration presented in `compression_config.json` is `CustomstepLrUpdateHook` though. `fixed_ratio` and `warmup_ratio` parts were removed in https://github.com/openvinotoolkit/training_extensions/pull/1400/commits/12e6951e58924c1885357b0729ce975a9f6e0229 as they were just for test purpose.

otx/algorithms/detection/adapters/mmdet/utils/config_utils.py
cih9088: This would be refactored in `mmcls` PR.

otx/algorithms/common/adapters/mmcv/nncf/utils.py
goodsong81: We can do it later. Thanks.

otx/algorithms/common/adapters/nncf/patches.py
goodsong81: Nice! Learning a lot from you :)

otx/algorithms/detection/tasks/nncf.py
goodsong81: Understood. NNCF works on the resulting model from 'task_adapt', no need to do it again. Thanks.

otx/algorithms/detection/init_venv.sh
goodsong81: Let's remove it in other PR.

goodsong81 (approved):
LGTM! Let's merge to base and fix issues, if any.

PR #1401, opened 2022-11-29
[Test] Change ote threshold of openvino test for cls

jaegukhyun:
This PR includes changing threshold for classification openvino export test.
Unlike other methods(det, seg), classification threshold for openvino export test is set as zero.
In this PR, the threshold will be changed to 0.2, same as detection


goodsong81 (approved):

sungmanc (approved):
LGTM

PR #1406, opened 2022-12-02
[feature/otx]: Supervised Contrastive Loss for Multiclass classification

arrufat:
This PR contains add the possibility to train a classifier using a custom model with a dual head:
- Classification head: standard
- SelfSL head: currently it uses the BarlowTwins loss, but could be used with other Self-SL losses.

Training in this way provides significant gains in the low data regime (10 samples per class).
In my experiments, the classification error is reduced by about 5% on average.

PS: This is my first PR, so I expect a lot of feedback.


train_cls.py
```diff
@@ -0,0 +1,249 @@
+import importlib
```
sovrasov: This file should be deleted prior to merging.

otx/algorithms/classification/configs/efficientnet_v2_s_cls_incr/model_selfsl.py
```diff
@@ -0,0 +1,26 @@
+# _base_ = [ "../../configs/base/models/efficientnet_v2.py" ]
```
sovrasov: Dead code

otx/algorithms/classification/configs/efficientnet_b0_cls_incr/template_selfsl.yaml
```diff
@@ -0,0 +1,62 @@
+# Description.
```
sovrasov: `train_type` option could be specified via otx CLI, so we just don't need one more copy of template with a different `train_type`.

Overall, training paradigm choice should be done via the `train_type` parameter (supervised, semu-sl, self-sl). Also, we have a subdivision into multiclass, multilabel and hierarchical methods. It's done in runtime and based on the input label schema (and it can't be specified via template in advance). This scheme is not perfect, but at least for now we try to keep only one template per classification model and also this is a requirement from the geti's architecture side.

otx/algorithms/classification/tasks/inference.py
```diff
@@ -302,6 +302,9 @@ def _init_recipe(self):
 
         if train_type == TrainType.INCREMENTAL:
             recipe = os.path.join(recipe_root, "incremental.yaml")
+        if train_type == TrainType.SELFSUPERVISED:
+            recipe = self.template_file_path.replace(os.path.basename(self.template_file_path), "class_selfsl.yaml")
```
sovrasov: This looks like a hack. The selfsl recipe should be located in the same folder as the other ones (inside the MPA repo).

otx/algorithms/classification/tasks/inference.py
```diff
@@ -325,6 +328,9 @@ def _init_model_cfg(self):
             cfg_path = os.path.join(base_dir, "model_hierarchical.py")
         else:
             cfg_path = os.path.join(base_dir, "model.py")
+        # Override the model configuration if we are in SELFSUPERVISED mode
+        if self._hyperparams.algo_backend.train_type == TrainType.SELFSUPERVISED:
```
sovrasov: This if could be the outer one:
```
if self._hyperparams.algo_backend.train_type == TrainType.SELFSUPERVISED:
   ...
else:
   base_dir = os.path.abspath(os.path.dirname(self.template_file_path))
   ...
```
Such structure is a bit more clear to extend in the future.

otx/algorithms/classification/adapters/mmcls/model/heads/selfsl_cls_head.py
```diff
@@ -0,0 +1,101 @@
+# Copyright (C) 2022 Intel Corporation
```
sovrasov: Right now we keep all the heads, losses, etc in the MPA repo. In the future this repo will be moved to otx anyway, but for now I'd keep all the custom modules together in one place.

sovrasov:
Overall, for me the self-sl workflow is not clear. Normally we launch training via otx train. This tool loads the template, creates a task object, dataset, then launches a training. What is the workflow for self-sl classification? I assume, this scenario consumes unlabeled data and then train() method of the classification task produces a feature extractor, saves it's weights.
After that we can launch a supervised training via otx train, and finetune the obtained pre-trained weights for multiclass / multilabel / hierarchical classification on a labelled dataset.
@arrufat as fat as I see, you initially intended to implement a different workflow, may be I didn't fully understand the requirements to the self-sl method, could you please share your vision?

jaegukhyun (changes requested):
I think this PR should be checked by pre-commit hook.
I recommend to use command 'tox -re pre-commit' to check pre-commit issues

arrufat:
> Overall, for me the self-sl workflow is not clear. Normally we launch training via otx train. This tool loads the template, creates a task object, dataset, then launches a training. What is the workflow for self-sl classification? I assume, this scenario consumes unlabeled data and then train() method of the classification task produces a feature extractor, saves it's weights. After that we can launch a supervised training via otx train, and finetune the obtained pre-trained weights for multiclass / multilabel / hierarchical classification on a labelled dataset. @arrufat as fat as I see, you initially intended to implement a different workflow, may be I didn't fully understand the requirements to the self-sl method, could you please share your vision?

Thank you for the feedback, I will address the issues you pointed out soon.

I think I was not clear about the motivation of this PR.
The goal is to improve classification accuracy with resource constraints: few samples per class and few epochs.

The main idea is drawn from the [Supervised Contrastive Learning Paper](https://arxiv.org/abs/2004.11362).
In that paper, they follow the scheme that you described:
- train the model using Supervised contrastive learning (contrastive loss with labels)
- save the weights.
- linear probe the model.

We decided that, if we could combine all of this in one step, we could use the SupCon loss as a guidance for the Cross-Entropy loss. That could lead to faster convergence and higher accuracy in this scenario:
- train a network with 2 heads
  - a head with classification loss (using the labels)
  - a head with SupCon loss (also using the labels)
- after training, discard the head with the SupCon loss

In my experiments, I tried the SupCon approach, and also experimented with different SSL methods. Replacing the SupCon loss with the [Barlow Twins loss](https://arxiv.org/abs/2103.03230) lead to the most significant and consistent gains across datasets and configurations. The thing is, due to the way the Barlow Twins loss works, we're not using the labels for that head.

I agree that the name of the method (SelfSL Classifier) might not be the best I could've come up with, but that's what it is: a combination of a classification + SelfSL method. I am open to suggestions, and maybe I should not use the `TrainType.SELFSUPERVISED` for this task, either.

sovrasov:
Indeed, `TrainType.SELFSUPERVISED` is not the right option for this method. So it looks like we could integrate it into the existing supervised learning recipe then. I'd propose making an additional parameter in classification configurable parameters (see training_extensions/external/model-preparation-algorithm/mpa_tasks/apis/classification/config.py + configuration.yaml). It will be visible in the geti UI, but not from the first glance, and even not from the second... At the same time, I guess we can't just enable this method by default. Introducing dataset size threshold parameter is also an option (SupCon is always on if we have less than x data).

kprokofi:
I agree with @sovrasov 
We don't need to add a new training type. Self-supervised it is something different that we intend to add as a feature in the next Q1-Q2
We can extend existent incremental learning training type.
We may add necessary functionality like losses, heads into MPA and add additional parameters in yaml config for multiclass classification. It will be good to introduce switching on / off rule for this method too.

otx/algorithms/classification/configs/efficientnet_b0_cls_incr/template_selfsl.yaml
arrufat: I couldn't find how to specify the `train_type` in the docs. Would you mind showing me how to do it?

sovrasov: All the hyperparameters in template are exported to command line. The mechanism is described here:  https://github.com/openvinotoolkit/training_extensions/blob/develop/QUICK_START_GUIDE.md#model-template-specific-parameters 
But I just found out that this parameter is marked as non editable, also it has a wrong description. To adjust it via CLI we have to make some modifications in configuration.yaml, but it's a subject of discussion. Sorry for confusion, for now you can add additional templates.

arrufat: Thank you :)

goodsong81 (changes requested):
Thanks for the new feature.
But we don't think this is a new train type, rather it's an optional accuracy improvement for multi-class classification. In experimental results, it seems it could even replace default supervised one.
So I'd like to suggest that integrate this feature as 'optional enablement' via configurable parameters. You can refer to `enable_tilng` h-param in https://github.com/openvinotoolkit/training_extensions/blob/46b831bcdb7db8e3dc08e819c37969de55c53b4d/external/mmdetection/detection_tasks/apis/detection/configuration.py#L183

otx/algorithms/classification/utils/data.py
```diff
@@ -133,11 +133,11 @@ def __init__(
     @staticmethod
     def _load_text_annotation(annot_path, data_dir):
```
jaegukhyun: Could you explain why this function should be changed?

otx/algorithms/classification/configs/configuration.yaml
```diff
@@ -305,6 +305,7 @@ algo_backend:
     header: train type
     options:
       INCREMENTAL: "INCREMENTAL"
+      SUPERVISEDCONTRASTIVE: "SUPERVISEDCONTRASTIVE"
```
jaegukhyun: As Songki said, we would like to recommend to add supcon option in configuration.yaml like tiling method in detection

otx/algorithms/classification/utils/data.py
arrufat: Ah, I can revert this change. This allows loading multiclass datasets from JSON files, and not only from a directory tree. It was useful to subsample the datasets in my experiments. I will revert it, sorry.

jaegukhyun (changes requested):
I left two commend. First is same with Songki's request. Second is about the reason why loading dataset part is changed

arrufat:
@goodsong81 @jaegukhyun how does the last change look like?

jaegukhyun:
@arrufat Let me take a look, Thank you for your effort!!

otx/algorithms/classification/configs/efficientnet_b0_cls_incr/template_supcon.yaml
```diff
@@ -0,0 +1,64 @@
+# Description.
```
jaegukhyun: This is same with template.yaml. OTX TF have agreed that one template.yaml per one model. You can generate templates for your experiments, but that templates can be pushed.

otx/algorithms/classification/configs/efficientnet_v2_s_cls_incr/template_supcon.yaml
```diff
@@ -0,0 +1,64 @@
+# Description.
```
jaegukhyun: As I mentioned, you should remove template_*.yaml

otx/algorithms/classification/configs/mobilenet_v3_large_075_cls_incr/template_supcon.yaml
```diff
@@ -0,0 +1,55 @@
+# Description.
```
jaegukhyun: Same above

otx/algorithms/classification/configs/efficientnet_v2_s_cls_incr/template_supcon.yaml
arrufat: Ah, that parameter can be set via the command line?

otx/algorithms/classification/configs/mobilenet_v3_large_1_cls_incr/template_supcon.yaml
```diff
@@ -0,0 +1,64 @@
+# Description.
```
jaegukhyun: Same above

otx/algorithms/classification/configs/efficientnet_b0_cls_incr/model_supcon.py
```diff
@@ -0,0 +1,26 @@
+"""EfficientNet-B0 for multi-class config."""
```
jaegukhyun: This is optional, how about "EfficientNet-B0 config for multi-class with contrastive loss for smaller dataset"

arrufat: Yes, that sounds better, I will update the descriptions of all models :)

otx/algorithms/classification/tasks/inference.py
```diff
@@ -319,14 +322,19 @@ def _init_recipe(self):
     # error log : ValueError: Unexpected type of 'data_loader' parameter
     def _init_model_cfg(self):
         base_dir = os.path.abspath(os.path.dirname(self.template_file_path))
+
+        model_base_name = "model"
+        if self._hyperparams.learning_parameters.enable_supcon:
+            model_base_name = "model_supcon"
+
         if self._multilabel:
```
jaegukhyun, line 330: Is there model_supcon_multilabel.py in template directory?

otx/algorithms/classification/tasks/inference.py
```diff
@@ -319,14 +322,19 @@ def _init_recipe(self):
...
-            cfg_path = os.path.join(base_dir, "model_multilabel.py")
+            cfg_path = os.path.join(base_dir, model_base_name + "_multilabel.py")
```
jaegukhyun: Is there model_supcon_multilabel.py in template dir?

otx/algorithms/classification/tasks/inference.py
```diff
@@ -303,6 +303,9 @@ def _init_recipe(self):
         if train_type == TrainType.INCREMENTAL:
             recipe = os.path.join(recipe_root, "incremental.yaml")
 
+        if self._hyperparams.learning_parameters.enable_supcon:
```
jaegukhyun: If supcon only supports incremental learning(fine-tuning), this blocks should be inside if train_type == TrainType.INCREMENTAL. Because semi-sl and self-sl will be joined

jaegukhyun (changes requested):
I left some comments, I think we are so close to add this feature!!

otx/algorithms/classification/tasks/inference.py
arrufat: There will be soon, once I push the changes in the MPA.
I think this PR should be on hold until the MPA changes are merged.
Then I will make another PR in the MPA adding support for multilabel and hierarchical classification.

Or, if you prefer, I can remove this and make another PR here later.

otx/algorithms/classification/tasks/inference.py
arrufat: Right...

otx/algorithms/classification/tasks/inference.py
jaegukhyun: Yes, let's make another PR for expand SupCon to multilabel and hierarchical classification

jaegukhyun (dismissed):
@sovrasov @goodsong81 It looks good to me. I want you to review this PR once.
@arrufat Is there any plan for test SupCon feature?

arrufat:
> @sovrasov @goodsong81 It looks good to me. I want you to review this PR once. @arrufat Is there any plan for test SupCon feature?

Yes, I need to add integration tests, I just haven't got there yet.

otx/algorithms/classification/tasks/inference.py
```diff
@@ -319,14 +321,19 @@ def _init_recipe(self):
...
+            model_base_name = "model_supcon"
+
         if self._multilabel:
             cfg_path = os.path.join(base_dir, "model_multilabel.py")
         elif self._hierarchical:
             cfg_path = os.path.join(base_dir, "model_hierarchical.py")
         else:
-            cfg_path = os.path.join(base_dir, "model.py")
-        cfg = MPAConfig.fromfile(cfg_path)
+            cfg_path = os.path.join(base_dir, model_base_name + ".py")
```
sungmanc: How about using one additional `elif` sentence for `supcon` to unify the codes?

AS IS
```
if self._hyperparams.learning_parameters.enable_supcon:
            model_base_name = "model_supcon"

        if self._multilabel:
            cfg_path = os.path.join(base_dir, "model_multilabel.py")
        elif self._hierarchical:
            cfg_path = os.path.join(base_dir, "model_hierarchical.py")
        else:
            cfg_path = os.path.join(base_dir, model_base_name + ".py")
``` 

TO BE
```
        if self._multilabel:
            cfg_path = os.path.join(base_dir, "model_multilabel.py")
        elif self._hierarchical:
            cfg_path = os.path.join(base_dir, "model_hierarchical.py")
        elif self._hyperparameters.learning_parameters.enable_supcon:
            cfg_path = os.path.join(base_dir, "model_supcon.py")
        else:
            cfg_path = os.path.join(base_dir, "model.py")
```

sungmanc: And actually, I want to abbreviate `self._hyperparameters.learning_parameters.enable_supcon` --> `self._supcon` by adding some codes.

otx/algorithms/common/tasks/training_base.py
```diff
@@ -108,6 +108,7 @@ def _run_task(self, stage_module, mode=None, dataset=None, **kwargs):
         if dataset is not None:
             train_data_cfg = Stage.get_train_data_cfg(self._data_cfg)
             train_data_cfg["data_classes"] = data_classes
+            train_data_cfg["classes"] = data_classes
```
sungmanc: Why we need `train_data_cfg["classes"]` ? can't we use existing one(`train_data_cfg["data_classes"]`)?

sungmanc (commented):
I left some comments, please check them.

otx/algorithms/common/tasks/training_base.py
arrufat: When I modified the data loading part to be able to load images from a JSON file for the multiclass classification case (https://github.com/openvinotoolkit/training_extensions/pull/1406/commits/cf929c3ab77d3791863337e34494a4012f1e8b97), somehow, it complained that it wanted to see the classes using the `classes`, so I just added that.

I need to revert those two changes, since that was only made for me to be able to easily subsample classification datasets without haveing to duplicate the tree structure several times.

otx/algorithms/classification/tasks/inference.py
arrufat: I wrote it like this because the `enable_supcon` toggle is not at the same level (or exclusive) with the other `TrainType`s, but complimentary. So, it seems misleading to me.

Moreover, in an upcoming PR, there will be `model_supcon_multilabel` and `model_supcon_hierarchical`.

sungmanc: > I wrote it like this because the `enable_supcon` toggle is not at the same level (or exclusive) with the other `TrainType`s, but complimentary. So, it seems misleading to me.
> 
> Moreover, in an upcoming PR, there will be `model_supcon_multilabel` and `model_supcon_hierarchical`.

If so, the name of `model_supcon.py` is weird to me because the goal of `model_****.py`  is to divide the `TrainType`.
Similarly, `model_supcon_multilabel` and `model_supcon_hierarchical` are also weird. As you said, although `supcon` represents the complementary method and `multilabel` represents the `TrainType`, they'll be located the same level.

sungmanc: However, actually, in my case, I don't have ideas for a better way, now. We need to think about more

arrufat: In that case, we could remove the underscore: `modelsupcon_****.py`.
Then we would be more explicit about the fact that it's a different model (at least for training) and keep the underscore as a field separator for the `TrainType`.

arrufat:
Updated to latest `feature/otx`.

otx/algorithms/classification/tasks/inference.py
```diff
@@ -352,11 +352,18 @@ def _init_recipe(self):
 
         if train_type == TrainType.INCREMENTAL:
             recipe = os.path.join(recipe_root, "incremental.yaml")
+            if self._hyperparams.learning_parameters.enable_supcon:
+                recipe = os.path.join(recipe_root, "supcon.yaml")
```
jaegukhyun, line 360: We don't have supcon.yaml. If you want to change hyperparameters for supcon, you should change template.yaml manually before training experiments.

jaegukhyun (changes requested):
I add a comment which is related with template.yaml name for supcon. Basically we use one template for one model policy. Therefore if you want to use supcon for your experiments, then you should change hyper parameters in template.yaml manually. Let's discuss about automatic hyper-parameter changes for SupCon later.
And one more thing, I want you to check pre-commit error in github action.

otx/algorithms/classification/tasks/inference.py
arrufat: But we do have `supcon.yaml` already in `feature/otx`:
https://github.com/openvinotoolkit/training_extensions/blob/feature/otx/otx/recipes/stages/classification/supcon.yaml

jaegukhyun: Ah, sorry about that. I confused with template and recipe.

arrufat: No worries, it's very confusing.

sovrasov:
> I add a comment which is related with template.yaml name for supcon. Basically we use one template for one model policy. Therefore if you want to use supcon for your experiments, then you should change hyper parameters in template.yaml manually. Let's discuss about automatic hyper-parameter changes for SupCon later. And one more thing, I want you to check pre-commit error in github action.

Since the new supcon parameter in configuration is editable, it should be accessible via OTX CLI and users can modify it without editing template.yaml. Overall, the changes look good to me.

jaegukhyun (dismissed):
I think we can merge this PR, after solving conflicts
@sungmanc @harimkang @supersoob @JihwanEom Can anyone review this PR once more?

arrufat:
@jaegukhyun conflicts resolved :)

jaegukhyun:
There are pylint issues, could you check it?
pylint...................................................................Failed
- hook id: pylint
- exit code: 4

************* Module otx.algorithms.classification.tasks.inference
otx/algorithms/classification/tasks/inference.py:356:16: W0201: Attribute 'task_pipeline_path' defined outside __init__ (attribute-defined-outside-init)
otx/algorithms/classification/tasks/inference.py:363:12: W0201: Attribute 'task_pipeline_path' defined outside __init__ (attribute-defined-outside-init)
otx/algorithms/classification/tasks/inference.py:362:12: W0201: Attribute 'task_model_dir' defined outside __init__ (attribute-defined-outside-init)

otx/algorithms/classification/tasks/inference.py
```diff
@@ -351,11 +351,18 @@ def _init_recipe(self):
...
+                self.task_pipeline_path = os.path.join(
```
jaegukhyun: We agreed on variable name for pipeline and model.
It should be self.model_dir, and self.pipeline_path. Please refer segmentation parts. Therefore I want you to change these variable name

jaegukhyun (commented):
I left one comment

otx/algorithms/classification/tasks/inference.py
arrufat: I am not sure I follow, I did it like in the selfsl case.

arrufat: I didn't update the `self.model_dir` to contain `supcon` because it won't work when the task is `multilabel` or `hierarchical`. In my next PR, which adds support for those two tasks, I will do it.

jaegukhyun: Does selfsl use self.model_dir and pipeline_path?
            self.model_dir = os.path.join(self.model_dir, "selfsl")
            pipeline_path = os.path.join(os.path.dirname(pipeline_path), "selfsl/data_pipeline.py")

arrufat: yes

jaegukhyun: Then why supcon generate new variable task_model_dir and task_pipeline_path?

arrufat: I don't understand your question. Supcon needs its own model and data pipeline.

jaegukhyun: My point is why we should change variable name from self.model_dir to self.task_model_dir and from self.pipeline_path to self.task_pipeline_path.
Current feature/otx branch already merged Sunchul's cls-selfsl PR and he agree the variable names self.model_dir and self.pipeline_path

arrufat: Ah, I see it now, it was changed after I merged the first time, I will fix that. I didn't make this names up, they were carried out from a previous version. Sorry for the confusion.

jaegukhyun: No problem!

jaegukhyun (dismissed):

supersoob (dismissed):

goodsong81 (dismissed):
LGTM.

tests/integration/cli/classification/test_classification.py
```diff
@@ -102,6 +102,14 @@ def test_otx_train(self, template, tmp_dir_path):
         args1["--load-weights"] = f"{template_work_dir}/trained_{template.model_template_id}/weights.pth"
         otx_train_testing(template, tmp_dir_path, otx_dir, args1)
 
+    @e2e_pytest_component
```
jaegukhyun, line 105: I think it's better to test this before test_otx_train because below eval, export test expect the weight comes from normal incremental learning

arrufat: Ok, got it.

jaegukhyun (approved):
LGTM, Thank you for your effort, if you attach your local test, we will check the results.
I think classification test is enough

arrufat:
Test summary without this PR:
image

Test summary after this PR:
image

supersoob (approved):

harimkang (approved):
Nice work! LGTM

PR #1412, opened 2022-12-05
[Bug-fix] Convert NaN to num to make visible in geti UI

supersoob:
- Convert NaN to num to make visible in geti UI
- https://jira.devtools.intel.com/browse/CVS-93509
- This was caused because loss and sharpness were NaN value, so made xs and ys of learning_curve nan_to_num like detection task.


supersoob:
run ote-test

goodsong81 (approved):
LGTM.

sovrasov (approved):

sovrasov:
Perhaps we don't need this PR if we would keep back-merging strategy from release branch to develop (se we can propagate #1413 to develop later).

PR #1413, opened 2022-12-05
[release/0.4.0][Bug-fix] Convert NaN to num to make visible in geti UI

supersoob:
- Convert NaN to num to make visible in geti UI
- https://jira.devtools.intel.com/browse/CVS-93509
- This was caused because loss and sharpness were NaN value, so made xs and ys of learning_curve nan_to_num like detection task.


goodsong81 (approved):
LGTM.

sovrasov (approved):

PR #1402, opened 2022-11-29
Class-wise saliency map generation for the detection task

negvet:
Enable class-wise saliency map generation for the detection case.

Tested with all detection models from `external\model-preparation-algorithm\configs\detection\` on PASCAL VOC dataset

related PR:
https://github.com/openvinotoolkit/model_preparation_algorithm/pull/97

Some examples (mobilenetv2_ssd):
image

PR #1408, opened 2022-12-05
[OTE][Develop][XAI] Detection fix two stage bbox_head error

dongkwan-kim01:
This is the fix of the error issued in [issue](https://github.com/openvinotoolkit/training_extensions/pull/1402#issuecomment-1336631895), which fixes the `simple_test` function in the class`SAMDetectorMixin`.
And resolving the tiling detection length Error regarding the output size assertion.

PR #1414, opened 2022-12-05
[OTE][Release][XAI] Detection fix two stage bbox_head error

dongkwan-kim01:
This is the fix of the error issued in [issue](https://github.com/openvinotoolkit/training_extensions/pull/1402#issuecomment-1336631895), which fixes the `simple_test` function in the class`SAMDetectorMixin`.
And resolving the tiling detection length Error regarding the output size assertion.

### DIFF
--- a/external/model-preparation-algorithm/mpa_tasks/apis/detection/task.py
+++ b/external/model-preparation-algorithm/mpa_tasks/apis/detection/task.py
@@ -138,9 +138,9 @@ class DetectionInferenceTask(BaseTask, IInferenceTask, IExportTask, IEvaluationT
         output = results["outputs"]
         metric = output["metric"]
         predictions = output["detections"]
-        assert len(output["detections"]) == len(output["feature_vectors"]), (
+        assert len(output["detections"]) == len(output["feature_vectors"]) == len(output["saliency_maps"]), (
             "Number of elements should be the same, however, number of outputs are "
-            f"{len(output['detections'])}, {len(output['feature_vectors'])})"
+            f"{len(output['detections'])}, {len(output['feature_vectors'])}, and {len(output['saliency_maps'])}"
         )
         prediction_results = zip(predictions, output["feature_vectors"], output["saliency_maps"])
         return prediction_results, metric
--- a/external/model-preparation-algorithm/mpa_tasks/apis/detection/task.py
+++ b/external/model-preparation-algorithm/mpa_tasks/apis/detection/task.py
@@ -309,15 +309,13 @@ class DetectionInferenceTask(BaseTask, IInferenceTask, IExportTask, IEvaluationT
             if saliency_map is not None:
                 if saliency_map.ndim == 2:
                     # Single saliency map per image, support e.g. EigenCAM use case
-                    actmap = get_actmap(
-                        saliency_map, (dataset_item.width, dataset_item.height)
-                    )
+                    actmap = get_actmap(saliency_map, (dataset_item.width, dataset_item.height))
                     saliency_media = ResultMediaEntity(
                         name="Saliency Map",
                         type="saliency_map",
                         annotation_scene=dataset_item.annotation_scene,
                         numpy=actmap,
-                        roi=dataset_item.roi
+                        roi=dataset_item.roi,
                     )
                     dataset_item.append_metadata_item(saliency_media, model=self._task_environment.model)
                 elif saliency_map.ndim == 3:
@@ -326,22 +324,22 @@ class DetectionInferenceTask(BaseTask, IInferenceTask, IExportTask, IEvaluationT
                     num_saliency_maps = saliency_map.shape[0]
                     if num_saliency_maps == len(labels) + 1:
                         # Include the background as the last category
-                        labels.append(LabelEntity('background', Domain.DETECTION))
+                        labels.append(LabelEntity("background", Domain.DETECTION))
                     for class_id, class_wise_saliency_map in enumerate(saliency_map):
-                        actmap = get_actmap(
-                            class_wise_saliency_map, (dataset_item.width, dataset_item.height)
-                        )
+                        actmap = get_actmap(class_wise_saliency_map, (dataset_item.width, dataset_item.height))
                         class_name_str = labels[class_id].name
                         saliency_media = ResultMediaEntity(
                             name=class_name_str,
                             type="saliency_map",
                             annotation_scene=dataset_item.annotation_scene,
-                            numpy=actmap, roi=dataset_item.roi
+                            numpy=actmap,
+                            roi=dataset_item.roi,
                         )
                         dataset_item.append_metadata_item(saliency_media, model=self._task_environment.model)
                 else:
-                    raise RuntimeError(f'Single saliency map has to be 2 or 3-dimensional, '
-                                       f'but got {saliency_map.ndim} dims')
+                    raise RuntimeError(
+                        f"Single saliency map has to be 2 or 3-dimensional, " f"but got {saliency_map.ndim} dims"
+                    )
 
     def _patch_data_pipeline(self):
         base_dir = os.path.abspath(os.path.dirname(self.template_file_path))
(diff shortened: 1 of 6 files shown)


sungmanc:
run ote-test

sungmanc:
Check the results at first: https://ci-ote.iotg.sclab.intel.com/job/ote/job/pr-ote-test/809/

dongkwan-kim01:
run ote-test

external/model-preparation-algorithm/mpa_tasks/apis/detection/task.py
```diff
@@ -138,9 +138,9 @@ def _infer_detector(
         output = results["outputs"]
         metric = output["metric"]
         predictions = output["detections"]
-        assert len(output["detections"]) == len(output["feature_vectors"]) == len(output["saliency_maps"]), (
+        assert len(output["detections"]) == len(output["feature_vectors"]), (
             "Number of elements should be the same, however, number of outputs are "
-            f"{len(output['detections'])}, {len(output['feature_vectors'])}, and {len(output['saliency_maps'])}"
+            f"{len(output['detections'])}, {len(output['feature_vectors'])})"
         )
```
negvet, line 144: Does it cause test failure?

Yes, total number of saliency maps is `len(dataset) * N` where N is the number of classes. But single saliency map is actually a 3-dimentional tensor [N, H, W], where N is the number of classes. So basically `len(output["feature_vectors"]) == len(output["saliency_maps"])` actually should be `True`.

sungmanc:
@dongkwan-kim01, Is there any difference between this PR and [PR#1408](https://github.com/openvinotoolkit/training_extensions/pull/1408)?

negvet:
Are you sure that it fails because of the number of saliency maps?

Basically saliency map tensor created for every dataset point. It has dimensions [N, H, W]. Where N is the total number of classes for torch task or number of predicted classes for the OV task.

But the number of saliency maps should be equal to `len(dataset)` in any case I believe

sungmanc:
> Are you sure that it fails because of the number of saliency maps?
> 
> Basically saliency map tensor created for every dataset point. It has dimensions [N, H, W]. Where N is the total number of classes for torch task or number of predicted classes for the OV task.
> 
> But the number of saliency maps should be equal to `len(dataset)` in any case I believe

Then, could you check this FAIL case?  (http://validationreports.sclab.intel.com:8006/reports/test_result/638e0a31fd6d37c8e261ba3a)

The error log was as below:
image


The same errors were occurred in detection + tiling cases. @eugene123tw do you have any idea for this issue?
image

sungmanc:
Test results: http://validationreports.sclab.intel.com:8006/reports/build_number_report?test_session_build_number=pr-ote-test-810&environment=iotg-dev-workstation-08

Most FAIL cases were solved. Although there were still two FAIL cases, it was not related to this PR. How about merging MPA first and this PR to OTE? @harimkang @dongkwan-kim01 @goodsong81 @negvet

harimkang (dismissed):
Checking the test results, one is a score drop in cls (train vs export) problem, and one is OOM. I think the existing issue appears to have been resolved.

I have confirmed [this](https://ci-ote.iotg.sclab.intel.com/job/ote/job/pr-ote-test/810/)

goodsong81 (dismissed):
LGTM. I've approved MPA PRs.
@sungmanc Could you merge MPA PRs?
@dongkwan-kim01 Could you update MPA commits after merge?

dongkwan-kim01:
I've talked with Evgeny, but in the tile detection case, it seems like many saliency maps are exported for each tile.
So exception handle which `saliency_maps = [saliency_maps[i] for i in range(dataset.num_samples)]`, existed in the previous `develop`, can avoid the assertion error but maybe not make the appropriate saliency map(maybe saliency maps of first few tiles).
Maybe saliency maps merging function or getting the saliency map from the original image(before tiling) would be needed.

> LGTM. I've approved MPA PRs. @sungmanc Could you merge MPA PRs? @dongkwan-kim01 Could you update MPA commits after merge?

Yes, I'll update after the merge but I think we need some fixes in the saliency map for tiling case.. Maybe @negvet investigating and make another PR for it.

negvet:
run ote-test

negvet:
> I've talked with Evgeny, but in the tile detection case, it seems like many saliency maps are exported for each tile. So exception handle which `saliency_maps = [saliency_maps[i] for i in range(dataset.num_samples)]`, existed in the previous `develop`, can avoid the assertion error but maybe not make the appropriate saliency map(maybe saliency maps of first few tiles). Maybe saliency maps merging function or getting the saliency map from the original image(before tiling) would be needed.
> Yes, I'll update after the merge but I think we need some fixes in the saliency map for tiling case.. Maybe @negvet investigating and make another PR for it.

Actually it works just fine. Not that good, but functionally correct.
The first tile is the full image size - so `saliency_maps = [saliency_maps[i] for i in range(dataset.num_samples)]` actually provide the saliency maps for the whole images, which is correct (I deleted this line of code accidently, @dongkwan-kim01 correctly get it back).

The other problem here is that the model is not able to distinguish objects at full resolution (after being trained at tiles):
image
But it still can see the object occurrences (you can see the difference in object density at saliency maps) and can distinguish the background (it is SSD model)

It would be good to revisit this and merge saliency maps from tiles instead of generating it from full-res image. But for now it is fine IMO.

ote_sdk/ote_sdk/utils/tiler.py
```diff
@@ -79,7 +80,7 @@ def predict(self, image: np.ndarray):
             detections = np.append(detections, output, axis=0)
             # cache full image feature vector and saliency map at 0 index
             if i == 0:
-                features = feats
+                features = copy.deepcopy(feats)
```
negvet, line 84: `feats` is mutable, this helps to keep it from being overwritten when `i > 0`. This bug persisted before - seems that saliency maps (along with `feature_vector`) did not generate correctly before for the tiling case.

ote_sdk/ote_sdk/utils/tiler.py
```diff
@@ -128,7 +129,7 @@ def predict_tile(
             if "feature_vector" in raw_predictions or "saliency_map" in raw_predictions:
                 features = [
                     raw_predictions["feature_vector"].reshape(-1),
-                    raw_predictions["saliency_map"],
+                    raw_predictions["saliency_map"][0],
                 ]
```
negvet, line 133: Just removing the batch dimension, as in `BaseInferencerWithConverter.predict()`

wonjuleee (dismissed):
LGTM

dongkwan-kim01:
@negvet Thank you! So the original one was correct. Thank you for your confirmation.

harimkang (dismissed):

sungmanc (dismissed):

dongkwan-kim01:
run ote-test

dongkwan-kim01:
image
Since there some code change regarding Evgeny's commit and assertion revert, I ran the tile-detection in `tests/ote_cli/external/model-preparation-algorithm/ote_cli/test_tiling_detection.py` it locally and passed but also running the ote-test.

external/model-preparation-algorithm/tests/test_xai.py
```diff
@@ -9,7 +9,6 @@
 from mmdet.models import build_detector
 
 from mpa.modules.hooks.auxiliary_hooks import ReciproCAMHook, DetSaliencyMapHook
-from mpa.det.stage import DetectionStage
```
negvet, line 13: For me this test did not pass without this import locally. @dongkwan-kim01 could you please double check?

sungmanc: @dongkwan-kim01 Could you check this? I want to merge this branch

dongkwan-kim01: @negvet @sungmanc Yes I'll check it. Sorry for late response

dongkwan-kim01: You're right it's making error. I've added with flake8 exception

sungmanc (dismissed):

sungmanc:
Test results: http://validationreports.sclab.intel.com:8006/reports/build_number_report?test_session_build_number=pr-ote-test-815&environment=iotg-dev-workstation-10

harimkang (dismissed):

harimkang (approved):

sungmanc (approved):

PR #1403, opened 2022-11-29
[OTX] Enabling GPU execution for exported code

GalyaZalesskaya:
CVS-96658, CVS-96566

PR #1421, opened 2022-12-07
[OTE][Develop][bugfix] Fix ote deploy failed tests

GalyaZalesskaya:
Fixed ote deploy tests affected with changes from this [PR](https://github.com/openvinotoolkit/training_extensions/pull/1403).
Updated OTE SDK commit used in exportable code demo requirements

### DIFF
--- a/ote_sdk/ote_sdk/usecases/exportable_code/demo/requirements.txt
+++ b/ote_sdk/ote_sdk/usecases/exportable_code/demo/requirements.txt
@@ -1,3 +1,3 @@
 openvino==2022.1.0
 openmodelzoo-modelapi @ git+https://github.com/openvinotoolkit/open_model_zoo/@releases/2022/SCv1.1#egg=openmodelzoo-modelapi&subdirectory=demos/common/python
-ote-sdk @ git+https://github.com/openvinotoolkit/training_extensions/@561581d1f2118764237520a1520f46a1e9c87298#egg=ote-sdk&subdirectory=ote_sdk
+ote-sdk @ git+https://github.com/openvinotoolkit/training_extensions/@c29f1b385d5eabcf07bc9a6e9c340fd0dbf67c8d#egg=ote-sdk&subdirectory=ote_sdk


GalyaZalesskaya:
run ote-test

GalyaZalesskaya:
Build 819 in progress

goodsong81 (changes requested):
Why don't we just track the 'develop' branch in this develop branch?

sungmanc:
> Why don't we just track the 'develop' branch in this develop branch?

I agree, we should use `branch` nameinstead of `commit` number.

ote_sdk/ote_sdk/usecases/exportable_code/demo/requirements.txt
```diff
@@ -1,3 +1,3 @@
 openvino==2022.1.0
 openmodelzoo-modelapi @ git+https://github.com/openvinotoolkit/open_model_zoo/@releases/2022/SCv1.1#egg=openmodelzoo-modelapi&subdirectory=demos/common/python
-ote-sdk @ git+https://github.com/openvinotoolkit/training_extensions/@561581d1f2118764237520a1520f46a1e9c87298#egg=ote-sdk&subdirectory=ote_sdk
+ote-sdk @ git+https://github.com/openvinotoolkit/training_extensions/@c29f1b385d5eabcf07bc9a6e9c340fd0dbf67c8d#egg=ote-sdk&subdirectory=ote_sdk
```
sungmanc, line 3: As @goodsong81 suggestion, How about using branch name (i.e. `develop`) instead of commit number ?
If so, we could avoid the further mistakes

sovrasov: What if the develop branch contains some changes which are not compatible with ote? Can we control it?

goodsong81: - develop branch -> @develop (ensuring latest sync)
- release branches -> specific commit or tag (ensuring version match)
What do you think?

sovrasov: I'm fine, but actually proper testing should resolve the consistency problem (i.e. at some moment a faulty PR was merged).

If someone introduces breaking deploy-related changes into ote_sdk, and we have a link to develop here, the CI will be red, because it doesn't merge automatically to develop branch. To get a green CI, PR author should switch this commit, pass the CI, then switch to develop again, then merge. I think this would be an overhead. This time CI failure was caused by a review mistake, we should just be more careful when merging and that's all we can do when managing such self-dependencies.

sovrasov: At the other hand we could think about automatic commit switching for testing PRs, but this requires some efforts at the CI side.

eugene123tw: What if there is a bug in OTE SDK, and we need to fix it? `develop` in requirement is going to be one version behind the hotfix, right?

At the moment, the commit is used as a tag, and instead of tracking `develop` branch, I think it'd be better if we track version numbers instead.

sungmanc: @eugene123tw I agree, as you said if we want to fix the bug located in `OTE SDK`, using the branch name is hard to maintain. At this moment, using version number is the appropriate way and consider a better way to reduce the mistake that could be happened during the PR review.

GalyaZalesskaya: Thank you for the discussion and your thoughts. I replaced commit number with branch name

GalyaZalesskaya:
run ote-tests

GalyaZalesskaya:
run ote-test

GalyaZalesskaya:
https://ci-ote.iotg.sclab.intel.com/job/ote/job/pr-ote-test/841/

goodsong81 (changes requested):
Sorry for bothering you again, but I think the commit hash is the best option for now as discussed in this PR.

- Release branch -> release tag
- Develop branch -> specific commit

Could you apply the latest commit, again?

GalyaZalesskaya:
@goodsong81  I did it. I'm sorry, I got confused with discussion conclusions

harimkang (approved):

goodsong81 (approved):

PR #1425, opened 2022-12-08
[OTX/Semi-SL-Seg] Enable training and evaluation for semi supervised learning in segmentation

supersoob:
**This PR includes**
- Make templates and configs folder for semi-sl
- Update trainer/inferrer types corresponding to train_type
- Revisit is_task_adapt flag

TODOs
- [x] Pre-commit tests
- [x] Test codes


otx/algorithms/segmentation/tasks/inference.py
```diff
@@ -187,31 +188,34 @@ def _init_recipe(self):
...
             raise NotImplementedError(f"Train type {train_type} is not implemented yet.")
         if train_type == TrainType.SEMISUPERVISED:
-            if self._data_cfg.get("data", None) and self._data_cfg.data.get("unlabeled", None):
-                recipe = os.path.join(recipe_root, "semisl.py")
-            else:
-                logger.warning("Cannot find unlabeled data.. convert to INCREMENTAL.")
-                train_type = TrainType.INCREMENTAL
-
+            recipe = os.path.join(recipe_root, "semisl.py")
         if train_type == TrainType.INCREMENTAL:
```
kprokofi, line 204: "elif" is better

kprokofi (commented):
minor.

otx/algorithms/segmentation/configs/ocr_lite_hrnet_18/model.py
```diff
@@ -22,7 +22,6 @@
 
 model = dict(
     type="ClassIncrSegmentor",
-    is_task_adapt=True,
```
jaegukhyun, line 25: Doesn't semi-sl support task adaptation?
As far as I know, task adapt is just for sampler setting

otx/algorithms/segmentation/configs/ocr_lite_hrnet_18/semisl/__init__.py
```diff
@@ -0,0 +1,15 @@
+"""Initialization of OCR-Lite-HRnet-18 model for Segmentation Task."""
```
jaegukhyun: How about clarify that this is semisl model?

otx/algorithms/segmentation/configs/ocr_lite_hrnet_18/semisl/model.py
```diff
@@ -0,0 +1,22 @@
+"""Model configuration of OCR-Lite-HRnet-18 model for Segmentation Task."""
```
jaegukhyun: Same, how about add description that this is model for semisl

jaegukhyun (changes requested):
Generally it looks good to me, thank you for your effort. I left a few comment.

otx/algorithms/segmentation/configs/ocr_lite_hrnet_18/model.py
supersoob: is_task_adapt is not the flag for task adaptation for sampler setting. It is redundant arg just for logging.

otx/algorithms/segmentation/tasks/inference.py
supersoob: I changed code, please look at it again.

supersoob:
@jaegukhyun @kprokofi @JihwanEom I revised codes based on your reviews. I request you to review again. I resolved your comments. If you need to add comments or still have a questions to resolved question, unresolve it or add some.

otx/algorithms/segmentation/configs/base/models/semi_mt_segmentor.py
```diff
@@ -0,0 +1,10 @@
+"""Segmentor config for semi-supervised learning."""
+
+model = dict(
+    type="MeanTeacher",
+    orig_type="EncoderDecoder",
+    unsup_weight=0.1,
+    train_cfg=dict(mix_loss=dict(enable=False, weight=0.1)),
+    test_cfg=dict(mode="whole", output_scale=5.0),
```
JihwanEom: Do we need `output_scale`?

otx/algorithms/segmentation/configs/ocr_lite_hrnet_18/semisl/__init__.py
```diff
@@ -0,0 +1,15 @@
+"""Initialization of OCR-Lite-HRnet-18 model for Semi-SL Segmentation Task."""
```
JihwanEom, line 1: I think we don't need semi-sl for `OCR-Lite-HRNet-18` architecture, this backbone is deprecated. We only need to leave template.yaml to be scanned well the template id when the user upgrade from Geti 1.0 to Geti 1.1

otx/algorithms/segmentation/configs/ocr_lite_hrnet_18/semisl/model.py
```diff
@@ -0,0 +1,22 @@
+"""Semi-SL model configuration of OCR-Lite-HRnet-18 model for Segmentation Task."""
```
JihwanEom, line 1: Same as above.

otx/algorithms/segmentation/configs/base/models/semi_mt_segmentor.py
```diff
@@ -0,0 +1,10 @@
+"""Segmentor config for semi-supervised learning."""
+
+model = dict(
+    type="MeanTeacher",
+    orig_type="EncoderDecoder",
+    unsup_weight=0.1,
+    train_cfg=dict(mix_loss=dict(enable=False, weight=0.1)),
```
JihwanEom: As far as I know, `mix_loss` is designed for OCR head.. :( 
Could you verify my opinion and remove if I'm right?

otx/algorithms/segmentation/tasks/inference.py
```diff
@@ -78,6 +78,9 @@ def __init__(self, task_environment: TaskEnvironment):
         self.freeze = True
         self.metric = "mDice"
         self._label_dictionary = {}  # type: Dict
+        self.train_type = None
```
sungchul1: If `train_type` is used globally and it is never modified except for L206, how about `self.train_type = self._hyperparams.algo_backend.train_type` here instead of L190 and L213?

otx/algorithms/segmentation/tasks/inference.py
```diff
@@ -78,6 +78,9 @@ def __init__(self, task_environment: TaskEnvironment):
...
+        self.base_model_dir = None
+        self.base_pipeline_path = None
```
sungchul1: AFAIK, `base` is used as root configs in otx and I think `base_` is a prefix that can cause confusion.
What about other prefixes, such as `task_`, etc.?

sungchul1:
I left some comments during referring to this for self-sl.

otx/algorithms/segmentation/tasks/inference.py
supersoob: Make sense. How about revising them too in detection? @jaegukhyun. It's not requirement to name as same across tasks, we may work on other PR about uniform naming and architectures.

otx/algorithms/segmentation/configs/base/models/semi_mt_segmentor.py
supersoob: I just move the cfg the old mpa cfg. But it's obvious that it needs test_cfg. Is there any problem with this value?

otx/algorithms/segmentation/configs/ocr_lite_hrnet_18/semisl/__init__.py
supersoob: We resolved this discussion in the group chat.

otx/algorithms/segmentation/configs/base/models/semi_mt_segmentor.py
supersoob: It was not designed for OCR. It was using inside each head and it was just specified in cfg to make sure to disable mix_loss for test.

otx/algorithms/segmentation/tasks/inference.py
supersoob: It won't work because self._hyperparams is not declared yet. The reason why I defined train_type here is to avoid pylint attribute-defined-outside-init error.

sungchul1 (dismissed):
LGTM

supersoob:
image
Local test results -> seg passed all cli tests. I hope we can merge this asap, because it would affect the seg task without this code because MPA PR is merged.

kprokofi (dismissed):

otx/algorithms/segmentation/configs/base/models/semi_mt_segmentor.py
```diff
@@ -0,0 +1,10 @@
+"""Segmentor config for semi-supervised learning."""
```
sungmanc: How about making `semisl` folder inside of `otx/algorithms/segmentation/configs/base/models`?
I can't imagine the meaning of `mt`  in this case. 

If don't want to make `semisl` folder, changing name of this file is needed for easy understanding. 
(i.e. Since this python file was already located in the `segmentation` folder, we can delete segmentor for this case. `semi_mt_segmentor.py` --> `semi_mean_teacher.py` or `mean_teacher.py`

otx/algorithms/segmentation/tasks/inference.py
harimkang: Wouldn't this be possible by declaring it after super().__init__?

otx/algorithms/segmentation/configs/base/models/semi_mt_segmentor.py
harimkang: I agree with sungman on this. But rather than creating a separate folder, just mean_teacher.py looks better.

otx/algorithms/segmentation/tasks/inference.py
harimkang: I agree with sungchul's opinion. This is the path of `data_pipeline`, so I think the word `data` should be included.

harimkang (changes requested):
I left only minor comments. However, this PR seems to need to be fixed by focusing on the test results.
Most of the fail is the current opencv environment problem, but the issue below seems to be a problem related to the current PR. could you please check it?
image

otx/algorithms/segmentation/tasks/inference.py
supersoob: I don't know whether to necessarily change it.. Could you explain why?

otx/algorithms/segmentation/tasks/inference.py
supersoob: this is outdated.

otx/algorithms/segmentation/tasks/inference.py
harimkang: I don't know why we need to avoid the attribute-defined-outside-init error. Does train_type affect super().__init__?

supersoob:
> I left only minor comments. However, this PR seems to need to be fixed by focusing on the test results. Most of the fail is the current opencv environment problem, but the issue below seems to be a problem related to the current PR. could you please check it? image

It is not due to this PR. It failed all seg task including incremental learning after MPA XAI PR merged. It seems like it has to be dealt in MPA side in other PR.

otx/algorithms/segmentation/tasks/inference.py
supersoob: Actually, self.train_type is re-defined inside in the _init_recipe. If I add the self.train_type code in init, it's code duplication. I rather chose to avoid attribute-defined-outside-init error here by ignoring.

supersoob:
@jaegukhyun @kprokofi @JihwanEom @sungmanc @harimkang I revised codes based on your reviews. I request you to review again. I resolved your comments. If you need to add comments or still have a questions to resolved question, unresolve it or add some.

sungmanc (approved):
LGTM

harimkang (approved):

jaegukhyun (approved):

PR #1426, opened 2022-12-08
[feature/otx] Changes in available backbone output formats

harimkang:
## Summary
Change auto collection of backbone list to json format list of available backbones.

## This PR includes
- Create `otx/cli/builder/supported_backbone` folder that include available backbone list (json file)
image
- If there is a related json file, use that data for `otx find & build`
- Updated required args to show available options in `otx find` as well
image
- Modified to build backbone configuration using the first value as default when selectable options are provided among required args
- Add raise error statement for exceptions

### DIFF
--- a/otx/cli/builder/builder.py
+++ b/otx/cli/builder/builder.py
@@ -29,7 +29,7 @@ from mmcv.utils import Registry, build_from_cfg
 from mpa.utils.config_utils import MPAConfig
 
 from otx.cli.registry import Registry as OTXRegistry
-from otx.cli.utils.importing import get_backbone_registry
+from otx.cli.utils.importing import get_available_backbone_list, get_backbone_registry
 
 DEFAULT_MODEL_TEMPLATE_ID = {
     "CLASSIFICATION": "Custom_Image_Classification_EfficinetNet-B0",
@@ -61,6 +61,8 @@ def update_backbone_args(backbone_config, registry, skip_missing=False):
     Also, it distinguishes the argment needed for the build to add convenience to the user.
     """
     backbone_function = registry.get(backbone_config["type"])
+    if not backbone_function:
+        raise ValueError(f"{backbone_config['type']} is not supported backbone")
     required_option = {}
     required_args, missing_args = [], []
     args_signature = inspect.signature(backbone_function)
@@ -69,7 +71,6 @@ def update_backbone_args(backbone_config, registry, skip_missing=False):
         if arg_key == "out_indices":
             use_out_indices = True
         if arg_value.default is inspect.Parameter.empty:
-            # TODO: How to get argument type (or hint or format)
             required_args.append(arg_key)
             if hasattr(backbone_function, "arch_settings"):
                 arg_options = [str(option) for option in backbone_function.arch_settings.keys()]
@@ -115,6 +116,27 @@ def update_backbone_args(backbone_config, registry, skip_missing=False):
         )
     backbone_config["use_out_indices"] = use_out_indices
     return missing_args
+
+
+def patch_missing_args(backbone_config, missing_args):
+    """Patch backbone's required arg configuration."""
+    updated_missing_args = []
+    backbone_type = backbone_config["type"]
+    backend, _ = Registry.split_scope_key(backbone_type)
+    backbone_data = get_available_backbone_list(backend)
+    if backbone_type not in backbone_data:
+        return missing_args
+    backbone_data = backbone_data[backbone_type]
+    for arg in missing_args:
+        if "options" in backbone_data and arg in backbone_data["options"]:
+            backbone_config[arg] = backbone_data["options"][arg][0]
+            print(
+                f"[otx build] '{arg}' can choose between: {backbone_data['options'][arg]}"
+                f"\n[otx build] '{arg}' default value: {backbone_config[arg]}"
+            )
+        else:
+            updated_missing_args.append(arg)
+    return updated_missing_args
 
 
 def update_in_channel(model_config, out_channels):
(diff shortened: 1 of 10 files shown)


otx/cli/builder/builder.py
```diff
@@ -117,6 +118,27 @@ def update_backbone_args(backbone_config, registry, skip_missing=False):
...
 
+def patch_missing_args(backbone_config, missing_args):
+    """Patch backbone's required arg configuration."""
+    updated_missing_args = []
+    backbone_type = backbone_config["type"]
+    backend, _ = Registry.split_scope_key(backbone_type)
+    backbone_data = get_available_backbone_list(backend)
+    if backbone_type not in backbone_data:
+        return missing_args
+    backbone_data = backbone_data[backbone_type]
```
goodsong81: backbone_data -> available_backbones ?
backbone_data = available_backbones[backbone_type] ?

otx/cli/tools/find.py
```diff
@@ -93,20 +80,24 @@ def main():
...
+                    if arg in meta_data["options"]:
+                        output_arg += f"={meta_data['options'][arg]}"
+                    required_args.append(output_arg)
+                row = [
+                    row_index + 1,
+                    backbone_type,
+                    ", ".join(required_args) if required_args else "",
+                    ", ".join(meta_data["available"]) if meta_data["available"] else "",
+                ]
+                backbone_table.add_row(row)
```
goodsong81, line 101: Nit: Better to be factored out as a function. Not mandatory :)

goodsong81 (dismissed):
Thank you for the PR, what a nice work, indeed.
I'd like to point out some future refactoring suggestions.

- Not sure of 'algorithm' or 'backend', but anyway it should be the source of available/supported backbone list.
- CLI / registry should just gather info from algo/backend.
- get_backbones(), build_model(), etc could be encapsulated in algo/backend logic with proper interface.
- Then, we can support find/build also for anomaly with their existing 1~2 backbones w/o a lot of if-elses in CLI module.

otx/cli/builder/builder.py
harimkang: Edited as per your comment.

otx/cli/tools/find.py
harimkang: Yes this is something to consider. Added a TODO comment

harimkang:
> * Not sure of 'algorithm' or 'backend', but anyway it should be the source of available/supported backbone list.

-> I'm confused about the names too. I'll think about it some more, but I don't think the `algorithm` is appropriate, so I'm currently using it as a `backend`.
> * CLI / registry should just gather info from algo/backend.

-> Yes, it is. Functions that currently provide info are in utils or builder. These also need to be sorted out.
> * get_backbones(), build_model(), etc could be encapsulated in algo/backend logic with proper interface.

-> Functions are mainly divided into `builder.py` and `importing.py` of utils, and I also think that this should be changed into an appropriate structure.
> * Then, we can support find/build also for anomaly with their existing 1~2 backbones w/o a lot of if-elses in CLI module.

-> I agree. It looks like a change is needed for each CLI in a way that can be used for all tasks. I'm also thinking of a way to build by applying the parameters of the backbone through the args of the CLI as the next task. i will consider this as well.

Thanks for the nice comment!

goodsong81 (approved):
LGTM.

tests/integration/cli/test_cli.py
```diff
@@ -30,8 +31,8 @@ def tmp_dir_path():
 class TestToolsOTXCLI:
     @e2e_pytest_component
     def test_otx_find(self):
-        otx_find_testing(otx_dir)
+        otx_find_testing()
 
     @e2e_pytest_component
     def test_otx_build(self, tmp_dir_path):
-        otx_build_testing(tmp_dir_path, otx_dir, build_backbone_args)
+        otx_build_testing(tmp_dir_path, build_backbone_args)
```
sungmanc, line 38: Do you plan to implement a test code for checking all available backbones?
Actually, the most challenging point of this work is not making errors for our available backbones.

otx/cli/utils/importing.py
```diff
@@ -52,8 +62,35 @@ def get_backbone_list(backend):
...
 
+def get_available_backbone_list(backend):
+    """Gather available backbone list from json file."""
+    available_backbone_path = os.path.join(get_otx_root_path(), "cli/builder/supported_backbone")
+    available_backbone_path = os.path.join(available_backbone_path, f"{backend}.json")
+    available_backbones = {}
+    if os.path.exists(available_backbone_path):
+        with open(available_backbone_path, "r", encoding="UTF-8") as f:
+            available_backbones = json.load(f)
+        available_backbones = available_backbones["backbones"]
```
sungmanc, line 73: mmcls, mmdet, mmseg backbones don't need `required` or `options` or `available` option?

sungmanc (approved):
LGTM. Currently, I think the current `Builder` has too many functions. we can refactor the `builder` in the near future. (not mandatory for this PR)

tests/integration/cli/test_cli.py
harimkang: It is possible, including the issues you mentioned, the issue of increasing the test time is also likely to occur. Let's discuss later.

otx/cli/utils/importing.py
harimkang: All mmcls, mmdet, and mmseg backbones use corresponding `options`, and `required` are used for both find and build. This is already created in the form of a json file.

sungmanc: Ah, I missed that, Thanks

PR #1431, opened 2022-12-12
[feature/otx] Unifying the data_pipeline policy of all tasks

harimkang:
## Summary
- Add base_data_pipeline_path in template for base data pipeline for classification & segmentation
- Import the data pipeline of `MPA recipe` to `OTX` level
- Data pipeline note priority: Preferentially use the data pipeline in the order below
    1. data_pipeline.py in same folder as template
    2. base_data_pipeline_path in template
    3. MPA recipe's data pipeline
   
## Why is this needed?
- OTX users cannot change the current data pipeline for any task other than detection.
- Bring the data pipeline configuration of each task to OTX, and put the commonly used ones in `configs/base/data`.
- Necessity of data pipeline in some models for public backbone building.


supersoob (dismissed):
Thanks for moving pipeline files to otx. Overall, it looks good. I'd say that it would be good that data_cfg is moved as well and replace base_data_pipeline_path to base_data_path which contains data pipeline cfg together. But let's do it gradually.

sungmanc (dismissed):
LGTM but minor opinion for naming. I think `data_pipeline.py` is more attract instead of `cls_incr_data_pipeline.py`.

otx/algorithms/classification/configs/base/data/cls_incr_data_pipeline.py
```diff
@@ -0,0 +1,45 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions
+# and limitations under the License.
+
+# pylint: disable=invalid-name
+
+img_norm_cfg = dict(mean=[123.675, 116.28, 103.53], std=[58.395, 57.12, 57.375], to_rgb=True)
```
wonjuleee: why don't you use `__img_norm_cfg` instead of `img_norm_cfg`?

otx/algorithms/common/adapters/mmcv/utils.py
```diff
@@ -70,12 +70,15 @@ def config_from_string(config_string: str) -> Config:
...
+def patch_data_pipeline(config: Config, base_dir: str, base_data_pipeline: str = ""):
     """Replace data pipeline to data_pipeline.py if it exist."""
-    data_pipeline_path = os.path.join(base_dir, "data_pipelne.py")
+    data_pipeline_path = os.path.join(base_dir, "data_pipeline.py")
     if os.path.exists(data_pipeline_path):
         data_pipeline_cfg = Config.fromfile(data_pipeline_path)
         config.merge_from_dict(data_pipeline_cfg)
+    elif os.path.exists(base_data_pipeline):
+        data_pipeline_cfg = Config.fromfile(base_data_pipeline)
+        config.merge_from_dict(data_pipeline_cfg)
```
wonjuleee, line 77: how about using `data_pipeline.py` when `base_data_pipeline' doesn't exist?
I meant that we can use
if os.path.exists(base_data_pipeline):
 ...
elif os.path.exists(os.path.join(base_dir, "data_pipeline.py")):
 ...
else:
 raise error

otx/algorithms/detection/adapters/mmdet/utils/config_utils.py
```diff
@@ -264,16 +263,6 @@ def patch_evaluation(config: Config):
...
-# TODO Replace this with function in common
-def patch_data_pipeline(config: Config, template_file_path: str):
-    """Update data_pipeline configs."""
-    base_dir = os.path.abspath(os.path.dirname(template_file_path))
-    data_pipeline_path = os.path.join(base_dir, "data_pipeline.py")
-    if os.path.exists(data_pipeline_path):
-        data_pipeline_cfg = Config.fromfile(data_pipeline_path)
-        config.merge_from_dict(data_pipeline_cfg)
-
-
```
wonjuleee, line 276: why should this be relocated to `utils.py`?

otx/cli/builder/builder.py
```diff
@@ -200,14 +200,23 @@ def build_task_config(self, task_type, model_type=None, workspace_path=None, otx
...
-        # Copy Data config
+        # Copy Data pipeline config
         if os.path.exists(os.path.join(template_dir, "data_pipeline.py")):
             data_pipeline_config = MPAConfig.fromfile(os.path.join(template_dir, "data_pipeline.py"))
             data_pipeline_config.dump(os.path.join(workspace_path, "data_pipeline.py"))
+        elif os.path.exists(os.path.join(template_dir, template_config.base_data_pipeline_path)):
+            data_pipeline_config = MPAConfig.fromfile(
+                os.path.join(template_dir, template_config.base_data_pipeline_path)
+            )
+            data_pipeline_config.dump(os.path.join(workspace_path, "data_pipeline.py"))
```
wonjuleee, line 205: I recommend to correct here so as to the above.

harimkang:
> 

@supersoob 
Thanks for the comment! However, I think that the `base_data_path` you suggested can be confusing because it feels like a `data-root` or a `path of a dataset`. If you have any other good suggestions, please comment.

harimkang:
> > 
> 
> @supersoob Thanks for the comment! However, I think that the `base_data_path` you suggested can be confusing because it feels like a `data-root` or a `path of a dataset`. If you have any other good suggestions, please comment.

@sungmanc It's a suggestion worth considering. What about base_data_pipeline.py? Would data_pipeline.py be better?

harimkang:
> LGTM but minor opinion for naming. I think `data_pipeline.py` is more attract instead of `cls_incr_data_pipeline.py`.

@sungmanc It's a suggestion worth considering. What about base_data_pipeline.py? Would data_pipeline.py be better?

otx/algorithms/classification/configs/base/data/cls_incr_data_pipeline.py
harimkang: That's a great suggestion! In line with this, i will go through and modify all the variables in `data_pipeline.py`.

otx/algorithms/common/adapters/mmcv/utils.py
harimkang: First of all, I prioritized `data_pipeline.py` in the template folder. If the file does not exist, it is configured to look for `base_data_pipeline`.

I think both have pros and cons, do you think `base_data_pipeline` should be given priority? 
+
I got the idea from your suggestion!
There seems to be a way to combine both to find only `base_data_pipeline` without finding `data_pipeline.py` in template folder (This can be replaced with just `base_data_pipeline_path="./data_pipeline.py"`.) in the template.

otx/algorithms/detection/adapters/mmdet/utils/config_utils.py
harimkang: The function used only in `detection` has been moved to `common` utils so that it can be commonly used in all tasks.

otx/cli/builder/builder.py
harimkang: I will modify it in such a way that it does not directly search for `data_pipeline.py` in template, but only searches for `base_data_pipeline_path`. Thanks for the nice comment!

otx/algorithms/common/tasks/training_base.py
```diff
@@ -167,6 +167,14 @@ def template_file_path(self):
         """Model Template file path."""
         return self._task_environment.model_template.model_template_path
 
+    @property
+    def base_data_pipeline_path(self):
```
jaegukhyun: Is this function used?

jaegukhyun (changes requested):
Generally LGTM, I left one command. I'll revisit after reflecting Wonju's request

otx/algorithms/classification/configs/base/data/cls_incr_data_pipeline.py
harimkang: Edited as suggested.

otx/algorithms/common/adapters/mmcv/utils.py
harimkang: I modified things related to this based on `data_pipeline_path` (rename base_data* -> data_).

otx/algorithms/common/tasks/training_base.py
harimkang: Yes this is called as `self.data_pipeline_path` in the related task.

sungmanc:
> hat about base_data_pipeline.py

`base_data_pipeline.py` also seems good. I just want to delete `cls_incr`, because all of our tasks assume the `cls_incr` condition.

harimkang:
> > hat about base_data_pipeline.py
> 
> `base_data_pipeline.py` also seems good. I just want to delete `cls_incr`, because all of our tasks assume the `cls_incr` condition.

Yes I modified this to `data_pipeline.py`. (https://github.com/openvinotoolkit/training_extensions/pull/1431/commits/231273edd6f9bd75949bbd3208f19f350faa54a3)

otx/algorithms/action/adapters/mmaction/utils/config_utils.py
```diff
@@ -35,12 +35,12 @@
 
 
 @check_input_parameters_type()
-def patch_config(config: Config, base_dir: str, work_dir: str, task_type: TaskType):
+def patch_config(config: Config, data_pipeline_path: str, work_dir: str, task_type: TaskType):
```
harimkang, line 38: I modified two files of action-related codes. Could you please check again?

goodsong81 (dismissed):
Changes look good to me.
However, I don't think users need to be able to directly modify data pipeline (except for the advance users like OTX developers).
Rather specific pipelines are tied to model & train type so that users would just select model and/or train type. Some details should be exposed via h-params, and OTX would adapt pipeline settings, e.g. image resolution, etc.

sungmanc:
> I don't think users need to be able to directly modify data pipeline (except for the advance users like OTX developers)

I agree, however, we need to clarify which configuration is the final thing that overwrites other configurations for developers who are using OTX. (maybe on the documentation or docstring?)

goodsong81 (approved):

JihwanEom (approved):

jaegukhyun (approved):

PR #1435, opened 2022-12-13
[OTX-MMCLS] Enable NNCF

cih9088:
### [Related MPA PR](https://github.com/openvinotoolkit/model_preparation_algorithm/pull/106)

**Changes**
- update mmcls version
- enable NNCF for mmcls models
- refactoring
- remove `force` keyword in `_initialize()`
- pickable NNCF model build function for multi-gpu use case
- proper use of `drop_last` by `{subset}_dataloader`


otx/algorithms/classification/adapters/mmcls/data/datasets.py
```diff
@@ -279,7 +279,6 @@ def evaluate(
                 if k in metrics:
                     eval_results[k] = v
 
-        eval_results["accuracy"] = mAP_value
```
goodsong81, line 282: Is this duplicate?

otx/algorithms/classification/adapters/mmcls/nncf/builder.py
```diff
@@ -0,0 +1,170 @@
...
+    remove_from_configs_by_type(custom_hooks, "TaskAdaptHook")
+    remove_from_configs_by_type(custom_hooks, "AdaptiveTrainSchedulingHook")
+
+    for hook in get_configs_by_dict(custom_hooks, dict(type="OTXProgressHook")):
+        time_monitor = hook.get("time_monitor", None)
+        if (
+            time_monitor
+            and getattr(time_monitor, "on_initialization_end", None) is not None
+        ):
+            time_monitor.on_initialization_end()
```
goodsong81, line 168: What is the purpose of this block?

otx/algorithms/classification/adapters/mmcls/nncf/registers.py
```diff
@@ -0,0 +1,17 @@
...
+if is_nncf_enabled():
+    from nncf.torch import register_module
+    from timm.models.layers.conv2d_same import Conv2dSame
+
+    # Register custom modules.
+    # Users of nncf should manually check every custom
+    # layer with weights which should be compressed and
+    # in case such layers are not wrapping by nncf,
+    # wrap such custom module by yourself.
+    register_module(ignored_algorithms=[])(Conv2dSame)
```
goodsong81, line 17: Could you explain a bit? `Conv2dSame` is used in our models but not wrapped by NNCF?

otx/algorithms/classification/tasks/nncf.py
```diff
@@ -14,14 +14,50 @@
...
 
-class OTXClassificationNNCFTask(ClassificationNNCFTask):
-    """Task for compressing classification models using NNCF."""
 
-    def __init__(self, task_environment: TaskEnvironment):  # pylint: disable=useless-parent-delegation
+class ClassificationNNCFTask(NNCFBaseTask, ClassificationInferenceTask):
+    @check_input_parameters_type()
+    def __init__(self, task_environment: TaskEnvironment):
         super().__init__(task_environment)
+        self._label_dictionary = dict(enumerate(self._labels, 1))
```
goodsong81, line 39: Just curious, why start from 1?

otx/algorithms/common/adapters/mmcv/nncf/hooks.py
```diff
@@ -24,6 +27,17 @@ def before_run(self, runner):
         if runner.rank == 0:
             runner.logger.info(self.compression_ctrl.statistics().to_str())
 
+    @master_only
+    def after_run(self, runner):
+        compression_state = self.compression_ctrl.get_compression_state()
+        for algo_state in compression_state.get("ctrl_state", {}).values():
+            if not algo_state.get("scheduler_state"):
+                algo_state["scheduler_state"] = {"current_step": 0, "current_epoch": 0}
```
goodsong81, line 35: Needed but sometimes missing?

otx/algorithms/common/tasks/nncf_base.py
```diff
@@ -0,0 +1,368 @@
...
+        config.merge_from_dict(self._model_cfg)
+        self._recipe_cfg.custom_hooks = custom_hooks
+
+        def update(d, u):
+            for k, v in u.items():
+                if isinstance(v, Mapping):
+                    d[k] = update(d.get(k, {}), v)
+                else:
+                    d[k] = v
+            return d
```
goodsong81, line 346: Could you describe why we need this?

goodsong81 (approved):
Great work, nice refactoring. LGTM, but one concern is that it looks hard to merge/rebase changes in feature/otx branch due to refactoring work.

Could you answer a few questions of mine for understanding?
Then, let's merge to base branch, and move on to the next step.

otx/algorithms/classification/adapters/mmcls/data/datasets.py
cih9088: I have no idea where this came from but I believe this is a bug becuase `mAP_value` is only defiend if `mAP` is in a list of `metrics` and `mAP_value` is assigned to `eval_results` with a key `mAP`. Please see [these lines](https://github.com/cih9088/training_extensions/blob/2816c96595186135281a36733e56fb7e1cdd09b3/otx/algorithms/classification/adapters/mmcls/data/datasets.py#L272-L274).

otx/algorithms/classification/adapters/mmcls/nncf/builder.py
cih9088: This block is for [this callback](https://github.com/openvinotoolkit/training_extensions/blob/feature/otx-new-mmcv/base/otx/algorithms/common/tasks/nncf_base.py#L277-L281). `OptimizationProgressCallback` has an additional method [`on_initialization_end`](https://github.com/openvinotoolkit/training_extensions/blob/9002dda17ae0801fadf35990ed2bf86500a78b31/otx/algorithms/common/utils/callback.py#L117-L121), which needs to be called after initialization of NNCF model.

otx/algorithms/classification/adapters/mmcls/nncf/registers.py
cih9088: Yes, exactly. NNCF only supports native pytorch modules. `Conv2dSame` module is used in a `EfficientNet-V2-S` classification model and this module is not wrapped by NNCF because it is defined by user, here `timm` library. This is a shortcomings of NNCF as one has to register every custom modules to be wrapped by NNCF. 
[Here](https://github.com/openvinotoolkit/nncf/blob/develop/docs/Usage.md#compression-of-custom-modules) is a relative section in NNCF docs

otx/algorithms/common/adapters/mmcv/nncf/hooks.py
cih9088: Not sure this is sometimes missing. I just copied it from an original code block.

otx/algorithms/common/tasks/nncf_base.py
cih9088: This inline function is to update nested `dict` because naive `update` method of `dict` does not consider nested values.

otx/algorithms/classification/tasks/nncf.py
cih9088: Thanks for catching it. This attribute is only for segmentation task. It is considered a bug and it would be removed in mmdeploy PR as this PR has been merged.

I think starting from 1 is because 0 is for background.

cih9088:
> Great work, nice refactoring. LGTM, but one concern is that it looks hard to merge/rebase changes in feature/otx branch due to refactoring work.

Yes, that would be a quite big huddle but I believe I can handle it.

PR #1438, opened 2022-12-13
[OTX] Add warmstart for classification

sungchul1:
This PR includes
* Warmstart for classification recipe
* Modify learning rate max value to use larger learning rate (0.45 for warmstart) than 0.1
* Add `SELFSUPERVISED` in `algo_backend.train_type.options` in `otx/algorithms/classification/configs/configuration.yaml`
* Enable to handle iter runner
* Disable validation during warmstart
* Enable to only use train set without label
* Disable evaluation after warmstart (there is an error because warmstart task doesn't have validation set)

Integration test : `tests/integration/cli/classification/test_classification.py`

FYI.
This PR has dependency with https://github.com/openvinotoolkit/model_preparation_algorithm/pull/108.

TODOs
- [x] Refactoring
- [x] Docstring


otx/algorithms/classification/adapters/mmcls/data/datasets.py
```diff
@@ -411,3 +412,48 @@ def evaluate(
...
+        self.pipeline2 = Compose([build_from_cfg(p, PIPELINES) for p in pipeline['view1']])
+
+    def __len__(self):
+        """Get dataset length."""
+        return len(self.otx_dataset)
+
+    def __getitem__(self, index: int):
+        """Get item from dataset."""
+        dataset = self.otx_dataset
+        item = dataset[index]
```
jaegukhyun, line 438: How about item = self.otx_dataset[index]?

otx/algorithms/classification/adapters/mmcls/models/heads/contrastive_head.py
```diff
@@ -0,0 +1,54 @@
+# pylint: disable=missing-module-docstring, unused-argument
```
jaegukhyun: Adding short description would be more propriate than disable pylint

otx/algorithms/classification/configs/efficientnet_b0_cls_warmstart/model.py
```diff
@@ -0,0 +1,41 @@
...
+            type="SelfSLMLP",
+            in_channels=256,
+            hid_channels=4096,
+            out_channels=256,
+            with_avg_pool=False
+        )
+    )
+)
+
+custom_hooks = [
```
jaegukhyun: We normally locate custom hooks in recipes. If warmstart needs this hook regardless models, than this could be added to warmstart recipe

otx/algorithms/common/tasks/training_base.py
```diff
@@ -101,15 +101,16 @@ def _run_task(self, stage_module, mode=None, dataset=None, **kwargs):
...
-        # update model config -> model label schema
-        data_classes = [label.name for label in self._labels]
-        model_classes = [label.name for label in self._model_label_schema]
-        self._model_cfg["model_classes"] = model_classes
-        if dataset is not None:
-            train_data_cfg = Stage.get_train_data_cfg(self._data_cfg)
-            train_data_cfg["data_classes"] = data_classes
-            new_classes = np.setdiff1d(data_classes, model_classes).tolist()
-            train_data_cfg["new_classes"] = new_classes
+        if not self._warmstart:
```
jaegukhyun: Could you explain why warmstart avoid this block?

jaegukhyun (changes requested):
Thank you for your hard work. It is a big PR indeed, there are several things we have to discuss
Before we start, I want to recommend to check pre-commit issue in this PR, and also recommend to follow pylint except inevitable cases for more unified and informative code.

And the followed things should be discussed with OTX TF 
1. Is it appropriate to generate separate folder for sefl-sl? Actually we have agreed on one template for one model policy
2. How we test self-sl properly? Selfsl normally requires lots of epochs to train, however it is not appropriate for test because it requires tons of time.
3. Can we implement more scalable code for warmstart? There are many if else statement for warmstart, it doesn't seem scalable. After discussing about what is different between warmstart algorithm and normal incremental or semi-supervised learning, we can agree on how to divide warmstart stage more appropriatly.

otx/algorithms/classification/adapters/mmcls/models/heads/contrastive_head.py
sungchul1: Done.

otx/algorithms/classification/configs/efficientnet_b0_cls_warmstart/model.py
sungchul1: Done.

otx/algorithms/common/tasks/training_base.py
sungchul1: Because warmstart use dataset without label, we don't need to set classes (it is unnecessary.).
But these classes will be ignored when setting `SelfSLDataset`, it's okay passing it.
If you think it is better to remove if statement, I'll fix it.

sungchul1:
@jaegukhyun Thanks for reviewing.
1. It would be better to decide it after discussion.
2. Self-sl consumes lots of resources (time, memory, ...) as you mentioned, but warmstart is conducted with small epochs (e.g. 10 epochs) and it doesn't take much time than we expected. But I have a concern about different input size between experiment and this PR (I mentioned in https://github.com/openvinotoolkit/model_preparation_algorithm/pull/108).
3. I agree. If there is better ways to be scalable we can apply them.

otx/algorithms/common/tasks/training_base.py
sungchul1: I removed this block for compatibility with warmstart for semantic segmentation.

otx/algorithms/classification/tasks/inference.py
```diff
@@ -306,16 +311,21 @@ def _init_recipe_hparam(self) -> dict:
...
             optimizer=ConfigDict(lr=self._hyperparams.learning_parameters.learning_rate),
             lr_config=lr_config,
             early_stop=early_stop,
             data=ConfigDict(
                 samples_per_gpu=int(self._hyperparams.learning_parameters.batch_size),
                 workers_per_gpu=int(self._hyperparams.learning_parameters.num_workers),
             ),
-            runner=ConfigDict(max_epochs=int(self._hyperparams.learning_parameters.num_iters)),
         )
+        if self._warmstart:
```
sungchul1: set specific runner

otx/algorithms/classification/tasks/inference.py
```diff
@@ -345,6 +355,9 @@ def _init_recipe(self):
         if train_type == TrainType.INCREMENTAL:
             recipe = os.path.join(recipe_root, "incremental.yaml")
 
+        if train_type == TrainType.SELFSUPERVISED:
+            recipe = os.path.join(recipe_root, "warmstart.yaml")
```
sungchul1: warmstart -> selfsl?

otx/algorithms/classification/adapters/mmcls/models/pretrainers/byol.py
```diff
@@ -0,0 +1,196 @@
...
+        backbone (dict): Config dict for module of backbone ConvNet.
+        neck (dict): Config dict for module of deep features to compact feature vectors.
+            Default: None.
+        head (dict): Config dict for module of loss functions. Default: None.
+        pretrained (str, optional): Path to pre-trained weights. Default: None.
+        base_momentum (float): The base momentum coefficient for the target network.
+            Default: 0.996.
+    """
+
+    def __init__(
```
sungchul1, line 38: use state_dict_hook

otx/algorithms/common/adapters/mmcv/hooks.py
```diff
@@ -567,3 +568,78 @@ def after_train_iter(self, runner: BaseRunner):
         if isnan(runner.outputs["loss"].item()):
             logger.warning("Early Stopping since loss is NaN")
             runner.should_stop = True
+
+
+@HOOKS.register_module()
+class MomentumUpdateHook(Hook):
```
sungchul1: -> EMAMomentumUpdateHook

tests/integration/cli/classification/test_classification.py
```diff
@@ -33,6 +34,18 @@
 )
 from tests.test_suite.e2e_test_system import e2e_pytest_component
 
+# Warmstart using data w/ 'intel', 'openvino', 'opencv' classes
+args_w = {
+    "--train-data-roots": "data/text_recognition/IL_data",
+    "train_params": [
+        "params",
+        "--learning_parameters.num_iters",
+        "100",
```
sungchul1, line 43: -> 10

tests/integration/cli/classification/test_classification.py
```diff
@@ -94,8 +107,42 @@ def tmp_dir_path():
...
+        for mode in ["train", "val", "unlabeled"]:
+            to_save_data_args["data"][mode] = {}
+            for property in ["ann-files", "data-roots"]:
+                key = f"--{mode}-{property}"
+                if key == "--train-data-roots":
+                    to_save_data_args["data"][mode][property] = args_w[key]
+                else:
+                    to_save_data_args["data"][mode][property] = None
+        yaml.dump(to_save_data_args, open("./data.yaml", "w"), default_flow_style=False)
+        args_w["--data"] = "./data.yaml"
```
sungchul1, line 138: - remove data.yaml after test

wonjuleee (approved):
I think that the only the part using pipeline1 and pipeline2 would be confusing and others are fine. how about specifying the view0 and view1 as other Self-SLs do?

sungchul2:
> I think that the only the part using pipeline1 and pipeline2 would be confusing and others are fine. how about specifying the view0 and view1 as other Self-SLs do?

PR #1443, opened 2022-12-15
[OTX] Make train type configurable in OTX CLI

supersoob:
This PR will make train type configurable in OTX CLI train. It is required to integrate semi and self recipes.

command example)
`otx train {template} {data args} params --algo_backend.train_type=SEMISUPERVISED`

### DIFF
--- a/otx/algorithms/classification/configs/configuration.yaml
+++ b/otx/algorithms/classification/configs/configuration.yaml
@@ -299,7 +299,7 @@ algo_backend:
   train_type:
     affects_outcome_of: TRAINING
     default_value: INCREMENTAL
-    description: Training schema for the MPA task
+    description: Training scheme option that determines how to train the model
     editable: True
     enum_name: TrainType
     header: train type
--- a/otx/algorithms/classification/configs/configuration.yaml
+++ b/otx/algorithms/classification/configs/configuration.yaml
@@ -297,7 +297,7 @@ algo_backend:
   description: parameters for algo backend
   header: Algo backend parameters
   train_type:
-    affects_outcome_of: NONE
+    affects_outcome_of: TRAINING
     default_value: INCREMENTAL
     description: Quantization preset that defines quantization scheme
     editable: false
@@ -305,6 +305,7 @@ algo_backend:
     header: train type
     options:
       INCREMENTAL: "INCREMENTAL"
+      SEMISUPEVISED: "SEMISUPERVISED"
     type: SELECTABLE
     ui_rules:
       action: DISABLE_EDITING
--- a/otx/algorithms/classification/configs/configuration.yaml
+++ b/otx/algorithms/classification/configs/configuration.yaml
@@ -299,8 +299,8 @@ algo_backend:
   train_type:
     affects_outcome_of: TRAINING
     default_value: INCREMENTAL
-    description: Quantization preset that defines quantization scheme
-    editable: false
+    description: Training schema for the MPA task
+    editable: True
     enum_name: TrainType
     header: train type
     options:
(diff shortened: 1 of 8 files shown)


goodsong81 (changes requested):
Suggestion for the description.

- Do not use `MPA` term in general from OTX era
- schema -> scheme or type

ex) Training scheme option that determines how to train the model

goodsong81 (dismissed):
Not intended as-is application, but thank you for the update :)
LGTM.

jaegukhyun (dismissed):
LGTM, Thank you for fast modification

jaegukhyun (approved):

JihwanEom (approved):

sungmanc (approved):
LGTM

otx/algorithms/classification/configs/configuration.yaml
```diff
@@ -297,14 +297,15 @@ algo_backend:
   description: parameters for algo backend
   header: Algo backend parameters
   train_type:
-    affects_outcome_of: NONE
+    affects_outcome_of: TRAINING
     default_value: INCREMENTAL
-    description: Quantization preset that defines quantization scheme
-    editable: false
```
sovrasov, line 303: Usually T/F values start with non-capital letter in these config yamls. I'm not sure if the parser of these files in geti case sensitive or not, but we can keep such values as non-capital at least for consittency.
Also, I'd like to point out that the yaml files normally are auto-generated from the BaseConfig class  (otx/algorithms/common/configs/training_base.py) and we don't need to modify them manually, but I forgot where is located the corresponding script in geti repo, so we can skip that step. Anyway, BaseConfig class should be modified accordingly (see section BaseAlgoBackendParameters).

sovrasov (changes requested):
https://github.com/openvinotoolkit/training_extensions/blob/feature/otx/otx/algorithms/common/configs/training_base.py#L256 should be modified as well

otx/algorithms/common/configs/training_base.py
```diff
@@ -255,7 +255,7 @@ class BaseAlgoBackendParameters(ParameterGroup):
         train_type = selectable(
             default_value=TrainType.INCREMENTAL,
             header="train type",
-            description="training schema for the MPA task",
+            description="Training scheme option that determines how to train the model",
             editable=False,
```
sovrasov, line 259: Should this option be true? There is inconsistency with the yaml files.

otx/algorithms/classification/configs/configuration.yaml
supersoob: I know that we can revise it from baseconfig but some tasks like action does not support semi or other mode. I don't think that we need to modify the baseconfig, since we are adding the train type for each task one by one. And capital or non-capital doesn't matter, which is not consistent already in same config file.

sovrasov: Basically there should be no mess between python - yaml configs. If some parameters have different options for different tasks means that we should move them down in the python configs hierarchy. This may not be important for CLI, but in case of geti all the configs should be aligned. Since we don't use this branch in geti, let's merge it as is.
@supersoob could you provide the output of `otx train otx/algorithms/classification/configs/efficientnet_b0_cls_incr/template.yaml params -h`?

sovrasov (approved):

otx/algorithms/classification/configs/configuration.yaml
supersoob: Great. You made me think about geti side. It still requires remaining action task to be use only incremental so inconsistency is inevitable as of now. About inconsistency and capital aligning, I will write down it on my refactoring plan backlog. This is the output of the commands you gave.
image
image

PR #1451, opened 2022-12-19
[OTX/Datumaro] Enable OTX Action task using Datumaro dataset

jaegukhyun:
This PR is for linking Datumaro dataset and OTX action training backend


otx/algorithms/action/adapters/mmaction/data/dataset.py
```diff
@@ -44,9 +45,36 @@ def __init__(self, otx_dataset, labels):
...
-            return len(self.otx_dataset)
+            return len(self.video_info)
+
+        def _update_meta_data(self):
+            """Update video metadata of each item in self.otx_dataset."""
+            video_info = {}
+            start_index = 0
+            for idx, item in enumerate(self.otx_dataset):
+                metadata = item.get_metadata()[0].data
+                if metadata.video_id in video_info:
```
sungmanc: How about adding a detailed doc string on this for easy understanding ?

i.e. 
```
# Check whether the current item's video_id is already included or not.
# If, included,  ...
# Else, ...
```

sungmanc: To be honest, understanding below codes is little bit hard.

otx/algorithms/action/adapters/mmaction/data/dataset.py
```diff
@@ -44,9 +45,36 @@ def __init__(self, otx_dataset, labels):
...
+                    video_info[metadata.video_id]["total_frames"] += 1
```
sungmanc: What is the purpose of this code?

otx/algorithms/action/adapters/mmaction/data/dataset.py
```diff
@@ -44,9 +45,36 @@ def __init__(self, otx_dataset, labels):
...
+                    video_info[metadata.video_id]["start_index"] = start_index
+                else:
+                    if len(item.get_annotations()) > 0:
+                        label = int(item.get_roi_labels(self.labels)[0].id)
+                    else:
+                        label = None
+                    ignored_labels = np.array([self.label_idx[lbs.id] for lbs in item.ignored_labels])
+                    video_info[metadata.video_id] = {
+                        "total_frames": 1,
```
sungmanc: Is there no problem to set the total_frames as 1?

otx/algorithms/action/adapters/mmaction/data/dataset.py
```diff
@@ -119,83 +135,6 @@ def prepare_test_frames(self, idx: int) -> dict:
         :param idx: int, Index of data.
         :return dict: Testing data after pipeline with new keys introduced by pipeline.
         """
-        item = copy(self.data_infos[idx])  # Copying dict(), not contents
-        self.pre_pipeline(item)
+        item = copy(self.video_infos[idx])  # Copying dict(), not contents
+        item["modality"] = self.modality
```
sungmanc: Maybe not related to this PR, however, using `is_rgb`? instead of `modality`? 
For the first time when I saw the `modality`, I assumed that this variable is related to `video` or `image`.

otx/algorithms/action/adapters/mmaction/data/det_dataset.py
```diff
@@ -39,7 +41,7 @@
 root_logger = get_root_logger()
 
 
-# pylint: disable=too-many-instance-attributes
+# pylint: disable=too-many-instance-attributes, too-many-arguments, invalid-name, super-init-not-called, too-many-locals
```
sungmanc: In my opinion, disabling `super-init-not-called` is not appropriate.

otx/algorithms/action/adapters/mmaction/data/dataset.py
```diff
@@ -44,9 +45,36 @@ def __init__(self, otx_dataset, labels):
             self.otx_dataset = otx_dataset
```
sungmanc: I prefer to use type-hint i.e. ```def __init__(self, otx_dataset: DatasetEntity, labels: Labels)```

otx/algorithms/action/adapters/mmaction/data/det_dataset.py
```diff
@@ -50,36 +52,132 @@ class OTXActionDetDataset(AVADataset):
...
     class _DataInfoProxy:
-        def __init__(self, otx_dataset, labels):
-            self.otx_dataset = otx_dataset
+        def __init__(self, otx_dataset, labels, fps, test_mode):
+            self.otx_dataset = copy.deepcopy(otx_dataset)
             self.labels = labels
             self.label_idx = {label.id: i for i, label in enumerate(labels)}
+            self.fps = fps
+            self.data_root = "/" + os.path.join(
+                *os.path.abspath(self.otx_dataset[0].media._Image__file_path).split("/")[:-4]
```
sungmanc: Can't we use `self.otx_dataset[0].media.path` instead of `self.otx_dataset[0].media._Image__file_path)?

sungmanc: Is there no problem to use `self.data_root = "/" + ...`  ? in this case, all images should be located under the `/` ?

otx/algorithms/action/adapters/mmaction/data/det_dataset.py
```diff
@@ -50,36 +52,132 @@ class OTXActionDetDataset(AVADataset):
...
+                metadata = item.get_metadata()[0].data
+                if metadata.video_id in video_info:
+                    video_info[metadata.video_id]["start_index"] = start_index
+                    if metadata.frame_idx < video_info[metadata.video_id]["timestamp_start"]:
+                        video_info[metadata.video_id]["timestamp_start"] = metadata.frame_idx
+                    if metadata.frame_idx > video_info[metadata.video_id]["timestamp_end"]:
+                        video_info[metadata.video_id]["timestamp_end"] = metadata.frame_idx
+                else:
+                    video_info[metadata.video_id] = {
+                        "start_index": idx,
```
sungmanc, line 117: "start_index" is not related to frame of video? just for indexing of item?

otx/algorithms/action/adapters/mmaction/data/det_dataset.py
```diff
@@ -50,36 +52,132 @@ class OTXActionDetDataset(AVADataset):
...
+                        "timestamp_start": metadata.frame_idx,
+                        "timestamp_end": metadata.frame_idx,
+                    }
+                    start_index = idx
+
+            remove_indices = []
+            for idx, item in enumerate(self.otx_dataset):
```
sungmanc, line 124: Could you describe the below code block for my understanding?

otx/algorithms/action/adapters/mmaction/data/dataset.py
```diff
@@ -15,13 +15,14 @@
 # and limitations under the License.
 
 from copy import copy
-from typing import Any, Dict, List, Sequence
+from typing import List, Sequence
```
sungmanc: How about changing this filename to `cls_dataset.py`?

sungmanc (commented):
I left some comments, we need to add a docstring to complex code and type-hint to each class.

otx/algorithms/action/adapters/mmaction/data/dataset.py
jaegukhyun: Docstring is added

otx/algorithms/action/adapters/mmaction/data/dataset.py
jaegukhyun: This code is for write total frame of the video, it assumes that DatasetItemEntities are sorted by video and frame index

otx/algorithms/action/adapters/mmaction/data/dataset.py
jaegukhyun: Yes it is because DatasetItemEntities are sorted by video and frame index

otx/algorithms/action/adapters/mmaction/data/dataset.py
jaegukhyun: Modality means whether this data is from rgb image or optical flow. I added explanation

otx/algorithms/action/adapters/mmaction/data/det_dataset.py
jaegukhyun: Unfortunately, OTXActionDetDataset is very different from AVADataset in mmaction2 therefore, it can't call super.__init__, It can be possible to remove inheritance of AVADataset, but I think it is beyond of this PR. BTW there are several super-init-not-called disable in current otx branch

otx/algorithms/action/adapters/mmaction/data/det_dataset.py
jaegukhyun: There isn't path method for Image entity, therefore I added it.

otx/algorithms/action/adapters/mmaction/data/dataset.py
jaegukhyun: Type hint is added

otx/algorithms/action/adapters/mmaction/data/det_dataset.py
jaegukhyun: start_index has same purpose with action classification. It is offset of the video

otx/algorithms/action/adapters/mmaction/data/det_dataset.py
jaegukhyun: I added docstring of this function

otx/algorithms/action/adapters/mmaction/data/dataset.py
jaegukhyun: Changed!

otx/algorithms/action/adapters/mmaction/data/det_dataset.py
sungmanc: I see, thanks for comment

otx/algorithms/action/tasks/inference.py
```diff
@@ -296,6 +295,14 @@ def _get_metric(self, output_resultset):
             return MetricsHelper.compute_f_measure(output_resultset)
         raise NotImplementedError(f"{self._task_type} is not supported in action task")
 
+    def remove_empty_frames(self, dataset):
```
sungmanc: Some functions still remain without typehint.

sungmanc (changes requested):
In my understandingd, we don't have strict rules for type-hint. However, I think we should prepare type-hint for all functions. What do you think about it?

+Could you share the local or CI test results?

otx/algorithms/action/tasks/inference.py
jaegukhyun: Given that restricted time, I think all function would be impossible, but I'll add typehint as much as I can.
And I think local test for action task is enough for this PR, what do you think?

sungmanc: > Given that restricted time, I think all function would be impossible, but I'll add typehint as much as I can. And I think local test for action task is enough for this PR, what do you think?

Agree with your opinion, we need to continuously refactor for readability. 
Local test for action tasks (cls, det) is enough + pre-commit tests

jaegukhyun: Cool, I wll update local test results after adding type hint.

jaegukhyun:
ACTION_CLASSIFICATION
image
ACTION_DETECTION
image

Here's local test of this PR(only action part)

sungmanc (approved):
LGTM, thanks for the effort

PR #1452, opened 2022-12-19
[OTX/Anomaly] Add changes from external to otx

ashwinvaidya17:
# Description

- Adds changes from external/anomaly to otx/algorithms.
- Note: the pre-commit tests are failing on this branch. But I am also getting these errors on the target branch. I think the CI is working now. Maybe we can update the target branch with fixes for pre-commit. I'll update this branch again after that.


otx/algorithms/anomaly/tasks/inference.py
```diff
@@ -30,6 +30,16 @@
...
+from ote_sdk.entities.datasets import DatasetEntity
+from ote_sdk.entities.inference_parameters import InferenceParameters
+from ote_sdk.entities.metrics import NullPerformance, Performance, ScoreMetric
+from ote_sdk.entities.model import (
+    ModelEntity,
+    ModelFormat,
+    ModelOptimizationType,
+    ModelPrecision,
+    OptimizationMethod,
+)
```
goodsong81: This might be mistake :)

otx/algorithms/anomaly/tasks/train.py
```diff
@@ -14,25 +14,24 @@
...
+from adapters.anomalib.logger import get_logger
+from anomalib.models import AnomalyModule, get_model
 from anomalib.utils.callbacks import (
     MetricsConfigurationCallback,
     MinMaxNormalizationCallback,
 )
+from ote_sdk.entities.datasets import DatasetEntity
+from ote_sdk.entities.model import ModelEntity
+from ote_sdk.entities.train_parameters import TrainParameters
+from ote_sdk.usecases.tasks.interfaces.training_interface import ITrainingTask
```
goodsong81: Same here.

otx/algorithms/anomaly/tasks/train.py
```diff
@@ -68,11 +66,10 @@ def train(
 
         logger.info("Training Configs '%s'", config)
 
-        datamodule = OTXAnomalyDataModule(config=config, dataset=dataset, task_type=self.task_type)
+        datamodule = OTEAnomalyDataModule(config=config, dataset=dataset, task_type=self.task_type)
```
goodsong81: -> OTX

goodsong81 (changes requested):
Thank you for the PR. Could you revise a little bit?
Don't mind the code check. I've fixed them. (You may want to rebase)

ashwinvaidya17:
Anomaly tests 
image

ashwinvaidya17:
@goodsong81 thanks for the review! I missed those. I've updated the PR.

goodsong81 (approved):
LGTM, thank you!
Would the DRAEM be dealt with separately?
Or, no need to be applied to feature/otx?

otx/algorithms/anomaly/configs/base/configuration.py
```diff
@@ -48,6 +49,16 @@ class LearningParameters(ParameterGroup):
         header = string_attribute("Learning Parameters")
         description = header
 
+        # Editable is set to false as WideResNet50 is very large for
+        # onnx's protobuf (2gb) limit. This ends up crashing the export.
+        backbone = selectable(
```
djdameln: if the configurable parameters are changed, should the `configuration.yaml` files be updated as well?

djdameln (commented):
I have the same question as Songki, did you take DRAEM into consideration? 

By adding the `backbone` parameter to the base config, that parameter is now exposed to all models, including DRAEM. But DRAEM does not support different backbones, so this parameter should not be exposed.

otx/algorithms/anomaly/configs/base/configuration.py
ashwinvaidya17: I think I was looking at a different PR for the diff. I have removed backbone from configuration.py and added the changes for nncf in configuration.py. But apart from that I don't think there are other changes between external and otx for DRAEM.

otx/algorithms/anomaly/configs/base/configuration.py
```diff
@@ -19,6 +19,7 @@
 from attr import attrs
 
 from otx.algorithms.anomaly.configs.base.configuration_enums import (
+    ModelBackbone,
```
djdameln: I guess we don't need to import this if it's not used

goodsong81: I'll take care. Thanks!

PR #1457, opened 2022-12-20
[OTX] Add warmstart for classification

sungchul1:
Continued at https://github.com/openvinotoolkit/training_extensions/pull/1438

**FYI**
Self-SL doesn't use early stopping, but otx uses it as its default.
Therefore, a temporary condition was added to avoid setting early stopping at [link](https://github.com/openvinotoolkit/training_extensions/blob/feature/otx-warmstart-cls/otx/algorithms/common/tasks/training_base.py#L238-L241).


sungchul1:
Update: temp condition for managing early stopping is not working on other tasks (anomaly, action, ...) because there is no selfsl task. It is updated to consider `self._hyperparams.algo_backend.train_type` instead of `self._selfsl`. ([link](https://github.com/openvinotoolkit/training_extensions/blob/feature/otx-warmstart-cls/otx/algorithms/common/tasks/training_base.py#L238-L241))

otx/algorithms/common/adapters/mmcv/hooks.py
```diff
@@ -567,3 +568,78 @@ def after_train_iter(self, runner: BaseRunner):
         if isnan(runner.outputs["loss"].item()):
             logger.warning("Early Stopping since loss is NaN")
             runner.should_stop = True
+
+
+@HOOKS.register_module()
+class EMAMomentumUpdateHook(Hook):
```
harimkang, line 574: Putting all the hook classes in hooks.py seems to make the file lengthy and uncomfortable to look at. How about creating a subfolder called `adapters/mmcv/hook` to separate them?

harimkang (dismissed):
It is difficult to confirm because the CI is not normal, but if it is confirmed that only the test passes well, the rest looks good. Some variable names are confusing, but I think it is not a problem of this PR, but a problem of commonly used variable names.
Thank you for your hard work!

otx/algorithms/classification/tasks/inference.py
```diff
@@ -69,8 +69,12 @@ def __init__(self, task_environment: TaskEnvironment):
         else:
             self._labels = task_environment.get_labels(include_empty=False)
         self._empty_label = get_empty_label(task_environment.label_schema)
+
+        self.task_model_dir = None
+        self.task_pipeline_path = None
         self._multilabel = False
         self._hierarchical = False
+        self._selfsl = False
```
harimkang, line 77: Wouldn't this variable be commonly used by other tasks? If we need to use other tasks as well, it would be nice to declare them in Common's `BaseTask`.

otx/algorithms/classification/tasks/inference.py
```diff
@@ -339,14 +353,19 @@ def _init_recipe(self):
         if train_type == TrainType.INCREMENTAL:
             recipe = os.path.join(recipe_root, "incremental.yaml")
 
+        if train_type == TrainType.SELFSUPERVISED:
+            recipe = os.path.join(recipe_root, "selfsl.yaml")
+            self.task_model_dir = os.path.join(self.task_model_dir, "selfsl")
+            self.task_pipeline_path = os.path.join(os.path.dirname(self.task_pipeline_path), "selfsl/data_pipeline.py")
```
harimkang, line 359: Is there any reason why this variable shouldn't use `data_pipeline_path` as it is?

otx/algorithms/common/adapters/mmcv/hooks.py
sungchul1: I agree. I think we have to consider the situation that additional hooks (or other adapters) will be saved.

otx/algorithms/classification/tasks/inference.py
sungchul1: Actually, that variable is used several times in classification, but just once in segmentation.
Currently, I think it is only for classification.

otx/algorithms/classification/tasks/inference.py
sungchul1: I reflected [this commit](https://github.com/openvinotoolkit/training_extensions/pull/1425/commits/68c7d2ebcd271e6e63c54b4e323216a6cd9d5521) and haven't reflected the recent commits for those variable names.
I'll update them at segmentation task.

otx/algorithms/classification/adapters/mmcls/data/pipelines.py
```diff
@@ -62,3 +64,76 @@ def __call__(self, results: Dict[str, Any]):
...
+
+    Use this instead of mmcls's because there is no `hue` parameter in mmcls ColorJitter.
+    """
+
+    def __call__(self, results):
+        """Callback function of OTXColorJitter.
+
+        :param results: Inputs to be transformed.
+        """
+        results["img"] = np.array(self.forward(Image.fromarray(results["img"])))
```
sungmanc, line 107: AFAIK, we agreed to avoid this kind of duplication (numpy --> PIL --> numpy). do you have a plan to avoid it?

otx/algorithms/common/tasks/training_base.py
```diff
@@ -234,7 +235,10 @@ def _initialize(self, export=False):
         else:
             self._recipe_cfg.pop("adaptive_validation_interval", None)
 
-        self.set_early_stopping_hook()
+        # TODO (sungchul): it will be removed after an update that applies hparam.yaml
```
sungmanc, line 238: Maybe in the current status, hparam.yaml is applied. How about deleting?

tests/integration/cli/classification/test_classification.py
```diff
@@ -519,3 +521,58 @@ def test_pot_optimize(self, template, tmp_dir_path):
     @pytest.mark.parametrize("template", templates, ids=templates_ids)
     def test_pot_eval(self, template, tmp_dir_path):
         pot_eval_testing(template, tmp_dir_path, otx_dir, args_h)
+
+
+# tmp: create & remove data.yaml to only use train-data-roots
```
sungmanc, line 526: If this temporary function is needed to test `self-sl` task, how about making separate `test_self.py`? not included in `test_classification.py`.

sungmanc (commented):
Thanks for PR, I just left a few comments

tests/integration/cli/classification/test_classification.py
jaegukhyun: For the semi-supervised learning, we agreed on one test file for one task, however if self-sl has many difference with other tasks, then it's okay to make separate test file

jaegukhyun (dismissed):
Generally LGTM, but there are some codes to refactor. I think you know that enough, so I expect you to make another PR for refactoring whole selfsl tasks(cls, seg)

otx/algorithms/classification/adapters/mmcls/data/pipelines.py
sungchul1: It also remains on segmentation task. As @jaegukhyun said, It would be better to be progressed with another PR.

otx/algorithms/common/tasks/training_base.py
sungchul1: Could you explain where hparam.yaml is applied?

tests/integration/cli/classification/test_classification.py
sungchul1: I think there is no need to create a separate test file because of this temp function.
It will be removed when single argument (e.g. --train-data-roots) can be used.

otx/algorithms/classification/adapters/mmcls/models/pretrainers/byol.py
```diff
@@ -0,0 +1,210 @@
...
+import torch
+import torch.distributed as dist
+from mmcls.models.builder import CLASSIFIERS, build_backbone, build_head, build_neck
+from mpa.utils.logger import get_logger
+from torch import nn
+
+logger = get_logger()
+
+
+@CLASSIFIERS.register_module()
```
supersoob: Can we revise the base folder name as classifiers rather than pretrainers? Otherwise, do you intend to name it?

otx/algorithms/classification/configs/configuration.yaml
```diff
@@ -33,7 +33,7 @@ learning_parameters:
       make it unstable.
     editable: true
     header: Learning rate
-    max_value: 0.1
+    max_value: 1.0
```
supersoob, line 36: Doesn't it affect the training lr scheduler not only for self-sl but other train type?

otx/algorithms/classification/adapters/mmcls/models/pretrainers/byol.py
sungchul1: I newly named it as pretrainer because it doesn't look like classifier.
But if it could be against policy, I'll modify it.

otx/algorithms/classification/tasks/inference.py
supersoob: Yeah, when I went over this again, task_ is a little vague, because it can make a confusion like task_type (detection/classificaition...). I recommend to use model_dir and data_pipeline_path.

otx/algorithms/common/tasks/training_base.py
sungmanc: Ah then, this file is not used ?
image

otx/algorithms/classification/adapters/mmcls/data/pipelines.py
sungmanc: Why do we move on to another PR for the refactoring task? @jaegukhyun

otx/cli/tools/train.py
```diff
@@ -207,20 +207,21 @@ def main():
         args.save_model_to = "./models"
     save_model_data(output_model, args.save_model_to)
 
-    validation_dataset = dataset.get_subset(Subset.VALIDATION)
-    predicted_validation_dataset = task.infer(
-        validation_dataset.with_empty_annotations(),
-        InferenceParameters(is_evaluation=False),
-    )
+    if data_config["data"]["val"]["data-roots"]:  # TODO (sungchul): refactoring
```
supersoob: Is todo resolved, or not? If resolved please remove comments.

otx/algorithms/common/tasks/training_base.py
jaegukhyun: hparam.yaml is used when otx builder build workspace. Builder will modify base template, which is oriented to incremental learning, to match own training type

tests/integration/cli/classification/test_classification.py
sungmanc: I prefer that temporary variables or functions should not be included in the existing one even though they will be deleted. However, it can be just my preference. So, if you don't agree, please ignore this. I want to make test_**.py only include the test cases. not samples

otx/algorithms/classification/adapters/mmcls/data/pipelines.py
sungchul1: I don't think how they are refactored.
I think we can add PR as 'optimize data pipelines inherited from torchvision' and discuss it there.

otx/algorithms/common/tasks/training_base.py
supersoob: Yeah. It is not used as of now. That otx builder which is not currently supported fully will adaptively overwrite the config from hparam.yaml. I said to Sungchul to leave configs as is, expecting them to process when otx build is supported. But seems like just unsetting the early stopping is needed. I think it's okay to good to leave this codes.

otx/algorithms/classification/adapters/mmcls/data/pipelines.py
jaegukhyun: As we have a limited time for feature freeze, and this refactoring task will be changed after merging segmentation. 
I think it would be better to discuss how we can make a better code after merging new feature PRs(self-sl-cls, self-sl-seg).
However, if you think it is necessary changes, then it should be handled in this PR

otx/algorithms/common/tasks/training_base.py
sungmanc: If so, hparam.yaml is already applied to change the template. So, we can delete the (TODO) tag

supersoob: Ah I think TODO tag is needed because hparam support to disable early stopping too. Self-sl needs to ensure to run some training epochs so I added comments to leave as is.

jaegukhyun: And this should be revisited. If we move self.train_type, which is in the each algo's inferencer, to this training_base, we can use it in this block

otx/algorithms/classification/configs/configuration.yaml
supersoob: That was my bad. I realized it just max value for ranging learning rate.

jaegukhyun: As we use lr scheduler to reduce lr, max_value might be fine, but it is needed to check lr scheduler algorithm in the mpa.

otx/algorithms/common/tasks/training_base.py
sungmanc: Maybe I missed something..! Okay then leave it as is

otx/algorithms/classification/configs/configuration.yaml
sungmanc: Since this is just `max_value`. it couldn't affect to others

otx/algorithms/classification/adapters/mmcls/data/pipelines.py
sungmanc: Didn't we have implemented a function to change this line? I'm aware of there was implemented function. If it didn't exist, I agreed with your opinions. 

+ just wondering: Self-SL cls/seg feature is included in OTXv1.0 feature freeze?

supersoob: Agree with jaeguk. Adding to this, it seems like only pipelines using for self-sl just registered here, while others in mpa transform files. I think we better discuss later.

otx/cli/tools/train.py
sungchul1: We decided to leave this part alone. I'll remove the comment.

supersoob (dismissed):

sungmanc (dismissed):
Sorry for my mis-synced questions, LGTM

sungchul1:
@wonjuleee @harimkang @jaegukhyun @sungmanc @supersoob 
Sorry to ask for your approval again. I missed isort when updating `pretrainers` → `classifiers`.

Test was only failed at `tests/integration/cli/classification/test_classification.py::TestToolsMPAMultilabelClassification::test_otx_eval_openvino[Custom_Image_Classification_MobileNet-V3-large-1x]`. I think we can merge this PR, what do you think?

harimkang (approved):
I agree with you!

LGTM!

jaegukhyun (approved):

supersoob (approved):

PR #1460, opened 2022-12-22
[OTX] otx find & build with OMZ backbones + Refactor codes

harimkang:
## Summary
Implemented otx find & build for OMZ backbone and refactored existing find and build code.

### What this PR contains
- Find OMZ backbones
image
- Classification build with OMZ backbones
- Classification build with mmdet, mmseg, torchvision backbones
- Refactoring otx find & build codes


otx/cli/builder/supported_backbone/mmdet.json
```diff
@@ -14,36 +14,60 @@
           "regnetx_12gf"
         ]
       },
-      "available": ["ATSS", "YOLOX", "Segmentation", "Mask-RCNN"]
+      "available": [
```
sungmanc, line 17: How about unifying all JSON files into single file? (i.e. `otx/cli/builder/supported_backbones.json`)
I'm not sure whether the location of JSON files that has a dependency on frameworks is good or not since they are located under the `otx/cli` and `builder` that could be represented such a general term.

otx/cli/builder/builder.py
```diff
@@ -301,7 +275,15 @@ def build_model_config(self, model_config_path, backbone_config_path, output_pat
 
         # Build Backbone
         backbone = build_from_cfg(backbone_config, otx_registry, None)
-        out_channels = get_backbone_out_channels(backbone)
+        if model_config.model.get("task", None) == "classification":
```
sungmanc: How about converting all `string variables` to the `TaskType` class?
(i.e. "classification" --> "TaskType.CLASSIFICATION")

Then, we can represents all kinds of tasks supported by Geti and also can avoid hard-coded string format.

otx/cli/builder/builder.py
```diff
@@ -301,7 +275,15 @@ def build_model_config(self, model_config_path, backbone_config_path, output_pat
...
+            # Update model layer's in/out configuration in ClsStage.configure_model
+            out_channels = -1
+            if hasattr(model_config.model, "head"):
+                model_config.model.head.in_channels = -1
+        else:
```
sungmanc, line 284: Maybe, we need to consider Action tasks, in the near future.
So, I just prefer the `elif` sentence in here to handle the `detection` and `segmentation`

otx/algorithms/common/adapters/mmcv/models/backbones/torchvision_backbones.py
```diff
@@ -131,7 +131,7 @@ def multioutput_forward(self, x):
             print(temp_s)
         if i == last_stage:
             break
-    return outputs
+    return tuple(outputs)
```
sungmanc, line 134: Just wondering: can we handle all forward function implemented on torchvision backbones by using above 3 functions? (resnet, shufflenet, multioutput forward)

otx/cli/builder/builder.py
harimkang: Perhaps I could use something like `str(TaskType.CLASSIFICATION)` instead of `classification`. I'll fix it.

otx/cli/builder/supported_backbone/mmdet.json
harimkang: There are parts that I think this is also a design choice. Agreeing with some of the comments, the downside to keeping it in one file is that it doesn't have good visibility. One file becomes very long and I suspect that if there is a problem with that file it could affect the whole framework's backbone find and build. what do you think?

otx/cli/builder/builder.py
harimkang: yes, I think this is something to consider. I will test it and if it is not a problem, I will fix it with `elif` & `else` (raise Error).

otx/algorithms/common/adapters/mmcv/models/backbones/torchvision_backbones.py
harimkang: Most `torchvision backbones` use layers in `self.layers` or `self.features`. In such cases, it is patched to use `multioutput forward`, and resnet and shufflenet are configured slightly differently, so additional forward patching was required.

jaegukhyun (dismissed):
I have no big objection for this PR, it looks good to me. As otx build and otx find support all task, all training type, and even all training backends(mmX, torchvision, openvino), it will contain many if, else statements. Therefore, at some point, it will be needed to re-design builder.py for more scalable and simple codes.

harimkang:
> I have no big objection for this PR, it looks good to me. As otx build and otx find support all task, all training type, and even all training backends(mmX, torchvision, openvino), it will contain many if, else statements. Therefore, at some point, it will be needed to re-design builder.py for more scalable and simple codes.

Yes, this is a PR that includes a re-design to reduce the `if` and `else` statements you mentioned in the current state. Nevertheless, There may still be some areas that are not general, but I think we can gradually improve this. thanks for the comment :)

otx/cli/builder/builder.py
harimkang: As a result of checking this, there are cases where there is no task variable for each model. Looks like I need to align this later. So I think we should leave it as it is for now.

otx/cli/builder/supported_backbone/mmdet.json
sungmanc: Yeah, it can differ from person to person. Let's leave it as is.

JihwanEom (dismissed):
LGTM

otx/cli/utils/importing.py
```diff
@@ -43,91 +44,51 @@ def get_impl_class(impl_path):
...
     if os.path.exists(available_backbone_path):
         with open(available_backbone_path, "r", encoding="UTF-8") as f:
             available_backbones = json.load(f)
         available_backbones = available_backbones["backbones"]
-    elif backend in ("otx", "pytorchcv"):
-        backbone_list = get_backbone_list(backend)
-        registry, _ = get_backbone_registry(backend)
+    elif backend == "pytorchcv" and importlib.util.find_spec(backend):
+        backbone_list = importlib.import_module(f"{backend}.model_provider")._models
         backbone_format = {"required": [], "options": {}, "available": []}
```
eunwoosh, line 56: How about deleting this line?

otx/cli/utils/importing.py
```diff
@@ -43,91 +44,51 @@ def get_impl_class(impl_path):
...
-        torchvision_backbones = importlib.import_module(torchvision_backbone_module)
-        return [f"{backbone}" for backbone in torchvision_backbones.TORCHVISION_MODELS.keys()]
-    raise ValueError(f"{backend} cannot be imported.")
-
-
-def get_available_backbone_list(backend):
-    """Gather available backbone list from json file."""
+    """Gather available backbone list from json file & imported lib."""
     available_backbone_path = os.path.join(get_otx_root_path(), "cli/builder/supported_backbone")
     available_backbone_path = os.path.join(available_backbone_path, f"{backend}.json")
```
eunwoosh: How about combine this line with right above code as below.
available_backbone_path = os.path.join(get_otx_root_path(), "cli/builder/supported_backbone", f"{backend}.json")

otx/cli/utils/tests.py
```diff
@@ -562,57 +562,70 @@ def otx_find_testing():
...
+        "build",
+        "--model",
+        os.path.join(task_workspace, "model.py"),
+        "--backbone",
+        os.path.join(task_workspace, "backbone.yaml"),
+        "--workspace-root",
+        task_workspace,
+    ]
+    check_run(command_line)
+
```
eunwoosh, line 626: how about adding some asserts code to check whether code work as expected?

otx/cli/tools/find.py
```diff
@@ -49,15 +50,53 @@ def parse_args():
...
-        action="append",
-        help=f"The currently supported options: {SUPPORTED_BACKBONE_BACKENDS}.",
+        nargs="+",
+        help=f"The currently supported options: {list(SUPPORTED_BACKBONE_BACKENDS.keys())}.",
     )
 
     return parser.parse_args()
 
 
+def generate_backbone_rows(index, backbone_type, meta_data):
```
eunwoosh: How about adding type hinting?

otx/cli/utils/importing.py
harimkang: Your review has been reflected. thank you :)

otx/cli/utils/importing.py
harimkang: It's corrected.

otx/cli/utils/tests.py
harimkang: Added this.

otx/cli/tools/find.py
harimkang: Added this too.

eunwoosh (approved):
LGTM. It can be merged after checking Pre-merge test

JihwanEom (approved):

harimkang:
As a result of locally double-checking the failed cases in the CI test, it is expected that this PR is not a problem.

- Tiling Instance-Segmentation
image
- Segmentation
image

I will merge this.

PR #1466, opened 2022-12-26
[OTX] Introduce mmdeploy to export cls/seg/det models

cih9088:
### [Related MPA PR](https://github.com/openvinotoolkit/model_preparation_algorithm/pull/124)

**Changes**
- Update openvino version
- introduced `mmdeploy`
- refactored to be `mmdeploy` friendly 
- removed colour format change in openvino classification model


otx/algorithms/classification/adapters/mmcls/__init__.py
```diff
@@ -16,6 +16,11 @@
 
 from .data import MPAClsDataset
 
+# FIXME: openvino pot library adds stream handlers to root logger
+# which makes annoying duplicated logging
+from mmcls.utils import get_root_logger
+get_root_logger().propagate = False
```
goodsong81, line 22: Nice catch!

otx/algorithms/classification/adapters/mmcls/nncf/builder.py
```diff
@@ -155,6 +140,7 @@ def build_model(config: Config, from_scratch: bool = False):
             },
         )
     )
+    remove_from_configs_by_type(custom_hooks, "UnlabeledDataHook")
```
goodsong81: No NNCF support for Semi-SL?

otx/algorithms/classification/adapters/mmcls/utils/builder.py
```diff
@@ -0,0 +1,40 @@
...
+def build_classifier(
+    config: Config,
+    checkpoint: Optional[str] = None,
+    device: Union[str, torch.device] = "cpu",
+    cfg_options: Optional[Union[Config, ConfigDict]] = None,
+    from_scratch: bool = False,
+):
+    """Creates a model, based on the configuration in config.
+    Note that this function consumes/updates 'load_from' attribute of 'config'.
+    """
```
goodsong81, line 22: Similar functions in MPA PR?

otx/algorithms/classification/tasks/inference.py
```diff
@@ -160,15 +163,20 @@ def export(self, export_type: ExportType, output_model: ModelEntity):
         output_model.optimization_type = ModelOptimizationType.MO
 
         stage_module = "ClsExporter"
-        results = self._run_task(stage_module, mode="train", precision="FP32", export=True)
```
goodsong81, line 163: Where would we handle the precision? Actually, Geti is expecting OTE to support FP16 and/or FP32 according to configurable parameters.

otx/algorithms/common/adapters/mmcv/nncf/runners.py
```diff
@@ -119,7 +119,10 @@ def validation_fn(self, *args, **kwargs):
         # Get metric from runner's attributes that set in EvalHook.evaluate() function
         all_metrics = getattr(self, "all_metrics", {})
         if len(all_metrics) == 0:
-            return 0.0
+            evalhook = [hook for hook in self.hooks if getattr(hook, '_do_evaluate', None)]
+            assert len(evalhook) == 1
+            evalhook[0]._do_evaluate(self)
+            all_metrics = getattr(self, "all_metrics", {})
```
goodsong81, line 125: Is this for 'before run' evaluation?

otx/algorithms/common/tasks/training_base.py
```diff
@@ -304,6 +312,80 @@ def _init_recipe_hparam(self) -> dict:
...
+                # so that `to_rgb=True` means a format change to BGR instead.
+                # Conventionally, OpenVINO IR expects a image in BGR format
+                # but OpenVINO IR under OTX assumes a image in RGB format.
+                #
+                # `to_rgb=True` -> a model was trained with images in BGR format
+                #                  and a OpenVINO IR needs to reverse input format from RGB to BGR
+                # `to_rgb=False` -> a model was trained with images in RGB format
+                #                   and a OpenVINO IR does not need to do a reverse
+                if normalize_cfg.get("to_rgb", False):
+                    options["flags"] += ["--reverse_input_channels"]
```
goodsong81, line 342: I agree with this logic. But we need to check external circumstances.
@sungmanc could you double check? We may use to_rgb=False for Geti input (RGB numpy? need to check) and to_rgb=True for Datumaro input (BGR numpy, if using OpenCV). In this case, we may want to turn on --reverse_input_channels always as the model itself is always expecting RGB image.

otx/algorithms/common/tasks/training_base.py
```diff
@@ -304,6 +312,80 @@ def _init_recipe_hparam(self) -> dict:
...
+                # default is static shape to prevent an unexpected error
+                # when converting to OpenVINO IR
+                deploy_cfg.backend_config.model_inputs = [
+                    ConfigDict(opt_shapes=ConfigDict(input=[1, 3, *size]))
+                ]
+
+            patch_input_preprocessing(deploy_cfg)
+            if not deploy_cfg.backend_config.get("model_inputs", []):
+                patch_input_shape(deploy_cfg)
+
```
goodsong81, line 386: For now, IR input shape is statically defined in deployment.py? (or base_xxx)

otx/algorithms/detection/adapters/mmdet/nncf/patches.py
```diff
@@ -120,3 +121,20 @@ def wrap_register_module(self, fn, *args, **kwargs):
...
+    for fn_path, record_dicts in FUNCTION_REWRITER._registry._rewrite_records.items():
+        if fn_path.startswith("torch"):
+            continue
+        obj, obj_cls = import_function(fn_path)
+        fn_name = fn_path.split(".")[-1]
+        if should_wrap(obj_cls, fn_name, HEADS_TARGETS):
+            for record_dict in record_dicts:
+                record_dict["_object"] = partial(
+                    no_nncf_trace_wrapper, None, record_dict["_object"]
+                )
```
goodsong81, line 140: Could you explain this block for my understanding?
I'm not sure what this is doing.

otx/algorithms/detection/adapters/openvino/model_wrappers/openvino_models.py
```diff
@@ -95,3 +105,46 @@ def _get_outputs(self) -> Dict:
...
+    def find_layer_bboxes_output(layers):
+        filter_outputs = [name for name, data in layers.items() if len(data.shape) == 3 and data.shape[-1] == 5]
+        if not filter_outputs:
+            raise ValueError('Suitable output with bounding boxes is not found')
+        if len(filter_outputs) > 1:
+            raise ValueError('More than 1 candidate for output with bounding boxes.')
+        return filter_outputs[0]
+
+    def __call__(self, outputs):
+        # FIXME: here, batch dim of IR must be 1
```
goodsong81, line 131: Is this a restriction from our implementation?

otx/algorithms/detection/adapters/openvino/model_wrappers/openvino_models.py
```diff
@@ -95,3 +105,46 @@ def _get_outputs(self) -> Dict:
                     output_match_dict[output_name] = node_name
                     break
         return output_match_dict
+
+
+class BatchBoxesLabelsParser:
+    def __init__(self, layers, input_size, labels_layer='labels', default_label=0):
```
goodsong81, line 111: Could you let me know the purpose?

requirements/anomaly.txt
```diff
@@ -1,4 +1,4 @@
 # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
 # Anomaly Requirements.                                                       #
 anomalib==0.3.6
-pytorch-lightning>=1.6.0,<1.7.0
\ No newline at end of file
+pytorch-lightning>=1.6.0,<1.7.0
```
goodsong81, line 4: Any change?

requirements/openvino.txt
```diff
@@ -2,6 +2,6 @@
 # OpenVINO Requirements.                                                      #
 nncf==2.3.0
 onnx==1.10.1
-openmodelzoo-modelapi @ git+https://github.com/openvinotoolkit/open_model_zoo/@releases/2022/SCv1.1#egg=openmodelzoo-modelapi&subdirectory=demos/common/python
-openvino==2022.2.0
-openvino-dev==2022.2.0
\ No newline at end of file
+openmodelzoo-modelapi @ git+https://github.com/openvinotoolkit/open_model_zoo/@releases/2022/3#egg=openmodelzoo-modelapi&subdirectory=demos/common/python
```
goodsong81, line 5: For now, we are controlling the PyPI package of OMZ Model APIs. I'll build & upload 2022.3.0 to PyPI soon.

otx/algorithms/classification/adapters/openvino/model_wrappers/openvino_models.py
```diff
@@ -81,20 +81,6 @@ def _get_outputs(self):
...
-        resized_image = self.resize(inputs, (self.w, self.h))
-        resized_image = cv2.cvtColor(resized_image, cv2.COLOR_RGB2BGR)
-        meta.update({"resized_shape": resized_image.shape})
-        if self.resize_type == "fit_to_window":
-            resized_image = pad_image(resized_image, (self.w, self.h))
-        resized_image = self.input_transform(resized_image)
-        resized_image = self._change_layout(resized_image)
-        dict_inputs = {self.image_blob_name: resized_image}
-        return dict_inputs, meta
-
```
goodsong81, line 97: Not used?

goodsong81 (approved):
What a work! Thank you for your effort.
Happy to see the NNCF/OV tests are re-enabled :)
Left a few question & comments.
Happy new year!

otx/algorithms/classification/adapters/openvino/model_wrappers/openvino_models.py
cih9088: My intension was to remove image format conversion to `BGR` format. Without format conversion, this method is exactly [the same](https://github.com/openvinotoolkit/open_model_zoo/blob/master/demos/common/python/openvino/model_zoo/model_api/models/image_model.py#L112-L147) with one in `super`

requirements/anomaly.txt
cih9088: I added new line at the end of the files, which is a sort of [standard](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline).

otx/algorithms/classification/adapters/mmcls/utils/builder.py
cih9088: Yes, almost the same function exists in MPA and it was for MPA users who run things directly from MPA. If MPA is merged into OTX, we could discard it in MPA.

otx/algorithms/classification/tasks/inference.py
cih9088: The precision would be handled in [here](https://github.com/cih9088/training_extensions/blob/feature/otx-new-mmcv/mmdeploy/otx/algorithms/common/tasks/training_base.py#L263-L266). I doubt that Geti could configure precision as this logic always export `FP32` model.

otx/algorithms/common/adapters/mmcv/nncf/runners.py
cih9088: Yes, exactly. Note that evaluation hook does not implement the `before_run` method.

otx/algorithms/common/tasks/training_base.py
cih9088: It is hard coded in each model's `deployment.py`. This logic is to patch `input shape` from `data.test.pipeline.Resize` in case it is not defined in `deployment.py`.

otx/algorithms/detection/adapters/openvino/model_wrappers/openvino_models.py
cih9088: [`BoxesLabelParser` in OMZ](https://github.com/openvinotoolkit/open_model_zoo/blob/99a43f8bee1fbb4df27523ca1bf5993c66cf7a03/demos/common/python/openvino/model_zoo/model_api/models/ssd.py#L118-L150) assumes that no batch dimension exists. It was fine with the legacy export but it is not with `mmdepoy` export. `mmdeploy` patches internal NMS logic of `mmdet` to preserve batch dimension. NMS from `mmdet` [returns](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/core/post_processing/bbox_nms.py#L31-L33) `(num_det, 5)` whereas NMS from `mmdeploy` [returns](https://github.com/open-mmlab/mmdeploy/blob/master/mmdeploy/codebase/mmdet/core/post_processing/bbox_nms.py#L204-L206) `(batch, num_det, 5)`

otx/algorithms/detection/adapters/openvino/model_wrappers/openvino_models.py
cih9088: Yes. Even though I allowed the detection models [have dynamic batch size](https://github.com/cih9088/training_extensions/blob/feature/otx-new-mmcv/mmdeploy/otx/algorithms/detection/configs/detection/cspdarknet_yolox/deployment.py#L8), I just assumed batch size of model input is 1 as the legacy exports model with static batch size of 1.

otx/algorithms/detection/adapters/mmdet/nncf/patches.py
cih9088: When export the model with `mmdeploy`, `mmdeploy` patches model's methods as you know. If some methods, which are required to be excluded when tracing NNCF graph, are patched, we must wrap those methods with `no_nncf_trace_wrapper`  as well.

otx/algorithms/classification/adapters/mmcls/nncf/builder.py
cih9088: Thanks for catching it and it is fixed in https://github.com/openvinotoolkit/training_extensions/pull/1466/commits/77a62c7196e065b8ced9b62e00226471ac83dc80. Please note that I haven't tested it on other tasks beside cls/det/seg.

otx/algorithms/classification/tasks/inference.py
goodsong81: You're right. I just wanted know how to enable one day. :) Thank you.

otx/algorithms/common/tasks/training_base.py
goodsong81: I'm thinking of exposing input size options to user. This adaption might be helpful then. Thanks!

PR #1471, opened 2022-12-29
[Hot-fix][OTX-MPA-Integration] Import semi-sl trainer & fix valuable references in staticmethods

supersoob:
This PR is for fixing importing errors about finding semi-sl trainer and for fixing valuable references in static-methods in detection stages.


sungmanc (approved):
LGTM, Thanks for quick resolvation

PR #1472, opened 2022-12-30
[OTX] Evaluate a model before training starts

eunwoosh:
### Summary
Before training starts, validate model weight initial model weight. It's for case where loaded model weight or resumed more weight is better than trained model weight.

### This PR includes

- add a hook to execute validation step before training starts.

### DIFF
--- a/otx/mpa/modules/hooks/eval_before_train_hook.py
+++ b/otx/mpa/modules/hooks/eval_before_train_hook.py
@@ -1,0 +1,48 @@
+# Copyright (C) 2022 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
+
+from mmcv.runner import HOOKS
+from mmcv.runner import EvalHook as SegEvalHook
+from mmcv.runner import Hook
+from mmdet.core.evaluation.eval_hooks import EvalHook as DetEvalHook
+from mpa.modules.hooks.checkpoint_hook import CheckpointHookWithValResults
+from mpa.modules.hooks.eval_hook import CustomEvalHook as ClsEvalHook
+
+
+@HOOKS.register_module()
+class EvalBeforeTrainHook(Hook):
+    """Hook to evaluate and save model weight before training."""
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self._executed = False
+
+    def before_train_epoch(self, runner):
+        """Execute the evaluation hook before training"""
+        if not self._executed:
+            for hook in runner.hooks:
+                if self.check_eval_hook(hook):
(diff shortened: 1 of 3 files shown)


otx/mpa/modules/hooks/eval_before_train_hook.py
```diff
@@ -0,0 +1,56 @@
...
+
+@HOOKS.register_module()
+class EvalBeforeTrainHook(Hook):
+    """Hook to evaluate and save model weight before training."""
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self._executed = False
+
+    def before_train_epoch(self, runner):
```
supersoob, line 26: I see you execute the existing hook methods dynamically once before training. Isn't it okay to move this before_run()? And I just wonder whether it's gonna be more complicated if we do not make use of existing hook and write codes for it. Calling after_train_epoch of hook in before_train_epoch is hard to understand workflow.

eunwoosh: As far as I know, some function modifies the model in `before_run()`. So I use `before_train_epoch()`.
But I'll check that whether I can use `beore_run()`.
and for second question, the reason why I use exiting hook is that I want the initial model weight can be chosen if initial weight is best model weight. Each validation hook, which I use, validates model score and saves model weight if needed. So, I just want to use this way for initial model to be chosen as best weight naturally.

harimkang (approved):
Is there any way other than hooks? This PR itself doesn't seem to be a problem, but I think there may be other more efficient ways.

eunwoosh:
> Is there any way other than hooks? This PR itself doesn't seem to be a problem, but I think there may be other more efficient ways.

Thanks for your comment! Of course, there are other ways to implement it. but evaluation hook of each task keeps best score and save best weight if needed. So, I think this is best way to exploit it. With other ways, it should be done by itself validating initial model weight, compare with final weight and select better one.

otx/mpa/modules/hooks/eval_before_train_hook.py
eunwoosh: I checked whether before_run() also works well but I found that accuracies of consecutive epochs are exactly same.
I tried to find why it happens, but I didn't able to find a reason.
I think it's better to use before_train_epoch() and change it to before_run() if needed.

eunwoosh:
There is a single failed case and it is `TestToolsAnomalySegmentation.test_otx_eval_openvino[ote_anomaly_segmentation_stfpm]`
score difference is not quite big as below.
image

supersoob (approved):

jaegukhyun (approved):

PR #1473, opened 2023-01-02
[OTX] Recover semi-sl for multi-class classification 

JihwanEom:
**This PR includes**
- Add SeparateAug pipeline element to support Semi-SL for multi-class classification
- Refactor data_pipeline to improve readability
- Replaced dictionary-shaped pipeline with unlabeled data hook
- Support unlabeled_batch size control via CLI
- Make templates and configs folder for semi-sl
- Handle inference.py & train.py to support `SEMISUPERVISED` train_type
- Delete unnecessary files (`cls_data.py` and `semisl_pipeline.py`) in MPA
- Tests for multi-class Semi-SL


otx/algorithms/classification/configs/base/data/data_pipeline.py
```diff
@@ -17,29 +17,35 @@
...
 
-train_pipeline = [
-    dict(type="Resize", size=__img_size),
+__train_pipeline = [
+    dict(type="Resize", size=__resize_target_size),
     dict(type="RandomFlip", flip_prob=0.5, direction="horizontal"),
     dict(type="AugMixAugment", config_str="augmix-m5-w3"),
     dict(type="RandomRotate", p=0.35, angle=(-10, 10)),
-    dict(type="ToNumpy"),
+    dict(type="PILImageToNDArray", keys=["img"]),
```
harimkang, line 27: This pipeline is known as the supervised learning pipeline. Any reason to change?

otx/algorithms/classification/tasks/train.py
```diff
@@ -108,7 +108,7 @@ def train(
         if train_parameters is not None:
             update_progress_callback = train_parameters.update_progress  # type: ignore
         self._time_monitor = TrainingProgressCallback(update_progress_callback)
-        self._learning_curves = defaultdict(OTXLoggerHook.Curve)
+        self._learning_curves = defaultdict(OTXLoggerHook.Curve)  # type: ignore
```
harimkang: What does the `# type: ignore` mean?

otx/mpa/cls/stage.py
```diff
@@ -80,6 +80,21 @@ def is_mmov_model(k, v):
             cfg.merge_from_dict(data_cfg)
         self.configure_data(cfg, training, **kwargs)
 
+        if training:
+            if "unlabeled" in cfg.data and cfg.train_type == "SEMISUPERVISED":
+                update_or_add_custom_hook(
+                    cfg,
+                    ConfigDict(
+                        type="UnlabeledDataHook",
```
harimkang, line 88: What does `UnlabeledDataHook` do?

otx/mpa/cls/stage.py
```diff
@@ -80,6 +80,21 @@ def is_mmov_model(k, v):
...
+                        unlabeled_data_cfg=cfg.data.unlabeled,
+                        samples_per_gpu=cfg.data.unlabeled.pop("samples_per_gpu", cfg.data.samples_per_gpu),
```
harimkang, line 90: Is this batch size for `labeled sample`?

otx/mpa/modules/models/classifiers/semisl_classifier.py
```diff
@@ -81,27 +29,22 @@ def forward_train(self, imgs, **kwargs):
...
             raise ValueError("'extra_0' does not exist in the dataset")
-        target = kwargs["gt_label"]
+        target = kwargs["gt_label"].squeeze()
         unlabeled_data = kwargs["extra_0"]
-
         x = {}
         x["labeled"] = self.extract_feat(imgs)
 
-        img_uw = unlabeled_data["weak"]["img"]
+        img_uw = unlabeled_data["img_weak"]
```
harimkang: It's minor, but I think that this is an `img` and the one used for strong augmentation should be `img_strong`. Since `img_weak` is an image that eventually uses the same augmentation as the labeled image, so I think `img_strong` should be rather distinct.

otx/mpa/modules/models/heads/semisl_cls_head.py
```diff
@@ -136,3 +138,140 @@ def forward_train(self, x, gt_label):
...
+            # compute unsupervised loss
+            lu = self.compute_loss(logits_u_s, pseudo_label, avg_factor=len(logits_u_s)) * mask
+        losses["loss"] = lx + self.unlabeled_coef * lu
+
+        # compute accuracy
+        acc = self.compute_accuracy(logits_x, gt_label)
+        losses["accuracy"] = {f"top-{k}": a for k, a in zip(self.topk, acc)}
+        return losses
+
+    def forward_train(self, x, gt_label):
```
harimkang, line 156: Is there a difference between the corresponding functions (forward_train, etc..) for Linear and Nonlinear? If not, this is considered duplicate code.

tests/integration/cli/classification/test_classification.py
```diff
@@ -100,7 +100,7 @@ def tmp_dir_path():
     templates_ids = [template.model_template_id for template in templates]
 
 
-class TestToolsMPAClassification:
+class TestToolsClassification:
```
harimkang: It would be better if the multi-class classification had a distinct name.

harimkang (commented):
Added a few comments.

otx/algorithms/classification/configs/base/data/data_pipeline.py
JihwanEom: To align with semi-sl pipeline. I think we don't need to maintain `ToNumpy`.

otx/mpa/cls/stage.py
JihwanEom: That enables ComposedDL

otx/mpa/cls/stage.py
JihwanEom: If we don't have any input for unlabeled batch size, batch size of labeled sample will be used.

otx/algorithms/classification/configs/base/data/data_pipeline.py
harimkang: I don't understand. As far as I know, `ToNumpy` is a pipeline provided by `mmcls`. Do we need to modify this if nothing else than `PILImageToNDArray`? Any difference?

otx/mpa/modules/models/heads/semisl_cls_head.py
JihwanEom: Nonlinear uses `classifier` instead of `fc`. It should be separated. Also, inheriting linear head for non-linear head has been looked weird (I've tried but it showed poor readability).

otx/mpa/cls/stage.py
harimkang: How to check? or Where to check if unlabeled_batch_size have or not?

otx/mpa/modules/models/heads/semisl_cls_head.py
harimkang: Then, how about creating a `SemiLinearHead` that has a separate `SemiSLHead` and inherits `SemiSLHead` and `LinearHead`? (NonLinear as well)

- SemiSLHead : Head for SemiSL algorithm
- SemiLinearHead: SemiSLHead + LinearHead
- SemiNonLinearHead: SemiSLHead + NonLinearHead

If you make it like this, the class will increase by one, but I think you can handle redundant code.
How is it?

otx/mpa/modules/models/classifiers/semisl_classifier.py
JihwanEom: I partially agree, but to change `img_strong`, we need to change data pipeline flow from bottom. 
I distinguished weak augmented image and strong augmented image by branching out at different time.
But for `img_strong`, absolutely another implementation will be needed.

harimkang: It's very confusing.
- Labeled image -> img : weak pipeline
- Unlabeled image
  - img : strong pipeline
  - weak_img: weak pipeline
 
Isn't it a part that can be solved by changing the key in the pipeline?

otx/algorithms/classification/tasks/train.py
JihwanEom: Fixed.

otx/mpa/modules/models/heads/semisl_cls_head.py
jaegukhyun: I agree with harim, if head gets linear or non-linear flag, it can reduce lots of codes

jaegukhyun (changes requested):
Generally LGTM. Thanks for your effort. I agree with Harim's opinion about SemiCLSHead. It has lots of duplicated code, and it can be reduced if SemiCLSHead knows user wants whether linear or nonlinear head

otx/algorithms/classification/configs/base/data/data_pipeline.py
JihwanEom: `PILImageToNDArray` does same work with `ToNumpy`, but it supports multiple key inputs.

otx/mpa/cls/stage.py
JihwanEom: Check here with `pop`

otx/mpa/modules/models/heads/semisl_cls_head.py
JihwanEom: I'll refactor this file.

otx/mpa/modules/models/classifiers/semisl_classifier.py
JihwanEom: Yes, that can't be solved by only changing key name. To use `img_strong` key, time-independent image branching pipeline is needed.

tests/integration/cli/classification/test_classification.py
JihwanEom: Reflected.

otx/mpa/modules/models/heads/semisl_cls_head.py
JihwanEom: Refactored. Please review again.

otx/algorithms/classification/configs/efficientnet_v2_s_cls_incr/semisl/hparam.yaml
```diff
@@ -0,0 +1,22 @@
+# Hyperparameters.
```
supersoob, line 1: where is unlabeled_batch_size here? Is it okay to get the default unlabeled_batch_size configured in configuration.yaml?

otx/algorithms/classification/tasks/inference.py
```diff
@@ -342,6 +354,10 @@ def _init_recipe(self):
             if not self._multilabel and not self._hierarchical:
                 if self._data_cfg.get("data", None) and self._data_cfg.data.get("unlabeled", None):
                     recipe = os.path.join(recipe_root, "semisl.yaml")
+                    self.task_model_dir = os.path.join(self.task_model_dir, "semisl")
+                    self.task_pipeline_path = os.path.join(
```
supersoob: Please align it as model_dir, pipeline_path.

otx/mpa/modules/models/heads/semisl_cls_head.py
```diff
@@ -6,47 +6,24 @@
...
+from otx.mpa.modules.models.heads.non_linear_cls_head import NonLinearClsHead
 
-@HEADS.register_module()
-class SemiSLClsHead(LinearClsHead):
-    """Semi-SL for Classification Head
 
-    This ClsHead is a classification linear head based on the FixMatch algorithm.
-    This head uses the dynamic threshold for each class calculated based on
-    the confidence value of the model.
+class SemiClsHead:
```
supersoob, line 12: I'm very sorry but I still can't understand your refactored code as well. I think it would better inherit both LinearClsHead and NonLinearClsHead here and remove all SemiLinearClsHead or SemiNonLinearClsHead, remaining only this class. If that's impossible, it's okay to leave it as it. Plus, I think SemiSLClsHead is more proper name.

otx/mpa/modules/models/classifiers/semisl_classifier.py
harimkang: The current pipeline structure is quite inconvenient. I'll think about it some more.

otx/algorithms/classification/configs/base/data/data_pipeline.py
harimkang: So this is like `ToNumpy` but for multiple image keys supporting (for `semisl`)?

otx/algorithms/classification/configs/efficientnet_v2_s_cls_incr/semisl/hparam.yaml
JihwanEom: Updated. thank you!

otx/algorithms/classification/tasks/inference.py
JihwanEom: Reflected.

otx/algorithms/classification/configs/base/data/data_pipeline.py
JihwanEom: Yes, right

otx/algorithms/classification/configs/base/data/semisl/data_pipeline.py
```diff
@@ -0,0 +1,68 @@
...
+
+__train_pipeline = [
+    *__common_pipeline,
+    dict(type="PILImageToNDArray", keys=["img"]),
+    dict(type="Normalize", **__img_norm_cfg),
+    dict(type="ImageToTensor", keys=["img"]),
+    dict(type="ToTensor", keys=["gt_label"]),
+    dict(type="Collect", keys=["img", "gt_label"]),
+]
+
```
sungmanc, line 39: Maybe I think we can use below code (just suggestion).

For the `data_pipeline.py`.
```python3
__common_pipeline = [
    dict(type="Resize", size=__resize_target_size),
    dict(type="RandomFlip", flip_prob=0.5, direction="horizontal"),
    dict(type="AugMixAugment", config_str="augmix-m5-w3"),
    dict(type="RandomRotate", p=0.35, angle=(-10, 10)),
]
__unlabeled_strog_pipeline = [
    dict(type="MPARandAugment"),
    dict(type="PILImageToNDArray"),
    dict(type="ImageToTensor"),
    dict(type="Collect"),
]

__unlabeled_weak_pipeline = [
    dict(type="PILImageToNDArray"),
    dict(type="ImageToTensor"),
    dict(type="Collect"),
]
```

And then, we can use
```python3
img = common_pipeline(img)
strong_img = unlabeled_strong_pipeline(img)
weak_img = unlabeled_weak_pipeline(img)
```

sungmanc (commented):
Please make type hint, docstring for your code. 
Since tox's pre-commit test will ignore the otx/mpa, need to check with local pre-commit test without tox

otx/algorithms/classification/configs/base/data/semisl/data_pipeline.py
JihwanEom: Reflected. Please review again

otx/mpa/modules/models/classifiers/semisl_classifier.py
JihwanEom: Updated. Please review again.

otx/algorithms/classification/adapters/mmcls/data/pipelines.py
```diff
@@ -137,3 +139,64 @@ def __repr__(self):
...
+            raise ValueError("key 'img' should be included in pairs argumenent")
+        self.pipelines = {key: Compose(pipeline) for key, pipeline in pairs.items()}
+
+    def __call__(self, results):
+        """Callback function of SeparateAug."""
+        for key, pipeline in self.pipelines.items():
+            if key == "img":
+                output = pipeline(copy.deepcopy(results))
+            else:
+                output[key] = pipeline(results)["img"]
```
harimkang: I have one question, the order of this is a bit confusing. Is it that `img` goes through the weak pipeline first, and then copying for strong img?

JihwanEom: No, `results` is **not augmented, original image**. 
output["img"] will be an image augmented by weak pipeline,
output[key] will be an image augmented by strong pipeline. This code does not result weak+strong (jointed in one) augmented image.

harimkang: So, isn't this method completely different from the last commit? Isn't it call the common(weak) pipeline twice each?

In my memory, didn't we discuss that the common pipeline was carried out with one img and then only added after that?

JihwanEom: I understood that you wanted to have time independent data pipeline and merge into one element. Am I wrong?
What is your desired code? Please give me detailed description. I'll follow that.

JihwanEom: Reflected. Please review again for this part. I'll rebase after review.

tests/integration/cli/classification/test_classification.py
```diff
@@ -234,6 +234,22 @@ def test_otx_multi_gpu_train(self, template, tmp_dir_path):
         otx_train_testing(template, tmp_dir_path, otx_dir, args1)
 
 
+class TestToolsMultiClassSemiSLClassification:
+    @e2e_pytest_component
+    @pytest.mark.parametrize("template", templates, ids=templates_ids)
+    def test_otx_train(self, template, tmp_dir_path):
+        args_semisl = args0.copy()
+        args_semisl["--unlabeled-data-roots"] = args["--train-data-roots"]
+        args_semisl["train_params"].extend(["--algo_backend.train_type", "SEMISUPERVISED"])
```
jaegukhyun, line 236: This may harm original args0

sungmanc (dismissed):

otx/algorithms/classification/configs/efficientnet_b0_cls_incr/semisl/hparam.yaml
```diff
@@ -0,0 +1,24 @@
...
+      batch_size:
+        default_value: 16
+        auto_hpo_state: POSSIBLE
+      unlabeled_batch_size:
+        default_value: 48
+      learning_rate:
+        default_value: 0.00245
+        auto_hpo_state: POSSIBLE
+      early_stop_start:
+        default_value: 50
```
kprokofi, line 15: Isn't it too much to start early_stop? What if we have simple small dataset. What is the reason to apply this heuristic?

kprokofi (commented):
question

otx/algorithms/classification/configs/efficientnet_b0_cls_incr/semisl/hparam.yaml
jaegukhyun: I guess it comes from experiments. Maybe semisl models converge slow. I think we should add skipping early stop count if model's accuracy is below the threshold. It will be useful for any tasks and training algorithms. But my suggestion is beyond this PR. Let's discuss it later

JihwanEom: Thanks Jaeguk, Jaeguk explained instead of me perfectly :) `early_stop` related parameters are based on experiment results, even we start from 50 epoch, it requires under 4.0x training time with stable performance gain (10~15%).

tests/integration/cli/classification/test_classification.py
JihwanEom: Replaced `copy()` with `copy.deepcopy`. Thank you!

otx/algorithms/classification/configs/efficientnet_b0_cls_incr/semisl/hparam.yaml
harimkang: According to the results of the last PoC, at least 50 epochs must be run to ensure better performance than supervised. An agreement has been reached to produce this. So, it is correct that it should go into 50 epoch now, and it can be modified if there is an upgrade of the algorithm in the future.

harimkang (approved):
Thank you for your hard work. If all classification test results pass, I think it's OK to merge.

JihwanEom:
Passed all of the cls/det/seg cli local tests.

Classification
image

Detection
image

Segmentation
image

jaegukhyun (approved):

PR #1475, opened 2023-01-02
[OTX] Bugfix: multi GPU raise error when num_workers isn't set as 0.

eunwoosh:
### Summary

- fix a bug that error is raised when multi gpu training with none zero num_workers.

### Reason of the bug
When process is spawned, deafult multi porcess method is set as "spawn". It raises error when `dataloader` is used with num_workers > 0.
This is because dataloader has a DatasetItemEntity which has a thread lock attribute and thread lock is unpickleable.
I don't know exact reason, but when forking a new process, unpickleable argument can be passed to new process.

### DIFF
--- a/otx/cli/utils/multi_gpu.py
+++ b/otx/cli/utils/multi_gpu.py
@@ -169,6 +169,7 @@ class MultiGPUManager:
             output_path (str): output path where task output are saved.
             multi_gpu_port (str): port for communication between multi GPU processes.
         """
+        mp.set_start_method(None, True)  # set multi porcess method as default
         gpus_arg_idx = sys.argv.index("--gpus")
         for _ in range(2):
             sys.argv.pop(gpus_arg_idx)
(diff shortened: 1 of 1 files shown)


otx/cli/utils/multi_gpu.py
```diff
@@ -169,6 +169,7 @@ def run_child_process(train_func: Callable, rank: int, gpu_ids: List[int], outpu
             output_path (str): output path where task output are saved.
             multi_gpu_port (str): port for communication between multi GPU processes.
         """
+        mp.set_start_method(None, True)  # set multi porcess method as default
```
goodsong81: So, what is this line is doing? None? True?

eunwoosh: This code set multi process method as default by force.
In the Unix, default multi process method is "fork". So, it's set as "fork".

goodsong81: Yeah, but your intention is well hidden.
mp.set_start_method(method="fork", force=True) <- Isn't it better?

eunwoosh: You're right. I'll update it.

harimkang (approved):

goodsong81 (approved):

PR #1480, opened 2023-01-04
[OTX] Add test dataset for action classification and action detection

jaegukhyun:
Add datumaro format test dataset for action classification and action detection
These dataset come from original ote/data/custom_action_recognition/


sungmanc (approved):
LGTM.

supersoob (approved):

PR #1482, opened 2023-01-04
[DEV] Cache numpy for VideoFrame

goodsong81:


external/mmdetection/detection_tasks/extension/utils/pipelines.py
```diff
@@ -25,6 +27,12 @@
 from ..datasets import get_annotation_mmdet_format
 
 
+# Clean up cache directory per process launch
+_CACHE_DIR = "/tmp/img-cache"
```
sovrasov: tempfile.TemporaryDirectory? Or you intend to save this cache across runs?

sungmanc: Can we delete the temporary directory at every training? AFAIK, /tmp directory is cleaned at every booting.
In this PR, since we save lots of images in cache-directory, deleting it at every training is needed. 
And for this case, we can't use tempfile.TemporaryDirectory.

JihwanEom: Reflected. Please refer #1491

external/mmsegmentation/segmentation_tasks/extension/utils/pipelines.py
```diff
@@ -40,8 +44,8 @@ def __init__(self, to_float32: bool = False):
 
     @check_input_parameters_type()
     def __call__(self, results: Dict[str, Any]):
-        dataset_item = results['dataset_item']
-        img = dataset_item.numpy
+        # Get image (possibly from cache)
+        img = get_image(results, _CACHE_DIR.name, to_float32=self.to_float32)
```
sovrasov, line 48: If we rename `get_image` function to something like `get_image_cached`, the comment above won't be required.

JihwanEom: That can be conflicted `load_image_from_cache` in data_utils.py meaningfully. I think it will be better to remove comment.

goodsong81:
@JihwanEom will create a new PR for this.

PR #1484, opened 2023-01-05
[OTX] Refactor mpa stage folder structure and inheritance in segmentation

supersoob:
This PR only includes refactoring codes for segmentation

- Refactor mpa stage folder structure
- Remove unclear inheritance
- Separate into base SegStage, incremental SegStage, Semisl SegStage
- Fix bug in TestToolsMPASemiSLSegmentation in CLI test


supersoob:
image
local test results for segmentation.

otx/mpa/seg/__init__.py
```diff
@@ -17,3 +17,4 @@
 # flake8: noqa
 from . import evaluator, exporter, inferrer, stage, trainer
 from .semisl import inferrer, trainer
+from .incremental import inferrer, trainer
```
jaegukhyun: Is this okay? What happens someone call from otx.mpa.seg import trainer?

otx/mpa/seg/incremental/stage.py
```diff
@@ -0,0 +1,43 @@
...
+
+logger = get_logger()
+
+
+class IncrSegStage(SegStage):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    def configure_task(self, cfg, training, **kwargs):
+        """Adjust settings for task adaptation"""
```
jaegukhyun, line 20: I think this docstring does not match anymore.
Does task adaptation is handle in basic stage?
Then this function is for incremental learning settings

jaegukhyun (changes requested):
I left a comment about initialization, and I have a question about base trainer and inferrer
Do they inherits from base stage or incremental stage?

otx/mpa/seg/__init__.py
supersoob: I missed that point. Thanks. It has to call inferrer and trainer inside incremental, but it wasn't. I will fix.

supersoob: Actually it would not matter because each trainer and inferrer was using other class name.

jaegukhyun (dismissed):

otx/mpa/seg/__init__.py
sungmanc: Do we need these kinds of imports? Maybe we can remove all kinds of import by adding `__init__.py` inside of each folders. And in this point, we can see the `__init__.py` under `incremental` folder, but can't find in `semisl` folder. we need to unify the file arhictecture.

sungmanc (commented):
Could you check this PR with local pre-commit tests?

Currently, `tox` is not working for `otx/mpa` and `otx/recipes`. So, we should check those folders by using `pre-commit run all-files`. not using `tox -re/-e pre-commit`.

otx/mpa/seg/incremental/stage.py
```diff
@@ -0,0 +1,43 @@
...
+        super().configure_task(cfg, training, **kwargs)
+
+        self.configure_incremental(cfg, self.org_model_classes, self.model_classes)
+
+    def configure_incremental(self, cfg, org_model_classes, model_classes):
```
eugene123tw: Could we rename this method to sth like `add_new_labels`? Configuration is a broad concept, but we are basically updating labels. And arguments to `old_labels`, `new_labels`? We are also mixing labels and classes, but in SDK we use labels, we should stick to one naming convention. 

In `otx.mpa.seg.stage.SegStage` we have `configure_classes` method. Could we not use that method?

otx/mpa/seg/incremental/stage.py
```diff
@@ -0,0 +1,43 @@
...
+from mmcv import ConfigDict
+
+from otx.mpa.seg.stage import SegStage
+from otx.mpa.utils.config_utils import update_or_add_custom_hook
+from otx.mpa.utils.logger import get_logger
+
+logger = get_logger()
+
+
+class IncrSegStage(SegStage):
```
eugene123tw, line 15: `otx.mpa.stage.Stage` looks a bit ambiguous. It seems like a base runner (e.g. BaseRunner), but it doesn't really explain/comment on what it does. And to make things more complex, we have a `BaseTask` that looks like another baserunner. I suggest we either keep `BaseTask` or `Stage` (rename to sth clearer) but not both. For instance, `BaseTaskRunner` and each task can inherit from this (e.g. `SegTaskRunner(BaseTaskRunner)`, ...). If we need to extend to incremental learning we can extend this `IncrementalSegRunner(SegTaskRunner, IncrementalRunner)`

otx/mpa/seg/incremental/trainer.py
```diff
@@ -0,0 +1,17 @@
...
+from otx.mpa.utils.logger import get_logger
+
+from .stage import IncrSegStage
+from otx.mpa.seg.trainer import SegTrainer
+
+logger = get_logger()
+
+
+@STAGES.register_module()
+class IncrSegTrainer(IncrSegStage, SegTrainer):
```
eugene123tw, line 15: If this class doesn't do anything could we not have it? Could we not have an `IncrementalSegRunner(SegTaskRunner, IncrementalRunner)` which `SegTaskRunner` has 2 methods 1.) `SegTaskRunner.train(dataloader)` 2.) `SegTaskRunner.inference(dataloader)`? In this case, we can lose all `xTrainer`, `xInferer` classes

otx/algorithms/segmentation/tasks/inference.py
```diff
@@ -217,11 +217,10 @@ def _init_recipe(self):
         logger.info(f"initialized recipe = {recipe}")
 
     def _update_stage_module(self, stage_module: str):
-        if self.train_type == TrainType.SEMISUPERVISED:
-            if stage_module == "SegTrainer":
-                return "SemiSegTrainer"
-            if stage_module == "SegInferrer":
-                return "SemiSegInferrer"
+        module_prefix = {TrainType.SEMISUPERVISED: "Semi", TrainType.INCREMENTAL: "Incr"}
```
eugene123tw: It'd be better if we have a `SegTask` that has 2 methods - `train` and `inference`, so we don't need 2 more classes (`xTrainer` and `xInferrer` for every task). For seg, we have `SegTrainer`, `SegInferrer`, SemiSegTrainer, SemiSegInferrer, IncrSegTrainer, and IncrSegInferrer, that's overkill. Instead, we can have `SegTask`, `SemiSegTask`, `IncrementalSegTask`. From 6 to 3. 

I also don't think this a good way to get class using string, it's easily breakable, and the intention is unclear. Could we not tell which task class we are training? If parsed the task type, and the exact task CLI from `train.py`. Could we not get the class directly from that 2 pieces of information?

eugene123tw (commented):
Hi @supersoob Thanks for creating this PR. I have added a few comments. But most comments are down to design and principle. IMHO, we should think about redesigning the interfaces and base modules as it's difficult to navigate and difficult to read what's going on.

otx/mpa/seg/incremental/stage.py
supersoob: Your comments are totally right. I agree that we don't need to keep stage as long as we only maintain otx now and it has to be disappear in the future. But I'd like to let you the specific design for the runners are not decided yet because we are in the middle of refactoring stage to get to the otxv2 design. I wanted to remove the current unclear inheritance which is very different from other tasks (look at det/cls, they all different and confusing.) We discussed to keep the same folder structure and I'd say that it has to be, whenever. And this clarity and structured stages will surely help refactor for the final design for the future, since this PR includes the separation of configuration with respect to train type. As conclusion, this is not a PR for refactoring whole runners design and I hope we can have some discussion soon for how to merge runners and stages because it is very huge task which requires to have the very different structure compared to the current.

otx/mpa/seg/incremental/stage.py
supersoob: I'd like to leave this method name as configure_incremental. Because it is just used when incremental learning is enabled like special sampler for incremental learning. I think configure_classes is more close to add_new_labels. I will revise it.

otx/mpa/seg/incremental/trainer.py
supersoob: I hope you refer to my comment above.

otx/algorithms/segmentation/tasks/inference.py
supersoob: I hope you refer to my comment above as well. And about module name, I will revise it to get the value with full name.

otx/mpa/seg/__init__.py
supersoob: semisl has it, but with no content. I want to ask @sungmanc if it's okay to partially update the convention and follow pre-commit test rules, then I will revise my updated code to match with otx format including `__init__`.

eugene123tw (dismissed):
Thanks, Soobee, I agree this PR has mitigated gaps and cleared some degree of ambiguity. I'm interested in otxv2 and what the changes are. I hope we can refactor this asap so anything builds on top is stable.

jaegukhyun (dismissed):

eugene123tw (dismissed):

supersoob:
Hi, I found the potential bug in test cases for semisl and revised it. I also aligned the semi-sl trainer and inferrer name with det. I hope you review again.
Local test results ->
image

otx/algorithms/segmentation/tasks/inference.py
```diff
@@ -212,11 +212,10 @@ def _init_recipe(self):
         logger.info(f"initialized recipe = {recipe}")
 
     def _update_stage_module(self, stage_module: str):
-        if self._train_type == TrainType.SEMISUPERVISED:
-            if stage_module == "SegTrainer":
-                return "SemiSegTrainer"
-            if stage_module == "SegInferrer":
-                return "SemiSegInferrer"
+        module_prefix = {TrainType.SEMISUPERVISED: "SemiSL", TrainType.INCREMENTAL: "Incr"}
```
harimkang, line 215: As far as I know, there is `selfsl` for segmentation, but does it just use `SegTrainer`?

otx/mpa/seg/__init__.py
harimkang: This code still seems to have a problem. This is because inferrer and trainer are imported redundantly, so only semisl is imported here. Looks like you need to fix it.

otx/mpa/seg/semisl/stage.py
```diff
@@ -11,17 +11,17 @@
...
 
     def configure_task(self, cfg, training, **kwargs):
         """Adjust settings for task adaptation"""
         super().configure_task(cfg, training, **kwargs)
 
-        self.configure_semi(cfg, **kwargs)
+        self.configure_semisl(cfg, **kwargs)
 
-    def configure_semi(self, cfg, **kwargs):
+    def configure_semisl(self, cfg, **kwargs):
```
harimkang: Could this function be merged with `configure_task`?

harimkang (commented):
I left a few comments.

otx/algorithms/segmentation/tasks/inference.py
supersoob: It does not have its own stage.py and trainer. It works well by SegTrainer.

otx/mpa/seg/__init__.py
supersoob: Well,, it is not. Because the class name is totally different, but i will revise it like OTX init.

otx/mpa/seg/semisl/stage.py
supersoob: I hope to leave as is. configure_task is collections of configurations, so as other stages. It is good to seperate the method functionality as minimum.

jaegukhyun: How about configure_data? Actually detection does

otx/mpa/seg/semisl/stage.py
```diff
@@ -11,17 +11,17 @@
 logger = get_logger()
 
 
-class SemiSegStage(SegStage):
+class SemiSLSegStage(SegStage):
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
```
harimkang, line 16: since `SemiSLSegStage` inherits `SegStage`, doesn't it work well without `init`?

otx/mpa/seg/semisl/stage.py
harimkang: I feel that.. the name `configure_semisl` doesn't sound right.. I don't really have any ideas, but the method of overriding an existing function looks good.

otx/mpa/seg/__init__.py
sungmanc: > semisl has it, but with no content. I want to ask @sungmanc if it's okay to partially update the convention and follow pre-commit test rules, then I will revise my updated code to match with otx format including `__init__`.

Ah, I missed the MPA architecture for `__init__.py`. I didn't realize that most of the modules in MPA used direct import(i.e. `from .incremental import inferred, trainer`). I agreed with the current implementation. Let's discuss MPA refactoring later.

For the Harim's opinion, I also agree with this idea, how about using the one like as below:
``` from otx.mpa.seg.incremental.inferrer import IncrSegInferrer ```

@harimkang @supersoob

sungmanc: Refactoring all kind of `__init__.py` included in MPA could be handled by another PR because we also need to enable `mypy` and `pylint`, `flake8` for MPA.

supersoob: okay:)) I realized I got it wrong. I was obsessed by my thought. Sorry. I will revise it as just sungman said.

otx/mpa/seg/semisl/stage.py
supersoob: yeah working well. it already had init file previously but with empty content so didn't have to make it here.

otx/mpa/seg/semisl/stage.py
supersoob: I will just remove and merge it both in incremental and semisl. It would be good to seperate the function when more method or config is added and realized that the name configure_task covers all itself. Thanks for comments.

supersoob:
@harimkang @sungmanc @jaegukhyun I changed based on your reviews.
Local test results -> 
image

harimkang (approved):

sungmanc (approved):

jaegukhyun (approved):

kprokofi (approved):

supersoob:
Failed 2 cases which is not related to this PR
image

PR #1485, opened 2023-01-05
[OTX] Align data pipeline configs & OTX build with sub train type

harimkang:
## Summary
- As discussed in the OTX group, each model has a data pipeline config file. It assumes that `model.py` and `data_pipeline.py` are in the same folder and raises an Error if they are not present.
- Implemented OTX workspace builds for all training type(SemiSL and SelfSL).

## Contents
- [X] Align data pipeline config patching in tasks
- [X] OTX build
  - [X] classification incremental, selfsl build & train
  - [X] detection incremental, semisl build & train
  - [X] segmentation incremental, semisl build & train
- [X] Regression Test -> ALL PASSED (CLI test in local)
image

## How to use
```
# TASK = {classification, detection, segmentation, instance_segmentation}
# TRAIN_TYPE = {incremental(default), semisl, selfsl}
otx build --task {TASK} --train-type {TRAIN_TYPE}
```


harimkang:
@sungchul2 I tried segmentation-selfSL with any training data, but there is a key error on the mpa side. As far as I remember, it seems to only work with certain dataset structures. Is that because of that? (Should I not care? It works until training, but after that, an error occurs on the dataset load side.)

harimkang:
First of all, I'm going to run a test locally. In the meantime, please check if there are any problems with the code structure.

harimkang:
image
All CLI test PASSED in local

otx/algorithms/classification/tasks/inference.py
```diff
@@ -330,41 +330,26 @@ def _init_recipe(self):
...
 
-        if train_type == TrainType.SELFSUPERVISED:
+        if self._train_type == TrainType.INCREMENTAL:
+            recipe = os.path.join(recipe_root, "incremental.yaml")
+        elif self._train_type == TrainType.SELFSUPERVISED:
             recipe = os.path.join(recipe_root, "selfsl.yaml")
-            self.model_dir = os.path.join(self.model_dir, "selfsl")
-            pipeline_path = os.path.join(os.path.dirname(pipeline_path), "selfsl/data_pipeline.py")
+        elif self._train_type == TrainType.SEMISUPERVISED:
+            recipe = os.path.join(recipe_root, "semisl.yaml")
```
sungmanc: Even though already checked by line#335, how about adding the below lines?
```python3
else:
    raise NotImplementedError()
```

otx/algorithms/common/tasks/training_base.py
```diff
@@ -46,6 +46,11 @@
 from otx.mpa.utils.logger import get_logger
 
 logger = get_logger()
+TRAIN_TYPE_DIR_PATH = {
+    TrainType.INCREMENTAL.name: ".",
```
sungmanc, line 50: Is there any better way to avoid using `"."` ?

otx/algorithms/classification/tasks/inference.py
sungmanc: Same opinion for : https://github.com/openvinotoolkit/training_extensions/pull/1485/files#diff-6b1f990ed67dacee34cd5e1a9ef8545c98171d60c417698936e7f4a7973f21a8R274 and https://github.com/openvinotoolkit/training_extensions/pull/1485/files#diff-38970adf7b9d362c1a674ca03fa4e639def08cbe36c4da5daaeea4f2f5bc6b54R200

otx/cli/builder/builder.py
```diff
@@ -174,15 +180,28 @@ def build_task_config(self, task_type, model_type=None, workspace_path=None, otx
         template_config = MPAConfig.fromfile(template.model_template_path)
         template_config.hyper_parameters.base_path = "./configuration.yaml"
 
+        # Configuration of Train Type value
+        sub_rel_path = "."
```
sungmanc: if the objective of this line is just for initializing, how about `""`?

sungmanc: When I first meet `"."` , I think `.` has a meaning.

sungmanc: +I prefer to use `train_type_rel_path`, `train_type_dir` than `sub_rel_path`, `sub_workspace_dir`, if there is no problem

sungmanc (commented):
With respect to architectural uniformity, please re-check the below opinion.

classification and segmentation tasks used `configs/base/data/data_pipeline` or `configs/base/data/selfsl/data_pipeline.py`.  However, detection doesn't use this architecture. There is no `data_pipeline.py` under the `detection/configs/base/`. How about making same architecture for each tasks?


+I also left some comments

otx/algorithms/classification/tasks/inference.py
sungchul1: If so, added condition is not used.
So I think it looks better to move L335 to the last if-else statement as @sungmanc said, what do you think?

otx/cli/builder/builder.py
sungchul1: I guess it is used to set `model_dir` to current `template_dir` or `sub_workspace_dir` to current `workspace_path`.
But the newly assigned paths are set through `os.path.join`, so it would be okay to use `""` instead of `"."` as @sungmanc said.

otx/cli/builder/builder.py
```diff
@@ -174,15 +180,28 @@ def build_task_config(self, task_type, model_type=None, workspace_path=None, otx
...
+        if train_type.lower() != "incremental":
+            sub_rel_path = train_type.lower()
```
sungchul1: We already get `train_type` as lower case at [link](https://github.com/openvinotoolkit/training_extensions/pull/1485/files#diff-bf8911e3241497afed1050a4d4a1995b8dc64417cc0edfaca70668d32dc72421R35).

https://github.com/openvinotoolkit/training_extensions/blob/a9a79a273e830f74e6685a7de066d3b260393aa0/otx/cli/tools/build.py#L34-L36

I think `.lower()` is a duplicated operation.

sungchul1 (commented):
LGTM. I left some minor comments.

sungchul1:
Although det and seg use `patch_color_conversion` in `otx/algorithms/common/adapters/mmcv/utils.py`, cls still uses this function. (also `patch_datasets` and `patch_evaluation` of det and seg are in each `utils/config_utils.py`, but cls's is here.)
https://github.com/openvinotoolkit/training_extensions/blob/a9a79a273e830f74e6685a7de066d3b260393aa0/otx/algorithms/classification/tasks/inference.py#L390-L461
It looks better to match them, do you have a plan to refactor them?

otx/algorithms/classification/tasks/inference.py
harimkang: yes i agree, I will move L335 to the corresponding else statement.

otx/algorithms/common/tasks/training_base.py
harimkang: There are currently no ideas that come to mind. There would be a way to create a sub-dir called `Incremental` for unification, but I don't think it's appropriate.

otx/cli/builder/builder.py
harimkang: Yes, I will modify the initial value to "" as suggested.

otx/cli/builder/builder.py
harimkang: yes i will fix it! thanks

harimkang:
> It looks better to match them, do you have a plan to refactor them?

Regarding classification, it is difficult for me to give a definite answer. @supersoob @JihwanEom Could you possibly answer this question?

otx/algorithms/classification/tasks/inference.py
harimkang: I fixed this in that commit (https://github.com/openvinotoolkit/training_extensions/pull/1485/commits/509e1eecb806ef1ed3a66878822c0081b37a8c9b). Please check again!

otx/cli/builder/builder.py
harimkang: Fixed! https://github.com/openvinotoolkit/training_extensions/pull/1485/commits/598606871255bd2ceeddcaecec3ed016f64b198a

otx/cli/builder/builder.py
harimkang: Fixed! https://github.com/openvinotoolkit/training_extensions/pull/1485/commits/598606871255bd2ceeddcaecec3ed016f64b198a

harimkang:
> classification and segmentation tasks used `configs/base/data/data_pipeline` or `configs/base/data/selfsl/data_pipeline.py`. However, detection doesn't use this architecture. There is no `data_pipeline.py` under the `detection/configs/base/`. How about making same architecture for each tasks?

Is your suggestion to remove `data_pipeline.py` from `configs/base`?
It doesn't matter, but I remember talking about it at the meeting to do it this way. and also said that data_pipeline will change its location and become invisible later. Now, we have specified the base path in cls/seg, but do you think it should be changed? (I think, rather than copying and using the same code-pipeline, the current method of specifying it through the base path looks better.)

-> This is a temporary solution, so any ideas are welcome!

harimkang:
CI is not working, so local testing is in progress.

sungmanc:
> bout it at the meeting to do it this way. and also said that data_pipeline will change its location and become invisible later. Now, we have specified the base path in cls/seg, but do you think it should be changed? (I think, rather than copying and using the same code-pipeline, the current method of specifying it through the base path looks better.)

Agree with your opinion. Architectural unification should be included in TODO item. 
My concern is when we can start this process...

jaegukhyun (approved):
LGTM

sungchul1 (approved):
LGTM

sungmanc (approved):
LGTM, merge this after checking the test results

harimkang:
[CLI Test in Local]
- classification: H-label Failed (not related this PR, reported to the group)
  - multi-class
image
  - multi-label
image
  - self-SL
image
- detection (+semisl): All PASSED
image
  - tiling: All PASSED
image
  - instance-seg: All PASSED
image
  - ins-seg tiling: All PASSED
image
- segmentation(+semisl, selfsl): 1 Failed (score drop)
image
- action
  - classification: All PASSED
image
  - detection: All PASSED
image
- CLI
image

PR #1393, opened 2022-11-25
[OTE] Update OpenVINO to 2022.3 release

sovrasov:
- Update OV
- Update NNCF, corresponding fixes in d-o-r
- Minor fixes in readmes
- Changes from #1487

PR #1487, opened 2023-01-05
[DEV] Fix nncf config for padim model

AlexanderDokuchaev:


sovrasov:
@AlexanderDokuchaev thanks for the PR. Unfortunately, it cannot be tested without #1393, so I added your commit there

sovrasov:
These changes were applied in https://github.com/openvinotoolkit/training_extensions/pull/1393

PR #1488, opened 2023-01-05
[DRAFT] Task implementation for region-based anomaly detection

djdameln:
This PR adds the task implementation for region-based anomaly detection (or RKDE: Region-based Kernel Density Estimation).

- Adds the config files and `template.yaml` for the new task.
- The task was updated to accommodate the most recent changes on the anomalib side (mainly related to callbacks and some variable/parameter names)
- Some changes were needed to the data and inference functionality of the anomaly task to accommodate this new model:
    - In the `OTXAnomalibDataset` class, we now pass the raw box annotations to anomalib as GT (Previously we converted to mask first).
    - In the inference callback, we now process the box predictions from the anomalib and convert those into box annotations (previously this was done by converting masks to boxes).


sungmanc:
If this PR is under WIP, could you change this PR's status to DRAFT ?

PR #1491, opened 2023-01-06
[DEV] Save numpy array as compressed npz in videoframe cache & Exception handling

JihwanEom:
To reduce space complexity, changed to use compressed npz when save numpy array.
+) Add exception handling for file lock
+) Use tempfile.TemporaryDirectory (cached file automatically will be removed after training)
+) Prevent caching in evaluation phase


sungmanc (approved):
LGTM

PR #1500, opened 2023-01-09
[OTX] Refactor stage structure in detection

jaegukhyun:
This PR includes refactoring stage in mpa/det.

- Add trainer and inferrer to mpa/det/incremental
- Move task adapt part to base stage, so that incremental stage is only responsible for incremental learning
- Base trainer, inferrer inherits from base stage
- Semi-SL stage inherits from base stage


otx/algorithms/detection/tasks/inference.py
```diff
@@ -303,11 +303,13 @@ def _init_test_data_cfg(self, dataset: DatasetEntity):
         return data_cfg
 
     def _update_stage_module(self, stage_module):
-        if self._train_type == TrainType.SEMISUPERVISED:
-            if stage_module == "DetectionTrainer":
-                stage_module = "SemiSLDetectionTrainer"
-            elif stage_module == "DetectionInferrer":
-                stage_module = "SemiSLDetectionInferrer"
+        if stage_module in ["DetectionTrainer", "DetectionInferrer"]:
```
kprokofi: let's create prefix dict, as Soobee did. Because, when we will have more training types, "if" will be spreading out
image
Also, I suggest using assert checking "self.train_type": `assert` self.train_type in prefix_dict, f"{self.train_type} do not have own trainer or inferrer"`

otx/mpa/det/incremental/inferrer.py
```diff
@@ -0,0 +1,36 @@
...
+
+from .stage import IncrDetectionStage
+
+logger = get_logger()
+
+
+@STAGES.register_module()
+class IncrDetectionInferrer(IncrDetectionStage, DetectionInferrer):
+    def __init__(self, **kwargs):
+        IncrDetectionStage.__init__(self, **kwargs)
```
kprokofi, line 17: Maybe we need to initialize `DetectionInferrer` too

otx/mpa/det/incremental/trainer.py
```diff
@@ -0,0 +1,35 @@
+# Copyright (C) 2022 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
+
+import glob
+import os.path as osp
```
kprokofi: Do we need all of this imports if we create only one class without any new methods/attributes?

otx/mpa/det/incremental/trainer.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+logger = get_logger()
+
+
+# FIXME DetectionTrainer does not inherit from stage
+@STAGES.register_module()
+class IncrDetectionTrainer(IncrDetectionStage, DetectionTrainer):
+    def __init__(self, **kwargs):
+        IncrDetectionStage.__init__(self, **kwargs)
```
kprokofi, line 17: Same about `DetectionTrainer` initialization

otx/mpa/det/incremental/__init__.py
```diff
@@ -2,6 +2,8 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
+from .inferrer import IncrDetectionInferrer
 from .stage import IncrDetectionStage
```
kprokofi: Do we need to import Stage now?

otx/mpa/det/semisl/stage.py
```diff
@@ -37,36 +37,21 @@ def configure_data(self, cfg, data_cfg, training, **kwargs):
                 )
 
     def configure_task(self, cfg, training, **kwargs):
```
kprokofi, line 39: Why do we inherit  `SemiSLDetectionStage` from IncrDetectionStage? I thought this PR should refactor inheritance and create base Stage. Other stages will be inherited from this Base one.

kprokofi (commented):
Overall, seems structure for semisl is broken. SemiSL Stage should be inherited from base Stage -> then we can get rid of `configure_task` duplication

otx/mpa/det/incremental/inferrer.py
jaegukhyun: IncrDetectionStage.__init__() and DetectionInferrer.__init__() is same. They both initialize DetectionStage

otx/algorithms/detection/tasks/inference.py
jaegukhyun: Revise it, thanks

otx/mpa/det/incremental/trainer.py
jaegukhyun: I missed that. Thanks. I removed them

otx/mpa/det/incremental/__init__.py
jaegukhyun: I removed them, thanks

otx/mpa/det/semisl/stage.py
jaegukhyun: Great point, I have a mistake here. Thank you for your comment

supersoob (dismissed):
LGTM!

supersoob (approved):

JihwanEom (approved):

sungmanc (approved):

PR #1502, opened 2023-01-09
[OTX] Add SupCon for semantic segmentation

sungchul1:
# This PR includes 
- Add supcon for semantic segmentation modules, configs
- Enable detcon when training w/ fp16
- Update selfsl & supcon data_pipeline to use `DefaultFormatBundle`

## TODOs

- [x] experiments for reproducibility and improvement
- [x] change target branch from `feature/selfsl/cls` to `feature/otx` after merging https://github.com/openvinotoolkit/training_extensions/pull/1406
- [ ] ~unit test~ 
→ It will be implemented in other PR because this PR is so huge and it's unit test depends on self-sl for semantic segmentation.
- [x] integration test
`tests/integration/cli/segmentation/test_segmentation.py::TestSegmentationCLI::test_otx_train_supcon`

## To be considered in the future
- If with incremental setting, `model.num_classes` is automatically set from `256` to `len(model_classes)`. if not, `model.num_classes` remains `256`. https://github.com/openvinotoolkit/training_extensions/blob/cc643d3dd4ab9f8811d1a8e9a77959b05725b318/otx/mpa/seg/stage.py#L136-L138

### DIFF
--- a/otx/algorithms/classification/tasks/inference.py
+++ b/otx/algorithms/classification/tasks/inference.py
@@ -328,7 +328,7 @@ class ClassificationInferenceTask(
                 ConfigDict(warmup_iters=warmup_iters) if warmup_iters > 0 else ConfigDict(warmup_iters=0, warmup=None)
             )
 
-        if params.enable_early_stopping and self._recipe_cfg.get("evaluation", None):  # type: ignore
+        if params.enable_early_stopping is not None and self._recipe_cfg.get("evaluation", None) is not None:
             early_stop = ConfigDict(
                 start=int(params.early_stop_start),
                 patience=int(params.early_stop_patience),
--- a/otx/algorithms/classification/tasks/inference.py
+++ b/otx/algorithms/classification/tasks/inference.py
@@ -328,7 +328,7 @@ class ClassificationInferenceTask(
                 ConfigDict(warmup_iters=warmup_iters) if warmup_iters > 0 else ConfigDict(warmup_iters=0, warmup=None)
             )
 
-        if params.enable_early_stopping:
+        if params.enable_early_stopping and self._recipe_cfg.get("evaluation", None):  # type: ignore
             early_stop = ConfigDict(
                 start=int(params.early_stop_start),
                 patience=int(params.early_stop_patience),
--- a/otx/algorithms/classification/tasks/inference.py
+++ b/otx/algorithms/classification/tasks/inference.py
@@ -330,7 +330,7 @@ class ClassificationInferenceTask(
 
         early_stop = False
         if self._recipe_cfg is not None:
-            if params.enable_early_stopping is not None and self._recipe_cfg.get("evaluation", None) is not None:
+            if params.enable_early_stopping and self._recipe_cfg.get("evaluation", None):
                 early_stop = ConfigDict(
                     start=int(params.early_stop_start),
                     patience=int(params.early_stop_patience),
--- a/otx/algorithms/classification/tasks/inference.py
+++ b/otx/algorithms/classification/tasks/inference.py
@@ -328,14 +328,14 @@ class ClassificationInferenceTask(
                 ConfigDict(warmup_iters=warmup_iters) if warmup_iters > 0 else ConfigDict(warmup_iters=0, warmup=None)
             )
 
-        if params.enable_early_stopping is not None and self._recipe_cfg.get("evaluation", None) is not None:
-            early_stop = ConfigDict(
-                start=int(params.early_stop_start),
-                patience=int(params.early_stop_patience),
-                iteration_patience=int(params.early_stop_iteration_patience),
-            )
-        else:
-            early_stop = False
+        early_stop = False
+        if self._recipe_cfg is not None:
+            if params.enable_early_stopping is not None and self._recipe_cfg.get("evaluation", None) is not None:
+                early_stop = ConfigDict(
+                    start=int(params.early_stop_start),
+                    patience=int(params.early_stop_patience),
+                    iteration_patience=int(params.early_stop_iteration_patience),
+                )
 
         if self._recipe_cfg.runner.get("type").startswith("IterBasedRunner"):  # type: ignore
             runner = ConfigDict(max_iters=int(params.num_iters))
--- a/otx/algorithms/common/adapters/mmcv/hooks.py
+++ b/otx/algorithms/common/adapters/mmcv/hooks.py
@@ -707,11 +707,19 @@ class SwitchPipelineHook(Hook):
 
     @check_input_parameters_type()
     def before_train_epoch(self, runner: BaseRunner):
+        if self.interval == 1:
+            # Always keep `TwoCropTransform` enabled.
+            return
+            
         dataset = self.get_dataset(runner)
         self.decide_pipeline(dataset.pipeline.transforms)
 
     @check_input_parameters_type()
     def after_train_iter(self, runner: BaseRunner):
+        if self.interval == 1:
+            # Always keep `TwoCropTransform` enabled.
+            return
+            
         if self.cnt < self.interval-1:
             # Instead of using `runner.every_n_iters` or `runner.every_n_inner_iters`,
             # this condition is used to compare `self.cnt` with `self.interval` throughout the entire epochs.
@@ -722,9 +730,9 @@ class SwitchPipelineHook(Hook):
             for transform in dataset.pipeline.transforms:
                 if transform.__class__.__name__ == "TwoCropTransform":
                     if not transform.is_both:
-                        # if self.cnt == self.interval-1, there are two cases,
-                        # 1. self.cnt was updated in L696, so is_both must be on.
-                        # 2. if iter was already conducted, is_both must be off.
+                        # If `self.cnt` == `self.interval`-1, there are two cases,
+                        # 1. `self.cnt` was updated in L704, so `is_both` must be on for the next iter.
+                        # 2. if the current iter was already conducted, `is_both` must be off.
                         transform.is_both = True
                     else:
                         transform.is_both = False
--- a/otx/algorithms/common/adapters/mmcv/hooks.py
+++ b/otx/algorithms/common/adapters/mmcv/hooks.py
@@ -685,14 +685,15 @@ class TwoCropTransformHook(Hook):
     def __init__(self, interval: int = 1, by_epoch: bool = False):
         assert interval > 0, f"interval (={interval}) must be positive value."
         if by_epoch:
-            raise NotImplementedError(f"by_epoch is not implemented.")
+            raise NotImplementedError("by_epoch is not implemented.")
 
         self.interval = interval
         self.cnt = 0
 
     @check_input_parameters_type()
     def _get_dataset(self, runner: BaseRunner):
-        if hasattr(runner.data_loader.dataset, 'dataset'):
+        """Get dataset to handle `is_both`."""
+        if hasattr(runner.data_loader.dataset, "dataset"):
             # for RepeatDataset
             dataset = runner.data_loader.dataset.dataset
         else:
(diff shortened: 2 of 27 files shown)


sungchul1:
- [x] `2023-02-01T04:41:22.7597434Z FAILED tests/integration/cli/classification/test_classification.py::TestMultiClassClassificationCLI::test_otx_train_selfsl[Custom_Image_Classification_EfficinetNet-B0]`
- [ ] `2023-02-01T04:41:22.7598667Z FAILED tests/integration/cli/detection/test_instance_segmentation.py::TestInstanceSegmentationCLI::test_nncf_optimize[Custom_Counting_Instance_Segmentation_MaskRCNN_EfficientNetB2B]`
- [x] `2023-02-01T04:41:22.7600116Z FAILED tests/integration/cli/segmentation/test_segmentation.py::TestSegmentationCLI::test_otx_train_selfsl[Custom_Semantic_Segmentation_Lite-HRNet-18-mod2_OCR]`

otx/algorithms/classification/tasks/inference.py
```diff
@@ -328,7 +328,7 @@ def _init_recipe_hparam(self) -> dict:
                 ConfigDict(warmup_iters=warmup_iters) if warmup_iters > 0 else ConfigDict(warmup_iters=0, warmup=None)
             )
 
-        if params.enable_early_stopping:
+        if params.enable_early_stopping and self._recipe_cfg.get("evaluation", None):  # type: ignore
```
sungmanc: Why use `# type: ignore`? to avoid mypy?, If so, it's not desirable.

If this is for mypy, how about using below?
`if params.enable_early_stopping is not None and self._recipe_cfg.get("evaluation", None) is not None:`

otx/algorithms/common/adapters/mmcv/hooks.py
```diff
@@ -666,3 +666,84 @@ class ForceTrainModeHook(Hook):
     def before_train_epoch(self, runner):
         """Make sure to put a model in a training mode before train epoch."""
         runner.model.train()
+
+
+@HOOKS.register_module()
+class TwoCropTransformHook(Hook):
```
sungmanc, line 672: Just for curiosity, as far as I remember, the similar function was also implemented for Self-SL classification. Is it wrong? I just want to avoid the duplicate implementation

sungmanc: Could you convert the doc-string format as Google's one? We'll use this format as standard doc-string format

Ref.
https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

otx/algorithms/common/adapters/mmcv/hooks.py
```diff
@@ -666,3 +666,84 @@ class ForceTrainModeHook(Hook):
...
+            # for RepeatDataset
+            dataset = runner.data_loader.dataset.dataset
+        else:
+            dataset = runner.data_loader.dataset
+
+        return dataset
+
+    # pylint: disable=inconsistent-return-statements
+    @check_input_parameters_type()
+    def _find_two_crop_transform(self, transforms):
```
sungmanc: Need type hint

sungchul1:
`FAILED tests/integration/cli/detection/test_instance_segmentation.py::TestInstanceSegmentationCLI::test_nncf_optimize[Custom_Counting_Instance_Segmentation_MaskRCNN_EfficientNetB2B]`
Is there any way to solve this fail?

sungmanc:
> `FAILED tests/integration/cli/detection/test_instance_segmentation.py::TestInstanceSegmentationCLI::test_nncf_optimize[Custom_Counting_Instance_Segmentation_MaskRCNN_EfficientNetB2B]` Is there any way to solve this fail?

References:
https://github.com/openvinotoolkit/training_extensions/pull/1574#issuecomment-1413321980

According to the @cih9088 's opinion, This FAIL only occurred in CI. could you check this? @yunchu

cih9088:
> According to the @cih9088 's opinion, This FAIL only occurred in CI. could you check this? @yunchu

Can I access the CI server to debug this or at least container image for CI? ~~It happens in Geti as well and a somewhat related ticket has been filed. (CVS-102888)~~

otx/algorithms/common/tasks/training_base.py
```diff
@@ -57,7 +57,7 @@
 
 logger = get_logger()
 TRAIN_TYPE_DIR_PATH = {
-    TrainType.INCREMENTAL.name: ".",
+    TrainType.INCREMENTAL.name: "",
```
jaegukhyun: @harimkang Is it okay to change this. AFAIK there is specific reason for this from building side

otx/algorithms/segmentation/tasks/inference.py
```diff
@@ -187,7 +187,12 @@ def _init_recipe(self):
         logger.info(f"train type = {self._train_type}")
 
         if self._train_type in RECIPE_TRAIN_TYPE:
-            recipe = os.path.join(recipe_root, RECIPE_TRAIN_TYPE[self._train_type])
+            if self._train_type == TrainType.INCREMENTAL and self._hyperparams.learning_parameters.enable_supcon:
+                recipe = os.path.join(recipe_root, "supcon.py")
+                if "supcon" not in self._model_dir:
```
jaegukhyun, line 192: Is this check necessary? I think we always need to change our model directory to template_dir/supcon_model_dir

jaegukhyun (commented):
Generally LGTM. Thank you for your hard work. I left some commands. 
It would be nice to re-check ci integration test from rebased with current develop branch

otx/algorithms/common/adapters/mmcv/hooks.py
sungchul1: > Just for curiosity, as far as I remember, the similar function was also implemented for Self-SL classification. Is it wrong? I just want to avoid the duplicate implementation

This hook is to handle interval (: when supcon loss is applied) in `TwoCropTransform` implemented in `otx/algorithms/segmentation/adapters/mmseg/data/pipelines.py` and this `TwoCropTransform` is similar with the other one implemented by adria in `otx/mpa/modules/datasets/pipelines/transforms/twocrop_transform.py`. There are obvious differences between both modules, but it is possible to combine them to the single module. But, this process is not a scope of this PR and it can be proceeded when merging mpa to otx.

sungchul1: > Could you convert the doc-string format as Google's one? We'll use this format as standard doc-string format
> 
> Ref. https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

But other modules also use the same docstring format. I can modify it as you said, but it is not a scope of this PR and I think it would be better to modify it with other modules. What do you think?

otx/algorithms/common/tasks/training_base.py
sungchul1: > @harimkang Is it okay to change this. AFAIK there is specific reason for this from building side

If so, I'll return it.

otx/algorithms/common/adapters/mmcv/hooks.py
sungmanc: > > Could you convert the doc-string format as Google's one? We'll use this format as standard doc-string format
> > Ref. https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html
> 
> But other modules also use the same docstring format. I can modify it as you said, but it is not a scope of this PR and I think it would be better to modify it with other modules. What do you think?

We're currently focusing on adding type-hint and doc-string. Although pre-existed code didn't include proper type-hint and doc-string, we suggested adding them to new PRs. So, please do not consider pre-existed code and modify just yours.

otx/algorithms/segmentation/tasks/inference.py
sungchul1: I often occur the error that `self._model_dir` for training was remained until evaluation and `supcon` was added one more and `self._model_dir` became `.../supcon/supcon` whenever I ran the test.
SupCon for cls uses the similar condition to avoid overlapped path update.
https://github.com/openvinotoolkit/training_extensions/blob/8894d3d811ce1c770af3c1e3e74ec825a9ba16af/otx/algorithms/classification/tasks/inference.py#L380-L388
If there is another way to handle it, I can update it. Could you recommend it?

otx/algorithms/common/adapters/mmcv/hooks.py
sungchul1: Done.

otx/algorithms/classification/tasks/inference.py
sungchul1: Done.

sungchul1:
```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

otx/algorithms/classification/tasks/inference.py:331: error: "None" has no attribute "get"  [attr-defined]
Found 1 error in 1 file (checked 626 source files)
```
It seems that `self._recipe_cfg` is `None` and mypy cannot call `get` method. Is there anything to resolve it?

sungmanc:
``` python
if self._recipe_cfg is not None:
    if params.enable_early_stopping is not None and self._recipe_cfg.get("evaluation", None) is not None:
        sth....***
```

Or, how about the below?

``` python
if self._recipe_cfg is not None:
    if params.enable_early_stopping is not None and self._recipe_cfg.get("evaluation") is not None:
        sth....***
```

supersoob: I think we need to add `assert self._recipe_cfg is not None` at L233 as defined in _init_recipe_hparam, or check 'if self._recipe_cfg is not None' here.

otx/algorithms/segmentation/tasks/inference.py
supersoob: I glanced supcon support including classification, I feel like supcon is supported as train type already even if it has enable_supcon arg itself. (which has its own recipe file but the contents are almost same as incremental) I understand that model and data pipeline is totally different so it must be reloaded here but about other things, I personally recommend to refine all config for supcon in stage such as making _configure_supcon methods, injecting enable_supcon cfg in the incremental recipe file without making a recipe file. I think we need some discussion about that. But as of now, it's okay to handle these things (and consider classification supcon as well) after merging this PR.

otx/algorithms/classification/tasks/inference.py
sungchul1: Ah, `self._recipe_cfg` is None during mypy only, not main execution.
So, we don't need to check if `self_recipe_cfg` is None.
I don't know the reason....

supersoob: I know, mypy checks the all the possibility that recipe cfg is none because it was defined as none initially. I think it's okay to add ignore: attr-defined for skipping mypy check.

sungmanc: > Ah, `self._recipe_cfg` is None during mypy only, not main execution. So, we don't need to check if `self_recipe_cfg` is None. I don't know the reason....

mypy consider all possible cases and doesn't like dynamic type changing. So, we need to explicit check like `if self._recipe_cfg is not None`. Then, mypy also knows `self._recipe_cfg` could not be None. 

Actually, in the training base (`BaseTask`), `self._recipe_cfg` is defined as None.
https://github.com/openvinotoolkit/training_extensions/blob/97e13ccf0275cbdc4005610e2c6c4259d0723d39/otx/algorithms/common/tasks/training_base.py#L106

otx/algorithms/common/adapters/mmcv/hooks.py
sungchul1: Done.

otx/algorithms/classification/tasks/inference.py
sungchul1: Thanks for your kind explanation! @supersoob @sungmanc 
If adding `if self._recipe_cfg is not None` is more stable, I'll add it.

supersoob (dismissed):
LGTM. Thanks for your hard work. I think it's okay to merge if all ci tests are passed. Let's discuss refactoring later.

sungmanc (dismissed):

sungchul1:
@supersoob @sungmanc There was some pylint issue and I fixed it. Could you approve it again?

supersoob (approved):

sungmanc (approved):

PR #1503, opened 2023-01-10
[DEV] Replace imwrite/imread with imencode/imdecode in videoframe caching

JihwanEom:
- To unify `filename` to `f` expression, opencv method is replaced with imencode/imdecode.

### DIFF
--- a/external/model-preparation-algorithm/mpa_tasks/utils/data_utils.py
+++ b/external/model-preparation-algorithm/mpa_tasks/utils/data_utils.py
@@ -60,7 +60,7 @@ def get_image(results: Dict[str, Any], cache_dir: str, to_float32=False):
                 fcntl.flock(f, fcntl.LOCK_UN)
                 f.close()
             except Exception:
-                pass
+                return None
 
     def save_image_to_cache(img: np.array, filename: str):
         try:
@@ -76,7 +76,7 @@ def get_image(results: Dict[str, Any], cache_dir: str, to_float32=False):
                 fcntl.flock(f, fcntl.LOCK_UN)
                 f.close()
             except Exception:
-                pass
+                return None
 
     subset = results["dataset_item"].subset
     if is_training_subset(subset) and is_video_frame(results["dataset_item"].media):
--- a/external/model-preparation-algorithm/mpa_tasks/utils/data_utils.py
+++ b/external/model-preparation-algorithm/mpa_tasks/utils/data_utils.py
@@ -44,38 +44,48 @@ def get_image(results: Dict[str, Any], cache_dir: str, to_float32=False):
         return subset.name in ["TRAINING", "VALIDATION"]
 
     def load_image_from_cache(filename: str, to_float32=False):
-        f = open(filename, "rb")
-        fcntl.flock(f, fcntl.LOCK_SH)
         try:
+            f = open(filename, "rb")
+            fcntl.flock(f, fcntl.LOCK_SH)
             cached_img = np.asarray(bytearray(f.read()))
+            fcntl.flock(f, fcntl.LOCK_UN)
+            f.close()
             cached_img = cv2.imdecode(cached_img, cv2.IMREAD_COLOR)
             if to_float32:
                 cached_img = cached_img.astype(np.float32)
             return cached_img
         except Exception as e:
             logger.warning(f"Skip loading cached {filename} \nError msg: {e}")
-        finally:
-            fcntl.flock(f, fcntl.LOCK_UN)
-            f.close()
-    
+            try:  # to handle the case that f is not defined: failed open(filename, "rb")
+                fcntl.flock(f, fcntl.LOCK_UN)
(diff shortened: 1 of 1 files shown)


goodsong81 (approved):
LGTM, but can we bring `open()` into try/catch block to ignore I/O error?
(meaning just use original VideoFrame in case of I/O error, possibly due to full disk space, etc)

sungmanc (approved):

JihwanEom:
> LGTM, but can we bring `open()` into try/catch block to ignore I/O error?
> (meaning just use original VideoFrame in case of I/O error, possibly due to full disk space, etc)

Sure, reflected. Please review again.

external/model-preparation-algorithm/mpa_tasks/utils/data_utils.py
```diff
@@ -44,27 +44,31 @@ def is_training_subset(subset):
...
-                cached_img = cv2.imread(filename)
-                if to_float32:
-                    cached_img = cached_img.astype(np.float32)
-                return cached_img
-            except Exception as e:
-                logger.warning(f"Skip loading cached {filename} \nError msg: {e}")
-            finally:
-                fcntl.flock(f, fcntl.LOCK_UN)
+        f = open(filename, "rb")
+        fcntl.flock(f, fcntl.LOCK_SH)
```
goodsong81: Can we catch exceptions from these lines?

external/model-preparation-algorithm/mpa_tasks/utils/data_utils.py
```diff
@@ -44,27 +44,31 @@ def is_training_subset(subset):
...
-        with open(filename, "wb") as f:
-            fcntl.flock(f, fcntl.LOCK_EX)
-            try:
-                cv2.imwrite(filename, img=img)
-            except Exception as e:
-                logger.warning(f"Skip caching for {filename} \nError msg: {e}")
-            finally:
-                fcntl.flock(f, fcntl.LOCK_UN)
+        f = open(filename, "wb")
+        fcntl.flock(f, fcntl.LOCK_EX)
```
goodsong81: Same here.

JihwanEom:
Oh I misunderstood comment, I'll revise that

external/model-preparation-algorithm/mpa_tasks/utils/data_utils.py
JihwanEom: Revised. Please review again. Thank you!

external/model-preparation-algorithm/mpa_tasks/utils/data_utils.py
```diff
@@ -44,34 +44,48 @@ def is_training_subset(subset):
...
+                cached_img = cached_img.astype(np.float32)
+            return cached_img
+        except Exception as e:
+            logger.warning(f"Skip loading cached {filename} \nError msg: {e}")
+            try:  # to handle the case that f is not defined: failed open(filename, "rb")
                 fcntl.flock(f, fcntl.LOCK_UN)
-    
+                f.close()
+            except Exception:
+                pass
```
goodsong81: Can't we use `finally`?
Anyway, at least we can return `None` explicitly.

JihwanEom: Do you mean use `finally` instead of stacked `try~except`?
I intended to handle `f = open(filename, "rb")`, `f` will not be defined if failed to `open`.

Updated to return None explicitly! Thank you

external/model-preparation-algorithm/mpa_tasks/utils/data_utils.py
```diff
@@ -44,34 +44,48 @@ def is_training_subset(subset):
...
-                cached_img = cv2.imread(filename)
-                if to_float32:
-                    cached_img = cached_img.astype(np.float32)
-                return cached_img
-            except Exception as e:
-                logger.warning(f"Skip loading cached {filename} \nError msg: {e}")
-            finally:
+            cached_img = np.asarray(bytearray(f.read()))
+            fcntl.flock(f, fcntl.LOCK_UN)
+            f.close()
```
eunwoosh, line 52: I think it's better to separate fcntl.flock and file open part as each `try except`.

external/model-preparation-algorithm/mpa_tasks/utils/data_utils.py
```diff
@@ -44,34 +44,48 @@ def is_training_subset(subset):
...
+            try:  # to handle the case that f is not defined: failed open(filename, "rb")
                 fcntl.flock(f, fcntl.LOCK_UN)
-    
+                f.close()
+            except Exception:
+                return None
+
     def save_image_to_cache(img: np.array, filename: str):
-        with open(filename, "wb") as f:
+        try:
```
eunwoosh, line 66: same as above.

JihwanEom:
Close by #1555

PR #1504, opened 2023-01-10
[OTX] Simple auto-config class design and implementation for classification task

sungmanc:
# Summary
* This PR includes the simple auto-configuration functionality for Classification task with `otx build`
* By this PR, user can build with just single argument (i.e. `otx build --train-data-roots=${some_path}`)
* Will add other tasks (Detection, Segmentation)
* Will add other CLI (`otx train`)

# Description
Class Design and Pseudo code for workflow
image


otx/cli/tools/build.py
```diff
@@ -49,13 +53,68 @@ def main():
...
     args = parse_args()
+    args.task = args.task.upper() if args.task is not None else args.task
 
     builder = Builder()
 
-    # Build with task_type -> Create User workspace
     otx_root = get_otx_root_path()
 
-    if args.task and args.task.upper() in SUPPORTED_TASKS:
+    # Auto configuration & Datumaro Helper
```
harimkang: How about creating a variable that can check if this is auto-config?
`auto_config = (args.task is None) & args.train_data_roots`

harimkang: And my suggestion is below.

1. if auto_config -> detect task (task update)
2. if task:
    - create workspace (build_task_config)
3. if args.train_data_roots 
    1. generate datumaro dataset
    2. [if not val_data_roots] auto-split
    3. copy to workspace
    4. data.yaml update

...
If this is the case, it seems that we don't have to do folder management in build.py.
Scenario is below:
if auto_config -> create workspace & dataset generation & data.yaml update
if not auto_config
    - if task -> create workspace
    - if backbone -> generate backbone config
    - if backbone & model -> update model config
There may be something I missed.
What do you think?

sungmanc: @harimkang 

At first, making the `is_autoconfig_enabled` to check whether auto_config is on or not is a good idea.
For the workflow, I agree with you. However, I'm not sure that it is good to create a workspace inside of `build_task_config`.

When I checked that name(`build_task_config`), I didn't think this function also created the workspace.
So, for this, maybe we need to separate the function like `builder.create_workspace`.

sungmanc: > How about creating a variable that can check if this is auto-config? `auto_config = (args.task is None) & args.train_data_roots`

auto_config will be enabled when `args.train_data_roots is not None` regardless of `args.task`

harimkang: > @harimkang 
> 
> 
> 
> At first, making the `is_autoconfig_enabled` to check whether auto_config is on or not is a good idea.
> 
> For the workflow, I agree with you. However, I'm not sure that it is good to create a workspace inside of `build_task_config`.
> 
> 
> 
> When I checked that name(`build_task_config`), I didn't think this function also created the workspace.
> 
> So, for this, maybe we need to separate the function like `builder.create_workspace`.
> 
> 

What is the difference between `build_task_config` and `create_workspace` ?  To me, the two functions seem to be functions of the same role.  (Renaming that function looks good.)
If want to use workspaces that share the same name, how about adding an option to the existing function to set existing=True?

sungmanc: > > @harimkang
> > At first, making the `is_autoconfig_enabled` to check whether auto_config is on or not is a good idea.
> > For the workflow, I agree with you. However, I'm not sure that it is good to create a workspace inside of `build_task_config`.
> > When I checked that name(`build_task_config`), I didn't think this function also created the workspace.
> > So, for this, maybe we need to separate the function like `builder.create_workspace`.
> 
> What is the difference between `build_task_config` and `create_workspace` ? To me, the two functions seem to be functions of the same role. (Renaming that function looks good.) If want to use workspaces that share the same name, how about adding an option to the existing function to set existing=True?

IMO, `build_task_config` looks a kind of handling configurations about task, `create_workspace` seems just create the workspace, not handling any configurations. (`build_task_config()` --> `create_workspace()`) is my suggestion. 

`existing=True` is also good option for me :) 

Thanks for giving me detailed workflow, I'll reflect your comments and then please re-check the PRs

harimkang: yep good!  Thinking about it again, `build_task_config` can be used as a function that updates the configuratuon in the folder if existing is true.

`build_task_config` with exist=false -> create initial workspace with initial configurations
`build_task_config` with exist=true
-> update configs of workspace

After clearing this up, the `exist` option seems more necessary!

sungmanc: Add `exist` variable. if you have any other comments please let me know
https://github.com/openvinotoolkit/training_extensions/pull/1504/commits/0b6f07dd48323327cbd9f358faba84f1186a2df7

otx/cli/builder/builder.py
```diff
@@ -220,11 +220,11 @@ def build_task_config(
                 str(train_type_dir / "compression_config.json"),
             )
 
-        print(f"[otx build] Create OTX workspace: {str(workspace_path.absolute())}")
-        print(f"\tTask Type: {template.task_type}")
-        print(f"\tLoad Model Template ID: {template.model_template_id}")
-        print(f"\tLoad Model Name: {template.name}")
-        print(f"\tYou need to edit that file: {str(workspace_path.absolute() / 'data.yaml')}")
+        print(f"[*] Load Model Template ID: {template.model_template_id}")
```
kprokofi, line 213: why [*]?

otx/cli/tools/build.py
```diff
@@ -49,29 +53,88 @@ def main():
...
+            args.workspace_root += f"-{args.model}"
+    workspace_path = Path(args.workspace_root)
+
+    # Auto configuration & Datumaro Helper
+    # First, find the format of train dataset and import it by using DatasetManager
+    # Second, find the task type
+    # Third, check the whether val_data_roots is None or not None
+    # If val_data_roots is None, auto_split() is trigerred.
+    # Forth, save the dataset with it's configuraiton(.yaml) file
+    is_autoconfig_enabled = args.train_data_roots
```
kprokofi: I think we need to define a different condition to enable auto-config. When we don't have template passed to the CLI and train-data-root is available. If we use only train-data-root checking it is not right, since we can have template path and val-data-roots as well to run ordinary training

otx/core/config/manager.py
```diff
@@ -0,0 +1,110 @@
...
+            "SEGMENTATION": ["common_semantic_segmentation", "voc", "cityscapes", "ade20k2017", "ade20k2020"],
+        }
+
+    def get_task_type(self, data_format: str) -> str:
+        """Detect task type.
+
+        For some datasets (i.e. COCO, VOC, MVTec), can't be fully automated.
+        Because those datasets have several format at the same time.
+        (i.e. for the COCO case, object detection and instance segmentation annotations coexist)
+        In this case, the task_type will be selected to default value.
```
kprokofi, line 64: what is the default value?

otx/core/config/manager.py
```diff
@@ -0,0 +1,110 @@
...
+
+        For action tasks, currently action_classification is default.
+
+        If Datumaro supports the Kinetics, AVA datasets, MVTec, _is_cvat_format(), _is_mvtec_format()
+        functions will be deleted.
+        """
+
+        task = ""
+        if data_format == "multi-cvat":
```
kprokofi: why do we check it here if we store this in `self.task_data_dict` and  it can be checked in the loop below?

kprokofi (commented):
first round, few comments

otx/cli/builder/builder.py
sungmanc: No special reason, just want to pretty log :)

otx/core/config/manager.py
sungmanc: For the COCO, VOC detection is default and I'll remove INSTANCE_SEGMENTATION task because Datumaro can't auto split for the instance segmentation dataset

otx/core/config/manager.py
sungmanc: Agree, it can be deleted. Thanks !

otx/cli/tools/build.py
sungmanc: This block is dependent on `template` and `task` arguments because it just find the `data_format`, `task_type`, and `auto_split` and it'll ignore auto-detected `task_format` if user input their `--task` argument directly.

Let's assume the user's input arguments.  
Left: User's input argument, Right: configuration setting (A: auto-config, U: user input)

1. --train-data-roots --> train dataset (A) + val dataset (A) + task_type (A) + template (A)
2. --train-data-roots + --val-data-roots --> train dataset (U) + val dataset (U) + task_type (A) + template (A)
3. --train-data-roots + --task --> train_dataset (A) + val_dataset (A) + task_type (U) + template (A)
4. --train-data--roots + --task + --template --> train_dataset (A) + val_dataset (A) + task_type (U) + template (U)

harimkang: sorry, but I still didn't understand that `train_data_roots` is always set to autoconfig when it comes in.
If `--task` and `--train-data-roots` come in together, wouldn't it be possible to reduce task detection?
Isn't the time complexity of the function that detects the task in the dataset high?

otx/cli/tools/build.py
```diff
@@ -49,29 +61,85 @@ def main():
...
+        print(f"[*] Train task type: {train_task_type}")
+        if args.task:
+            assert (
+                args.task == train_task_type
+            ), f"Dataset format({train_data_format}) can't be used for {args.task} task."
+
+        # Overwrite the args.task to train_task_type to select default template
+        args.task = train_task_type if args.task is None else args.task
+
+        # Create workspace
```
harimkang: Couldn't we get out of the if statement related to auto config from here?
I think lines 100-135 have little to do with auto-config.
What do you think?

otx/cli/tools/build.py
```diff
@@ -49,29 +61,85 @@ def main():
...
+        args.workspace_root = set_workspace(args.workspace_root, args.task, args.model)
```
harimkang: If this code leave out the if statement like my comment, I think this can be replaced with a form that uses `build_task_config`.

otx/cli/builder/builder.py
```diff
@@ -155,12 +156,11 @@ def build_task_config(
...
-            workspace_path = f"./otx-workspace-{task_type}"
-            if model_type:
-                workspace_path += f"-{model_type}"
-        workspace_path = workspace_path if isinstance(workspace_path, Path) else Path(workspace_path)
-        workspace_path.mkdir(exist_ok=False)
+        # Check whether the workspace is existed or not
+        # FIXME: better variable name? is_autoconfig_enabled?
+        workspace_path = Path(workspace_path)
+        if not exist:
+            workspace_path.mkdir(exist_ok=False)
```
harimkang: It looks like this could be replaced with `workspace_path.mkdir(exist_ok=exist)` without needing an if statement.

otx/cli/tools/build.py
sungmanc: > sorry, but I still didn't understand that `train_data_roots` is always set to autoconfig when it comes in. If `--task` and `--train-data-roots` come in together, wouldn't it be possible to reduce task detection? Isn't the time complexity of the function that detects the task in the dataset high?

Maybe we need to discuss for user's use case. Thanks for your opinion @harimkang , @kprokofi 
Here is what I wanted for this auto-configuration functionality.

If the user inputs `--train-data-roots`, then the user might want to `automatic auto split` for the dataset.
So, if the user only inputs `--train-data-roots`, I think OTX build must trigger the `auto_config`.

In this case, if the user additionally gives `task` or `model`, then OTX build overwrites the `task`, and `mode` to pre-detected ones.
That's the reason why I use only `--train-data-roots` to trigger the `auto config`.

BTW, there are also some users who want to not use `auto-config` even if they give only `--train-data-roots`.
So, in this circumstance, I think we need to think more about what is our default setting.

otx/core/data/manager/dataset_manager.py
```diff
@@ -0,0 +1,142 @@
...
+import os
+from typing import List, Tuple
+
+import datumaro
+from datumaro.components.dataset import Dataset, DatasetSubset
+from datumaro.components.dataset_base import DatasetItem
+from datumaro.plugins.splitter import Split
+
+
+class DatasetManager:
```
eunwoosh, line 17: If it doesn't need class format, how about defining them as just functions.

otx/core/data/manager/dataset_manager.py
```diff
@@ -0,0 +1,142 @@
...
+        return data_item.media.path
+
+    @staticmethod
+    def get_classification_label(data_item: DatasetItem) -> str:
+        """Returns the classfication label (multi-class)."""
+        assert len(data_item.annotations) == 1, "Not implemented for multi-label"
+        return data_item.annotations[0].label
+
+    @staticmethod
+    def export_dataset(dataset: Dataset, output_dir: str, data_format: str, save_media=True):
```
eunwoosh, line 74: Is there any reason to wrap a single line as a function? I think just origin single line also has good readability.

otx/core/data/manager/dataset_manager.py
```diff
@@ -0,0 +1,142 @@
...
+        """Export the Datumaro Dataset."""
+        return Dataset.export(dataset, output_dir, data_format, save_media=save_media)
+
+    @staticmethod
+    def import_dataset(data_root: str, data_format: str) -> dict:
```
eunwoosh, line 79: same as above.

otx/core/data/manager/dataset_manager.py
```diff
@@ -0,0 +1,142 @@
...
+        cvat_format = sorted(["images", "annotations.xml"])
+        for sub_folder in os.listdir(path):
+            # video_0, video_1, ...
+            sub_folder_path = os.path.join(path, sub_folder)
+            # files must be same with cvat_format
+            if os.path.isdir(sub_folder_path):
+                files = sorted(os.listdir(sub_folder_path))
+                if files != cvat_format:
+                    return False
+        return True
```
eunwoosh, line 118: How about adding exception handling? ex1) empty direcotry, ex2) path not exist, etc.

otx/core/data/manager/dataset_manager.py
```diff
@@ -0,0 +1,142 @@
...
+        |--ground_truth
+        |--train
+        |--test
+
+        will be deprecated soon.
+        """
+
+        mvtec_format = sorted(["ground_truth", "train", "test"])
+        folder_list = []
+        for sub_folder in os.listdir(path):
```
eunwoosh, line 137: same as above

otx/cli/utils/tests.py
```diff
@@ -728,3 +729,24 @@ def otx_build_backbone_testing(root, backbone_args):
     assert (
         model_config["model"]["backbone"]["type"] == backbone
     ), f"{model_config['model']['backbone']['type']} != {backbone}"
+
+
+def otx_build_auto_config(root, otx_dir, args):
```
eunwoosh: I know other test function doesn't have type hinting, but how about adding it to be good starting point to increase a readability?

otx/cli/utils/tests.py
```diff
@@ -728,3 +729,24 @@ def otx_build_backbone_testing(root, backbone_args):
...
+    command_line = [
+        "otx",
+        "build",
+    ]
+
+    for option in args:
```
eunwoosh: how about changing this code like `for option, val in args.items():`?

otx/core/config/manager.py
```diff
@@ -0,0 +1,109 @@
...
+
+        If Datumaro supports the Kinetics, AVA datasets, MVTec, _is_cvat_format(), _is_mvtec_format()
+        functions will be deleted.
+        """
+
+        task = ""
+        # pick task type
+        for task_key, data_value in self.task_data_dict.items():
+            if data_format in data_value:
+                task = task_key
```
eunwoosh:
```suggestion
               return task
```

otx/core/config/manager.py
```diff
@@ -0,0 +1,109 @@
...
+    def write_data_with_cfg(self, datasets: Dict[str, IDataset], data_format: str, workspace_dir: str):
+        """Save the splitted dataset and data.yaml to the workspace."""
+        default_data_folder_name = "dataset"
+
+        data_config = self._create_empty_data_cfg()
+        for phase, dataset in datasets.items():
+            dst_dir_path = os.path.join(
+                workspace_dir,
+                default_data_folder_name,
+                str(phase),
```
eunwoosh: As typehinting show, phase is string already. Is there any reason to call str again?

otx/cli/builder/builder.py
```diff
@@ -208,10 +208,11 @@ def build_task_config(
         template_config.dump(str(workspace_path / "template.yaml"))
 
         # Create Data.yaml
-        data_subset_format = {"ann-files": None, "data-roots": None}
-        data_config = {"data": {subset: data_subset_format.copy() for subset in ("train", "val", "test")}}
-        data_config["data"]["unlabeled"] = {"file-list": None, "data-roots": None}
-        mmcv.dump(data_config, str(workspace_path / "data.yaml"))
+        if (workspace_path / "data.yaml").exists() is False:
```
eunwoosh:
```suggestion
        if not (workspace_path / "data.yaml").exists():
```

otx/cli/builder/builder.py
```diff
@@ -208,10 +208,11 @@ def build_task_config(
...
+            data_subset_format = {"ann-files": None, "data-roots": None}
```
eunwoosh: I think this code part is exactly same as `ConfigManager._create_empty_data_cfg()`. How about making it static function and reusing it?

otx/cli/tools/build.py
```diff
@@ -19,17 +19,29 @@
...
 
 from otx.cli.builder import Builder
 from otx.cli.utils.importing import get_otx_root_path
+from otx.core.config.manager import ConfigManager
+from otx.core.data.manager.dataset_manager import DatasetManager
 
 SUPPORTED_TASKS = ("CLASSIFICATION", "DETECTION", "INSTANCE_SEGMENTATION", "SEGMENTATION")
 SUPPORTED_TRAIN_TYPE = ("incremental", "semisl", "selfsl")
 
+def set_workspace(path, task, model):
```
eunwoosh: This function return path if path is not None. then, how about removing path as argument and call this function `if path is None`?

otx/core/config/manager.py
```diff
@@ -0,0 +1,137 @@
...
+        train_data_format = DatasetManager.get_data_format(train_data_root)
+        train_dataset = DatasetManager.import_dataset(data_root=train_data_root, data_format=train_data_format)
+        self.splitted_dataset = DatasetManager.auto_split(
+            task=train_task,
+            dataset=DatasetManager.get_train_dataset(train_dataset),
+            split_ratio=[("train", 0.8), ("val", 0.2)],
+        )
+        print("[*] Auto-split enabled.")
+
+    def write_data_with_cfg(
```
harimkang, line 88: How about not generating `data.yaml` in `builder.py` and always managing it here?

Here, it seems good to create a `data.yaml` containing `train_data_roots` or `val_data_roots` if they are updated, and create an empty `data.yaml` if they don't exist.

otx/cli/builder/builder.py
harimkang: This is what I said in the comment above. Deleting this if statement and its contents also looks good.

otx/core/data/manager/dataset_manager.py
sungmanc: `os.listdir` could show the proper error message in case of ex1, ex2.  So, I think we don't need to add exception handling.

otx/core/data/manager/dataset_manager.py
sungmanc: https://github.com/openvinotoolkit/training_extensions/pull/1504#discussion_r1086191982

otx/cli/builder/builder.py
sungmanc: Agree with this and the above comments, I also prefer this option.

otx/core/data/manager/dataset_manager.py
eunwoosh: In my opinion, it's better that the function returns false at those cases. Do you think raising error is better?

sungmanc: Yes, IMO, if the user gives an empty data directory or empty path, raising an error is better to analyze the root-cause because ex1, ex2 are kinds of abnormal cases

sungmanc:
@harimkang , @eunwoosh , I edited codes according to your comments, please check https://github.com/openvinotoolkit/training_extensions/pull/1504/commits/b2e44287703d51ed86ecd0d6c4c1b11574d740fb

harimkang (dismissed):
LGTM,
Recently, tests have been added (for builder), so there may be conflicts.
Could you please check it out?

otx/cli/builder/builder.py
harimkang: What do you think of this comment?

sungmanc:
> LGTM, Recently, tests have been added (for builder), so there may be conflicts. Could you please check it out?

I already rebase from latest feature/otx branch. Did you find any problems?
image

harimkang:
> LGTM, Recently, tests have been added (for builder), so there may be conflicts. Could you please check it out?

I meant unit tests and tests/integration/test_cli.py . If pass all of them, you can ignore them :)

sungmanc:
> > LGTM, Recently, tests have been added (for builder), so there may be conflicts. Could you please check it out?
> 
> I meant unit tests and tests/integration/test_cli.py . If pass all of them, you can ignore them :)

Got ALL PASS for tests/integration/test_cli.py :)

eunwoosh (dismissed):
LGTM. Thanks for your effort!

sungmanc:
Finished adding type-hints for `tests.py`, and removed useless arguments.

Tests for `test_cli.py`
image

sungmanc:
> LGTM. Thanks for your effort!

Thanks for your comments 👍

otx/cli/builder/builder.py
sungmanc: Since we changed the workflow of `builder.py`, `exist` argument is not needed anymore.
image

eunwoosh (dismissed):

otx/cli/builder/builder.py
```diff
@@ -140,9 +140,9 @@ class Builder:
     def build_task_config(
         self,
         task_type: str,
+        workspace_root: str,
         model_type: str = None,
         train_type: str = "incremental",
-        workspace_path: Union[Path, str] = None,
```
harimkang, line 145: As far as I remember, in unit tests, this is passed as a `Path`, but wouldn't this be a problem?

harimkang (dismissed):
LGTM!

otx/cli/builder/builder.py
sungmanc: If so, I'll change the unit-test soon !

sungmanc: I edited unit-test also :)

sungmanc:
@harimkang @eunwoosh , Sorry for asking for your approval frequently..! 
I edited unit-test cases for builder cli test.

eunwoosh (approved):

harimkang (approved):

PR #1507, opened 2023-01-10
[ENHANCE] Fix numpy_softmax NaN issue

goodsong81:
- Fixed indexing bug
- Enhanced numerical stability of softmax

Signed-off-by: Songki Choi <songki.choi@intel.com>


jaegukhyun (dismissed):

sungmanc (dismissed):

goodsong81:
Please do not merge. Still found NaN.

otx/algorithms/classification/adapters/openvino/model_wrappers/openvino_models.py
```diff
@@ -141,17 +141,8 @@ def sigmoid_numpy(x: np.ndarray):
...
-    # But it results in "ValueError: zero-size array to reduction operation maximum which has no identity"
-    inf_ind = np.isinf(x)
-    total_infs = np.sum(inf_ind)
-    if total_infs > 0:
-        x[inf_ind] = 1.0 / total_infs
-        x[~inf_ind] = 0
-    else:
-        x /= np.sum(x) + eps
-    return x
+    x = np.exp(x - np.max(x))
```
sovrasov, line 128: How this line is supposed to remove Nans from the input (if they come from the model)?
In my short example it produced all Nans:
```
>>> tst_a = np.array([0,9,2.2,float('nan')])
array([0. , 9. , 2.2, nan])
>>> np.exp(tst_a-np.max(tst_a))
array([nan, nan, nan, nan]) 
```
Also, exp(x) produces infs sometimes, so I added handling of it. In case if the input vector doesn't contain Nans, the result of exp(x) shouldn't contain them as well (we may have only overflow (inf) or underflow (0) here). The next operation is protected from zero division by +eps, so it doesn't produce Nans as well.

sungmanc: If there isn't NaN in the input, we can meet the overflow problem when the input value is too large. In this case, we can avoid the overflow issue by subtracting the max value although it couldn't guarantee the exact same value.

image

harimkang (approved):

jaegukhyun (approved):

PR #1469, opened 2022-12-28
[Documentation][feature/otx] Add Object Detection Tutorial

GalyaZalesskaya:
This is the first iteration for the Tutorial section and Object Detection tutorials. 
### What was done:
 
- [x] Updated the structure of Tutorials in initial templates, adding splitting for different tasks. Now the structure looks like this:
image

- [x] Wrote the Object Detection tutorial for training, validation, exporting and optimizing. 

- [x] Align proposed tutorial changes with changes in the [Explanation section PR](https://github.com/openvinotoolkit/training_extensions/pull/1467)
- [x] Clean up cross-pages links & typos & outputs results

TO BUILD DOCS:
```
pip install -U sphinx
pip install pydata-sphinx-theme
pip install sphinx-panels
pip install sphinx-copybutton
pip install sphinx-tabs
```
Then:
```
cd docs
make html
```

### For future PRs:
- The usage of Datumaro integration
- Quick Start
- Demo & Deploy tutorials

PR #1510, opened 2023-01-10
[OTX] Update OTX detection requirements

GalyaZalesskaya:
This PR contains a few code fixes, discovered during [documentation writing](https://github.com/openvinotoolkit/training_extensions/pull/1469). 

- [x] Added missing mmcls and timm packages to the detection requirements
- [x] Make it possible to pass annotation paths via `--data data.yaml` parameter even if it doesn't have `unlabeled` annotation

### DIFF
--- a/otx/cli/tools/train.py
+++ b/otx/cli/tools/train.py
@@ -174,7 +174,7 @@ def main():  # pylint: disable=too-many-branches
             "ann_file": data_config["data"]["val"]["ann-files"],
             "data_root": data_config["data"]["val"]["data-roots"],
         }
-    if data_config["data"]["unlabeled"]["data-roots"]:
+    if "unlabeled" in data_config["data"] and data_config["data"]["unlabeled"]["data-roots"]:
         data_roots["unlabeled_subset"] = {
             "data_root": data_config["data"]["unlabeled"]["data-roots"],
             "file_list": data_config["data"]["unlabeled"]["file-list"],
--- a/otx/cli/tools/train.py
+++ b/otx/cli/tools/train.py
@@ -186,7 +186,7 @@ def main():  # pylint: disable=too-many-branches
         train_data_roots=data_roots["train_subset"]["data_root"],
         val_data_roots=data_roots["val_subset"]["data_root"] if data_config["data"]["val"]["data-roots"] else None,
         unlabeled_data_roots=data_roots["unlabeled_subset"]["data_root"]
-        if "unlabeled" in data_config and data_config["data"]["unlabeled"]["data-roots"]
+        if "unlabeled" in data_config["data"] and data_config["data"]["unlabeled"]["data-roots"]
         else None,
     )
     dataset = dataset_adapter.get_otx_dataset()
--- a/otx/cli/tools/train.py
+++ b/otx/cli/tools/train.py
@@ -186,7 +186,7 @@ def main():  # pylint: disable=too-many-branches
         train_data_roots=data_roots["train_subset"]["data_root"],
         val_data_roots=data_roots["val_subset"]["data_root"] if data_config["data"]["val"]["data-roots"] else None,
         unlabeled_data_roots=data_roots["unlabeled_subset"]["data_root"]
-        if data_config["data"]["unlabeled"]["data-roots"]
+        if "unlabeled" in data_config and data_config["data"]["unlabeled"]["data-roots"]
         else None,
     )
     dataset = dataset_adapter.get_otx_dataset()
--- a/requirements/detection.txt
+++ b/requirements/detection.txt
@@ -1,5 +1,5 @@
 mmcv-full==1.6.1
 mmdet==2.25.1
 pytorchcv
-mmcls==0.24
+mmcls==0.24.0
 timm
--- a/requirements/detection.txt
+++ b/requirements/detection.txt
@@ -1,5 +1,5 @@
 mmcv-full==1.6.1
 mmdet==2.25.1
 pytorchcv
-mmcls==0.24.0
+mmcls==0.23.2
 timm
--- a/requirements/detection.txt
+++ b/requirements/detection.txt
@@ -1,3 +1,5 @@
 mmcv-full==1.6.1
 mmdet==2.25.1
 pytorchcv
+mmcls==0.24
+timm
--- a/requirements/detection.txt
+++ b/requirements/detection.txt
@@ -2,4 +2,4 @@ mmcv-full==1.6.1
 mmdet==2.25.1
 pytorchcv
 mmcls==0.23.2
-timm
+timm==0.6.12


requirements/detection.txt
```diff
@@ -1,3 +1,5 @@
 mmpycocotools
+mmcls==0.15.0
```
sungmanc: `mmcls` and `timm` are needed for `build.py` and `find.py` ?

otx/cli/tools/train.py
```diff
@@ -175,7 +175,7 @@ def main():  # pylint: disable=too-many-branches
             "ann_file": data_config["data"]["val"]["ann-files"],
             "data_root": data_config["data"]["val"]["data-roots"],
         }
-    if data_config["data"]["unlabeled"]["data-roots"]:
+    if "unlabeled" in data_config["data"] and data_config["data"]["unlabeled"]["data-roots"]:
```
sungmanc, line 177: The objective of this line is to avoid the error caused when "unlabeled" is not in `data_config["data"]`?
IMO, If there is no "unlabeled" key in `data_config["data"]`, this line still makes an error.

sungmanc (commented):
Left some comments, + could you use the upper character for the first character of commit?
(i.e. check if "unlabeled" in data_config --> Check if "unlabeled" in data_config)

otx/cli/tools/train.py
kprokofi: It will not raise an error since we have "lazy" if in Python.

requirements/detection.txt
kprokofi: Agree, why do we need `timm` and `mmcls` in detection task?

GalyaZalesskaya: The `timm` package is needed for [efficientnetv2 MMCV backbone](https://github.com/openvinotoolkit/training_extensions/blob/feature/otx/otx/algorithms/common/adapters/mmcv/models/backbones/efficientnetv2.py). Although we don't use it in detection templates it is called during MMCV models initialization

GalyaZalesskaya: `mmcls` is called during MPA hooks initialization: https://github.com/openvinotoolkit/training_extensions/blob/adc16410ced3ecfb7524cc00d5d5afd76ebaa99c/otx/mpa/modules/hooks/recording_forward_hooks.py#L22

kprokofi: hmm, seems strange to me. We need to get rid of this. Because we support stand alone environments for each task - but it is required to install mmcls for detection task.

otx/cli/tools/train.py
GalyaZalesskaya: @sungmanc Yes. If we pass `--data data.yaml` parameter with `data.yaml` without `unlabeled` annotation, this field will mot present in data_config because of [configure_dataset implementation](https://github.com/openvinotoolkit/training_extensions/blob/adc16410ced3ecfb7524cc00d5d5afd76ebaa99c/otx/cli/utils/config.py#L39), so the error will occur. But it will work if we pass the `--train-ann-files` and `--train-data-roots` paths.
 
No, it doesn't end up with error. The same construction with the checking field in dict and then accessing it in the same `if` condition is used in [configure_dataset implementation](https://github.com/openvinotoolkit/training_extensions/blob/adc16410ced3ecfb7524cc00d5d5afd76ebaa99c/otx/cli/utils/config.py#L52) as well

GalyaZalesskaya:
> Left some comments, + could you use the upper character for the first character of commit? (i.e. check if "unlabeled" in data_config --> Check if "unlabeled" in data_config)

Thank you, Sungman. I did it, but can you clarify if it is the common rule for committing to OTX?

requirements/detection.txt
sungmanc: > hmm, seems strange to me. We need to get rid of this. Because we support stand alone environments for each task - but it is required to install mmcls for detection task.

Agree with this opinion, we need to remove this later. But, not this time.

sungmanc:
> > Left some comments, + could you use the upper character for the first character of commit? (i.e. check if "unlabeled" in data_config --> Check if "unlabeled" in data_config)
> > Thank you, Sungman. I did it, but can you clarify if it is the common rule for committing to OTX?

No special rules for us, but the one thing is that we just use a capital letter for the first character of commit and PR title.

requirements/detection.txt
harimkang: As far as I know, `timm` is installed from `mmcls>=0.16.0`. I know that upgrading the mmcls version will solve. (we don't need to install timm seperately)
BTW, Isn't this a problem that needs to be solved by cleaning up the mpa import? -> Importing `mmcls` from `recording_forward_hooks` itself could be wrong.  (detection do not need to use `mmcls`)

GalyaZalesskaya: @harimkang Thank you for your note, I checked on practice, and `timm` isn't installed with higher `mmcls`. Also, it doesn't have it in [requirements](https://github.com/open-mmlab/mmclassification/tree/master/requirements).
Regarding `mmcls` installation, previously it was installed during `mpa`  installation in [detection requirements](https://github.com/openvinotoolkit/training_extensions/blob/be582a20581d73eacfcb8036f473e4a9f140f951/requirements/detection.txt), so it wasn't so obvious what we need `mmcls` to run detection. 
So yes, import clean up in mpa import can be an option.

sungmanc (dismissed):

requirements/detection.txt
```diff
@@ -3,3 +3,5 @@ mmcv-full==1.6.1
 #otxdet==0.3.2
 mmdet==2.25.1
 pytorchcv
+mmcls==0.23.2
+timm
```
JihwanEom: Don't we need to specify the version for stability?

GalyaZalesskaya: Good point, done

sungmanc (changes requested):
Could you please rebase this PR?

+ according to @kprokofi 's opinion, please also edit this
image

GalyaZalesskaya:
> Could you please rebase this PR?
> 
> * according to @kprokofi 's opinion, please also edit this
>   image

@sungmanc @kprokofi Thank you for your comments, I rebased and added changes to Datumaro part as well

otx/cli/tools/train.py
```diff
@@ -186,7 +186,7 @@ def main():  # pylint: disable=too-many-branches
         train_data_roots=data_roots["train_subset"]["data_root"],
         val_data_roots=data_roots["val_subset"]["data_root"] if data_config["data"]["val"]["data-roots"] else None,
         unlabeled_data_roots=data_roots["unlabeled_subset"]["data_root"]
-        if data_config["data"]["unlabeled"]["data-roots"]
+        if "unlabeled" in data_config and data_config["data"]["unlabeled"]["data-roots"]
```
kprokofi:
```suggestion
        if "unlabeled" in data_config["data"] and data_config["data"]["unlabeled"]["data-roots"]
```

kprokofi (changes requested):

sungmanc (approved):
Thanks for quick fix :)

kprokofi (approved):

PR #1519, opened 2023-01-12
[OTX][TEST] Add unit test for `otx.cli.builder`

harimkang:
## Summary
Add new unit-test for functions in `otx.cli.builder`
Consolidate `conftest.py` into one
Replace `os.path` with `pathlib.Path` (otx/cli/builder/builder.py).

## Results
Target code coverage for `otx.cli.builder` reached. (>=80%)
```
pytest --cov-report term-missing --cov=otx/cli/builder/ tests/unit/cli/builder/test_cli_builder.py
---------- coverage: platform linux, python 3.9.13-final-0 -----------
Name                          Stmts   Miss  Cover   Missing
-----------------------------------------------------------
otx/cli/builder/__init__.py       2      0   100%
otx/cli/builder/builder.py      172      5    97%   157-159, 305, 314
-----------------------------------------------------------
TOTAL                           174      5    97%
========== 24 passed, 4 warnings in 24.56s ==========
```


tests/unit/cli/builder/test_cli_builder.py
```diff
@@ -0,0 +1,365 @@
...
+
+        <b>Steps</b>
+        1. Generate backbone config file (mmcls.MMOVBackbone)
+        2. Raise ValueError with wrong output_path
+        """
+        otx_builder = Builder()
+
+        # 1. Generate backbone config file (mmcls.MMOVBackbone)
+        tmp_backbone_path = os.path.join(tmp_dir_path, "backbone.yaml")
+        backbone_type = "mmcls.MMOVBackbone"
```
eugene123tw: Instead of hardcoding it, could we move this variable with pytest.mark.parametrize decorator? It'd be easier to extend if we have different backbone types in the future.

```py
@pytest.mark.parametrize("tmp_dir_path, backbone", [tmp_dir_path, "mmcls.MMOVBackbone"] )
def test_builder_build_backbone_config(self, tmp_dir_path: str, backbone: str):
```

tests/unit/cli/builder/test_cli_builder.py
```diff
@@ -0,0 +1,365 @@
...
+        # 2. Raise ValueError with wrong output_path
+        tmp_backbone_path = os.path.join(tmp_dir_path, "wrong.path")
+        with pytest.raises(ValueError):
+            otx_builder.build_backbone_config(backbone_type, tmp_backbone_path)
+
+    @e2e_pytest_unit
+    def test_builder_build_model_config(self, tmp_dir_path):
+        """
+        <b>Description:</b>
+        Check "Builder.build_model_config" function that update model config with new backbone is working well
```
eugene123tw: Probably not related to the unit test itself, but the design.

so this function:`Builder.build_model_config(self, model_config_path, backbone_config_path)` builds a model from the model config path. And is it true that it also updates the backbone of the model? 

Wouldn't it makes more sense if we build a model from OTX model builder and call `otx_builder.udpate_backbone(backbone_type: str)` to update the backbone in the model instance?

tests/unit/cli/builder/test_cli_builder.py
```diff
@@ -0,0 +1,365 @@
...
+from otx.cli.utils.importing import get_otx_root_path
+from otx.mpa.utils.config_utils import MPAConfig
+from tests.test_suite.e2e_test_system import e2e_pytest_unit
+
+
+class TestOTXCLIBuilder:
+    @e2e_pytest_unit
+    def test_builder_build_task_config(self, tmp_dir_path):
+        """
+        <b>Description:</b>
```
eugene123tw: Is it a limitation of our test CI so we have to write HTML tags for formatting?

tests/unit/cli/builder/test_cli_builder.py
harimkang: It looks good. I will apply this for the future. :)

tests/unit/cli/builder/test_cli_builder.py
harimkang: Yes, the inputs (model.py & backbone.yaml) and outputs (model.py) of this are configurations, but actually build backbone, measure the channels using fake data, and update the necessary config value of the model.
we can think of it as a function that `merges the model and the backbone` and `emits the updated model`.

It's probably okay to configure it with a function named `update_backbone`, but the concern is that we aim to update based on the `backbone's configuration file`, not targeting the `backbone_type`. so it doesn't seem appropriate to receive only `backbone_type` as an argment. Or it seems that it can be divided into separate functions inside.

To sum up, there are two ideas for a more understandable design: 
1. Rename this function to `update_backbone` (same inside code)
2. How to build a file config and put a separate function inside to reduce redundant code and split it into two functions. (`build_model_config` -> call `update_backbone(backbone_type: str)`) - I'll have to check if this is possible or good design while implementing it myself.

What is your recommendation?

tests/unit/cli/builder/test_cli_builder.py
harimkang: I know there is no problem with CI. Currently, I brought the code docstring format of `tests/unit/api` as it is, but this PR can be a starting point because there is no fixed thing, and OTX does not actually have a unit-test. 
Is there a format you recommend?

harimkang:
@eugene123tw thanks for the comment! OTX unit-test is currently almost non-existent, and it is just the beginning. I think this PR could be a format for unit-test. If you have a better design or format, please recommend!

tests/unit/cli/builder/test_cli_builder.py
harimkang: Reflected on this, https://github.com/openvinotoolkit/training_extensions/pull/1519/commits/73bb47915e2071fbbcde0174c3087588032330c5

tests/unit/cli/builder/test_cli_builder.py
```diff
@@ -0,0 +1,365 @@
...
+            "model_type": model_type,
+            "workspace_path": workspace_path,
+            "otx_root": otx_root,
+        }
+        otx_builder.build_task_config(**inputs)
+        assert os.path.exists(os.path.join(workspace_path, "template.yaml"))
+        template = MPAConfig.fromfile(os.path.join(workspace_path, "template.yaml"))
+        assert template.name.lower() == model_type
+
+        # 6. Raise ValueError when build workspace with wrong model_type argments
```
eunwoosh: I think there are many test cases in a single test case. how about separating them as separate test case?

tests/unit/cli/builder/test_cli_builder.py
```diff
@@ -0,0 +1,365 @@
...
+
+        <b>Steps</b>
+        1. Create Classification custom workspace
+        2. Raising Error of building workspace with already created path
+        3. Update hparam.yaml with train_type="selfsl"
+        4. Raising ValueError with wrong train_type
+        5. Build workspace with model_type argments
+        6. Raise ValueError when build workspace with wrong model_type argments
+        """
+        otx_builder = Builder()
```
eunwoosh:
```
otx_builder = Builder()
otx_root = get_otx_root_path()
```
how about making these parts as fixtures?

tests/unit/cli/builder/test_cli_builder.py
```diff
@@ -0,0 +1,365 @@
...
+
+        # 4. Update model config without backbone's out_indices
+        backbone_config = mmcv.load(tmp_backbone_path)
+        assert backbone_config["backbone"].pop("out_indices") == (3,)
+        mmcv.dump(backbone_config, tmp_backbone_path)
+        otx_builder.build_model_config(tmp_model_path, tmp_backbone_path)
+        updated_model_config = MPAConfig.fromfile(tmp_model_path)
+        assert "out_indices" in updated_model_config.model.backbone
+
+        # 5. Update model config with backbone's pretrained path
```
eunwoosh: As I said above, how about separating many tests as each test cases?

tests/unit/cli/builder/test_cli_builder.py
```diff
@@ -0,0 +1,365 @@
...
+        2. Check backbone input_size change to 128
+        """
+
+        # 1. Check default input_size = 64
+        backbone = MockBackbone()
+        expected_result = [64, 64]
+        out_channels = get_backbone_out_channels(backbone)
+        assert out_channels == expected_result
+
+        # 2. Check backbone input_size change
```
eunwoosh: As I said above, how about separating many tests as each test cases?

tests/unit/cli/builder/test_cli_builder.py
```diff
@@ -0,0 +1,365 @@
...
+        2. Update required Args in Backbone (Check Missing Args)
+        3. Update Args with out_indices
+        4. Update backbone using the backbone name from the backbone list (Check updating with options)
+        5. Update backbone using the backbone name from the backbone list (Check updating without options)
+        6. Raise ValueError with unexpected backbone
+        """
+        registry = MockRegistry(name="test")
+        backbone = MockBackbone
+
+        # 1. Update backbone args (Case without Required Args)
```
eunwoosh: As I said above, how about separating many tests as each test cases?

tests/unit/cli/builder/test_cli_builder.py
```diff
@@ -0,0 +1,365 @@
...
+
+        <b>Steps</b>
+        1. Remove model.neck.in_channels (GlobalAveragePooling)
+        2. Update model.neck.in_channels
+        3. Update model.decode_head.in_channels & in_index (segmentation case)
+        4. Update model.head.in_channels
+        5. Raise NotImplementedError with unexpected model key
+        """
+
+        # 1. Remove model.neck.in_channels (GlobalAveragePooling)
```
eunwoosh: As I said above, how about separating many tests as each test cases?

jaegukhyun (dismissed):
LGTM I think it is a good start to define unit test of OTX.
In addition I also have a doubt for HTML style docstring. I think we can change this time

eunwoosh (commented):
Thank you for effort. This can be good starting point to implement the unit tests.
I think if you separate some code as fixture, code can be more readable and less duplicated.

tests/unit/cli/builder/test_cli_builder.py
harimkang: This is a test design choice. I first proposed a 1 test function design for 1 function. What you say makes sense. we can do that. but i think, It seems that discussions between groups are necessary.

harimkang: Anyone please comment here!

tests/unit/cli/builder/test_cli_builder.py
harimkang: Those objects are not considered general objects that are only used in a small number of test cases.
I'm afraid if I subtract this as fixtures, end up with too many fixtures later. I think we need to narrow the criteria down to something more general.

harimkang:
@eunwoosh Thank you for review & the great suggestion. I understand all of your comments and I can make that happen. It seems that the consent of the group agreement is required. (Or we can decide..?) It seems that both the your suggested method and the current method have pros and cons.

tests/unit/cli/builder/test_cli_builder.py
eunwoosh: Actually, when I learned unit test in the book, I read that "One unit test case should test only one thing.".
Based on this, I think separation is better choice. but as you said, this is a design choice in a way. So, it's good to listen other opinions.

sungmanc: I prefer @eunwoosh 's opinion. ("One test with one method call and one statement that asserts or verifies the behavior was as expected")

harimkang: I reflected the opinions of Eunwoo and Sungman. Please confirm. https://github.com/openvinotoolkit/training_extensions/pull/1519/commits/fc590a79fddc0d6c695a186f904762ae4aea5f92

I split the 6 test functions into 24.

tests/unit/cli/builder/test_cli_builder.py
harimkang: This is also bundled as an internal fixture of the class. https://github.com/openvinotoolkit/training_extensions/pull/1519/commits/fc590a79fddc0d6c695a186f904762ae4aea5f92

tests/unit/cli/builder/test_cli_builder.py
harimkang: To avoid the docstring of the HTML tags method, the docstring of the existing code (in otx) has been borrowed as it is. https://github.com/openvinotoolkit/training_extensions/pull/1519/commits/fc590a79fddc0d6c695a186f904762ae4aea5f92

harimkang:
> LGTM I think it is a good start to define unit test of OTX. In addition I also have a doubt for HTML style docstring. I think we can change this time

Considering Eugene and your comments, I modified this to match the docstring of the original `otx` code.
https://github.com/openvinotoolkit/training_extensions/pull/1519/commits/fc590a79fddc0d6c695a186f904762ae4aea5f92

eunwoosh (dismissed):
Nice work!

tests/conftest.py
```diff
@@ -15,3 +20,20 @@
...
+
+
+@pytest.fixture(scope="session")
+def tmp_dir_path():
+    with TemporaryDirectory() as tmp_dir:
+        yield Path(tmp_dir)
+
+
+@pytest.fixture(autouse=True)
+def set_default_tmp_path(tmp_dir_path):
```
samet-akcay:
```suggestion
def set_default_tmp_path(tmp_dir_path: Path) -> Generator[None, None, None]:
```

jaegukhyun (dismissed):

tests/conftest.py
```diff
@@ -15,3 +20,20 @@
 
 def pytest_addoption(parser):
     otx_pytest_addoption_insertion(parser)  # noqa: F405
+
+
+@pytest.fixture(scope="session")
+def tmp_dir_path():
```
samet-akcay:
```suggestion
def tmp_dir_path() -> Generator[Path, None, None]:
```

tests/unit/cli/builder/test_cli_builder.py
```diff
@@ -0,0 +1,420 @@
...
+    <Steps>
+        1. Update model config with mmcls.ResNet backbone (default model.backbone: otx.OTXEfficientNet)
+        2. Raise ValueError with wrong model_config_path
+        3. Raise ValueError with wrong backbone_config_path
+        4. Update model config without backbone's out_indices
+        5. Update model config with backbone's pretrained path
+    """
+
+    @pytest.fixture(autouse=True)
+    def setup(self, tmp_dir_path):
```
samet-akcay: Is `tmp_dir_path` still a `Path` object? If not can we use Path object since it is much more convenient than `str` and `os`

samet-akcay:
```suggestion
    def setup(self, tmp_dir_path: Union[str, Path]) -> None:
```

tests/unit/cli/builder/test_cli_builder.py
```diff
@@ -0,0 +1,420 @@
...
+        self.otx_builder = Builder()
+        self.otx_root = get_otx_root_path()
+        self.tmp_dir_path = tmp_dir_path
```
samet-akcay:
```suggestion
        self.tmp_dir_path = tmp_dir_path if isinstance(tmp_dir_path, Path) else Path(tmp_dir_path)
```

tests/unit/cli/builder/test_cli_builder.py
```diff
@@ -0,0 +1,420 @@
...
+        """Create Classification custom workspace."""
+        workspace_path = os.path.join(self.tmp_dir_path, "test_builder_build_task_config_create_workspace")
+        inputs = {"task_type": "classification", "workspace_path": workspace_path, "otx_root": self.otx_root}
+        self.otx_builder.build_task_config(**inputs)
+        assert os.path.exists(workspace_path)
+        assert os.path.exists(os.path.join(workspace_path, "configuration.yaml"))
+        assert os.path.exists(os.path.join(workspace_path, "template.yaml"))
+        assert os.path.exists(os.path.join(workspace_path, "model.py"))
+        assert os.path.exists(os.path.join(workspace_path, "data.yaml"))
+        assert os.path.exists(os.path.join(workspace_path, "data_pipeline.py"))
```
samet-akcay: This function could be a lot shorter with the use of `Path`

```suggestion
    def test_builder_build_task_config_create_workspace(self) -> None:
        """Create Classification custom workspace."""
        workspace_path = self.tmp_dir_path / "test_builder_build_task_config_create_workspace"
        inputs = {"task_type": "classification", "workspace_path": workspace_path, "otx_root": self.otx_root}
        self.otx_builder.build_task_config(**inputs)
        
        assert workspace_path.exists()
        assert (workspace_path / "configuration.yaml").exists()
        assert (workspace_path / "template.yaml").exists()
        assert (workspace_path / "model.yaml").exists()
        assert (workspace_path / "data.yaml").exists()
        assert (workspace_path / "data_pipeline.yaml").exists()
```

tests/unit/cli/builder/test_cli_builder.py
```diff
@@ -0,0 +1,420 @@
...
+
+    @e2e_pytest_unit
+    def test_builder_build_task_config_reuse_same_path(self):
```
samet-akcay:
```suggestion
def test_builder_build_task_config_reuse_same_path(self) -> None:
```

tests/unit/cli/builder/test_cli_builder.py
```diff
@@ -0,0 +1,420 @@
...
+        """Raising Error of building workspace with already created path."""
+        workspace_path = os.path.join(self.tmp_dir_path, "test_builder_build_task_config_create_workspace")
```
samet-akcay: Would it be possible to switch to pathlib here too? Or does `build_task_config` always expect a `str`?

tests/unit/cli/builder/test_cli_builder.py
```diff
@@ -0,0 +1,420 @@
...
+        assert os.path.exists(os.path.join(workspace_path, "configuration.yaml"))
+        assert os.path.exists(os.path.join(workspace_path, "template.yaml"))
+        assert os.path.exists(os.path.join(workspace_path, "data.yaml"))
+        template = MPAConfig.fromfile(os.path.join(workspace_path, "template.yaml"))
+        expected_template_train_type = {"default_value": "SELFSUPERVISED"}
+        assert template.hyper_parameters.parameter_overrides.algo_backend.train_type == expected_template_train_type
+        model_dir = os.path.join(workspace_path, train_type)
+        assert os.path.exists(model_dir)
+        assert os.path.exists(os.path.join(model_dir, "model.py"))
+        assert os.path.exists(os.path.join(model_dir, "data_pipeline.py"))
```
samet-akcay:
```suggestion
    def test_builder_build_task_config_normal_train_type(self) -> None:
        """Update hparam.yaml with train_type="selfsl"."""
        workspace_path = self.tmp_dir_path / "test_builder_build_task_config_check_update_hparams"
        train_type = "selfsl"
        inputs = {
            "task_type": "classification",
            "train_type": train_type,
            "workspace_path": str(workspace_path),
            "otx_root": self.otx_root,
        }
        self.otx_builder.build_task_config(**inputs)
        assert workspace_path.exists()
        assert (workspace_path / "configuration.yaml")
        assert (workspace_path / "template.yaml")
        assert (workspace_path / "data.yaml")

        template = MPAConfig.fromfile(str(workspace_path, "template.yaml"))
        expected_template_train_type = {"default_value": "SELFSUPERVISED"}
        assert template.hyper_parameters.parameter_overrides.algo_backend.train_type == expected_template_train_type

        model_dir = workspace_path / train_type
        assert model_dir.exists()
        assert (model_dir / "model.py").exists()
        assert (model_dir / "data_pipeline.py").exists()
```

tests/unit/cli/builder/test_cli_builder.py
```diff
@@ -0,0 +1,420 @@
...
+
+    @e2e_pytest_unit
+    def test_builder_build_task_config_abnormal_train_type(self):
```
samet-akcay:
```suggestion
def test_builder_build_task_config_abnormal_train_type(self) -> None:
```

tests/unit/cli/builder/test_cli_builder.py
```diff
@@ -0,0 +1,420 @@
...
+        """Raising ValueError with wrong train_type."""
+        workspace_path = os.path.join(self.tmp_dir_path, "test_builder_build_task_config_abnormal_train_type")
```
samet-akcay:
```suggestion
        workspace_path = self.tmp_dir_path / "test_builder_build_task_config_abnormal_train_type"
```

tests/unit/cli/builder/test_cli_builder.py
```diff
@@ -0,0 +1,420 @@
...
+        train_type = "unexpected"
+        inputs = {
+            "task_type": "classification",
+            "train_type": train_type,
+            "workspace_path": workspace_path,
```
samet-akcay, line 110:
```suggestion
            "workspace_path": str(workspace_path),
```

samet-akcay: and same for the rest of the functions if possible

tests/unit/cli/builder/test_cli_builder.py
```diff
@@ -0,0 +1,420 @@
...
+class MockBackbone(nn.Module):
+    def __init__(self):
+        super(MockBackbone, self).__init__()
+
+    def forward(self, x):
+        return x
+
+
+class MockRegistry(Registry):
+    def __init__(self, name, parent=None, scope=None):
```
samet-akcay: Can you provide type hints for these args to improve readability please?

tests/unit/cli/builder/test_cli_builder.py
```diff
@@ -0,0 +1,420 @@
...
+        expected_extra_value = "!!!!!!!!!!!INPUT_HERE!!!!!!!!!!!"
+        assert "extra" in backbone_config
+        assert backbone_config["extra"] == expected_extra_value
+
+    @e2e_pytest_unit
+    def test_update_backbone_args_abnormal_backbone_type(self):
+        """Raise ValueError with unexpected backbone."""
+        backbone_config = {
+            "type": "unexpected",
+        }
```
samet-akcay: is the line length greater than our limit here?

tests/unit/cli/builder/test_cli_builder.py
```diff
@@ -0,0 +1,420 @@
+"""Unit-Test Case for otx.cli.builder.builder."""
```
samet-akcay, line 1: Just to follow the same directory structure, would it be an idea to place this test into `tests/unit/otx/cli/builder/test_cli_builder.py`?

samet-akcay: Or do you think all of the unit tests are already part of `otx` directory, hence we don't need to create `tests/unit/otx`?

tests/unit/cli/builder/test_cli_builder.py
```diff
@@ -0,0 +1,420 @@
...
+        backbone_config = mmcv.load(tmp_backbone_path)
+        assert backbone_config["backbone"]["type"] == backbone_type
+
+    @e2e_pytest_unit
+    @pytest.mark.parametrize("backbone_type", ["mmcls.MMOVBackbone"])
+    def test_builder_build_backbone_config_abnormal_output_path(self, backbone_type):
+        """Raise ValueError with wrong output_path."""
+        tmp_backbone_path = os.path.join(self.tmp_dir_path, "wrong.path")
+        with pytest.raises(ValueError):
+            self.otx_builder.build_backbone_config(backbone_type, tmp_backbone_path)
```
samet-akcay: If I'm missing anything, these lines are not covered by the test cases. This would be an important test case. Would it be possible to write a test case for this one as well?

samet-akcay (changes requested):
@harimkang, thanks a lot for your effort! I have few comments, majority of which could be organised into the following items:

- We could utilize `pathlib.Path` instead of `os.path` to significantly simplify the functions. I have added some suggestions already, but it would be great if you could apply the same thing to the rest of the functions.
- Can we add the type hints whenever possible please?
- Would it be possible to add a test case to cover Lines 157-159, which seems to be an important part?

tests/unit/cli/builder/test_cli_builder.py
eugene123tw: > What is your recommendation?

I think it'd be better if we call it `merge_backbone`, and keep the parameters as it is now. 

A remark while I was trying builder out:  
It's a great feature! That you can configure a model in this way. However, in terms of user experience, I feel model building is interactive and CLI is limited to the interactive process. I've been looking into [pytermgui](https://github.com/bczsalba/pytermgui) and [textual](https://github.com/Textualize/textual), they are terminal GUI. I hope we can bring this to another level i.e. building a model in terminal GUI. Maybe not now, but great if we add this feature.

tests/conftest.py
harimkang: reflected this.

tests/conftest.py
harimkang: This has also been reflected.

tests/unit/cli/builder/test_cli_builder.py
harimkang: I added a type hint with `Union[Path, str]`.

harimkang: I added this type hint.

tests/unit/cli/builder/test_cli_builder.py
harimkang: reflected this.

tests/unit/cli/builder/test_cli_builder.py
harimkang: The method of using `os.path` in test code and builder.py to be tested has been replaced with `pathlib.Path`. (`os.path` -> `pathlib.Path`)

tests/unit/cli/builder/test_cli_builder.py
harimkang: Added type hints.

tests/unit/cli/builder/test_cli_builder.py
harimkang: Modified to use `str` and `pathlib.Path` by replacing `os.path` in `builder.py` including `build_task_config` with `pathlib.Path`.

tests/unit/cli/builder/test_cli_builder.py
harimkang: This has also been corrected.

tests/unit/cli/builder/test_cli_builder.py
harimkang: Added type hints.

tests/unit/cli/builder/test_cli_builder.py
harimkang: It was modified by using `Path.joinpath`.

tests/unit/cli/builder/test_cli_builder.py
harimkang: Modified builder.py to receive `pathlib.Path`.

tests/unit/cli/builder/test_cli_builder.py
harimkang: Added type hints.

tests/unit/cli/builder/test_cli_builder.py
harimkang: No, I checked and fixed it!

tests/unit/cli/builder/test_cli_builder.py
harimkang: Your second comment is correct.
This doesn't matter, but since the target of tests is the `otx` folder (since other folders are not in the target), there seems to be no need to put `otx` in `tests/unit`.

tests/unit/cli/builder/test_cli_builder.py
harimkang: I'm sorry but I didn't understand this comment. Are you saying it would be nice to add some more test cases related to this? (wrong output path test)

harimkang:
> @harimkang, thanks a lot for your effort! I have few comments, majority of which could be organised into the following items:
> 
> * We could utilize `pathlib.Path` instead of `os.path` to significantly simplify the functions. I have added some suggestions already, but it would be great if you could apply the same thing to the rest of the functions.
> * Can we add the type hints whenever possible please?
> * Would it be possible to add a test case to cover Lines 157-159, which seems to be an important part?

Thanks for the detailed review comments. I have no doubt that your review will always improve the quality of the code!

1. As per your comment, I removed the use of `os.path` and replaced it with `pathlib.Path` in builder.py & test code. (This is actually the first time I've heard of it. I will have to use this in the future.)
2. Added type hints as much as possible. All hints of files modified in PR have been added.
3. This test code still doesn't cover lines 157-159. This is the code to help create a workspace folder in `otx build` run root. I think we can consider the case of creating and deleting a folder in a place **other than `tmp`**, but I'm not sure. Would it be better to write test code to add and clear a folder to root?

Thank you for the review, and please feel free to leave any additional comments. :)

tests/unit/cli/builder/test_cli_builder.py
harimkang: Your opinion looks good. I renamed `build_model_config` to `merge_backbone`.
I also think it is difficult to maximize and show our strengths with the current CLI.
It would be nicer if the combination of OTX workspace and several commands would work as a terminal GUI like you suggested. (Guess there's a lot more to think about!)

thanks for the comment :)

otx/cli/builder/builder.py
```diff
@@ -140,27 +142,28 @@ def build_task_config(
...
         """
 
         # Create OTX-workspace
         if workspace_path is None:
             workspace_path = f"./otx-workspace-{task_type}"
             if model_type:
                 workspace_path += f"-{model_type}"
-        os.makedirs(workspace_path, exist_ok=False)
+        workspace_path = workspace_path if isinstance(workspace_path, Path) else Path(workspace_path)
+        Path.mkdir(workspace_path, exist_ok=False)
```
samet-akcay: You could even do this, which is even easier:
```suggestion
        workspace_path.mkdir(exist_ok=False)
```

otx/cli/builder/builder.py
```diff
@@ -184,92 +187,102 @@ def build_task_config(
...
             train_type_rel_path = train_type
-        model_dir = os.path.join(os.path.abspath(template_dir), train_type_rel_path)
-        if not os.path.exists(model_dir):
+        model_dir = template_dir.absolute().joinpath(train_type_rel_path)
+        if not model_dir.exists():
             raise ValueError(f"[otx build] {train_type} is not a type supported by OTX {task_type}")
-        train_type_dir = os.path.join(workspace_path, train_type_rel_path)
-        os.makedirs(train_type_dir, exist_ok=True)
+        train_type_dir = workspace_path.joinpath(train_type_rel_path)
+        Path.mkdir(train_type_dir, exist_ok=True)
```
samet-akcay: [Optional] : Nice thing with pathlib is that it's super flexible, so even the following is possible 
```suggestion
        train_type_dir = workspace_path / train_type_rel_path
        train_type_dir.mkdir(exist_ok=True)
```

tests/unit/cli/builder/test_cli_builder.py
samet-akcay: Thanks for the clarification!

samet-akcay (commented):
Thanks @harimkang for swiftly addressing my comments!

As I mentioned, I try to promote `pathlib` as much as possible since it significantly reduces the complexity of the code.

Apologies for being annoying, I have one final comment. Instead of using `joinpath`, it is possible to simplify the code even further by doing the following:
```python
workspace_path.joinpath("configuration.yaml")    # GOOD
workspace_path / "configuration.yaml"            # BETTER, EVEN WORKS WITH WINDOWS!
```

samet-akcay:
> This test code still doesn't cover lines 157-159. This is the code to help create a workspace folder in otx build run root. I think we can consider the case of creating and deleting a folder in a place other than tmp, but I'm not sure. Would it be better to write test code to add and clear a folder to root?

As long as we do the clean up properly, I guess it could be ok. The last we would want is to pollute the root. 🤣 Another idea would be to create another tmp directory, maybe. Not sure how feasible it is though. 

You've already achieved 97% percent coverage, but I wanted to ensure how we handle if there is a wrong path, which would be an edge case. If you think this is not that important, we could ignore this for now

otx/cli/builder/builder.py
harimkang: I modified the code as suggested for a better quality!

otx/cli/builder/builder.py
harimkang: First time using it, but it looks good. thank you! Modified.

harimkang:
> > This test code still doesn't cover lines 157-159. This is the code to help create a workspace folder in otx build run root. I think we can consider the case of creating and deleting a folder in a place other than tmp, but I'm not sure. Would it be better to write test code to add and clear a folder to root?
> 
> As long as we do the clean up properly, I guess it could be ok. The last we would want is to pollute the root. 🤣 Another idea would be to create another tmp directory, maybe. Not sure how feasible it is though.
> 
> You've already achieved 97% percent coverage, but I wanted to ensure how we handle if there is a wrong path, which would be an edge case. If you think this is not that important, we could ignore this for now

I tried to temporarily change the location of the root in a way like `os.chdir`, but it doesn't work properly. If anyone knows how to make only certain functions of pytest work in a different root, please let me know. If I find out how, I'll add it. :)

samet-akcay:
> > > This test code still doesn't cover lines 157-159. This is the code to help create a workspace folder in otx build run root. I think we can consider the case of creating and deleting a folder in a place other than tmp, but I'm not sure. Would it be better to write test code to add and clear a folder to root?
> > 
> > 
> > As long as we do the clean up properly, I guess it could be ok. The last we would want is to pollute the root. 🤣 Another idea would be to create another tmp directory, maybe. Not sure how feasible it is though.
> > You've already achieved 97% percent coverage, but I wanted to ensure how we handle if there is a wrong path, which would be an edge case. If you think this is not that important, we could ignore this for now
> 
> I tried to temporarily change the location of the root in a way like `os.chdir`, but it doesn't work properly. If anyone knows how to make only certain functions of pytest work in a different root, please let me know. If I find out how, I'll add it. :)

Alright, let's not spend more time on this now, let's be pragmatic. Can you add a `TODO: Jira Ticket No: Cover these lines`. We could address it later. We have more important stuff :D

eugene123tw (dismissed):

harimkang:
> 
> Alright, let's not spend more time on this now, let's be pragmatic. Can you add a `TODO: Jira Ticket No: Cover these lines`. We could address it later. We have more important stuff :D

Created Jira ticket and added TODO comment! Thanks for the review!

samet-akcay (approved):
Looking good! Thanks a lot for your effort!

sungmanc:
> @harimkang, thanks a lot for your effort! I have few comments, majority of which could be organised into the following items:
> 
> * We could utilize `pathlib.Path` instead of `os.path` to significantly simplify the functions. I have added some suggestions already, but it would be great if you could apply the same thing to the rest of the functions.
> * Can we add the type hints whenever possible please?
> * Would it be possible to add a test case to cover Lines 157-159, which seems to be an important part?

I agreed with that `pathlib.Path` is more convenient than `os.path`. However, I'm not sure we should change `os.path` to `pathlib.Path`. Unifying the representation for `path` is needed. So, let's discuss it more in the near future.

harimkang:
> I agreed with that `pathlib.Path` is more convenient than `os.path`. However, I'm not sure we should change `os.path` to `pathlib.Path`. Unifying the representation for `path` is needed. So, let's discuss it more in the near future.

Maybe we should switch to this when we need to support WINDOWS. :D

harimkang:
After Datumaro Integration -> No Conflict
unit-test related builder
image
intg-test with build & find
image

@sungmanc, @eunwoosh, @jaegukhyun could you please approve this?

sungmanc (approved):

PR #1520, opened 2023-01-12
[OTX] consider multi GPU when setting drop last

eunwoosh:
### Summray

- setting drop_laste considering multi GPU

### Detail
Although samplers_per_gpu * num_gpus is bigger than number of images, drop_last is set as 0, then max_iter is calculated as 0, which makes a bug.

### DIFF
--- a/otx/algorithms/classification/tasks/inference.py
+++ b/otx/algorithms/classification/tasks/inference.py
@@ -8,6 +8,7 @@ import os
 from typing import Optional
 
 import numpy as np
+import torch.distributed as dist
 from mmcv.utils import ConfigDict
 
 from otx.algorithms.classification.configs import ClassificationConfig
@@ -454,7 +455,11 @@ class ClassificationInferenceTask(
             # In train dataset, when sample size is smaller than batch size
             if subset == "train" and self._data_cfg:
                 train_data_cfg = Stage.get_data_cfg(self._data_cfg, "train")
-                if len(train_data_cfg.get("otx_dataset", [])) < self._recipe_cfg.data.get("samples_per_gpu", 2):
+                num_gpus = dist.get_world_size() if self.distributed else 1
+                if (
+                    len(train_data_cfg.get("otx_dataset", []))
+                    < self._recipe_cfg.data.get("samples_per_gpu", 2) * num_gpus
+                ):
                     cfg.drop_last = False
 
             cfg.domain = domain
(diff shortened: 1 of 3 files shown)


otx/algorithms/classification/tasks/inference.py
```diff
@@ -454,7 +455,11 @@ def patch_color_conversion(pipeline):
             # In train dataset, when sample size is smaller than batch size
             if subset == "train" and self._data_cfg:
                 train_data_cfg = Stage.get_data_cfg(self._data_cfg, "train")
-                if len(train_data_cfg.get("otx_dataset", [])) < self._recipe_cfg.data.get("samples_per_gpu", 2):
+                num_gpus = dist.get_world_size() if dist.is_initialized() else 1
```
JihwanEom: How about `num_worlds` for cpu env?

otx/mpa/cls/trainer.py
```diff
@@ -105,7 +106,11 @@ def run(self, model_cfg, model_ckpt, data_cfg, **kwargs):
         drop_last = False
         dataset_len = len(otx_dataset) if otx_dataset else 0
         # if task == h-label & dataset size is bigger than batch size
-        if train_data_cfg.get("hierarchical_info", None) and dataset_len > cfg.data.get("samples_per_gpu", 2):
+        num_gpus = dist.get_world_size() if self.distributed else 1
```
JihwanEom: Same as above

JihwanEom (dismissed):
LGTM. Let's merge after passing TCs.

eunwoosh:
prmerge pass
All CI tests are passed.

otx/algorithms/classification/tasks/inference.py
eunwoosh: Nice point. I'll update it

otx/mpa/cls/trainer.py
eunwoosh: Nice point. I'll update it

eunwoosh:
As you can see all CI tests are passed and I just changed variable names from `num_gpus` to `num_worlds`.
Could you approve if you think it's okay?

JihwanEom (approved):

sungmanc (approved):
Don't we need to consider detection/segmentation tasks?

eunwoosh:
> Don't we need to consider detection/segmentation tasks?

Yes, I checked that there is code to set it as True, but only classification task has.

PR #1521, opened 2023-01-12
[OTX] Learning rate bugfix on ReduceLROnPlateau

JihwanEom:
## This PR includes:
- Fix learning rate reverting bug after dropping LR in ROP which was found in detection task.

### What was the root cause?
- ROP didn't consider **adaptive interval**. 
- If current epoch is not multiple of `self.interval` in ROP, `_should_check_stopping` returns `False`.
- Then, by `if` statement at line 505, ROP returns `base_lr` instead of reduced `self.current_lr`.
image

### How I fixed
- Define `self.currnt_lr` in initial epoch and return `self.current_lr` instead of `base_lr` when runner is not in check timing (every interval or epoch).

### Verify ROP
image
- By non-deterministic training, learning rate can differ but note that ROP works well on both tasks by this PR.

### DIFF
--- a/otx/algorithms/common/adapters/mmcv/hooks.py
+++ b/otx/algorithms/common/adapters/mmcv/hooks.py
@@ -491,13 +491,10 @@ class ReduceLROnPlateauLrUpdaterHook(LrUpdaterHook):
         self.key_indicator = key_indicator
         self.compare_func = self.rule_map[self.rule]
 
-    def _is_check_timing(self, runner):
-        """Called _is_check_timing in ReduceLROnPlateauLrUpdaterHook."""
+    def _is_check_timing(self, runner) -> bool:
+        """Check whether current epoch or iter is multiple of self.interval."""
         check_time = self.every_n_epochs if self.by_epoch else self.every_n_iters
-        if not check_time(runner, self.interval):
-            # No evaluation during the interval.
-            return False
-        return True
+        return check_time(runner, self.interval)
 
     @check_input_parameters_type()
     def get_lr(self, runner: BaseRunner, base_lr: float):
--- a/otx/algorithms/common/adapters/mmcv/hooks.py
+++ b/otx/algorithms/common/adapters/mmcv/hooks.py
@@ -491,7 +491,7 @@ class ReduceLROnPlateauLrUpdaterHook(LrUpdaterHook):
         self.key_indicator = key_indicator
         self.compare_func = self.rule_map[self.rule]
 
-    def _is_check_timing(self, runner) -> bool:
+    def _is_check_timing(self, runner: BaseRunner) -> bool:
         """Check whether current epoch or iter is multiple of self.interval."""
         check_time = self.every_n_epochs if self.by_epoch else self.every_n_iters
         return check_time(runner, self.interval)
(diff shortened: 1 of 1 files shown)


otx/algorithms/common/adapters/mmcv/hooks.py
```diff
@@ -491,8 +491,8 @@ def _init_rule(self, rule, key_indicator):
         self.key_indicator = key_indicator
         self.compare_func = self.rule_map[self.rule]
 
-    def _should_check_stopping(self, runner):
-        """Called _should_check_stopping in ReduceLROnPlateauLrUpdaterHook."""
+    def _is_check_timing(self, runner):
```
eunwoosh: I think this name is a little confusing. how about using `time_to_use` or maybe what you think better?

otx/algorithms/common/adapters/mmcv/hooks.py
```diff
@@ -491,8 +491,8 @@ def _init_rule(self, rule, key_indicator):
...
+        """Called _is_check_timing in ReduceLROnPlateauLrUpdaterHook."""
```
eunwoosh: This is meaningless. it's better to change it more meaningful.

otx/algorithms/common/adapters/mmcv/hooks.py
```diff
@@ -491,8 +491,8 @@ def _init_rule(self, rule, key_indicator):
...
         check_time = self.every_n_epochs if self.by_epoch else self.every_n_iters
         if not check_time(runner, self.interval):
```
eunwoosh: how about change this line to `return check_time(runner, self.interval)` and remove below lines?

jaegukhyun (dismissed):

otx/algorithms/common/adapters/mmcv/hooks.py
sungmanc: For the naming, I don't have any good idea. `time_to_use` also looks good to me. However, please put the return type.

```
def _is_check_timing(self, runner) -> bool:
```

sungmanc (dismissed):
Thanks for quick remedy

otx/algorithms/common/adapters/mmcv/hooks.py
harimkang: I also recommend adding type hints as well.

JihwanEom: Added type hint, but I love current name :) please go with as-is.

otx/algorithms/common/adapters/mmcv/hooks.py
JihwanEom: Reflected. Thank you.

otx/algorithms/common/adapters/mmcv/hooks.py
JihwanEom: Reflected. Thank you for good idea!

eunwoosh (approved):
LGTM. let's check the CI result.

harimkang (approved):

JihwanEom:
Close by #1533

PR #1523, opened 2023-01-13
[OTX] Decide best model using fixed metric

eunwoosh:
### Summary

- use a fixed metric when deciding best model weight

### Detail
Current classification task decides a best model weight by summing up all metric scores (top1, top5, each class score, etc.). This PR change it to use a single metric to decide.


JihwanEom (approved):
LGTM. Let's see test results.

sungmanc (approved):
Nice works

goodsong81 (approved):

PR #1527, opened 2023-01-16
[OTX][TEST] Disable multi gpu test temporary

eunwoosh:
### Summary

- Disable multi GPU test case until resolving a bug

### Detail
There is a bug that process is stuck during multi GPU test in CI machine. After resolving the bug, All those tests will be re-enabled again.

### DIFF
--- a/tests/integration/cli/classification/test_classification.py
+++ b/tests/integration/cli/classification/test_classification.py
@@ -246,7 +246,7 @@ class TestToolsMultiClassClassification:
         pot_eval_testing(template, tmp_dir_path, otx_dir, args)
 
     @e2e_pytest_component
-    @pytest.mark.skip  # disable multi GPU test until resolving a bug
+    @pytest.mark.skip  # TODO: Enable after fix this - https://jira.devtools.intel.com/browse/CVS-101246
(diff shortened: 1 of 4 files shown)


tests/integration/cli/classification/test_classification.py
```diff
@@ -246,6 +246,7 @@ def test_pot_eval(self, template, tmp_dir_path):
         pot_eval_testing(template, tmp_dir_path, otx_dir, args)
 
     @e2e_pytest_component
+    @pytest.mark.skip  # disable multi GPU test until resolving a bug
```
harimkang: How about create Jira-Ticket & Write TODO comments here with Ticket num?

eunwoosh: Good idea. I updated.

sungmanc (dismissed):

goodsong81 (changes requested):
There is something like skipe(reason="CVS-XXXX")
Please look for other test cases.

goodsong81 (approved):

yunchu (approved):

JihwanEom (approved):

PR #1448, opened 2022-12-16
[OTX] Drop legacy-resolver

cih9088:
### Changes
- Since pip version is not restricted to `21.2.1` anymore, we can drop `legacy-resolver`. This slow version resolve issue has been fixed in version of `22.0`.

[Release note of pip 22.0](https://pip.pypa.io/en/stable/news/#v22-0)
> Optimize installation order calculation to improve performance when installing requirements that form a complex dependency graph with a large amount of edges. ([#10557](https://github.com/pypa/pip/issues/10557))

PR #1528, opened 2023-01-16
[OTX][HOT-FIX] Keep pre-installed torch

cih9088:
New resolver introduced in https://github.com/openvinotoolkit/training_extensions/pull/1448 related issue.

If `file-links` are not specified explicitly, pip gets info from pypi only and reinstall torch, etc. with version from pypi.


sungmanc (approved):

goodsong81 (approved):
Actually, init_venv.sh is kind of temporary solution for developers.
It will be finally removed, and this issue will be resolved if we omit the pytorch from requirement item and treat it as prerequisite just like mmcv frameworks.
BTW, OK by then. :)

PR #1497, opened 2023-01-09
[OTX] Replace forked mm libraries with public ones

cih9088:
**I highly encourage you to look into this PR thoroughly as this PR impacts on all tasks in OTX.**

**Changes**
- replaced forked mm libraries with public ones
- employed `mmdeploy` to export model to OpenVINO IR
- enabled multi-node distributed training by extending https://github.com/openvinotoolkit/training_extensions/pull/1392
- used existing hooks (`AdaptiveTrainSchedulingHook`, `EvalHook`) to evaluate model before and after training instead of createing new one https://github.com/openvinotoolkit/training_extensions/pull/1472
- refactored (align code base across cls/det/seg to future refactoring)
- fixed bugs

**Tests**
- all pre-commit tests and pytests are passed

**Related PRs**
- mmdet ([OTX](https://github.com/openvinotoolkit/training_extensions/pull/1355))
- mmseg ([OTX](https://github.com/openvinotoolkit/training_extensions/pull/1400), [MPA](https://github.com/openvinotoolkit/model_preparation_algorithm/pull/96))
- mmcls ([OTX](https://github.com/openvinotoolkit/training_extensions/pull/1435), [MPA](https://github.com/openvinotoolkit/model_preparation_algorithm/pull/106))
- mmdeploy ([OTX](https://github.com/openvinotoolkit/training_extensions/pull/1466), [MPA](https://github.com/openvinotoolkit/model_preparation_algorithm/pull/124))

PR #1530, opened 2023-01-16
[OTX][HOT-FIX] Revert otx cli tools with data.yaml check

harimkang:
## Summary
Due to the effect of that https://github.com/openvinotoolkit/training_extensions/pull/1497, data.yaml of otx-workspace becomes unusable, so I revert this.

### DIFF
--- a/otx/cli/tools/eval.py
+++ b/otx/cli/tools/eval.py
@@ -56,12 +56,9 @@ def parse_args():
     parser = argparse.ArgumentParser()
     if not os.path.exists("./template.yaml"):
         parser.add_argument("template")
-    parser.add_argument("--data", required=False)
+    parser.add_argument("--data", required=False, default="./data.yaml")
     parsed, _ = parser.parse_known_args()
-    required = True
-    if parsed.data is not None:
-        assert os.path.exists(parsed.data)
-        required = False
+    required = not os.path.exists(parsed.data)
 
     parser.add_argument(
         "--test-ann-files",
--- a/otx/cli/tools/optimize.py
+++ b/otx/cli/tools/optimize.py
@@ -61,12 +61,9 @@ def parse_args():
     parser = argparse.ArgumentParser()
     if not os.path.exists("./template.yaml"):
         parser.add_argument("template")
-    parser.add_argument("--data", required=False)
+    parser.add_argument("--data", required=False, default="./data.yaml")
     parsed, _ = parser.parse_known_args()
-    required = True
-    if parsed.data is not None:
-        assert os.path.exists(parsed.data)
-        required = False
+    required = not os.path.exists(parsed.data)
 
     parser.add_argument(
         "--train-ann-files",
--- a/otx/cli/tools/train.py
+++ b/otx/cli/tools/train.py
@@ -69,12 +69,9 @@ def parse_args():
     parser = argparse.ArgumentParser()
     if not os.path.exists("./template.yaml"):
         parser.add_argument("template")
-    parser.add_argument("--data", required=False)
+    parser.add_argument("--data", required=False, default="./data.yaml")
     parsed, _ = parser.parse_known_args()
-    required = True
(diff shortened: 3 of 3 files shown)


cih9088:
The problem of original implementation was that `./data.yaml` used while testing sometimes does not get removed.
If `./data.yaml` is not removed, all the following tests would be failed. I didn't look into the root cause though.
Revert it for now and if a similar issue comes up, lets fix them then.

harimkang:
> The problem of original implementation was that `./data.yaml` used while testing sometimes does not get removed. If `./data.yaml` is not removed, all the following tests would be failed. I didn't look into the root cause though. Revert it for now and if a similar issue comes up, lets fix them then.

Oh, there was such an issue. Let me take a look.

sungmanc:
@harimkang @cih9088 , Did you execute the parallel tests ? (conduct >= 2 classification/detection/segmentation tests at the same time).

When I executed 2 task tests, I got the same error as @cih9088 's said.

harimkang:
> @harimkang @cih9088 , Did you execute the parallel tests ? (conduct >= 2 classification/detection/segmentation tests at the same time).
> 
> When I executed 2 task tests, I got the same error as @cih9088 's said.

Hi @sungchul2, I think the issue is caused by the function used in the test below. Could you please confirm?
https://github.com/openvinotoolkit/training_extensions/blob/3e5ff263aeb58384c372fd65da0c2c8928edcfbd/tests/integration/cli/classification/test_classification.py#L615

cih9088:
> When I executed 2 task tests, I got the same error as @cih9088 's said.

Yes, exactly. As I said, if you have `.data.yaml` in your CWD, because of whatever reasons, like pytest didn't remove it properly or one executed tests parallel, which means at some point you got `data.yaml` in your CWD, you would have an error.

harimkang:
> > When I executed 2 task tests, I got the same error as @cih9088 's said.
> 
> Yes, exactly. As I said, if you have `.data.yaml` in your CWD, because of whatever reasons, like pytest didn't remove it properly or one executed tests parallel, which means at some point you got `data.yaml` in your CWD, you would have an error.

As sungman said, if we run multiple tests at the same time, it seems to be reproduced. I think maybe the cause I found is correct. I guess this needs to be addressed as I don't think test should affect root.

sungmanc:
> > When I executed 2 task tests, I got the same error as @cih9088 's said.
> 
> Yes, exactly. As I said, if you have `.data.yaml` in your CWD, because of whatever reasons, like pytest didn't remove it properly or one executed tests parallel, which means at some point you got `data.yaml` in your CWD, you would have an error.

Let's assume this case,

test process 1: classification test
test process 2: detection test

Then, test process 1 made `data.yaml` during the `self-sl` test. In the mean time, test process 2 might get error because there was `data.yaml` file made by test process 1.

So, if we don't use parallel tests, then it will be fine

sungmanc:
> > > When I executed 2 task tests, I got the same error as @cih9088 's said.
> > 
> > 
> > Yes, exactly. As I said, if you have `.data.yaml` in your CWD, because of whatever reasons, like pytest didn't remove it properly or one executed tests parallel, which means at some point you got `data.yaml` in your CWD, you would have an error.
> 
> As sungman said, if we run multiple tests at the same time, it seems to be reproduced. I think maybe the cause I found is correct. I guess this needs to be addressed as I don't think test should affect root.

Agreed. and this problem will be resolved after the merge of Datumaro PR. @sungchul2 , @cih9088 , @harimkang .
So, how about keep this as is? 

Because, when we use the Datumaro, we don't need to directly make `data.yaml`  for test case.

Making `.yaml` file will be removed
image

sungchul1:
I agree. After Datumaro integration, I think we can remove the function.

sungmanc (approved):

eunwoosh (approved):

PR #1532, opened 2023-01-17
[OTX] Multi-GPU: run a thread to check parent process alive when child process starts

eunwoosh:
### Summary

- run a thread to check a parent process alive when child process starts

### DIFF
--- a/otx/cli/utils/multi_gpu.py
+++ b/otx/cli/utils/multi_gpu.py
@@ -24,6 +24,7 @@ import time
 from contextlib import closing
 from typing import Callable, List, Optional
 
+import psutil
 import torch
 import torch.distributed as dist
 import torch.multiprocessing as mp
@@ -237,7 +238,22 @@ class MultiGPUManager:
 
         MultiGPUManager.initialize_multigpu_train(rdzv_endpoint, rank, local_rank, gpu_ids, world_size)
 
+        threading.Thread(target=MultiGPUManager.check_parent_processes_alive, daemon=True).start()
+
         train_func()
+
+    @staticmethod
+    def check_parent_processes_alive():
+        """Check parent process is alive and if not, exit by itself."""
+        cur_process = psutil.Process()
+        parent = cur_process.parent()
+        while True:
+            time.sleep(1)
+            if not parent.is_running():
+                break
+
+        logger.warning("Parent processe is terminated abnormally. process exits.")
+        cur_process.kill()
 
     def _spawn_multi_gpu_processes(self, output_path: str) -> List[mp.Process]:
         processes = []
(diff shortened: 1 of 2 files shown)


eunwoosh:
I checked that all multi GPU tests are passed.
multi GPU test result

requirements/base.txt
```diff
@@ -19,3 +19,4 @@ scikit-learn==0.24.*
 Shapely>=1.7.1,<=1.8.0
 hpopt==0.3.0
 imagesize==1.4.1
+psutil
```
goodsong81, line 23: BSD 3-Clause "New" or "Revised" License.
https://github.com/giampaolo/psutil/blob/master/LICENSE

Looks OK but @yunchu could you check this?

otx/cli/utils/multi_gpu.py
```diff
@@ -237,8 +238,23 @@ def run_child_process(
...
+        threading.Thread(target=MultiGPUManager.check_parent_processes_alive, daemon=True).start()
+
         train_func()
 
+    @staticmethod
+    def check_parent_processes_alive():
+        """Check parent process is alive and if not, exit by itself."""
+        cur_process = psutil.Process()
+        parent = cur_process.parent()
+        while True:
```
jaegukhyun, line 250: I have a little concern about this. Is there no need to add a timeout?

jaegukhyun (commented):
I have a comment about "while True:"

otx/cli/utils/multi_gpu.py
eunwoosh: Yes. this function is always started as daemon. So, when parent process exits, then this process also exits automatically.

otx/cli/utils/multi_gpu.py
```diff
@@ -237,8 +238,23 @@ def run_child_process(
...
+            time.sleep(1)
+            if not parent.is_running():
+                break
+
+        logger.warning("Parent processe is terminated abnormally. process exits.")
```
JihwanEom: Simple typo.. `processe` -> `process`, and `process exits.` -> `Process exits.` (Capital)

eunwoosh: Thanks for pointing. I will update it.

JihwanEom (approved):

requirements/base.txt
goodsong81: @yunchu It's a reminder. Could you please confirm when you are available?

goodsong81 (approved):
LGTM. I would wait for @yunchu's comment.

requirements/base.txt
yunchu: @goodsong81 @eunwoosh 
There is no issue to use BSD 3-Clause "New" or "Revised" License. 
The only thing required is that adding this dependency to the BOM to update OSPDT for the next release.

PR #1533, opened 2023-01-17
[OTX][HOT-FIX] Learning rate bugfix on ReduceLROnPlateau

JihwanEom:
## This PR includes:
- Fix learning rate reverting bug after dropping LR in ROP which was found in detection task.

### What was the root cause?
- ROP didn't consider **adaptive interval**. 
- If current epoch is not multiple of `self.interval` in ROP, `_should_check_stopping` returns `False`.
- Then, by `if` statement at line 505, ROP returns `base_lr` instead of reduced `self.current_lr`.
image

### How I fixed
- Define `self.currnt_lr` in initial epoch and return `self.current_lr` instead of `base_lr` when runner is not in check timing (every interval or epoch).

### Verify ROP
image
- By non-deterministic training, learning rate can differ but note that ROP works well on both tasks by this PR.

### DIFF
--- a/otx/algorithms/common/adapters/mmcv/hooks.py
+++ b/otx/algorithms/common/adapters/mmcv/hooks.py
@@ -492,9 +492,9 @@ class ReduceLROnPlateauLrUpdaterHook(LrUpdaterHook):
         self.compare_func = self.rule_map[self.rule]
 
     def _is_check_timing(self, runner: BaseRunner) -> bool:
-        """Check whether current epoch or iter is multiple of self.interval."""
+        """Check whether current epoch or iter is multiple of self.interval, skip during warmup interations."""
         check_time = self.every_n_epochs if self.by_epoch else self.every_n_iters
-        return check_time(runner, self.interval)
+        return check_time(runner, self.interval) or (self.warmup_iters > runner.iter)
 
     @check_input_parameters_type()
     def get_lr(self, runner: BaseRunner, base_lr: float):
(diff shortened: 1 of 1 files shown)


supersoob (dismissed):

sungmanc (dismissed):

otx/algorithms/common/adapters/mmcv/hooks.py
```diff
@@ -491,25 +491,22 @@ def _init_rule(self, rule, key_indicator):
...
         """Called get_lr in ReduceLROnPlateauLrUpdaterHook."""
-        if not self._should_check_stopping(runner) or self.warmup_iters > runner.iter:
-            return base_lr
-
         if self.current_lr < 0:
             self.current_lr = base_lr
 
-        if hasattr(runner, self.metric):
-            score = getattr(runner, self.metric, 0.0)
+        if not self._is_check_timing(runner) or self.warmup_iters > runner.iter:
```
eunwoosh: how about move `self.warmup_iters > runner.iter` in `self._is_check_timing()`?

JihwanEom: Thank you for suggestion, but if we move `self.warmup_iters > runner.iter` in `_is_check_timing()`, we need if else statement in that method. 
I believe doing one job for one function is the way to improve readability. As plan B, we can split if not~ or statement with two if statements.  ex) `if not self.is_check_timing(runner) ~ if self.warmup_iters > runner.iter`

JihwanEom: Reflected, thank you for your wisdom :)

eunwoosh (approved):

jaegukhyun (approved):

PR #1534, opened 2023-01-17
[FEATURE] Integrate multi-gpu training for semi-supervised learning and self-supervised learning

supersoob:
### Summary

This PR is to enable multi-gpu training for semi-supervised learning and self-supervised learning.
- [x] Detection SEMI
- [x] Segmentation SEMI/SELF
- [x] Classification SEMI/SELF/SUPCON

**Main Issues**

- All tensors in losses has to have its local rank num because it is broadcasted or ring-scattered to other gpu. In det semisl, there were some tensors (ps_recall, ps_) without device num
- In self-sl classification, ckpt loading incurs issues that `
RuntimeError: storage has wrong size`. Because two models are needed to be built. After the first model(online_model) is loaded and saved to cache, the second model(target_model) will bring the ckpt from the cache because pretrained url is the same since they are same model. Such access to the cache can be done concurrently and make it corrupt during serialization between two processes.

### How to test

- Run e2e or integration test for each task

### Checklist

- [x] I submit my changes into the `develop` branch
- [x] I have updated the [documentation](https://github.com/openvinotoolkit/training_extensions/tree/develop/docs) accordingly

### License

```python
#
```


otx/algorithms/classification/adapters/mmcls/models/classifiers/byol.py
```diff
@@ -50,7 +52,11 @@ def __init__(
 
         # build backbone
         self.online_backbone = build_backbone(backbone)
-        self.target_backbone = build_backbone(backbone)
+
+        target_backbone_cfg = copy.deepcopy(backbone)
+        target_backbone_cfg["pretrained"] = None
```
jaegukhyun, line 57: Could you explain why the target backbone's pretrain should be False?

jaegukhyun (commented):
Generally LGTM, I left a comment for question

eunwoosh:
Could you add Test Case for this with skip flag? I'll re-enable multi GPU TCs after checking.

otx/algorithms/classification/adapters/mmcls/models/classifiers/byol.py
supersoob: I detected mmcv bug that ckpt corruption when consecutive ckpt loading happens. And actually target backbone does not need to load ckpt pretrained weight. So made it not incur the corresponding error.

supersoob:
@jaegukhyun @eunwoosh could you check again?

jaegukhyun (approved):

sungmanc (approved):

PR #1541, opened 2023-01-18
[OTX][Hot-Fix] Fix self-sl seg issue

sungmanc:
# Summary
For self-supervised learning, there is no argument `--val-data-roots`.
In this case, `seg/trainer.py` need to check the whether `val` data is included in data configuration to avoid making `val_dataloader`

Here is the local test result for Self-SL segmentation
image

### DIFF
--- a/otx/mpa/det/trainer.py
+++ b/otx/mpa/det/trainer.py
@@ -59,12 +59,6 @@ class DetectionTrainer(DetectionStage):
         if 'val' in cfg.data:
             cfg.data.val_dataloader.samples_per_gpu = 1
 
-        # FIXME: scale_factors is fixed at 1 even batch_size > 1 in simple_test_mask
-        # Need to investigate, possibly due to OpenVINO
-        if "roi_head" in model_cfg.model:
-            if "mask_head" in model_cfg.model.roi_head:
-                cfg.data.val_dataloader.samples_per_gpu = 1
-
         if hasattr(cfg, "hparams"):
             if cfg.hparams.get("adaptive_anchor", False):
                 num_ratios = cfg.hparams.get("num_anchor_ratios", 5)
--- a/otx/mpa/det/trainer.py
+++ b/otx/mpa/det/trainer.py
@@ -56,7 +56,8 @@ class DetectionTrainer(DetectionStage):
         datasets = [build_dataset(cfg.data.train)]
 
         # FIXME: Currently detection do not support multi batch evaluation. This will be fixed
-        cfg.data.val_dataloader.samples_per_gpu = 1
+        if 'val' in cfg.data:
+            cfg.data.val_dataloader.samples_per_gpu = 1
 
         # FIXME: scale_factors is fixed at 1 even batch_size > 1 in simple_test_mask
         # Need to investigate, possibly due to OpenVINO
--- a/otx/mpa/seg/trainer.py
+++ b/otx/mpa/seg/trainer.py
@@ -57,7 +57,9 @@ class SegTrainer(SegStage):
         datasets = [build_dataset(cfg.data.train)]
 
         # FIXME: Currently segmentor does not support multi batch evaluation.
-        cfg.data.val_dataloader.samples_per_gpu = 1
+        # For the Self-SL case, there is no val data. So, need to check the 
+        if 'val' in cfg.data:
+            cfg.data.val_dataloader.samples_per_gpu = 1
 
         # Dataset for HPO
         hp_config = kwargs.get("hp_config", None)


cih9088 (approved):

cih9088 (dismissed):
Could you apply the change in det/trainer.py as well for completeness?

goodsong81 (dismissed):

sungmanc:
> Could you apply the change in seg/trainer.py as well for completeness?

No, problems, I'll

sungmanc:
@goodsong81 @cih9088 , I applied this solution to `det/train.py` to consider further self-sl detection.
Please recheck this :)

otx/mpa/det/trainer.py
```diff
@@ -56,7 +56,8 @@ def run(self, model_cfg, model_ckpt, data_cfg, **kwargs):
         datasets = [build_dataset(cfg.data.train)]
 
         # FIXME: Currently detection do not support multi batch evaluation. This will be fixed
-        cfg.data.val_dataloader.samples_per_gpu = 1
+        if 'val' in cfg.data:
+            cfg.data.val_dataloader.samples_per_gpu = 1
 
         # FIXME: scale_factors is fixed at 1 even batch_size > 1 in simple_test_mask
         # Need to investigate, possibly due to OpenVINO
```
cih9088: You missed this block (line 62-66)! Please include it as well. :)

sungmanc: I'm going to remove line 62-66 because if there is a validation dataset, `val_dataloader.samplers_per_gpu` will be set to 1.

sungmanc: What do you think?

cih9088: Yes, then no objection on that. I also left this part for the reference because there is a `FIXME` I didn't consider at all.
If this `FIXME` is considered `FIXED`, then I believe we can safely get rid of it.

cih9088 (dismissed):

otx/mpa/det/trainer.py
sungmanc: Actually, in the previous version, there is no problem to use `val_dataloader.samplers_per_gpu` > 1 for detection, classification except for instance segmentation, semantic segmentation. However, I'm also not sure the latest mmdetection version. So, let's keep as is with `FIXME` tag.

otx/mpa/seg/trainer.py
```diff
@@ -57,7 +57,9 @@ def run(self, model_cfg, model_ckpt, data_cfg, **kwargs):
         datasets = [build_dataset(cfg.data.train)]
 
         # FIXME: Currently segmentor does not support multi batch evaluation.
-        cfg.data.val_dataloader.samples_per_gpu = 1
+        # For the Self-SL case, there is no val data. So, need to check the 
+        if 'val' in cfg.data:
+            cfg.data.val_dataloader.samples_per_gpu = 1
```
sungchul1, line 62: Could you explain why this change is needed for seg? I don't know the reason because mmseg already uses `samples_per_gpu=1` when creating val_dataloader. 

https://github.com/openvinotoolkit/mmsegmentation/blob/ac8e04d73857af3bff4c27df62ac69d1c51ff3ac/mmseg/apis/train.py#L65-L74

cih9088: **This is a new feature from public mm libraries to construct different dataloader per subset.**

Since global options for dataloader, such as `samples_per_gpu`, etc., would be deprecated in the future as `mmdet` is on the way, MPA explicitly [construct dataloader config](https://github.com/openvinotoolkit/training_extensions/blob/8b1732239eeedde534f1f36e7d89d8951c4d5bb3/otx/mpa/seg/trainer.py#L104) for each subset right before the training. [The method](https://github.com/openvinotoolkit/training_extensions/blob/8b1732239eeedde534f1f36e7d89d8951c4d5bb3/otx/mpa/stage.py#L309-L346) just pops the global options and assigns them to each dataloader config.

Specifying dataloader config would [override the default](https://github.com/open-mmlab/mmsegmentation/blob/master/mmseg/apis/train.py#L157-L162) which is 1 for validations's `samples_per_gpu`, MPA should set to `1` explicitly.

sungmanc: My changing part is to avoid the `cfg.data.val_dataloader.samples_per_gpu=1`.
And after checking the mmseg, I also think we can remove the line #60-62 for the segmentation.

sungmanc: > Specifying dataloader config would [override the default](https://github.com/open-mmlab/mmsegmentation/blob/master/mmseg/apis/train.py#L157-L162) which is 1 for validations's samples_per_gpu, MPA should set to 1 explicitly.

Then you mean that we need to explicitly let users know by adding this line although the default value is set to 1?

cih9088: @sungmanc This should be as is. Unless, it will build dataloader with `samples_per_gpu` from global dataloader option which might not be `1`.

sungmanc: Understood, thanks for kind explanation. what do you think? @sungchul2 

If you also agree, we can merge this PR

sungchul1: I agree. Thanks for your kind summary!

sungchul1 (dismissed):

cih9088 (dismissed):

sungmanc:
@cih9088 @sungchul2 @goodsong81 , after rebase, need to re-approve.  thanks in advance.

cih9088 (approved):

sungchul1 (approved):

PR #1542, opened 2023-01-18
[OTX] remove temporary work directory after training

eunwoosh:
### Summary

- This PR makes temporary work directory removed after training
- Also fix a bug that CI is stuck while HPO tests is running after error occurs

### Detail
Current OTX deletes a work directory when garbage collector decides to delete a train task. So, it's uncertain when and whether the directory is deleted. After merging this PR, work directory always deleted at the last part of train.py of CLI if `work-dir` is not specified by user. If user specifies `work-dir`, then OTX assumes that user needs `work-dir` and not delete it.

### Reason of the bug that CI is stuck while HPO test case is running
While a trial of HPO is running, garbage collector deletes train task, which result in executing `__del__()`  in the class.
Then, work directory is deleted, and logger hook raises error that logger text file is missing.
I don't know exact reason why garbage collector deletes the task, but I checked that this PR fixes this bug.

### DIFF
--- a/otx/algorithms/action/tasks/inference.py
+++ b/otx/algorithms/action/tasks/inference.py
@@ -304,7 +304,8 @@ class ActionInferenceTask(BaseTask, IInferenceTask, IExportTask, IEvaluationTask
 
     def unload(self):
         """Unload the task."""
-        self.finalize()
+        if self._work_dir_is_temp:
+            self._delete_scratch_space()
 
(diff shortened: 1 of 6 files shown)


sungmanc (approved):
It must be nice solution, LGTM

goodsong81 (approved):
Nice!

PR #1516, opened 2023-01-11
[OTE][dev] Fix tiling OpenVINO backward compatibility

eugene123tw:

PR #1511, opened 2023-01-10
[OTE][dev] Copy Feature Vector

eugene123tw:

PR #1543, opened 2023-01-18
[OTX] Tiling feature in OTX

eugene123tw:
* Detection Test
Screenshot from 2023-01-23 16-17-58

* Inst Seg Test
Screenshot from 2023-01-23 08-31-28

* Tiling Detection Test
Screenshot from 2023-01-23 10-19-18

* Tiling Inst-Seg Test
Screenshot from 2023-01-23 12-09-55


goodsong81 (dismissed):
Thank you for the update. LGTM.
@jaegukhyun Could you check the changes in common part?

otx/algorithms/common/tasks/training_base.py
```diff
@@ -473,6 +475,13 @@ def _load_model_ckpt(self, model: Optional[ModelEntity]):
             if model_data.get("anchors"):
                 self._anchors = model_data["anchors"]
 
+            tiling_parameters = model_data.get("config", None).get("tiling_parameters", None)
```
jaegukhyun: What happens if model_data do not contain "config" attribute? Can we get something from None?

otx/algorithms/detection/adapters/mmdet/utils/builder.py
```diff
@@ -28,8 +28,6 @@ def build_detector(
 
     from mmdet.models import build_detector as origin_build_detector
 
-    model_cfg = deepcopy(config.model)
```
jaegukhyun: Could you explain why this part is removed?

otx/algorithms/detection/adapters/mmdet/utils/config_utils.py
```diff
@@ -241,6 +241,10 @@ def update_pipeline(cfg):
             continue
         config.data[f"{subset}_dataloader"] = config.data.get(f"{subset}_dataloader", ConfigDict())
 
+        remove_from_config(config.data[subset], "ann_file")
+        remove_from_config(config.data[subset], "img_prefix")
+        remove_from_config(config.data[subset], "classes")  # Get from DatasetEntity
```
jaegukhyun, line 246: Why normal detection datasets works without these config modification?

otx/algorithms/detection/utils/data.py
```diff
@@ -464,24 +464,22 @@ def format_list_to_str(value_lists: list):
     return f"[{str_value[:-2]}]"
 
 
-def adaptive_tile_params(dataset: DatasetEntity, object_tile_ratio=0.01, rule="avg") -> Dict:
+def adaptive_tile_params(hyper_parameters, dataset: DatasetEntity, object_tile_ratio=0.01, rule="avg"):
```
jaegukhyun: Could you add type_hint of hyper_parameters and add it to the description too?

otx/mpa/stage.py
```diff
@@ -293,12 +293,15 @@ def configure_samples_per_gpu(
...
         distributed: bool = False,
     ):
-        task_lib_module = importlib.import_module(f"{MODEL_TASK[cfg.model_task]}.datasets")
-        dataset_builder = getattr(task_lib_module, "build_dataset")
 
         dataloader_cfg = cfg.data.get(f"{subset}_dataloader", ConfigDict())
         samples_per_gpu = dataloader_cfg.get("samples_per_gpu", cfg.data.get("samples_per_gpu", 1))
-        dataset_len = len(build_dataset(cfg, subset, dataset_builder))
+
+        if getattr(cfg.data[subset], "dataset", None):
```
jaegukhyun: I think hasattr is more appropriate

otx/api/entities/tensor.py
```diff
@@ -21,7 +21,7 @@ class TensorEntity(IMetadata):
 
     def __init__(self, name: str, numpy: np.ndarray):
         self.name = name
-        self._numpy = numpy
+        self._numpy = np.copy(numpy)
```
jaegukhyun, line 25: Could you explain why we need copy of numpy?

otx/cli/utils/io.py
```diff
@@ -103,6 +103,8 @@ def read_model(model_configuration: ModelConfiguration, path: str, train_dataset
             full_path = os.path.join(os.path.dirname(path), key)
             if os.path.exists(full_path):
                 model_adapters[key] = ModelAdapter(read_binary(full_path))
+            else:
+                model_adapters[key] = ModelAdapter(bytes())
```
jaegukhyun, line 107: Could you explain why we need this for tiling?

jaegukhyun (commented):
Generally looks good to me too. I have a few questions about the code. 
Also, it would be great if you add local test results of test_detection, test_instance_segmentation, test_tiling. Since current ci do not work properly.

otx/algorithms/detection/adapters/mmdet/utils/builder.py
eugene123tw: It's duplicated: https://github.com/openvinotoolkit/training_extensions/blob/01f2bd9c1b7c87a59b0653391f9fc8b5036e7383/otx/algorithms/detection/adapters/mmdet/utils/builder.py#L34

otx/algorithms/detection/adapters/mmdet/utils/config_utils.py
eugene123tw: Tiling uses a different data pipeline (`tile_pipeline.py`). And `ImageTilingDataset` doesn't take `ann_file`, , `img_prefix`, and `classes` as arguments.

otx/algorithms/detection/utils/data.py
eugene123tw: Good Point. Added!

otx/algorithms/detection/adapters/mmdet/utils/builder.py
eugene123tw: I will revert this change to avoid conflict as Inhyuk also found this issue in his PR

https://github.com/openvinotoolkit/training_extensions/pull/1553/files#diff-d9bdf7888477389f59524b4f06c95e46d030d1fd25191158dc6a9b6dd1188f6bL31

otx/api/entities/tensor.py
eugene123tw: It's reflecting the change in develop branch to feature/otx: 
https://github.com/openvinotoolkit/training_extensions/pull/1511

I didn't create a new PR as @goodsong81 suggested I add here

otx/cli/utils/io.py
eugene123tw: It's the same reason as previous remark: https://github.com/openvinotoolkit/training_extensions/pull/1516

otx/mpa/stage.py
eugene123tw: Thanks! I also think that's better.

eugene123tw (commented):
@jaegukhyun Thanks for reviewing. I have added my comments and local test results.

otx/algorithms/common/tasks/training_base.py
eugene123tw: Good point! I have made some changes.

otx/algorithms/detection/adapters/mmdet/utils/config_utils.py
JihwanEom: Can we remove these lines by adding `**kwargs` in `ImageTilingDataset`?

otx/mpa/modules/datasets/task_adapt_dataset.py
```diff
@@ -10,8 +10,16 @@
...
-    """Dataset wrapper for task-adative evaluation."""
+class TaskAdaptEvalDataset:
+    """Dataset wrapper for task-adaptive evaluation.
+
+    Attributes:
+        model_classes:
+        **kwargs:
+
+    Notes:
+        TODO[Eugene/Jaeguk]: check if this class is still valid in OTX
```
JihwanEom, line 21: As far as I know, this class is not used on OTX. (it was designed for previous incremental learning in MPA) BTW, could you tell me why you updated this file? Is this required for tiling?

otx/mpa/stage.py
```diff
@@ -293,12 +293,15 @@ def configure_samples_per_gpu(
...
         distributed: bool = False,
     ):
-        task_lib_module = importlib.import_module(f"{MODEL_TASK[cfg.model_task]}.datasets")
-        dataset_builder = getattr(task_lib_module, "build_dataset")
 
         dataloader_cfg = cfg.data.get(f"{subset}_dataloader", ConfigDict())
         samples_per_gpu = dataloader_cfg.get("samples_per_gpu", cfg.data.get("samples_per_gpu", 1))
-        dataset_len = len(build_dataset(cfg, subset, dataset_builder))
+
+        if hasattr(cfg.data[subset], "dataset"):
```
JihwanEom: @jaegukhyun Can we use `unwrap_dataset`? I can't remember exactly :( help me..

otx/mpa/modules/datasets/task_adapt_dataset.py
eugene123tw: It's not required for tiling, however, it's still being used.
 https://github.com/openvinotoolkit/training_extensions/blob/67105dcdd302699f2d39ebca6831561ba6f16692/otx/mpa/det/stage.py#L184

Since it's going to be used anyway according to Harim and Jaeguk. It's better including it in data pipeline. Therefore there will be a PR doing this as noted with TODO.

otx/algorithms/detection/adapters/mmdet/utils/config_utils.py
eugene123tw: We can. But I don't prefer passing unnecessary kwargs to objet classes. I think the better way to solve this is to remove ann_file, img_prefix and classes from data pipeline (it's better to solve it from the root than patching).

otx/mpa/modules/datasets/task_adapt_dataset.py
harimkang: @JihwanEom As Eugene said, it is still used in detection.

otx/mpa/stage.py
jaegukhyun: I'm not familiar with unwrap_dataset. @eugene123tw How about using get_data_cfg from otx.mpa.stage?

jaegukhyun (commented):
I left one comment about getting length of dataset

jaegukhyun (approved):
LGTM! Thank you for your effort

JihwanEom (approved):

goodsong81 (approved):

otx/mpa/deploy/apis.py
```diff
@@ -195,6 +195,11 @@ def helper(*args, **kwargs):
                 "input_data",
                 {"shape": (128, 128, 3), "file_path": None},
             )
+
+            # patch test_pipeline in case of missing LoadImageFromOTXDataset
+            if cfg.data.test.pipeline[0].type != "LoadImageFromOTXDataset":
+                cfg.data.test.pipeline.insert(0, mmcv.ConfigDict(dict(type="LoadImageFromOTXDataset")))
```
harimkang, line 201: Hi @eugene123tw,
What is the background for this code being added? Due to this, the data pipeline is built and throws an error even during the current `otx export`. (current export, this code works and LoadImageFromOTXDataset is called and an error is raised there.)
Is this what we need for tiling?

PR #1544, opened 2023-01-19
[OTX][Hotfix] Remove base_path in semisl hparam.yaml

JihwanEom:
To support otx build -> otx train in semi-sl, base_path should not be included in hparam.yaml.

### DIFF
--- a/otx/algorithms/classification/configs/efficientnet_b0_cls_incr/semisl/hparam.yaml
+++ b/otx/algorithms/classification/configs/efficientnet_b0_cls_incr/semisl/hparam.yaml
@@ -1,6 +1,5 @@
 # Hyperparameters.
 hyper_parameters:
-  base_path: ../../configuration.yaml
   parameter_overrides:
     learning_parameters:
       batch_size:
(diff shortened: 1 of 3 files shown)


sungmanc:
Could you explain with a more detailed description? I'm not sure why `base_path` should be excluded

JihwanEom:
> Could you explain with a more detailed description? I'm not sure why `base_path` should be excluded

When we run `otx train --task classification --train-type semisl`, configuration.yaml is created in the same directory with template.yaml. 
So base path should be `./configuration.yaml` but if we have `base_path` in semisl/hparams.yaml, `otx build` overwrites `base_path` in template.yaml, otx can't find configuration.yaml like below image. (it tries to find `../../configuration.yaml`)

image

sungmanc:
> > Could you explain with a more detailed description? I'm not sure why `base_path` should be excluded
> 
> When we run `otx train --task classification --train-type semisl`, configuration.yaml is created in the same directory with template.yaml. So base path should be `./configuration.yaml` but if we have `base_path` in semisl/hparams.yaml, `otx build` overwrites `base_path` in template.yaml, otx can't find configuration.yaml like below image. (it tries to find `../../configuration.yaml`)
> 
> image

Thanks a lot for kind explanation. Thanks

eunwoosh (approved):
LGTM

sungmanc (approved):

PR #1550, opened 2023-01-19
[OTX] copy fixes in classification from develop

sovrasov:
- Active score
- Indexing in the hierarchical part

### DIFF
--- a/otx/algorithms/classification/adapters/openvino/model_wrappers/openvino_models.py
+++ b/otx/algorithms/classification/adapters/openvino/model_wrappers/openvino_models.py
@@ -146,8 +146,8 @@ def activate_multihead_output(logits: np.ndarray, multihead_class_info: dict):
         logits[logits_begin:logits_end] = softmax_numpy(logits[logits_begin:logits_end])
 
     if multihead_class_info["num_multilabel_classes"]:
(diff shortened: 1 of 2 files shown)


goodsong81 (dismissed):
Thank you for the fix!

jaegukhyun (commented):
Your OTE PR contains a change in external/deep-object-reid/torchreid_tasks/utils.py. Don't this change need to patch here?

sovrasov:
> Your OTE PR contains a change in external/deep-object-reid/torchreid_tasks/utils.py. Don't this change need to patch here?

As far as I know, Sungman has deleted the whole external directory from otx, so we don't need patching d-o-r and other similar folders anymore.

sungmanc:
> > Your OTE PR contains a change in external/deep-object-reid/torchreid_tasks/utils.py. Don't this change need to patch here?
> 
> As far as I know, Sungman has deleted the whole external directory from otx, so we don't need patching d-o-r and other similar folders anymore.

Yes, now we don't use any external modules for OTX :)

sovrasov:
pytest -s -v tests/integration/cli:
image

JihwanEom (approved):

jaegukhyun (approved):

PR #1554, opened 2023-01-20
[OTX][Hot-fix] Add init_weights call in the custom atss head

jaegukhyun:
After changing mmcv framework, init_weights function call in atss head is disappeared.
That raise large classification loss at the early stage. 
This fix add init_weights call in the init_layer function in atss_head. 

Before
image

After
image

### DIFF
--- a/otx/mpa/modules/models/heads/custom_atss_head.py
+++ b/otx/mpa/modules/models/heads/custom_atss_head.py
@@ -36,15 +36,6 @@ class CustomATSSHead(CrossDatasetDetectorHead, ATSSHead):
         self.bg_loss_weight = bg_loss_weight
         self.use_qfl = use_qfl
 
-    def _init_layers(self):
-        """Initialize layers of the ATSS head.
-
-        Current mmdetection requires to call init_weights function to reflect init_cfg
-        """
-
-        super(CustomATSSHead, self)._init_layers()
-        self.init_weights()
-
     @force_fp32(apply_to=("cls_scores", "bbox_preds", "centernesses"))
     def loss(self, cls_scores, bbox_preds, centernesses, gt_bboxes, gt_labels, img_metas, gt_bboxes_ignore=None):
         """Compute losses of the head.
--- a/otx/mpa/modules/models/heads/custom_atss_head.py
+++ b/otx/mpa/modules/models/heads/custom_atss_head.py
@@ -36,6 +36,15 @@ class CustomATSSHead(CrossDatasetDetectorHead, ATSSHead):
         self.bg_loss_weight = bg_loss_weight
         self.use_qfl = use_qfl
 
+    def _init_layers(self):
+        """Initialize layers of the ATSS head.
+
+        Current mmdetection requires to call init_weights function to reflect init_cfg
+        """
+
+        super(CustomATSSHead, self)._init_layers()
+        self.init_weights()
+
     @force_fp32(apply_to=("cls_scores", "bbox_preds", "centernesses"))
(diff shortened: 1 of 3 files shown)


otx/mpa/modules/models/heads/custom_atss_head.py
```diff
@@ -36,6 +36,15 @@ def __init__(
         self.bg_loss_weight = bg_loss_weight
         self.use_qfl = use_qfl
 
+    def _init_layers(self):
```
harimkang: Is this a function called from the mmdetection side?

harimkang (dismissed):

otx/mpa/modules/models/heads/custom_atss_head.py
```diff
@@ -36,6 +36,15 @@ def __init__(
...
+        """Initialize layers of the ATSS head.
+
+        Current mmdetection requires to call init_weights function to reflect init_cfg
+        """
+
+        super(CustomATSSHead, self)._init_layers()
+        self.init_weights()
+
```
goodsong81: Is this a right way to call init_weghts()? (Sorry, I don't know, just question)
Is it different among other models & heads... or mmcv modules?

otx/mpa/modules/models/heads/custom_atss_head.py
jaegukhyun: Yes it is in the latest mmdet's atss_head class

otx/mpa/modules/models/heads/custom_atss_head.py
jaegukhyun: You mean, we should check other structure and give general solution? I made this PR only for ATSS since, SSD and YOLOX does not give this kind of issue after mmcv framework change. I'll check whether other structure call init_weights. If they not call this func, then I'll find why those two structures work fine

cih9088:
How about calling `init_weights()` in a model level when [building a model](https://github.com/openvinotoolkit/training_extensions/blob/feature/otx/otx/algorithms/detection/adapters/mmdet/utils/builder.py) for all tasks? It would be beneficial in terms of training as each mm libraries defined fine-grained `init_cfg` per modules to boost training. I skimmed through forked libraries and found out that they called `init_weights()` inside of `__init__()` which means calling `init_weights()` was implicit at that time but now users should call it explicitly. I couldn't find why they changed the design in this way.

jaegukhyun:
@cih9088 Thank you for your opinion. I guess that is what @goodsong81 mentioned. Let me modify this after validation on some custom datasets

cih9088:
[This mmcv PR](https://github.com/open-mmlab/mmcv/pull/780) was about forcing users to call `init_weights()` explicitly.
I believe we should call it in model build function.

> Please inform users to call model.init_weight() after building models in tutorals.

goodsong81 (approved):
Looks much better to me, thanks.
BTW, any problem for log level?

@JihwanEom @supersoob Could you check the `init_weight()` call in classification & segmentation?

harimkang (approved):

jaegukhyun:
> Looks much better to me, thanks. BTW, any problem for log level?
> 
> @JihwanEom @supersoob Could you check the `init_weight()` call in classification & segmentation?

init_weights function print logs about whole initialized parameters, and it looks very annoying

JihwanEom:
> Looks much better to me, thanks. BTW, any problem for log level?
> 
> @JihwanEom @supersoob Could you check the `init_weight()` call in classification & segmentation?

`init_weight()` is not called in classification task, so classifier head is not initialized with given method, initialized only with random. 
I'll create PR to resolve this issue and conduct to benchmark performance on both incremental/semi-sl.

PR #1555, opened 2023-01-20
[DEV] Remove filelock in videoframe caching

JihwanEom:
This PR removes filelock used in videoframe caching by using temporary filename.

When the process starts to save image to cache, image will be saved as `TRAINING-000001-tmp.png` and will be renamed as `TRAINING-000001.png` after saving.
`load_image_from_cache` only tries to find renamed `TRAINING-000001.png`, so cannot load saving (temporary) image.

### DIFF
--- a/external/model-preparation-algorithm/mpa_tasks/utils/data_utils.py
+++ b/external/model-preparation-algorithm/mpa_tasks/utils/data_utils.py
@@ -42,9 +42,6 @@ def get_image(results: Dict[str, Any], cache_dir: str, to_float32=False):
     def is_training_subset(subset):
         return subset.name in ["TRAINING", "VALIDATION"]
 
-    def is_cached_or_caching(filename, tmp_filename):
-        return os.path.exists(filename) or os.path.exists(tmp_filename)
-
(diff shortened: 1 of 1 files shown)


goodsong81 (approved):
Nice one!
BTW, can we encapsulate "-tmp" logic in `is_cached()` and `cache_image()`?

JihwanEom:
> Nice one! BTW, can we encapsulate "-tmp" logic in `is_cached()` and `cache_image()`?

Reflected, thank you for suggesstion.

jaegukhyun (approved):

sungmanc (commented):
Is there no problem to use `jpg` for loading? AFAIK, `jpg` format showed the lower performance than `png` although more faster

goodsong81 (approved):

JihwanEom:
> Is there no problem to use `jpg` for loading? AFAIK, `jpg` format showed the lower performance than `png` although more faster

Sorry for confusion by typos in description. Proposed code lines does not use `jpg` format, only read/write `png` format.

sungmanc (approved):
What a good feature !

PR #1556, opened 2023-01-20
[OTX] Fix a timing to compare score and add bad_count in ReduceLROnPlateauLrUpdaterHook

eunwoosh:
### Summary

- modify a timing to compare score after evaluation
- modify an order of print_log to infrom hook status after updating `patience` and `best score`

### Detail
**This PR is for fixing below two bugs.** 
**Bug1 : increase `bad_count` during warmup iteration**
This is due to `_is_check_timing()`. it's fixed to return False during warmup iterations.

**Bug2: compare score and increase `bad_count` right before evaluating model**
This is because `ReduceLROnPlateauLrUpdaterHook` is executed at `before_train_epoch` but eval hook is executed at `after_train_epoch`. To remedy it, the code is fixed to compare score and add `bad_count` at next epoch (or iter) after multiples of interval epoch(iter).

### DIFF
--- a/otx/algorithms/common/adapters/mmcv/hooks.py
+++ b/otx/algorithms/common/adapters/mmcv/hooks.py
@@ -493,8 +493,16 @@ class ReduceLROnPlateauLrUpdaterHook(LrUpdaterHook):
 
     def _is_check_timing(self, runner: BaseRunner) -> bool:
         """Check whether current epoch or iter is multiple of self.interval, skip during warmup interations."""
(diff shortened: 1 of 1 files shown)


jaegukhyun (dismissed):
LGTM. Thank you for your effort

supersoob (dismissed):

JihwanEom (dismissed):
LGTM. Thanks for your effort!

JihwanEom (approved):

jaegukhyun (approved):

PR #1557, opened 2023-01-20
[Hot-fix][OTX] Exclude background label in OTX dataset in segmentation

supersoob:
This PR is to resolve the training performance degradation in segmentation due to inclusion of background label and annotation when converting to OTXdataset.
After excluding the background label in OTX dataset annotation, the training was working properly and val mdice and accuracy is increased as before.
Still the issue about final performance value logging as '-1' is not resolved. So after finding the root cause, that will be dealt in other PR.


sungmanc (dismissed):

sungmanc:
So much thanks for nice bug resolving !

eunwoosh (dismissed):
Nice work!

sungmanc (approved):

eunwoosh (approved):

PR #1560, opened 2023-01-20
[OTX][Hot-fix] Fix instance segmentation error

sungmanc:
# Changes
* Fix mis-input labels for instance segmentation.
* Add a function that removes unused label entities to make the correct `Label Schema`.
* Add `task check` logic for the detection adapter: To handle instance segmentation, detection, both.

# Details
* To consider the semantic segmentation, Datumaro shouldn't include the `background` for labels 
and the label file has `{1: 'person'}` format. In this case `labels` like as below:
```python3
self.label_entities = [LabelEntity(1, 'person', ...)]
```

So, we need to subtract 1 from the polygon's label to match the polygon's label with `label.entities`. 
```python3
d_polygon.label -= 1
```

* For instance segmentation, we don't need to consider the background label. So, it must be implemented like as below:
```python3
labels=[ScoredLabel(label=self.label_entities[annotation.label])],
```

### DIFF
--- a/otx/core/data/adapter/base_dataset_adapter.py
+++ b/otx/core/data/adapter/base_dataset_adapter.py
@@ -250,5 +250,5 @@ class BaseDatasetAdapter(metaclass=abc.ABCMeta):
                     for i in range(0, len(annotation.points), 2)
                 ]
             ),
-            labels=[ScoredLabel(label=self.label_entities[annotation.label - 1])],
+            labels=[ScoredLabel(label=self.label_entities[annotation.label])],
         )
--- a/otx/core/data/adapter/base_dataset_adapter.py
+++ b/otx/core/data/adapter/base_dataset_adapter.py
@@ -252,3 +252,18 @@ class BaseDatasetAdapter(metaclass=abc.ABCMeta):
             ),
             labels=[ScoredLabel(label=self.label_entities[annotation.label])],
         )
+
+    def remove_unused_label_entities(self, used_labels: List):
+        """Remove unused label from label entities.
+        
+        Because label entities will be used to make Label Schema,
+        If there is unused label in Label Schema, it will hurts the model performance.
+        So, remove the unused label from label entities.
+
+        Args:
+            used_labels (List): list for index of used label
+        """
+        clean_label_entities = []
+        for used_label in used_labels:
+            clean_label_entities.append(self.label_entities[used_label])
+        self.label_entities = clean_label_entities
--- a/otx/core/data/adapter/base_dataset_adapter.py
+++ b/otx/core/data/adapter/base_dataset_adapter.py
@@ -255,7 +255,7 @@ class BaseDatasetAdapter(metaclass=abc.ABCMeta):
 
     def remove_unused_label_entities(self, used_labels: List):
         """Remove unused label from label entities.
-        
+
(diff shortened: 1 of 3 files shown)


sovrasov (dismissed):

supersoob (dismissed):
Thanks for the hard work!

sungmanc:
Here I attached the performance difference and I think we can merge this. @sovrasov @supersoob .
Could you check?

Before merging Datumaro (Commit: b1a8c036d2f5cfb994b3d77f93c9a24e03dc2f04)
image

This PR
image

supersoob (dismissed):

otx/core/data/adapter/base_dataset_adapter.py
```diff
@@ -250,5 +250,20 @@ def _get_polygon_entity(self, annotation: DatumaroAnnotation, width: int, height
...
+        If there is unused label in Label Schema, it will hurts the model performance.
+        So, remove the unused label from label entities.
+
+        Args:
+            used_labels (List): list for index of used label
+        """
+        clean_label_entities = []
+        for used_label in used_labels:
+            clean_label_entities.append(self.label_entities[used_label])
+        self.label_entities = clean_label_entities
```
harimkang: This seems to give a no newline error. Does the pre-commit pass?

sungmanc: I missed the pre-commit test ! However, after the pre-commit, no big differences..
https://github.com/openvinotoolkit/training_extensions/pull/1560/commits/b259db3da9520682dd1b4e35b985c1c84905de35


image

eugene123tw (approved):

otx/core/data/adapter/segmentation_dataset_adapter.py
```diff
@@ -41,9 +42,13 @@ def get_otx_dataset(self) -> DatasetEntity:
...
                             for d_polygon in datumaro_polygons:
                                 if d_polygon.label != 0:
+                                    d_polygon.label -= 1
                                     shapes.append(self._get_polygon_entity(d_polygon, image.width, image.height))
+                                    used_labels.append(d_polygon.label)
 
                     dataset_item = DatasetItemEntity(image, self._get_ann_scene_entity(shapes), subset=subset)
                     dataset_items.append(dataset_item)
 
+        # Remove unused labels from label entities
```
sovrasov, line 52: Does this comment bring any additional information? The method name looks verbose enough

sungmanc: I added a detailed description inside the function, do you want to add a detailed comment on this line?

harimkang (approved):
LGTM!

eugene123tw (approved):

otx/core/data/adapter/segmentation_dataset_adapter.py
sungmanc: @sovrasov , If you have other opinions, could you let me know? I'll apply it. If not, do I merge this PR?

sovrasov: I actually wanted to tell that we don't need this and similar comments at all, since we have a clear function name and a docstring. Feel free to merge this.

PR #1562, opened 2023-01-25
[OTX][Hot-fix] Add init_weights call in the mmcls builder.py

JihwanEom:
Same bug with #1554, but the classifier head was not initialized until now. (not caused by public mmcv PR #1497 )

### DIFF
--- a/otx/algorithms/classification/adapters/mmcls/utils/builder.py
+++ b/otx/algorithms/classification/adapters/mmcls/utils/builder.py
@@ -8,7 +8,11 @@ from typing import Optional, Union
 
 import torch
 from mmcv.runner import load_checkpoint
-from mmcv.utils import Config, ConfigDict
+from mmcv.utils import Config, ConfigDict, get_logger
+
+from otx.mpa.utils.logger import LEVEL
+
+logger = get_logger("mmcls")
 
 
 def build_classifier(
@@ -31,6 +35,9 @@ def build_classifier(
 
     model_cfg = deepcopy(config.model)
     model = origin_build_classifier(model_cfg)
+    logger.setLevel("WARNING")
+    model.init_weights()
+    logger.setLevel(LEVEL)
     model = model.to(device)
 
     checkpoint = checkpoint if checkpoint else config.pop("load_from", None)
--- a/otx/algorithms/segmentation/adapters/mmseg/utils/builder.py
+++ b/otx/algorithms/segmentation/adapters/mmseg/utils/builder.py
@@ -8,7 +8,11 @@ from typing import Optional, Union
 
 import torch
 from mmcv.runner import load_checkpoint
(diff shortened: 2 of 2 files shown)


otx/algorithms/classification/adapters/mmcls/utils/builder.py
```diff
@@ -31,6 +35,9 @@ def build_classifier(
 
     model_cfg = deepcopy(config.model)
     model = origin_build_classifier(model_cfg)
+    logger.setLevel("WARNING")
+    model.init_weights()
```
harimkang, line 39: What is the difference between `init_weight` and `load_checkpoint` (line 45)?

harimkang:
Shouldn't this be segmentation as well?

otx/algorithms/classification/adapters/mmcls/utils/builder.py
JihwanEom: `init_weight` calls initialization methods like `Normal init` or `Linear init`. `load_checkpoint` loads incoming checkpoint from Imagenet or `load-from` argument.

harimkang (dismissed):
LGTM

supersoob (changes requested):
I don't think it's good to add the init_weights for seg in this PR. Plus, It already calls it inside the model backbone and heads. If you adds the line, it is redundantly added it and produces the redundant logs. Please revert it @harimkang @JihwanEom

JihwanEom:
> 

Could you check again with log? `decode_head.conv_seg` is not initialized properly. Also, there will be no additional logs because the log level has been updated.

image

supersoob:
> > 
> 
> Could you check again with log? `decode_head.conv_seg` is not initialized properly.
> 
> image

I don't understand what you want me to check and why you think it is not initialized properly. init_weights does not load load_from arg, what we actually want to load.

harimkang:
> I don't think it's good to add the init_weights for seg in this PR. Plus, It already calls it inside the model backbone and heads. If you adds the line, it is redundantly added it and produces the redundant logs. Please revert it @harimkang @JihwanEom

@jaegukhyun I think this is no different from detection. Do you have any advice?

JihwanEom:
I'll revert segmentation part, let's talk on other PR for segmentation.

JihwanEom:
> I don't think it's good to add the init_weights for seg in this PR. Plus, It already calls it inside the model backbone and heads. If you adds the line, it is redundantly added it and produces the redundant logs. Please revert it @harimkang @JihwanEom

When I add `init_weight` in builder.py, the accuracy was improved (41.75 -> 45.13 on mbv3). Check it out next time when you have a room. Sorry for inconvinience.

supersoob (approved):

harimkang (approved):

PR #1537, opened 2023-01-18
[OTX][HOT-FIX] Make clear distinction between OTX and MPA

cih9088:
It will fix an error when running `**/*_sample.py`

PR #1564, opened 2023-01-25
[OTX][Hot-fix] Fix learning curve to put training metrics statistics into output

supersoob:
- This PR is to fix learning curve to put training metrics statistics into output and fix -1 value in final performance evaluation
- This is caused because OTXLoggerHook.Curve is not static or class instance. So it cannot bring the eval logs to learning curve after the training. I fixed it by making curves as a class instance and refer it to learning curve after training.
- Since learning curve is important to record the model statistics for geti UI, we need to fix it smartly.
- No bug was found in action tasks. (Jihwan)


supersoob:
I just made curves as a class instance to resolve it. If you have a better idea, please share.

Before)
image

After fix)
image

harimkang (commented):
Thank you for finding and fixing the edge case.
Is there a reason the action is not included in this PR?
(@JihwanEom Is action something different?, The code looks the same...)

+ `otx/algorithms/common/tasks/training_base.py` also has the same code.

JihwanEom:
> Thank you for finding and fixing the edge case. Is there a reason the action is not included in this PR? (@JihwanEom Is action something different?, The code looks the same...)
> 
> * `otx/algorithms/common/tasks/training_base.py` also has the same code.

Same opinion with you, for my best understand, soobee asked me just double checking for action task. It will be updated soon.

harimkang:
> > Thank you for finding and fixing the edge case. Is there a reason the action is not included in this PR? (@JihwanEom Is action something different?, The code looks the same...)
> > 
> > * `otx/algorithms/common/tasks/training_base.py` also has the same code.
> 
> Same opinion with you, for my best understand, soobee asked me just double checking for action task. It will be updated soon.

Thank you for your confirmation. I think it would be best to work on this PR at once to manage PR history.

JihwanEom:
There was no bug on action tasks by  https://github.com/openvinotoolkit/training_extensions/pull/1564/commits/2a8686675b86e850dd0a219b7162b075ce2a9cec
https://github.com/openvinotoolkit/training_extensions/blob/feature/otx/otx/algorithms/action/tasks/train.py#L154
I also checked it still works well with changes in this PR, no side effect is expected for action tasks.
image

JihwanEom (dismissed):

eunwoosh:
The reason why it happens is that `self._learning_curves` is deepcopied. Original intention is passing the `self._learning_curves` and save the training history in the passed `self._learning_curves`. But current OTX deepcopies the `self._recipe_cfg` where `self._learning_curves` is saved and also mmdetection deepcopied `cfg`. One way to avoid this is making a subclass inheriting `defaultdict` and override a `__deepcopy__` to return itself as it is (i.e. not copied at all) as below and using it instead original `defaultdict`.
image

supersoob:
> The reason why it happens is that `self._learning_curves` is deepcopied. Original intention is passing the `self._learning_curves` and save the training history in the passed `self._learning_curves`. But current OTX deepcopies the `self._recipe_cfg` where `self._learning_curves` is saved and also mmdetection deepcopied `cfg`. One way to avoid this is making a subclass inheriting `defaultdict` and override a `__deepcopy__` to return itself as it is (i.e. not copied at all) as below and using it instead original `defaultdict`. image

I had a sync with Eunwoo. Making curves a class instance would have the same overhead but it looks like Eunwoo's approach(making uncopiable dict) is cleaner and more refined. So I would revise as Eunwoo's suggestions. How do you think? @JihwanEom @harimkang  Besides, I'm worried that the deepcopy of cfg would bring many side effects. Let's fix one by one.

otx/algorithms/common/utils/utils.py
```diff
@@ -16,13 +16,22 @@
...
 import yaml
 
 from otx.api.utils.argument_checks import YamlFilePathCheck, check_input_parameters_type
 
 
+class UncopiableDefaultDict(defaultdict):
+    """Uncopiable defaultdict."""
+
+    def __deepcopy__(self, memo):
+        """deepcopy."""
```
eunwoosh: I'm not sure but doesn't error occur because first ltter is lower case in pre-commit?

otx/algorithms/action/tasks/train.py
```diff
@@ -120,7 +120,7 @@ def train(
         else:
             update_progress_callback = default_progress_callback
         self._time_monitor = TrainingProgressCallback(update_progress_callback)
-        self._learning_curves = DefaultDict(OTXLoggerHook.Curve)
+        self._learning_curves = UncopiableDefaultDict(OTXLoggerHook.Curve)
```
eunwoosh: Is this needed even if `self._learning_curves` is initialized in `__init__()` in `BaseTask`?

otx/algorithms/common/utils/utils.py
supersoob: No error in pre-commit.

otx/algorithms/action/tasks/train.py
supersoob: I fixed it. please check again.

eunwoosh (dismissed):
Thanks for your work!

sungmanc:
@cih9088 , could you review this PR? this PR might be related with `deepcopy` for `cfg`.

otx/algorithms/common/utils/utils.py
```diff
@@ -16,13 +16,22 @@
...
 import yaml
 
 from otx.api.utils.argument_checks import YamlFilePathCheck, check_input_parameters_type
 
 
+class UncopiableDefaultDict(defaultdict):
+    """Uncopiable defaultdict."""
+
+    def __deepcopy__(self, memo):
+        """Deepcopy."""
```
JihwanEom, line 31: Could you update docstring more specifically? I believe it's hard to understand for other contributors.

jaegukhyun (dismissed):

otx/algorithms/common/utils/utils.py
supersoob: If you have some, please give and I will update. I think it's enough to understand its functionality as other methods in this file.

harimkang: I agree with Jihwan's opinion. The purpose of the docstring is to communicate what the function does. The current annotations make it difficult to understand the functionality.
It's hard to understand the meaning of `__deepcopy__` in `UncopiableDefaultDict` without looking deeply.

supersoob: Then, could you please give some idea to update? I can't brainstorm more than "defauldict type object to avoid deepcopy."

harimkang: I think your comment ("Defauldict type object to avoid deepcopy.") is enough. :)
or "A function that returns itself to avoid deepcopy." ..?

supersoob: I fixed it. Thanks.

cih9088:
I'm not sure this is the right direction. The purpose of #1537 was to make a clear boundary between OTX and MPA because they are completely different modules as I see them. I'm not sure OTX will absorb MPA later but for now, for me, they are not.
If you felt it in the same way, making this sort of stuff as a return from MPA would be an appropriate design choice.
If not, let's revert #1537 and make everything easy.

supersoob:
> I'm not sure this is the right direction. The purpose of #1537 was to make a clear boundary between OTX and MPA because they are completely different modules as I see them. I'm not sure OTX will absorb MPA later but for now, for me, they are not. If you felt it in the same way, making this sort of stuff as a return from MPA would be an appropriate design choice. If not, let's revert #1537 and make everything easy.

OTX will absorb MPA later. Anyway, I see the deepcopy existed before #1537 and I was heard that it was needed for other reason than boundary between OTX and MPA. So this error was caused before #1537 because of `deepcopy(self._recipe_cfg)`. If you are saying that you will remove all deepcopy for cfg in tasks. I can close this PR. If not, I'd like to merge this bug-fix.

cih9088:
> OTX will absorb MPA later. Anyway, I see the deepcopy existed before #1537 and I was heard that it was needed for other reason than boundary between OTX and MPA. So this error was caused before #1537 because of `deepcopy(self._recipe_cfg)`. If you are saying that you will remove all deepcopy for cfg in tasks. I can close this PR. If not, I'd like to merge this bug-fix.

Deepcoyping `recipe_cfg` was in my commit anyway when replacing forked mm libraries to make a boundary. #1537 was to extend this boundary to `data_cfg` as well.
I'm not saying that I'm going to get rid of `deepcopy` part but we, not me of course as I'm merely a contributor and my words carry no weights, need a discussion whether removing `deepcopy` or not to make things easier.

JihwanEom (approved):

supersoob:
We discussed to remain deepcopy for cfg for avoiding causing the other issues. I checked CI tests errors are not caused by this PR. (num_classes issue in det/dataset item issue in multi-class is not relevant to this PR) I hope you approve or review if you have other opinions.

jaegukhyun (approved):

PR #1570, opened 2023-01-26
[OTX] Updated for fixing ci issues

yunchu:
- updated ci dockerfile to omit otx dependencies
- added some helper scripts to start/stop/check ci container
- revised tox.ini to specify some package dependencies as prerequisite
- runtime editing requirement files for torch + cuda versioning will not work
- added test-timeout to the ote-sdk streamer test case (prevent hanging)
- revised workflow to run the pre-commit job to the GH hosted runner

Signed-off-by: Lee, Yunchu <yunchu.lee@intel.com>

### DIFF
--- a/.ci/README.md
+++ b/.ci/README.md
@@ -46,7 +46,7 @@
    - `Enter the name of the runner group to add this runner to:`
      - Just press Enter key for default
    - `Enter the name of runner:`
-     - Input the runner name which would be used by Github. If you want to replace an exsiting runner to this one, input the same runner's name.
+     - Input the runner name which would be used by Github. If you want to replace an existing runner to this one, input the same runner's name.
    - [`Optional`] If you input the existing runner's name to replace in the step above, confirmation prompt will be shown.
      - Input 'y' for the replacing
      - `Enter name of work folder:`
--- a/.ci/README.md
+++ b/.ci/README.md
@@ -1,51 +1,91 @@
 # Guide to Setting up the CI using the Docker images
 
-1. Build the docker image using the Dockerfile in the .ci directory.
+1. Build the docker image using the `build.sh` in the .ci directory.
    Make sure you are in the root directory of `training_extensions`.
 
    ```bash
-   sudo docker build --build-arg HTTP_PROXY="$http_proxy" --build-arg HTTPS_PROXY="$https_proxy" --build-arg NO_PROXY="$no_proxy" . -t otx-ci -f .ci/Dockerfile
+   training_extensions$ .ci/build.sh --help
+     USAGE: .ci/build.sh <tag> [Options]
+     Options
+        -p|--push           Push built image(s) to registry
+        -u|--url            url to get Github actions-runner package
+        -c|--cuda           Specify CUDA version
+        -h|--help           Print this message
    ```
 
-   Here, `otx_ci` is the name of the image.
-
-1. Create and start a container
+   Below example builds an image using actions-runner v2.299.1 based on `NVIDIA CUDA 11.1.1` image and tag it as `2.299.1`.
 
    ```bash
-   sudo docker run --gpus all -i -t -d --name otx-ci-container otx-ci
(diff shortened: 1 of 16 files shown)


.ci/README.md
```diff
@@ -1,51 +1,91 @@
...
-   mkdir actions-runner && cd actions-runner
+   training_extensions$ .ci/start-runner.sh otx-ci-container <github-token> <codacy-token> -g 0
+   ```
+
+   Once it executed, An interactive session for the runner configuration will be started and you should input some information to each prompt.
+
+   - `Enter the name of the runner group to add this runner to:`
+     - Just press Enter key for default
+   - `Enter the name of runner:`
+     - Input the runner name which would be used by Github. If you want to replace an exsiting runner to this one, input the same runner's name.
```
JihwanEom: A simple typo, `exsiting` -> `existing`

requirements/action.txt
```diff
@@ -1,4 +1,4 @@
-mmpycocotools
+# mmpycocotools
```
JihwanEom, line 1: Is it intended commenting out instead of removing `mmpycocotools`?

requirements/detection.txt
```diff
@@ -1,4 +1,4 @@
-mmpycocotools
+# mmpycocotools
```
JihwanEom, line 1: Same as above.

JihwanEom:
Thank you very much for your effort !!

.ci/README.md
yunchu: thanks!

requirements/action.txt
yunchu: i've removed all dependencies of the mmpycocotools from requirements for resolving some conflict while setting testenv but i'm not sure whether it could make any side effect or not. so i've left them with comment and wanted to let the task owner know what was changed on dependency.

.github/workflows/pre_merge.yml
```diff
@@ -18,26 +18,30 @@ concurrency:
...
         uses: actions/checkout@v3
+      - name: Set up Python
+        uses: actions/setup-python@v3
+        with:
+          python-version: 3.8
       - name: Install dependencies
         run: python -m pip install tox
       - name: Code Quality Checks
-        run: tox -re pre-commit
+        run: tox -e pre-commit
```
sungmanc, line 32: Is it okay when someone changes the requirements?

yunchu: AFAIK, the dependency changes will re-create env and -r is an option for the "force re-create"

requirements/action.txt
JihwanEom: Thank you for kind explanation! @sungmanc do you know the history of `mmpycocotools`? 😮

JihwanEom (approved):

.github/workflows/pre_merge.yml
sungmanc: Thanks !

requirements/action.txt
sungmanc: @JihwanEom , in my knowledge, mmpycocotools is not used anymore since we changed mmdetection as a latest version

sungmanc (approved):

eunwoosh:
I have a question. After this PR is applied, should we use python3.8 and torch1.8.2 when we use pip install or execute tox?

yunchu:
> I have a question. After this PR is applied, should we use python3.8 and torch1.8.2 when we use pip install or execute tox?

yes, i think so. and i'm not sure whether current implementation can work with different version of python. i've failed when i tested on the python 3.7 or 3.9 from the develop branch before.
and different torch version can be used if it works as expected. one thing I can confirm is that torch, torchvision, mmcv-full and cuda version should be matched for working correctly.
for the tox, i think it would be used only for run the testing not for the develop so i think all developers don't need to care about it. (please correct me if i'm wrong)

harimkang:
> I have a question. After this PR is applied, should we use python3.8 and torch1.8.2 when we use pip install or execute tox?

In fact, I have been experimenting and testing on python3.9.. So maybe it still works with python3.9? :)

PR #1572, opened 2023-01-27
[OTX][Hot-fix] Inst-seg, Det + Simple-Autoconfig test Hotfix

sungmanc:
# Summary
* Fix `detection_adapter.py` to resolve inst-seg, det issue: wrong implementation
* Fix `test_cli.py`: not use workspace in otx_root, use /tmp directory to save the workspace

### DIFF
--- a/otx/cli/utils/tests.py
+++ b/otx/cli/utils/tests.py
@@ -733,7 +733,7 @@ def otx_build_backbone_testing(root, backbone_args):
 
 
 def otx_build_auto_config(root, otx_dir: str, args: Dict[str, str]):
-    workspace_root = os.path.join(root, "otx-workspace-CLASSIFICATION")
+    workspace_root = os.path.join(root, "otx-workspace")
     command_line = [
         "otx",
         "build",
--- a/otx/cli/utils/tests.py
+++ b/otx/cli/utils/tests.py
@@ -732,10 +732,13 @@ def otx_build_backbone_testing(root, backbone_args):
     ), f"{model_config['model']['backbone']['type']} != {backbone}"
 
 
-def otx_build_auto_config(otx_dir: str, args: Dict[str, str]):
+def otx_build_auto_config(root, otx_dir: str, args: Dict[str, str]):
+    workspace_root = os.path.join(root, "otx-workspace-CLASSIFICATION")
     command_line = [
         "otx",
         "build",
+        "--workspace-root",
+        workspace_root
     ]
 
     for option, val in args.items():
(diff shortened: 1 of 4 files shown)


otx/cli/utils/tests.py
```diff
@@ -732,10 +732,13 @@ def otx_build_backbone_testing(root, backbone_args):
     ), f"{model_config['model']['backbone']['type']} != {backbone}"
 
 
-def otx_build_auto_config(otx_dir: str, args: Dict[str, str]):
+def otx_build_auto_config(root, otx_dir: str, args: Dict[str, str]):
+    workspace_root = os.path.join(root, "otx-workspace-CLASSIFICATION")
```
eunwoosh: Is this function executed only for classification?

sungmanc: For now, yes. I'll add for detection, segmentation later :)

eunwoosh: Then how about using general name to consider that?

sungmanc: It sounds good, thanks for opinion. --> Edited

eunwoosh (dismissed):

otx/core/data/adapter/detection_dataset_adapter.py
```diff
@@ -36,16 +36,18 @@ def get_otx_dataset(self) -> DatasetEntity:
                     image = Image(file_path=datumaro_item.media.path)
                     shapes = []
                     for ann in datumaro_item.annotations:
-                        if self.task_type is TaskType.DETECTION and ann.type == AnnotationType.polygon:
+                        if (
+                            str(self.task_type) is TaskType.INSTANCE_SEGMENTATION.name
```
cih9088: Why do we need `str(*)` and `*.name`?

harimkang: Can't we just use `self.task_type is TaskType.INSTANCE_SEGMENTATION`?

harimkang (dismissed):
When I run it locally, `test_cli.py` seems to work fine in `/tmp`. Thanks for the correction.
Everything looks good except for the question about the modification with an if statement comparing str to name.

otx/core/data/adapter/detection_dataset_adapter.py
harimkang: This part was corrected after discussing with sungman.
https://github.com/openvinotoolkit/training_extensions/pull/1572/commits/19c016e8cd4d03a38bb489e2d933c0724f108ebe

cih9088 (dismissed):

cih9088 (approved):

harimkang (approved):

PR #1573, opened 2023-01-27
[OTX] Execute cleanup function instead unload

eunwoosh:
### Summary

- execute cleanup function instead unload

### Detail
There is a bug that segmentation fault is thrown while anomaly test cases are running in CI. This is because `unload()` in anomaly task throws segmentation fault on purpose in docker environment. To avoid this, separate code part to remove work directory and use it.

### DIFF
--- a/otx/algorithms/anomaly/tasks/inference.py
+++ b/otx/algorithms/anomaly/tasks/inference.py
@@ -88,7 +88,11 @@ class InferenceTask(IInferenceTask, IEvaluationTask, IExportTask, IUnload):
         self.base_dir = os.path.abspath(os.path.dirname(template_file_path))
 
         # Hyperparameters.
-        self.project_path: str = output_path if output_path is not None else tempfile.mkdtemp(prefix="otx-anomalib")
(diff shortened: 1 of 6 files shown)


sungmanc (approved):
Nice work !

otx/algorithms/anomaly/tasks/inference.py
```diff
@@ -337,8 +341,7 @@ def _is_docker() -> bool:
 
     def unload(self) -> None:
         """Unload the task."""
-        if os.path.exists(self.config.project.path):
-            shutil.rmtree(self.config.project.path, ignore_errors=False)
+        self.cleanup()
 
         if self._is_docker():
```
harimkang, line 346: Wouldn't it still give a segmentation fault on this code? It looks like we need to check in CI.

harimkang: This confirmed that the existing issue was not reproducible in CI.

harimkang (approved):

PR #1576, opened 2023-01-27
[OTX][Hotfix] Fix OpenVINO export bug and dataset import bug

eugene123tw:
* Patch test pipeline in DetectionExporter instead of universal Naive Exporter

### DIFF
--- a/otx/core/data/adapter/detection_dataset_adapter.py
+++ b/otx/core/data/adapter/detection_dataset_adapter.py
@@ -42,9 +42,6 @@ class DetectionDatasetAdapter(BaseDatasetAdapter):
                         if self.task_type is TaskType.DETECTION and ann.type == AnnotationType.bbox:
                             shapes.append(self._get_normalized_bbox_entity(ann, image.width, image.height))
                             used_labels.append(ann.label)
-                        elif ann.type == AnnotationType.polygon:
-                            shapes.append(self._get_polygon_entity(ann, image.width, image.height))
-                            used_labels.append(ann.label)
 
                     dataset_item = DatasetItemEntity(image, self._get_ann_scene_entity(shapes), subset=subset)
                     dataset_items.append(dataset_item)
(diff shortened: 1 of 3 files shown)


harimkang (approved):
Thanks for the quick fix! 
Now, CI is working normally, so if there is nothing wrong with the test results, I think it can be merged. (The CI test seems to take about two hours, I'll take a look when it's over.)

harimkang:
스크린샷 2023-01-27 오후 11 58 35
There is one instance-segmentation nncf test fail. It seems to be a fail that has nothing to do with this PR. (i'm not sure..)
@cih9088 Could you take a look when you have some time?
https://github.com/openvinotoolkit/training_extensions/actions/runs/4023635943/jobs/6915085992

cih9088 (commented):
Why do we need tiling for `test` and `val` datasets? Aren't these datasets supposed to be normal datasets not wrapped ones? Should `test` dataset be unwrapped dataset at least?

cih9088:
> (i'm not sure..) @cih9088 Could you take a look when you have some time? 

It seems it happens from time to time. I can't reproduce it.

eugene123tw:
> Why do we need tiling for `test` and `val` datasets? Aren't these datasets supposed to be normal datasets not wrapped ones? Should `test` dataset be unwrapped dataset at least?

@cih9088  I'm not quite sure I understand your questions correctly. But if a model is trained with tiling, then the same logic should apply to inference otherwise you wouldn't get the desired results.

cih9088:
> @cih9088 I'm not quite sure I understand your questions correctly. But if a model is trained with tiling, then the same logic should apply to inference otherwise you wouldn't get the desired results.

Sorry about the confusion. I thought tiling was a sort of training time augmentation.
Using wrapped dataset for test should be somehow generalized like you tried to implement it in `apis.py`. Could you add `TODO` comment?

cih9088 (approved):

PR #1581, opened 2023-01-30
[OTX] Handle wrapper dataset when exporting model

cih9088:
related #1576

### DIFF
--- a/otx/mpa/det/exporter.py
+++ b/otx/mpa/det/exporter.py
@@ -3,7 +3,6 @@
 #
 
 import numpy as np
-from mmcv import ConfigDict
 from mmcv.runner import wrap_fp16_model
 
 from otx.mpa.exporter_mixin import ExporterMixin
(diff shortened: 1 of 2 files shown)


eugene123tw (dismissed):
Thanks for adding this and making it generic. I only have one minor comment as I think tiling is a special case, and possibly most dataset wrappers would retain `LoadImage` in their dataset pipeline, in this case, we might encounter unexpected failure or other edge cases. 

The reason I patched this in detection is just that I think it's an edge (special) case and it might not happen to other tasks.

harimkang:
Thank you for your hard work. Could you please apply the latest commit? It is necessary for CI test to run. :)

cih9088:
> Thanks for adding this and making it generic. I only have one minor comment as I think tiling is a special case, and possibly most dataset wrappers would retain `LoadImage` in their dataset pipeline, in this case, we might encounter unexpected failure or other edge cases.

I believe it is highly unlikely that pipeline of wrappers have `LoadImage` transformation. Instead, it would be included in *inner* dataset pipeline as `tiling` wrapper case. For other datasets, not wrapper, this block would not executed as there is no `dataset` or `datasets` keywords in config. Only caveat here is, I think, complicated pipeline logic unlike `tiling` but let's leave it for future work as there is no other tasks using dataset wrappers when testing in OTX for now.

eugene123tw (dismissed):

sungmanc (dismissed):

sungmanc (approved):

eugene123tw (approved):

PR #1584, opened 2023-01-30
[OTX][HOT-FIX] Resolve replacement backbone issue

harimkang:
## Summary
- Resolve replacement backbone issue with build.py
- Resolve mmcls init_cfg isssue

## Test Results (tests/integration/cli/test_cli.py)
image

### DIFF
--- a/otx/cli/tools/build.py
+++ b/otx/cli/tools/build.py
@@ -66,6 +66,7 @@ def main():
     builder = Builder()
 
     otx_root = get_otx_root_path()
+    update_data_cfg = False
 
     # Auto-configuration
     config_manager = ConfigManager()
@@ -75,6 +76,7 @@ def main():
             args.task = task_type
         if args.val_data_roots is None:
             config_manager.auto_split_data(args.train_data_roots, args.task)
+        update_data_cfg = True
 
     if args.task and args.task in SUPPORTED_TASKS:
         # Build with task_type and create user workspace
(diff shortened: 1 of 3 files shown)


jaegukhyun (dismissed):
LGTM, Any test results?

otx/cli/tools/build.py
```diff
@@ -103,11 +103,12 @@ def main():
                 raise ValueError("The selected backbone has inputs that the user must enter.")
             builder.merge_backbone(args.model, args.backbone)
 
-    config_manager.write_data_with_cfg(
-        workspace_dir=args.workspace_root,
-        train_data_roots=args.train_data_roots,
-        val_data_roots=args.val_data_roots,
-    )
+    if args.workspace_root is not None:
```
sungmanc: If a user gives `--backbones` with `--workspace_root`, it still causes the problem.

So, I suggest moving this (line#107-111) into line#90. In this case, we also could move line#81-82 to line#78

harimkang: Thanks for finding the problem with this code.
But your suggestion still generates `data.yaml` in unintended circumstances. 
In my opinion, supplementing the if statement in line 106 looks good.
How about `if not args.workspace_root and config_manager.splitted_dataset:` ?

harimkang: Oh, there is a problem with my suggestion. Let's think about it some more.

harimkang: Could you please check this https://github.com/openvinotoolkit/training_extensions/pull/1584/commits/c396f181c15821ee8c82fbb9daec5da9fa897bf3 ?

harimkang: It works as intended for the edge case you commented.

sungmanc: Looks good :)

sungmanc (approved):

harimkang:
> LGTM, Any test results?

image
I think this works fine, but lacks test cases (train after build -> X). I will add it later.

jaegukhyun (approved):

PR #1592, opened 2023-01-31
[OTX] Enable auto configuration in train.py

sungmanc:
# Summary
* Edit the builder, build to enable auto-config in `otx train`
* To avoid the duplicate implementation, I modified pre-exisited `build.py` to use `build` function in `train.py`.
* TODO
  * Need to consider save_output
  * Need to consider edge case


otx/cli/builder/builder.py
```diff
@@ -155,25 +156,27 @@ def build_task_config(
...
-            if len(template_lst) == 0:
-                raise ValueError(
-                    f"[otx build] {model_type} is not a type supported by OTX {task_type}."
-                    f"\n[otx build] Please refer to 'otx find --template --task_type {task_type}'"
-                )
-            template = template_lst[0]
-        else:
-            template = otx_registry.get(DEFAULT_MODEL_TEMPLATE_ID[task_type.upper()])
-        template_dir = Path(template.model_template_path).parent
+        if not template:
```
harimkang: Wouldn't it be possible to remove the contents of this if statement as well? What do you think?
Does the case where template is None still exist?

otx/cli/builder/builder.py
```diff
@@ -155,25 +156,27 @@ def build_task_config(
...
+                template = template_lst[0]
+            else:
+                template = otx_registry.get(DEFAULT_MODEL_TEMPLATE_ID[task_type.upper()])
+
+        assert template is not None
+        template_dir: Path = Path(template.model_template_path).parent
 
         # Copy task base configuration file
-        task_configuration_path = template_dir / template.hyper_parameters.base_path
+        task_configuration_path = template_dir / str(template.hyper_parameters.base_path)
```
harimkang: Is this causing a problem now? As far as I know, ModelTemplate's hyper_parameters.base_path is str or os.path.

otx/cli/tools/build.py
```diff
@@ -57,56 +58,102 @@ def parse_args():
...
-            train_data_roots=args.train_data_roots,
-            val_data_roots=args.val_data_roots,
+            workspace_dir=workspace_root,
+            train_data_roots=train_data_roots,
+            val_data_roots=val_data_roots,
         )
 
     # Build Backbone related
-    if args.backbone:
+    if backbone:
```
harimkang, line 127: How about tying it into another function from here on out? (line 121- 133)

otx/cli/tools/build.py
```diff
@@ -57,56 +58,102 @@ def parse_args():
...
+    template: ModelTemplate = None,
+):
+    """Makes configuration files (data.yaml, template.yaml, ...) in workspace directory.
+
+    Args:
+        builder (Builder): Builder class
+        train_data_roots (str, optional): The path of training dataset. Defaults to None.
+        val_data_roots (str, optional): The path of validation dataset. Defaults to None.
+        task (str, optional): The name of task (i.e. classification, detection, segmentation, ..). Defaults to None.
+        train_type (str, optional): The type of training (i.e. incremental, semi-sl, ..). Defaults to "incremental".
```
JihwanEom: Are both `semi-sl` and `semisl` supported?

otx/cli/tools/build.py
```diff
@@ -29,9 +30,9 @@
 SUPPORTED_TRAIN_TYPE = ("incremental", "semisl", "selfsl")
 
 
-def set_workspace(task, model):
-    """Set workspace path according to path, task, model arugments."""
-    path = f"./otx-workspace-{task}"
+def set_workspace(task: str, model: str = None, root: str = None, name: str = "otx-workspace"):
+    """Set workspace path according to arugments."""
```
JihwanEom:
```suggestion
    """Set workspace path according to arguments."""
```

otx/cli/tools/train.py
```diff
@@ -144,15 +135,50 @@ def parse_args():
...
-    # Dynamically create an argument parser based on override parameters.
-    args, template, hyper_parameters = parse_args()
+    parser = get_parser()
+
+    args, _ = parser.parse_known_args()
+    train_workspace_path = os.getcwd()
+
+    has_template_yaml = Path(args.template).exists()
+    has_data_yaml = Path(args.data).exists()
+    was_built_before = has_template_yaml and has_data_yaml
```
harimkang: how about `use_workspace = Path(args.template).exists() and Path(args.data).exists()`?
and `args.data` and `args.template` are always non-None, right?

otx/cli/tools/train.py
```diff
@@ -144,15 +135,50 @@ def parse_args():
...
+            train_data_roots=args.train_data_roots,
+            val_data_roots=args.val_data_roots,
+            workspace_root=train_workspace_path,
+            otx_root=get_otx_root_path(),
+            template=template,
+        )
+
+    # Update configurations made by builder
+    # When users input template argument, need to overwrite
+    if args.template != "./template.yaml":
```
harimkang: Could lines 152-165 have anything to do with this code?
If so, I think it would be okay to attach this to line 167 as an else.

otx/cli/tools/train.py
```diff
@@ -144,15 +135,50 @@ def parse_args():
...
+        template = find_and_parse_model_template(args.template)
+    else:
+        template = find_and_parse_model_template(os.path.join(train_workspace_path, "template.yaml"))
+    args.data = os.path.join(train_workspace_path, "data.yaml")
+    data_config = configure_dataset(args)
+
+    # Get hyper parameters schema.
+    hyper_parameters = template.hyper_parameters.data
+    assert hyper_parameters
```
harimkang: From what I remember, we decided to avoid using asserts as much as possible. How about raising issue?

otx/cli/tools/build.py
sungmanc: I'm sorry, I can't understand what you want to said

sungmanc: In my doc-string, I just typed `semi-sl`

otx/cli/builder/builder.py
sungmanc: It makes mypy error. Because `template` can be None.

otx/cli/tools/build.py
harimkang: I think it's a typo error. Isn't `semisl` correct?

otx/cli/tools/train.py
sungmanc: In the below line, we need to check whether there is `template.yaml` or not. So, I implemented this way.
The default value of `args.data` and `args.template` is `./data.yaml` and `./template.yaml`, respectively.

otx/cli/builder/builder.py
harimkang: Then shouldn't we raise an error when the template is None?

otx/cli/tools/build.py
sungmanc: This is the preference problem. someone can write the Semi Supervised Learning as Semi-SL, SemiSL, semi-sl, semisl, semi-supervised learning. 

image

otx/cli/tools/train.py
harimkang: Yes, I understand. How about using `use_workspace` instead of `was_built_before`?

sungmanc: For the name of variable, Sure :)

otx/cli/tools/build.py
harimkang: I was talking about this.
image
We have to pick one of these.
Due to the unusual folder structure of configs, a fixed wording has to be used.

JihwanEom: Sorry for confusing comment, absolutely same opinion with Harim.

sungmanc: Understood, will edit this

otx/cli/builder/builder.py
sungmanc: Sure, when user execute `otx build --train-data-roots={}`, the template is None. In this case, we need to find template

otx/cli/builder/builder.py
sungmanc: > Then shouldn't we raise an error when the template is None?

For usecase scenario, template can't be None at this line. What I meant is just for mypy checking.
Since one of the argument of `build_task_config` is ```template: ModelTemplate = None```,
mypy recognized template can be None. So, I explicitly gave a type.

Do you have other idea to resolve this mypy issue?

otx/cli/tools/build.py
sungmanc: Thanks :)

otx/cli/tools/build.py
sungmanc: I agree, Maybe we need to refactor the `builder` and `build` and `train`. How about doing this another PR?

otx/cli/builder/builder.py
```diff
@@ -144,6 +144,7 @@ def build_task_config(
         model_type: str = None,
         train_type: str = "incremental",
         otx_root: str = ".",
+        template: ModelTemplate = None,
```
jaegukhyun, line 147: How about Optional[ModelTemplate]

otx/cli/tools/train.py
sungmanc: We should separate this line as is.

Here are the possible use cases.

1. `otx build` --> `otx train --train-data-roots={COCO}`
2. `otx build` --> `otx train --train-data-roots={COCO} --template={SSD}`
3. `otx train --train-data-roots={COCO}`

So, the first thing we have to do is to check whether the workspace exists or not. (`if not was_built_before`).

Let's assume the scenario2, then we need to check the user's input argument (`args.template`).
In this scenario, `otx build` made a `template.yaml` that used ATSS model.
If user gives SSD template to OTX (i.e. `otx train --template={SSD}`), we need to train the model with SSD template.

To be honset, imagining all use cases is quite complex, so if you have any other idea, please let me know

otx/cli/tools/train.py
sungmanc: Sure, no problem. I just used previous implemented code [line62](https://github.com/openvinotoolkit/training_extensions/pull/1592/files#diff-d4366781a54d3bbd5d4c49115d58e31a837715214588d4409e388a359bbf2b9eL62)

otx/cli/tools/train.py
```diff
@@ -144,15 +135,50 @@ def parse_args():
...
+
+    args, _ = parser.parse_known_args()
+    train_workspace_path = os.getcwd()
+
+    has_template_yaml = Path(args.template).exists()
+    has_data_yaml = Path(args.data).exists()
+    was_built_before = has_template_yaml and has_data_yaml
+    if not was_built_before:
+        # Prepare build
+        train_workspace_path = os.path.join(args.save_model_to)
```
eunwoosh: isn't `os.path.join(args.save_model_to)` same as just `args.save_model_to`?
If then, why don't you write a code just as `train_workspace_path = args.save_model_to`?

otx/cli/tools/build.py
```diff
@@ -57,56 +58,102 @@ def parse_args():
...
+    builder: Builder,
+    train_data_roots: str = None,
+    val_data_roots: str = None,
+    task: str = None,
+    train_type: str = "incremental",
+    workspace_root: str = None,
+    model: str = None,
+    backbone: str = None,
+    save_backbone_to: str = None,
+    otx_root: str = ".",
```
eunwoosh: How about using `otx.cli.utils.importing.get_otx_root_path` rather than getting it as argument?

otx/cli/tools/build.py
```diff
@@ -57,56 +58,102 @@ def parse_args():
...
-    args = parse_args()
-    args.task = args.task.upper() if args.task is not None else args.task
-
-    builder = Builder()
-
-    otx_root = get_otx_root_path()
+# pylint: disable=too-many-arguments
+def build(
+    builder: Builder,
+    train_data_roots: str = None,
```
eunwoosh: I think type hinting `Optional` needs to be added to some arguments of this function.

otx/cli/tools/train.py
eunwoosh: There is a change that a bug exists, I think. If user give the template argument which doesn't exists then, this code can find the template because builder makes a template in current directory.

otx/cli/tools/train.py
```diff
@@ -42,29 +45,16 @@
...
-    else:
-        pre_parser.add_argument("template")
-        parsed, _ = pre_parser.parse_known_args()
-        template_path = parsed.template
-    # Load template.yaml file.
-    template = find_and_parse_model_template(template_path)
-    # Get hyper parameters schema.
-    hyper_parameters = template.hyper_parameters.data
-    assert hyper_parameters
+    parser.add_argument("--template", required=False, default="./template.yaml", help="Template file.")
```
eunwoosh, line 69: Is there any reason why chaning template argument from position to optional? (i.e. why `template` -> `--template`?)

otx/cli/tools/build.py
```diff
@@ -57,56 +58,102 @@ def parse_args():
...
-
-    args = parse_args()
-    args.task = args.task.upper() if args.task is not None else args.task
-
-    builder = Builder()
-
-    otx_root = get_otx_root_path()
+# pylint: disable=too-many-arguments
+def build(
+    builder: Builder,
```
eunwoosh, line 65: How about initializing builder with `Builder()` in the function rather than getting it as an argument?

otx/cli/builder/builder.py
eunwoosh: Type hinting `Optional` need to be added in this line, I think.
and also need to add an information of this argument in below comment.

otx/cli/tools/train.py
sungmanc: Thanks for comment, I'll edit

otx/cli/tools/build.py
sungmanc: I agree

otx/cli/tools/build.py
sungmanc: Thanks for comment, I'll edit

otx/cli/tools/train.py
sungmanc: From now, `template` argument should be optional, because we can use `otx train --train-data-roots={}` without `template`.

otx/cli/tools/build.py
sungmanc: When user read the `train.py`, I think explicitly constructing `builder` in `train.py` is better option.

I think we need to refactor for overall classes. So, how about leave as is? we can change overall flow

sungmanc:
@harimkang , BTW, could you let me know the reason why we don't use the assert?
IMO, assert is more simple than raise because it can be written in single line.

@eunwoosh , @harimkang  I reflected your comments to the latest commit or left my answers.

harimkang:
> @harimkang , BTW, could you let me know the reason why we don't use the assert?
> 
> IMO, assert is more simple than raise because it can be written in single line.
> 
> 
> 
> @eunwoosh , @harimkang  I reflected your comments to the latest commit or left my answers.

As far as I know, `assert` won't work if `__debug__`(in python) is false.  After some searching, it is recommended to use it mainly in test code.  We also discussed using raise in group meetings. @goodsong81

sungmanc:
> > @harimkang , BTW, could you let me know the reason why we don't use the assert?
> > IMO, assert is more simple than raise because it can be written in single line.
> > @eunwoosh , @harimkang  I reflected your comments to the latest commit or left my answers.
> 
> As far as I know, `assert` won't work if `__debug__`(in python) is false. After some searching, it is recommended to use it mainly in test code. We also discussed using raise in group meetings. @goodsong81

Thanks for clarification :)

goodsong81:
For production code, explicit "if some condition, raise exception" should be preferred.
But, let's don't find and fix all existing asserts :)

otx/cli/tools/train.py
```diff
@@ -101,6 +91,7 @@ def parse_args():
     parser.add_argument(
         "--save-model-to",
         required=False,
+        default="results",
```
kprokofi, line 108: Maybe "output_model" or "trained_model"

otx/cli/tools/build.py
```diff
@@ -57,56 +59,100 @@ def parse_args():
...
+            missing_args = builder.build_backbone_config(backbone, save_backbone_to)
+            backbone = save_backbone_to
+        if model:
             if missing_args:
                 raise ValueError("The selected backbone has inputs that the user must enter.")
-            builder.merge_backbone(args.model, args.backbone)
+            builder.merge_backbone(model, backbone)
+
+
+def main():
```
kprokofi, line 142: Do we need main functions? For what reasons? If we want to test that everything is working, maybe unit tests is a better choice?
Overwise, this main function also should be supported and it adds code to the file

otx/cli/builder/builder.py
```diff
@@ -155,33 +156,37 @@ def build_task_config(
...
+                if len(template_lst) == 0:
+                    raise ValueError(
+                        f"[otx build] {model_type} is not a type supported by OTX {task_type}."
+                        f"\n[otx build] Please refer to 'otx find --template --task_type {task_type}'"
+                    )
+                template = template_lst[0]
+            else:
+                template = otx_registry.get(DEFAULT_MODEL_TEMPLATE_ID[task_type.upper()])
+
+        if template is None:
```
kprokofi: We already checked "if not template". Is it still can be None?
As I see in the code it can't be None

otx/cli/tools/train.py
```diff
@@ -144,15 +135,62 @@ def parse_args():
...
+def main():
     """Main function that is used for model training."""
-    # Dynamically create an argument parser based on override parameters.
-    args, template, hyper_parameters = parse_args()
+    parser = get_parser()
+
+    args, _ = parser.parse_known_args()
+    train_workspace_path = os.getcwd()
+
+    default_workspace_components = {"tempate_path": "./template.yaml", "data_path": "./data.yaml"}
```
kprokofi:
```suggestion
    default_workspace_components = {"template_path": "./template.yaml", "data_path": "./data.yaml"}
```

kprokofi (commented):
image
It is not working as expected I guess. If I pass the template path it should use my tempate not the default one
But even with this default template the error occurs
image

otx/cli/tools/build.py
```diff
@@ -57,56 +59,100 @@ def parse_args():
...
-
-    builder = Builder()
-
+# pylint: disable=too-many-arguments
+def build(
+    builder: Builder,
+    train_data_roots: Optional[str] = None,
+    val_data_roots: Optional[str] = None,
+    task: Optional[str] = None,
+    train_type: Optional[str] = "incremental",
```
eunwoosh:
```suggestion
    train_type: str = "incremental",
```

otx/cli/builder/builder.py
```diff
@@ -141,9 +141,10 @@ def build_task_config(
         self,
         task_type: str,
         workspace_path: Path,
-        model_type: str = None,
-        train_type: str = "incremental",
-        otx_root: str = ".",
+        model_type: Optional[str] = None,
+        train_type: Optional[str] = "incremental",
```
eunwoosh:
```suggestion
        train_type: str = "incremental",
```

otx/cli/tools/train.py
```diff
@@ -144,15 +135,62 @@ def parse_args():
...
+            template = None
+        # Build
+        builder = Builder()
+        build(
+            builder=builder,
+            train_data_roots=args.train_data_roots,
+            val_data_roots=args.val_data_roots,
+            workspace_root=train_workspace_path,
+            template=template,
+        )
```
eunwoosh, line 209: I'm not sure but, doesn't `template` variable have to re-initialized as newly build template.yaml?

otx/cli/tools/train.py
sungmanc: Since this folder also includes configurations (i.e.`template.yaml`, `model.py`), I named this name as `results`

otx/cli/tools/build.py
sungmanc: This is the CLI interface for `otx build`. So, if user gives arguments as `otx build --train-data-roots={}`, this main function will be executed.

BTW, @harimkang , @eunwoosh , and I discussed for the current class diagram especially the relationship between `Builder` and `ConfigManager`. As a results, we'll refactoring the overall class relationship and responsibility.

otx/cli/tools/train.py
sungmanc: Could you explain more detail? 

If you mean, `template` variable don't have to re-initialized in `build` function ?

If a user doesn't use workspace, then we need to build first. Otherwise, if the user use workspace, then we can use the `template.yaml` located in workspace.

otx/cli/builder/builder.py
sungmanc: Thanks, edited

otx/cli/builder/builder.py
sungmanc: Thanks, edited

otx/cli/tools/build.py
sungmanc: Thanks, edited

otx/cli/builder/builder.py
harimkang: In fact, in this line, template cannot be None, right?
But is it correct that if there is no current str, mypy will raise an error?

otx/cli/builder/builder.py
```diff
@@ -155,33 +156,34 @@ def build_task_config(
...
+        task_configuration_path = template_dir / str(template.hyper_parameters.base_path)
         shutil.copyfile(task_configuration_path, str(workspace_path / "configuration.yaml"))
         # Load Model Template
         template_config = MPAConfig.fromfile(template.model_template_path)
         template_config.hyper_parameters.base_path = "./configuration.yaml"
 
         # Configuration of Train Type value
         train_type_rel_path = ""
-        if train_type != "incremental":
+        if train_type is not None and train_type != "incremental":
```
harimkang: There seems to be no case where None is entered in train_type, but isn't it?

harimkang (dismissed):
It looks good. However, a full refactoring will follow in a later PR.

otx/cli/tools/build.py
harimkang: @kprokofi This is what I agree to. The main function is actually the way all CLIs use it now, so we'll leave it there, but we're going to work on reducing the code in main function. Maybe one day we can get rid of it..?

otx/cli/tools/train.py
harimkang: What you said seems to be a problem with the current structure. The use case definition we're trying to provide doesn't seem right. For this, we need to proceed with the refactoring we discussed. :)

otx/cli/tools/train.py
eunwoosh: I mean as below.
```
        template = find_and_parse_model_template(args.template)
        builder = Builder()
        build(
            builder=builder,
            train_data_roots=args.train_data_roots,
            val_data_roots=args.val_data_roots,
            workspace_root=train_workspace_path,
            template=template,
        )
        template = find_and_parse_model_template(os.path.join(train_workspace_path, "template.yaml") <= not necessary?
```

otx/cli/tools/train.py
sungmanc: Yes, we should define the discrete use cases to design the system. 
But, I think the above 3 use cases can be included as a real use case.

otx/cli/builder/builder.py
sungmanc: This is caused by `template.hyper_parameters`. The `hyper_parameters.base_path` can be None.
So, we need to define the explicit type (i.e. `str`)

image

otx/cli/tools/train.py
harimkang: Yes, but I don't think we need to satisfy all three of them in this PR. :)
First of all, this PR should only be considered number 3.

otx/cli/tools/build.py
sungmanc: Hmm, if we remove the `main()` function, is there any way to support the CLI?

otx/cli/builder/builder.py
harimkang: Then isn't `str(None)` itself strange too?

sungmanc: I don't know what happens in inner part.. However, if I remove the `str()`, I got message below:
image

otx/cli/tools/build.py
harimkang: In the `if __name__ == "__main__":` part, it would be nice to divide it into several functions with distinct functions other than main.

if __name__ == "__main__":
    get_args()
    config()
    train()

otx/cli/builder/builder.py
sungmanc: It'll be removed ! according to @eunwoosh 's opinion, I revmoed `Optional` for `train_type`. So, I `train_type is not None` can be removed

otx/cli/builder/builder.py
harimkang: Probably because this function can now receive a ModelTemplate.
Then it seems appropriate to check `if base_path is None`.

otx/cli/tools/train.py
sungmanc: Agree with you, although I tried to handle all 3 cases (maybe there are other use cases), I might miss smth.
Let's define the exact use cases and refactoring

otx/cli/tools/train.py
sungmanc: Good point, I missed. I edited

otx/cli/builder/builder.py
sungmanc: resolved. with checking the base_path

otx/cli/tools/build.py
sungmanc: Ah, sure, make sense

otx/cli/tools/train.py
sungmanc: @eunwoosh , Could you check this https://github.com/openvinotoolkit/training_extensions/pull/1592/commits/6d763e1826429a1b41760fab70550fe47e63f245?

sungmanc:
> image It is not working as expected I guess. If I pass the template path it should use my tempate not the default one But even with this default template the error occurs image

I edited some codes according to @harimkang @eunwoosh and your(@kprokofi ) comments and I added integration tests to check the CLI interfaces. https://github.com/openvinotoolkit/training_extensions/pull/1592/commits/3fa0bb3649af3ed5c026c1437680d04e13929829

Most of comments were included now. If you have any other comments, please let me know

otx/cli/tools/train.py
eunwoosh: Is template file made in current directory after build? then if args.template doesn't exists and isn't default path (i.e. not ./template.yaml), then args.template is loaded even if newly made template path is "./template.yaml"

sungmanc: > Is template file made in current directory after build? then if args.template doesn't exists and isn't default path (i.e. not ./template.yaml), then args.template is loaded even if newly made template path is "./template.yaml"

Yes, however, it will cause the error later since the template doesn't exist.

sungmanc: If user gives an correct path for `args.template`, then we need to use this argument for training.

So, I think #L180 is necessary. If a user gives a not existed template, then the user must meet the error. Because OTX can't training with not-existed template

eunwoosh: I think if `args.template = default_workspace_components["template_path"]` right after build, then user don't have to meet the error. Also, user get f"Can't find {args.template}, the default template will be used to train." message before build. So user may think training works well without any error, I think. How do you think about?

sungmanc: > I think if `args.template = default_workspace_components["template_path"]` right after build, then user don't have to meet the error.

Actually, we can't avoid the error when a user gives a weird template path. And in this case, making an error is more natural thing instead of selecting default template.

After editing the code as you mentioned, If a user gives `otx train --template=${NOT EXISTED TEMPLATE} --train-data-roots=${PATH}` in the workspace, user also meet the error. 

> Also, user get f"Can't find {args.template}, the default template will be used to train." message before build.

Even in the current implementation, the user receives this message before `build()`.

kprokofi:
I have questions regarding this: As we discussed at the meeting about auto-config functionality - we agreed that we will be using .txt files to determine and memories split for the classification dataset. However, I can see a new folder created in the results folder with copied images
image
Could you explain, why is it impossible to memorize split in other (not memory consuming) way. What if I have 1 mln images?

We can't use now `otx train path/to/template ....`, now is only possible: `otx train --template path/to/template`?

kprokofi:
Next thing:
image
Why can't I run my training again if I have folder "results"? I think it is not what the user expects to see

kprokofi:
Also a question:
As I understand, `otx` train can't be launched without passing `--val-data-root` if we don't want auto-split
I thought Datumaro will recognize validation set automatically if we have one.

otx/cli/tools/train.py
eunwoosh: As far as I understand, User can execute train command without error even if passing wrong template argument.
Below is how I understand code flow. I know I may misunderstand. If then, I'll really appreciate if you let me know that :)
```
    ...
    has_template_yaml = Path(default_workspace_components["template_path"]).exists() => doesn't exists, so it's False
    has_data_yaml = Path(default_workspace_components["data_path"]).exists()

    use_workspace = has_template_yaml and has_data_yaml
    if not use_workspace:
        # Prepare build
        train_workspace_path = args.save_model_to

        # If an user gives a weird path, then automatically selects default template by using build function.
        if Path(args.template).exists():
            template = find_and_parse_model_template(args.template)
        else: => because it doesn't exist, this branch is executed. So, template is initialized as None
            print(f"Can't find {args.template}, the default template will be used to train. ")
            template = None
        # Build
        builder = Builder()
        build(
            builder=builder,
            train_data_roots=args.train_data_roots,
            val_data_roots=args.val_data_roots,
            workspace_root=train_workspace_path,
            template=template, => build is executed with None template.
        )

    # Update configurations made by builder
    # When an user gives template argument, need to overwrite it.
    if args.template != default_workspace_components["template_path"]: => even if template is made automatically in build, this branch is executed and raise error.
        template = find_and_parse_model_template(args.template)
    else:
        template = find_and_parse_model_template(os.path.join(train_workspace_path, "template.yaml"))
    ...
```

sungmanc:
> I have questions regarding this: As we discussed at the meeting about auto-config functionality - we agreed that we will be using .txt files to determine and memories split for the classification dataset. However, I can see a new folder created in the results folder with copied images image Could you explain, why is it impossible to memorize split in other (not memory consuming) way. What if I have 1 mln images?

Currently, I need more time to use this way (saving only annotation files), I'll change the saving manner as according to our discussion. Datumaro supports lots of datasets and what I found is that it's very complex because import/export/split functions for each dataset are quite different. So, supporting all datasets with auto-configuration is quite big tasks.

> We can't use now otx train path/to/template ...., now is only possible: otx train --template path/to/template?

Yes, from now, since `template` is an optional variable,  we need to use `otx train --template path/to/template`.

> Also a question:
As I understand, otx train can't be launched without passing --val-data-root if we don't want auto-split
I thought Datumaro will recognize validation set automatically if we have one.

No, OTX train can be launched without `--val-data-roots`. CLI can't recognize whether user want to use `auto-split` or not.
So, If user gives only `--train-data-roots` then auto-split will be enabled. However, if user gives `--train-data-roots` and `--val-data-roots`, then auto-split will not be executed.

Thanks for many comments, I think we need to reduce the gap between ours. So, how about making separate sync. meeting fo the auto-configuration and current CLIs? If you agree, I'll arrange

eunwoosh (dismissed):
LGTM. Thanks for your work!

harimkang:
As a result of the CI test, there are currently many fail cases. Could you please confirm?

kprokofi:
@sungmanc , Sorry for so many questions :(
But I noticed one more problem during testing auto-config/auto-split:
When I run this command using `data.yaml`:
image
It will result in error:
image
Here is `data.yaml`:
image
This data created by `otx  build` 
However, If I run this CLI command:
image
Everything is working fine!

I hope the problem in data.yaml handling... Could you take a look?

sungmanc:
> As a result of the CI test, there are currently many fail cases. Could you please confirm?

I also checked the CI test results, we need to discuss about the `build`

sungmanc:
@kprokofi , I added the `test_data_roots` argument for `build` CLI. So, if you build with `test_data_roots`, then your problem might be resolved. In your case, you only have `train_data_roots` and `val_data_roots` after build. So, `eval` can't be executed since there is no `test_data_roots` argument.

@harimkang , After some modification, I got results below (all pass in unit-test, 1FAIL case in integ. test: nncf optimize)
image

@eunwoosh, Overall logic was changed after your review, so could you please re-check?

harimkang:
@kprokofi I think this PR (https://github.com/openvinotoolkit/training_extensions/pull/1641) could replace this. Could you please check it out?

PR #1596, opened 2023-01-31
[CI] Enable coverage report with Codecov

yunchu:
Replaced coverage report from Codacy to Codecov

Note that some tox usage is updated for the "pre-merge" env.
to run testing on your local machine, you need to set the testing target folder to the tox command like below.
(the default value is "tests/unit tests/integration")
```
tox -e pre-merge -- <testing-target-path>
```

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -1,0 +1,20 @@
+name: Daily-Regression Test
+
+on:
+  workflow_dispatch: # run on request (no need for PR)
+  schedule:
+    # every 4AM from Tue to Sat
+    - cron: "0 4 * * 2-6"
+
+jobs:
+  Regression-Tests:
+    runs-on: [self-hosted, linux, x64, dev]
+    timeout-minutes: 600
+    if: github.ref == 'ref/heads/develop'
+    steps:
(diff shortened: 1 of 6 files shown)


sungmanc (dismissed):

samet-akcay:
@yunchu, great progress!

I have two questions:
- When I check https://app.codecov.io/gh/openvinotoolkit, I cannot see `training_extensions` there. Is it because we need to set this up first, or the repo has not yet been approved?
- With this codecov configuration, will it be possible to work on forked repos and still upload the coverage for PRs?
The reason for asking the second question is that ideally developers/PR makers should be working on forks rather than branches on the main repo. One potential roadblock, however, would be that coverage report may not be uploaded for PRs made via forks (a problem that anomalib has for example).

yunchu:
@samet-akcay 
actually it's not clear yet for me since I want to keep checking the status after merged it into the default branch. I'm an admin for this repo so I think I can add it to the codecov but I did not do it yet. please give little time to check all things are working as expected without having sign up to the external app.

goodsong81 (approved):
LGTM!

sungmanc (approved):
Thanks for the nice work

PR #1605, opened 2023-02-01
[OTX][Unit-test]Add unit tests for checking otx.algorithms.action.adapters.mmaction.models

jaegukhyun:
This PR is for adding unit test codes to check otx.algorithms.action.adapters.mmaction.models

**Modification in otx**
otx/algorithms/action/adapters/mmaction/models/backbones/x3d.py: It is deleted since we don't need adapted X3D for otx after changing mmcv version.
otx/algorithms/action/adapters/mmaction/models/detectors/fast_rcnn.py: 
1. Change order of inputs for ONNXPool3D: It is bug fix, the current one is right
2. Remove inference when the model is not under onnx export: When I firstly designed the forward_infer function, it is for both inference and onnx export. However the current implementation is only for onnx export purpose. Therefore I removed the unnecessary inference part.

**Coverage:**
image


harimkang (changes requested):
image
This is an error related to this PR. Could you please check it?

jaegukhyun:
After rebase, the errors disappear
image

harimkang (approved):

supersoob (approved):

PR #1612, opened 2023-02-02
[OTX][Unit-tests] Unit tests for otx/algorithms/segmentation/tasks

supersoob:
This PR includes
- Unit tests for otx/algorithms/segmentation/tasks
- Minor bug fixes in openvino files ([CVS-103090](https://jira.devtools.intel.com/browse/CVS-103090))

[Coverage report]
image
FYI, inference.py has too many private methods used in common/training_base.py. Its coverage will be increased after making unit tests in common.


harimkang (dismissed):
LGTM!
It would be nice to run the CI test reflecting the latest commit and check if the unit test passes normally in CI.

tests/unit/algorithms/segmentation/tasks/test_segmentation_inference.py
```diff
@@ -0,0 +1,116 @@
...
+)
+
+
+class TestOTXSegTaskInference:
+    @pytest.fixture(autouse=True)
+    def setup(self) -> None:
+        model_template = parse_model_template(os.path.join(DEFAULT_SEG_TEMPLATE_DIR, "template.yaml"))
+        hyper_parameters = create(model_template.hyper_parameters.data)
+        task_env = init_environment(hyper_parameters, model_template)
+        self.seg_train_task = SegmentationInferenceTask(task_env)
```
eunwoosh: I think this code makes something in `/tmp` directory and not deletes it because you don't set `output_path` when initializing `SegmentationInferenceTask`. How about making temporary directory using [tempfile](https://docs.python.org/3/library/tempfile.html) and set it as `output_path` argument to delete it automatically after test?

tests/unit/algorithms/segmentation/test_helpers.py
```diff
@@ -0,0 +1,114 @@
...
+from otx.api.entities.shapes.rectangle import Rectangle
+from otx.api.entities.task_environment import TaskEnvironment
+from tests.test_helpers import generate_random_annotated_image
+
+DEFAULT_SEG_TEMPLATE_DIR = os.path.join("otx/algorithms/segmentation/configs", "ocr_lite_hrnet_18_mod2")
+
+labels_names = ("rectangle", "ellipse", "triangle")
+
+
+def create_model():
```
eunwoosh: I think this function can be modified as fixture because it doesn't get argument. I think that difference between use case for fixture and utility function for test code is that first one provides fixed result and second one can provide various results depending on argument. But some of functions in this file including not only this but also some of functions below provides fixed result and this can be used as fixture, which result in making them easier to use because not need to import and call them but just set them as argument. How do you think about?

supersoob: I revised based on your review. Could you check again?

tests/unit/algorithms/segmentation/tasks/test_segmentation_inference.py
supersoob: I revised based on your review. Could you check again?

eunwoosh (approved):
LGTM. Thanks for applying my comment!

supersoob:
all tests are passed except for existing one in inst-seg.
image

jaegukhyun (approved):
LGTM, It would be much better if there are docstrings for explaining test procedure and which results are checked

PR #1619, opened 2023-02-03
[DEV] Videoframe caching

JihwanEom:
## This PR includes:
- Apply videoframe caching for cls/det/seg tasks.
- `disable=consider-using-with` was inevitable by dynamically changed name of cache directory in `__call__`
- `disable=broad-except` is suggested to handle all the possible exceptions.


otx/algorithms/common/utils/data.py
```diff
@@ -135,3 +141,60 @@ def get_old_new_img_indices(labels, new_classes, dataset):
...
+            if to_float32:
+                cached_img = cached_img.astype(np.float32)
+            return cached_img
+        except Exception as e:  # pylint: disable=broad-except
+            logger.warning(f"Skip loading cached {filename} \nError msg: {e}")
+            return None
+
+    def save_image_to_cache(img: np.array, filename: str) -> None:
+        tmp_filename = filename.replace(".png", "-tmp.png")
+        if os.path.exists(filename) or os.path.exists(tmp_filename):  # if image is cached or caching
```
eunwoosh, line 173: I ran below code and get error as below image.
```
def write_file():
    print("write file start!")
    if os.path.exists("temp_test.txt"):
        return
    with open("temp_test.txt", "w") as f:
        f.write("fake")
    os.replace("temp_test.txt", "test.txt")


if __name__ == '__main__':
    write_arr = [Process(target=write_file) for _ in range(4)]

    for val in write_arr:
        val.start()

    time.sleep(10)

    for val in write_arr:
        val.terminate()
    for val in write_arr:
        val.join()

```
error

I think there is a chance to save caching file in multiple processes at the same time even if it should be rare.
So, In my opinion, this is not a perfect solution and need to find other ways.

JihwanEom: Thank you for suggestion, but this issue can be resolved by `if os.path.exists("temp_test.txt") or os.path.exists("test.txt"):`, similar with https://github.com/openvinotoolkit/training_extensions/pull/1619/files#diff-0c3a869f418bb133848a27abb82c04919ec84d13d29fb7f18d3b6f4eac2dd8e4R173.
Could you double check?

JihwanEom: Sorry, I could find same issue with 5000+ trials. I'll take a look.

JihwanEom: Hmm, handled by adding `if os.path.exists(tmp_name) and not os.path.exists(tgt_name)` before `os.replace`.

```
import tempfile
import os
import time
from multiprocessing import Process

_CACHE_DIR = tempfile.TemporaryDirectory(prefix="test-cache-")

def write_file():
    tmp_name = f"{_CACHE_DIR.name}/temp_test.txt"
    tgt_name = f"{_CACHE_DIR.name}/test.txt"
    if os.path.exists(tmp_name) or os.path.exists(tgt_name):
        return
    with open(tmp_name, "w") as f:
        f.write("fake")
    if os.path.exists(tmp_name) and not os.path.exists(tgt_name):
        os.replace(tmp_name, tgt_name)


if __name__ == '__main__':
    write_arr = [Process(target=write_file) for _ in range(4)]

    for val in write_arr:
        val.start()

    time.sleep(0.01)

    for val in write_arr:
        val.terminate()
    for val in write_arr:
        val.join()
```

eunwoosh (dismissed):
LGTM

harimkang (changes requested):
There seems to be a problem with the test (CI). After resolving this, please request a review again.

JihwanEom:
Passed all the local CI tests. Can reproduce with `tox -re pre-merge`

image

goodsong81 (dismissed):
Great!

JihwanEom:
> Great!

Sorry for requesting review again :( my mistake..

harimkang:
> > Great!
> 
> Sorry for requesting review again :( my mistake..

Sorry, but this PR still doesn't seem to work in CI. Please confirm.

sungmanc (approved):
@harimkang , now it successfully works. @JihwanEom could you rebase it with latest develop?
image

eunwoosh (approved):

harimkang (approved):

PR #1631, opened 2023-02-06
[DEV] Remove unused hpo code

eunwoosh:
### Summray

- Remove code to make hpo dataset which is unused.

### Detail
Making hpo dataset is done in "hpo.py" now. So, I removed unused code.
I checked normal training and hpo works well in all three tasks.


supersoob (approved):
LGTM

jaegukhyun (approved):

PR #1635, opened 2023-02-07
Remove xfail for 83124

AlexanderDokuchaev:
It looks like bug it was fixed a long time ago.

anomaly_cl
anomaly_det
anomaly_seg


goodsong81 (approved):
Thank you for the fix.

sungmanc (approved):

PR #1617, opened 2023-02-02
[OTX] Fixes to h-cls

sovrasov:
- [x] Fix ignore in case of multilabel cls
- [x] Copy hierarchical cls fix from OTE

PR #1637, opened 2023-02-07
[OTX] hotfix device mismatch

cih9088:


sungmanc (dismissed):

eunwoosh (dismissed):

goodsong81 (commented):
Could this similar block in multi-label cls be problematic, too?
image

cih9088:
> Could this similar block in multi-label cls be problematic, too?

Fixed all device mismatch errors caused by https://github.com/openvinotoolkit/training_extensions/pull/1617

goodsong81 (approved):
Thank you for the super quick fix!

PR #1639, opened 2023-02-07
[OTX][Hot-fix] Fix conflicts of same module names in test by adding init file

supersoob:
When collecting pytests test files, it searches upwards until it can find the last folder which still contains an __init__.py file in order to find the package root. The conflict happened between tests/unit/algorithms/action/test_helpers.py and unit/algorithms/segmentation/test_helpers.py. It mainly can resolve by using different name or adding __init__.py for each folder. I resolved it by adding __init__.py for each.


goodsong81 (approved):
Thank you for the quick fix!

sungmanc (approved):

PR #1641, opened 2023-02-07
[OTX] OTX CLI draft refactoring & training with auto-configuration

harimkang:
## Summary
Clarify the role of the ConfigManager and simplify the CLI code. (Refactoring Overall OTX CLI)
- :o: **Compatible with all existing commands**. (otx commands with template)
- :o: Depending on the condition, it additionally supports commands without a template. (**Auto-configuration**)
- :o: Now **otx creates a workspace as a result of every `otx train` command**. (Omit if already)

## How to use
```
# Auto-Configuration with dataset (in the workspace or not)
(otx) otx train --train-data-roots {data-path}
# And we can use all existing commands as well.
(otx) otx train {template} [--args] [params]
```

### Use-cases
image

### Sequence diagram of refactored `otx train`
image

## :heavy_exclamation_mark: Future Works
1. We are going to work on **splitting the `ConfigManager` into different classes**. (because OTX CLI has become simple, but the role of `ConfigManager` has become too large.)
2. We are going to discuss and refactor the task output. (Includes model weights, logs, configuration, and TensorBoard logs, etc.)
3. Add feature for Template replacement support within the workspace (Currently, otx does not support template intended replacement after creating a workspace.)
4. Add Test Cases


otx/cli/manager/config_manager.py
```diff
@@ -0,0 +1,405 @@
...
+                splitted_dataset = self.dataset_manager.auto_split(
+                    task=task,
+                    dataset=train_dataset,
+                    split_ratio=[("train", 0.8), ("val", 0.2)],
+                )
+            else:
+                print(f"[*] Found validation data in your dataset in {data_roots}. It'll be used as validation data.")
+                splitted_dataset = {"train": train_dataset, "val": val_dataset}
+        else:
+            print(f"[*] Current auto-split can't support the {self.data_format} format.")
```
eunwoosh, line 199: I think raising error at this point it better even if error is raised later. How do you think about?

otx/cli/manager/config_manager.py
```diff
@@ -0,0 +1,405 @@
...
+            task_type (str): The task_type registered in the registry. Used for filtering.
+            model (str, optional): The task_type registered in the registry. Used for filtering. Defaults to None.
+
+        Returns:
+            ModelTemplate: Selected model template.
+        """
+        otx_registry = OTXRegistry(self.otx_root).filter(task_type=task_type)
+        if model:
+            template_lst = [temp for temp in otx_registry.templates if temp.name.lower() == model.lower()]
+            if len(template_lst) == 0:
```
eunwoosh:
```suggestion
            if not template_lst:
```

It's recommended way to check empty list in [PEP 8 style guide](https://www.python.org/dev/peps/pep-0008).

otx/cli/manager/config_manager.py
```diff
@@ -0,0 +1,405 @@
...
+                raise ValueError(
+                    f"[*] {model} is not a type supported by OTX {task_type}."
+                    f"\n[*] Please refer to 'otx find --template --task_type {task_type}'"
+                )
+            template = template_lst[0]
+        else:
+            template = otx_registry.get(DEFAULT_MODEL_TEMPLATE_ID[task_type.upper()])
+        return template
+
+    def build_workspace(self, new_workspace_path: str = None) -> None:
```
eunwoosh:
```suggestion
    def build_workspace(self, new_workspace_path: Optional[Union[str, Path]] = None) -> None:
```

otx/cli/manager/config_manager.py
```diff
@@ -0,0 +1,405 @@
...
+        """Create OTX workspace with Template configs from task type.
+
+        This function provides a user-friendly OTX workspace and provides more intuitive
+        and create customizable templates to help users use all the features of OTX.
+        task_type: The type of task want to get (str)
```
eunwoosh: I think the comments to explain argument aren't aligned with current function.

otx/cli/utils/parser.py
```diff
@@ -78,7 +84,7 @@ def _get_default_metavar_for_optional(self, action):
         return action.dest.split(".")[-1].upper()
 
 
-def add_hyper_parameters_sub_parser(parser, config, modes=None):
+def add_hyper_parameters_sub_parser(parser, config, modes=None, return_sub_parser=False):
```
eunwoosh, line 90: How about just making this function always return and remove `return_sub_parser` argument?

harimkang:
image

otx/cli/utils/tests.py
```diff
@@ -742,11 +762,27 @@ def otx_build_auto_config(root, otx_dir: str, args: Dict[str, str]):
...
+    for option, value in args.items():
+        if option == "template":
+            command_line.extend([args[option]])
+        elif option in ["--train-data-roots", "--val-data-roots"]:
+            command_line.extend([option, f"{os.path.join(otx_dir, value)}"])
+    command_line.extend(["--save-model-to", f"{work_dir}"])
+    command_line.extend(["--work-dir", f"{work_dir}"])
+    command_line.extend(args["train_params"])
+    check_run(command_line)
+    shutil.rmtree(work_dir)
```
eunwoosh: How about using TemporaryDirectory in [tempfile](https://docs.python.org/3/library/tempfile.html) instead deleting manually?

otx/cli/tools/export.py
```diff
@@ -90,7 +95,9 @@ def main():
 
     task.export(ExportType.OPENVINO, exported_model)
 
-    os.makedirs(args.save_model_to, exist_ok=True)
+    if "save_model_to" not in args or not args.save_model_to:
+        args.save_model_to = str(config_manager.workspace_root / "model-exported")
+    Path(args.save_model_to).mkdir(exist_ok=True)
```
eunwoosh:
```suggestion
    Path(args.save_model_to).mkdir(exist_ok=True, parents=True)
```
Isn't it needed?

otx/cli/tools/build.py
```diff
@@ -37,11 +35,35 @@ def set_workspace(task, model):
...
 
-def parse_args():
+def get_args():
     """Parses command line arguments."""
+    # TODO: Declaring pre_parser to get the template
+    pre_parser = argparse.ArgumentParser(add_help=False)
+    pre_parser.add_argument("template", nargs="?", default=None)
+    parsed, _ = pre_parser.parse_known_args()
+    template = parsed.template
     parser = argparse.ArgumentParser()
```
eunwoosh: I think code L41 ~ L49 is duplicated across many files in otx/cli/tools. How about making these part as a function (manybe in the otx/utils/parser,py?) and using it?

otx/cli/tools/deploy.py
```diff
@@ -78,9 +91,11 @@ def main():
 
     deployed_model = ModelEntity(None, environment.get_model_configuration())
 
+    if "save_model_to" not in args or not args.save_model_to:
+        args.save_model_to = str(config_manager.workspace_root / "model-deployed")
     os.makedirs(args.save_model_to, exist_ok=True)
     task.deploy(deployed_model)
-    with open(os.path.join(args.save_model_to, "openvino.zip"), "wb") as write_file:
+    with open(str(Path(args.save_model_to) / "openvino.zip"), "wb") as write_file:
```
eunwoosh:
```suggestion
    with open(Path(args.save_model_to) / "openvino.zip", "wb") as write_file:
```
open can get Path.

otx/cli/utils/parser.py
harimkang: That would be nice too, but this is currently only used in train.py.

otx/cli/utils/tests.py
harimkang: As far as I know, this uses a TemporaryDirectory. 
@sungmanc Any other reason for this erasing code?

otx/cli/manager/config_manager.py
harimkang: This will be addressed after further discussion. In the current PR, I decided to leave it as it is for now. Most of the case the training doesn't work, or it might work with selfsl.

otx/cli/manager/config_manager.py
harimkang: Here new_workspace_path is intended as str. It is intended to be passed directly from args. I'll leave it like this.

otx/cli/utils/parser.py
eunwoosh: Yes, you're right. But others which use this function just don't use return value then. Actually, this is a one of design choices. So if you think current implementation is better than just let it be and resolve this conversation.

otx/cli/manager/config_manager.py
eunwoosh:
```suggestion
    def build_workspace(self, new_workspace_path: Optional[str] = None) -> None:
```
Then, how about this?

otx/cli/utils/parser.py
harimkang: Yes your suggestion is good also but doing so requires a lot of changes, There doesn't seem to be a need to put in that much effort right now.

otx/cli/manager/config_manager.py
harimkang: reflected this. https://github.com/openvinotoolkit/training_extensions/pull/1641/commits/3fcf067f3ae5f70826b0df66bdb702db0a04c358

otx/cli/manager/config_manager.py
harimkang: Removed this. https://github.com/openvinotoolkit/training_extensions/pull/1641/commits/3fcf067f3ae5f70826b0df66bdb702db0a04c358

otx/cli/utils/tests.py
harimkang: Removed this. https://github.com/openvinotoolkit/training_extensions/pull/1641/commits/3fcf067f3ae5f70826b0df66bdb702db0a04c358

otx/cli/tools/export.py
harimkang: reflected this. https://github.com/openvinotoolkit/training_extensions/pull/1641/commits/3fcf067f3ae5f70826b0df66bdb702db0a04c358

otx/cli/tools/build.py
harimkang: reflected this. -> parser.py https://github.com/openvinotoolkit/training_extensions/pull/1641/commits/3fcf067f3ae5f70826b0df66bdb702db0a04c358

otx/cli/tools/deploy.py
harimkang: reflected this. https://github.com/openvinotoolkit/training_extensions/pull/1641/commits/3fcf067f3ae5f70826b0df66bdb702db0a04c358

otx/cli/manager/config_manager.py
harimkang: reflected this. https://github.com/openvinotoolkit/training_extensions/pull/1641/commits/56c882a32f8d2476219990539bd17ef17903629e

harimkang:
@eunwoosh Thanks for the detailed review. Most of your suggestions have been reflected.

otx/cli/tools/export.py
sungmanc: It allows the user can input the more complex path (e.g.`results/checkpoints`), not mandatory.

otx/cli/utils/tests.py
sungmanc: I couldn't remember the reason, maybe I added this line when we don't allow the duplicate build. So, it can be removed.

otx/cli/manager/config_manager.py
eunwoosh: Thanks. But comment to explain `new_workspace_path` argument is also removed. Could you add it?

harimkang: Originally there was no explanation for it.. :)

eunwoosh: I see. But I think it's better to add to make it easier to understand.

harimkang: Added!

otx/cli/tools/build.py
```diff
@@ -37,11 +35,26 @@ def set_workspace(task, model):
     return path
 
 
-def parse_args():
+def get_args():
```
sungmanc, line 38: How about add `--test-data-roots` also

sungmanc (dismissed):
Thanks for the nice work !

otx/cli/tools/build.py
harimkang: This is something I was missing. Added!

eunwoosh (dismissed):
Thanks for your work! This can be first step to organize current complex cli implementation.

goodsong81 (dismissed):
What an enhancement!
Looking forward to see rebuild, retrain in the near future :)

harimkang:
@sungmanc @goodsong81 @eunwoosh Sorry for the inconvenience, some test dataset paths were changed recently, causing 2 FAIL cases in CI, but this commit https://github.com/openvinotoolkit/training_extensions/pull/1641/commits/875db1d277eaff855d746247057edfbac52b27e9 fixed it.
image

sungmanc (dismissed):

sungmanc (approved):

goodsong81 (approved):

samet-akcay:
Late to the party, but thanks for this PR. Looks promising!

PR #1644, opened 2023-02-08
Remove unused files in mpa/modules

supersoob:
This PR is for removing the unused files in mpa/modules. I tested and checked cli tests over all tasks. I will make the following PR for removing unused files in mpa/recipe after this. If there are files in use, please comment.


jaegukhyun (approved):
Looks good, let's see the integration test results

JihwanEom:
Could you run e2e test locally? It will be better to check broad coverage.

supersoob:
I ran the local e2e tests, and failed issues are anomalib import issue caused by torch update and fq issues which is known issues. I think it's okay to merge. How do you think? @JihwanEom 
image
image
image
image
image
image

sungmanc (approved):

PR #1638, opened 2023-02-07
[DEV] Unit tests for classification

sovrasov:

PR #1646, opened 2023-02-08
[DEV][HOTFIX] Add accuracy metric in multi label classification

eunwoosh:
### Summary

- Add `accuracy` metric in multi label classification

### Detail
Currently, there is no `accuracy` metric in multi label classification, which results in error during HPO.

### DIFF
--- a/otx/algorithms/classification/adapters/mmcls/data/datasets.py
+++ b/otx/algorithms/classification/adapters/mmcls/data/datasets.py
@@ -281,6 +281,7 @@ class MPAMultilabelClsDataset(MPAClsDataset):
                 if k in metrics:
                     eval_results[k] = v
 
+        eval_results["accuracy"] = eval_results["accuracy-mlc"]
         return eval_results
 
 


supersoob (approved):
Thanks for quick fix.

JihwanEom (approved):

otx/algorithms/classification/adapters/mmcls/data/datasets.py
```diff
@@ -281,6 +281,7 @@ def evaluate(
                 if k in metrics:
                     eval_results[k] = v
 
+        eval_results["accuracy"] = eval_results["accuracy-mlc"]
```
sovrasov, line 284: Hi @eunwoosh. Earlier in this file we have a [line ](https://github.com/openvinotoolkit/training_extensions/pull/1646/files#diff-24b4265f1eaea95b802d58f85dc3dd3d9b233776a1dc53b19d842865ba927593L247)
```
if "accuracy-mlc" in metrics:
```
That means `accuracy-mlc` is conditionally included into `eval_results` and sometimes this line could cause a failure.
I'll fix that in [#1638](https://github.com/openvinotoolkit/training_extensions/pull/1638)

eunwoosh: I forget about that. Thanks for pointing it.

PR #1647, opened 2023-02-08
[DEV] Edit the dummy datasets to toy datasets (CVS-103301)

sungmanc:
# Summary
* This PR is a remedy for [CVS-103301](https://jira.devtools.intel.com/browse/CVS-103301)
* Convert the detection dummy datasets to toy datasets
* Convert the segmentation dummy datasets to more easy ones.
* Change the test files
* Had no FAIL cases in instance_segmentation tests even `nncf_optimize` was included. (tests/integration/cli/detection)
image
image

* Got all PASSED in unit tests (tests/unit)


JihwanEom:
Does this PR not include dataset for semantic segmentation?

JihwanEom:
How about the accuracy? Is it not always 0.0?

sungmanc:
> How about the accuracy? Is it not always 0.0?

In the current stats, some models showed 0.0 F-measure. However, accuracy is not important for NNCF. If there is a single prediction, it is okay regardless of whether it is positive or not

JihwanEom:
> > How about the accuracy? Is it not always 0.0?
> 
> In the current stats, some models showed 0.0 F-measure. However, accuracy is not important for NNCF. If there is a single prediction, it is okay regardless of whether it is positive or not

Sounds good! Thank you for kind explanation.

sungmanc:
image

goodsong81 (approved):
Nice fix!

harimkang (approved):

PR #1648, opened 2023-02-08
[FEATURE] Add support for direct JSON ann files

eugene123tw:
* `otx train/eval --data-root /DATASET/coco`
* `otx train --train_ann_file /DATASET/coco/annotations/instances_train.json --val_ann_file /DATASET/coco/annotations/instances_val.json`
* `otx train/eval data.yaml`

data.yaml file definition:
```yaml
'data-root': '/DATASET/coco'
'train': '/DATASET/coco/annotations/instances_train_shorten_to_10.json'
'val': '/DATASET/coco/annotations/instances_val_shorten_to_10.json'
'test': '/DATASET/coco/annotations/instances_test_shorten_to_10.json'
```
* `otx eval --test-ann-file /DATASET/coco/annotations/instances_test_shorten_to_10.json`

### DIFF
--- a/otx/cli/tools/train.py
+++ b/otx/cli/tools/train.py
@@ -69,14 +69,19 @@ def parse_args():
     required = not os.path.exists(parsed.data)
 
     parser.add_argument(
-        "--train-data-roots",
+        "--data-root",
         required=required,
-        help="Comma-separated paths to training data folders.",
+        help="Comma-separated paths to data root folder.",
     )
     parser.add_argument(
-        "--val-data-roots",
+        "--train-ann-file",
         required=False,
-        help="Comma-separated paths to validation data folders.",
+        help="Comma-separated paths to training JSON.",
+    )
+    parser.add_argument(
+        "--val-ann-file",
+        required=False,
+        help="Comma-separated paths to validation JSON.",
     )
     parser.add_argument(
         "--unlabeled-data-roots",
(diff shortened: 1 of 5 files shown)


otx/cli/tools/train.py
```diff
@@ -69,14 +69,19 @@ def parse_args():
     required = not os.path.exists(parsed.data)
 
     parser.add_argument(
-        "--train-data-roots",
+        "--data-root",
```
harimkang, line 72: With this structure (data-root + subset's ann-file), the folder structure format of multi-class classification is not considered. What do you think?

eugene123tw: hmm, no problem. I'll check if I can support folder structure in multi-class classification

otx/core/data/adapter/base_dataset_adapter.py
```diff
@@ -98,40 +102,38 @@ def _import_dataset(
...
 
+        if train_ann_file:
             # Prepare subsets by using Datumaro dataset
-            for k, v in datumaro_dataset.subsets().items():
-                if "train" in k or "default" in k:
-                    dataset[Subset.TRAINING] = v
-                elif "val" in k:
-                    dataset[Subset.VALIDATION] = v
-            self.is_train_phase = True
+            dataset[Subset.TRAINING] = DatumaroDataset.import_from(
```
sungmanc, line 122: This is the hardest part when we used the Datumaro as an data frontend. Datumaro can't import annotation file directly. 
So, this line will not work.

sungmanc: So, if we want to use annotation file directly, we need to change the Datumaro's function or use previous CLI datasets.

sungmanc (commented):
Thanks a lot to propose your thoughts, we might need a lot of time to discuss this topic.

otx/core/data/adapter/base_dataset_adapter.py
harimkang: In my opinion, adding that feature to datumaro seems most appropriate. But before that, is there any other way to apply (I don't know if the previous method is supported..)?

eugene123tw: Screenshot from 2023-02-09 08-11-53
@sungmanc it works for me without changing anything :thinking:

sungmanc: > Screenshot from 2023-02-09 08-11-53 @sungmanc it works for me without changing anything 🤔

Wow, I had never thought about this way. Then, we can add direct annotation easily

eugene123tw: @sungmanc I did some tracing in the Datumaro code and seems like it works, but there might be some edge cases I haven't considered yet. Do you know what needs to be changed for a multi-classification folder structure?

sungmanc: > @sungmanc I did some tracing in the Datumaro code and seems like it works, but there might be some edge cases I haven't considered yet. Do you know what needs to be changed for a multi-classification folder structure?

I'll look into deeper after the OTXv1.0 ! For now, I don't have idea.

harimkang:
@sungmanc This PR is now an included feature, right?

sungmanc:
Applied for COCO format.

PR #1662, opened 2023-02-09
[FIX] Restore compatibility with latest Anomalib

djdameln:
This PR restores the compatibility of the anomaly tasks with the latest Anomalib release (v0.4.0rc1), and also prepares the task implementation for Anomalib's new bbox detection models.

### DIFF
--- a/otx/algorithms/anomaly/adapters/anomalib/callbacks/inference.py
+++ b/otx/algorithms/anomaly/adapters/anomalib/callbacks/inference.py
@@ -138,7 +138,7 @@ class AnomalyInferenceCallback(Callback):
                 probability = score.item()
                 annotations.append(Annotation(shape=shape, labels=[ScoredLabel(label=label, probability=probability)]))
             # get label
-            label = self.normal_label if len(annotations) == 0 else self.anomalous_label
+            label = self.anomalous_label if annotations else self.normal_label
             probability = pred_score if label.is_anomalous else 1 - pred_score
             # update dataset item
             dataset_item.append_annotations(annotations)
--- a/otx/algorithms/anomaly/adapters/anomalib/callbacks/inference.py
+++ b/otx/algorithms/anomaly/adapters/anomalib/callbacks/inference.py
@@ -47,7 +47,7 @@ class AnomalyInferenceCallback(Callback):
         self.task_type = task_type
         self.label_map = {0: self.normal_label, 1: self.anomalous_label}
 
-    def on_predict_epoch_end(self, _trainer: pl.Trainer, pl_module: AnomalyModule, outputs: List[Any]):
+    def on_predict_epoch_end(self, _trainer: pl.Trainer, _pl_module: AnomalyModule, outputs: List[Any]):
         """Call when the predict epoch ends."""
         # TODO; refactor Ignore too many locals
         # pylint: disable=too-many-locals
--- a/otx/algorithms/anomaly/adapters/anomalib/callbacks/inference.py
+++ b/otx/algorithms/anomaly/adapters/anomalib/callbacks/inference.py
@@ -18,17 +18,20 @@ from typing import Any, List
 
 import numpy as np
 import pytorch_lightning as pl
+import torch
 from anomalib.models import AnomalyModule
 from anomalib.post_processing import anomaly_map_to_color_map
 from pytorch_lightning.callbacks import Callback
+from torch import Tensor
 
 from otx.algorithms.anomaly.adapters.anomalib.logger import get_logger
+from otx.api.entities.annotation import Annotation
 from otx.api.entities.datasets import DatasetEntity
 from otx.api.entities.label import LabelEntity
 from otx.api.entities.model_template import TaskType
 from otx.api.entities.result_media import ResultMediaEntity
 from otx.api.entities.scored_label import ScoredLabel
-from otx.api.utils.anomaly_utils import create_detection_annotation_from_anomaly_heatmap
+from otx.api.entities.shapes.rectangle import Rectangle
 from otx.api.utils.segmentation_utils import create_annotation_from_segmentation_map
 
 logger = get_logger(__name__)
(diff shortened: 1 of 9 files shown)


otx/algorithms/anomaly/adapters/anomalib/data/data.py
```diff
@@ -93,23 +94,44 @@ def __getitem__(self, index: int) -> Dict[str, Union[int, Tensor]]:
...
         item: Dict[str, Union[int, Tensor]] = {}
         item = {"index": index}
-        if self.task_type in (TaskType.ANOMALY_CLASSIFICATION, TaskType.ANOMALY_DETECTION):
+        if self.task_type == TaskType.ANOMALY_CLASSIFICATION:
             # Detection currently relies on image labels only, meaning it'll use image
             #   threshold to find the predicted bounding boxes.
-            item["image"] = self.pre_processor(image=dataset_item.numpy)["image"]
+            item["image"] = self.transform(image=dataset_item.numpy)["image"]
+        elif self.task_type == TaskType.ANOMALY_DETECTION:
+            item["image"] = self.transform(image=dataset_item.numpy)["image"]
```
samet-akcay, line 102: Would box transforms not needed here?

otx/algorithms/anomaly/tasks/inference.py
```diff
@@ -192,13 +194,17 @@ def infer(self, dataset: DatasetEntity, inference_parameters: InferenceParameter
...
+            pixel_metrics=config.metrics.get("pixel"),
         )
-        callbacks = [progress, normalize, inference, metrics_configuration]
+        post_processing_configuration = PostProcessingConfigurationCallback(
+            normalization_method=NormalizationMethod.MIN_MAX,
+            threshold_method=ThresholdMethod.ADAPTIVE,
+            manual_image_threshold=config.metrics.threshold.manual_image,
+            manual_pixel_threshold=config.metrics.threshold.manual_pixel,
+        )
+        callbacks = [progress, normalize, inference, metrics_configuration, post_processing_configuration]
```
samet-akcay, line 207: Does order of the callbacks matter here? If yes, are we certain that it is in order?

otx/algorithms/anomaly/adapters/anomalib/callbacks/inference.py
```diff
@@ -49,49 +52,118 @@ def on_predict_epoch_end(self, _trainer: pl.Trainer, pl_module: AnomalyModule, o
...
+                    x1=box[0].item() / width,
+                    y1=box[1].item() / height,
+                    x2=box[2].item() / width,
+                    y2=box[3].item() / height,
+                )
+                label = self.label_map[label.item()]
+                probability = score.item()
+                annotations.append(Annotation(shape=shape, labels=[ScoredLabel(label=label, probability=probability)]))
+            # get label
+            label = self.normal_label if len(annotations) == 0 else self.anomalous_label
```
samet-akcay: A suggestion by `refurb`:
```suggestion
            label = self.anomalous_label if annotation else self.normal_label
```

otx/algorithms/anomaly/adapters/anomalib/data/data.py
```diff
@@ -93,23 +94,44 @@ def __getitem__(self, index: int) -> Dict[str, Union[int, Tensor]]:
...
+                        Tensor(
+                            [
+                                annotation.shape.x1 * width,
+                                annotation.shape.y1 * height,
+                                annotation.shape.x2 * width,
+                                annotation.shape.y2 * height,
+                            ]
+                        )
+                    )
+                if len(boxes) > 0:
```
samet-akcay: A suggestion by `refurb`
```suggestion
                if boxes:
```

otx/algorithms/anomaly/adapters/anomalib/callbacks/inference.py
```diff
@@ -44,54 +47,123 @@ def __init__(self, otx_dataset: DatasetEntity, labels: List[LabelEntity], task_t
...
+            self._process_classification_predictions(pred_labels, pred_scores)
+        elif self.task_type == TaskType.ANOMALY_DETECTION:
+            # collect detection predictions
+            pred_boxes = []
+            box_scores = []
+            box_labels = []
+            for output in outputs:
+                pred_boxes.extend(output["pred_boxes"])
+                box_scores.extend(output["box_scores"])
+                box_labels.extend(output["box_labels"])
```
samet-akcay, line 72: image
Maybe we could use list comprehension here, which is shorter and more efficient.

samet-akcay:
```suggestion
            pred_boxes = [output["pred_boxes"] for output in outputs]
            box_scores = [output["box_scores"] for output in outputs]
            box_labels = [output["box_labels"] for output in outputs]
```

samet-akcay (changes requested):
Thanks for your effort! I only have few minor comments/questions

otx/algorithms/anomaly/adapters/anomalib/data/data.py
djdameln: Box coordinates in OTX are normalized wrt image shape. The absolute values are computed in lines 111 to 114 based on the transformed image shape, so no need to include the boxes in the transform.

otx/algorithms/anomaly/adapters/anomalib/callbacks/inference.py
djdameln: This won't work because we have to flatten the list of lists. We could do the following:
```suggestion
            pred_boxes = [box for output in outputs for box in output["pred_boxes"]]
            box_scores = [score for output in outputs for score in output["box_scores"]]
            box_labels = [label for output in outputs for label in output["box_labels"]]
```
but I feel it would make the code less readable

otx/algorithms/anomaly/tasks/inference.py
djdameln: The order doesn't matter here. The only thing that matters is that normalization is applied before the inference callback returns the predictions, but since the normalization callback is active at `on_predict_batch_end`, and the inference callback at `on_predict_epoch_end`, this does not depend on the execution order of the callbacks.

The configuration callbacks are only active during setup, and the order in which metrics and post-processing is set up does not matter.

otx/algorithms/anomaly/adapters/anomalib/callbacks/inference.py
samet-akcay: oh I see. Yeah, then ignore this comment. I'll resolve it

otx/algorithms/anomaly/tasks/inference.py
samet-akcay: thanks. Similar to anomalib, a custom loop will be a quite nice solution here

samet-akcay (dismissed):
lgtm, thanks a lot for the effort

goodsong81:
Not changing anomalib version itself? If you would deal with the upgrade in other PR, please make sure that re-enable integration/e2e tests for anomaly task that had been disabled due to torch1.13.1 incompatibility. (Search for ` --ignore-glob=**/anomaly/*` in tox.ini

goodsong81 (approved):
Thanks for the quick fix!

harimkang (approved):
LGTM!

PR #1665, opened 2023-02-09
[TEST][ANOMALY] Add unit tests

ashwinvaidya17:
# Description

Add anomaly unit tests

# TODO

- [x] nncf task
- [x] openvino task
- [ ] exportable code


ashwinvaidya17:
We need to include changes from https://github.com/openvinotoolkit/training_extensions/pull/1662 so that it works in the CI

tests/unit/algorithms/anomaly/helpers/dummy_dataset.py
```diff
@@ -0,0 +1,172 @@
...
+from otx.api.entities.shapes.rectangle import Rectangle
+from otx.api.entities.subset import Subset
+
+
+def get_shapes_dataset(task_type: TaskType, one_each: bool = False) -> DatasetEntity:
+    """Get shapes dataset.
+
+    Args:
+        task_type (TaskType): Task type.
+        one_each (bool): Gets one each of each class. Defaults to False.
```
samet-akcay: Can you elaborate this? It took me a while to actually get what exactly it does.

tests/unit/algorithms/anomaly/helpers/dummy_model.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def configure_optimizers(self):
+        return None
+
+    def predict_step(self, batch: Any, batch_idx: int, dataloader_idx: int = 0) -> Any:
+        # Just return everything as anomalous
+        batch["anomaly_maps"] = batch["pred_masks"] = torch.ones(batch["image"].shape[0], 1, *batch["image"].shape[2:])
+        batch["pred_labels"] = batch["pred_scores"] = torch.ones((batch["image"].shape[0]))
+
+        return batch
```
samet-akcay:
```suggestion
    def configure_optimizers(self):
        return None

    def training_step(self, *args, **kwargs):
        pass

    def predict_step(self, batch: Any, batch_idx: int, dataloader_idx: int = 0) -> Any:
        # Just return everything as anomalous
        batch["anomaly_maps"] = batch["pred_masks"] = torch.ones(batch["image"].shape[0], 1, *batch["image"].shape[2:])
        batch["pred_labels"] = batch["pred_scores"] = torch.ones((batch["image"].shape[0]))
        return batch

    def validation_step(self, batch: Any, batch_idx: int, dataloader_idx: int = 0):
        return self.predict_step(batch, batch_idx, dataloader_idx)

    def test_step(self, batch: Any, batch_idx: int, dataloader_idx: int = 0):
        return self.predict_step(batch, batch_idx, dataloader_idx)
```

tests/unit/algorithms/anomaly/tasks/test_inference.py
```diff
@@ -0,0 +1,73 @@
...
+from otx.api.entities.model_template import TaskType
+from otx.api.entities.resultset import ResultSetEntity
+from otx.api.entities.subset import Subset
+from otx.api.usecases.tasks.interfaces.export_interface import ExportType
+from tests.unit.algorithms.anomaly.helpers.dummy_dataset import get_shapes_dataset
+from tests.unit.algorithms.anomaly.helpers.utils import create_task_environment
+
+
+class TestInferenceTask:
+    """Tests the method in the inference task."""
```
samet-akcay: Can you add a description here to explain the steps in details. What would be the scope of this test? This would serve as a future reference

tests/unit/algorithms/anomaly/adapters/callbacks/test_inference_callback.py
```diff
@@ -0,0 +1,26 @@
...
+
+# Copyright (C) 2021-2022 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+import pytorch_lightning as pl
+
+from otx.algorithms.anomaly.adapters.anomalib.callbacks import AnomalyInferenceCallback
+from otx.api.entities.model_template import TaskType
+from tests.unit.algorithms.anomaly.helpers.dummy_dataset import MockDataModule
```
samet-akcay: Just to follow the same naming convention, would it be an idea to name `MockDataModule` to `DummyDataModule`? Is there a specific motivation of naming it as `MockDataModule`?

tests/unit/algorithms/anomaly/adapters/callbacks/test_inference_callback.py
```diff
@@ -0,0 +1,26 @@
...
+from tests.unit.algorithms.anomaly.helpers.dummy_model import DummyModel
+
+
+class TestInferenceCallback:
+    @pytest.mark.parametrize(
+        "task_type", [TaskType.ANOMALY_CLASSIFICATION, TaskType.ANOMALY_DETECTION, TaskType.ANOMALY_SEGMENTATION]
+    )
+    def test_inference_callback(self, task_type):
```
djdameln, line 19: Can you add a docstring explaining what exactly is tested here?

djdameln, line 19: As discussed, we might want to add more tests to cover various use cases and edge cases (possibly in a separate PR). Some suggestion:
- all three task types (cls, det, seg)
- Are heatmaps stored correctly as metadata items?

tests/unit/algorithms/anomaly/adapters/data/test_dataset.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+@pytest.mark.parametrize(
+    "task_type", [TaskType.ANOMALY_CLASSIFICATION, TaskType.ANOMALY_DETECTION, TaskType.ANOMALY_SEGMENTATION]
+)
+def test_dataloaders(task_type):
+    """Tests whether the dataloaders can load the data correctly."""
+    datamodule = ShapesDataModule(task_type)
+    if task_type == TaskType.ANOMALY_CLASSIFICATION:
+        assert next(iter(datamodule.predict_dataloader())).keys() == {"image", "label", "index"}
```
djdameln: Should we also test the train, val and test dataloaders?

tests/unit/algorithms/anomaly/config/test_model_config_load.py
```diff
@@ -0,0 +1,64 @@
...
+        ("padim", PadimAnomalySegmentationConfig),
+        ("stfpm", STFPMAnomalyClassificationConfig),
+        ("stfpm", STFPMAnomalyDetectionConfig),
+        ("stfpm", STFPMAnomalySegmentationConfig),
+        ("draem", DraemAnomalyClassificationConfig),
+        ("draem", DraemAnomalyDetectionConfig),
+        ("draem", DraemAnomalySegmentationConfig),
+    ],
+)
+def test_model_template_loading(model_name, configurable_parameters):
```
djdameln, line 48: Could you add a docstring here?

tests/unit/algorithms/anomaly/config/test_model_config_load.py
```diff
@@ -0,0 +1,64 @@
...
+    configurable_parameters_yaml: ConfigurableParameters = create(configurable_parameters_yaml_str)
+
+    model_template = get_model_template(model_name)
+    hyper_parameters: dict = model_template.hyper_parameters.data
+    configurable_parameters_loaded = create(hyper_parameters)
+
+    assert configurable_parameters_yaml == configurable_parameters_loaded
+
+    # Confirm that we can create an anomalib config from the loaded yaml
+    get_anomalib_config(model_name, configurable_parameters_yaml)
```
djdameln: the comment mentions the loaded yaml, but the check targets the created yaml. I guess one of those should be changed.

tests/unit/algorithms/anomaly/helpers/dummy_dataset.py
```diff
@@ -0,0 +1,176 @@
...
+        self.config = OmegaConf.create({"dataset": {"image_size": [32, 32]}})
+
+
+class ShapesDataModule(OTXAnomalyDataModule):
+    def __init__(self, task_type: TaskType):
+        self.dataset = ShapesDataset(get_shapes_dataset(task_type), task_type)
+        self.labels = self.dataset.dataset.get_labels()
+
+    def predict_dataloader(self) -> DataLoader:
+        return DataLoader(self.dataset, shuffle=False, batch_size=32, num_workers=2, collate_fn=collate_fn)
```
djdameln: Not sure if it's a good idea to base our dataset tests on a subclass that overrides some stuff from the parent class. If I'm not wrong, our tests don't actually cover `OTXAnomalyDataModule.predict_dataloader()` now because it is overridden here.

tests/unit/algorithms/anomaly/tasks/test_inference.py
djdameln: ~~method~~ methods

tests/unit/algorithms/anomaly/tasks/test_inference.py
```diff
@@ -0,0 +1,73 @@
...
+        train_task.save_model(output_model)
+
+        new_dataset: DatasetEntity = get_shapes_dataset(task_type, one_each=True)
+        gt_val_dataset = new_dataset.get_subset(Subset.VALIDATION)
+        new_task_environment = create_task_environment(gt_val_dataset, task_type)
+        new_task_environment.model = output_model
+        inference_task = InferenceTask(new_task_environment, output_path=root)
+        pred_val_dataset = inference_task.infer(
+            gt_val_dataset.with_empty_annotations(), InferenceParameters(is_evaluation=True)
+        )
```
djdameln: Could you add some comments here explaining the different steps? This test is a bit hard to follow now

tests/unit/algorithms/anomaly/tasks/test_inference.py
```diff
@@ -0,0 +1,73 @@
...
+from tests.unit.algorithms.anomaly.helpers.utils import create_task_environment
+
+
+class TestInferenceTask:
+    """Tests the method in the inference task."""
+
+    @pytest.mark.parametrize(
+        "task_type", [TaskType.ANOMALY_CLASSIFICATION, TaskType.ANOMALY_DETECTION, TaskType.ANOMALY_SEGMENTATION]
+    )
+    def test_inference(self, task_type, tmpdir):
```
djdameln: This test case is quite long and tests a bunch of different things. Do you think it could be split into several smaller tests?

samet-akcay (commented):
Thank you for your effort. The code looks clean too! This would overall do the job to provide a decent coverage.

I have some comments, but i think we could address these in a follow-up PR for a refactor. 

- It would be great if we utilize pytest fixtures to setup data, model, train and the rest of the workflow. The advantage would be that we would initialize these once and use it throughout the tests. 
- It would be good to have the `assert` statements at the end of each test function. If we have multiple `assert` statements, this might indicate that we could split the test functions.

As I said, I'm providing these comments here for a future reference. Let's address these later.

tests/unit/algorithms/anomaly/tasks/test_train.py
```diff
@@ -0,0 +1,53 @@
...
+        dataset: DatasetEntity = get_shapes_dataset(task_type, one_each=True)
+        task_environment = create_task_environment(dataset, task_type)
+        # get model configuration
+        output_model = ModelEntity(
+            dataset,
+            task_environment.get_model_configuration(),
+        )
+
+        train_task = TrainingTask(task_environment, output_path=root)
+        train_task.train(dataset, output_model, TrainParameters())
```
djdameln: There seems to be some overlap between this test case and those in `test_nncf.py`, `test_openvino.py`. You could maybe consider using fixtures to abstract away some of the boilerplate code.

djdameln (changes requested):

tests/unit/algorithms/anomaly/tasks/test_train.py
```diff
@@ -0,0 +1,47 @@
...
+    def test_train_and_load(self, tmpdir, setup_task_environment):
+        """Tests the train method and check if it can be loaded correctly."""
+        root = str(tmpdir.mkdir("anomaly_training_test"))
+
+        # Get task environment
+        setup_task_environment = deepcopy(setup_task_environment)  # since fixture is mutable
+        task_environment = setup_task_environment.task_environment
+        output_model = setup_task_environment.output_model
+        dataset = setup_task_environment.dataset
+        task_type = setup_task_environment.task_type
```
djdameln, line 36: Not sure if this is an improvement, but maybe we can revisit this later

djdameln (approved):
Thanks. I feel the task tests could be simplified and improved a bit further, but I don't want to block you so let's merge this for now.

tests/unit/algorithms/anomaly/tasks/test_train.py
ashwinvaidya17: Yeah that was my concern as well. Ideally we should combine all the tasks into a single file and run them in stages to avoid redundant computation.

samet-akcay (approved):

PR #1666, opened 2023-02-09
[ENHANCE] Consolidate hpopt into OTX

eunwoosh:
### Summary

- Moved hpopt files at otx/hpo directory.
- Unnecessary files weren't moved.
- Unit test files for hpopt were also moved. (test coverage : 87%)
- Updated requirements/base.txt
- Changed `import hpopt...` to `import otx.hpo...`.
- Added some utility function to otx/algorithms/common/utils/utils.py


otx/hpo/__init__.py
```diff
@@ -0,0 +1,25 @@
+"""HPO package."""
+
+# Copyright (C) 2022 Intel Corporation
```
JihwanEom, line 3:
```suggestion
# Copyright (C) 2023 Intel Corporation
```

otx/hpo/hpo_base.py
```diff
@@ -0,0 +1,305 @@
+"""HPO algorithm abstract class."""
+
+# Copyright (C) 2022 Intel Corporation
```
JihwanEom, line 3:
```suggestion
# Copyright (C) 2023 Intel Corporation
```

otx/hpo/hpo_base.py
```diff
@@ -0,0 +1,305 @@
...
+        subset_ratio (Optional[Union[float, int]]): ratio to how many train dataset to use for each trial.
+                                     The lower value is, the faster the speed is.
+                                     But If it's too low, HPO can be unstable.
+        min_subset_size (int) : Minimum size of subset. Default value is 500.
+        resume (bool): resume flag decide to use previous HPO results.
+                       If HPO completed, you can just use optimized hyper parameters.
+                       If HPO stopped in middle, you can resume in middle.
+        prior_hyper_parameters (Optional[Union[Dict, List[Dict]]]) = Hyper parameters to try first.
+        acceptable_additional_time_ratio (Union[float, int]) = Decide how much additional time can be acceptable.
+    """
```
JihwanEom, line 61: `batch_size_name` is not explained.

otx/hpo/hpo_base.py
```diff
@@ -0,0 +1,305 @@
...
+        batch_size_name: Optional[str] = None,
+        resume: bool = False,
+        prior_hyper_parameters: Optional[Union[Dict, List[Dict]]] = None,
+        acceptable_additional_time_ratio: Union[float, int] = 1.0,
+    ):
+        # pylint: disable=too-many-arguments, too-many-locals
+        check_mode_input(mode)
+        check_positive(full_dataset_size, "full_dataset_size")
+        check_positive(num_full_iterations, "num_full_iterations")
+        if not 0 < non_pure_train_ratio <= 1:
```
JihwanEom, line 88: How about 
```suggestion
        if not (0 < non_pure_train_ratio <= 1):
```

otx/hpo/hpo_base.py
```diff
@@ -0,0 +1,305 @@
...
+            raise ValueError("non_pure_train_ratio should be between 0 and 1." f" Your value is {non_pure_train_ratio}")
+        if maximum_resource is not None:
+            check_positive(maximum_resource, "maximum_resource")
+        if num_trials is not None:
+            check_positive(num_trials, "num_trials")
+        check_positive(num_workers, "num_workers")
+        if subset_ratio is not None:
+            if not 0 < subset_ratio <= 1.0:
```
JihwanEom: Same as above, 
```suggestion
            if not (0 < subset_ratio <= 1):
```

otx/hpo/hpo_base.py
```diff
@@ -0,0 +1,305 @@
...
+                raise ValueError(
```
JihwanEom, line 100: Error message is different with `non_pure_train_ratio`, could you unify these?

otx/hpo/hpo_base.py
```diff
@@ -0,0 +1,305 @@
...
+        raise NotImplementedError
+
+    @abstractmethod
+    def is_done(self):
+        """Check whether HPO algorithm is done."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def get_next_sample(self):
+        """Get next sample to trian."""
```
JihwanEom:
```suggestion
        """Get next sample to train."""
```

otx/hpo/hpo_base.py
```diff
@@ -0,0 +1,305 @@
...
+        return self._id
+
+    @property
+    def configuration(self):
+        """Configuration to train with."""
+        return self._configuration
+
+    @property
+    def iteration(self):
+        """Iteration to use for trianing."""
```
JihwanEom:
```suggestion
        """Iteration to use for training."""
```

otx/hpo/hpo_base.py
```diff
@@ -0,0 +1,305 @@
...
+    ) -> Optional[Union[float, int]]:
+        """Get best score of the trial.
+
+        Args:
+            mode (str, optional): Decide which is better between highest score or lowest score. Defaults to "max".
+            resource_limit (Optional[Union[float, int]], optional): Find a best score among the score at resource
+                                                                    lower than this value. Defaults to None.
+
+        Returns:
+            Optional[Union[float, int]]: Best score. If there is not score, return None.
```
JihwanEom:
```suggestion
            Optional[Union[float, int]]: Best score. If there is no score, return None.
```

otx/hpo/hpo_runner.py
```diff
@@ -0,0 +1,386 @@
+"""HPO runner and resource manager class."""
+
+# Copyright (C) 2022 Intel Corporation
```
JihwanEom, line 3:
```suggestion
# Copyright (C) 2023 Intel Corporation
```

otx/hpo/hpo_runner.py
```diff
@@ -0,0 +1,386 @@
...
+    def have_available_resource(self):
+        """Check that there is available resource."""
+        raise NotImplementedError
+
+
+class CPUResourceManager(ResourceManager):
+    """Resource manager class for CPU.
+
+    Args:
+        num_parallel_trial (int, optional): How many trials to trun parallelly. Defaults to 4.
```
JihwanEom:
```suggestion
        num_parallel_trial (int, optional): How many trials to run in parallel. Defaults to 4.
```

otx/hpo/hpo_runner.py
```diff
@@ -0,0 +1,386 @@
...
+    num_parallel_trial: Optional[int] = None,
+    num_gpu_for_single_trial: Optional[int] = None,
+    available_gpu: Optional[str] = None,
+) -> ResourceManager:
+    """Get an appropriate resource manager depending on current environment.
+
+    Args:
+        resource_type (Literal["gpu", "cpu"]): Which type of resource to use.
+                                               If can be changed depending on environment.
+        num_parallel_trial (Optional[int]): How many trials to trun parallelly. It's used for CPUResourceManager.
```
JihwanEom:
```suggestion
        num_parallel_trial (Optional[int]): How many trials to run in parallel. It's used for CPUResourceManager.
```

otx/hpo/hpo_runner.py
```diff
@@ -0,0 +1,386 @@
...
+                                            Defaults to None.
+        num_gpu_for_single_trial (Optional[int]): How many GPUs is used for a single trial.
+                                                  It's used for GPUResourceManager. Defaults to None.
+        available_gpu (Optional[str]): How many GPUs are available. It's used for GPUResourceManager. Defaults to None.
+
+    Raises:
+        ValueError: If resource_type is neither 'gpu' nor 'cpu, then raise an error.
```
JihwanEom:
```suggestion
        ValueError: If resource_type is neither 'gpu' nor 'cpu', then raise an error.
```

otx/hpo/hpo_runner.py
```diff
@@ -0,0 +1,386 @@
...
+
+class HpoLoop:
+    """HPO loop manager to run trials.
+
+    Args:
+        hpo_algo (HpoBase): HPO algorithms.
+        train_func (Callable): Function to train a model.
+        resource_type (Literal['gpu', 'cpu'], optional): Which type of resource to use.
+                                                         If can be changed depending on environment. Defaults to "gpu".
+        num_parallel_trial (Optional[int], optional): How many trials to trun parallelly.
```
JihwanEom:
```suggestion
        num_parallel_trial (Optional[int], optional): How many trials to run in parallel.
```

otx/hpo/hpo_runner.py
```diff
@@ -0,0 +1,386 @@
...
+                                                    It's used for CPUResourceManager. Defaults to None.
+        num_gpu_for_single_trial (Optional[int], optional): How many GPUs is used for a single trial.
```
JihwanEom:
```suggestion
        num_gpu_for_single_trial (Optional[int], optional): How many GPUs are used for a single trial.
```

otx/hpo/hpo_runner.py
```diff
@@ -0,0 +1,386 @@
...
+        pipe1, pipe2 = self._mp.Pipe(True)
+        process = self._mp.Process(
+            target=_run_train,
+            args=(
+                self._train_func,
+                trial.get_train_configuration(),
+                partial(_report_score, pipe=pipe2, trial_id=trial.id),
+            ),
+        )
+        os.environ = origin_env
```
JihwanEom, line 102: It's hard for me to understand `origin_env = os.environ` and `os.environ = origin_env`. Could you explain for this? Thank you!

otx/hpo/hpo_runner.py
```diff
@@ -0,0 +1,386 @@
...
+    available_gpu: Optional[str] = None,
+):
+    """Run the HPO loop.
+
+    Args:
+        hpo_algo (HpoBase): HPO algorithms.
+        train_func (Callable): Function to train a model.
+        resource_type (Literal['gpu', 'cpu'], optional): Which type of resource to use.
+                                                         If can be changed depending on environment. Defaults to "gpu".
+        num_parallel_trial (Optional[int], optional): How many trials to trun parallelly.
```
JihwanEom:
```suggestion
        num_parallel_trial (Optional[int], optional): How many trials to run in parallel.
```

otx/hpo/hpo_runner.py
```diff
@@ -0,0 +1,386 @@
...
+                                                      It's used for CPUResourceManager. Defaults to None.
+        num_gpu_for_single_trial (Optional[int], optional): How many GPUs is used for a single trial.
```
JihwanEom:
```suggestion
        num_gpu_for_single_trial (Optional[int], optional): How many GPUs are used for a single trial.
```

otx/hpo/utils.py
```diff
@@ -0,0 +1,124 @@
+"""HPO utility."""
+
+# Copyright (C) 2022 Intel Corporation
```
JihwanEom:
```suggestion
# Copyright (C) 2023 Intel Corporation
```

otx/hpo/utils.py
```diff
@@ -0,0 +1,124 @@
...
+
+    Args:
+        value (Any): value to validate.
+        variable_name (Optional[str], optional): name of value. It's used for error message. Defaults to None.
+        error_message (Optional[str], optional): Error message to use when type is different. Defaults to None.
+
+    Raises:
+        ValueError: If value is negative, the error is raised.
+    """
+    if value < 0:
```
JihwanEom: Hmm.. can we merge `check_positive` and `check_not_negative` in one function? or.. we can design the function that checks the variable is zero or not.

otx/hpo/utils.py
```diff
@@ -0,0 +1,124 @@
...
+    """Validate that mode is 'max' or 'min'.
+
+    Args:
+        mode (str): string to validate.
+
+    Raises:
+        ValueError: If 'mode' is not both 'max' and 'min', the error is raised.
+    """
+    if mode not in ["max", "min"]:
+        raise ValueError("mode should be max on min.\n" f"Your value : {mode}")
```
JihwanEom:
```suggestion
        raise ValueError("mode should be max or min.\n" f"Your value : {mode}")
```

otx/hpo/hyperband.py
```diff
@@ -0,0 +1,959 @@
+"""Hyperband implementation."""
+
+# Copyright (C) 2022 Intel Corporation
```
JihwanEom, line 3:
```suggestion
# Copyright (C) 2023 Intel Corporation
```

otx/hpo/hyperband.py
```diff
@@ -0,0 +1,959 @@
...
+    check_positive,
+    left_is_better,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def _check_reduction_factor_value(reduction_factor: int):
+    if reduction_factor < 2:
+        raise ValueError("reduction_factor should be at least 2.\n" f"your value : {reduction_factor}")
```
JihwanEom:
```suggestion
        raise ValueError("reduction_factor should be greater than 2.\n" f"your value : {reduction_factor}")
```

jaegukhyun (dismissed):

otx/hpo/hyperband.py
```diff
@@ -0,0 +1,959 @@
...
+
+
+class AshaTrial(Trial):
+    """ASHA trial class.
+
+    Args:
+        trial_id (Any): Id of the trial.
+        configuration (Dict): Configuration for the trial.
+        train_environment (Optional[Dict]): Train environment for the trial.
+                                            For, example, subset ratio can be includes. Defaults to None.
```
JihwanEom:
```suggestion
                                            For example, subset ratio can be included. Defaults to None.
```

otx/hpo/hyperband.py
```diff
@@ -0,0 +1,959 @@
...
+        return self._bracket
+
+    @bracket.setter
+    def bracket(self, val: int):
+        """Setter for bracket."""
+        check_not_negative(val, "bracket")
+        self._bracket = val
+
+    def save_results(self, save_path: str):
+        """Save a result of the trail at 'save_path'."""
```
JihwanEom:
```suggestion
        """Save a result of the trial at 'save_path'."""
```

otx/hpo/hyperband.py
```diff
@@ -0,0 +1,959 @@
...
+
+        self._reduction_factor = reduction_factor
+        self._num_required_trial = num_required_trial
+        self._resource = resource
+        self._trials: List[AshaTrial] = []
+        self._rung_idx = rung_idx
+
+    @property
+    def num_required_trial(self):
+        """Necessary trials for the rung."""
```
JihwanEom:
```suggestion
        """Number of required trials for the rung."""
```

otx/hpo/hyperband.py
```diff
@@ -0,0 +1,959 @@
...
+                best_score = trial_score
+
+        return best_trial
+
+    def need_more_trials(self) -> bool:
+        """Check whether the rung needs more trials."""
+        return self.num_required_trial > self.get_num_trials()
+
+    def get_num_trials(self) -> int:
+        """Number of trails the rung has."""
```
JihwanEom:
```suggestion
        """Number of trials the rung has."""
```

otx/hpo/hyperband.py
```diff
@@ -0,0 +1,959 @@
...
+        if best_trial is not None:
+            self._rungs[rung_idx + 1].add_new_trial(best_trial)
+
+        return best_trial
+
+    def get_next_trial(self) -> Optional[AshaTrial]:
+        """Get next trial to train.
+
+        Returns:
+            Optional[AshaTrial]: Next trial to train. There is no trial to trian, then return None.
```
JihwanEom:
```suggestion
            Optional[AshaTrial]: Next trial to train. There is no trial to train, then return None.
```

otx/hpo/hyperband.py
```diff
@@ -0,0 +1,959 @@
...
+            if not bracket.is_done():
+                next_sample = bracket.get_next_trial()
+                if self._asynchronous_bracket and next_sample is None:
+                    continue
+                break
+
+        return next_sample
+
+    def _make_trial_to_estimate_resource(self) -> AshaTrial:
+        """Trial to esimate a maximum resource or minimum resource."""
```
JihwanEom:
```suggestion
        """Trial to estimate a maximum resource or minimum resource."""
```

otx/hpo/hyperband.py
```diff
@@ -0,0 +1,959 @@
...
+
+    def save_results(self):
+        """Save a ASHA result."""
+        for idx, bracket in self._brackets.items():
+            save_path = osp.join(self.save_path, str(idx))
+            os.makedirs(save_path, exist_ok=True)
+            bracket.save_results(save_path)
+
+    def auto_config(self) -> List[Dict[str, Any]]:
+        """Configure ASHA automatically aligning with possible resoure.
```
JihwanEom:
```suggestion
        """Configure ASHA automatically aligning with possible resource.
```

otx/hpo/__init__.py
eunwoosh: This code is written at that time.

otx/hpo/hpo_base.py
eunwoosh: This code is written at that time.

otx/hpo/hpo_base.py
eunwoosh: I originally implemented as you suggested, but it raised an error in pre-commit.

otx/hpo/hpo_base.py
eunwoosh: I originally implemented as you suggested, but it raised an error in pre-commit.

otx/hpo/hpo_base.py
eunwoosh: I'll update it.

otx/hpo/hpo_base.py
eunwoosh: I'll update it.

otx/hpo/hpo_runner.py
eunwoosh: This code is written at that time.

otx/hpo/hpo_runner.py
eunwoosh: I'll update it.

otx/hpo/hpo_runner.py
eunwoosh: I'll update it.

otx/hpo/hpo_runner.py
eunwoosh: I'll update it.

otx/hpo/hpo_runner.py
eunwoosh: I'll update it.

otx/hpo/utils.py
eunwoosh: I think there is no reason to merge it, because as far as I know, the smaller a function, the better. And I also think current implementation is better to understand than merging it.

otx/hpo/utils.py
eunwoosh: I'll update it.

otx/hpo/hyperband.py
eunwoosh: I'll update it.

otx/hpo/hyperband.py
eunwoosh: I'll update it.

otx/hpo/hyperband.py
eunwoosh: I'll update it.

otx/hpo/hyperband.py
eunwoosh: I'll update it.

otx/hpo/hyperband.py
eunwoosh: I'll update it.

otx/hpo/hyperband.py
eunwoosh: I'll update it.

otx/hpo/hyperband.py
eunwoosh: I'll update it.

otx/hpo/hyperband.py
eunwoosh: I'll update it.

otx/hpo/hyperband.py
eunwoosh: This code is written at that time.

otx/hpo/hpo_runner.py
eunwoosh: Purpose of the code is passing appropriate environment variables to child process. But I think I forget to deepcopy the `os.environ`. Thanks for pointing it!

eunwoosh:
@JihwanEom , thanks for detailed review! I'll update what I commented as 'I'll update it'. Could you check that?

otx/hpo/utils.py
```diff
@@ -0,0 +1,124 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions
+# and limitations under the License.
+
+from typing import Any, List, Optional, Tuple, Union
+
+
+def check_type(
```
sungmanc: This function is a common function also used by other codes. So, how about moving this under the `common`?

otx/hpo/utils.py
```diff
@@ -0,0 +1,124 @@
...
+        mode (str): string to validate.
+
+    Raises:
+        ValueError: If 'mode' is not both 'max' and 'min', the error is raised.
+    """
+    if mode not in ["max", "min"]:
+        raise ValueError("mode should be max or min.\n" f"Your value : {mode}")
+
+
+def left_is_better(val1, val2, mode: str) -> bool:
```
sungmanc: Suggest to change the name : better --> larger.

otx/hpo/hpo_runner.py
```diff
@@ -0,0 +1,387 @@
...
+        else:
+            self._available_gpu.extend(self._usage_status[trial_id])
+            del self._usage_status[trial_id]
+
+    def have_available_resource(self):
+        """Check that there is available resource."""
+        return len(self._available_gpu) >= self._num_gpu_for_single_trial
+
+
+def get_resource_manager(
```
sungmanc: Even when we consider the GPU training, CPU, GPU managers are required. 
Maybe this function can automatically select the proper manager, users might want to use CPU, GPU manager for the GPU training case.

This comment also consider the further works. (auto-config by checking the HW resource)

otx/hpo/hpo_runner.py
```diff
@@ -0,0 +1,387 @@
...
+
+import torch
+
+from otx.hpo.hpo_base import HpoBase, Trial, TrialStatus
+from otx.hpo.utils import check_positive
+
+logger = logging.getLogger(__name__)
+
+
+class ResourceManager(ABC):
```
sungmanc: This abstract class is awesome ! However, I think we need to make separate `.py` file for this if we consider the further implementation.

sungmanc (commented):
To be honest, since there were too many attributes in each class, I couldn't totally understand.
However, the overall implementation design is incredible. 

In some cases, you used `Any` type for mypy, I think we need to avoid using `Any` type. When we used `Any` for typing, it has actually no meaning to use type-hint.

otx/hpo/utils.py
eunwoosh: Actually, it's for comparing score. the lower the better in some metric like loss and there is the other things like accuracy to the contrary. So how to let it be?

otx/hpo/utils.py
eunwoosh: I agree with you. But I don't know what exactly `common` means. Could you let me know which one is your meaning?

1. making otx/hpo/common directory and using it.
2. making otx/common directory and using it.
3. using otx/algorithm/common directory.

otx/hpo/utils.py
sungmanc: Then, adding `metric` as a suffix is better option to avoid the confusion. `def left_is_better_metric`

otx/hpo/utils.py
sungmanc: Both of 2, 3 looks good to me. How about https://github.com/openvinotoolkit/training_extensions/blob/hpopt_consolidation/otx/algorithms/common/utils/utils.py ?

otx/hpo/hpo_runner.py
eunwoosh: It also uses the CPU manager when user selects CPU manager even if GPU is available. Could you check again?

otx/hpo/hpo_runner.py
eunwoosh: I see. I'll reorganized classes as you said.

otx/hpo/utils.py
eunwoosh: Honestly, Topology looks a little wrong (otx/hpo uses otx/algorithm/common). But as you said, moving it to make it available to be used in other place is good choice. Maybe we can discuss about moving otx/algorithm/common to otx/common later. For now, I'll move it to otx/algorithm/common.

otx/hpo/utils.py
eunwoosh: In my perspective, honestly, it's not confusing. Then how about to rename it to `left_value_is_better`? 
and I think that it also can be moved to `common`. how about letting it be as general name and making ii usable for general purpose?

otx/hpo/hpo_runner.py
sungmanc: When we assume that the GPU training environment, then users might want to control CPU and GPU at the same time. So, in this case, this function should return both `CPUResourceManager` and `GPUResourceManager`. 

However, this function couldn't return the both of them. is it intended? if so, it's okay

eunwoosh: My intention is that `GPUResourceManager` is used for training w/ GPU and `CPUResourceManager ` is used for training w/ CPU. For now, `CPUResourceManager` doesn't manages resources considering real CPU resources. It just gets how many trials will be run in parallel. So it's not meaningful to use both of them.

harimkang (dismissed):

goodsong81 (dismissed):
LGTM! Welcome HPO :)

eunwoosh:
> To be honest, since there were too many attributes in each class, I couldn't totally understand. However, the overall implementation design is incredible.
> 
> In some cases, you used `Any` type for mypy, I think we need to avoid using `Any` type. When we used `Any` for typing, it has actually no meaning to use type-hint.

@sungmanc,  Thanks for your compliment and I also understand what you said about lots of attributes. It could be better to divide some big classes to small classes, but it would be quite a big work. So, I'll try to refactor them after thinking some ideas later :)
Anyway, I reflected some of your comments. Could you check about that?

JihwanEom (dismissed):

JihwanEom (approved):

otx/hpo/utils.py
sungmanc: No objection

sungmanc (approved):

PR #1674, opened 2023-02-10
[FIX] Add `SelfSLSegmentationDatasetAdapter`

sungchul1:
## Summary
`create_pseudo_masks` in `otx/algorithms/segmentation/adapters/mmseg/utils/data_utils.py` was implemented to create pseudo masks for DetCon if they aren't created before and import them before training. However, after datumaro integration, OTX CLI dataset disappeared and now this function is not called anywhere.
In this PR, `SelfSLSegmentationDatasetAdapter` is newly added to create pseudo masks and import them. The reason of why adding the new adapter is required is that created masks cannot be loaded through datumaro due to their own directory.

## Update
- [x] add modules to create and import pseudo masks
- [x] add an adapter
- [ ] ~add condition not to skip background class during creating polygon~
Don't consider it in this PR.
- [x] update other modules (action/tools/sample_classification.py, action/tools/sample_detection.py, cli/utils/hpo.py) that are affected by these changes
- [x] unit test
image


sungchul1:
@eunwoosh I updated loading `dataset_adapter` in `cli/utils/hpo.py`.
@jaegukhyun I updated loading `dataset_adapter` in `action/tools/sample_classification.py` and `action/tools/sample_detection.py`.
Could you take a look? If there are some parts to be updated, please let me know.

otx/cli/utils/hpo.py
```diff
@@ -696,6 +696,7 @@ def _prepare_hyper_parameter(self):
     def _prepare_dataset_adapter(self):
         dataset_adapter = get_dataset_adapter(
             self._task.task_type,
+            self._model_template.hyper_parameters.parameter_overrides["algo_backend"]["train_type"]["default_value"],
```
eunwoosh, line 699: So, Intention of this code is to make hpo always use `INCREMENTAL` train type, which is default value. Am I right?

otx/core/data/adapter/segmentation_dataset_adapter.py
```diff
@@ -52,3 +66,104 @@ def get_otx_dataset(self) -> DatasetEntity:
...
+        So, it is required to manually load and set annotations.
+
+        Args:
+            train_data_roots (Optional[str]): Path for training data.
+            pseudo_mask_dir (str): Directory to save pseudo masks. Defaults to "detcon_mask".
+
+        Returns:
+            DatumaroDataset: Datumaro Dataset
+        """
+        assert train_data_roots
```
eunwoosh: Could you change this line to using `raise`? As far as I know, it's recommended to use `raise` instead `assert` in OTX.

otx/core/data/adapter/segmentation_dataset_adapter.py
```diff
@@ -52,3 +66,104 @@ def get_otx_dataset(self) -> DatasetEntity:
...
+        dataset[Subset.TRAINING] = DatumaroDataset.import_from(train_data_roots, format="image_dir")
+        self.is_train_phase = True
+
+        # Load pseudo masks
+        img_dir = None
+        total_labels = []
+        flag_create_mask = False
+        for item in dataset[Subset.TRAINING]:
+            img_path = item.media.path
+            if img_dir is None:
```
eunwoosh, line 175: As far as I understand, `img_dir` is always None. Is this `if`` statement necessary?

otx/core/data/adapter/segmentation_dataset_adapter.py
```diff
@@ -52,3 +66,104 @@ def get_otx_dataset(self) -> DatasetEntity:
...
+        pseudo_mask_roots = train_data_roots.replace(img_dir, pseudo_mask_dir)  # type: ignore
+        if flag_create_mask:
+            # Save dataset_meta.json for newly created pseudo masks
+            meta = {"label_map": {i + 1: f"target{i+1}" for i in range(max(total_labels) + 1)}}
+            with open(os.path.join(pseudo_mask_roots, "dataset_meta.json"), "w", encoding="UTF-8") as f:
+                json.dump(meta, f, indent=4)
+
+        # Make categories for pseudo masks
+        label_map = parse_meta_file(os.path.join(pseudo_mask_roots, "dataset_meta.json"))
+        dataset[Subset.TRAINING]._data._categories = make_categories(label_map)
```
eunwoosh: I think it is wrong to use the protected attribute in class.

otx/core/data/adapter/__init__.py
```diff
@@ -17,54 +17,80 @@
...
 import importlib
 
+from otx.algorithms.common.configs.training_base import TrainType
 from otx.api.entities.model_template import TaskType
 
 ADAPTERS = {
     TaskType.CLASSIFICATION: {
-        "module_name": "classification_dataset_adapter",
-        "class": "ClassificationDatasetAdapter",
+        "INCREMENTAL": {
```
jaegukhyun, line 25: 1. Doesn't classification selfsl task require adapter?
2. Can classification selfsl task get any adapter in this structure?

jaegukhyun (dismissed):
LGTM, but Eunwoo's comments totally make sense. Could you take a look?

otx/cli/utils/hpo.py
sungchul1: No. It is for `train_type`.
I checked when `train_type` is set to `SELFSUPERVISED` in cli, `default_value` is set to `SELFSUPERVISED` although `default_value` is `INCREMENTAL` in `configuration.yaml`.
We can change `default_value` in cli.

otx/core/data/adapter/segmentation_dataset_adapter.py
sungchul1: I inserted if statement to update `img_dir` in the first loop.
`img_dir` is updated only in the first loop, and it is not updated afterwards and is used in the function.
https://github.com/openvinotoolkit/training_extensions/blob/47c7c0fdba103fbb43ddad6b8f80a3498f5ed8b0/otx/core/data/adapter/segmentation_dataset_adapter.py#L110-L112

Do you think it would be better to update `img_dir` every iteration?

otx/core/data/adapter/segmentation_dataset_adapter.py
sungchul1: Done.

otx/core/data/adapter/segmentation_dataset_adapter.py
sungchul1: Done.

jaegukhyun (dismissed):

jaegukhyun:
Could you rebase once more?

sungchul1:
@jaegukhyun Could you take a look at this unit test?
All fails showed this
`RuntimeError: expand(torch.FloatTensor{[1, 1, 1, 33, 33]}, size=[1, -1, -1, -1]): the number of sizes provided (4) must be greater or equal to the number of dimensions in the tensor (5)`

JihwanEom:
> @jaegukhyun Could you take a look at this unit test? All fails showed this `RuntimeError: expand(torch.FloatTensor{[1, 1, 1, 33, 33]}, size=[1, -1, -1, -1]): the number of sizes provided (4) must be greater or equal to the number of dimensions in the tensor (5)`

#1747 will fix this issue, please rebase after merging that PR.

sungchul1:
@jaegukhyun @eunwoosh @sungmanc Could you check it? All tests were passed.

jaegukhyun (approved):

JihwanEom (approved):

PR #1676, opened 2023-02-10
[Fix] Missing config.json

eugene123tw:
## This PR includes:
- Apply try-except when config.json is missing for detection task. 
- Add `e` as good names for pylint.basic (For using the convention `Exception as e`)
- `disable=broad-except` looks reasonable, codes in try statement can cause various errors (`json.JSONDecodeError`, `TypeError`, `ValueError`, `AttributeError`). Both for stability and readability, `disable=broad-except` was adopted.


goodsong81 (approved):
Thank you for the quick fix!

sungmanc (approved):

PR #1682, opened 2023-02-13
[Test] Add unit tests for otx.algorithms.action.utils

jaegukhyun:
This PR includes adding unit tests for otx.algorithms.action.utils

**Coverage**
image

### DIFF
--- a/tests/unit/algorithms/action/utils/test_action_convert_public_data_to_cvat.py
+++ b/tests/unit/algorithms/action/utils/test_action_convert_public_data_to_cvat.py
@@ -5,7 +5,6 @@
 #
 
 import numpy as np
-from lxml import etree
 
 from otx.algorithms.action.utils.convert_public_data_to_cvat import (
     convert_ava_dataset_to_datumaro,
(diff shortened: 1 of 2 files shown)


eunwoosh (dismissed):

tests/unit/algorithms/action/utils/test_action_convert_public_data_to_cvat.py
```diff
@@ -0,0 +1,171 @@
+"""Unit Test for otx.algorithms.action.utils.convert_public_data_to_cvat."""
+
+# Copyright (C) 2023 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
+
+import numpy as np
+from lxml import etree
```
harimkang: This seems like a security issue. Can you check it out?
image

sungmanc: > This seems like a security issue. Can you check it out? image

Thanks for the nice finding ! 
After some researches, I found that we are in the safe environment since we are using self-made XML files. 

This vulnerability comes from maliciously crafted XML.

Reference: https://docs.python.org/3/library/xml.html#xml-vulnerabilities
According to the reference, we also don't have other candidates when we want to parse XML

jaegukhyun: At this time, removing lxml in original code is too hard. Therefore I remove lxml in the test codes first.

harimkang (approved):

eunwoosh (approved):

PR #1689, opened 2023-02-14
[TEST] Weekly Test

sungmanc:
# Summary
* Edit `test_suite/run_test_command.py` to validate the model performance.
  *  Add arguments in existed test cases ( criteria, reg_threshold, result_dict ) to compare with criteria.
  *  Add test function for evaluation.

* Add `tests/regression/{TASKS}/test_***.py` 
    * All other cases used in E2E or Integration tests will be ignored.
* Add `regression_config.json` to save the model's performance metric and data_path
  * Since we need to fix some algorithms, I set all criteria as 0.00 to make all PASS cases (criteria value will be updated)
  * Some test cases are not included (e.g., action detection export, nncf, ...)
  
* Add `test_helpers` for regression tests to load the configuration
* Add `summarize_test_results.py` and edit the `tox.ini` to generate the result.csv by automatically.

* After weekly test (executed by `tox -e weekly-test`), we can find the results csv files on `.tox` folder.
image

### Result examples
image

image


goodsong81 (dismissed):
LGTM in general. Thank you for the great effort in such a short time.
Leaving a few comments. (not mandatory to address in this PR)
- As we discussed online, we need some aggregation feature for performance numbers.
- I support @eunwoosh's refactoring to rule out test codes from coverage computation. We need to move common tests code in tests/test_suite at some point.

otx/cli/utils/tests.py
```diff
@@ -763,3 +764,36 @@ def otx_train_auto_config(root, otx_dir: str, args: Dict[str, str]):
...
+
+    with open(performance_json_path) as read_file:
+        trained_performance = json.load(read_file)
+
+    model_criteria = criteria[template.name]
+    modified_criteria = model_criteria - (model_criteria * threshold)
+    for k in trained_performance.keys():
+        print(f"trained_performance: {trained_performance[k]}, modified_criteria:{modified_criteria}")
+        assert (
+            trained_performance[k] <= modified_criteria
```
kprokofi: As I see that criteria checks that model performance <= 0.95?
Why? I thought we will train all  our templates and record exact numbers (or small performance range).
why do we need: `model_criteria - (model_criteria * threshold)` as we always store model_criteria as 1.0 in `json`.
I don't understand what this test checks

tests/regression/classification/test_classification.py
```diff
@@ -0,0 +1,266 @@
...
+
+    @e2e_pytest_component
+    @pytest.mark.parametrize("template", templates, ids=templates_ids)
+    def test_otx_train_semisl(self, template, tmp_dir_path):
+        tmp_dir_path = tmp_dir_path / "multi_class_cls/test_semisl"
+        config_semisl = load_regression_configuration(otx_dir, "classification", "semi_supervised", "multi_class")
+        args_semisl = config_semisl["data_path"]
+
+        args_semisl["train_params"] = []
+        args_semisl["train_params"].extend(["--algo_backend.train_type", "SEMISUPERVISED"])
```
kprokofi: maybe we can avoid assigning to the empty list first?

otx/cli/utils/tests.py
sungmanc: @kprokofi , This PR is still WIP. 
For the current status, since we didn't know the current model's performance, we need to check it first.
So, I want to make FAIL for all test cases. That's the reason why I set the 1.0 value. 

After measuring the current model's performance, I'll edit the criteria value. Thanks for question !

tests/regression/classification/test_classification.py
sungmanc: I agree. In this time, I just want to make fast implementation. :)

sungmanc: So, I might miss many things. Please give your opinion 👍

tests/regression/detection/test_detection.py
```diff
@@ -0,0 +1,263 @@
...
+# Detection
+detection_regression_config = load_regression_configuration(otx_dir, TASK_TYPE, TRAIN_TYPE, LABEL_TYPE)
+detection_data_args = detection_regression_config["data_path"]
+detection_data_args["train_params"] = ["params", "--learning_parameters.num_iters", REGRESSION_TEST_EPOCHS]
+
+
+class TestRegressionDetection:
+    def setup_method(self):
+        self.label_type = LABEL_TYPE
+        self.performance = {}
```
eunwoosh, line 58: It's just a question. As far as I understand, `self.performance` isn't used across the functions of class. Why it's initialized in `setup_method` instead of initializing it as local variable in each test function?

tests/regression/regression_test_helpers.py
```diff
@@ -0,0 +1,116 @@
...
+        Dict[str, Any]: The dictionary that includes data roots
+    """
+    root_path = Path(otx_dir)
+    with open(root_path / ("tests/regression/regression_config.json"), "r") as f:
+        reg_config = json.load(f)
+    return reg_config
+
+
+def load_regression_configuration(
+    otx_dir: str, task_type: str, train_type: str = "", label_type: str = ""
```
eunwoosh, line 103: How about using `typing.Literal` instead of using `str` for `task_type, `train_type and `label_type?

tests/test_suite/run_test_command.py
```diff
@@ -237,7 +240,9 @@ def otx_eval_testing(template, root, otx_dir, args):
     assert os.path.exists(f"{template_work_dir}/trained_{template.model_template_id}/performance.json")
 
 
-def otx_eval_openvino_testing(template, root, otx_dir, args, threshold):
+def otx_eval_openvino_testing(
+    template, root, otx_dir, args, threshold=0.0, criteria={}, reg_threshold=0.10, result_dict={}
```
eunwoosh: IMHO, it's better to return values instead of saving it in `result_dict` because current structure make it hard to understand which values are used on caller side. How do you think about that?

tests/regression/action/test_action_detection.py
```diff
@@ -0,0 +1,74 @@
...
+otx_dir = os.getcwd()
+templates = Registry("otx/algorithms/action").filter(task_type=TASK_TYPE.upper()).templates
+templates_ids = [template.model_template_id for template in templates]
+
+
+result_dict = get_result_dict(TASK_TYPE)
+result_dir = f"/tmp/regression_test_results/{TASK_TYPE}"
+Path(result_dir).mkdir(parents=True, exist_ok=True)
+
+# Action Classification
```
eunwoosh:
```suggestion
# Action Detection
```

tests/regression/detection/test_detection.py
```diff
@@ -0,0 +1,263 @@
...
+
+# Detection
+detection_regression_config = load_regression_configuration(otx_dir, TASK_TYPE, TRAIN_TYPE, LABEL_TYPE)
+detection_data_args = detection_regression_config["data_path"]
+detection_data_args["train_params"] = ["params", "--learning_parameters.num_iters", REGRESSION_TEST_EPOCHS]
+
+
+class TestRegressionDetection:
+    def setup_method(self):
+        self.label_type = LABEL_TYPE
```
eunwoosh, line 57: How about using global variable `LABEL_TYPE` instead of `self.label_type`?

eunwoosh (commented):
Thanks for your work! This regression test can make OTX stable :)
I left some comments. please take a look.

tests/regression/action/test_action_classification.py
```diff
@@ -0,0 +1,134 @@
+"""Tests for Action Classification with OTX CLI."""
+# Copyright (C) 2022 Intel Corporation
```
jaegukhyun, line 2: Need to change copyright

tests/regression/action/test_action_classification.py
```diff
@@ -0,0 +1,134 @@
...
+TRAIN_TYPE = "supervised"
+LABEL_TYPE = "multi_class"
+
+otx_dir = os.getcwd()
+templates = Registry("otx/algorithms/action").filter(task_type=TASK_TYPE.upper()).templates
+templates_ids = [template.model_template_id for template in templates]
+
+
+result_dict = get_result_dict(TASK_TYPE)
+result_dir = f"/tmp/regression_test_results/{TASK_TYPE}"
```
jaegukhyun, line 43: It's simple question, is there a logic to delete result_dir?

tests/regression/anomaly/test_anomaly_classificaiton.py
```diff
@@ -0,0 +1,214 @@
...
+
+otx_dir = os.getcwd()
+templates = Registry("otx/algorithms/anomaly").filter(task_type=TASK_TYPE.upper()).templates
+templates_ids = [template.model_template_id for template in templates]
+
+result_dict = get_result_dict(TASK_TYPE)
+result_dir = f"/tmp/regression_test_results/{TASK_TYPE}"
+Path(result_dir).mkdir(parents=True, exist_ok=True)
+
+# Detection
```
jaegukhyun: I think these kind of comments contain wrong information and useless, how about removing theses comments

jaegukhyun (commented):
Generally LGTM, I left a few comments

tests/regression/regression_test_helpers.py
sungmanc: Do this brings an advantage?

tests/regression/detection/test_detection.py
sungmanc: Nothing special,

Class variables defined in setup or teardown methods are always re-initialized at every class function.
Since I want to save the result data at every test function, I explicitly defined `self.performance` in the setup function.

tests/regression/regression_test_helpers.py
eunwoosh: Yes, it brings a similar advantage as same with type hinting. Although which values are available is explained in comment, it also can be explained by type hinting using `Literal`.

tests/test_suite/run_test_command.py
sungmanc: It could be. The reason why I selected this manner is I don't want to make duplicate functions in `run_test_command.py`.
For the `otx_eval_openvino_testing` or such other test functions was implemented as no-return. So, I just followed the way

tests/regression/detection/test_detection.py
sungmanc: For the classification case, we need to use `self.label_type` since other test classes have different label_type.
However, for the detection case, we can avoid the `self.label_type`, I'm going to reflect it

i.e. 
``` python3
class TestRegressionMultiClassClassification:
    def setup_method(self):
        self.label_type = "multi_class"
        self.performance = {}

class TestRegressionMultiLabelClassification:
    def setup_method(self):
        self.label_type = "multi_label"
        self.performance = {}
```

tests/regression/action/test_action_classification.py
sungmanc: As far as I know, crontab cleans the /tmp folder regularly.

tests/regression/anomaly/test_anomaly_classificaiton.py
sungmanc: 100% Agree, Thanks

tests/regression/action/test_action_detection.py
sungmanc: Thanks

tests/test_suite/run_test_command.py
eunwoosh: Then how about making function return necessary things instead of modifying `result_dict`? You don't have to make duplicated function also.
And I forget one thing. I think you need to set default value of `criteria` as None. Using dictionary as default value could be dangerous. Please refer [this](https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python)

sungmanc: > Then how about making function return necessary things instead of modifying `result_dict`? You don't have to make duplicated function also. And I forget one thing. I think you need to set default value of `criteria` as None. Using dictionary as default value could be dangerous. Please refer [this](https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python)

For the return value, I want to change the code architecture later, not in this PR because we need to get test results within 1 week. 

In the case of dictionary initialization, I'll revise it

jaegukhyun (dismissed):

non_anomaly_results.csv
```diff
@@ -0,0 +1,61 @@
+,Task type,Train type,Label type,Model,train,export,deploy,nncf,pot,train E2E Time (Sec.),export E2E Time (Sec.),deploy E2E Time (Sec.),nncf E2E Time (Sec.),pot E2E Time (Sec.),train Eval Time (Sec.),export Eval Time (Sec.),deploy Eval Time (Sec.),nncf Eval Time (Sec.),pot Eval Time (Sec.)
```
ashwinvaidya17: Are these meant to be commited? What if we upload the csv file from the CI as an artefact?

tests/test_suite/run_test_command.py
```diff
@@ -767,3 +818,66 @@ def otx_train_auto_config(root, otx_dir: str, args: Dict[str, str]):
...
+        assert (
+            trained_performance[k] >= modified_criteria
+        ), f"Current model performance: ({trained_performance[k]}) < criteria: ({modified_criteria})."
+
+    result_dict["Model size (MB)"] = round(
+        os.path.getsize(f"{template_work_dir}/trained_{template.model_template_id}/weights.pth") / 1e6, 2
+    )
+
+
+def otx_eval_e2e_train_time(train_time_criteria, e2e_train_time, template, threshold=0.10):
```
ashwinvaidya17, line 868: Can you add a docstring to these methods to explain these tests. I am not familiar with them so I am not aware of what is `train_time_criteria` and `e2e_train_time`

tests/regression/anomaly/test_anomaly_classificaiton.py
```diff
@@ -0,0 +1,240 @@
...
```
ashwinvaidya17: Might be leftover

tests/regression/regression_test_helpers.py
```diff
@@ -0,0 +1,139 @@
...
+    result: Dict[str, Union[str, int, float]] = {
+        "data_path": "",
+        "model_criteria": 0,
+    }
+
+    if "anomaly" not in task_type:
+        if train_type == "" or label_type == "":
+            raise ValueError()
+        result["regression_criteria"] = reg_config["regression_criteria"][task_type][train_type][label_type]
+        result["kpi_e2e_train_time_criteria"] = reg_config["kpi_e2e_train_time_criteria"][task_type][train_type][
```
ashwinvaidya17, line 127: I am not aware of the terminology. What is kpi here?

tests/regression/regression_test_helpers.py
```diff
@@ -0,0 +1,139 @@
...
+    root_path = Path(otx_dir)
+    with open(root_path / ("tests/regression/regression_config.json"), "r") as f:
+        reg_config = json.load(f)
+    return reg_config
+
+
+def load_regression_configuration(
+    otx_dir: str, task_type: str, train_type: str = "", label_type: str = ""
+) -> Dict[str, Union[str, int, float]]:
+    """Load dataset path according to task, train, label types.
```
ashwinvaidya17, line 105: Maybe this docstring needs to be updated.

tests/regression/regression_test_helpers.py
```diff
@@ -0,0 +1,139 @@
+import json
```
ashwinvaidya17, line 15: We need to add license information here

tests/regression/anomaly/test_anomaly_classificaiton.py
```diff
@@ -0,0 +1,240 @@
...
+# Configurations for regression test.
+TASK_TYPE = "anomaly_classification"
+SAMPLED_ANOMALY_DATASET_CATEGORIES = random.sample(ANOMALY_DATASET_CATEGORIES, 1)
+
+otx_dir = os.getcwd()
+templates = Registry("otx/algorithms/anomaly").filter(task_type=TASK_TYPE.upper()).templates
+templates_ids = [template.model_template_id for template in templates]
+
+result_dict = get_result_dict(TASK_TYPE)
+result_dir = f"/tmp/regression_test_results/{TASK_TYPE}"
```
ashwinvaidya17, line 46: Personally, I would prefer fixtures here as they will ensure that the temporary directories are cleaned even if the test fails or is interrupted. Since we don't restart the CI is possible that tmp is not cleared. In this case we assume that overwrite will solve the issue.

```python
@pytest.fixture(scope="class")
def result_dir():
    with TemporaryDirectory(f"regression_test_results/{TASK_TYPE}") as tmpdir:
        yield tmpdir
```

> _Note:_ This is an example out of my head so I haven't tested it. This snippet might require some modification.

Other option is this: https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html#the-tmp-path-fixture

tests/regression/anomaly/test_anomaly_classificaiton.py
```diff
@@ -0,0 +1,240 @@
...
+
+        self.performance[template.name][TIME_LOG["train_time"]] = round(train_elapsed_time, 3)
+        self.performance[template.name][TIME_LOG["infer_time"]] = round(infer_elapsed_time, 3)
+        result_dict[TASK_TYPE]["train"][category].append(self.performance)
+
+    @e2e_pytest_component
+    @pytest.mark.parametrize("template", templates, ids=templates_ids)
+    @pytest.mark.parametrize("category", SAMPLED_ANOMALY_DATASET_CATEGORIES)
+    def test_otx_train_kpi_test(self, template, category):
+        results = result_dict[TASK_TYPE]["train"][category]
```
ashwinvaidya17, line 104: Can you add a docstring explaining what this test does?

tests/regression/anomaly/test_anomaly_segmentation.py
```diff
@@ -0,0 +1,238 @@
...
+    otx_eval_e2e_train_time,
+    otx_eval_openvino_testing,
+    otx_export_testing,
+    otx_train_testing,
+    pot_eval_testing,
+    pot_optimize_testing,
+)
+
+# Configurations for regression test.
+TASK_TYPE = "anomaly_segmentation"
```
ashwinvaidya17, line 38: Might be quite a lot of refactor for this PR but since the difference between the three tests is the task_type, how about we use pytest.parameterize and just keep a single file?

ashwinvaidya17 (commented):
Thanks for the efforts. I have a few questions.

non_anomaly_results.csv
sungmanc: Sorry, this file isn't uploaded. I'll remove it

tests/regression/regression_test_helpers.py
sungmanc: We have several tests for maintaining the OTX

1. Pre-merge: Unit-test + Integration-test
2. Daily: E2E test
3. Weekly: Regression + KPI + Stability test

Regression: measure the model performance w.r.t. accuracy
KPI: measure the model performance w.r.t. training time and evaluation time

tests/regression/regression_test_helpers.py
sungmanc: Agree. we need discussion for this part

tests/regression/regression_test_helpers.py
sungmanc: Done

tests/regression/anomaly/test_anomaly_classificaiton.py
sungmanc: For now, we want to keep the regression test results. If we can make automatic csv report to CI, we can change this part as you suggested

tests/regression/anomaly/test_anomaly_segmentation.py
sungmanc: Good point. However, it needs lots of changes in the current PR. 
As you know, although many parts of this PR need refactoring, we should get the regression / KPI test results until the next release. 

So, I want to suggest to postpone the refactoring process after merging this.

tests/regression/anomaly/test_anomaly_classificaiton.py
sungmanc: Done

tests/regression/anomaly/test_anomaly_classificaiton.py
sungmanc: Done

tests/test_suite/run_test_command.py
sungmanc: Done

tests/regression/regression_test_helpers.py
sungmanc: I think `str` is more explicit than `Literal`. It can be different from person by person. So, want to keep as is.

eunwoosh (approved):

sungmanc:
@chuneuny-emily , could you check this PR?

ashwinvaidya17 (approved):
Thanks

PR #1692, opened 2023-02-14
[TEST][Bugfix] Make test passed regardless of order of the list

eunwoosh:
There is a chance to fail `test_get_image_files_in_dir()` in tests/unit/cli/utils/test_io.py even if it works well due to order of the list.
This PR fixed it.


sungmanc (approved):

goodsong81 (approved):
Ah, nice catch! I also observed the failure at times randomly.

harimkang (approved):

PR #1693, opened 2023-02-14
[FIX] Remove mmcv dependency in OTX CLI

harimkang:
## Summary
Modify OTX CLI to work even in an environment without mmcv dependency (for anomaly).

Confirmed works in the `pre-installed torch + pip install -e .[anomaly]` environment.


goodsong81 (approved):
Thank you for super-quick fix.
Is OmegaConf enough to replace mmcv.Config in common & general use-case?
Then, let's use if for task area. mmcv modules should be referenced only in adapter modules. :)

sungmanc (approved):

harimkang:
> Thank you for super-quick fix. Is OmegaConf enough to replace mmcv.Config in common & general use-case? Then, let's use if for task area. mmcv modules should be referenced only in adapter modules. :)

@goodsong81 In fact, yaml files are covered by `OmegaConf`, but Python files (such as `model.py` or `data_pipeline.py`) are not covered, so they depend on mmcv Config. However, the current anomaly does **not use py files in config**, so it was only temporarily (not general solution) modified to not use mmcv Config in anomaly env.

harimkang:
Thank you for review. If there are no other comments, I will merge after checking all the integration test results.

JihwanEom (approved):

PR #1694, opened 2023-02-14
[Enhance] Add in-memory caching in dataloader

vinnamkim:
- Related issue: https://jira.devtools.intel.com/browse/CVS-103077
 - Add in-memory caching logic into data loading.
 - Update CLI argument parser for this feature.
 - Add tests for this feature.

PR #1699, opened 2023-02-15
[FIX] Add yaml recipes to package_data

vinnamkim:
From https://github.com/openvinotoolkit/training_extensions/pull/1694#issuecomment-1430627172, we observed that our package setup doesn't include `otx/**/*.yaml` when installing our package. This PR fixes it.


goodsong81 (dismissed):
Thank you for the fix!

JihwanEom (dismissed):

goodsong81 (dismissed):

goodsong81 (approved):

PR #1703, opened 2023-02-15
[FIX] Bugfix for classification & align filename for samples

JihwanEom:
## This PR includes:
- Remove exhausting logs from `model.init_weights()` in sample
- Delete unnecessary validation before NNCF for classification

### DIFF
--- a/otx/algorithms/classification/tools/classification_sample.py
+++ b/otx/algorithms/classification/tools/classification_sample.py
@@ -42,7 +42,7 @@ torch.cuda.manual_seed_all(SEED)
 torch.backends.cudnn.deterministic = True
 torch.backends.cudnn.benchmark = False
 
-logger = get_logger(name="sample")
+logger = get_logger(name="mmcls")
 
 parser = argparse.ArgumentParser(description="Sample showcasing the new API")
 parser.add_argument("template_file_path", help="path to template file")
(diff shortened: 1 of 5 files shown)


goodsong81 (approved):
Thank you for the fix.
Was the verbose logging due to "sample" logger?
(Could you explain a bit?)

JihwanEom:
> Thank you for the fix. Was the verbose logging due to "sample" logger? (Could you explain a bit?)

Yes, that's right.

`mmcv.utils.get_logger` initializes a new logger with `name` parameter. But we already have logger named `mmcls` by https://github.com/openvinotoolkit/training_extensions/blob/develop/otx/algorithms/classification/adapters/mmcls/utils/builder.py#L15. So, there will be two loggers by `logger = get_logger('sample')`. `['sample', 'mmcls']`

Even though the log level is changed temporary to `WARNING` by https://github.com/openvinotoolkit/training_extensions/blob/develop/otx/algorithms/classification/adapters/mmcls/utils/builder.py#L38, if we have initialized two loggers, the first one dumps `init_weights` log by https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/base_module.py#L96.

Therefore, even though the log level of `mmcls` logger is changed to `WARNING`, `sample` logger dumps exhausting logs. mmcv seems like supposing one logger in runtime.

PR #1705, opened 2023-02-15
[FIX] Remove background and ignored label in VOC and common semantic seg dataset in label entity set

supersoob:
This PR is for

- Remove bg label in VOC and common semantic seg dataset in label entity set
- Remove "ignored" label in VOC
- Make semantic segmentation dataset adapter generate VOC and common dataset polygon by proper label mapping


jaegukhyun (approved):
LGTM, but we cannot assure that VOC and common seg dataset works with its background label. Could you attach comparison exp results before and after this PR?

sungmanc (approved):
Thanks for the nice work !

sungmanc:
> LGTM, but we cannot assure that VOC and common seg dataset works with its background label. Could you attach comparison exp results before and after this PR?

This PR is the upgrade version of the previous one. So, we don't need to compare. 
The logic that handles the background is the same as the previous one. Also, this PR includes that additional check logic for ignored labels and for the dataset that didn't include the background.

PR #1707, opened 2023-02-15
[DOC] Update README and quick start guide

kprokofi:
This PR intends to update README and Quick guide

* I deleted CLI Readme and API one and move the description of the CLI commands to the main README. I think it now looks simpler for users and all the necessary information stored in one place
* I aligned the quick start guide with the latest CLI interface
* I deleted mention of the python notebook  since it will be substituted soon with the API tutorial in the documentation


sungmanc (dismissed):
Such a incredible work !

QUICK_START_GUIDE.md
```diff
@@ -2,13 +2,13 @@
 
 ## Prerequisites
 
-Current version of OTX was tested under following environments
+The current version of OTX was tested under the following environments
```
sovrasov: Environment? we mention only one

kprokofi:
I fixed pre-commit tests 
@sungmanc, @harimkang, @goodsong81 , could you please review?

kprokofi:
Pre-merge is failed, but it doesn't depend on this PR. This is only documentation PR

sungmanc (dismissed):
LGTM

QUICK_START_GUIDE.md
```diff
@@ -26,435 +26,377 @@ Current version of OTX was tested under following environments
...
-
 #### Development environment
 
 ```bash
-# Create development environment
-# -- need to fix '38' in tox.ini if other python version needed
-tox devenv -e pre-merge venv/otx
+# Create a development environment
+# -- need to fix '38' in tox.ini if another python version needed
+tox --devenv venv/otx -e pre-merge
```
goodsong81: We are specifying tox==4.4.5, and it's `tox devenv` in the (latest) version. :)

goodsong81 (changes requested):
Great! Thank you for the update.
I'd like to request a tiny change, though.
In addition, I would create a PR soon to align CUDA version to 11.7 actually.

QUICK_START_GUIDE.md
kprokofi: OK. Thanks!

goodsong81 (dismissed):

kprokofi:
I will reflect changes after Harim fix tomorrow. Please, do not merge now since we found some bug in CLI

QUICK_START_GUIDE.md
```diff
@@ -26,435 +26,377 @@ Current version of OTX was tested under following environments
...
+This line will create an object detection `Custom_Object_Detection_Gen3_SSD` model template with ResNet backbone from [mmdetection](https://github.com/open-mmlab/mmdetection):
 
 ```bash
-otx train template
+(otx) ...$ otx build --template Custom_Object_Detection_Gen3_SSD --backbone mmdet.ResNet --train-data-roots <path/to/train/root> --val-data-roots <path/to/val/root>
+
 ```
 
-And with the `--help` command along with `template`, you can list additional information, such as its parameters common to all model templates and model-specific hyper parameters.
+> **_Note:_** each task supports different dataset formats. To know more about which formats are supported by each task, refer to [explanation section](https://openvinotoolkit.github.io/training_extensions/explanation/index.html) in the documentation.
```
GalyaZalesskaya: To run trainings we need to have datasets in a certain format. For example, for detection task we need the following dataset format:
(ROOT)
```
|---- annotations       
  |---- instances_train.json
  |---- instances_val.json
|---- images       
  |---- train               
    |---- 000.jpg       
  |---- val
```
Can we add here the format example for Detection? Because it's easy to skip this part

QUICK_START_GUIDE.md
```diff
@@ -26,435 +26,377 @@ Current version of OTX was tested under following environments
...
+  --backbone BACKBONE   Available Backbone Type can be found using 'otx find --backbone {framework}'. If there is an already created backbone configuration yaml file, enter the corresponding path.
+```
 
-`train` command requires `template` as a positional arguement. it could be taken from the output of the `find` command above.
+This line will create an object detection `Custom_Object_Detection_Gen3_SSD` model template with ResNet backbone from [mmdetection](https://github.com/open-mmlab/mmdetection):
 
 ```bash
-otx train template
+(otx) ...$ otx build --template Custom_Object_Detection_Gen3_SSD --backbone mmdet.ResNet --train-data-roots <path/to/train/root> --val-data-roots <path/to/val/root>
+
```
GalyaZalesskaya, line 173: Can you please add here, that in case of omitting `--val-data-roots`, the val folder will be splitted automatically according to annotations_val, and corresponding ``splitted dataset`` folder will be added to workspace, so its paths can be used for training/evaluation?

QUICK_START_GUIDE.md
GalyaZalesskaya: Due to the fact that dataset format is crucial for us, maybe even we can put the note with the appropriate format and annotation naming in the beginning (before describing all commands). I understand that the `tutorials` is a better place for that, but inappropriate format is a source of all issues down the road

QUICK_START_GUIDE.md
```diff
@@ -26,435 +26,377 @@ Current version of OTX was tested under following environments
...
+  --base-rank BASE_RANK
+                        Base rank of the current node workers.
+  --world-size WORLD_SIZE
+                        Total number of workers in a worker group.
+  --data DATA           The data.yaml path want to use in train task.
+
 ```
 
-#### Model template-specific parameters
+It is also possible to start training by just passing the path to dataset roots, then the auto-configuration will be enabled. If you created a workspace with `otx build`, the training process can be started (in the workspace directory) just with `otx train` command without any additional options.
```
GalyaZalesskaya: I'd suggest to move this paragraph after code below, because overwise it looks confusing

QUICK_START_GUIDE.md
```diff
@@ -26,435 +26,377 @@ Current version of OTX was tested under following environments
...
+
+`otx train` trains a model (a particular model template) on a dataset and saves results in two files:
+
+- `weights.pth` - a model snapshot
+- `label_schema.json` - a label schema used in training, created from a dataset
+
+The results will be saved in `./model` folder by default. It can be modified by `--save-model-to` option. These files can be used by other commands: `export`, `eval`, `demo`, etc.
 
-#### Common parameters
+`otx train` receives `template` as a positional argument. Also, the path to train and val data root should be passed to the CLI to start training. The template name could be taken from the output of the `find` command above.
```
GalyaZalesskaya: Passing paths is a really tricky part here. Can you please add that --train-data-root can contain the root path to a dataset with  `annotations` and `images`, where images are split into `train` and `val`, so the script finds it automatically?

This is a convenient feature but it's really tricky part, and I even don't know, how to document it properly 🤨

GalyaZalesskaya (commented):
Thank you for this update that we really need now

QUICK_START_GUIDE.md
kprokofi: We make quick start general overview. I leave a link to explanation section where all supported dataset formats can be found for each task. However, in build and train sections I will cover some corner cases

kprokofi:
@GalyaZalesskaya I reflected your comments:
1) I move "note" for the dataset formats to the beginning.
2) I cover corner cases with `otx build` and `otx train` with some use cases including auto-split
3) I added example with COCO dataset when we have val and train in one root folder on how to pass roots in that case

@goodsong81 @harimkang @sungmanc I aligned changes from Harim fix (thanks for that!) - now I think PR is ready to merge

harimkang (approved):
Thank you for your efforts. That's a good Doc!

goodsong81 (approved):
Great!

QUICK_START_GUIDE.md
```diff
@@ -26,385 +26,381 @@ Current version of OTX was tested under following environments
...
-
 #### Development environment
 
 ```bash
-# Create development environment
-# -- need to fix '38' in tox.ini if other python version needed
-tox devenv -e pre-merge venv/otx
+# Create a development environment
+# -- need to fix '38' in tox.ini if another python version needed
+tox devenv venv/otx -e pre-merge
```
goodsong81, line 67: Nit: `python -m pip install -r requirements/dev.txt` is actually omitted here, but users could just hit `pip install tox`.

PR #1708, opened 2023-02-15
[ENHANCE] Parametrize saliency maps dumping in export parametrizable

GalyaZalesskaya:
Added `dump_features` parameter in `otx export` not to calculate and dump saliency maps and representation vectors for each export. 

Checked export and inference of pth and IR model for:
- Classification task (classic classification - MobileNet-V3-large-1x)
- Detection task (ATSS model)

Expected result: improved efficiency of deployed models if no explanation is needed 

TODO
1. Implement dumping feature according to flag for Segmentation task, Action Detection and Anomaly task
2. Tests


otx/cli/tools/export.py
```diff
@@ -83,8 +88,9 @@ def main():
     task = task_class(task_environment=environment)
 
     exported_model = ModelEntity(None, environment.get_model_configuration())
+    # args.dump_features = True
```
negvet:
```suggestion
```

otx/cli/tools/export.py
```diff
@@ -40,6 +40,11 @@ def get_args():
         "--save-model-to",
         help="Location where exported model will be stored.",
     )
+    parser.add_argument(
+        "--dump_features",
+        action="store_true",
+        help="Whether to return feature vector and saliency map for explanation purposes.",
+    )
```
negvet, line 47: Personally, like Galina did, I would also default it to False (to make IR infer fast by default). But not sure about that.

What do you think @goodsong81?

negvet (changes requested):
Minor comment from my side

negvet:
Unit tests are required, although can be added in a separate PR

otx/algorithms/classification/tasks/inference.py
```diff
@@ -195,7 +195,7 @@ def unload(self):
         self.cleanup()
 
     @check_input_parameters_type()
-    def export(self, export_type: ExportType, output_model: ModelEntity):
+    def export(self, export_type: ExportType, output_model: ModelEntity, dump_features: bool = False):
```
sovrasov, line 218: The new parameter should be added to `IExportTask` as well.

otx/cli/tools/export.py
goodsong81: I agree w/ default=False.

JihwanEom: I also agree, for classification, IR inference time was quiet long in proportion to dataset size.

JihwanEom:
Could you also check pre-commit issues?

GalyaZalesskaya:
@JihwanEom @sovrasov Your comments were reflected
@negvet I'm currently facing issues with running unit-tests, I'll make a separate PR with unit tests for this

negvet (dismissed):
LGTM, please check pre-commits

otx/algorithms/action/tasks/inference.py
```diff
@@ -307,7 +307,7 @@ def unload(self):
             self._delete_scratch_space()
 
     @check_input_parameters_type()
-    def export(self, export_type: ExportType, output_model: ModelEntity):
```
sovrasov, line 310: Why do we need kwargs here? They are not used

GalyaZalesskaya: Yes, we don't use kwargs (or dump_features) in action and anomaly tasks, but since I changed the [line in cli/export](https://github.com/openvinotoolkit/training_extensions/blob/5de193dd4080117a2e850e56345de3b1cd95a242/otx/cli/tools/export.py#L92) and added this parameter to parent IExport class, I need to have it in child classes as well, even if I don't use it.

It's an inconsistent behavior, that will be fixed after the release

sovrasov: In that case, we shouldn't break the interface conformance and make the method signature unclear.
My proposal is to follow the interface, but raise a NotImplementedError if dump_features is false.

GalyaZalesskaya: It's a great suggestion to keep the interface and code clear, I was wondering how to achieve it. Thank you for this idea.

GalyaZalesskaya:
@JihwanEom @sovrasov @negvet Tests were fixed, can you review again please?

goodsong81 (approved):
Thank you!

sungmanc (approved):

kprokofi (approved):

sovrasov (approved):

harimkang:
@GalyaZalesskaya I have a suspicion that instance-segmentation export is not working since this PR was merged.
Can you check it out?
i follow below steps:
```
otx train Custom_Counting_Instance_Segmentation_MaskRCNN_ResNet50 --train-data-roots tests/assets/car_tree_bug --val-data-roots tests/assets/car_tree_bug
cd otx-workspace-INSTANCE_SEGMENTATION
otx export
```

i got runtime error
RuntimeError: number of output names provided (5) exceeded number of outputs (3)

PR #1712, opened 2023-02-16
[FIX] Data configuration path in data.yaml to absolute path

harimkang:
## Summary
Modify data paths of `data.yaml` to be entered as absolute paths instead of relative paths
Both absolute and relative paths can be received in commands, but they are displayed as absolute paths in data.yaml.

Thank you for the reporting @jaegukhyun 

++ Unit test fail that occurs in develop has been hot-fixed here.

### DIFF
--- a/otx/cli/utils/config.py
+++ b/otx/cli/utils/config.py
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions
 # and limitations under the License.
 
-import os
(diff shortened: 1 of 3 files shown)


sungmanc (approved):

jaegukhyun (approved):
Thank you for quick fix!

eunwoosh (approved):

PR #1717, opened 2023-02-16
[ENHANCE] Refine dependency versions and requirement structure

goodsong81:
- Upgrade the default CUDA version of tox.ini: 11.6 -> 11.7 (aligned w/ Geti)
- Upgrade Datumrao version: 0.4.0.1 -> 1.0.0rc1
- Separate requirements/api.txt from base.txt (not to install irrelevant dependencies to apps and packages that use only `otx.api`
  - Now, `pip install .` would install requirements only for `otx.api`.


requirements/base.txt
```diff
@@ -1,20 +1,12 @@
...
-opencv-python>=4.5
 prettytable
 protobuf>=3.20.0
-pymongo==3.12.0
 pyyaml
-scikit-learn==0.24.*
-Shapely>=1.7.1,<=1.8.0
 imagesize==1.4.1
-datumaro==0.4.0.1
+datumaro==1.0.0rc1
```
sungmanc, line 9: I'm not sure Datumaro1.0.0 perfectly fits with the previous version. However, ,maybe there is no problem. let's check the test results.

sungmanc (approved):

requirements/base.txt
goodsong81: Yeah, if there is some problem, we can rollback as I separated api requirement (w/o datumaro) so that there will be no version conflict. Anyway, 1.0.0rc1 is already integrated to Geti develop, though.

eunwoosh (approved):

PR #1721, opened 2023-02-16
[FIX] Resolve FQ issue, Minor refactoring for DatasetAdapter

sungmanc:
# Summary
* Resolve FQ issues occurred in E2E tests
* Minor refactoring for `DatasetAdapter`: Use `self.data_type` instead of `self.data_type_candidates[0]`.
* Now, there are 3FAIL cases regarding the openvino in E2E test.

image


goodsong81 (approved):
Thank you for the fix. It was annoying.

PR #1724, opened 2023-02-16
[ENHANCE] Parametrize: (1) saliency map postprocessing and (2) explain targets

negvet:
Add `process_saliency_maps` and `explain_predicted_classes` to `InferenceParameters`

In this PR, by default, saliency maps are not getting processed and explanations are provided only for predicted classes (both for torch and ov tasks).

Result: this helps to overcome compute and memory overhead on the OTX side.

Please note that explain load of task.infer will be deprecated after using task.explain by Geti.

TODO: tests


otx/algorithms/classification/tasks/openvino.py
```diff
@@ -227,6 +227,7 @@ def infer(
         dataset_size = len(dataset)
         for i, dataset_item in enumerate(dataset, 1):
             predicted_scene, probs, saliency_map, repr_vector, act_score = self.inferencer.predict(dataset_item.numpy)
+            item_labels = predicted_scene.annotations[0].get_labels()
             dataset_item.append_labels(predicted_scene.annotations[0].get_labels())
```
sovrasov: dataset_item.append_labels(item_labels)

otx/algorithms/classification/tasks/openvino.py
```diff
@@ -262,14 +264,16 @@ def explain(
         dataset_size = len(dataset)
         for i, dataset_item in enumerate(dataset, 1):
             predicted_scene, _, saliency_map, _, _ = self.inferencer.predict(dataset_item.numpy)
+            item_labels = predicted_scene.annotations[0].get_labels()
             dataset_item.append_labels(predicted_scene.annotations[0].get_labels())
```
sovrasov: ditto

otx/algorithms/detection/tasks/inference.py
```diff
@@ -202,8 +207,9 @@ def _explain_detector(
             dataset=dataset,
             explainer=explain_parameters.explainer if explain_parameters else None,
         )
+        detections = results["outputs"]["detections"]
         explain_results = results["outputs"]["saliency_maps"]
-        return explain_results
+        return detections, explain_results
```
sovrasov, line 217: Does these return items correspond to the docstring and type annotation of the method?

otx/algorithms/detection/tasks/inference.py
```diff
@@ -353,36 +359,47 @@ def _update_stage_module(self, stage_module):
...
-
+            shapes = self._get_shapes(all_results, dataset_item.width, dataset_item.height, confidence_threshold)
             dataset_item.append_annotations(shapes)
 
             if feature_vector is not None:
                 active_score = TensorEntity(name="representation_vector", numpy=feature_vector.reshape(-1))
                 dataset_item.append_metadata_item(active_score, model=self._task_environment.model)
 
             if saliency_map is not None:
+                labels = list(self._labels)
```
sovrasov: Isn't self._labels a list already or it's a tuple? We can use copy() if it's a list

otx/algorithms/classification/tasks/openvino.py
negvet: Fixed!

negvet: Fixed!

otx/algorithms/classification/tasks/openvino.py
negvet: Fixed!

otx/algorithms/detection/tasks/inference.py
negvet: Thanks for the notice, fixed

otx/algorithms/detection/tasks/inference.py
negvet: Right, `self._labels` is a list. But I need to modify it. Below is a bit more justification.

If I will do `labels = self._labels` and then modify `labels` then `self._labels` would also be changed because `labels` is not a new list by refer to `self._labels`.
`list()` creates a new list that I can safely modify.

sovrasov: It's ok, but I just meant that the implicit list() call looks like a type conversion, while copy() emphasizes that we are working within the same data type.

negvet: Oh I see what you mean. Will update. Thanks!

sovrasov: Yeah, let's keep it pythonic! copy = list(a) feels like a call of a copy constructor in c++.

sovrasov (commented):
Will tests be added in this PR? Or due to the tight schedule we have to postpone it?

negvet:
> Will tests be added in this PR? Or due to the tight schedule we have to postpone it?

Yeah, not sure I will be in time adding new tests. Still need to look at CI though.
New tests will be added separately.

tests/integration/api/xai/test_api_xai_sanity.py
```diff
@@ -38,27 +38,37 @@
 
 save_model_to = "/tmp/otx_xai/"
```
sovrasov: Is there any specific reason in preferring a hardcoded path instead of tempfile.TemporaryDirectory?

negvet: No there is no, fixed!

sovrasov (dismissed):

tests/integration/api/xai/test_api_xai_sanity.py
```diff
@@ -4,13 +4,13 @@
 
 import os
 import os.path as osp
+import tempfile
 
 import pytest
 import torch
 
-from otx.algorithms.classification.tasks import (
+from otx.algorithms.classification.tasks import (  # ClassificationOpenVINOTask,
```
GalyaZalesskaya, line 12: I believe, this comment was left by accident

GalyaZalesskaya (dismissed):
One minor comments, that can be fixed afterwards

tests/integration/api/xai/test_api_xai_sanity.py
negvet: Yeah, it is possible to remove it, but if you check below there is commented OV task. I will fix it later. So `# ClassificationOpenVINOTask` is kept for it.

sovrasov (approved):

GalyaZalesskaya (approved):

wonjuleee (approved):
LGTM

PR #1727, opened 2023-02-17
[FIX] Detection and Base adapter Hot-fix

sungmanc:
# Summary
* Previous the base dataset adapter weirdly imports the validation set. 
  * For the validation set, train+val+test datasets were included as validation dataset. 
  * --> Fixed by checking the subset name when import https://github.com/openvinotoolkit/training_extensions/compare/sungmanc/hotfix-detection-adapter?expand=1#diff-b38447408c6f8042dce39bb79e294c843264c17d406f61322e7ba8970c59a60aR148

* When there is abnormal annotation (e.g, width=0 or height=0 in bbox annotation), OTX raises error.
  * --> Fixed by checking the abnormal annotation
https://github.com/openvinotoolkit/training_extensions/compare/sungmanc/hotfix-detection-adapter?expand=1#diff-b38447408c6f8042dce39bb79e294c843264c17d406f61322e7ba8970c59a60aR204


otx/core/data/adapter/base_dataset_adapter.py
```diff
@@ -188,6 +201,9 @@ def _prepare_label_information(
 
         return {"category_items": category_items, "label_groups": label_groups, "label_entities": label_entities}
 
+    def _is_normal_bbox(self, x1: float, y1: float, x2: float, y2: float) -> bool:
+        return x1 != x2 and y1 != y2
```
jaegukhyun: How about x1 < x2 and y2 < y2

jaegukhyun (commented):
LGTM, I left a single comment

otx/core/data/adapter/base_dataset_adapter.py
sungmanc: Better idea! Edited.

jaegukhyun (approved):

goodsong81 (approved):
Nice catch.

PR #1734, opened 2023-02-17
[FIX]Fix coco to Datumaro multilabel converter

sovrasov:


JihwanEom (approved):

sungmanc (approved):
Nice catch ! Thanks, Vlad :)

PR #1737, opened 2023-02-18
[TEST] Unit test for otx/mpa/*.py

supersoob:
This PR includes unit test files for otx/mpa/*.py

image
image
image
image


harimkang (approved):
LGTM! Doesn't this require __init__.py?

samet-akcay:
> LGTM! Doesn't this require **init**.py?

@harimkang , `__init__.py` is required if there are tests share same test names. If the names are different, no need for `__init__.py`

JihwanEom (approved):

PR #1738, opened 2023-02-19
[FIX] Integer based sort for label.id_ in label_schema instead of string

JihwanEom:
## This PR includes:
- Sort `labels` in `label_schema` following `id_` attribute in `LabelEntity` **based on integer instead of string**. 
- Fix inference bug on action tasks when the number of class is higher than 10. (Occurred by string sorted `self._labels`. inference task is supposing integer sorted `self._labels`. (https://github.com/openvinotoolkit/training_extensions/blob/develop/otx/algorithms/action/tasks/inference.py#L382, https://github.com/openvinotoolkit/training_extensions/blob/develop/otx/algorithms/action/tasks/inference.py#L416)

## Examples:
Before: `self._labels = [LabelEntity(0), LabelEntity(1), LabelEntity(10), LabelEntity(11), ... LabelEntity(2)]`
After: `self._labels = [LabelEntity(0), LabelEntity(1), LabelEntity(2), LabelEntity(3), ... LabelEntity(10)]`


jaegukhyun (approved):

goodsong81 (approved):

sungmanc (approved):
Nice work !

PR #1739, opened 2023-02-19
[FIX] Keep `ote_config_helper` name for external backward compatibility

goodsong81:
- Keep `ote_config_helper` for Geti / Geti SDK
- Use replacy `otx_config_helper` use in OTX by `cfg_helper` (same module w/ diff name)


goodsong81:
CI might have some problem. Local tests are all green.
image

PR #1740, opened 2023-02-19
[ENHANCE] Support HMDB51/UCF101 converting to CVAT format for action classification

JihwanEom:
### Action classification
```
python3 otx/algorithms/action/utils/convert_public_data_to_cvat.py \
--task action_classification \
--src_path ./data/hmdb51/rawframes \
--dst_path ./data/hmdb51/CVAT/train \
--ann_file ./data/hmdb51/hmdb51_train_split_1_rawframes.txt \
--label_map ./data/hmdb51/label_map.txt
```

### Action detection
```
python3 otx/algorithms/action/utils/convert_public_data_to_cvat.py \
--task action_detection \
--src_path ./data/ava_dataset/frames \
--dst_path ./data/ava_dataset/CVAT/train \
--ann_file ./data/ava_dataset/annotations/train.csv
```

- [x] HMDB51 convert
- [x] UCF101 convert
- [x] Jester convert
- [x] AVA convert
- [x] Update unit test codes
image


otx/algorithms/action/utils/convert_public_data_to_cvat.py
```diff
@@ -302,18 +326,35 @@ def read_ava_csv(csv_path):
     return annot_info
 
 
-def main(src_path, dst_path, task):
+def parse_args():
```
harimkang, line 343: Why does this code take separate inputs?

JihwanEom: It's WIP, will be handled after @jaegukhyun's update.

harimkang:
@goodsong81 @sungmanc  Is it the right way to have a separate script like this? I think this should be supported at the CLI level?  Or is this code that should be in datumaro?

goodsong81:
> @goodsong81 @sungmanc Is it the right way to have a separate script like this? I think this should be supported at the CLI level? Or is this code that should be in datumaro?

Not the best way, but it's for OTX users, not for Datumaro users.
Until Datumaro could support multi-video formats directly, this is not-that-bad workaround for a while. @sungmanc @jaegukhyun  Please correct me, if any.

sungmanc:
> @goodsong81 @sungmanc Is it the right way to have a separate script like this? I think this should be supported at the CLI level? Or is this code that should be in datumaro?

As @goodsong81 said, this is a temporary solution to support the multi-video CVAT format.
While Datumaro could support the CVAT format, it couldn't support multi-video CVAT format required at our scenario.
So, as a temporary solution, we decided to support this kind of script. 

For the multi-label classification, we also provides a script for dataset converting (e.g, COCO --> multi-label COCO classification). This is also one of the temporary solution.

harimkang:
> > @goodsong81 @sungmanc Is it the right way to have a separate script like this? I think this should be supported at the CLI level? Or is this code that should be in datumaro?
> 
> As @goodsong81 said, this is a temporary solution to support the multi-video CVAT format. While Datumaro could support the CVAT format, it couldn't support multi-video CVAT format required at our scenario. So, as a temporary solution, we decided to support this kind of script.
> 
> For the multi-label classification, we also provides a script for dataset converting (e.g, COCO --> multi-label COCO classification). This is also one of the temporary solution.

Okay. So this is a temporary solution for now and this is code that will be fixed in a different way, right? I understand.

sungmanc:
> > > @goodsong81 @sungmanc Is it the right way to have a separate script like this? I think this should be supported at the CLI level? Or is this code that should be in datumaro?
> > 
> > 
> > As @goodsong81 said, this is a temporary solution to support the multi-video CVAT format. While Datumaro could support the CVAT format, it couldn't support multi-video CVAT format required at our scenario. So, as a temporary solution, we decided to support this kind of script.
> > For the multi-label classification, we also provides a script for dataset converting (e.g, COCO --> multi-label COCO classification). This is also one of the temporary solution.
> 
> Okay. So this is a temporary solution for now and this is code that will be fixed in a different way, right? I understand.

100% right.

jaegukhyun (approved):

goodsong81 (approved):

PR #1748, opened 2023-02-20
[TEST] ov unit test

cih9088:
Note that not all files are covered although the coverage is 88%
### Results
image

### DIFF
--- a/otx/cli/builder/supported_backbone/omz.mmcls.json
+++ b/otx/cli/builder/supported_backbone/omz.mmcls.json
@@ -1,9 +1,9 @@
 {
   "backbones": {
     "mmcls.MMOVBackbone": {
-      "required": ["model_path"],
+      "required": ["model_path_or_model"],
       "options": {
-        "model_path": [
+        "model_path_or_model": [
           "omz://alexnet",
           "omz://caffenet",
           "omz://densenet-121",
--- a/otx/mpa/modules/ov/models/mmov_model.py
+++ b/otx/mpa/modules/ov/models/mmov_model.py
@@ -4,6 +4,7 @@
 
 from typing import Dict, List, Optional, Union
 
+import openvino.runtime as ov
 import torch
 
 from .ov_model import OVModel
@@ -13,7 +14,7 @@ from .parser_mixin import ParserMixin
 class MMOVModel(OVModel, ParserMixin):
     def __init__(
         self,
-        model_path: str,
+        model_path_or_model: Union[str, ov.Model],
         weight_path: Optional[str] = None,
         inputs: Optional[Union[Dict[str, Union[str, List[str]]], List[str], str]] = None,
         outputs: Optional[Union[Dict[str, Union[str, List[str]]], List[str], str]] = None,
@@ -23,7 +24,7 @@ class MMOVModel(OVModel, ParserMixin):
         parser = kwargs.pop("parser", None)
         parser_kwargs = kwargs.pop("parser_kwargs", {})
         inputs, outputs = super().parse(
-            model_path=model_path,
+            model_path_or_model=model_path_or_model,
             weight_path=weight_path,
             inputs=inputs,
             outputs=outputs,
@@ -32,7 +33,7 @@ class MMOVModel(OVModel, ParserMixin):
         )
 
         super().__init__(
-            model_path=model_path,
+            model_path_or_model=model_path_or_model,
             weight_path=weight_path,
             inputs=inputs,
             outputs=outputs,
--- a/otx/mpa/modules/ov/omz_wrapper.py
+++ b/otx/mpa/modules/ov/omz_wrapper.py
@@ -7,6 +7,7 @@ import os
 import shutil
 import string
 import sys
+import time
 from pathlib import Path
 
 import requests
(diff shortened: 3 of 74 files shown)


otx/mpa/modules/ov/models/mmov_model.py
```diff
@@ -13,7 +14,7 @@
 class MMOVModel(OVModel, ParserMixin):
     def __init__(
         self,
-        model_path: str,
+        model_path_or_model: Union[str, ov.Model],
```
goodsong81, line 17: Reason for this change?
Type annotations are missing in other similar changed lines.

otx/mpa/modules/ov/omz_wrapper.py
```diff
@@ -163,6 +155,23 @@ def update_model(model):
     model.subdirectory_ori = model.subdirectory
     model.subdirectory = Path(m.hexdigest())
 
+    # FIXME: a bug from openvino-dev==2022.3.0
+    # It has been fixed on master branch.
+    # After upgrading openvino-dev, we can remove this temporary patch
+    if getattr(model, "conversion_to_onnx_args") and not [
+        arg for arg in model.conversion_to_onnx_args if arg.startswith("--model-path")
+    ]:
+        model.conversion_to_onnx_args.append("--model-path=")
```
goodsong81, line 164: Can you explain a bit about the bug and fix?

harimkang (dismissed):
You have worked so hard! I mainly looked at the changes in `MMOVBackbone`, but there seems to be no problem because there is almost no change in replacing the backbone.

@cih9088 As I remember, the actual omz pre-trained weights are obtained from somewhere other than the init_weights function, right?

supersoob (dismissed):
Thanks for your hard work.

goodsong81 (changes requested):
Great coverage improvement. Thank you so much for your contribution.
BTW, there is a integration test failure regarding model build.
@cih9088 @harimkang Could you guys handle the error?
https://github.com/openvinotoolkit/training_extensions/actions/runs/4220949816/jobs/7328806452#step:4:31539

harimkang:
> Great coverage improvement. Thank you so much for your contribution.
> BTW, there is a integration test failure regarding model build.
> @cih9088 @harimkang Could you guys handle the error?
> https://github.com/openvinotoolkit/training_extensions/actions/runs/4220949816/jobs/7328806452#step:4:31539

@cih9088 I think the error currently occurring is because the variable name of the parameter has been changed. 
could you please update this(https://github.com/openvinotoolkit/training_extensions/blob/develop/otx/cli/builder/supported_backbone/omz.mmcls.json)? Older versions are looking for a parameter called model_path.

otx/mpa/modules/ov/omz_wrapper.py
cih9088: OMZ internally converts a pytorch model to onnx with [pytorch_to_onnx.py](https://github.com/openvinotoolkit/open_model_zoo/blob/master/tools/model_tools/src/openvino/model_zoo/internal_scripts/pytorch_to_onnx.py) script. If a pytorch model is not defined in pre-installed libraries, such as torchvision, one needs a separate model definition file to load weights like this.
```
import CustomModel
model = CustomModel()
model.load_state_dict(state_dict)
```
`--model-path` argument is to indicate where `CustomModel` is defined and it is missing if a model definition is in pre-installed libraries. This missing value converted to `None` value and it raises errors in [here](https://github.com/openvinotoolkit/open_model_zoo/blob/ef62bb332b1d5eb59f71f4af057c0bc897bc21dc/tools/model_tools/src/openvino/model_zoo/internal_scripts/pytorch_to_onnx.py#L49) which it tries to concatenate `None` with `List[str]`. It has been fixed by https://github.com/openvinotoolkit/open_model_zoo/pull/3639. 

The workaround here is forcing to add argument `--model-path` to convert the missing value to empty list `[]` so that the concatenation is like this `[] + List[str]`.

otx/mpa/modules/ov/models/mmov_model.py
cih9088: Thank you for missing annotations. 
This argument can now accept `openvino.runtime.Model` object not just `xml` path.

goodsong81 (dismissed):

otx/cli/builder/supported_backbone/omz.mmcls.json
```diff
@@ -1,7 +1,7 @@
 {
   "backbones": {
     "mmcls.MMOVBackbone": {
-      "required": ["model_path"],
+      "required": ["model_path_or_model"],
       "options": {
         "model_path": [
```
harimkang: Could you also fix the `model_path` here as well?

wonjuleee (dismissed):
Looks great!

otx/cli/builder/supported_backbone/omz.mmcls.json
cih9088: This has been fixed in https://github.com/openvinotoolkit/training_extensions/pull/1748/commits/29ce2402ad29ffefced8f110add51b82498a41a3 already. Are you referring something else?

harimkang: What I said is line6. You need to change line6 as well. :)

harimkang:
@cih9088 @goodsong81 If only my comment is resolved, all other tests are passing. If that is resolved, I think this can be merged. I think we only need to check `tests/integration/cli/test_cli.py`.

goodsong81:
> @cih9088 @goodsong81 If only my comment is resolved, all other tests are passing. If that is resolved, I think this can be merged. I think we only need to check `tests/integration/cli/test_cli.py`.

Sure, I would merge right away once verified. Please attach a screen shot after fix. :)

goodsong81 (approved):

cih9088:
@goodsong81 @harimkang 
I can confirm tests are passed now with `pytest tests/integration/cli/test_cli.py`.
image

harimkang (approved):

PR #1750, opened 2023-02-20
[FIX] Resolved an issue with rebuilding on backbone replacement

harimkang:
## Summary
Avoid unnecessary rebuilds when replacing the backbone


sungmanc (approved):
LGTM

JihwanEom (approved):

PR #1757, opened 2023-02-20
[WIP] Emily/doc otx1.0

chuneuny-emily:


docs/source/guide/get_started/introduction.rst
```diff
@@ -0,0 +1,27 @@
+Introduction
```
kprokofi, line 1: I prefer to move it to the main index.rst before the table of content + add logo -> it will be the first page that users will see

PR #1696, opened 2023-02-14
[ENHANCE] Use FastBiasCorrection for OCR_HRNET_18_MOD2

daniil-lyakhov:
Speeds up model calibration from ~12h to 20 mins without significant acc drop
 
 Results on V0.5.0 with appropriate IS (dataset - kvasir_seg)

 # | FP32 | INT8 | diff
-- | -- | -- | --
Metric | 0.86799 | 0.8643 | 0.00369

PR #1758, opened 2023-02-20
HRNet IS for POT update

daniil-lyakhov:
HRNets now using dynamic shapes in OpenvinoIR, so names was being changed


otx/algorithms/segmentation/configs/ocr_lite_hrnet_18_mod2/pot_optimization_config.json
```diff
@@ -5,90 +5,89 @@
       "params": {
         "preset": "mixed",
         "target_device": "ANY",
-        "use_fast_bias": false,
```
goodsong81: As I remember, the fast bias was disabled recently. Is this re-enable intentional?

goodsong81 (dismissed):
Thank you so mush for the fix, and the script, too!

otx/algorithms/segmentation/configs/ocr_lite_hrnet_18_mod2/pot_optimization_config.json
daniil-lyakhov: That is strange because we were enabling `use_fast_bias` by removing this attr from config. And there is no `use_fast_bias` in  develop https://github.com/openvinotoolkit/training_extensions/blob/develop/otx/algorithms/segmentation/configs/ocr_lite_hrnet_18_mod2/pot_optimization_config.json

This diff looks broken, I already removed this line. Maybe I made previous PR (https://github.com/openvinotoolkit/training_extensions/pull/1696) in a wrong branch?

daniil-lyakhov: Ahh, I didn't rebase it, one moment

goodsong81 (approved):

sovrasov (approved):

PR #1759, opened 2023-02-20
[FIX] Change self-sl for cls batch size from 256 to 64

sungchul1:
batch size=256 was set for experiments with CIFAR datasets (input size=32), but when it is applied to datasets with input size=224 OOM occurs.
In this PR, batch size in self-sl recipes is set from 256 to 64.


JihwanEom:
Could you let us know how much memory was needed when we use 256 batch size? I'm curious about that :)

sungmanc:
> Could you let us know how much memory was needed when we use 256 batch size? I'm curious about that :)

Actually, we couldn't measure the memory when we used 256 batch . 
When I used 64 batch, the memory usage as below:
image

sungmanc (approved):
Thanks for quick fix

JihwanEom (approved):

PR #1761, opened 2023-02-20
[Tests] Add tests for losses

ashwinvaidya17:
# Description

- Adds tests for cross focal loss, l2sp loss and utils
- I didn't add cross entropy loss with ignore as I didn't understand it very well. It depends on base pixel loss and mpa pixel base so these aren't being tested as well.


sungmanc (approved):
Thanks a lot ! Such a incredible coverage improvement !

sovrasov (approved):

PR #1772, opened 2023-02-21
[FIX] Enable Semi-SL + Auto-split

sungmanc:
# Summary

## Problems
* This issue was reported by @supersoob 
* If an user inputs the `--train-data-roots`, `--unlabeled-data-roots` , and `params --algobackend.train_type "SEMISUPERVISED"`,
`unlabeled_data_root` written in `data.yaml` is `null` although the user explicitly inputs `--unlabeled-data-roots`.

``` bash
otx train otx/algorithms/classification/configs/efficientnet_b0_cls_incr/template.yaml --train-data-roots tests/assets/imagenet_dataset --unlabeled-data-roots tests/assets/imagenet_dataset_class_incremental params --algo_backend.train_type SEMISUPERVISED
```

* When auto-split is enabled, there is no logic to check the `--unlabeled-data-roots`. So, in this case, `--unlabeled-data-roots` written in `data.yaml` is set to `null`.

## Solutions
* If the user gives `--unlabeled-data-roots` argument, it will be added into data.yaml with absolute value.
* After merging this PR, we could expect data.yaml as like below:
image


supersoob (dismissed):
Thank you very much :)

supersoob (approved):

goodsong81 (approved):

PR #1798, opened 2023-02-22
[DOC] HPO documentation

eunwoosh:
PR for HPO documentation

### DIFF
--- a/docs/source/guide/explanation/additional_features/hpo.rst
+++ b/docs/source/guide/explanation/additional_features/hpo.rst
@@ -22,10 +22,31 @@ Simpe HPO Exapmle
 This exapmle provides a step-by-step guide on how to use Hyper-Parameter Optimization (HPO) for classification tasks. In this example, we will optimize the learning rate and batch size using HPO.
 
 =========================
-1. Set hpo_config.yaml
+1. Build workspace
 =========================
 
-Before running HPO, you can configure HPO using the hpo_config.yaml file. This file contains all the information that the HPO module needs, including the hyperparameters that you want to optimize. The file is located in the same directory as the `template.yaml` file and comes with default values.
+First, let's build a workspace. You can do this by running the following command:
+
+.. code-block::
+
+    (otx) ...$ otx build --train-data-roots data/flower_photos --model MobileNet-V3-large-1x
+
+    [*] Load Model Template ID: Custom_Image_Classification_MobileNet-V3-large-1x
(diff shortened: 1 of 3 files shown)


jaegukhyun (commented):
Overall explanation for HPO contents look good to me, however I prefer to explain overall hpo configure steps after using otx build. What do you think?

sungmanc:
> Overall explanation for HPO contents look good to me, however I prefer to explain overall hpo configure steps after using otx build. What do you think?

Agree with this opinion, however, we can add the detailed explanation in another PR. So, I prefer to merge like as is

sungmanc (dismissed):

eunwoosh:
Thanks for comments @jaegukhyun , @sungmanc . I added that step in docs. please check that :)

docs/source/guide/explanation/additional_features/hpo.rst
```diff
@@ -1,4 +1,184 @@
...
+Key features of OpenVINO™ Training Extensions include:
+
+- **Easy usability** : By using time as the control parameter, OpenVINO™ Training Extensions offers a straightforward and intuitive interface for users.
+
+- **Auto-config** : The automatic configuration feature sets internal control parameters automatically, ensuring that HPO finishes within the given time constraint.
+
+- **Scalability** : OpenVINO™ Training Extensions offers both sequential and parallel methods, making it scalable for different training environments. If you have multiple GPUs, you can take advantage of all available GPU resources to accelerate HPO.
+
+**************
+Simpe HPO Exapmle
```
kprokofi: It should be located in tutorials/advanced/
The explanation section is only for describing algorithm

1) Please, add description about what algorithm is used for HPO with link to articles and briefly how it works
2) Add overview of the HPO config and what parameters can be optimized by HPO and how to set them (how to set time limits, lower and higher boundaries for parameters, etc.)
3) Add just one line general example how to enable it here

Everything else should be moved to tutorials

kprokofi (changes requested):

docs/source/guide/explanation/additional_features/hpo.rst
eunwoosh: I applied your comment. Could you check it?

kprokofi (approved):
Thank you very much! Such a deep explanation section for HPO and good tutorial as well

harimkang (approved):

PR #1811, opened 2023-02-23
[FIX] Add selfsl condition in eval

sungchul1:
## Issue
Although `train_type == SELFSUPERVISED`, if `--val-data-root` is set, evaluation will work.
But, selfsl doesn't need evaluation.

## Change
Add the condition checking `train_type` to avoid evaluation during selfsl.

### DIFF
--- a/otx/cli/tools/train.py
+++ b/otx/cli/tools/train.py
@@ -210,8 +210,10 @@ def main():  # pylint: disable=too-many-branches
     save_model_data(output_model, args.save_model_to)
     print(f"[*] Save Model to: {args.save_model_to}")
 
-    if config_manager.data_config["val_subset"]["data_root"] and \
-        hyper_parameters.algo_backend.train_type.value != "SELFSUPERVISED":
+    if (
+        config_manager.data_config["val_subset"]["data_root"]
+        and hyper_parameters.algo_backend.train_type.value != "SELFSUPERVISED"
+    ):
         validation_dataset = dataset.get_subset(Subset.VALIDATION)
         predicted_validation_dataset = task.infer(
             validation_dataset.with_empty_annotations(),
--- a/otx/cli/tools/train.py
+++ b/otx/cli/tools/train.py
@@ -210,7 +210,8 @@ def main():  # pylint: disable=too-many-branches
     save_model_data(output_model, args.save_model_to)
     print(f"[*] Save Model to: {args.save_model_to}")
 
-    if config_manager.data_config["val_subset"]["data_root"]:
+    if config_manager.data_config["val_subset"]["data_root"] and \
+        hyper_parameters.algo_backend.train_type.value != "SELFSUPERVISED":
         validation_dataset = dataset.get_subset(Subset.VALIDATION)
         predicted_validation_dataset = task.infer(
             validation_dataset.with_empty_annotations(),


jaegukhyun (dismissed):
1. Do we need to change unit-test?
2. Should this PR head head to release branch?

sungchul1:
> 1. Do we need to change unit-test?
> 2. Should this PR head head to release branch?

There is no unit test for evaluation part, but if needed, I'll add it.
And I'm not sure which branch is proper. Please give me a suggestion.

otx/cli/tools/train.py
```diff
@@ -210,7 +210,10 @@ def main():  # pylint: disable=too-many-branches
     save_model_data(output_model, args.save_model_to)
     print(f"[*] Save Model to: {args.save_model_to}")
 
-    if config_manager.data_config["val_subset"]["data_root"]:
+    if (
+        config_manager.data_config["val_subset"]["data_root"]
+        and hyper_parameters.algo_backend.train_type.value != "SELFSUPERVISED"
```
harimkang, line 215: Is there any case where `hyper_parameters.algo_backend.train_type.value` is not read properly? 
and
Is there any other way to give this as None inside the `ConfigManager`? I'm confused about the case of giving `val-data-roots` in self-sl.

sungmanc:
Does this item urgent? If not, how about closing this PR? we're going to refactor overall DatasetManager to improve the stability of Auto-Config and Dataset Argument.

+ Adding special `if` logic in the `train.py` for only `self-sl` is not desirable. All exception occurred by specific data, train, task type should be handled by `DatasetManager` or `ConfigManager`

otx/cli/tools/train.py
sungchul1: If there is any other way to set None to `config_manager.data_config["val_subset"]["data_root"]` during self-sl, it doesn't matter.
If there is the demand to use data.yaml created from `otx build` of supervised learning, there is `val-data-root` and it will be set. Self-sl doesn't use it, but unnecessary evaluation after training will be conducted.

sungchul1:
> Does this item urgent? If not, how about closing this PR? we're going to refactor overall DatasetManager to improve the stability of Auto-Config and Dataset Argument.
> 
> * Adding special `if` logic in the `train.py` for only `self-sl` is not desirable. All exception occurred by specific data, train, task type should be handled by `DatasetManager` or `ConfigManager`

It is not critical. If there is any way to resolve it in `DatasetManager` or `ConfigManager`, this PR can be closed.

sungmanc:
> > Does this item urgent? If not, how about closing this PR? we're going to refactor overall DatasetManager to improve the stability of Auto-Config and Dataset Argument.
> > 
> > * Adding special `if` logic in the `train.py` for only `self-sl` is not desirable. All exception occurred by specific data, train, task type should be handled by `DatasetManager` or `ConfigManager`
> 
> It is not critical. If there is any way to resolve it in `DatasetManager` or `ConfigManager`, this PR can be closed.

Thanks for quick reply, we're going to cover this problem by changing the `DatasetManager` or `ConfigManager`.
For maintaining the simplicity of `train.py`, we don't want to add specific condition in the CLI (train, eval, export, ...).py.

PR #1813, opened 2023-02-23
[FIX] nncf statistic error

cih9088:
### Summary
It makes sure to take the onnx branch when extracting ROI for two stage model series.

### Jira ticket
[CVS-104503](https://jira.devtools.intel.com/browse/CVS-104503)

### DIFF
--- a/otx/algorithms/detection/adapters/mmdet/nncf/patches.py
+++ b/otx/algorithms/detection/adapters/mmdet/nncf/patches.py
@@ -21,6 +21,7 @@ from mmdet.models.roi_heads.mask_heads.fcn_mask_head import FCNMaskHead
 
 from otx.algorithms.common.adapters.nncf import (
     NNCF_PATCHER,
+    is_in_nncf_tracing,
     nncf_trace_wrapper,
     no_nncf_trace_wrapper,
 )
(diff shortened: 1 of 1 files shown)


sungmanc:
To make sure, 
Could this problem also be occurred in any `Twostage Architecture (i.e. Faster R-CNN, ...)`?
Could this patch prevent these problems ?

IMO, maybe the answer is right, isn't it?

sungmanc (dismissed):

harimkang (dismissed):
I have confirmed that the issue in the ticket is not reproducible, thank you.

harimkang:
BTW, this PR seems like a good idea to change and merged right into the release branch, do you agree? @goodsong81 @sungmanc

cih9088:
> To make sure, Could this problem also be occurred in any `Twostage Architecture (i.e. Faster R-CNN, ...)`? Could this patch prevent these problems ?
> 
> IMO, maybe the answer is right, isn't it?

I haven't tested them but if they used `SingleRoIExtracotr`, the answer would be yes.

sungmanc:
> BTW, this PR seems like a good idea to change and merged right into the release branch, do you agree? @goodsong81 @sungmanc

Exactly, we should do. @cih9088  could you change the target branch as releases/v1.0.0?

cih9088:
> BTW, this PR seems like a good idea to change and merged right into the release branch, do you agree?

Do you want me to change the target branch to v1.0.0?

sungmanc:
> > To make sure, Could this problem also be occurred in any `Twostage Architecture (i.e. Faster R-CNN, ...)`? Could this patch prevent these problems ?
> > IMO, maybe the answer is right, isn't it?
> 
> I haven't tested them but if they used `SingleRoIExtracotr`, the answer would be yes.

Thanks for giving the nice remedy

sungmanc (approved):

harimkang (approved):

goodsong81 (approved):

PR #1824, opened 2023-02-23
[FIX] Classification Auto-split bug fix

sungmanc:
# Summary
* When users follow the classification tutorial, auto-split will not properly works. So, train / val datasets will be the same. reported by @sungchul2 . This bug is fixed.
* Change unit-tests according to the fix
* Changed figure in the classification tutorial: to show the classification dataset
image

* Add ignore for detcon_mask made by unit-test


harimkang (dismissed):

sungchul1 (dismissed):

eunwoosh (dismissed):
LGTM

sungchul1 (dismissed):

supersoob (approved):

docs/source/guide/tutorials/base/how_to_train/classification.rst
```diff
@@ -121,7 +120,11 @@ Let's prepare an OpenVINO™ Training Extensions classification workspace runnin
 
   (otx) ...$ cd ./otx-workspace-CLASSIFICATION
 
+<<<<<<< HEAD
+It will create **otx-workspace-CLASSIFICATION** with all necessery configs for MobileNet-V3-large-1x, prepared ``data.yaml`` to simplify CLI commands launch and splitted dataset named ``splitted_dataset``.
+=======
```
supersoob: Could you fix??

supersoob (changes requested):
you need to remove HEAD from merge conflict

sungmanc:
> you need to remove HEAD from merge conflict

Thanks very much, Done

supersoob (dismissed):

eunwoosh (dismissed):

sungchul1 (dismissed):

harimkang:
I can't figure out the cause from the log alone, but there seems to be something wrong with the changes.
maybe validation dataset..? (Fail from task.infer)
스크린샷 2023-02-24 오후 9 01 46

otx/core/data/manager/dataset_manager.py
```diff
@@ -31,12 +31,12 @@ def get_train_dataset(dataset: Dataset) -> DatasetSubset:
...
     @staticmethod
-    def get_val_dataset(dataset: Dataset) -> DatasetSubset:
+    def get_val_dataset(dataset: Dataset) -> Union[DatasetSubset, None]:
         """Returns validation dataset."""
         for k, v in dataset.subsets().items():
-            if "val" in k or "default" in k:
+            if "val" in k:
                 return v
-        raise ValueError("Can't find validation data.")
+        return None
```
harimkang, line 39: return None -> It may be the cause of the current issue..?

sungmanc: Thanks for investigation, I'll check it out

kprokofi: we may delete return None since all procedures in Python return None by default

sungmanc: Done, I deleted this line

sungmanc:
@harimkang , I found out the root cause of the FAIL. It is caused by wrong auto-split.
Since the number of images in tests/assets/imagenet_dataset is too small, the splitted image is not normal like below.
Train and val datasets have different classes.
image


So, I added the number of data, and then issue was resolved. 
Actually, since the previous implementation of auto-config is weird, we couldn't find any problem in integration tests.. :(

otx/core/data/manager/dataset_manager.py
sungmanc: @kprokofi , no return is prohibited by mypy.. So, I left as is

kprokofi (approved):

goodsong81 (approved):

PR #1834, opened 2023-02-27
[FIX][Release1.0] Fix action task sample codes and unit tests

jaegukhyun:
This PR includes hot fix for runtime errors in action tasks sample codes
**Main Cause**
Changed dataset adapter requires train_type(INCREMENTAL, SEMSUPERVISED, etc). However according changes in action sample code are invalid.

**Solution**
Add new codes for valid train_type searching in model template class.

**Changes**
1. Change sample codes to search train_type properly.
2. Change unit test codes to reflect otx side changes.

### DIFF
--- a/otx/algorithms/action/tools/sample_classification.py
+++ b/otx/algorithms/action/tools/sample_classification.py
@@ -49,9 +49,12 @@ VAL_DATA_ROOTS = "tests/assets/cvat_dataset/action_classification/train"
 
 def load_test_dataset(model_template):
     """Load Sample dataset for detection."""
+
+    algo_backend = model_template.hyper_parameters.parameter_overrides["algo_backend"]
+    train_type = algo_backend["train_type"]["default_value"]
     dataset_adapter = get_dataset_adapter(
         model_template.task_type,
-        model_template.train_type,
+        train_type,
         train_data_roots=TRAIN_DATA_ROOTS,
         val_data_roots=VAL_DATA_ROOTS,
     )
--- a/otx/algorithms/action/tools/sample_detection.py
+++ b/otx/algorithms/action/tools/sample_detection.py
@@ -46,9 +46,11 @@ VAL_DATA_ROOTS = "tests/assets/cvat_dataset/action_detection/train"
 
 def load_test_dataset(model_template):
     """Load Sample dataset for detection."""
+    algo_backend = model_template.hyper_parameters.parameter_overrides["algo_backend"]
+    train_type = algo_backend["train_type"]["default_value"]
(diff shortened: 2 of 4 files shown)


goodsong81 (approved):
Thank you for the quick fix.

harimkang (approved):

otx/algorithms/action/tools/sample_classification.py
```diff
@@ -49,9 +49,12 @@ def parse_args():
 
 def load_test_dataset(model_template):
     """Load Sample dataset for detection."""
+
+    algo_backend = model_template.hyper_parameters.parameter_overrides["algo_backend"]
+    train_type = algo_backend["train_type"]["default_value"]
     dataset_adapter = get_dataset_adapter(
         model_template.task_type,
-        model_template.train_type,
+        train_type,
```
eunwoosh, line 57: As far as I understand, this `parameter_overrides` value is merged after `params = create(model_template.hyper_parameters.data)`.
So, I think it's better to get `params` as an argument for generality. How do you think about?

jaegukhyun: So do you suggest to require params argument in the sample code? or do you suggest to change the way to get train_type from model template?

eunwoosh: My intention is passing params (made from otx/algorithms/action/tools/sample_classification.py#79) to `load_test_dataset()` So it needs to move `dataset, labels_schema = load_test_dataset(model_template)` after `params = create(model_template.hyper_parameters.data)` and change it to `dataset, labels_schema = load_test_dataset(params)`

jaegukhyun: I got it. In that case, can we get task_type from params?

eunwoosh: hmm.. I think `params` doesn't have that information. it may need one more argument for `task_type`.

jaegukhyun: Then I think passing model_template looks simple and appropriate.

eunwoosh: My intention is just considering some exceptional case where `algo_backend` or `train_type` isn't in `template.yaml` but in `configuration.yaml` (I know it isn't currently, but I just consider changes in the future). I know it's a just sample code. So, if you think it doesn't matter to consider that, then I'll follow with your opinion :)

jaegukhyun: Yeah, thank you for your kind review. I think train_type will be existed for a long time. I'll keep this in my mind.

PR #1839, opened 2023-02-27
[HOTFIX][RELEASE1.0] Fix missing classes in the first incremental round & cleanup classification utils

sovrasov:


goodsong81 (approved):
Thank you for the fix. Working well on Geti at least for SL scenario.

PR #1841, opened 2023-02-28
[HOTFIX][RELEASE1.0] Fixed label_scheme mismatch in classification

harimkang:
## Summary
Fix classification label schema mismatch

A problem occurs because the class setting in incremental training and the class setting in the exporter are different. (Occurs when keep using the same Task object, can see it in the sample code.)

However, I'm not familiar with the `mpa/cls` code, so please take a look at @JihwanEom  and @supersoob  even after the merge.

### DIFF
--- a/otx/mpa/cls/incremental/stage.py
+++ b/otx/mpa/cls/incremental/stage.py
@@ -4,7 +4,7 @@
 
 from mmcv import ConfigDict
 
-from otx.mpa.cls.stage import ClsStage, Stage
+from otx.mpa.cls.stage import ClsStage
 from otx.mpa.utils.config_utils import update_or_add_custom_hook
 from otx.mpa.utils.logger import get_logger
 
@@ -35,9 +35,7 @@ class IncrClsStage(ClsStage):
     # noqa: C901
     def configure_task_adapt(self, cfg, training, **kwargs):
         """Configure for Task Adaptation Task"""
-
-        self.adapt_type = cfg["task_adapt"].get("op", "REPLACE")
-        train_data_cfg = Stage.get_data_cfg(cfg, "train")
+        train_data_cfg = self.get_data_cfg(cfg, "train")
         if training:
             if train_data_cfg.type not in CLASS_INC_DATASET:
                 logger.warning(f"Class Incremental Learning for {train_data_cfg.type} is not yet supported!")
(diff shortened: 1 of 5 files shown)


otx/mpa/cls/incremental/stage.py
```diff
@@ -35,35 +35,27 @@ def configure_task(self, cfg, training, **kwargs):
...
-        train_data_cfg = Stage.get_data_cfg(cfg, "train")
+        train_data_cfg = self.get_data_cfg(cfg, "train")
         if training:
             if train_data_cfg.type not in CLASS_INC_DATASET:
                 logger.warning(f"Class Incremental Learning for {train_data_cfg.type} is not yet supported!")
             if "new_classes" not in train_data_cfg:
                 logger.warning('"new_classes" should be defined for incremental learning w/ current model.')
 
             if cfg.model.type in WEIGHT_MIX_CLASSIFIER:
+                cfg.task_adapt.final = self.model_classes
```
eunwoosh: I think `cfg.task_adapt.final` isn't used anywhere. How about deleting this line?

otx/mpa/cls/stage.py
```diff
@@ -143,25 +143,28 @@ def configure_task(self, cfg, training, **kwargs):
...
+            model_classes = org_model_classes + [cls for cls in data_classes if cls not in org_model_classes]
+        else:
+            raise KeyError(f"{self.task_adapt_op} is not supported for task_adapt options!")
+
+        # Model architecture
+        cfg.model.head.num_classes = len(model_classes)
+
+        self.org_model_classes = org_model_classes
         self.model_classes = model_classes
         self.data_classes = data_classes
```
eunwoosh: As above, this line isn't used anywhere, I think. How about deleting it?

eunwoosh (commented):
LGTM in general. I submitted a few tiny comments. please take a look.

otx/mpa/cls/incremental/stage.py
harimkang: Done

otx/mpa/cls/stage.py
harimkang: Done

eunwoosh (approved):

goodsong81 (approved):
Nice catch!

PR #1842, opened 2023-02-28
[FEATURE] Add embedding of inference configuration to IR for classification

sovrasov:
- config.json is now embedded into IR xml
- exportable code tries to use the config from IR first, then, in case of fail, uses config.json


sovrasov:
@Wovchena @AlexKoff88 please, have a look at these changes. If it's ok, we can extend to other tasks.

sovrasov:
@goodsong81 in the requirements of exportable code we have a pypi version of otx1.0 now, but I guess this change leaked from release branch. Can we switch otx version back to a github commit?

otx/algorithms/detection/utils/utils.py
```diff
@@ -98,3 +99,26 @@ def generate_label_schema(label_names: Sequence[str], label_domain: Domain = Dom
     label_schema.add_group(exclusive_group)
     label_schema.add_group(empty_group)
     return label_schema
+
+
+def get_det_model_api_configuration(label_schema: LabelSchemaEntity, task_type: TaskType, confidence_threshold: float):
+    """Get ModelAPI config."""
+    omz_config = {}
+    if task_type == TaskType.DETECTION:
+        omz_config[("model_info", "model_type")] = "ssd"
```
AlexKoff88, line 108: How about YoloX which is supported in Model API?

sovrasov: We don't have a standalone inferencer for yolox: `OpenVINODetectionInferencer` and `OTX_SSD` are used for `TaskType.DETECTION`. I can make a workaround here and somehow check if the input model is yolox , but in otx inferencer logic it is not required.

otx/algorithms/detection/utils/utils.py
```diff
@@ -98,3 +99,26 @@ def generate_label_schema(label_names: Sequence[str], label_domain: Domain = Dom
...
+    if task_type == TaskType.INSTANCE_SEGMENTATION:
+        omz_config[("model_info", "model_type")] = "instance_segmentation"
```
AlexKoff88: The current version of Model API supports "MaskRCNN" and "Yolact"  model types for instance segmentation.

AlexKoff88 (dismissed):

goodsong81:
> @goodsong81 in the requirements of exportable code we have a pypi version of otx1.0 now, but I guess this change leaked from release branch. Can we switch otx version back to a github commit?

Actually, it was intentional. If we have some update, you can use commit based install for sure (until release point :))

otx/api/usecases/exportable_code/demo/demo_package/model_container.py
```diff
@@ -28,7 +29,14 @@ class ModelContainer:
...
     def __init__(self, model_dir: Path, device="CPU") -> None:
-        self.parameters = get_parameters(model_dir / "config.json")
+        model_adapter = OpenvinoAdapter(create_core(), get_model_path(model_dir / "model.xml"), device=device)
+
+        try:
+            config_data = model_adapter.model.get_rt_info(["otx_config"])
+            self.parameters = json.loads(config_data)
+        except RuntimeError:
+            self.parameters = get_parameters(model_dir / "config.json")
+
```
goodsong81, line 39: Could you check if this block could guarantee the backward compatibility for models exported from OTE0.x (which is being used in Geti1.0~3)?
cc: @eugene123tw

yunchu:
@sovrasov please set the target milestone for this PR. I guess it would be 1.1.0.

otx/api/usecases/exportable_code/demo/demo_package/model_container.py
sovrasov: I checked that openvino throws exactly this error if rt info is missing. Also, this new field ("otx_config") exists only for classification now, so exportable code tests for other components would fails in case of incompatibility. Let's see the build results.

AlexKoff88 (dismissed):

otx/api/usecases/exportable_code/demo/demo_package/model_container.py
sovrasov: @goodsong81 the tests were passed, indicating that IRs without embedded config don't crash the exportable code demos.

goodsong81: @eugene123tw What was the BW-compatibility strategy for models even before the "config.json" era?

@sovrasov Could you check the last commit of `releases/v0.5.0`? We may need one more safeguard. Could you apply it if it's missing here & there for completeness?

sovrasov: @goodsong81 the fix you mentioned was already applied in detection openvino task and it's not related to this PR:
https://github.com/openvinotoolkit/training_extensions/blob/develop/otx/algorithms/detection/tasks/openvino.py#L333
https://github.com/openvinotoolkit/training_extensions/commit/7dc7b8b42874b2a5df50a81a849381906cc17ee7

goodsong81 (dismissed):
Looks OK to me in general. Thank you for update!

goodsong81:
@eugene123tw Could you have a look?

AlexKoff88:
Any ETA on the merge?

goodsong81 (approved):
LGTM, Thanks!

PR #1851, opened 2023-03-03
[FIX] Fix label mismatch of evaluation and validation with large dataset in semantic segmentation

supersoob:
**Issue**
(1) 'otx eval' (inference step & final evaluation score after training) shows us very low score for models trained with dataset with multiple classes(>10)
(2) The order of classes written in validation log is wrong with dataset with multiple classes(>10)
(3) **[TODO]** Semantic segmentation always produces lower evaluation score compared to the best validation score -> (Found solution but need some discussion)

**Root Causes**
(1)(2) : When converting segmentation mask to otx 2d numpy array(mapping to class index), it str-sorts the labels with id key. https://github.com/openvinotoolkit/training_extensions/blob/6cba8c9f42ef55171f41fc6e6c63d3367acf730e/otx/api/utils/segmentation_utils.py#L70
image
Due to this, when num_classes is above 10, the order is newly made like this. That caused mismatch with unsorted label schema.json which is saved in order of initial dataset_meta.json. Plus, it caused mismatch in actual prediction order from model head output.

(3) : Two factors in this issue. Background ignored and soft_threshold. The final evaluation score does not include the background label score, which never be the same with the best validation score. soft_threshold is set to 0.5 in default. It ignores the max score in prediction below the soft_threshold. But in eval hook(validation), soft_threshold is not considered which is same as threshold 0.0.

**Solution**
(1) : label dictionary are sorted before evaluation in `inference.py` rather than removing the `sorted(labels)` in `mask_from_annnotation` because anomalib also uses this func and can be affected to Geti which generates random bytes id.
(2) : `project_labels` are sorted before converting gt mask to otx mask and `self.CLASSES` are realigned with sorted one.
(3) : **_To be updated in other PR (need to discuss)_**

**Checklists**
- [x] [WIP] Unit test code
- [x] [WIP] CLI Test

### DIFF
--- a/otx/algorithms/segmentation/adapters/mmseg/data/dataset.py
+++ b/otx/algorithms/segmentation/adapters/mmseg/data/dataset.py
@@ -107,7 +107,6 @@ class OTXSegDataset(CustomDataset, metaclass=ABCMeta):
                 ann_info=dict(labels=self.labels),
                 ignored_labels=ignored_labels,
             )
-            breakpoint()
 
             return data_info
 
@@ -274,4 +273,4 @@ class MPASegDataset(OTXSegDataset, metaclass=ABCMeta):
                     self.label_map[i] = -1
                 else:
                     self.label_map[i] = classes.index(c)
-            breakpoint()
+            # breakpoint()
--- a/otx/algorithms/segmentation/adapters/mmseg/data/dataset.py
+++ b/otx/algorithms/segmentation/adapters/mmseg/data/dataset.py
@@ -45,7 +45,6 @@ def get_annotation_mmseg_format(dataset_item: DatasetItemEntity, labels: List[La
     :param labels: List of labels in the project
     :return dict: annotation information dict in mmseg format
     """
-
     gt_seg_map = mask_from_dataset_item(dataset_item, labels)
     gt_seg_map = gt_seg_map.squeeze(2).astype(np.uint8)
 
@@ -108,6 +107,7 @@ class OTXSegDataset(CustomDataset, metaclass=ABCMeta):
                 ann_info=dict(labels=self.labels),
                 ignored_labels=ignored_labels,
             )
+            breakpoint()
 
             return data_info
 
@@ -129,6 +129,7 @@ class OTXSegDataset(CustomDataset, metaclass=ABCMeta):
         dataset_labels = self.otx_dataset.get_labels(include_empty=False)
         self.project_labels = self.filter_labels(dataset_labels, classes)
         self.CLASSES, self.PALETTE = self.get_classes_and_palette(classes, None)
+        self.CLASSES = [classes[0]] + [label.name for label in sorted(self.project_labels)] # FIXME : how to deal with id sorting mismatch
 
         # Instead of using list data_infos as in CustomDataset, this implementation of dataset
         # uses a proxy class with overriden __len__ and __getitem__; this proxy class
@@ -270,3 +271,4 @@ class MPASegDataset(OTXSegDataset, metaclass=ABCMeta):
                     self.label_map[i] = -1
                 else:
                     self.label_map[i] = classes.index(c)
+            breakpoint()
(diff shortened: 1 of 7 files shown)


sungmanc:
@ashwinvaidya17 , could you double check the effect of `mask_from_annnotation` function ? As Soobee said, the sort logic looks weird. https://github.com/openvinotoolkit/training_extensions/blob/6cba8c9f42ef55171f41fc6e6c63d3367acf730e/otx/api/utils/segmentation_utils.py#L50

@sungchul2 , maybe you also have knowledge about the segmentation issue, do you have any idea or comments?

otx/algorithms/segmentation/adapters/mmseg/data/dataset.py
```diff
@@ -263,6 +263,10 @@ def __init__(self, **kwargs):
         else:
             classes = []
         super().__init__(otx_dataset=otx_dataset, pipeline=pipeline, classes=classes)
+
+        # TODO [Soobee] : Need to align again when background label is removed
+        self.CLASSES = ["background"] + [label.name for label in self.project_labels]
```
harimkang: I think it would be nice to add some code here to check if this has a class called `background`, what do you think?

supersoob: Good. will revise it.

sungmanc: IMO, we need to consider the variants such as `Background` or `back ground`...

sungmanc (dismissed):
Thanks for the nice work. BTW, unit-test checking is needed

sungchul1 (dismissed):
LGTM.
But it would be better to handle it in `mask_from_annnotation` if there is a way to solve it in `mask_from_annnotation`.

kprokofi:
Unfortunately, I tested your branch and nothing changed
otx eval outputs much lower score than expected
Many classes aren't predicted at all during training:
image
otx eval:
image

I'm looking now also in that problem, but I think sorting classes didn't solve this issue

kprokofi:
during otx eval it seems like there is no labels in dataset at all:
image

supersoob:
> Unfortunately, I tested your branch and nothing changed otx eval outputs much lower score than expected Many classes aren't predicted at all during training: image otx eval: image
> 
> I'm looking now also in that problem, but I think sorting classes didn't solve this issue

@kprokofi Hmm.. My VOC dataset is working well. Could you share your model checkpoint and validation dataset? FYI, this will be completely solved when my follow-up PR is merged (please see (3) in this PR description -> bg label is not still included and soft threshold is 0.5 in default)

supersoob:
> during otx eval it seems like there is no labels in dataset at all: image

And it's natural that MPASegDataset for test dataset does not include any labels except for bg because otx eval cli passes validation dataset to infer function with_empty_annotations(). Score measurement is conducted with dataset with actual annotations in otx task side not in mpa.

kprokofi:
@supersoob thank you for this update. Testing cityscapes I found out mismatch in labels caused by:
image
There is no need to shift labels in that case
Could you please remove it? (I mean this "k+1") - simple delete this line. 
training_extensions/otx/core/data/adapter/segmentation_dataset_adapter.py

sungmanc:
> @supersoob thank you for this update. Testing cityscapes I found out mismatch in labels caused by: image There is no need to shift labels in that case Could you please remove it? (I mean this "k+1") - simple delete this line. training_extensions/otx/core/data/adapter/segmentation_dataset_adapter.py

Currently, Cityscapes dataset is not supported due to the background label issue.

supersoob:
> @supersoob thank you for this update. Testing cityscapes I found out mismatch in labels caused by: image There is no need to shift labels in that case Could you please remove it? (I mean this "k+1") - simple delete this line. training_extensions/otx/core/data/adapter/segmentation_dataset_adapter.py

Thank you for testing with various dataset. But unfortunately, every background insertion in otx is root cause for running cityscapes. Even if I shift the label, cityscape won't work as long as I don't remove all bg label in otx which is a very huge job. And that solution affects to voc and other dataset with bg. I plan to stabilize voc first and then look at cityscapes.. I hope to merge this PR if you don't have performance problem with voc and custom seg with >10 labels(with backgrounds)

jaegukhyun (approved):

kprokofi (approved):
Let's merge it now! Thank you for solving labels mismatch problem

PR #1854, opened 2023-03-03
[FIX] Fix datumaro to otx classification adapter

sovrasov:
This fix resolves multilabel classification performance drop.


sungmanc (approved):
Thanks for fixing the multi-label problem.

jaegukhyun (approved):

PR #1860, opened 2023-03-07
[Docs] Add in-memory caching guide to quick_start_guide/cli_commands

vinnamkim:
Add in-memory cache CLI command guide to the Quick Start Guide - CLI commands page.

### DIFF
--- a/docs/source/guide/index.rst
+++ b/docs/source/guide/index.rst
@@ -21,8 +21,7 @@ Guide
   :hidden:
   :caption: Explanation
 
-  explanation/algorithms/index
-  explanation/additional_features/index
+  explanation/index
 
 
 .. toctree::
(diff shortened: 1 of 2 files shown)


docs/source/guide/index.rst
```diff
@@ -21,8 +21,7 @@ Guide
   :hidden:
   :caption: Explanation
 
-  explanation/algorithms/index
-  explanation/additional_features/index
+  explanation/index
```
sungmanc, line 24: Why this line is changed?

sungmanc (approved):
LGTM w/ minor question

docs/source/guide/index.rst
vinnamkim: I found that there already exists a parent document with `toctree` for those two contents. It would be cleaner structure to link to the parent document than to link to the children.

https://github.com/openvinotoolkit/training_extensions/blob/329aead3ae9f714778bfa2d05da1db96d1fb897a/docs/source/guide/explanation/index.rst?plain=1#L20-L25

sungmanc: Thanks for explanation

goodsong81 (approved):
Thank you for the update!

PR #1861, opened 2023-03-07
[FIX][Releasev0.5.1]Remove use_adaptve_interval term checking part

jaegukhyun:
Current release 0.5.0 do not use adaptive interval hook if there is no use_adaptive_interval in recipe file.
Although we expect our default is using adaptive interval hook, but there is no use_adaptive_interval flag in detection's recipe file
Therefore the default behavior of detection is not using adaptive interval hook, and this causes severe underfitting of detection model. 
To fix this issue without modifying mpa submodule, this PR deletes the checking part of use_adaptive_interval flag.
From this PR, all task will use adaptive interval hook if there is adaptive_validation_interval attributes in recipe files.

### DIFF
--- a/external/model-preparation-algorithm/mpa_tasks/apis/task.py
+++ b/external/model-preparation-algorithm/mpa_tasks/apis/task.py
@@ -185,14 +185,6 @@ class BaseTask:
                 del self._model_cfg["fp16"]
         self._precision = [ModelPrecision.FP32]
 
-        # Add/remove adaptive interval hook
-        if self._recipe_cfg.get("use_adaptive_interval", False):
-            self._recipe_cfg.adaptive_validation_interval = self._recipe_cfg.get(
-                "adaptive_validation_interval", dict(max_interval=5)
-            )
-        else:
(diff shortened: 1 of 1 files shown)


jaegukhyun:
run ote-test

eunwoosh:
I'm not sure, but as far as I know, only detection task uses it. Doesn't this PR affect to other tasks?

goodsong81 (dismissed):
Thank you for the quick fix. It looks OK to me, but let's see the test results.

jaegukhyun:
> I'm not sure, but as far as I know, only detection task uses it. Doesn't this PR affect to other tasks?

I though the other task might use adaptive interval, but you are right. It seems that only detection uses this feature

goodsong81 (approved):

goodsong81:
http://validationreports.sclab.intel.com:8006/reports/build_number_report?test_session_build_number=pr-ote-test-921&environment=iotg-dev-workstation-08
2 failures are not relevant to this change.

PR #1862, opened 2023-03-07
[FIX] Enable VOC dataset in OTX

supersoob:
**Issue**

- VOC dataset is not properly supported in OTX
- Cannot exactly detect the train/val subset specified in VOC dataset

**Root Cause**

- When get_train_dataset from datumaro format, it checks key includes any "train" str. Thus, it first bring the subset dataset named "trainval", not "train"
- When get_val_dataset from datumaro format, it checks key includes any "val" str. Thus, it first bring the subset dataset named "trainval", not "val"
- Only if masks annotation exists, dataset item is valid. but it included the shape with no annotation.

**Solution**

- Resolved by checking whole words, "train" and "val" first


sungmanc (dismissed):
Nice work :)

kprokofi:
Will it also fix COCO dataset bug, when it uses the same "trainval" instead of "train" and "val" (for instance segmentation)?

sungmanc:
> Will it also fix COCO dataset bug, when it uses the same "trainval" instead of "train" and "val" (for instance segmentation)?

Yes, if user prepare "instances_trainval.json", then we can avoid including this annotation to train annotation.

kprokofi (dismissed):

jaegukhyun (dismissed):

PR #1863, opened 2023-03-07
[CI][SDL] Added Bandit scanning job to code-scan wf

yunchu:


goodsong81 (dismissed):
LGTM!

.github/workflows/code_scan.yml
```diff
@@ -7,6 +7,26 @@ on:
     - cron: "0 18 * * 1-5"
 
 jobs:
+  Bandit:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
```
eunwoosh: It couldn't matter, but why using actions/checkout@v2 instead of actions/checkout@v3?

.github/workflows/code_scan.yml
```diff
@@ -7,6 +7,26 @@ on:
...
+      - name: Set up Python
+        uses: actions/setup-python@v2
```
eunwoosh: As same in above context, why using actions/setup-python@v2 instead of actions/setup-python@v3?

ipas_default.config
```diff
@@ -92,6 +92,14 @@ tests:
 skips:
   [ 'B101', 'B102', 'B103', 'B104', 'B105', 'B106', 'B107', 'B108', 'B110', 'B112', 'B201', 'B501', 'B502', 'B503', 'B504', 'B505', 'B506', 'B507', 'B601', 'B602', 'B603', 'B604', 'B605', 'B606', 'B607', 'B608', 'B609', 'B610', 'B611', 'B701', 'B702', 'B703']
 
+
+# Added to exclude some path which are not actual source code for this project
```
eunwoosh, line 96: isn't it better to use `targets = ['otx']` instead of using `exclude_dirs`?
I don't know well about this, so it could be wrong.

eunwoosh (dismissed):
LGTM. I'm curious about some code so I left some comments. Please take a look :)

.github/workflows/code_scan.yml
yunchu: @eunwoosh  Thanks for your finding!

ipas_default.config
yunchu: @eunwoosh I don't know either but I think it's just "include" vs "exclude" not "good" or "better". If we use "target" attribute for configuration, we need to add "tests" path as well.

eunwoosh (approved):

sungmanc (approved):

PR #1865, opened 2023-03-07
[FEATURE] Use datumaro format in anomaly dataset adapters 

djdameln:
This PR updates the dataset adapter for the anomaly tasks to make use of the MVTec data format that was recently added to Datumaro. 

- Adds autosplitting support for anomaly tasks. Autosplitting for anomaly tasks is a bit different conmpared to other tasks, because the validation set has to be subsampled from the test set, not the training set. I had to add a case for this in `ConfigManager.auto_split_data()`
- `AnomalyClassificationDatasetAdapter`, `AnomalySegmentationDatasetAdapter` and `AnomalyDetectionDatasetAdapter` were merged into a single class `AnomalyDatasetAdapter`. This is possible because the differences between the task types is already accounted for in the type of annotations present in the Datumaro dataset object.
- Because Datumaro now supports MVTec format, the dataset adapter now works with the `_import_dataset` method from the base class.


otx/cli/manager/config_manager.py
```diff
@@ -231,14 +232,21 @@ def auto_split_data(self, data_roots: str, task: str):
...
+        test_dataset = self.dataset_manager.get_test_dataset(dataset)
         splitted_dataset = None
         if self.data_format in AUTOSPLIT_SUPPORTED_FORMAT:
             if val_dataset is None:
-                splitted_dataset = self.dataset_manager.auto_split(
-                    task=task,
-                    dataset=train_dataset,
-                    split_ratio=[("train", 0.8), ("val", 0.2)],
-                )
+                if self.data_format == "mvtec":
```
sungmanc: It is just an opinion, how about adding this logic into the `DatasetManager`?
The objective of `ConfigManager` controls the overall configurations, so adding logic regarding the dataset to `DatasetManager` is preferable to me. What do you think?

sungmanc (changes requested):
Overall, LGTM w/ left a minor opinion. However, need to check the FAIL cases occurred from unit-test CI.

otx/cli/manager/config_manager.py
djdameln: I agree that the dataset manager would be a more suitable place for this logic. But since auto splitting is performed on the test set for anomaly datasets and on the train set for other datasets, we would have to change the signature of `DatasetManager.auto_split`. Effectively we would have to move almost the entire contents of `ConfigManager.auto_split_data` to `DatasetManager.auto_split`. It's fine by me, but I don't know if this would break the intended design.

sungmanc: Agreed, there were many special conditions in `ConfigManager`.. Let's keep it as is.
BTW, could you resolve FAIL cases?

djdameln: I'm working on it :)

otx/core/data/adapter/anomaly_dataset_adapter.py
```diff
@@ -77,150 +59,54 @@ def _prepare_anomaly_label_information(self) -> List[LabelEntity]:
...
+            elif ann.type == AnnotationType.bbox:
+                if self._is_normal_bbox(ann.points[0], ann.points[1], ann.points[2], ann.points[3]):
+                    shapes.append(self._get_normalized_bbox_entity(ann, width, height))
+        return shapes
 
     def get_otx_dataset(self) -> DatasetEntity:
-        """Convert DatumaroDataset to DatasetEntity for Anomaly classification."""
-        normal_label, abnormal_label = self._prepare_anomaly_label_information()
-        self.label_entities = [normal_label, abnormal_label]
+        """Convert DatumaroDataset to DatasetEntity for Anomaly segmentation."""
```
ashwinvaidya17: Anomaly tasks maybe?

tests/unit/core/data/adapter/test_anomaly_adapter.py
```diff
@@ -31,13 +27,13 @@ def setup_method(self):
         self.val_data_roots: str = os.path.join(self.root_path, data_root_dict["val"])
         self.test_data_roots: str = os.path.join(self.root_path, data_root_dict["test"])
 
-        self.train_dataset_adapter = AnomalyClassificationDatasetAdapter(
+        self.train_dataset_adapter = AnomalyDatasetAdapter(
```
ashwinvaidya17: Probably not a part of this PR but the only difference I see between these tests are the task types. Maybe we can just use pytest parameterize and reduce the classes? But this is optional.

ashwinvaidya17 (commented):
Thanks for the efforts. The code looks much simpler now. I have two very minor comments.

otx/core/data/adapter/anomaly_dataset_adapter.py
djdameln: yep, this is a leftover from development

djdameln:
@sungmanc can you have a look at these failed tests in the CI? Everything passes fine on my local machine, and these tests that are now failing were previously passing with the exact same configuration.

sungmanc:
> @sungmanc can you have a look at these failed tests in the CI? Everything passes fine on my local machine, and these tests that are now failing were previously passing with the exact same configuration.

@wonjuleee, could you check this issue? When I checked this issue, the error occurred from `import_from` function.
Although Datumaro successfully recognized the dataset type (`MvtecClassificationBase`), It couldn't read labels properly. 

Our test data (tests/assets/anomaly/shapes) is exactly the same as the single category `MVTec` dataset (e.g. MVTec/bottle). 
image

Are we missing something?

sungmanc:
@djdameln , could you recheck with your local test? when I checked with your branch in local machine, I also got errors. 
Since we used Datumaro installed via pipy, it's status can't be changed. 

If you have experienced successful tests, then the code might be changed.

djdameln:
> @djdameln , could you recheck with your local test? when I checked with your branch in local machine, I also got errors.
Since we used Datumaro installed via pipy, it's status can't be changed.
>
> If you have experienced successful tests, then the code might be changed.

@sungmanc no matter what I try, the tests pass locally. I tried cloning the repo in a new folder and installing with a clean env, and the tests pass. I don't know what else I could try to solve this issue.

ashwinvaidya17:
I tried this on my system as well. The tests are passing locally. Even I am not able to reproduce the issue thrown in the CI
tests

sungmanc:
@ashwinvaidya17 , @djdameln . Okay I'll re-try local tests and look for this issue. 
Thanks for reporting

PR #1868, opened 2023-03-08
[FIX] hang issue when tracing a stack in certain scenario

cih9088:
### Changes
- use a primitive library `sys` instead of a wrapper `inspect` to trace a call stack.


sungmanc (approved):

eunwoosh (approved):
Thanks for your work. I checked that hanging TC works well after the changes is applied.

JihwanEom (approved):

supersoob (approved):

PR #1870, opened 2023-03-08
[FIX][POT] Set stat_requests_number parameter to 1

nikita-savelyevv:
Set POT `stat_requests_number` parameter to 1 in order to lower RAM footprint

Related ticket: 105234


goodsong81 (approved):
Thank you for the fix!

PR #1871, opened 2023-03-09
[CI] Upgrade actions/setup-python version to v4

yunchu:
Upgrade actions/setup-python version to v4 to avoid error from some deprecated usage of actions.

### DIFF
--- a/.github/workflows/code_scan.yml
+++ b/.github/workflows/code_scan.yml
@@ -13,7 +13,7 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v3
       - name: Set up Python
-        uses: actions/setup-python@v3
+        uses: actions/setup-python@v4
         with:
           python-version: 3.8
       - name: Install dependencies
--- a/.github/workflows/pre_merge.yml
+++ b/.github/workflows/pre_merge.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v3
       - name: Set up Python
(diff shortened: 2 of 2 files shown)


goodsong81 (approved):
Thank you for the update.
Just one question: can we set-up multiple python version?
(for python3.10 migration in the near future, and multiple python version tests for non-near future :))

eunwoosh (approved):

yunchu:
> Thank you for the update. Just one question: can we set-up multiple python version? (for python3.10 migration in the near future, and multiple python version tests for non-near future :))

Of course it's possible.

samet-akcay:
@goodsong81, @yunchu, I agree with both of you. It would be good to support multiple python versions. Luckily tox is capable of doing this via 

```tox
[tox]
envlist = py38, py39, py10
```

The question at which stage of the CI we test this. pre-merge or nightly?

goodsong81:
> @goodsong81, @yunchu, I agree with both of you. It would be good to support multiple python versions. Luckily tox is capable of doing this via
> 
> ```
> [tox]
> envlist = py38, py39, py10
> ```
> 
> The question at which stage of the CI we test this. pre-merge or nightly?

In my local test of above feature last week, tox just used the system (or current virtual env) python versions.
So, if there is no Python3.10 installed, py310 does not work.
Please correct me as I'm not that expert in this subject :)
Anyway, that's why I'm asking.
If possible, I prefer nightly runs in parallel.

PR #1872, opened 2023-03-09
[FIX] Training error when batch size is 1

cih9088:
### Summary
When the last batch size is 1, training batch normalization is a nonsense. We should drop it instead.

### Changes
- `drop_last=True` if the last batch size is 1.

### DIFF
--- a/otx/mpa/stage.py
+++ b/otx/mpa/stage.py
@@ -306,8 +306,17 @@ class Stage(object):
 
         if distributed:
             dataset_len = dataset_len // dist.get_world_size()
+
+        # set batch size as a total dataset
+        # if it is smaller than total dataset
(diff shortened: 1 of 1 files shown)


sungmanc (approved):

goodsong81 (commented):
@JihwanEom Could you take a look? I remember that there was a bug fix related to drop_last option. Please double-check :)

JihwanEom:
> @JihwanEom Could you take a look? I remember that there was a bug fix related to drop_last option. Please double-check :)

It's okay for me (also passed h-label tests), the drop_last related issue was raised when 1 < sample_size < batch_size. LGTM!

JihwanEom (approved):

PR #1873, opened 2023-03-09
[FIX] Dataset Adapter: Avoid duplicated annotation and permit empty image

sungmanc:
### This PR includes
* VOC import error: VOC includes several annotation formats (label, bbox, mask). In this case, if we want to make detection dataset, then we need to filter out the mask, and label annotations. --> `len(shapes) >0`.
* However, if there are empty label image that has no annotations (len(annotations)=0), we need to add them to `DatasetItemEntity`. --> `len(datumaro_item.annotations)==0`.
* For the unlabeled dataset, it also be included in `DatasetItemEntity`. --> `subset == Subset.UNLABELED`

```python3
if len(shapes) > 0 or subset == Subset.UNLABELED or len(datumaro_item.annotations) == 0:
```
---
* To test the empty_label case, I modified the test data that includes the empty_label image.
* Implement unit-test to validate whether empty_label image is successfully included in `DatasetItemEntity` or not.

  image

* Test results
  image

Thanks for reporting this issue. @eunwoosh , @supersoob


tests/unit/core/data/test_helpers.py
```diff
@@ -86,6 +86,7 @@ def generate_datumaro_dataset_item(
     task: str,
     image_shape: np.array = np.array((5, 5, 3)),
     mask_shape: np.array = np.array((5, 5)),
+    empty_label: bool = False,
```
eunwoosh: I think `empty_label` isn't used anywhere. Is it used somewhere I don't know?

eunwoosh (commented):
I think dataset w/o annotations need to be filtered out during training stage. I run detection on this branch but it isn't.

otx/core/data/adapter/segmentation_dataset_adapter.py
```diff
@@ -92,6 +92,7 @@ def get_otx_dataset(self) -> DatasetEntity:
                                 shapes.append(self._get_polygon_entity(d_polygon, image.width, image.height))
                                 if d_polygon.label not in used_labels:
                                     used_labels.append(d_polygon.label)
+
                     if len(shapes) > 0 or subset == Subset.UNLABELED:
```
supersoob, line 96: I don't get why there is no if-statement `len(datumaro_item.annotations) == 0` in seg like det. Does seg include an image file listed in dataset subset but with no annotation well?

tests/unit/core/data/test_helpers.py
sungmanc: Thanks, I missed to delete it

otx/core/data/adapter/segmentation_dataset_adapter.py
sungmanc: For segmentation, we need to consider various cases.

If we regard some items that have no annotations as `background` label, then `len(datumaro_item.annotations)==0` shouldn't be applied.

sungmanc:
@eunwoosh , I added logic to avoid inputting the empty image to train dataset. And changed tests/assets.

otx/core/data/adapter/segmentation_dataset_adapter.py
sungmanc: In other words, an empty image also is predicted as `background` in the segmentation case. So, all segmentation data items have annotation to represents the background

otx/core/data/adapter/detection_dataset_adapter.py
```diff
@@ -50,7 +50,7 @@ def get_otx_dataset(self) -> DatasetEntity:
                         if ann.label not in used_labels:
                             used_labels.append(ann.label)
 
-                    if len(shapes) > 0 or subset == Subset.UNLABELED or len(datumaro_item.annotations) == 0:
+                    if len(shapes) > 0 or subset == Subset.UNLABELED or (subset == Subset.VALIDATION and len(datumaro_item.annotations)) == 0:
```
eunwoosh:
```suggestion
                    if len(shapes) > 0 or subset == Subset.UNLABELED or (subset != Subset.TRAINING and len(datumaro_item.annotations) == 0):
```

eunwoosh: how about checking it's not training dataset not to filter out image w/o annotation in test dataset.

sungmanc: I can't understand what you want to say.. could you explain more detail?

eunwoosh: So current change just keep images w/o annotation only in validation dataset but not in test dataset. To keep them in both case, how about checking `subset != Subset.TRAINING` instead of `subset == Subset.VALIDATION`?

eunwoosh (dismissed):

eunwoosh (approved):
LGTM.

sungmanc:
@supersoob , @jaegukhyun  could you recheck this?

supersoob (approved):

PR #1874, opened 2023-03-09
[FEATURE] Add train-type parameter to otx train

sovrasov:
Currently, to enable SSL training we have to execute oxt build at the first step.
At the same time, otx train allows passing --unlabeled-data-roots, which causes a failure if otx build hasn't been executed before.
Therefore, CLI of otx train is contradictive and incomplete. This PR brings back missing ability to launch different training types right from otx train CLI. 

This contradiction is caused by mixing two paradigms for launching a training: pre-configuration of a workspace and passing all the required arguments via CLI. The possible solutions are:
- restricting of otx train CLI.
- maintaining otx train CLI as a standalone and self-containing tool for configuring and launching trainings.

This PR follows the second option, but the first one deserves discussion as well.

### DIFF
--- a/otx/cli/manager/config_manager.py
+++ b/otx/cli/manager/config_manager.py
@@ -186,7 +186,7 @@ class ConfigManager:  # pylint: disable=too-many-instance-attributes
             if arg_algo_backend:
                 train_type = arg_algo_backend.get("train_type", {"value": "INCREMENTAL"})  # type: ignore
                 return train_type.get("value", "INCREMENTAL")
-            if self.mode in ("build", "train") and self.args.train_type:
+            if hasattr(self.args, "train_type") and self.mode in ("build", "train") and self.args.train_type:
                 self.train_type = self.args.train_type.upper()
                 if self.train_type not in TASK_TYPE_TO_SUB_DIR_NAME:
                     raise ValueError(f"{self.train_type} is not currently supported by otx.")
--- a/otx/cli/manager/config_manager.py
+++ b/otx/cli/manager/config_manager.py
@@ -186,7 +186,7 @@ class ConfigManager:  # pylint: disable=too-many-instance-attributes
             if arg_algo_backend:
                 train_type = arg_algo_backend.get("train_type", {"value": "INCREMENTAL"})  # type: ignore
                 return train_type.get("value", "INCREMENTAL")
-            if self.mode in ("build") and self.args.train_type:
+            if self.mode in ("build", "train") and self.args.train_type:
                 self.train_type = self.args.train_type.upper()
                 if self.train_type not in TASK_TYPE_TO_SUB_DIR_NAME:
                     raise ValueError(f"{self.train_type} is not currently supported by otx.")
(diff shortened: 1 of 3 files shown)


goodsong81 (dismissed):
LGTM. @harimkang Could you have a look?

sungmanc:
I'm not sure why we should use `otx build` for SSL training. Users could use `otx cli` if they give `train_params`.
We currently use `params --algo_backend.train_type` to enable the different train types. 

Example for SemiSL:
https://github.com/openvinotoolkit/training_extensions/blob/0e4a8b7929a1162fe06e57e34b1ac5a2f986dcdb/tests/e2e/cli/classification/test_classification.py#L291

Example for SelfSL: https://github.com/openvinotoolkit/training_extensions/blob/0e4a8b7929a1162fe06e57e34b1ac5a2f986dcdb/tests/e2e/cli/classification/test_classification.py#L715

otx/cli/manager/config_manager.py
```diff
@@ -186,7 +186,7 @@ def _get_train_type(self, ignore_args: bool = False) -> str:
             if arg_algo_backend:
                 train_type = arg_algo_backend.get("train_type", {"value": "INCREMENTAL"})  # type: ignore
                 return train_type.get("value", "INCREMENTAL")
-            if self.mode in ("build") and self.args.train_type:
+            if self.mode in ("build", "train") and self.args.train_type:
```
harimkang: I think we should first check whether train_type exists in args.
False should be returned if train_type does not exist.

harimkang:
> I'm not sure why we should use `otx build` for SSL training. Users could use `otx cli` if they give `train_params`.
> 
> We currently use `params --algo_backend.train_type` to enable the different train types. 
> 
> 
> 
> Example for SemiSL:
> 
> https://github.com/openvinotoolkit/training_extensions/blob/0e4a8b7929a1162fe06e57e34b1ac5a2f986dcdb/tests/e2e/cli/classification/test_classification.py#L291
> 
> 
> 
> Example for SelfSL: https://github.com/openvinotoolkit/training_extensions/blob/0e4a8b7929a1162fe06e57e34b1ac5a2f986dcdb/tests/e2e/cli/classification/test_classification.py#L715

I agree to some extent, but vlad seems to want to improve the inconvenience.  However, there can be conflicts between the two methods(use train_type / use params), and it seems that this could cause problems somewhere. (maybe not..?)

sovrasov:
> if self.mode in ("build", "train") and self.ar

`_get_train_type` method has a clear logic to extract training type from template & parameters: according to the code `--train_type` (when it's specified) will be used if `params --algo_backend.train_type` is not specified, so I don't expect any unpredictable behavior here.

sovrasov:
> I'm not sure why we should use `otx build` for SSL training. Users could use `otx cli` if they give `train_params`. We currently use `params --algo_backend.train_type` to enable the different train types.
> 
> Example for SemiSL:
> 
> https://github.com/openvinotoolkit/training_extensions/blob/0e4a8b7929a1162fe06e57e34b1ac5a2f986dcdb/tests/e2e/cli/classification/test_classification.py#L291
> 
> Example for SelfSL:
> 
> https://github.com/openvinotoolkit/training_extensions/blob/0e4a8b7929a1162fe06e57e34b1ac5a2f986dcdb/tests/e2e/cli/classification/test_classification.py#L715

`params --...` is also a correct way, but it's not documented and hard to discover (since otx train --help doesn't reveal that's hidden under the params section).

harimkang (approved):

sungmanc (approved):

sungmanc:
After this PR is merged, we need to consider eliminating `params --algo_backend.train_type` from the configuration.

PR #1880, opened 2023-03-14
[FIX] Arrange scale between bbox preds and bbox targets in ATSS

jaegukhyun:
**Cause**
Bbox encode or decode status is not matched between bbox preds and bbox targets in ATSS

**Solution**
Check self.reg_decoded_bbox value. If self.reg_decoded_bbox is True, then ATSS has to decode bbox_pred values since bbox_target values are not encoded

### DIFF
--- a/otx/mpa/modules/models/heads/custom_atss_head.py
+++ b/otx/mpa/modules/models/heads/custom_atss_head.py
@@ -168,18 +168,16 @@ class CustomATSSHead(CrossDatasetDetectorHead, ATSSHead):
             pos_centerness = centerness[pos_inds]
 
             centerness_targets = self.centerness_target(pos_anchors, pos_bbox_targets)
-            pos_decode_bbox_pred = self.bbox_coder.decode(pos_anchors, pos_bbox_pred)
-            pos_decode_bbox_targets = self.bbox_coder.decode(pos_anchors, pos_bbox_targets)
+            if self.reg_decoded_bbox:
+                pos_bbox_pred = self.bbox_coder.decode(pos_anchors, pos_bbox_pred)
 
             if self.use_qfl:
-                quality[pos_inds] = bbox_overlaps(
-                    pos_decode_bbox_pred.detach(), pos_decode_bbox_targets, is_aligned=True
-                ).clamp(min=1e-6)
+                quality[pos_inds] = bbox_overlaps(pos_bbox_pred.detach(), pos_bbox_targets, is_aligned=True).clamp(
+                    min=1e-6
+                )
 
             # regression loss
-            loss_bbox = self.loss_bbox(
-                pos_decode_bbox_pred, pos_decode_bbox_targets, weight=centerness_targets, avg_factor=1.0
-            )
+            loss_bbox = self.loss_bbox(pos_bbox_pred, pos_bbox_targets, weight=centerness_targets, avg_factor=1.0)
 
             # centerness loss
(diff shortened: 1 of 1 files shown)


otx/mpa/modules/models/heads/custom_atss_head.py
```diff
@@ -168,18 +168,16 @@ def loss_single(
             pos_centerness = centerness[pos_inds]
 
             centerness_targets = self.centerness_target(pos_anchors, pos_bbox_targets)
-            pos_decode_bbox_pred = self.bbox_coder.decode(pos_anchors, pos_bbox_pred)
-            pos_decode_bbox_targets = self.bbox_coder.decode(pos_anchors, pos_bbox_targets)
+            if self.reg_decoded_bbox:
```
eunwoosh, line 171:
```suggestion
            if not self.reg_decoded_bbox:
```
Isn't the contrary right?

jaegukhyun: I don't think so. Targets are encoded when self.reg_decoded_bbox is False. Therefore if self.reg_decoded_bbox is False, then the bbox_pred don't need to decode. However if self.reg_decoded_bbox is True, then the bbox_pred should be decoded since the bbox_targets are not encoded

eunwoosh (approved):
LGTM. Thanks for your work :)

sungmanc (changes requested):
Could you follow the PR template?
e.g. https://github.com/openvinotoolkit/datumaro/pull/854

eunwoosh:
Isn't same bugfix needed to release branch?

otx/mpa/modules/models/heads/custom_atss_head.py
```diff
@@ -168,18 +168,16 @@ def loss_single(
...
+                pos_bbox_pred = self.bbox_coder.decode(pos_anchors, pos_bbox_pred)
 
             if self.use_qfl:
-                quality[pos_inds] = bbox_overlaps(
-                    pos_decode_bbox_pred.detach(), pos_decode_bbox_targets, is_aligned=True
-                ).clamp(min=1e-6)
+                quality[pos_inds] = bbox_overlaps(pos_bbox_pred.detach(), pos_bbox_targets, is_aligned=True).clamp(
+                    min=1e-6
+                )
```
goodsong81, line 177: In my opinion, `reg_decoded_bbox` means that the model should output the decoded bbox directly. So, the `pos_bbox_targets` should be decoded to enforce model to produce decoded output from its regression head.

goodsong81 (changes requested):
I've left a comment. Please let me know your opinion.

otx/mpa/modules/models/heads/custom_atss_head.py
jaegukhyun: Whether the final outputs of model are decoded or not is handled in simple_test funciton. Since it is loss calculation funciton, I think the only thing to check is whether the predictions and targets are in same scale. 
If you're worrying is that the loss do not reflect the actual value of the model's prediction, please let me know

goodsong81: If it's in simple_test(), then it's alright.
Thank you for the fix.

goodsong81 (approved):

otx/mpa/modules/models/heads/custom_atss_head.py
jaegukhyun: image
For sure, it's in get_bboxes function in base_dense_head.py

PR #1883, opened 2023-03-14
[FIX] Fix the condition of supcon for seg

sungchul1:
### Summary
- When training with supcon and using `otx train ... --work-dir=otx-supcon-workspace` to create a workspace including `supcon` text in the path, otx cannot find `supcon` related files (`data_pipeline.py` and `model.py`).
    ```bash
    FileNotFoundError: data_pipeline: training_extensions/otx-supcon-workspace/./data_pipeline.py not founded
    ```
- Change the condition `"supcon" not in self._model_dir` → `not self._model_dir.endswith("supcon")`.
- It is just to find `supcon` directory in the workspace.

### How to test
```bash
$ pytest tests/unit/algorithms/segmentation/tasks/test_segmentation_inference.py::TestOTXSegTaskInference::test_init_recipe_supcon

============================================= 3 passed, 17 warnings in 1.50s =============================================
```

### DIFF
--- a/otx/algorithms/segmentation/tasks/inference.py
+++ b/otx/algorithms/segmentation/tasks/inference.py
@@ -199,7 +199,7 @@ class SegmentationInferenceTask(BaseTask, IInferenceTask, IExportTask, IEvaluati
             self._train_type in RECIPE_TRAIN_TYPE
             and self._train_type == TrainType.INCREMENTAL
             and self._hyperparams.learning_parameters.enable_supcon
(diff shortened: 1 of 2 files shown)


sungmanc (commented):
Want to suggest 
* making the unit-test for this exception case. 
* Add proper milestone for this PR

harimkang (approved):

sungmanc (approved):

PR #1891, opened 2023-03-15
Add Tile Classifier to enhance efficiency of tiny object detection

eugene123tw:
### Checklist

- [x] I submit my changes into the `develop` branch
- [x] I have added description of my changes into [CHANGELOG](https://github.com/openvinotoolkit/training_extensions/blob/develop/CHANGELOG.md)
- [x] I have [linked related issues](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)

### License

```python
#
```


otx/mpa/deploy/utils/onnx.py
eugene123tw, line 1: @cih9088 I had to add the following changes from mmdeploy, otherwise, there are dangling nodes in ONNX and OpenVINO would fail to partition my tile classifier model. I didn't dig too deep but it works, you might wanna have a look.

otx/core/data/adapter/detection_dataset_adapter.py
eugene123tw, line 1: @sungmanc I only wanna polygon but I found this code includes boxes for instance segmentation and I end up having twice many of the annotations I wanted.

otx/cli/utils/io.py
eugene123tw, line 1: @goodsong81 Do we need to change Geti backend since it has to read additional models? and who's responsible for this change?

otx/api/utils/tiler.py
```diff
@@ -118,16 +122,25 @@ def predict_tile(
...
-        if return_features:
-            if "feature_vector" in raw_predictions or "saliency_map" in raw_predictions:
-                features = (
-                    raw_predictions["feature_vector"].reshape(-1),
-                    raw_predictions["saliency_map"][0],
-                )
-        return features, output
+        run_prediction = True
+        if not return_features:
+            # TODO[EUGENE]: should we set the threshold (0.45) as a parameter?
```
eugene123tw: @goodsong81 do you think we should set this confidence threshold as a configurable parameter? The higher the faster but risker.

otx/algorithms/common/tasks/training_base.py
```diff
@@ -347,7 +347,7 @@ def _initialize_post_hook(self, options=None):
         pass
 
     @abc.abstractmethod
-    def _init_recipe(self):
+    def _init_recipe(self, dataset: Optional[DatasetEntity] = None):
```
eugene123tw: @harimkang I added this parameter to a function to consolidate tiling-related parameters in one place instead of scattering them throughout the codebase. 

However, this change also affects other classes that inherit from this function, as the new parameter is now part of the function's signature and must be supplied when calling it. This requires updating the implementations of those classes to include the new parameter and its functionality.

harimkang: So this is needed for tiling, so we need to add it to the BaseTask as well, right? I don't think there should be any major issues. If we add None as the default for the other tasks as well, shouldn't there be an issue?
@jaegukhyun What do you think?

otx/mpa/deploy/utils/onnx.py
cih9088: LGTM

otx/core/data/adapter/detection_dataset_adapter.py
```diff
@@ -40,8 +41,10 @@ def get_otx_dataset(self) -> DatasetEntity:
                             self.task_type in (TaskType.INSTANCE_SEGMENTATION, TaskType.ROTATED_DETECTION)
                             and ann.type == AnnotationType.polygon
                         ):
-                            if self._is_normal_polygon(ann):
+                            if isinstance(ann, Polygon):
```
sungmanc: I'm not sure why `ann.type == AnnotationType.polygon` couldn't filter out `Bbox`.

If you want to use `isinstance(ann, Polygon)`, then edit above `if` sentence like below.
```python3
if (
    self.task_type in (TaskType.INSTANCE_SEGMENTATION, TaskType.ROTATED_DETECTION) and isinstance(ann, Polygon)
):
```

yunchu:
@eugene123tw could you set the milestone for this PR?

otx/api/utils/tiler.py
goodsong81: Yeah, I think it's good to be included in config params with default values.

otx/cli/utils/io.py
goodsong81: As I said earlier, Geti does not use OTX CLI.
OTX should be responsible for
* Keep additional IR models using model.set_data()
* Extract, if any, additional IR models using model.get_data()
* Deploy, if any, additional IR models along with base one.
* Infer properly both in OpenVINO task and exportable code
* Keep backward compatibility for models deployed from early versions of OTX/Geti.

eugene123tw: @goodsong81 My question pertains to the input of new models. Specifically, in the OTX CLI, we use the read_model function to read models from a folder. To add two new OpenVINO models, I need to include them with the other files that are required for the `model_adapter_keys variable`. These files include "tile_classifier.xml", "tile_classifier.bin". However, I am unsure where and how Geti reads these additional files before creating `task_environment` and initializing `OpenVINODetectionTask`. And there's nowhere I could test the Geti workflow (read models -> create task environment -> initialise OpenVINO task -> run prediction) behaviour from OTX, should we contact the Geti backend team for further information?

otx/algorithms/detection/tasks/inference.py
```diff
@@ -96,6 +97,47 @@ def __init__(self, task_environment: TaskEnvironment, **kwargs):
...
+                                if layer_name.startswith("tile_classifier"):
+                                    found_tile_classifier = True
+                                    break
+                            if not found_tile_classifier:
+                                raise RuntimeError(
+                                    "Tile classifier is enabled but not found in the trained model. Please retrain your model."
+                                )
+                            hparams.tiling_parameters.enable_tile_classifier = loaded_tiling_parameters[
+                                "enable_tile_classifier"
+                            ]["value"]
```
goodsong81: Need refactoring. Too much tiling-specific logics in _load_model_ckpt()

otx/api/utils/tiler.py
goodsong81: Hmm, or we just set the default good value for a while :)

goodsong81 (commented):
You are already doing OK w/ model.set/get_data()
Model will be saved & loaded this data from model repo.

otx/cli/utils/io.py
goodsong81: AFAIK, Geti will store models w/ their metadata given by set_data().
So, we may not need to worry.
Rather we are dealing with additional data in CLI use cases as we don't have such functionality.
You may test it after OTX is merged to Geti.

otx/api/utils/tiler.py
eugene123tw: then should we leave it at 0.45?

otx/algorithms/detection/tasks/inference.py
eugene123tw: thanks for the comment, I've updated it.

otx/algorithms/common/tasks/training_base.py
jaegukhyun: I think there is no harm for overall structure, but I think this could increase complexity of BaseTask, which is complex enough. Let's see... I'll consider how we can refactor for our new task. At the current status, there is no big objection

jaegukhyun (commented):
I don't have any big objection for this PR, but I want to check integration test and e2e test for all tasks. @eugene123tw Could you attach your local e2e test for all tasks?

eugene123tw:
@jaegukhyun pre-merge tests and CLI tests finished, and codecov also attached. Let me know if there's any info you need. Thanks!

otx/algorithms/detection/adapters/mmdet/utils/config_utils.py
```diff
@@ -326,3 +327,58 @@ def cluster_anchors(recipe_config: Config, dataset: DatasetEntity):
...
+            config.get("data", ConfigDict()).get("test_dataloader", ConfigDict()).update(ConfigDict(samples_per_gpu=1))
+            config.get("data", ConfigDict(samples_per_gpu=1))
+
+        if hparams.tiling_parameters.enable_tile_classifier:
+            logger.info("Tile classifier enabled")
+            logger.info(f"Patch model from: {config.model.type} to CustomMaskRCNNTileOptimised")
+            config.model.type = "CustomMaskRCNNTileOptimised"
+
+            if config.model.backbone.type == "efficientnet_b2b":
+                lr = 0.002  # pylint: disable=invalid-name
```
jaegukhyun:
```suggestion
                learning_rate = 0.002
```

otx/algorithms/detection/adapters/mmdet/utils/config_utils.py
```diff
@@ -326,3 +327,58 @@ def cluster_anchors(recipe_config: Config, dataset: DatasetEntity):
...
+                logger.info(
+                    f"Patch {config.model.backbone.type} LR from: {hparams.learning_parameters.learning_rate} to {lr}"
```
jaegukhyun:
```suggestion
                    f"Patch {config.model.backbone.type} LR from: {hparams.learning_parameters.learning_rate} to {learning_rate}"
```

otx/algorithms/detection/adapters/mmdet/utils/config_utils.py
```diff
@@ -326,3 +327,58 @@ def cluster_anchors(recipe_config: Config, dataset: DatasetEntity):
...
+                )
+                hparams.learning_parameters.learning_rate = lr
```
jaegukhyun:
```suggestion
                hparams.learning_parameters.learning_rate = learning_rate
```

otx/algorithms/detection/adapters/mmdet/utils/config_utils.py
```diff
@@ -326,3 +327,58 @@ def cluster_anchors(recipe_config: Config, dataset: DatasetEntity):
...
+        logger.info("Tiling enabled")
+
+        if dataset and dataset.purpose != DatasetPurpose.INFERENCE and hparams.tiling_parameters.enable_adaptive_params:
+            adaptive_tile_params(hparams.tiling_parameters, dataset)
+            tiling_params = ConfigDict(
+                tile_size=int(hparams.tiling_parameters.tile_size),
+                overlap_ratio=float(hparams.tiling_parameters.tile_overlap),
+                max_per_img=int(hparams.tiling_parameters.tile_max_number),
+            )
+            config.update(
```
jaegukhyun, line 343: What if hparams.tiling_parameters.enable_adaptive_param is False? Is there another logic to add default tiling_params to config or there is no need to add tiling_params to config in that case?

otx/algorithms/detection/adapters/mmdet/utils/config_utils.py
```diff
@@ -326,3 +327,58 @@ def cluster_anchors(recipe_config: Config, dataset: DatasetEntity):
...
+                ConfigDict(
+                    data=ConfigDict(
+                        train=tiling_params,
+                        val=tiling_params,
+                        test=tiling_params,
+                    )
+                )
+            )
+
+        if dataset and dataset.purpose == DatasetPurpose.INFERENCE:
```
jaegukhyun, line 353: Don't we have setting samples_per_gpu to 1 for test_dataloader? If we don't have that logic just resolve this. I think I saw that logic in the stage

jaegukhyun (commented):
I left some comments, could you check them?

otx/algorithms/detection/adapters/mmdet/utils/config_utils.py
eugene123tw: If enable adaptive param is set to false then the default tiling parameters is used:  https://github.com/openvinotoolkit/training_extensions/blob/fd1e6c3cd27c35add8cae76ab474cffd9d5e0c5d/otx/algorithms/detection/configs/detection/configuration.yaml#L411

otx/algorithms/detection/adapters/mmdet/utils/config_utils.py
eugene123tw: It is set to 2 (or 4, I can't remember) in our current code, although I don't know where it is set yet. 

Before I change this universally to 1, I'd like to confirm check the batch size should be set to 1? I think this is set to 1 before, but somehow set to > 1 and I thought this is intentional, no?

otx/algorithms/detection/adapters/mmdet/utils/config_utils.py
jaegukhyun: Yes but those default parameters are updated to config? In this code config is updated only when adaptive tiling params is true

otx/algorithms/detection/adapters/mmdet/utils/config_utils.py
jaegukhyun: Okay, then we should set validation batch size in the task level, I'll keep this in my mind for refactoring

otx/algorithms/detection/adapters/mmdet/utils/config_utils.py
eugene123tw: Tiling parameters already exist in the config as `tile_pipeline.py` is being used instead of using `data_pipeline` https://github.com/openvinotoolkit/training_extensions/blob/5ae47128adc6d312d2b0ae1d29e8e1c0e1a1f5ae/otx/algorithms/common/tasks/training_base.py#L213

otx/core/data/adapter/detection_dataset_adapter.py
eugene123tw: yeah, you're right. I reverted the change.

jaegukhyun (dismissed):

otx/algorithms/common/tasks/training_base.py
sungmanc: Inserting `dataset` to `_init_recipe` function is not a good way for me since `recipe` is not related with dataset

How about using `_initialize` function?

otx/algorithms/detection/adapters/mmdet/models/detectors/custom_maskrcnn_tile_optimised.py
```diff
@@ -0,0 +1,312 @@
+"""Custom MaskRCNN detector with tile classifier."""
```
sungmanc: Most of contents included in this PR used American style word. So, want to suggest to change the title.
optimised --> optimized

otx/algorithms/detection/adapters/mmdet/models/detectors/custom_maskrcnn_tile_optimised.py
```diff
@@ -0,0 +1,312 @@
...
+
+
+class TileClassifier(torch.nn.Module):
+    """Tile classifier for the tile optimised MaskRCNN model."""
+
+    def __init__(self):
+        super().__init__()
+        self.fp16_enabled = False
+        self.features = nn.Sequential(
+            nn.Conv2d(3, 64, kernel_size=11, stride=4, padding=2),
```
sungmanc: Could you modularize this part? maybe we can use below style function

```python
def make_conv_block(in_channels, out_channels, k_size, stride, padding) -> nn.Sequential:
```

otx/algorithms/detection/adapters/mmdet/models/detectors/custom_maskrcnn_tile_optimised.py
```diff
@@ -0,0 +1,312 @@
...
+        self.avgpool = torch.nn.AdaptiveAvgPool2d((6, 6))
+        self.classifier = torch.nn.Sequential(
+            torch.nn.Linear(256 * 6 * 6, 256),
+            torch.nn.ReLU(inplace=True),
+            torch.nn.Linear(256, 256),
+            torch.nn.ReLU(inplace=True),
+            torch.nn.Linear(256, 1),
+        )
+
+        self.loss_fun = torch.nn.BCEWithLogitsLoss()
```
sungmanc:
```suggestion
        self.loss_func = torch.nn.BCEWithLogitsLoss()
```

otx/algorithms/detection/adapters/mmdet/models/detectors/custom_maskrcnn_tile_optimised.py
```diff
@@ -0,0 +1,312 @@
...
+        """Calculate BCE loss.
+
+        Args:
+            pred (torch.Tensor): _description_
+            target (torch.Tensor): binary target
+
+        Returns:
+            torch.Tensor: BCE loss
+        """
+        loss = self.loss_fun(pred, target)
```
sungmanc:
```suggestion
        loss = self.loss_func(pred, target)
```

otx/algorithms/detection/adapters/mmdet/utils/config_utils.py
```diff
@@ -326,3 +327,59 @@ def cluster_anchors(recipe_config: Config, dataset: DatasetEntity):
...
+            config.get("data", ConfigDict()).get("val_dataloader", ConfigDict()).update(ConfigDict(samples_per_gpu=1))
+            config.get("data", ConfigDict()).get("test_dataloader", ConfigDict()).update(ConfigDict(samples_per_gpu=1))
+            config.get("data", ConfigDict(samples_per_gpu=1))
+
+        if hparams.tiling_parameters.enable_tile_classifier:
+            logger.info("Tile classifier enabled")
+            logger.info(f"Patch model from: {config.model.type} to CustomMaskRCNNTileOptimised")
+            config.model.type = "CustomMaskRCNNTileOptimised"
+
+            if config.model.backbone.type == "efficientnet_b2b":
```
sungmanc, line 363: This patching seems to be avoided. It is easy to miss that the learning rate is changed.
Could we make this part to be more explicit way?

otx/api/configuration/helper/utils.py
```diff
@@ -187,3 +187,32 @@ def config_to_bytes(config: ConfigurableParameters) -> bytes:
...
+    """
+    for key, value in config.items():
+        if isinstance(value, dict):
+            if "value" in value:
+                config[key] = value["value"]
+            else:
+                flatten_config_values(value)
+
+
+def group_to_dict(config: dict):
```
sungmanc: If this function only targets the detection task, how about changing the name?

sungmanc (commented):
To be honest, I couldn't fully understand the logic. BTW, I left some comments

otx/algorithms/detection/adapters/mmdet/models/detectors/custom_maskrcnn_tile_optimised.py
eugene123tw: Done

otx/algorithms/detection/adapters/mmdet/models/detectors/custom_maskrcnn_tile_optimised.py
eugene123tw: I have changed this part. Although, not sure if it's the style you referred to.

otx/api/configuration/helper/utils.py
eugene123tw: I renamed it.

jaegukhyun (changes requested):
https://github.com/openvinotoolkit/training_extensions/pull/1955 Overall detection task is refactored. You need to move your PR's content to new task structure. I know this is quite annoying, but I hope you understand this situation.

jaegukhyun (changes requested):
Detection task test took so many long time. Could you check it in local?

otx/algorithms/detection/adapters/mmdet/task.py
```diff
@@ -199,6 +200,7 @@ def _init_task(self, dataset: Optional[DatasetEntity] = None, export: bool = Fal
         # Patch tiling parameters
         patch_tiling(self._recipe_cfg, self._hyperparams, dataset)
 
+        set_hyperparams(self._recipe_cfg, self._hyperparams)
```
jaegukhyun: Do we need this? Hyperparameters are updated in line#169.

jaegukhyun (commented):
Why don't you remove unnecessary changes after detection task refactoring?

otx/algorithms/detection/adapters/mmdet/task.py
eugene123tw: Thanks for letting me know! I added some changes.

otx/algorithms/detection/task.py
```diff
@@ -98,7 +98,44 @@ def __init__(self, task_environment: TaskEnvironment, output_path: Optional[str]
...
+                hparams.enable_tile_classifier = loaded_tiling_parameters["enable_tile_classifier"]["value"]
+
     def _load_model_ckpt(self, model: Optional[ModelEntity]):
+        """Load model checkpoint from model entity.
+
+        Args:
+            model (Optional[ModelEntity]): The model entity.
+
+        Returns:
+            _type_: _description_
```
jaegukhyun: Could you check this description?

jaegukhyun (commented):
I left one comment!

otx/algorithms/detection/task.py
eugene123tw: Done.

jaegukhyun (dismissed):

jaegukhyun (dismissed):

sungmanc (changes requested):
Need to resolve conflicts.
BTW, AFAIK, you have been experimenting with a lower learning rate to avoid the patching.
If the experimental score is quite good, will you introduce a new learning rate in this PR?

jaegukhyun (dismissed):

otx/algorithms/detection/task.py
```diff
@@ -290,6 +320,20 @@ def export(
             output_model.set_data("openvino.xml", f.read())
         with open(onnx_file, "rb") as f:
             output_model.set_data("model.onnx", f.read())
+
+        if self._hyperparams.tiling_parameters.enable_tile_classifier:
+            tile_classifier = None
+            for partition in outputs.get("partitioned"):
```
supersoob: Minor, but suggest -> outputs.get("partitioned", {})

otx/api/utils/tiler.py
```diff
@@ -68,87 +100,98 @@ def predict(self, image: np.ndarray):
...
+            image (np.ndarray): full size image
 
         Returns:
+            detection: prediction results
             features: saliency map and feature vector
-            output: single tile prediction
         """
+        tile_coords = self.tile(image)
+        if isinstance(self.classifier, Model):
+            tile_coords = self.filter_tiles_by_objectness(image, tile_coords)
```
supersoob: I think these lines are duplicated with `predict_sync`. Filtering tiles from tiling classifier uses `infer_sync`, which doesn't seems appropriate inside `predict_async`. I don't know well but if possible, it'd be better to move them into `predict` and then branch off into each `predict_sync` and `predict_async` with input of tile_coords. If not possible and hard, it's fine to let as is.

supersoob (commented):
Nice work:) Thanks for your efforts. I have minor comments.

tests/integration/cli/detection/test_tiling_instseg.py
```diff
@@ -34,11 +37,28 @@
         "4",
         "--tiling_parameters.enable_tiling",
         "1",
+        "--tiling_parameters.enable_tile_classifier",
```
jaegukhyun: Shouldn't we check the non-tile-classifier cli test?

eugene123tw: I could run additional testing, but doing so would double the testing time. My reasoning is that since enabling the tile classifier already covers the tiling test, running extra tests is unnecessary.

otx/algorithms/detection/task.py
eugene123tw: Thanks! That's a great catch!

otx/api/utils/tiler.py
eugene123tw: Thank you for your feedback. I agree that it'd be ideal to have the code structured in the way you suggested, but as you mentioned, it may not be practical to implement at this time and I need some time to think about the overall design. I appreciate your suggestion and I will add a TODO in the code to revisit this in the future when feasible.

eugene123tw:
> Need to resolve conflicts. BTW, AFAIK, you have been experimenting with a lower learning rate to avoid the patching. If the experimental score is quite good, will you introduce a new learning rate in this PR?

I have conducted a study comparing the learning rate on EfficientNet for the Vitens Coliform dataset. The results indicate a significant 5% accuracy gap between the different learning rates. It's worth noting that the training on the lower learning rate was terminated earlier due to early stop. While this outcome is not surprising, I believe that the 5% gap is substantial enough to keep the LR patching approach.

otx/api/utils/tiler.py
eugene123tw: Thanks! That's great advice. I've refactored accordingly.

goodsong81 (approved):
LTGM! Thank you for the great feature & enhancement.

sungmanc (approved):
Thanks for the work

PR #1892, opened 2023-03-16
[FIX] Handle unpickable update_progress_callback

cih9088:
### Summary
`update_progress_callback` is not always a pickable object.

### Checklist

### License

```python
#
```

### DIFF
--- a/otx/api/usecases/reporting/time_monitor_callback.py
+++ b/otx/api/usecases/reporting/time_monitor_callback.py
@@ -9,9 +9,14 @@
 import logging
 import math
 import time
-from typing import List, Optional
+from typing import List
 
-from otx.api.entities.train_parameters import UpdateProgressCallback
+import dill
+
+from otx.api.entities.train_parameters import (
+    UpdateProgressCallback,
+    default_progress_callback,
+)
 from otx.api.usecases.reporting.callback import Callback
 
 logger = logging.getLogger(__name__)
@@ -27,7 +32,7 @@ class TimeMonitorCallback(Callback):
         num_test_steps (int): amount of testing steps
         epoch_history (int): Amount of previous epochs to calculate average epoch time over
         step_history (int): Amount of previous steps to calculate average steps time over
-        update_progress_callback (Optional[UpdateProgressCallback]): Callback to update progress
+        update_progress_callback (UpdateProgressCallback): Callback to update progress
     """
 
     def __init__(
@@ -38,7 +43,7 @@ class TimeMonitorCallback(Callback):
         num_test_steps: int = 0,
         epoch_history: int = 5,
         step_history: int = 50,
-        update_progress_callback: Optional[UpdateProgressCallback] = None,
+        update_progress_callback: UpdateProgressCallback = default_progress_callback,
     ):
 
         self.total_epochs = num_epoch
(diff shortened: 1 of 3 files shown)


goodsong81 (approved):
Thank you for the fix!

otx/api/usecases/reporting/time_monitor_callback.py
```diff
@@ -67,6 +67,13 @@ def __init__(
 
         self.update_progress_callback = update_progress_callback
 
+    def __getstate__(self):
+        """Return state values to be pickled."""
+        state = self.__dict__
```
goodsong81: Shouldn't we copy? `state = self.__dict__.copy()`
BTW, tests are failing due to NoneType error possibly due to this function.
https://github.com/openvinotoolkit/training_extensions/actions/runs/4434177685
Can we just remove this from pickling without actual progress callback funcionality?

goodsong81 (changes requested):
There seems some test failures. Could you take a look?

otx/api/usecases/reporting/time_monitor_callback.py
cih9088: Thank you for point it out!
The pickled object now has default callback.

otx/api/usecases/reporting/time_monitor_callback.py
```diff
@@ -67,6 +72,29 @@ def __init__(
...
+            self.total_epochs,
+            self.train_steps,
+            self.val_steps,
+            self.test_steps,
+            self.epoch_history,
+            self.step_history,
+            self.update_progress_callback,
+        )
+        memo[id(self)] = result
+        return result
```
goodsong81: Could you let me know why this is needed?
(And the line#95)
Thanks!

goodsong81 (dismissed):

otx/api/usecases/reporting/time_monitor_callback.py
cih9088: `__deepcopy__` is a special method for `copy.deepcopy` and unless it is implemented, `copy.deepcopy` falls back to use `__[gs]etstate__` indirectly. Since we implemented `__getstate__`, `copy.deepcopy` will use `__getstate__`, which means the deep copied object could not be a homogeneous object as we replaced the unpickable `update_progress_callback` object with sensible default one `default_progress_callback`. To avoid this, we should explicitly implement `__deepcopy__` method.

`memo` keeps track of what has been copied. It is to prevent recursive infinite loop if the object is recursive. Please refer [here](https://docs.python.org/3/library/copy.html#copy.deepcopy).

goodsong81: Thank you for kind explanation. Understood clearly!

goodsong81 (dismissed):

goodsong81 (changes requested):
I've made an workaround for Geti integration and merged to release branch & Geti. https://github.com/openvinotoolkit/training_extensions/pull/1895
But I think this fix still needed. Could you redirect this PR to develop branch? It's not that urgent, though. Thank you for your contribution!

eunwoosh (approved):

goodsong81 (approved):

PR #1896, opened 2023-03-16
[FIX] Prevent torch 2.0.0 installation

JihwanEom:
### Summary
- Add `torchvision` & `torchtext` in `requirements/anomaly.txt` with fixed version (0.14.1) to prevent torch 2.0.0 installation

### How to test
With cuda 11.7 environment:
```
pip install torch==1.13.1 torchvision==0.14.1 --extra-index-url https://download.pytorch.org/whl/cu117
pip install -e .[full]
```

### Checklist

### License

- [ ] I submit _my code changes_ under the same [MIT License](https://github.com/openvinotoolkit/training_extensions/blob/develop/LICENSE) that covers the project.
- [ ] I have updated the license header for each file (see an example below)

```python
#
```

### DIFF
--- a/requirements/anomaly.txt
+++ b/requirements/anomaly.txt
@@ -1,6 +1,6 @@
 # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
 # Anomaly Requirements.
-torchvision<=0.14.1
-torchtext<=0.14.1
+torchvision<0.15.1
+torchtext<0.15.1
 anomalib==0.4.0
 kornia==0.6.9
--- a/requirements/anomaly.txt
+++ b/requirements/anomaly.txt
@@ -1,4 +1,6 @@
 # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
 # Anomaly Requirements.
+torchvision==0.14.1
+torchtext==0.14.1
 anomalib==0.4.0rc2  # FIXME: -> 0.4.0
 kornia==0.6.9
--- a/requirements/anomaly.txt
+++ b/requirements/anomaly.txt
@@ -1,6 +1,6 @@
 # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
 # Anomaly Requirements.
-torchvision==0.14.1
-torchtext==0.14.1
-anomalib==0.4.0rc2  # FIXME: -> 0.4.0
+torchvision<=0.14.1
+torchtext<=0.14.1
+anomalib==0.4.0
 kornia==0.6.9


requirements/anomaly.txt
```diff
@@ -1,4 +1,6 @@
 # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
 # Anomaly Requirements.
+torchvision==0.14.1
```
goodsong81: Do we need this, too?

JihwanEom: Yes, without this, torchvision==0.15.1 and torch==2.0.0 are installed.

ashwinvaidya17: I am a bit hesitant about fixed requirements. Can't we do something like `torchvision<0.15.1`? Also, does `torchtext` need to be specified?

djdameln: What if we fix the torch version instead and let the installer sort out the torchvision and torchtext versions?

goodsong81: > What if we fix the torch version instead and let the installer sort out the torchvision and torchtext versions?

Currently, we regard the torch as pre-requisite. Users might want to use torch1.12.1. I would prefer @ashwinvaidya17's approach.

JihwanEom: umm.. I'm also not exactly sure about these changes. This PR is hotfix. (Absolutely opened for good ideas!)
If we can control the acceptable range of version in anomaly side, it looks best to me. https://github.com/openvinotoolkit/anomalib/blob/v0.4.0rc2/requirements/base.txt#L15-L16

goodsong81: > I am a bit hesitant about fixed requirements. Can't we do something like `torchvision<0.15.1`? Also, does `torchtext` need to be specified?

torchtext>=0.15.1 would install torchdata that would force to install torch==2.0. Looks like mistake.

goodsong81: image

requirements/anomaly.txt
```diff
@@ -1,4 +1,6 @@
...
+torchtext==0.14.1
```
samet-akcay: well, now that you just modified this file, maybe you could also replace `anomalib==0.4.0rc2` with `anomalib==0.4.0` too? 😁

JihwanEom:
CI failed by codecov coverage upload issue, but passed all of unit tests. (known issue)

requirements/anomaly.txt
JihwanEom: Reflected! :)

goodsong81 (approved):
Nice catch! Thank you for the quick fix.

PR #1908, opened 2023-03-20
Add mmcls.VisionTransformer backbone support

harimkang:
### Summary

New Backbone Support
- mmcls.VisionTransformer
- mmcls.T2T_ViT
- mmcls.Conformer

mmcls.VisionTransformer support
- [X] otx find with mmcls transformer backbones
- [X] otx build with mmcls transformer backbones
- [X] otx train with mmcls transformer backbones

Things to know:
The current implementation uses `dump_feature` and `dump_saliency_map` as **False** in infer when using `VisionTransformerClsHead` due to the different output structure of transformer.

### How to test

```
(otx) otx find --task classification --backbone mmcls
(otx) otx build --task classification
(otx) cd otx-workspace-CLASSIFICATION
# Edit data.yaml
(otx) otx build --backbone mmcls.VisionTransformer
(otx) otx train
```

### Checklist

### License

- [X] I submit _my code changes_ under the same [MIT License](https://github.com/openvinotoolkit/training_extensions/blob/develop/LICENSE) that covers the project.
- [X] I have updated the license header for each file (see an example below)

```python
#
```

### DIFF
--- a/otx/algorithms/classification/configs/configuration.yaml
+++ b/otx/algorithms/classification/configs/configuration.yaml
@@ -10,7 +10,7 @@ learning_parameters:
       stable. A larger batch size has higher memory requirements.
     editable: true
     header: Batch size
-    max_value: 512
+    max_value: 2048
     min_value: 1
     type: INTEGER
     ui_rules:
--- a/otx/cli/builder/builder.py
+++ b/otx/cli/builder/builder.py
@@ -28,8 +28,8 @@ import torch
 from mmcv.utils import Registry, build_from_cfg
 from torch import nn
 
-from otx.api.entities.model_template import TaskType
 from otx.algorithms import TRANSFORMER_BACKBONES
+from otx.api.entities.model_template import TaskType
 from otx.cli.utils.importing import (
     get_backbone_list,
     get_backbone_registry,
@@ -102,8 +102,8 @@ def update_backbone_args(backbone_config: dict, registry: Registry, backend: str
 
 def update_channels(model_config: MPAConfig, out_channels: Any):
     """Update in_channel of head or neck."""
-    if hasattr(model_config.model, "neck"):
-        if model_config.model.neck.type == "GlobalAveragePooling":
+    if hasattr(model_config.model, "neck") and model_config.model.neck:
+        if model_config.model.neck.get("type", None) == "GlobalAveragePooling":
             model_config.model.neck.pop("in_channels", None)
         else:
             print(f"\tUpdate model.neck.in_channels: {out_channels}")
--- a/otx/cli/builder/builder.py
+++ b/otx/cli/builder/builder.py
@@ -29,6 +29,7 @@ from mmcv.utils import Registry, build_from_cfg
 from torch import nn
 
 from otx.api.entities.model_template import TaskType
+from otx.algorithms import TRANSFORMER_BACKBONES
 from otx.cli.utils.importing import (
     get_backbone_list,
     get_backbone_registry,
@@ -212,6 +213,12 @@ class Builder:
             out_channels = -1
             if hasattr(model_config.model, "head"):
                 model_config.model.head.in_channels = -1
+            # TODO: This is a hard coded part of the Transformer backbone and needs to be refactored.
+            if backend == "mmcls" and backbone_class in TRANSFORMER_BACKBONES:
+                if hasattr(model_config.model, "neck"):
+                    model_config.model.neck = None
+                if hasattr(model_config.model, "head"):
+                    model_config.model.head["type"] = "VisionTransformerClsHead"
         else:
             # Need to update in/out channel configuration here
             out_channels = get_backbone_out_channels(backbone)
--- a/otx/mpa/cls/inferrer.py
+++ b/otx/mpa/cls/inferrer.py
@@ -56,6 +56,7 @@ class ClsInferrer(ClsStage):
         dump_saliency_map = kwargs.get("dump_saliency_map", False)
         # TODO: It looks like we need to modify that code in an appropriate way.
         if model_cfg.model.head.get("type", None) == "VisionTransformerClsHead":
+            dump_features = False
             dump_saliency_map = False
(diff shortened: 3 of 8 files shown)


goodsong81 (dismissed):
Thank you for the quick fix.

otx/algorithms/classification/configs/configuration.yaml
```diff
@@ -10,7 +10,7 @@ learning_parameters:
       stable. A larger batch size has higher memory requirements.
     editable: true
     header: Batch size
-    max_value: 512
+    max_value: 2048
```
sungmanc, line 13: Why should we change the max batch size in this PR? is it related to transformer?

sungmanc (commented):
Shouldn't we check whether these backbones are successfully represented at `otx find` or not?

otx/algorithms/classification/configs/configuration.yaml
harimkang: Yes, that's right. It says we need to raise that max value to support the VisionTransformer-large model. (report from @JihwanEom)

harimkang:
> Shouldn't we check whether these backbones are successfully represented at `otx find` or not?

The existing branch cannot find mmcls.VisionTransformer in otx find, but it will be added in this PR.
image
However, I still need to verify that this works properly, so there may be additional changes in this PR.

harimkang:
@JihwanEom @sungchul2 This will not work with these settings at this time, please let us know what additional settings need. I want to make the command below work.
```
(otx) otx build --task classification
(otx) cd otx-workspace-CLASSIFICATION
# Edit data.yaml
(otx) otx build --backbone mmcls.VisionTransformer
(otx) otx train
```

otx/algorithms/classification/configs/configuration.yaml
JihwanEom: Not limited for ViTs, 512+ batch size configuration does not work on current develop..

sungchul1: Actually, I'd like to know why these min/max values are in config (batch size, lr, ...).
I think there are some cases that might require values that are out of the range, and in these cases, these restricted ranges are inconvenient and too complicated to handle from the user's point of view.

harimkang: I'm guessing that's probably an attribute for the geti UI? I agree with Sungchul's comment. But this is probably tied to geti. We need to check that first.

goodsong81 (changes requested):
Not to merge.

otx/mpa/cls/inferrer.py
```diff
@@ -53,6 +54,10 @@ def run(self, model_cfg, model_ckpt, data_cfg, **kwargs):
         model_builder = kwargs.get("model_builder", None)
         dump_features = kwargs.get("dump_features", False)
         dump_saliency_map = kwargs.get("dump_saliency_map", False)
+        # TODO: It looks like we need to modify that code in an appropriate way.
+        if model_cfg.model.head.get("type", None) == "VisionTransformerClsHead":
```
harimkang, line 58: @sungchul2 Due to the unusual output structure of the transformer, I first set dump_feature and dump_saliency_map to False. This may result in some things not working. However, I have no issues with training and exporting. (Of course, I've only checked with the toy dataset, so it seems like a good double check to make sure it's working).

otx/algorithms/classification/configs/configuration.yaml
sungchul1: If that's the reason, shouldn't it be set in this PR?
To change it, it seems we need to talk about it with geti, doesn't it?

goodsong81 (approved):
LGTM! Welcome new backbones!

otx/cli/builder/builder.py
```diff
@@ -212,6 +213,12 @@ def merge_backbone(
             out_channels = -1
             if hasattr(model_config.model, "head"):
                 model_config.model.head.in_channels = -1
+            # TODO: This is a hard coded part of the Transformer backbone and needs to be refactored.
+            if backend == "mmcls" and backbone_class in TRANSFORMER_BACKBONES:
+                if hasattr(model_config.model, "neck"):
+                    model_config.model.neck = None
+                if hasattr(model_config.model, "head"):
+                    model_config.model.head["type"] = "VisionTransformerClsHead"
```
sungchul1, line 221: Doesn't this part need to use `CustomClsHead`?'
I think modified `CustomClsHead` for `VisionTransformerClsHead` is needed.

otx/mpa/cls/inferrer.py
sungchul1: I checked there is no issue using cifar10.
But, are these hooks (`FeatureVectorHook`, `ReciproCAMHook`) not needed for ViT models?

sungchul1 (commented):
Thanks for your hard work. I left some questions and let me know the answers.

otx/cli/builder/builder.py
harimkang: Are you saying that we should support VisionTransformer in `CustomClsHead`?

harimkang:
> Thanks for your hard work. I left some questions and let me know the answers.

@goodsong81 I modified the max_value of batch size in configuration.yaml, will this affect Geti? 

@sungchul2 Hooks (`FeatureVectorHook`, `ReciproCAMHook`) may need to be added for 'otx explain' functionality, but I'm not familiar with those hooks, so I've disabled it for now. This is something that needs to be discussed, but for now training and exporting are fine. @goodsong81 Should we support the explain of the Transformer model in this PR?

otx/cli/builder/builder.py
sungchul1: No. I'm just curious.

harimkang: For now, I only modified the otx build to change the model.py head type to `VisionTransformerClsHead` when using `mmcls.VisionTransformer`.

sungchul1 (approved):

PR #1910, opened 2023-03-20
Fix OTX1.1 -> Geti1.4 integration issues

goodsong81:
### Summary

* Add dill to api dependency for `otx.api`-only use-case
  (Originally, installed by mmdeploy)
* Return None instead of raising exception
  (Geti does not implement `path` method)

### Checklist

### License

- [ ] I submit _my code changes_ under the same [MIT License](https://github.com/openvinotoolkit/training_extensions/blob/develop/LICENSE) that covers the project.
- [ ] I have updated the license header for each file (see an example below)

```python
#
```

### DIFF
--- a/otx/api/entities/media.py
+++ b/otx/api/entities/media.py
@@ -54,4 +54,4 @@ class IMedia2DEntity(IMediaEntity, metaclass=abc.ABCMeta):
     @property
     def path(self) -> Optional[str]:
         """Returns the path of the 2D Media object."""
-        raise NotImplementedError
+        return None
--- a/requirements/api.txt
+++ b/requirements/api.txt
@@ -9,3 +9,4 @@ pymongo==3.12.0
 scikit-learn==0.24.*
 Shapely>=1.7.1,<=1.8.0
 imagesize==1.4.1
+dill>=0.3.6


eunwoosh (approved):

harimkang (approved):

otx/api/entities/media.py
```diff
@@ -54,4 +54,4 @@ def width(self) -> int:
     @property
     def path(self) -> Optional[str]:
         """Returns the path of the 2D Media object."""
-        raise NotImplementedError
```
sovrasov, line 57: It's an abstract class, so every method should be overloaded in child classes. Why do we need changing this?

sovrasov: Is there any ticket to Geti SDK to fix that?

goodsong81: No Geti SDK or SC SDK tickets.
Actually, `path` method had been added for OTX-only use-case while OTX1.0 development. SC SDK does not implement it so that abstract method resulted in error. Here again, some module called `path` but it was actually sc_sdk.entities.Image2D. Removing `path` might need some time.
@sungmanc idea?

sungmanc: @goodsong81 @sovrasov 

To assure consistency with SC SDK, we could remove all `path` variable inside of `OTX`.
Maybe this could affect the `Action` tasks. I'll check it later and let you know

PR #1951, opened 2023-03-28
Segmentation task enhancement + new templates (SegNext)

kprokofi:
### Checklist

### License

```python
```


kprokofi:
- [x] Added SegNext family as new templates
- [x] Changed incremental recepie to enhance current models
- [x] Refactored head and add universal head to support different heads from mmseg with our interface
- [x] Fixed label sorting (consider changing sorting only in segmentation task, not in Entity)
- [x] Added annotations from file support for Common Semantic Segmentation format (it is useful for debugging OTX dataset and utilization of some academic datasets and masks which isn't supported by Shapely)
- [x] Added checkpoint patching to support pre-trained weights from mmseg
- [x] Fixed unit tests, pre-commit and pre-merge. Add new if required

otx/algorithms/common/adapters/mmcv/configs/backbones/deeplabv3plus_resnet50.py
```diff
@@ -0,0 +1,47 @@
+norm_cfg = dict(type="BN", requires_grad=True)
+model = dict(
```
sungmanc: This implementation seems to include all of model architecture although the folder name is `backbones`.
We need to divided each part to proper `folder`.

otx/algorithms/common/adapters/mmcv/tasks/stage.py
```diff
@@ -486,12 +487,34 @@ def get_model_classes(cfg):
     def get_model_ckpt(ckpt_path, new_path=None):
         """Return model ckpt from ckpt_path."""
         ckpt = CheckpointLoader.load_checkpoint(ckpt_path, map_location="cpu")
+        local_torch_hub_folder = torch.hub.get_dir()
         if "model" in ckpt:
             ckpt = ckpt["model"]
             if not new_path:
                 new_path = ckpt_path[:-3] + "converted.pth"
-            torch.save(ckpt, new_path)
+            if not torch.distributed.is_initialized() or dist.get_rank() == 0:
```
sungmanc: Have this line side effect to multi-gpu training? @eunwoosh

otx/algorithms/segmentation/adapters/mmseg/__init__.py
```diff
@@ -48,7 +48,7 @@
     "MPASegDataset",
     "LiteHRNet",
     "MMOVBackbone",
-    "CustomFCNHead",
+    "get_head",
```
sungmanc: Do we need to add `get_head` function in the `__init__.py`?

otx/algorithms/segmentation/adapters/mmseg/models/__init__.py
```diff
@@ -33,7 +33,7 @@
 __all__ = [
     "LiteHRNet",
     "MMOVBackbone",
-    "CustomFCNHead",
+    "get_head",
```
sungmanc: Do we need to add this line in `__init__.py`? Will we use `get_head` function as like module?

otx/algorithms/segmentation/adapters/mmseg/models/heads/mixin.py
```diff
@@ -23,42 +23,21 @@
...
-        if self.enable_out_norm:
-            feat = normalize(feat, dim=1, p=2)
-        if self.conv_seg is not None:
-            return self.conv_seg(feat)
-        return feat
-
-
-class AggregatorMixin(nn.Module):
-    """A class for creating an aggregator."""
+class Mixin(nn.Module):
```
sungmanc: The name of this class needs to be more specific. How about `SegHeadMixin` or `SegMixin`

otx/algorithms/segmentation/adapters/mmseg/models/__init__.py
sungmanc: Basically, `get_head()` function itself is useful to avoid duplicate implementation.

However, adding `get_head()` function in `__init__.py` and `heads/` looks inappropriate to me. 
It seems more pretty to locate `get_head()` function to `utils/` folder and implement `CustomOTXHead` in `heads/`, `__init__.py`.

Want to listen to other colleague's opinion. 
What do you think about this implementation? @supersoob , @eunwoosh

otx/algorithms/segmentation/configs/base/data/data_pipeline.py
```diff
@@ -37,12 +36,13 @@
...
     dict(type="Normalize", **__img_norm_cfg),
     dict(type="Pad", size=__crop_size, pad_val=0, seg_pad_val=255),
-    dict(type="RandomRotate", prob=0.5, degree=30, pad_val=0, seg_pad_val=255),
     dict(type="DefaultFormatBundle"),
     dict(type="Collect", keys=["img", "gt_semantic_seg"]),
 ]
+
 test_pipeline = [
     dict(type="LoadImageFromFile"),
+    dict(type="LoadFromFiles", flag=False),
```
sungmanc: Could we remove this line and then edit the `LoadImageFromFile`?

It seems to confuse users since `LoadImageFromFile` and `LoadFromFiles` looks very similar.

```python3
dict(type="LoadImageFromFile", use_otx_adapter=False)
```

sungmanc (commented):
Thanks for a huge contribution :) 
The first thing I want to say that please use the forked repository after this PR.
And I left some opinions and comments.

otx/algorithms/common/adapters/mmcv/tasks/stage.py
supersoob: I think this line is needed because only rank 0 should save the ckpt. Could you double-check? @eunwoosh

otx/algorithms/segmentation/adapters/mmseg/datasets/dataset.py
```diff
@@ -118,16 +119,18 @@ def __init__(
...
         test_mode: bool = False,
+        load_from_files: bool = False,
     ):
         self.otx_dataset = otx_dataset
         self.test_mode = test_mode
 
         self.ignore_index = 255
         self.reduce_zero_label = False
         self.label_map = None
+        self.load_from_files = load_from_files
```
supersoob: This load_from_files is for direct ann from ann file to prevent current datumaro mask disappearing issue. Is it right? So you set it False as default.

otx/api/entities/datasets.py
```diff
@@ -417,4 +417,4 @@ def get_labels(self, include_empty: bool = False) -> List[LabelEntity]:
             List[LabelEntity]: list of labels that appear in the dataset
         """
         label_set = set(itertools.chain(*[item.annotation_scene.get_labels(include_empty) for item in self]))
-        return list(label_set)
+        return sorted(list(label_set), key=lambda x: int(x.id))
```
supersoob: Could you explain why we need this? It looks like it can affects to other tasks.

otx/algorithms/segmentation/configs/resnet50_deeplabv3plus/template.yaml
harimkang, line 1: I believe the name `template.yaml` is probably ingested by Geti. Wouldn't this be a problem?

otx/api/utils/segmentation_utils.py
```diff
@@ -67,7 +75,6 @@ def mask_from_annotation(
         2d numpy array of mask
     """
 
-    labels = sorted(labels)  # type: ignore
```
supersoob, line 70: @ashwinvaidya17 @samet-akcay Could you double-check if this can affect to anomalib or geti? Because of this line, we had a problem with label mismatch because it was sorted in order of string in id. (e.g. id(0) -> id(1) -> id(10) -> id(11) -> ...) So we have aligned it in string order sort in all codes. But it seems like users will be confused if model head label order and label order in log is different from label json file that the user provided. Thus, we'd like to align it in the label order as user input.

otx/recipes/stages/segmentation/incremental.py
```diff
@@ -1,29 +1,29 @@
...
-        # clip=0.2,
-        # method='default',
-        max_norm=40,
-        norm_type=2,
-    ),
+    type="AdamW",
+    lr=0.0005,
+    betas=(0.9, 0.999),
+    weight_decay=1e-4,
+    paramwise_cfg={"bias_decay_mult ": 0.0, "norm_decay_mult ": 0.0},
```
supersoob: Could you explain what is this?

otx/algorithms/segmentation/adapters/mmseg/datasets/dataset.py
```diff
@@ -257,12 +261,18 @@ def __init__(self, **kwargs):
         if test_mode is False:
             self.img_indices = get_old_new_img_indices(classes, new_classes, otx_dataset)
 
+        if test_mode:
+            for pipe in pipeline:
+                if pipe["type"] == "LoadFromFiles":
```
supersoob: Actually I don't know what this is for.. "LoadFromFiles", Why especially only for test? Could you explain?

otx/algorithms/common/adapters/mmcv/configs/backbones/deeplabv3plus_resnet50.py
```diff
@@ -0,0 +1,47 @@
+norm_cfg = dict(type="BN", requires_grad=True)
```
supersoob: I hope you align the name "resnet50_deeplabv3plus" like template name.

otx/algorithms/segmentation/adapters/mmseg/models/heads/custom_otx_head.py
```diff
@@ -0,0 +1,32 @@
...
+# Copyright (C) 2023 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
+
+from mmseg.models.decode_heads.fcn_head import FCNHead
+from mmseg.models.decode_heads.sep_aspp_head import DepthwiseSeparableASPPHead
+
+from .mixin import Mixin
+
+KNOWN_HEADS = {"FCNHead": FCNHead, "ASPPHead": DepthwiseSeparableASPPHead}
```
supersoob: Acutally I'd like to revert OCRHead as well. It is effective for large dataset with lite hrnet. But not this PR's scope. Let me do later after PoC.

supersoob (commented):
Thanks for refactoring the FCN head. Nice work! I left some comments.

otx/algorithms/common/adapters/mmcv/configs/backbones/deeplabv3plus_resnet50.py
supersoob: I don't know what is right, but maybe deeplabv3plus_resnetXX is more appropriate because names of old templates are ocr_lite_hrnet_XX(backbone name later). Please align them.

otx/algorithms/segmentation/configs/resnet50_deeplabv3plus/template.yaml
```diff
@@ -0,0 +1,61 @@
+# Description.
+model_template_id: Custom_Semantic_Segmentation_ResNet50_ASPP
```
supersoob: confusing name.. deeplabv3plus_resnet50? or aspp_resent50?

supersoob: Like I commented above, I think we need to align the name of template and file name.

otx/algorithms/segmentation/configs/resnet50_deeplabv3plus/template.yaml
supersoob: I'm also curious about this. Is this okay to leave as template.yaml? I wonder what the next procedure is to update the template in geti.

otx/algorithms/segmentation/adapters/mmseg/tasks/stage.py
```diff
@@ -136,6 +142,17 @@ def configure_classes(self, cfg, task_adapt_op):
...
+    def configure_head(self, cfg):
+        """Change head to custom head to align with otx
+        inteface for class incremental learning."""
+        decode_head = cfg.model.get("decode_head", None)
+        auxiliary_head = cfg.model.get("auxiliary_head", None)
+
+        for head in (decode_head, auxiliary_head):
+            if head:
+                head.head_name = head.type
+                head.type = get_head
```
supersoob: I think this code flow is quite confusing.. I don't think head.type is to be custom function for mmcv-base implementation. I hope we can deal all in OTXCustomHead class.

otx/algorithms/segmentation/adapters/mmseg/models/__init__.py
supersoob: I agree with @sungmanc. I hope we can deal with head init in CustomOTXHead.

otx/algorithms/common/adapters/mmcv/tasks/stage.py
kprokofi: This line actually prevents problem with multi-gpu training. If we remove this line - the checkpoint will be corrupted. Adding line using only 1 gpu with rank=0 will solve this issue, in case of single gpu training - first conditional won't let training fall with error

otx/algorithms/segmentation/adapters/mmseg/__init__.py
kprokofi: No, we don't need - I will remove

otx/algorithms/segmentation/adapters/mmseg/models/__init__.py
kprokofi: CustomOTXHead should be inherited from mmseg head class on the go - so, it can be apart from function. I can create Class fabric instead and define __call__ method to return newly created class inherited from base head class from mmseg. What do you thing?

otx/algorithms/segmentation/adapters/mmseg/models/heads/mixin.py
kprokofi: Agree, will change

otx/algorithms/common/adapters/mmcv/tasks/stage.py
eunwoosh: As soobee said, it is necessary part. I checked that multi GPU training works well with this change. Thanks for detailed change Kirill.

otx/algorithms/segmentation/configs/base/data/data_pipeline.py
kprokofi: Hmmm, yes, I can add this to LoadImageFromFile thank you. My solution is a bad workaround, I agree

otx/algorithms/segmentation/configs/resnet50_deeplabv3plus/template.yaml
kprokofi: Let's rename new templates to experimental first. The next step is to adapt different training types and then discuss with Geti team about substituting default models

otx/algorithms/segmentation/adapters/mmseg/datasets/dataset.py
kprokofi: Yes, by default it is False.
This place is needed only for test mode, because for training part we control loading annotations through augmentation

otx/api/entities/datasets.py
kprokofi: I thought about that - and also falling unit test in classification proved this. I will move it to segmentation task only

otx/api/utils/segmentation_utils.py
kprokofi: I will revert it actually to
`sorted(labels, key=lambda x: int(x.id))`, because I will remove sorting in entity
@samet-akcay, please let me know if it will affect anomalib

otx/recipes/stages/segmentation/incremental.py
kprokofi: it is No Bias Decay, it will remove decay for norm layers and bias

otx/algorithms/common/adapters/mmcv/tasks/stage.py
```diff
@@ -486,12 +487,34 @@ def get_model_classes(cfg):
...
+            for name in ckpt:
+                if not name.startswith("backbone") and not "fc" in name and not "head" in name:
+                    new_name = "backbone." + name
+                    modified = True
+                else:
+                    new_name = name
+                new_ckpt[new_name] = ckpt[name]
+            if modified:
+                if not new_path:
+                    new_path = local_torch_hub_folder + "/__converted.pth"
```
eunwoosh:
```suggestion
                    new_path = osp.join(local_torch_hub_folder, "converted.pth")
```
I think using `converted.pth` is better because it could be overwritten by other task, which can save storage.

otx/algorithms/segmentation/adapters/mmseg/datasets/dataset.py
kprokofi: I will align with Sungman proposal and remove "LoadFromFiles". But, originally, it is only for test mode since we can't controll loading annotation through augmentation (loading function). It is happened inside dataset instead. So, I added artificial pipeline function to testing pipeline.
However, it is a bad workaround. I would rather add additional parameter to "LoadImage" pipe and will use it here as well

kprokofi: For training mode - it is controlled by loading function in pipelines, much simpler to change

otx/algorithms/common/adapters/mmcv/configs/backbones/deeplabv3plus_resnet50.py
kprokofi: OK

otx/algorithms/segmentation/adapters/mmseg/models/heads/custom_otx_head.py
kprokofi: I hope it will simply work by adding here new head and change config, but let's see

otx/algorithms/segmentation/adapters/mmseg/tasks/stage.py
kprokofi: `head.type` can be a function actually. It can be class, function or string (registered head in Registry).
As I  said above - I can create class instead, that will return new class in callback metrhod, but it will be quite similar to current flow

otx/algorithms/common/adapters/mmcv/tasks/stage.py
```diff
@@ -486,12 +487,34 @@ def get_model_classes(cfg):
...
+                torch.save(ckpt, new_path)
             return new_path
+
+        elif "state_dict" in ckpt:
+            ckpt = ckpt["state_dict"]
+            new_ckpt = OrderedDict()
+            modified = False
+            # patch pre-trained checkpoint for model
+            for name in ckpt:
+                if not name.startswith("backbone") and not "fc" in name and not "head" in name:
```
eunwoosh: Could you add a comment in code to describe why it's necessary?

otx/algorithms/segmentation/adapters/mmseg/datasets/dataset.py
eunwoosh: if it's right, then is it right to make direct annotation available, which result in bypassing datumaro?
I think fixing bug is a right decision. How do you think about, @sungmanc ?

otx/algorithms/segmentation/adapters/mmseg/models/heads/mixin.py
eunwoosh: IMHO, SegHeadMixin isn't also proper name. It doesn't include any information. It's essential for `CustomOTXHead`, then why don't you  merge this class with `CustomOTXHead`?

otx/algorithms/segmentation/adapters/mmseg/tasks/stage.py
```diff
@@ -136,6 +142,17 @@ def configure_classes(self, cfg, task_adapt_op):
...
         self.model_classes = model_classes
 
+    def configure_head(self, cfg):
+        """Change head to custom head to align with otx
+        inteface for class incremental learning."""
+        decode_head = cfg.model.get("decode_head", None)
+        auxiliary_head = cfg.model.get("auxiliary_head", None)
+
+        for head in (decode_head, auxiliary_head):
+            if head:
```
eunwoosh:
```suggestion
            if head is not None:
```

otx/algorithms/segmentation/adapters/mmseg/datasets/dataset.py
kprokofi: It is good option for debugging OTX dataset + for now it is the only one way to achieve same perfomance compared with mmseg. I added this as temporary workaround + useful debugging functionality for the future. Also, please, note, that some datasets can't be fixed inside Datumaro. For example, DIS5K dataset has granularity in the annotations that Shapely can't support at all, I guess:
OTX:
image
Original:
image

kprokofi: So, I think it is OK to add this functionality. Of course we won't use it by default - it is just for debugging, some corner cases, some datasets
We should fix OTX dataset to achieve same perfomance on such datasets as VOC, Cityscapes, etc.

otx/algorithms/common/adapters/mmcv/tasks/stage.py
kprokofi: I don't mind this option. Actually I use "/__converted.pth" to not overwrite other snapshots.
But I'm okay to use just "converted.pth"

otx/algorithms/common/adapters/mmcv/tasks/stage.py
kprokofi: Sure, I will add comment in the code

otx/api/utils/segmentation_utils.py
ashwinvaidya17: From a quick initial test it does not seem to cause as issue as each dataset item has only one label associated with it. So this should be fine for anomaly tasks.

kprokofi: Thank you for your reply!

otx/algorithms/segmentation/adapters/mmseg/models/heads/mixin.py
kprokofi: I renamed it to `SegMixinModule`. Unfortunately, we can't merge CustomOTXHead with SegMixinModule since CustomOTXHead should be inherited from base head class from mmseg on the go. It should be done in class factory or in function. Class factory isn't possible, unfortunately due to registry limits inside mmseg.

otx/algorithms/segmentation/adapters/mmseg/tasks/stage.py
kprokofi: Oh no, we can't use Class factory unfortunately. Only function is possible to support inheritance on the go. But I simplified it a bit

otx/algorithms/segmentation/adapters/mmseg/tasks/stage.py
```diff
@@ -136,6 +139,19 @@ def configure_classes(self, cfg, task_adapt_op):
...
+
+        decode_head = cfg.model.get("decode_head", None)
+        auxiliary_head = cfg.model.get("auxiliary_head", None)
+
+        for head in (decode_head, auxiliary_head):
+            if head is not None:
+                # substitute head.type with OTX Head factory function
+                head_type = head.get("type", None)
+                if head_type:
+                    head_type = otx_head_factory
```
supersoob: We need to subtitute head.type as otx_head_factory also if it's none

supersoob: if head is not None:
                # substitute head.type with OTX Head factory function
                head_type = head.get("type", None)
                if head_type is not None and head_type != "otx_head_factory":
                    head.base_type = head_type
                    head.type = otx_head_factory
                elif head_type is None:
                    head.type = otx_head_factory

otx/algorithms/segmentation/configs/configuration.yaml
```diff
@@ -149,7 +149,7 @@ learning_parameters:
     warning: null
   enable_early_stopping:
     affects_outcome_of: TRAINING
-    default_value: true
+    default_value: false
```
supersoob: I don't think we need to disable this. If disabled, otx always have to iterate 100 epochs, which consumes too much e2e time. Please make it true, and I'm conducting experiments to prevent from stopping training and I will go over this part by increasing patience.

otx/algorithms/segmentation/adapters/mmseg/datasets/dataset.py
sungmanc: I agree with both opinions. Although we can leave this part as is for now, we need to resolve this issue in the future.
Maybe this issue comes from `Polygon->Mask` conversion, so we need to pretty much time to resolve this issue.
So, let's keep it as is.

otx/algorithms/segmentation/adapters/mmseg/datasets/dataset.py
```diff
@@ -257,12 +261,16 @@ def __init__(self, **kwargs):
         if test_mode is False:
             self.img_indices = get_old_new_img_indices(classes, new_classes, otx_dataset)
 
+        for pipe in pipeline:
+            if pipe["type"] == "LoadImageFromFile" and "use_otx_adapter" in pipe:
```
supersoob:
```suggestion
            if pipe["type"] == "LoadAnnotationFromOTXDataset" and "use_otx_adapter" in pipe:            
```
because that pipeline is subtituted in stage into XXFromOTXDataset

otx/algorithms/segmentation/adapters/mmseg/datasets/pipelines/loads.py
```diff
@@ -29,6 +29,11 @@
 class LoadImageFromOTXDataset(load_image_base.LoadImageFromOTXDataset):
     """Pipeline element that loads an image from a OTX Dataset on the fly."""
 
+    @check_input_parameters_type()
+    def __init__(self, to_float32: bool = False, use_otx_adapter: bool = True):
+        self.use_otx_adapter = use_otx_adapter
```
supersoob, line 32: Does image load need use_otx_adapter parameter? I suggest keep this parameter only in `LoadAnnotationFromOTXDataset`

otx/algorithms/segmentation/adapters/mmseg/datasets/dataset.py
```diff
@@ -257,12 +261,16 @@ def __init__(self, **kwargs):
...
+                use_otx_adapter = pipe["use_otx_adapter"]
+
```
supersoob, line 256: Please add `break` here inside the if-statement. It doesn't need to iterate all for updating use_otx_adapter.

otx/api/utils/segmentation_utils.py
```diff
@@ -37,13 +39,26 @@ class index is zero. labels[0] matches pixel value 1, etc. The class index is
...
+    """Loads masks directly from annotation image.
+
+    Only Common Sematic Segmentation format is supported.
+    """
+
+    mask_form_file = dataset_item.media.path
+    if mask_form_file is not None:
+        mask_form_file = mask_form_file.replace("images", "masks")
+        mask = cv2.imread(mask_form_file, cv2.IMREAD_GRAYSCALE)
+        mask = np.expand_dims(mask, axis=2)
```
supersoob: I hope you add the line to replace the image format to png if mask is None because the raw mask file format for voc is png.

    if mask_form_file is not None:
        mask_form_file = mask_form_file.replace("images", "masks")
        mask = cv2.imread(mask_form_file, cv2.IMREAD_GRAYSCALE)
        if mask is None:
              mask_form_file = mask_form_file.replace("jpg", "png")
              mask = cv2.imread(mask_form_file, cv2.IMREAD_GRAYSCALE)
        mask = np.expand_dims(mask, axis=2)

yunchu:
please add label for this PR.

otx/algorithms/segmentation/adapters/mmseg/datasets/pipelines/loads.py
kprokofi: Unfortunately, for **test** dataset we can control this parameter only via LoadImageFromOTXDataset since we don't use LoadAnnotationFromOTXDataset there

otx/algorithms/segmentation/configs/configuration.yaml
kprokofi: Yes sorry, it was just for debugging. I forgot to revert it back

otx/api/utils/segmentation_utils.py
kprokofi: According to Common Semantic Segmentation format - masks and images should be with same extension. So, it is not required. Other formats are not supported for this feature. Also, please, note that it is mainly for debugging, supporting some custom fine-grained and for fair comparison between mmseg and otx. So, I'm not sure that it is worth to spend time and support other formats.

otx/algorithms/segmentation/configs/deeplabv3plus_resnet18/compression_config.json
```diff
@@ -0,0 +1,62 @@
+{
```
sovrasov: Is this config a copy-paste from other models?

kprokofi: Yes, for now it is the same as Lite-HRNet-s and Lite-HRNet-x

kprokofi:
As I can see  XAI unit tests are failed for classification
image

kprokofi:
@supersoob , @eunwoosh , @sungmanc I replied to comments and aligned with new OTX develop. Could you take a look?
I struggling to fix last issues in pre-merge tests, but anyway if you have any comments - please add

kprokofi:
Tests for segmentation are fixed now. Other tests are failed due to problems with CI. 
Previous commits were successful on those tests. So, I think PR can be merged if no other comments

otx/recipes/stages/segmentation/incremental.py
```diff
@@ -1,29 +1,36 @@
...
-optimizer = dict(_delete_=True, type="Adam", lr=1e-3, eps=1e-08, weight_decay=0.0)
-
-optimizer_config = dict(
+optimizer = dict(
+    _delete_=True,
+    type="AdamW",
+    lr=0.0005,
+    betas=(0.9, 0.999),
+    weight_decay=0.0001,
+    paramwise_cfg={"bias_decay_mult ": 0.0, "norm_decay_mult ": 0.0},
```
supersoob: I hope you to revert all config changes in recipe file which is globally used because I'm doing the task optimizing the current hr-net based models without changing the lr configs. As for now, I think we need detailed exp results on various datasets for all models if hp is needed to changed. If this lr config and optimizer is best for new template deeplab, it would be better to separate configs and keep it in its own model.py in deeplab template.

otx/algorithms/segmentation/configs/ocr_lite_hrnet_x_mod3/template.yaml
```diff
@@ -29,14 +29,14 @@ hyper_parameters:
         default_value: 8
```
supersoob: Please revert base config change in current hrnet.

otx/algorithms/segmentation/configs/configuration.yaml
```diff
@@ -70,7 +70,7 @@ learning_parameters:
     affects_outcome_of: TRAINING
```
supersoob: Please revert all base config changes.

supersoob (changes requested):
Thanks for your hard work. But I have some comments.

1. Let's not make a change in all configurations until we update deeplabv3 as default template. I think hp change has to be carefully done with experiments on various datasets considering geti scenarios. This config has adopted for geti scenario with small datasets. And the optimization task for current hrnet models are in progress which finds the best hp and implement some algorithms to enhance just for current models. I hope this PR to contain only template change. Until this template becomes the default, I think it would be better to keep the hp you found for new model deeplab.

2. I hope you can share the performance results in small dataset(12, 24) and public VOC running with new deeplabv3 on OTX repo. We need to consider the geti scenario first.

otx/recipes/stages/segmentation/incremental.py
kprokofi: My work also includes changes for HRNets. This recipe shows  better results on all data. So, this is for all models, not only for Resnet

otx/algorithms/segmentation/configs/configuration.yaml
kprokofi: It is better to extend early stopping patience  for 10 epochs and turn it on a bit later. Now, it exits from training to early sometimes

otx/recipes/stages/segmentation/incremental.py
kprokofi: It is a part of my PoC. Current recipe shows much worse results

otx/algorithms/segmentation/configs/ocr_lite_hrnet_x_mod3/template.yaml
kprokofi: I can, but 300 default epochs looks like overhead. Maybe we can decrease this number?

kprokofi:
> Thanks for your hard work. But I have some comments.
> 
> 1. Let's not make a change in all configurations until we update deeplabv3 as default template. I think hp change has to be carefully done with experiments on various datasets considering geti scenarios. This config has adopted for geti scenario with small datasets. And the optimization task for current hrnet models are in progress which finds the best hp and implement some algorithms to enhance just for current models. I hope this PR to contain only template change. Until this template becomes the default, I think it would be better to keep the hp you found for new model deeplab.
> 2. I hope you can share the performance results in small dataset(12, 24) and public VOC running with new deeplabv3 on OTX repo. We need to consider the geti scenario first.

About number 2: I prepared entire presentation comparing models with all results..

otx/recipes/stages/segmentation/incremental.py
supersoob: After we had a gpu utilization experiments, we were in optimization stage to enhance lite-hrnet-fcn models considering both geti and otx. We had many experiments like changing heads architecture and loss function on various datasets and finalized, so now implementing new loss and configs for current hrnet models. If configs are updated now, I don't know how it affects the performance. I'd like to suggest to keep all current configs and change them when deeplab is updated as default template or update its own configs only in model.py in deeplab templates. How do you think about this?

kprokofi: This recipe gives 20% gain to VOC, few precents to small version of Kvasir, also gain for KITTI and cityscapes. I conducted experiments on wide range datasets. If we return it back we will be again with poorly performed models and won't be able even share results publicly.
So, This PR intended to change training recipe as well and as I remember we discussed this on meeting with you and Eunwoo few weeks ago.

So, could you explain about your concern about GPU utilization? I don't think that new scheduler will affect it

JihwanEom:
Thank you for all the work you've done so far! I have a couple of comments.

First,  the configuration of HRNet has been updated through numerous experiments, including the geti scenario. However, in this pull request, all hyperparameters are being changed for every template. I agree with this update if it performs better than the current template on the geti scenario dataset. But, without such data **in OTX**, it may be challenging to update all hyperparameters.

Second, even if we don't update the hyperparameters, there are many code changes outside of the new template. If we don't minimize these changes, we may end up with unintended differences that result in different outcomes from previous experiments, requiring us to optimize performance again.

Therefore, I suggest adding a new template and applying new hyperparameters in it while keeping the HRNet configuration unchanged. Also, it's necessary to minimize code changes outside of adding a new template to prevent side effects.

goodsong81:
@kprokofi Thank you for the great enhancement! I think @kprokofi, @supersoob and @JihwanEom's opinions all make senses. Then, why don't we apply these changes step by step? I think the only concern here is too broad impact regardless of improvement itself.

- In this PR, new model change only
- Kirill to hold a review session
- Decision for old model recipe changes afterward

What do you think?

kprokofi:
> @kprokofi Thank you for the great enhancement! I think @kprokofi, @supersoob and @JihwanEom's opinions all make senses. Then, why don't we apply these changes step by step? I think the only concern here is too broad impact regardless of improvement itself.
> 
> * In this PR, new model change only
> * Kirill to hold a review session
> * Decision for old model recipe changes afterward
> 
> What do you think?

Step by step is OK. I will revert training recipe for now, however we will have still bad perfomance not compatible with other solutions. Maybe it is ok for 12-24 images, but OTX should cover wide range of datasets including few hundreds and thousands images.

I will train it once again comparing training recipes and will prepare pivot table with results

kprokofi:
I updated templates (Added SegNext) and revert changes in configuration for old models

kprokofi:
I checked IR conversion and backward compatibility with 1.0.0 mmseg release

kprokofi:
@goodsong81 , @supersoob , could you take a look?
I reverted changes for incremental recipe for now. I added separate training recipe for new models and added them as experimental_template.

kprokofi:
@supersoob I updated PR and resolved conflicts

supersoob (approved):
Nice work:) Thank you for updating. It looks good to me.

sungmanc (approved):
LGTM

PR #1970, opened 2023-04-04
Add legacy ssd-like IRs support

sovrasov:
### Summary

In old SSD-like IRs bbox and label layers have different shapes than in current implementation. Wrapper was fixes to support both IR formats.

### Checklist

### License

```python
```

### DIFF
--- a/otx/algorithms/detection/adapters/openvino/model_wrappers/openvino_models.py
+++ b/otx/algorithms/detection/adapters/openvino/model_wrappers/openvino_models.py
@@ -129,7 +129,11 @@ class BatchBoxesLabelsParser:
             self.labels_layer = None
             self.default_label = default_label
 
-        self.bboxes_layer = self.find_layer_bboxes_output(layers)
+        try:
+            self.bboxes_layer = self.find_layer_bboxes_output(layers)
+        except ValueError:
+            self.bboxes_layer = find_layer_by_name("boxes", layers)
+
         self.input_size = input_size
 
     @staticmethod
@@ -146,7 +150,8 @@ class BatchBoxesLabelsParser:
         """Parse bboxes."""
         # FIXME: here, batch dim of IR must be 1
         bboxes = outputs[self.bboxes_layer]
-        bboxes = bboxes.squeeze(0)
+        if bboxes.shape[0] == 1:
+            bboxes = bboxes.squeeze(0)
         assert bboxes.ndim == 2
         scores = bboxes[:, 4]
         bboxes = bboxes[:, :4]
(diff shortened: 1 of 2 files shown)


goodsong81 (changes requested):
Thank you so much for the fix. We need to be careful for this kind of change in the future. By the way, could you update the change log? It's a new convention and I think this change is worth to note. :)

goodsong81 (approved):
LGTM!

jaegukhyun (approved):
Is this for old version of openvino model inference? It would be nice if you add some summary and motivation of this work.

otx/algorithms/detection/adapters/openvino/model_wrappers/openvino_models.py
```diff
@@ -146,7 +150,8 @@ def __call__(self, outputs):
         """Parse bboxes."""
         # FIXME: here, batch dim of IR must be 1
         bboxes = outputs[self.bboxes_layer]
-        bboxes = bboxes.squeeze(0)
+        if bboxes.shape[0] == 1:
```
cih9088, line 153: Why do we need this condition? `squeeze` removes the first dim if the first dim is 1, does nothing otherwise.

otx/algorithms/detection/adapters/openvino/model_wrappers/openvino_models.py
```diff
@@ -156,7 +161,8 @@ def __call__(self, outputs):
             labels = outputs[self.labels_layer]
         else:
             labels = np.full(len(bboxes), self.default_label, dtype=bboxes.dtype)
-        labels = labels.squeeze(0)
+        if labels.shape[0] == 1:
```
cih9088, line 164: Why do we need this condition? `squeeze` removes the first dim if the first dim is 1, does nothing otherwise.

otx/algorithms/detection/adapters/openvino/model_wrappers/openvino_models.py
sovrasov: We can squeeze the first dimension only if it equals to 1. For old IRs from previous OTE versions that's not true and squeeze raises an exception.

otx/algorithms/detection/adapters/openvino/model_wrappers/openvino_models.py
sovrasov: We can squeeze the first dimension only if it equals to 1. For old IRs from previous OTE versions that's not true and squeeze raises an exception.

sovrasov:
> Is this for old version of openvino model inference? It would be nice if you add some summary and motivation of this work.

Added a brief description in summary

otx/algorithms/detection/adapters/openvino/model_wrappers/openvino_models.py
cih9088: Ahh. The object is numpy array not pytorch tesnor.

otx/algorithms/detection/adapters/openvino/model_wrappers/openvino_models.py
cih9088: Ahh. The object is numpy array not pytorch tesnor.

cih9088 (approved):

PR #1980, opened 2023-04-06
Re enable multi GPU test cases

eunwoosh:
### Summary
This PR includes:

- re-enable multi GPU test cases
- fix a minor bug of multi GPU test cases
- Handling a getting stuck when main process raises an error

**What is fixed bug about?**
After training is done, both rank 0 processes and other processes try to save a model weight using saved model weight by mmXX framework.
Because only rank 0 process can save a model weight every epoch in mmXX framework,
If other processes try to save a model weight after training before rank 0 process doesn't save a model weight yet, error is raised.
Fortunately, left steps after training can be executed by each process independently, it's not huge problem.
But I fixed the bug to remove unnecessary error log.

**Why existing main function in cli/train.py is wrapped?**
It's for handling a getting stuck when main process raises an error. Current implementation deals with case that main process 
accidently exit, by making child process kill itself when it's observed.
But, when main process with child processes raises an error, main process doesn't exit for waiting all child processes done.
For this reason, child process can't know main process runs normally or waits itself after raising an error.
To resolve this issue, only way I can found is registering callback function to raising error using `ExitStack`.
Because `ExitStack` only can be used with `with`, I wrapped the current main function.
If you have any other better idea about this, please let me know. I'll check it :) 

**pre-merge test result**
image
image
I ran enabled tests and all tests are passed.

### How to test
pytest -ra --showlocals tests/integration/cli -k "multi_gpu_train"
pytest -ra --showlocals tests/e2e/cli -k "multi_gpu_train"

### Checklist

### License

```python
```

### DIFF
--- a/otx/algorithms/segmentation/adapters/mmseg/models/losses/detcon_loss.py
+++ b/otx/algorithms/segmentation/adapters/mmseg/models/losses/detcon_loss.py
@@ -42,11 +42,11 @@ class DetConLoss(nn.Module):
 
     def get_distributed_tensors(self, target1, target2, batch_size, num_samples, num_features, device):
         """Grab tensors across replicas during distributed training."""
-        num_gpus = torch.cuda.device_count()
-        if num_gpus > 1 and torch.distributed.is_initialized() and self.use_replicator_loss:
+        if torch.distributed.is_initialized() and self.use_replicator_loss:
             # Grab tensor across replicas and expand first dimension
-            target1_large = [torch.zeros_like(target1) for _ in range(num_gpus)]
-            target2_large = [torch.zeros_like(target2) for _ in range(num_gpus)]
+            world_size = dist.get_world_size()
+            target1_large = [torch.zeros_like(target1) for _ in range(world_size)]
+            target2_large = [torch.zeros_like(target2) for _ in range(world_size)]
             dist.all_gather(target1_large, target1)
             dist.all_gather(target2_large, target2)
             target1_large = torch.cat(target1_large, dim=0)
--- a/otx/algorithms/segmentation/adapters/mmseg/models/losses/detcon_loss.py
+++ b/otx/algorithms/segmentation/adapters/mmseg/models/losses/detcon_loss.py
@@ -42,7 +42,7 @@ class DetConLoss(nn.Module):
 
     def get_distributed_tensors(self, target1, target2, batch_size, num_samples, num_features, device):
         """Grab tensors across replicas during distributed training."""
-        if torch.distributed.is_initialized() and self.use_replicator_loss:
+        if dist.is_initialized() and self.use_replicator_loss:
             # Grab tensor across replicas and expand first dimension
             world_size = dist.get_world_size()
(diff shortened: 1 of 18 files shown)


sungmanc (dismissed):

supersoob (dismissed):

eunwoosh:
Could you check what I modified, @sungchul2 ? I changed using `num_gpus` to `world_size` in otx/algorithms/segmentation/adapters/mmseg/models/losses/detcon_loss.py.

otx/algorithms/segmentation/adapters/mmseg/models/losses/detcon_loss.py
```diff
@@ -42,11 +42,11 @@ def __init__(
 
     def get_distributed_tensors(self, target1, target2, batch_size, num_samples, num_features, device):
         """Grab tensors across replicas during distributed training."""
-        num_gpus = torch.cuda.device_count()
-        if num_gpus > 1 and torch.distributed.is_initialized() and self.use_replicator_loss:
+        if torch.distributed.is_initialized() and self.use_replicator_loss:
```
sungchul1: Is `torch.distributed.is_initialized()` True when only using multi gpu?
And during multi gpu, do `torch.cuda.device_count()` and `dist.get_world_size()` have the same value?

----- added -----
What about when using the limited number of gpus with CUDA_VISIBLE_DEVICES=0,1 in the envs with 4 gpus?
And isn't unit test needed for this change?

eunwoosh: 1) Is torch.distributed.is_initialized() True when only using multi gpu? => yes
2) do torch.cuda.device_count() and dist.get_world_size() have the same value? => No. Although machine has 4 GPUs user can only use 2 GPUs for training. In that case, torch.cuda.device_count() is 4 and dist.get_world_size() is 2.

sungchul1: > 1. Is torch.distributed.is_initialized() True when only using multi gpu? => yes
> 2. do torch.cuda.device_count() and dist.get_world_size() have the same value? => No. Although machine has 4 GPUs user can only use 2 GPUs for training. In that case, torch.cuda.device_count() is 4 and dist.get_world_size() is 2.

Oh, if so, previous implementation was bug. Thanks for finding.
I left some more comments.

eunwoosh: Sorry for late. I realized there is an additional comment now.
What about when using the limited number of gpus with CUDA_VISIBLE_DEVICES=0,1 in the envs with 4 gpus? => In this case, both torch.cuda.device_count() and dist.get_world_size() is 2 same if user set `--gpus` as `0,1`.
And isn't unit test needed for this change? => I checked test_detcon_loss.py, but It isn't necessary. But updating it is better than let it be. I'll update it.

jaegukhyun (approved):

sungchul1 (approved):

PR #2010, opened 2023-04-14
Disable ignore mode in CLI

harimkang:
### Summary

Confirmed unintended loss patching in Instance-segmentation: Mask-RCNN was unintentionally using `FocalLoss` even without the ignore label. (This should use the original `CrossEntropyLoss`.)
image

In the CLI, there is currently no case where ignore label comes in, so we agreed to patch ignore to False in the CLI first.
This should be fixed in the future when ignore label is available in the OTX CLI. (This does not affect Geti or the api, as the actual template still uses ignore=True.)

### How to test

`otx build --task instance_segmentation`

### Checklist

### License

```python
```

### DIFF
--- a/otx/cli/manager/config_manager.py
+++ b/otx/cli/manager/config_manager.py
@@ -510,6 +510,11 @@ class ConfigManager:  # pylint: disable=too-many-instance-attributes
                     )
 
                     config = MPAConfig.fromfile(str(target_dir / file_name))
+                    # FIXME: In the CLI, there is currently no case for using the ignore label.
+                    # so the workspace's model patches ignore to False.
+                    if config.get("ignore", None):
+                        config.ignore = False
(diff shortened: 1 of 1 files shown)


sungmanc (approved):
BTW, is there any good way to memorize this? Maybe we could forget this in the future.

harimkang:
> BTW, is there any good way to memorize this? Maybe we could forget this in the future.

Yes, I agree. So I've put a FIXME on it, but do you have any further suggestions?

sungmanc:
> > BTW, is there any good way to memorize this? Maybe we could forget this in the future.
> 
> Yes, I agree. So I've put a FIXME on it, but do you have any further suggestions?

To be honest, I have no idea 🤣. That's the reason why I select approve button. Let's keep it as is

goodsong81 (approved):

PR #2014, opened 2023-04-14
Remove unused LabelGraph class

sovrasov:
### Summary
LabelGraph was removed.

### Checklist

### License

```python
```


harimkang (approved):
LGTM,
Is this a class that will never be used?

sovrasov:
> LGTM, Is this a class that will never be used?

This class just wraps the LabelTree and doesn't bring new functionality. Anyway, there are no plans to modify the implementation of label schema.

jaegukhyun (approved):
LGTM, could you add change log?

sovrasov:
> LGTM, could you add change log?

Do we actually need this in that case? The PR just removes dead code, which doesn't affect OTX functionality.

@goodsong81 what is our policy for the changelog? Do we need to log even microscopic changes which may be interesting only for developers and updating changelog is a thing that must happen in every PR, or we can hide similar PRs behind something like "other minor improvements and cleanup in otx core"? Omitting minor changes might be also an option if the changelog is targeted to otx users.

goodsong81:
> > LGTM, could you add change log?
> 
> Do we actually need this in that case? The PR just removes dead code, which doesn't affect OTX functionality.
> 
> @goodsong81 what is our policy for the changelog? Do we need to log even microscopic changes which may be interesting only for developers and updating changelog is a thing that must happen in every PR, or we can hide similar PRs behind something like "other minor improvements and cleanup in otx core"? Omitting minor changes might be also an option if the changelog is targeted to otx users.

No. As you said, we need to log anything worth noting to the change log.
If you think it's not that informative to OTX users, you can omit. :)

PR #2029, opened 2023-04-18
Add configuration to segmentation IR

sovrasov:
### Summary
This PR adds ModelAPI-specific meta information into segmentation model IR.

### Checklist

### License

```python
```

### DIFF
--- a/otx/algorithms/segmentation/utils/metadata.py
+++ b/otx/algorithms/segmentation/utils/metadata.py
@@ -5,6 +5,7 @@
 
 
 from mmcv.utils import ConfigDict
+
 from otx.api.entities.label_schema import LabelSchemaEntity
 
 
--- a/otx/algorithms/segmentation/utils/metadata.py
+++ b/otx/algorithms/segmentation/utils/metadata.py
@@ -1,0 +1,37 @@
+"""Utils for hadnling metadata of segmentation models."""
+
+# Copyright (C) 2022 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions
+# and limitations under the License.
+
+
+from mmcv.utils import ConfigDict
+from otx.api.entities.label_schema import LabelSchemaEntity
+
+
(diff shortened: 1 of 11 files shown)


AlexKoff88 (dismissed):

otx/algorithms/segmentation/utils/metadata.py
```diff
@@ -0,0 +1,37 @@
...
+    """Get ModelAPI config."""
+    omz_config = {}
+    omz_config[("model_info", "model_type")] = "segmentation"
+
+    omz_config[("model_info", "soft_threshold")] = hyperparams.postprocessing.soft_threshold
+    omz_config[("model_info", "blur_strength")] = hyperparams.postprocessing.blur_strength
+
+    all_labels = ""
+    for lbl in label_schema.get_labels(include_empty=False):
+        all_labels += lbl.name.replace(" ", "_") + " "
```
AlexKoff88, line 22: Hope this change won't cause any issues on the Geti side. Not sure who is the right person to comment here.

sovrasov: In the future, we have to assure that the process of conversion string labels to Geti's LabelEntity is reliable.

goodsong81 (approved):
Embedding itself here looks OK to me.
Would you somehow check whether the IR models from old OTX (Geti) without this segmentation metadata will be still working well without issues? I mean backward compatibility.

sovrasov:
> Embedding itself here looks OK to me. Would you somehow check whether the IR models from old OTX (Geti) without this segmentation metadata will be still working well without issues? I mean backward compatibility.

This information is not used by OTX, so such issues can not arise (like we don't observe it in case of classification and detection).

goodsong81:
> > Embedding itself here looks OK to me. Would you somehow check whether the IR models from old OTX (Geti) without this segmentation metadata will be still working well without issues? I mean backward compatibility.
> 
> This information is not used by OTX, so such issues can not arise (like we don't observe it in case of classification and detection).

Yes, I agree. When the new ModelAPIs will replace current OMZ ModelAPIs, we need to check such compatibility anyway. Thank you for clarification. :)

sovrasov:
> > > Embedding itself here looks OK to me. Would you somehow check whether the IR models from old OTX (Geti) without this segmentation metadata will be still working well without issues? I mean backward compatibility.
> > 
> > 
> > This information is not used by OTX, so such issues can not arise (like we don't observe it in case of classification and detection).
> 
> Yes, I agree. When the new ModelAPIs will replace current OMZ ModelAPIs, we need to check such compatibility anyway. Thank you for clarification. :)

yeah, integration of ModelAPI and it's consequences for geti versions compatibility is a standalone discussion. It's complicated, since MAPI doesn't know about config.json, so we may need patch IRs in a JIT manner.

jaegukhyun (approved):

AlexKoff88 (approved):

PR #2046, opened 2023-04-20
Docker: initial Dockerfile support

QSmally:
### Summary

This PR implements a basic Dockerfile in contrast to using Python virtual environment (as I personally had some dependency problems with it). It's basic, as in, it installs the CPU dependencies and leaves CUDA out of the mix. Maybe with build-args you can change some things but I don't have a Nvidia setup to test it.

### How to test

```bash
$ cd training_extensions/
$ docker build -t trainer .
$ docker run -it --rm trainer
trainer% otx # ...
```

### Checklist

### License

```python
```


samet-akcay:
@ashwinvaidya17, can you review this PR please?

Dockerfile
```diff
@@ -0,0 +1,18 @@
+
```
ashwinvaidya17, line 1: I am in favour of collecting all the dockerfiles in a single directory. While this is not related to the CI, maybe we can have a folder within .ci that holds all the docker files. Or we can perhaps create a folder in the root directory? What do you think @yunchu

Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+FROM ubuntu:20.04
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt update && \
+    apt install curl python3.9 python3.9-dev python3.9-distutils git g++ ffmpeg libsm6 libxext6 libgl1-mesa-glx -y && \
+    curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && \
+    python3.9 get-pip.py && \
```
ashwinvaidya17: Do you think we should symlink `python3.9` as `python` to make it convinient for users. Something like `ln -sf /usr/bin/python3.9 /usr/bin/python`

ashwinvaidya17 (commented):
Thanks for the PR. I have one comment regarding the placement of this file. The rest looks good.

Dockerfile
QSmally: I'm afraid a Dockerfile cannot be outside of its root build context with having `COPY . /training_extensions`? I don't think `COPY ..` is possible, also because there's a `.dockerignore` file in the repository's root directory, but please correct me if I'm wrong...

Additionally, I think it's better to have one Dockerfile for locally building any configured instance of the training extensions, with build args rather than multiple Dockerfiles? The build file may also run a setup script if necessary

ashwinvaidya17: Can't we use call the `build` from root and add `-f <path/Dockerfile`

ashwinvaidya17: Also, what if someone wants to have cuda support? In that case we might need to use nvidia's base image. Do you think this should be handled by cli arguments or by having multiple docker files?

QSmally: I'm not sure that keeps the build context to the repository's root directory, and I also don't see why you'd go through that extra step just to get the Dockerfile out of the root directory. I'm not sure about the Nvidia part as I haven't fiddled with it before, but good point

ashwinvaidya17: Yes, that's why I would like to know what other maintainers think about moving this to another folder. Currently it is not an issue but say in the future you need more configurations like a different ubuntu and cuda version. In that case it would be easier to have multiple docker files as they will have some significant differences. My reference here is the approach used in anomalib (see the first step https://github.com/openvinotoolkit/anomalib/tree/main/.ci#steps). We need containers for ubuntu 18 and 20 running different cuda versions. It keeps the build context while cleanly categorizing multiple docker files. [Lightning](https://github.com/Lightning-AI/lightning/tree/master/dockers) and [Torch](https://github.com/pytorch/pytorch/tree/main/.ci/docker) use similar approach. We can think of a future-proof design but if others are fine with this then I am okay with it as well.

QSmally: Instead of using a Nvidia image, perhaps we can conditionally install CUDA dependencies through the use of build args? It also allows you to use more build args to install different tools conditionally rather than switching between Dockerfiles

ashwinvaidya17: Sounds good! Let's roll with it. Maybe we can introduce a makefile in the future to make it easier to target different versions.

ashwinvaidya17 (dismissed):

Dockerfile
QSmally: Python isn't directly used by the user, so I don't think it's that big of an issue as of yet

goodsong81 (commented):
Thank you for the contribution @QSmally!
It looks generally OK but one only concern of mine is that the majority of use-cases of model training will run on GPUs. Current PR covers only Python3.9 + CPU.
@samet-akcay What do you think? Will we start from CPU and extend GPU support later?

Dockerfile
yunchu: > I am in favour of collecting all the dockerfiles in a single directory. While this is not related to the CI, maybe we can have a folder within .ci that holds all the docker files. Or we can perhaps create a folder in the root directory? What do you think @yunchu

@QSmally @ashwinvaidya17 
sorry for the delayed response. I think it's better to move different subfolder like 'docker' to hold dokerfile instead of using '.ci' because this one is not related with CI and the non-docker user doesn't need to know the context about it.

Dockerfile
yunchu: for more user's conveniences, how about get the python version through the `build-arg`? if it set to any supported python version as a default (e.g., python 3.10 - it's the latest version of the python which is supported by this project), most of users who does not care about the python version can use it without setting `build-arg` but someone who intended to use different version of python, they can set it via `build-arg` to build the docker image.

QSmally:
> Thank you for the contribution @QSmally!
> It looks generally OK but one only concern of mine is that the majority of use-cases of model training will run on GPUs. Current PR covers only Python3.9 + CPU.

Ty @goodsong81, @ashwinvaidya17, @yunchu,

I implemented build-arg support for `UBUNTU_VER`, `PYTHON_VER` and `SOURCE`. There's an issue where `python3.10-dev` and `python3.10-distutils` don't exist but I did get it working with Python 3.8 (along with Ubuntu 18.04). The default values if no args were given are `20.04`, `3.9` and `/whl/cpu` as given in the README.

However, I lack the installation knowledge and machine to get the GPU dependencies, so I'd like assistance with that

samet-akcay:
> > Thank you for the contribution @QSmally!
> > It looks generally OK but one only concern of mine is that the majority of use-cases of model training will run on GPUs. Current PR covers only Python3.9 + CPU.
> 
> Ty @goodsong81, @ashwinvaidya17, @yunchu,
> 
> I implemented build-arg support for `UBUNTU_VER`, `PYTHON_VER` and `SOURCE`. There's an issue where `python3.10-dev` and `python3.10-distutils` don't exist but I did get it working with Python 3.8 (along with Ubuntu 18.04). The default values if no args were given are `20.04`, `3.9` and `/whl/cpu` as given in the README.
> 
> However, I lack the installation knowledge and machine to get the GPU dependencies, so I'd like assistance with that

Thanks for the contribution @QSmally. In this case, GPU support could be addressed by a follow-up PR by the team later.

ashwinvaidya17 (dismissed):
Thanks

yunchu (dismissed):
LGTM. @QSmally, Thank you for your contributions!

goodsong81:
@ashwinvaidya17 @yunchu Could you help pass the Dockerfile style check? I think the CI is regarding the warning as error.
image

yunchu:
@goodsong81, we can update linter setting for dockerfile but I think it's better to resolve all warnings.
@QSmally, Can you apply some changes to resolve warning issues from linter? I recommend to use `pre-commit` tool to your development environment to get the direct feedback from the linters while generating the commits.

QSmally:
I rarely encounter a Dockerfile with pinned apt versions, so I'm not sure how (and if) we should update that, @yunchu/@goodsong81

image

yunchu:
@QSmally 
You can add the comment below to disable that warning on right before the `apt-get install` command in your Dockerfile.
```
# hadolint ignore=DL3008 
```

yunchu:
@QSmally I'm not sure but https://github.com/openvinotoolkit/training_extensions/blob/develop/.ci/Dockerfile this one could be the example for you.

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+ARG PYTHON_VER=3.9
+ARG SOURCE=https://download.pytorch.org/whl/cpu
+ENV DEBIAN_FRONTEND=noninteractive
+
+# hadolint ignore=DL3008
+RUN apt-get update && \
+    apt-get install curl python$PYTHON_VER python$PYTHON_VER-dev python$PYTHON_VER-distutils g++ ffmpeg libsm6 libxext6 libgl1-mesa-glx -y && \
+    curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && \
+    python$PYTHON_VER get-pip.py && \
+    rm get-pip.py
```
yunchu:
```suggestion
    rm get-pip.py && \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/*
```
changes above will remedy the issue below
```bash
Dockerfile:10 DL3009 info: Delete the apt-get lists after installing something
```

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+ARG UBUNTU_VER=20.04
+FROM ubuntu:$UBUNTU_VER
+
+ARG PYTHON_VER=3.9
+ARG SOURCE=https://download.pytorch.org/whl/cpu
+ENV DEBIAN_FRONTEND=noninteractive
+
+# hadolint ignore=DL3008
+RUN apt-get update && \
+    apt-get install curl python$PYTHON_VER python$PYTHON_VER-dev python$PYTHON_VER-distutils g++ ffmpeg libsm6 libxext6 libgl1-mesa-glx -y && \
```
yunchu:
```suggestion
    apt-get install --no-install-recommends -y && \
    curl python$PYTHON_VER python$PYTHON_VER-dev python$PYTHON_VER-distutils g++ ffmpeg libsm6 libxext6 libgl1-mesa-glx && \
```
changes like above will remedy the issue below
```bash
Dockerfile:10 DL3015 info: Avoid additional packages by specifying `--no-install-recommends`
```

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+    curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && \
+    python$PYTHON_VER get-pip.py && \
+    rm get-pip.py
+
+WORKDIR /training_extensions
+COPY . /training_extensions
+
+RUN pip install --no-cache-dir torch==1.13.1 torchvision==0.14.1 torchaudio==0.13.1 --extra-index-url $SOURCE && \
```
yunchu, line 22: I think thing below is the false positive
```bash
Dockerfile:19 SC2102 info: Ranges can only match single chars (mentioned due to duplicates).
```
So, we can omit it by adding comment below
```suggestion
# hadolint ignore=SC2102
RUN pip install --no-cache-dir torch==1.13.1 torchvision==0.14.1 torchaudio==0.13.1 --extra-index-url $SOURCE && \
```

yunchu (approved):

goodsong81 (approved):
Thank you for the contribution!

PR #2052, opened 2023-04-21
Tiling Spatial Concatenation for OpenVINO IR

eugene123tw:
### Summary

To optimize the inference process of an IR model, adjusting tiling parameters can be effective. For example, during training, 16x16 tiles with 100x100 pixels may be used. However, during inference, 8x8 tiles with 200x200 pixels can be used after reshaping the IR model once. This change can result in a 1/4 reduction in the number of calls, with expected internal parallelization in the OpenVINO runtime.

### Checklist

### License

```python
```

### DIFF
--- a/otx/algorithms/common/configs/training_base.py
+++ b/otx/algorithms/common/configs/training_base.py
@@ -190,6 +190,16 @@ class BaseConfig(ConfigurableParameters):
             affects_outcome_of=ModelLifecycle.TRAINING,
         )
 
+        auto_decrease_batch_size = configurable_boolean(
+            default_value=False,
+            header="Decrease batch size if current batch size isn't fit to CUDA memory.",
+            description="Find a proper batch size by training for an iteration with various batch size a few times.",
+            warning="Enabling this option could reduce the actual batch size if the current setting results in "
+            "out-of-memory error. The learning rate also could be adjusted according to the adapted batch size. "
+            "This process might take some extra computation time to try a few batch size candidates.",
+            affects_outcome_of=ModelLifecycle.TRAINING,
+        )
+
     @attrs
     class BasePostprocessing(ParameterGroup):
         """BasePostprocessing for OTX Algorithms."""
@@ -281,7 +291,7 @@ class BaseConfig(ConfigurableParameters):
             header="train type",
             description="Training scheme option that determines how to train the model",
             editable=False,
-            visible_in_ui=True,
+            visible_in_ui=False,
         )
 
         mem_cache_size = configurable_integer(
@@ -299,7 +309,7 @@ class BaseConfig(ConfigurableParameters):
             header="Scheme for storage cache",
             description="Scheme for storage cache",
             editable=False,
-            visible_in_ui=True,
+            visible_in_ui=False,
         )
 
     @attrs
--- a/otx/algorithms/common/configs/training_base.py
+++ b/otx/algorithms/common/configs/training_base.py
@@ -367,7 +367,7 @@ class BaseConfig(ConfigurableParameters):
             description="The purpose of the scale parameter is to optimize the performance and "
             "efficiency of tiling in OpenVINO IR during inference. By controlling the increase in tile size and "
             "input size, the scale parameter allows for more efficient parallelization of the workload and "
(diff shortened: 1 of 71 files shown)


otx/algorithms/common/configs/training_base.py
```diff
@@ -372,4 +372,20 @@ class BaseTilingParameters(ParameterGroup):
             affects_outcome_of=ModelLifecycle.NONE,
         )
 
+        ir_scale_factor = configurable_float(
```
sungmanc:
```suggestion
        tiling_ir_scale_factor = configurable_float(
```

sungmanc: The name `ir_scale_factor` is could be confusing. Since this variable is only used for tiling, I would suggest to change the name

otx/algorithms/detection/adapters/openvino/task.py
```diff
@@ -274,6 +274,7 @@ def __init__(
         tile_size: int = 400,
         overlap: float = 0.5,
         max_number: int = 100,
+        ir_scale_factor: float = 1.0,
```
sungmanc: Same with above

otx/algorithms/detection/configs/instance_segmentation/configuration.yaml
```diff
@@ -553,5 +553,23 @@ tiling_parameters:
     visible_in_ui: true
     warning: null
 
+  ir_scale_factor:
```
sungmanc: Same with above

tests/integration/cli/detection/test_tiling_detection.py
```diff
@@ -112,7 +112,6 @@ def test_otx_hpo(self, template, tmp_dir_path):
         otx_hpo_testing(template, tmp_dir_path, otx_dir, args)
 
     @e2e_pytest_component
-    @pytest.mark.skip(reason="CVS-98026")
```
sungmanc: So, Is tiling NNCF enable now?

tests/unit/algorithms/detection/tiling/test_tiling_detection.py
```diff
@@ -235,6 +257,60 @@ def test_patch_tiling_func(self):
         patch_tiling(cfg, hyper_parameters, self.otx_dataset)
 
     @e2e_pytest_unit
-    def test_openvino(self):
-        # TODO[EUGENE]: implement unittest for tiling prediction with openvino
-        pass
+    @pytest.mark.parametrize("scale_factor", [1, 1.5, 2, 3, 4])
+    def test_ir_scale_deploy(self, tmp_dir_path, scale_factor):
```
sungmanc: Same with above (function name)

sungmanc (commented):
Overall, LGTM. BTW, could you attach the experimental results in description?

otx/algorithms/common/configs/training_base.py
eunwoosh: I agree also

goodsong81: It's included in `tiling_parameters` group. Do we really need to add `tiling` for all tiling-related parameters? BTW, I'm neutral as other param has `tile` words anyway.

goodsong81 (changes requested):
Thank you for the great enhancement. LGTM if test failures are fixed.

tests/integration/cli/detection/test_tiling_detection.py
eugene123tw: @sungmanc I reverted the pytest skip, and will address this issue in a new PR: eugene/CVS-109001-fix-tiling-nncf

otx/algorithms/common/configs/training_base.py
eugene123tw: I didn't include the term "tiling" in the IR scale factor because while it's related to tiling, the concept primarily involves increasing the input shape of the IR model, and since it's under tiling group, it should be clear enough from my POV. Nevertheless, I understand it may cause confusion and I can make the change.

goodsong81 (approved):

jaegukhyun (approved):

PR #2106, opened 2023-05-04
Update concurrency settings for pre-merge

yunchu:
### Summary
By applying reusable workflow to the testing pipelines, the workflow level of concurrency setting leads unexpected job canceling when any job resulted as failure because all jobs shared the same concurrency group setting.
By this PR, those issues can be fixed.

* Pre-merge: Workflow level of concurrency setting leads unexpected job canceling.
To resolve this issue, concurrency setting needs to be moved to each job level.

* Daily: remove concurrency setting

### DIFF
--- a/.github/workflows/pre_merge.yml
+++ b/.github/workflows/pre_merge.yml
@@ -12,13 +12,12 @@ on:
       - synchronize
   workflow_dispatch: # run on request (no need for PR)
 
-# This is what will cancel the workflow concurrency
-concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
-  cancel-in-progress: true
-
 jobs:
   Code-Quality-Checks:
+    # This is what will cancel the job concurrency
+    concurrency:
+      group: ${{ github.workflow }}-Linting-${{ github.event.pull_request.number || github.ref }}
+      cancel-in-progress: true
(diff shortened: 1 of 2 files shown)


.github/workflows/pre_merge.yml
```diff
@@ -94,6 +97,10 @@ jobs:
           - task: "ano"
             test_dir: "tests/integration/cli/anomaly"
     name: Integration-Test-py310-${{ matrix.task }}
+    # This is what will cancel the job concurrency
```
jaegukhyun, line 100: Is this mean a single test failure can cancel the remaining integration tests?

goodsong81 (approved):
LGTM, thanks!
One question: why removing it from daily?

.github/workflows/pre_merge.yml
yunchu: Just meaning the level of concurrency setting. As you can see, it is defined inside of job section not the workflow level.

yunchu:
> LGTM, thanks! One question: why removing it from daily?

@goodsong81, I've added some background to the PR description. please check out them.
And I can update daily as well using the job level concurrency setting but in case of daily will be triggered once a day. I think the duplication of the same job will not be happened so the removing is the simplest solution for the daily workflow case.

jaegukhyun (approved):

PR #2113, opened 2023-05-08
Set fail-fast to false to avoid canceling on parallel jobs

yunchu:
### Summary

Default option of the `fail-fast` to the workflow is `true` so the jobs running as parallel (using matrix) will be
canceled when any instance completed as `failure`.
To avoid this, need to set `fail-fast: false` for the job runs as parallel.


sungmanc (approved):
LGTM

eunwoosh (approved):

PR #2119, opened 2023-05-09
Auto adapting batch size

eunwoosh:
### Summary

- Implement a feature to adapt a batch size according to current GPU memory size automatically.
- It supports action, classification, detection and segmentation tasks.

### Detail
This feature finds a batch size which uses almost GPU memory. It generally increases batch size, so, it can reduce a training time when comparing it to small batch size with same epochs. To make it work well, it uses some techniques. First, it tries to find just a big enough batch size instead of maximum runnable batch size. Finding a maximum value need quite much time although the advantage isn't huge compared to finding just a big enough batch size. Second, it estimates an optimal batch size using estimated equation. it can reduce much time compared to other methods. Finally, learning rate is increased sqrt(k) times, 
when increasing batch size k times. It's based on experiments and also, it's supported theoretically.

### How to test
You can use the feature by adding `--learning_parameters.auto_adapt_batch_size Full`.
You can also test the feature by running `pytest -k auto_adapt_batch_size tests/integration/cli/`

### Checklist

### License

```python
```

### DIFF
--- a/otx/algorithms/action/adapters/mmaction/task.py
+++ b/otx/algorithms/action/adapters/mmaction/task.py
@@ -48,6 +48,7 @@ from otx.algorithms.common.adapters.mmcv.utils.config_utils import (
     MPAConfig,
     update_or_add_custom_hook,
 )
+from otx.algorithms.common.configs.configuration_enums import BatchSizeAdaptType
 from otx.algorithms.common.utils.data import get_dataset
 from otx.algorithms.common.utils.logger import get_logger
 from otx.api.entities.datasets import DatasetEntity
@@ -295,17 +296,14 @@ class MMActionTask(OTXActionTask):
 
         validate = bool(cfg.data.get("val", None))
 
-        if (
-            self._hyperparams.learning_parameters.auto_decrease_batch_size
-            or self._hyperparams.learning_parameters.auto_adapt_batch_size
-        ):
+        if self._hyperparams.learning_parameters.auto_adapt_batch_size != BatchSizeAdaptType.NONE:
             train_func = partial(train_model, meta=deepcopy(meta), model=deepcopy(model), distributed=False)
             adapt_batch_size(
                 train_func,
                 cfg,
                 datasets,
                 validate,
-                not_increase=(not self._hyperparams.learning_parameters.auto_adapt_batch_size),
+                not_increase=(self._hyperparams.learning_parameters.auto_adapt_batch_size == BatchSizeAdaptType.SAFE),
             )
 
         train_model(
--- a/otx/algorithms/action/adapters/mmaction/task.py
+++ b/otx/algorithms/action/adapters/mmaction/task.py
@@ -295,9 +295,11 @@ class MMActionTask(OTXActionTask):
 
         validate = bool(cfg.data.get("val", None))
 
-        if self._hyperparams.learning_parameters.auto_decrease_batch_size:
+        if (self._hyperparams.learning_parameters.auto_decrease_batch_size
+            or self._hyperparams.learning_parameters.auto_adapt_batch_size):
             train_func = partial(train_model, meta=deepcopy(meta), model=deepcopy(model), distributed=False)
-            adapt_batch_size(train_func, cfg, datasets, validate)
+            adapt_batch_size(train_func, cfg, datasets, validate,
+                             not_increase=(not self._hyperparams.learning_parameters.auto_adapt_batch_size))
 
         train_model(
             model,
--- a/otx/algorithms/action/adapters/mmaction/task.py
+++ b/otx/algorithms/action/adapters/mmaction/task.py
@@ -295,11 +295,18 @@ class MMActionTask(OTXActionTask):
 
         validate = bool(cfg.data.get("val", None))
 
-        if (self._hyperparams.learning_parameters.auto_decrease_batch_size
-            or self._hyperparams.learning_parameters.auto_adapt_batch_size):
+        if (
+            self._hyperparams.learning_parameters.auto_decrease_batch_size
+            or self._hyperparams.learning_parameters.auto_adapt_batch_size
+        ):
             train_func = partial(train_model, meta=deepcopy(meta), model=deepcopy(model), distributed=False)
-            adapt_batch_size(train_func, cfg, datasets, validate,
-                             not_increase=(not self._hyperparams.learning_parameters.auto_adapt_batch_size))
+            adapt_batch_size(
+                train_func,
+                cfg,
+                datasets,
+                validate,
+                not_increase=(not self._hyperparams.learning_parameters.auto_adapt_batch_size),
+            )
 
         train_model(
             model,
--- a/otx/algorithms/action/configs/classification/configuration.yaml
+++ b/otx/algorithms/action/configs/classification/configuration.yaml
@@ -198,43 +198,29 @@ learning_parameters:
       type: UI_RULES
     visible_in_ui: true
     warning: This will automatically control the patience and interval when early stopping is enabled.
-  auto_decrease_batch_size:
-    affects_outcome_of: TRAINING
-    default_value: false
-    description: Find a proper batch size by training for an iteration with various batch size a few times.
-    editable: true
-    header: Decrease batch size if current batch size isn't fit to CUDA memory.
-    type: BOOLEAN
-    ui_rules:
-      action: DISABLE_EDITING
-      operator: AND
-      rules: []
-      type: UI_RULES
-    visible_in_ui: true
-    warning:
-      Enabling this option could reduce the actual batch size if the current setting results in out-of-memory error.
-      The learning rate also could be adjusted according to the adapted batch size.
-      This process might take some extra computation time to try a few batch size candidates.
   auto_adapt_batch_size:
     affects_outcome_of: TRAINING
-    default_value: false
-    description:
-      Find a big enough batch size using most of GPU memory by training with various batch size a few times.
-      It increases batch size generally, but it can decrease batch size if even default value isn't fit to
-      current GPU memory. After adapting batch size, learning rate is also updated.
+    default_value: None
+    description: Safe => Prevent GPU out of memory. Full => Find a batch size using almost GPU memory.
     editable: true
-    header: Adapt a batch size according to GPU memory.
-    type: BOOLEAN
+    enum_name: BatchSizeAdaptType
+    header: Decrease batch size if current batch size isn't fit to CUDA memory.
+    options:
+      NONE: "None"
+      SAFE: "Safe"
+      FULL: "Full"
+    type: SELECTABLE
     ui_rules:
       action: DISABLE_EDITING
       operator: AND
       rules: []
       type: UI_RULES
+    value: None
     visible_in_ui: true
     warning:
-      Batch size is updated to use most of GPU memory, which means batch size increases generally.
-      It can reduce overall training time by reducing number of back propgation execution,
-      but it could reduce model accuracy slightly although learning rate is also aligned to batch size.
+      Enabling this option can change the actual batch size depending on adapt type and current GPU Memory.
+      The learning rate also could be adjusted according to the adapted batch size. This process might change
+      a model performance and take some extra computation time to try a few batch size candidates.
   type: PARAMETER_GROUP
   visible_in_ui: true
 postprocessing:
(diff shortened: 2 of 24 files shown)


otx/algorithms/detection/adapters/mmdet/task.py
```diff
@@ -277,9 +277,18 @@ def _train_model(
 
         validate = bool(cfg.data.get("val", None))
 
-        if self._hyperparams.learning_parameters.auto_decrease_batch_size:
+        if (
```
harimkang: Does the same code need to be scattered across each task? 
Couldn't it just be declared in a commonly used task?

harimkang (commented):
I left one comment.

otx/algorithms/detection/adapters/mmdet/task.py
eunwoosh: It's hard to do, because a few things are different depending on tasks. For example, action task should set True to `distributed` argument. And each task has different training function (e.g. train_detector, train_segmentor, etc.).

sungmanc (commented):
Why do we need to separate `auto_decrease_batch_size` and `auto_adapt_batch_size` ?
How about unifying the parameters? `auto_adapt_batch_size` could include the concept of `auto_decrease_batch_size`.

eunwoosh:
> Why do we need to separate `auto_decrease_batch_size` and `auto_adapt_batch_size` ? How about unifying the parameters? `auto_adapt_batch_size` could include the concept of `auto_decrease_batch_size`.

IMHO, because they have different purpose. Purpose of `auto_decrease_batch_size` is decreasing batch size only when current batch size isn't fit to GPU memory. In contrast, Purpose of `auto_adapt_batch_size` is making training faster by increasing batch size. And I also concern that if unifying them, sem-seg and ins-seg will use `auto_adapt_batch_size` as default. Currently, `auto_decrease_batch_size` is turned on as default in those tasks because they are easy to make GPU OOM.

harimkang:
> 

How about getting a key named `auto_adapt_batch_size` as a str like.. `auto_adapt_batch_size in ("decrease", "auto")` (just example)
I don't see the need to have two separate parameters. Wouldn't it be nice to have multiple behaviors in one?

eunwoosh:
> > 
> 
> How about getting a key named `auto_adapt_batch_size` as a str like.. `auto_adapt_batch_size in ("decrease", "auto")` (just example) I don't see the need to have two separate parameters. Wouldn't it be nice to have multiple behaviors in one?

I think it's a good idea. I'll apply it after asking others' opinions.

otx/algorithms/action/adapters/mmaction/task.py
```diff
@@ -295,9 +295,18 @@ def _train_model(
 
         validate = bool(cfg.data.get("val", None))
 
-        if self._hyperparams.learning_parameters.auto_decrease_batch_size:
+        if (
+            self._hyperparams.learning_parameters.auto_decrease_batch_size
+            or self._hyperparams.learning_parameters.auto_adapt_batch_size
```
supersoob: I agree that these two parameter from cli is better to be combined. I suggest to make a string argument as ('maximum', 'auto') rather than ('decrease', 'auto'). But I suddenly got curious, if merge these config, what happens if user put the runnable batch size and enabled this with 'decrease'? Is batch size kept or automatically returned with the large batch or maximum batch size?

otx/algorithms/common/adapters/mmcv/utils/automatic_bs.py
```diff
@@ -84,19 +86,27 @@ def train_func_single_iter(batch_size):
...
         train_func=train_func_single_iter,
-        current_bs=default_bs,
-        trainset_size=len(datasets[0]),
+        default_bs=default_bs,
+        max_bs=len(datasets[0]),
     )
+    if not_increase:
+        new_batch_size = bs_search_ago.auto_decrease_batch_size()
+    else:
+        new_batch_size = bs_search_ago.find_big_enough_batch_size()
```
supersoob: Can we make it bs_search_algo.find() do the all and return the proper new_batch_size of all cases?

otx/algorithms/common/adapters/mmcv/utils/automatic_bs.py
```diff
@@ -84,19 +86,27 @@ def train_func_single_iter(batch_size):
 
     default_bs = _get_batch_size(cfg)
 
-    available_bs = adapt_torch_model_bs(
+    bs_search_ago = BsSearchAlgo(
```
supersoob: bs_search_ago -> bs_search_algo?

eunwoosh: Thanks for finding! I'll change it.

otx/algorithms/common/adapters/mmcv/utils/automatic_bs.py
eunwoosh: It's hard to do, because `auto_decrease_batch_size` only finds batch size lower than default value and `find_big_enough_batch_size` finds batch size which uses almost GPU memory.

otx/algorithms/action/adapters/mmaction/task.py
eunwoosh: If user set it as `decrease` then just checking defulat batch size is runnable with current GPU memory. it try to find a runnable batch size only when default batch size isn't runnable.

eunwoosh:
I combined separated arguments to a single one and applied other comments. Could you review my PR? @harimkang  @supersoob  @sungmanc

otx/algorithms/common/configs/configuration_enums.py
```diff
@@ -33,3 +33,11 @@ class StorageCacheScheme(ConfigurableEnum):
     JPEG_95 = "JPEG/95"
     PNG = "PNG"
     TIFF = "TIFF"
+
+
+class BatchSizeAdaptType(ConfigurableEnum):
+    """This Enum represents the type of adapting batch size."""
+
```
harimkang, line 45: Actually, I'm not sure those words (NONE, SAFE, FULL) are appropriate; it's hard to look at them and predict how they will behave.

eunwoosh: Honestly, making an appropriate name, which is meaningful but not verbose, is also quite hard to me. I had some of candidates and decide it together in the meeting. I also think it could be confusing to someone, but I think description can help them to understand.

harimkang: Yep.. It was already a done deal,,,, then I guess we need to comment this out here more detail about Type.

otx/algorithms/action/configs/classification/configuration.yaml
```diff
@@ -198,23 +198,29 @@ learning_parameters:
...
       operator: AND
       rules: []
       type: UI_RULES
+    value: None
     visible_in_ui: true
     warning:
-      Enabling this option could reduce the actual batch size if the current setting results in out-of-memory error.
-      The learning rate also could be adjusted according to the adapted batch size.
-      This process might take some extra computation time to try a few batch size candidates.
+      Enabling this option can change the actual batch size depending on adapt type and current GPU Memory.
```
sungmanc:
```suggestion
      Enabling this could change the actual batch size depending on the current GPU status.
```

otx/algorithms/common/configs/configuration_enums.py
eunwoosh: I agree on that. Then I'll add some comment there.

otx/algorithms/action/configs/classification/configuration.yaml
sungmanc: Since we had a description for the options (None, Safe, Full), we don't need to introduce it again.

otx/algorithms/action/configs/classification/configuration.yaml
```diff
@@ -198,23 +198,29 @@ learning_parameters:
       type: UI_RULES
     visible_in_ui: true
     warning: This will automatically control the patience and interval when early stopping is enabled.
-  auto_decrease_batch_size:
+  auto_adapt_batch_size:
     affects_outcome_of: TRAINING
-    default_value: false
-    description: Find a proper batch size by training for an iteration with various batch size a few times.
+    default_value: None
+    description: Safe => Prevent GPU out of memory. Full => Find a batch size using almost GPU memory.
```
sungmanc:
```suggestion
    description: Safe => Prevent GPU out of memory. Full => Find a max batch size with considering GPU status.
```

otx/algorithms/common/configs/configuration_enums.py
eunwoosh: done on [2dcb527](https://github.com/openvinotoolkit/training_extensions/pull/2119/commits/2dcb52706d932498d0863be6d596d597c82ce517)

sungmanc (commented):
LGTM, with minor comments

otx/algorithms/action/configs/classification/configuration.yaml
eunwoosh: As you know, it doesn't find maximum batch size, which is why I write as it is. How about just letting it be?

sungmanc: My suggestion is not mandatory. However, almost GPU memory seems weird to me. How about most of GPU memory?

otx/algorithms/action/configs/classification/configuration.yaml
eunwoosh: done on [942eb90](https://github.com/openvinotoolkit/training_extensions/pull/2119/commits/942eb90b1d54fbb92328bd7ba34720befdf54b73)

otx/algorithms/action/configs/classification/configuration.yaml
eunwoosh: done on [942eb90](https://github.com/openvinotoolkit/training_extensions/pull/2119/commits/942eb90b1d54fbb92328bd7ba34720befdf54b73)

sungmanc (dismissed):
Thanks very much for applying my minor suggestion :), LGTM.

harimkang (dismissed):

supersoob (dismissed):

otx/algorithms/common/adapters/torch/utils/bs_search_algo.py
```diff
@@ -12,66 +12,199 @@
...
+        """Decrease batch size if default batch size isn't fit to current GPU device.
+
+        Returns:
+            int: Proper batch size possibly decreased as default value isn't fit
+        """
+        available_bs = 0
+        current_bs = self._default_bs
+        lowest_unavailable_bs = self._default_bs + 2
+
+        while True:
```
harimkang, line 84: Maybe we don't need to specify a timeout value for this?

otx/algorithms/common/adapters/torch/utils/bs_search_algo.py
```diff
@@ -12,66 +12,199 @@
...
+        estimated_bs += 2
+        if estimated_bs > self._max_bs:
+            return self._default_bs
+        cuda_oom, bs_mem_usage = self._try_batch_size(estimated_bs)
+        if cuda_oom or bs_mem_usage > self._mem_upper_bound:
+            return self._default_bs
+
+        # estimate batch size using equation
+        estimation_pct = 0.82
+        while True:
```
harimkang, line 142: This is the same as the comment above.

otx/algorithms/common/configs/configuration_enums.py
```diff
@@ -33,3 +33,16 @@ class StorageCacheScheme(ConfigurableEnum):
     JPEG_95 = "JPEG/95"
     PNG = "PNG"
     TIFF = "TIFF"
+
+
+class BatchSizeAdaptType(ConfigurableEnum):
+    """This Enum represents the type of adapting batch size.
+
+    None : Not adapt batch size.
+    Safe : Find a batch size preventing GPU out of memory.
```
harimkang, line 42: I'm still a little confused, Safe is the decrease_only mode, right?
Sentence "Find a batch size preventing GPU out of memory" seems like it should apply to both Safe and Full.

otx/algorithms/common/adapters/torch/utils/bs_search_algo.py
eunwoosh: I think it always can exit loop  because upper bound decreases as it progresses and also lower bound increases also. And it works with default batch size, so it's doesn't use much time, I think.

otx/algorithms/common/adapters/torch/utils/bs_search_algo.py
eunwoosh: As you know, it exits the loop if criteria is meet or estimated batch size is already tried. So I think it can always exit the loop.

otx/algorithms/common/configs/configuration_enums.py
eunwoosh: Do you mean "Find a batch size preventing GPU out of memory" needs to be added to `Full` description?

harimkang (approved):

supersoob (approved):

PR #2133, opened 2023-05-11
Updated code-scan workflow

yunchu:
### Summary

snyk replaced to trivy

### DIFF
--- a/.github/workflows/code_scan.yml
+++ b/.github/workflows/code_scan.yml
@@ -9,38 +9,26 @@ on:
 jobs:
   Trivy-scan:
     runs-on: ubuntu-20.04
-    permissions:
-      security-events: write
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
-      - name: Run Trivy vulnerability scanner in repo mode
-        uses: aquasecurity/trivy-action@0.9.2
+      - name: Set up Python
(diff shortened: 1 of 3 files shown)


sungmanc (approved):

.github/workflows/code_scan.yml
```diff
@@ -9,38 +9,26 @@ on:
...
-    permissions:
-      contents: write
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@v3
-      - name: Run Trivy in GitHub SBOM mode and submit results to Dependency Graph
-        uses: aquasecurity/trivy-action@0.9.2
+          python-version: "3.10"
+      - name: Install dependencies
+        run: python -m pip install tox
```
eunwoosh, line 20: I'm not sure but isn't it necessary to consider tox version? because I watched tox is installed as below in other part.
```
python -m pip install -r requirements/dev.txt
```

yunchu: @eunwoosh In case of testing, need to install testing dependency as well using the requirements/dev.txt but for the code scanning, we don't need to install them. only required thing is all list of dependencies. So, I think it could reduce the time to be done for this workflow.

eunwoosh (approved):

PR #2148, opened 2023-05-14
Fix coordinates normalization for segmentation (release)

kprokofi:
duplicate: https://github.com/openvinotoolkit/training_extensions/pull/2146


eunwoosh (commented):
Nice catch! I think it's hard to find.
I have a one question. Don't we have to deal with the case where value of width or height is 1?
There's little chance but it's possible case. How do you think about?

goodsong81:
@kprokofi Could you please fix some unit test failures?

kprokofi:
> Nice catch! I think it's hard to find. I have a one question. Don't we have to deal with the case where value of width or height is 1? There's little chance but it's possible case. How do you think about?

If an image equals to 1x1 pixel, then coordinates will be (0,0) only, 1 pixel - it is right. But I am not sure that 1 pixel image is a valid example

eunwoosh:
> >
> If an image equals to 1x1 pixel, then coordinates will be (0,0) only, 1 pixel - it is right. But I am not sure that 1 pixel image is a valid example

Yes, Actually (0, 0) won't be a problem but zero division can be. How to add a code to handle that case?

kprokofi:
> > > 
> > 
> > 
> > If an image equals to 1x1 pixel, then coordinates will be (0,0) only, 1 pixel - it is right. But I am not sure that 1 pixel image is a valid example
> 
> Yes, Actually (0, 0) won't be a problem but zero division can be. How to add a code to handle that case?

For instance segmentation we have this check if the length of contour more than 2. We also use it for semantic segmentation exportable code. I added same check for segmentation utils. If our contour less than 2 it will be skipped
It will prevent division by 0. 
Overall, masks won't be appeared on image with 1 pixel - we can consider to change this behavior everywhere, but I don't think it is necessary. 1 pixel image it is more a bug than a use case

image

goodsong81 (approved):

sungmanc (approved):

eunwoosh (approved):

PR #2156, opened 2023-05-15
Semi-SL Semantic Segmentation. Prototype View.

kprokofi:
### Checklist

### License

```python
```


kprokofi:
This PR includes new solution for Semi-SL approach. It is implemented for new models, SegNext. In the next PRs documentation will be updated with validation metrics on some public datasets used for validation. Also, experiments for tuning some hyperparameters are in progress in the background mode. 
For now, these changes (Prototype based approach) can achieve the following result:

model | cityscapes 1/16 | kitty_54 | VOC 1/16 | DISK 1/4 | city4 1/16 | voc_12 | Mean Dice gain
-- | -- | -- | -- | -- | -- | -- | -- 
ham_segnext_t: SUP | 55.93 | 62.35 | 73.82 | 86.87 | 68.3 | 68 | 0 
ham_segnext_t: MT | 59.2 | 66.68 | 76.14 | 87.4 | 68.9 | 69.78 | + 2.14 
ham_segnext_t: Proto 0.1 | **60.2** | **67.1** | **77.2** | **87.6** | **69.21** | **70.4** | + **2.74** 

Experiments with bigger models in progress.
Some results:
SegNext-s:
model | cityscapes 1/16 | kitty_54 | VOC 1/16 | voc_12 | Mean Dice gain
-- | -- | -- | -- | -- | --
ham_segnext_s: MT | 67.02 | 68.11	| 79.5 | **75.11** | 0
ham_segnext_s: Proto 0.1 | **69.71** | **68.54** | **80.51** | 75 | **+ 1%**

otx/algorithms/segmentation/adapters/mmseg/models/segmentors/mean_teacher_segmentor.py
```diff
@@ -21,19 +23,42 @@ class MeanTeacherSegmentor(BaseSegmentor):
...
 
-    def __init__(self, orig_type=None, unsup_weight=0.1, semisl_start_iter=30, **kwargs):
+    def __init__(
+        self,
+        orig_type=None,
+        unsup_weight=0.1,
+        proto_weight=0.7,
+        aux_weight=0.1,
+        drop_percent=80,
+        num_iters_per_epoch=6000,
```
eunwoosh: It's just a question. Is it ok to set default value of this? I think it can be changed depending on dataset size.

otx/algorithms/segmentation/adapters/mmseg/models/segmentors/mean_teacher_segmentor.py
```diff
@@ -21,19 +23,42 @@ class MeanTeacherSegmentor(BaseSegmentor):
     It creates two models and ema from one to the other for consistency loss.
     """
 
-    def __init__(self, orig_type=None, unsup_weight=0.1, semisl_start_iter=30, **kwargs):
+    def __init__(
```
eunwoosh, line 36: Could you add comments for init arguments? I know it doesn't exist originally, but it can makes much easier to maintain

otx/algorithms/segmentation/adapters/mmseg/models/segmentors/mean_teacher_segmentor.py
```diff
@@ -21,19 +23,42 @@ class MeanTeacherSegmentor(BaseSegmentor):
...
         self.model_t = build_segmentor(cfg)
+        self.use_prototype_head = False
         self.unsup_weight = unsup_weight
-
+        if proto_head is not None and hasattr(self.model_s.decode_head, "_forward_feature"):
+            self.proto_net = ProtoNet(num_classes=self.model_s.decode_head.num_classes, **proto_head)
+            self.use_prototype_head = True
+        elif proto_head is not None:
+            logger.warning(
+                "Prototype head isn't supported by this model. "
```
eunwoosh, line 69: how about to add comment about what will happen in this case?

otx/algorithms/segmentation/adapters/mmseg/models/segmentors/mean_teacher_segmentor.py
```diff
@@ -42,6 +67,49 @@ def encode_decode(self, img, img_metas):
         """Encode and decode images."""
         return self.model_s.encode_decode(img, img_metas)
 
+    def decode_proto_network(
```
eunwoosh, line 83: As above, could you add comment for arguments?

otx/algorithms/segmentation/adapters/mmseg/models/segmentors/mean_teacher_segmentor.py
```diff
@@ -42,6 +67,49 @@ def encode_decode(self, img, img_metas):
...
+        self, sup_input, gt_semantic_seg, unsup_input=None, pl_from_teacher=None, reweight_unsup=1.0
+    ):
+        """Forward prototype network, compute proto loss."""
+        # supervised branch
+        head_features_sup = self.model_s.decode_head._forward_feature(sup_input)
```
eunwoosh, line 100: Could you avoid to use protected function?

otx/algorithms/segmentation/adapters/mmseg/models/segmentors/mean_teacher_segmentor.py
```diff
@@ -42,6 +67,49 @@ def encode_decode(self, img, img_metas):
...
+        proto_out_supervised = self.proto_net(head_features_sup, gt_semantic_seg)
+        loss_proto = self.proto_net.losses(**proto_out_supervised, seg_label=gt_semantic_seg)
+        self._update_summary_loss(loss_proto, loss_weight=self.proto_weight)
+        # unsupervised branch
+        if unsup_input and pl_from_teacher:
+            head_features_unsup = self.model_s.decode_head._forward_feature(unsup_input)
```
eunwoosh, line 106: same as above, Could you avoid to use protected function?

otx/algorithms/segmentation/adapters/mmseg/models/segmentors/mean_teacher_segmentor.py
```diff
@@ -61,37 +129,61 @@ def forward_dummy(self, img, **kwargs):
     def forward_train(self, img, img_metas, gt_semantic_seg, **kwargs):
         """Forward train."""
         self.count_iter += 1
-        if self.semisl_start_iter > self.count_iter or "extra_0" not in kwargs:
+        self.losses["sum_loss"] = 0.0
+        if self.semisl_start_iter >= self.count_iter or "extra_0" not in kwargs:
```
eunwoosh, line 174: Is it your intention to change `>` to `>=` ?

otx/algorithms/segmentation/adapters/mmseg/task.py
```diff
@@ -329,6 +340,12 @@ def _train_model(
         # Data
         datasets = [build_dataset(cfg.data.train)]
 
+        if self._train_type == TrainType.Semisupervised:
+            # forward the knowledge of num iters per epoch to model for filter loss
+            bs_per_gpu = cfg.data.train_dataloader["samples_per_gpu"]
+            actual_bs = bs_per_gpu * len(cfg.gpu_ids) if cfg.distributed else bs_per_gpu
```
eunwoosh: You need to use `torch.distributed.get_world_size()` instead of `len(cfg.gpu_ids)`.

JihwanEom:
Could you please clarify which paper you've implemented, and provide a link to it? Am I correct in understanding that your implementation is based on "Semi-supervised Semantic Segmentation with Prototype-based Consistency Regularization" (https://arxiv.org/pdf/2210.04388.pdf)?

If this is the case, would you consider adopting a step-by-step approach before fully integrating it?
image

As you might be aware, the Mean Teacher model and CutMix-seg are commonly used baselines in academia. It might be beneficial to first explore the performance and training time trade-off for CutMix-seg and then incrementally add Prototype view functionalities.

I would also suggest trying to keep changes to the MeanTeacherSegmentor to a minimum by defining other class for Prototype view, as it could serve as a standard baseline architecture for semi-supervised semantic segmentation.

kprokofi:
> 
@JihwanEom 
I use this paper as base idea and this one: https://arxiv.org/abs/2203.15102 for implementation reference

Unfortunately, Cutmix-seg performs poorly in OTX, I conducted experiments with that including different probability as well as Soobee did and saw accuracy degradation. But, CutOut helped and in final solution I changed CutMix to CutOut.

What do you mean step by step? I integrated Prototype network and enhanced a bit MT with CutOut and Filter pixels with high entropy. You can use standard MeanTeacher as always, just remove protohead from config. Also, old models (HrNets) use MT, I didn't change their configs.
It is impossible to use different Class, because the base method it is Mean Teacher.

Many changes in that PR is just "black" some files

otx/algorithms/segmentation/adapters/mmseg/models/segmentors/mean_teacher_segmentor.py
kprokofi: It will be changed automatically in task during configuration:
image

otx/algorithms/segmentation/adapters/mmseg/models/segmentors/mean_teacher_segmentor.py
kprokofi: Ok

otx/algorithms/segmentation/adapters/mmseg/models/segmentors/mean_teacher_segmentor.py
kprokofi: Unfortunately, it is impossible, I think. We need to follow mmseg interface. In mmseg heads we have "_forward_feature()" method to receive features from head without classifiers. It is supported by many heads including FCN and ASPP, for example

otx/algorithms/segmentation/adapters/mmseg/models/segmentors/mean_teacher_segmentor.py
kprokofi: Because EMA hook uses the rule of "epoch == epoch_to_start_update", "self.semisl_start_iter >= self.count_iter" this should be aligned with EMA hook. So, we need to set >= to avoid situations, when EMA hook is not initiated, but unsupervised part is started

JihwanEom:
> 

Thank you for the detailed explanation. I agree that `CutOut` can bring the promising performance improvements, but I believe that `CutMix` may work in our situation. Have you investigated the performance tendencies for both the lite-hrnet and SegNext templates? (or include ResNets?)

image

According to the CutMix-seg paper (https://arxiv.org/pdf/1906.01916.pdf), CutMix brings significant performance gains compared to CutOut even with only 100 labeled images. I don't think there is a significant difference in the environment between our situation and the one described in the paper.

Even if CutMix does not work due to unknown issues and we use CutOut instead, we need to assess the net benefit of using CutOut without combining it with Prototype view. That was my intention for taking a step-by-step approach. Could you share the experiment results if you have already?

You can continue using the standard MeanTeacher as usual, just remove protohead from the configuration. Also, the old models (HrNets) use MeanTeacher, and I haven't changed their configurations.
=> My suggestion was to consider inheriting MeanTeacherSegmentor in the new architecture of PrototypeViewSegmentor. This would ensure maintainability and reproducibility for not only MT and new models. I recommend creating a new file that defines PrototypeViewSegmentor and inheriting from MeanTeacherSegmentor, because it's a base method for many various algorithms for semi-sl on semantic segmentation as you said. But it's also my individual opinion as OTX developer, please share us if other one have good design ideas.

kprokofi:
> My suggestion was to consider inheriting MeanTeacherSegmentor in the new architecture of PrototypeViewSegmentor. This would ensure maintainability and reproducibility for not only MT and new models. I recommend creating a new file that defines PrototypeViewSegmentor and inheriting from MeanTeacherSegmentor, because it's a base method for many various algorithms for semi-sl on semantic segmentation as you said.

I consider your proposal and even tried to do that, but I faced a problem with that. If I do that -> I will copy all the code from forward_train method. Why should we do that? 
I moved everything related to Prototypes to `decode_proto_network`. And there are new lines of code in main forward method:
image
image
I think it is ambigious to create different Segmentor and copy almost all code there.
We also have second option, not copy, but use parent forward and then compute proto based forward, but in that case I should call self.model_s.extract_feat twice. It is double inference

I would consider leave it in the main MeanTeacher framework as ProtoNetwork is additional method rather than main one.

kprokofi:
I have some experiments comparing CutMix, CutOut and base algo
image

But I would like to conduct it one more time with different implementation of CutMix, SegNext-s and without early stopping

otx/algorithms/segmentation/adapters/mmseg/models/segmentors/mean_teacher_segmentor.py
eunwoosh: I see. But I think it's better to set `None` as default value or remove default value because it can prevent using it without setting `num_iters_per_epoch` later.

otx/algorithms/segmentation/configs/configuration.yaml
```diff
@@ -149,7 +149,7 @@ learning_parameters:
     warning: null
   enable_early_stopping:
     affects_outcome_of: TRAINING
-    default_value: true
+    default_value: false
```
supersoob: Why is changing base config needed? Early stopping should to be true. I hope not to change base configs in this PR.

otx/algorithms/segmentation/adapters/mmseg/models/segmentors/mean_teacher_segmentor.py
```diff
@@ -21,19 +23,42 @@ class MeanTeacherSegmentor(BaseSegmentor):
...
-    def __init__(self, orig_type=None, unsup_weight=0.1, semisl_start_iter=30, **kwargs):
+    def __init__(
+        self,
+        orig_type=None,
+        unsup_weight=0.1,
+        proto_weight=0.7,
+        aux_weight=0.1,
+        drop_percent=80,
+        num_iters_per_epoch=6000,
+        semisl_start_iter=1,
```
supersoob: If it starts by epoch like L42, it would be better to rename as semisl_start_epoch. And is it find to start it just after only 1 epoch?

otx/algorithms/segmentation/adapters/mmseg/models/segmentors/mean_teacher_segmentor.py
supersoob: It looks like if self.use_prototype_head is False, it runs with naive mean teacher method. So I think it would be good to add the message something like 'mean teacher without prototype is used for semi-supervised learning.'

otx/algorithms/segmentation/adapters/mmseg/models/segmentors/mean_teacher_segmentor.py
```diff
@@ -42,6 +67,49 @@ def encode_decode(self, img, img_metas):
...
+        with torch.no_grad():
+            teacher_feat = self.model_t.extract_feat(ul_w_img)
+            teacher_out = self.model_t._decode_head_forward_test(teacher_feat, ul_img_metas)
+            teacher_out = resize(
+                input=teacher_out, size=ul_w_img.shape[2:], mode="bilinear", align_corners=self.align_corners
+            )
+            teacher_prob_unsup = torch.softmax(teacher_out, axis=1)
+            _, pl_from_teacher = torch.max(teacher_prob_unsup, axis=1, keepdim=True)
+
+        # drop pixels with high entropy
```
supersoob, line 129: I have a question about dropping pixels with high entropy. Is this from original prototype based paper? Does it drop all pixel predictions masking it as ignored label 255 rather than using threshold 0.5 for prediction logit? Drop percent 80 looks so high, which the prediction per label is all dismissed.

otx/algorithms/segmentation/adapters/mmseg/models/segmentors/mean_teacher_segmentor.py
```diff
@@ -61,37 +129,61 @@ def forward_dummy(self, img, **kwargs):
...
-        for (key, value) in loss_decode_u.items():
-            if value is None:
+        self._update_summary_loss(loss_decode)
+        self._update_summary_loss(loss_decode_u, loss_weight=self.unsup_weight * reweight_unsup)
+
+        if self.use_prototype_head:
+            # for proto head we need to derive head features
+            self.decode_proto_network(x, gt_semantic_seg, x_u, pl_from_teacher, reweight_unsup)
+
+        if hasattr(self.model_s, "auxiliary_head"):
```
supersoob: Do we have a case to keep auxiliary head in OTX?

otx/algorithms/segmentation/adapters/mmseg/models/utils/proto_utils.py
```diff
@@ -0,0 +1,131 @@
...
+    """Projection head to transfrom features space used further for prototype learning."""
+
+    def __init__(self, dim_in, proj_dim=256):
+        super(ProjectionHead, self).__init__()
+
+        self.proj = nn.Sequential(nn.Conv2d(dim_in, dim_in, 1), nn.ReLU(inplace=True), nn.Conv2d(dim_in, proj_dim, 1))
+
+    def forward(self, x):
+        """Farward method."""
+        return l2_normalize(self.proj(x))
```
supersoob: I think it's okay to refactor it F.normalize(x, p=2, dim=-1) to reduce call depth.

otx/algorithms/segmentation/adapters/mmseg/models/utils/proto_utils.py
```diff
@@ -0,0 +1,131 @@
...
+        tensor: an n-dimensional `torch.Tensor`
+        mean: the mean of the normal distribution
+        std: the standard deviation of the normal distribution
+        a: the minimum cutoff value
+        b: the maximum cutoff value
+    Examples:
+        >>> w = torch.empty(3, 5)
+        >>> nn.init.trunc_normal_(w).
+    """
+    return _no_grad_trunc_normal_(tensor, mean, std, a, b)
```
supersoob: Same here. I think it's okay to keep either trunc_normal or no_grad_trunc_normal.

otx/algorithms/segmentation/adapters/mmseg/models/heads/proto_head.py
```diff
@@ -0,0 +1,140 @@
...
+    distributed_sinkhorn,
+    momentum_update,
+    trunc_normal_,
+)
+
+
+@HEADS.register_module()
+class ProtoNet(BaseDecodeHead):
+    """Prototype based head.
+
```
supersoob, line 28: I think we need to add reference paper link.

supersoob (commented):
I think it's okay to keep use_prototype_head in original MeanTeacherSegmentor and enable it by parameter because it were eventually based on mean teacher as current @kprokofi's implementation. About Cutmix, even though I checked the transformed image is alright, I couldn't see the gain of it (but only 1-2% drop). In my investigation, at that time there were some doubtable points. First, it was needed to fix any confidence threshold for pseudo label. Second, inference needed to be done in teacher model which might more generalized than student. And I doubted the cross entropy loss using as consistency loss is affecting it, which it uses prediction from the model trained with few labeled model that might not well-generalized as gt. I don't exactly know what caused cutmix make worse but I hope these helped some for you to refer.

supersoob:
Could you add unit tests for prototype head and for changed mean teacher?

requirements/segmentation.txt
```diff
@@ -6,3 +6,4 @@ scikit-image
 mmdeploy==0.14.0
 timm==0.6.12
 pytorchcv
+einops==0.6.1
```
supersoob, line 9: Do we need this for running semisl? Did you check this version no crush with other packages?

otx/algorithms/segmentation/adapters/mmseg/models/segmentors/mean_teacher_segmentor.py
kprokofi: It is actually from U2PL paper, as I remember. I utilized this method since it works pretty well. The naming of drop pixels = 80% is misleading. I will rename it. Actually 20% of pixels are droped from start of training and this number decreases during training.

`percent_unreliable = (100 - drop_percent) * (1 - self.count_iter / self.filter_pixels_iters)
drop_percent = 100 if percent_unreliable <= 0 else 100 - percent_unreliable
 thresh = np.percentile(entropy[pl_from_teacher != 255].detach().cpu().numpy().flatten(), drop_percent)
 thresh_mask = entropy.ge(thresh).bool() * (pl_from_teacher != 255).bool()`
 
 So in first epoch we drop 20% (we compute 0.8 percentile and drop everything greater). In the following epochs this percentage decreases.

otx/algorithms/segmentation/adapters/mmseg/models/segmentors/mean_teacher_segmentor.py
kprokofi: Some methods, for example, deeplabv3 ResNet family supports Aux head. 
Actually, I conducted experiments with that for HRNets and SegNext and saw the perfomance drop. So, I think we can avoid keeping AUX head.

otx/algorithms/segmentation/adapters/mmseg/models/utils/proto_utils.py
kprokofi: Agree

otx/algorithms/segmentation/adapters/mmseg/models/segmentors/mean_teacher_segmentor.py
kprokofi: Agree

otx/algorithms/segmentation/configs/configuration.yaml
kprokofi: Sorry, it is accidently. I used it for some experiments. I will revert it back

requirements/segmentation.txt
kprokofi: I don't see any problems with other packages. Yes, unfortunately, this library intensively used by ProtoNetwork. It allows us to compute tensor operations efficiently. It is very popular library by the way:
https://github.com/arogozhnikov/einops

otx/algorithms/segmentation/adapters/mmseg/models/segmentors/mean_teacher_segmentor.py
kprokofi: According to my experiments, start semi-sl after one epoch is enough

eunwoosh:
LGTM, but as Soobee said, could you add unit test?

kprokofi:
@JihwanEom 
Please, find below experiments with augmentations on bigger model. Unfortunately, cutmix always performs worse in our OTX. I tried 2 different implementations, but result is the same - CutOut looks better

<head>

</head>

<div style='direction:ltr'>



Model | Cityscapes | VOC | Kitty_57
-- | -- | -- | --
SegNext-b: MT | 70.88 | 82.05 | 70.45
SegNext-b: MT +   pixel_filter + cutout | 71.87 | 82.35 | 73.67
SegNext-b: MT +   pixel_filter + cutmix | 70.43 | 81.80 | 69.44
SegNext-b: MT +   pixel_filter + cutout + ProtoNet | 72.10 | 82.74 | 74.31



</div>


</body>

</html>

kprokofi:
I added unit tests + integration tests for Semi-SL and e2e for new model (we need to start validate it at least one template)
Could you take a look and merge if it looks good to you?

JihwanEom:
> @JihwanEom Please, find below experiments with augmentations on bigger model. Unfortunately, cutmix always performs worse in our OTX. I tried 2 different implementations, but result is the same - CutOut looks better
> 
> Model	Cityscapes	VOC	Kitty_57
> SegNext-b: MT	70.88	82.05	70.45
> SegNext-b: MT + pixel_filter + cutout	71.87	82.35	73.67
> SegNext-b: MT + pixel_filter + cutmix	70.43	81.80	69.44
> SegNext-b: MT + pixel_filter + cutout + ProtoNet	72.10	82.74	74.31

Okay, thank you so much for experiment results and kind explanation. I understood.

jaegukhyun:
Could you set milestone for this PR?

eunwoosh (approved):

kprokofi:
> Could you set milestone for this PR?

Done. Could you approve if it is OK? Let's finally merge this. 
In the following PR I will update documentation accordingly

supersoob (approved):

supersoob (approved):

supersoob (approved):

supersoob (approved):

PR #2195, opened 2023-05-24
Auto train type detection (Extension for auto-config functionality)

kprokofi:
### Checklist

### License

```python
```


kprokofi:
In this PR introduced:

- [x] Auto training type detection based on the given dataset to CLI. Introduced "Auto" default training type
- [x] Added tests for corner cases (Unit + integration)

Currently to start Semi-SL training or Self-SL we should explicitly set training type argument
We can avoid it and make it optional.

otx train MobileNet-V3-large-1x --train-data-roots data/flower_photos/train --val-data-roots data/flower_photos/val  --unlabeled-data-roots tests/assets/imagenet_dataset --train-type Semisupervised

We can simplify this command line to:

otx train MobileNet-V3-large-1x --train-data-roots data/flower_photos/train --val-data-roots data/flower_photos/val 

	1) If directory passed to "--train-data-roots" includes "unlabeled_images" folder and this task supports Semi-SL training -> start Semi-SL training
	2) We also can specify --unlabeled-data-roots explicitly, but remove "--train-type" option
	3) If we pass to "--train-data-roots" just folder with images or without validation subset -> we start Self-SL training

Implementation:
Modify "_get_train_type()" -> if train_type is not specified then check train-data-roots structure -> if only folder with images or no validation images then start Self-SL, if unlabeled_images are presented in args or in the directory, then Semi-SL, overwise incremental.

Finally, using auto-config functionality users can run Semi-SL and Self-SL in same way as for Supervised training:

otx train --train-data-roots data/flower_photos/train

However, to distinguish between auto_split functionality and Self-SL we should explicitly pass training_type.

The related documentation with all of use cases will be added in the next PR

harimkang (changes requested):
Hi @kprokofi , 
I think this is a great idea for a feature.
However, I have a concern about its use in API (Geti as well).
As you know, API also needs `configuration.yaml` and `template.yaml` inside our config. 
But **they don't use ConfigManager**. For-example, Geti doesn't use a CLI, they use the API directly.
So `Auto` in `configuration.yaml` and `template.yaml` can be a problem.
I think a different implementation should be proposed.

1. Either recognize Auto in the API and modify it to change (Not in ConfigManager)
or
2. Revert (configuration.yaml and template.yaml) and use Auto detect only for the CLI. (The default is Incremental, but we suggest giving CLI users an option called `Auto`.) 

I can think of at least two ways to do this.

eunwoosh:
Yes, I agree with Harim's comment. I think it's a nice feature, but there may be many things to consider including not only what Harim said but also other things.
For example, current PR considers it as semi-SL case to set only training data root, but I think it can be confusing to user who expects auto split. 
I think it's better to merge it after having a meeting to share this feature to team. Could you hold a meeting?

kprokofi:
> Yes, I agree with Harim's comment. I think it's a nice feature, but there may be many things to consider including not only what Harim said but also other things. For example, current PR considers it as semi-SL case to set only training data root, but I think it can be confusing to user who expects auto split. I think it's better to merge it after having a meeting to share this feature to team. Could you hold a meeting?

Sure, I will think today how to avoid changing default training_type. 
The problem that we have now - it is wrong logic for Self-SL. User should expect run training with Self-SL using only folder with images, but now we should create full dataset structure to run Self-SL. I think it is wrong and we need to reconsider handling this. Moreover documentation for Self-SL also wrong. It says, that it can be launched with images path only, but it is not working...
Let's hold a meeting on the next week

kprokofi:
- [x] Reverted train type back, make train_type = None by default 
- [x] Checked that SupCon runs via additional parameter and it is not related to algo_backend
- [x] Remove overlap with auto-split. Unit tests fixed

We need to add support for running training with only images when training type is Self-SL. This will be done in next PRs after investigation

otx/cli/manager/config_manager.py
```diff
@@ -243,6 +241,96 @@ def auto_task_detection(self, data_roots: str) -> str:
...
+            """Check if a directory contains only images."""
+            import glob
+
+            pngs = glob.glob(f"{dir}/*.png")
+            jpgs = glob.glob(f"{dir}/*.jpg")
+            return len(pngs) > 0 or len(jpgs) > 0
+
+        def _check_semisl_requirements(train_dir, unlabeled_dir, thershold=0.07):
+            """Check if quantity of unlabeled images is sufficient for Semi-SL learning."""
+            all_unlabeled_images = len(os.listdir(unlabeled_dir))
```
harimkang: it would be possible to read a file called unlabeled-image-list, but that doesn't seem to be considered.
Also, this is assuming that the unlabeled data folder doesn't have subdirectories - I know that OTX currently load all images in subdirectories as well.
@sungmanc If you have any additional comments, please leave them below.

otx/cli/manager/config_manager.py
```diff
@@ -243,6 +241,96 @@ def auto_task_detection(self, data_roots: str) -> str:
...
+            else:
+                raise ValueError(
+                    "Unlabeled images isn't a directory or it doesn't exist. "
+                    "Please, check command line and directory path."
+                )
+
+        unlabeled_folder_name = [
+            item
+            for item in os.listdir(path_to_train_data)
+            if (item.startswith("unlabeled") and os.path.isdir(os.path.join(path_to_train_data, item)))
```
harimkang: The folder name should always contain `unlabeled`, which is good for now, but we may need to think about a better way.

otx/cli/manager/config_manager.py
```diff
@@ -243,6 +241,96 @@ def auto_task_detection(self, data_roots: str) -> str:
...
+        ]
+        if unlabeled_folder_name:
+            # if in directory we have several unlabeled data folders -> retrive first
+            path_to_unlabeled_data = Path(path_to_train_data / unlabeled_folder_name[0])
+            if not _check_semisl_requirements(path_to_train_data, path_to_unlabeled_data):
+                print(
```
harimkang: As I understand it, this is a case where `unlabeled_data_roots` was **not taken** as an argument, but an unlabeled folder was found, and the image was determined to be insufficient.

Is a WARNING necessary in this case? I think it's more natural to print a WARNING when automatically declare `Semisupervised` outside of an if statement. (before line 328)

Isn't it more likely that users will think it's `Supervised` by default because user didn't declare `unlabeled_data_roots`?
I think the following is natural.
With unlabeled_data_roots -> Detect Incremental, Then WARNING
No unlabeled_data_roots -> Detect SemiSL, Then WARNING
(Of course, that could just be me.)

otx/cli/manager/config_manager.py
```diff
@@ -243,6 +241,96 @@ def auto_task_detection(self, data_roots: str) -> str:
...
+            print("[*] Selfsupervised training type detected")
+            self.train_type = "Selfsupervised"
+            return
+
+        # if user explicitly passed unlabeled images folder
+        if self.args.unlabeled_data_roots is not None:
+            unlabeled_path = Path(self.args.unlabeled_data_roots)
+            if Path.is_dir(unlabeled_path):
+                if not _check_semisl_requirements(path_to_train_data, unlabeled_path):
+                    print(
```
harimkang: In this case, isn't the code to update to Incremental missing?

harimkang (commented):
Thanks for the great feature. I left a few comments.

otx/cli/manager/config_manager.py
```diff
@@ -243,6 +241,96 @@ def auto_task_detection(self, data_roots: str) -> str:
...
+                    "Please, check command line and directory path."
+                )
+
+        unlabeled_folder_name = [
+            item
+            for item in os.listdir(path_to_train_data)
+            if (item.startswith("unlabeled") and os.path.isdir(os.path.join(path_to_train_data, item)))
+        ]
+        if unlabeled_folder_name:
+            # if in directory we have several unlabeled data folders -> retrive first
```
eunwoosh: Are there any reason of only retrieving first? How about getting all images if there are more than one directory.

otx/cli/manager/config_manager.py
```diff
@@ -243,6 +241,96 @@ def auto_task_detection(self, data_roots: str) -> str:
...
+                    )
+                self.train_type = "Semisupervised"
+                return
+            else:
+                raise ValueError(
+                    "Unlabeled images isn't a directory or it doesn't exist. "
+                    "Please, check command line and directory path."
+                )
+
+        unlabeled_folder_name = [
```
eunwoosh: I think `_check_semisl_requirements` can include this part also. How about moving all part related to check semisl into the function?

tests/unit/cli/manager/test_config_manager.py
```diff
@@ -477,6 +472,51 @@ def test__get_train_type(self, mocker):
...
+            assert False
+        except ValueError:
+            assert True
+        config_manager.args.train_data_roots = "non_exist_dir"
+        config_manager.args.val_data_roots = "tests/assets/classification_dataset"
+        config_manager.args.unlabeled_data_roots = None
+        try:
+            config_manager._get_train_type(ignore_args=False)
+            assert False
+        except ValueError:
```
eunwoosh: it also works but it can be replaced by `with pytest.raises(ValueError):`

tests/unit/cli/manager/test_config_manager.py
```diff
@@ -477,6 +472,51 @@ def test__get_train_type(self, mocker):
...
+            os.mkdir("tests/assets/classification_dataset/unlabeled_images")
+            config_manager.args.train_data_roots = "tests/assets/classification_dataset"
+            config_manager.args.val_data_roots = "tests/assets/classification_dataset"
+            # test empty unlabeled folder
+            assert config_manager._get_train_type(ignore_args=False) == "Incremental"
+            os.mkdir("tests/assets/classification_dataset/unlabeled_images/some")
+            # test unsufficient number of images
+            assert config_manager._get_train_type(ignore_args=False) == "Incremental"
+        finally:
+            shutil.rmtree("tests/assets/classification_dataset/unlabeled_images")
```
eunwoosh: If you want to use temporary directory removed automatically, you can use `tempfile.TemporaryDirectory`.

tests/unit/cli/manager/test_config_manager.py
eunwoosh, line 1: As you know, generally, each test includes single test case to increase a readability. Could you split the test into each test cases? I think current one includes multiple test cases.

eunwoosh (commented):
Thanks for your work! I left some comments. please take a look

otx/cli/manager/config_manager.py
kprokofi: 1) I added a check before passing to that function that unlabeled_dir is a dir, not a file
2) Yes, current implementation considered just a dir with images, @sungmanc will Datumaro scan directories recursively if there are some in the tree?

otx/cli/manager/config_manager.py
kprokofi: Yeah, I wanted to combine some names like unlabeled_data, unlabeled_images, unlabeled-...
The better choice of naming can be considered, of course

otx/cli/manager/config_manager.py
kprokofi: I will add in the documentation (in the next PR, I think) expected behavior related to auto configuration. So, if user will have unlabeled data folder with images - it is expected, that SemiSL should be started. However, in some cases the number of images is insufficient to start SemiSL, so I added a warning

After this feature will be merged declaration of unlabeled_data_roots in arguments will be optional and for some users maybe it will be more convenient to create unlabeled_images in the data root and always omit unlabeled_data_roots declaration to start Semi-SL

However, I will add a message like: "The Semi-SL training type detected. {folder_name} will be used for unlabeled_data_roots"

kprokofi: What I wanted to say, when _check_semisl_requirements returns False - I think we should print a message with warning, that unlabeled images is not enough for good Semi-SL perfomance

otx/cli/manager/config_manager.py
kprokofi: Yes, you are right, thanks
Need to check tests as well, why this case wasn't caught

otx/cli/manager/config_manager.py
kprokofi: It is a complicated scenario I think, because in that case we should copy all images in one place and then pass the path further. If we have large number of images - it will be time consuming, unfortunately. Because of that, I use only the first found folder.

tests/unit/cli/manager/test_config_manager.py
kprokofi: I used os.mkdir because I need folder with exact name and exact location. In tempfile I didn't find such functionality. Do you know other way how to do that?

tests/unit/cli/manager/test_config_manager.py
kprokofi: Ok, I will divide tests

otx/cli/manager/config_manager.py
sungmanc: > 1. I added a check before passing to that function that unlabeled_dir is a dir, not a file
> 2. Yes, current implementation considered just a dir with images, @sungmanc will Datumaro scan directories recursively if there are some in the tree?

Yes, Datumaro can. However, in the case of single-level tree architecture (like ImageNet classification), it is hard to distinguish between supervised classification and unsupervised one.

otx/cli/manager/config_manager.py
```diff
@@ -243,6 +241,96 @@ def auto_task_detection(self, data_roots: str) -> str:
...
+        Overwise set Incremental training type.
+        """
+
+        def _check_is_only_images(dir):
+            """Check if a directory contains only images."""
+            import glob
+
+            pngs = glob.glob(f"{dir}/*.png")
+            jpgs = glob.glob(f"{dir}/*.jpg")
+            return len(pngs) > 0 or len(jpgs) > 0
```
sungmanc: Unfortunately, we need to consider other variants and capitals.
JPEG, jpeg, PNG, JPG, etc.

otx/cli/manager/config_manager.py
sungmanc: For the warning message, I think using a logger is a more appropriate way since users could check the .log file whenever they want to see. Although .log file is not automatically saved now, I hope it will be fixed later.

sungmanc (dismissed):
Thanks for the good feature, BTW, as you know, this kind of automation could raise unexpected errors and I have a little concern about this.  However, it is not crucial and just a minor concern.

otx/cli/manager/config_manager.py
kprokofi: I added recursive scan of the folder with unlabeled images to count the number

otx/cli/manager/config_manager.py
kprokofi: Done

otx/cli/manager/config_manager.py
kprokofi: I added logging.warning instead of print

otx/cli/manager/config_manager.py
kprokofi: Since we have to omit this part of functionality - the conversation can be resolved

kprokofi:
I found a problem with dataset detection format when using unlabeled data inside dataset folder. Also in imagenet format this folder will be detected as additional label class, so, unfortunately we can't omit path to unlabeled images in CLI

harimkang (dismissed):
LGTM, Now we just need to fix it so that the test passes.

tests/unit/cli/manager/test_config_manager.py
eunwoosh: I think you forgot to do that. Could you check it again?

kprokofi: Sorry, maybe I didn't understand you right. I divided tests cases to "test_auto_semisl_detection" and "test_auto_selfsl_detection". Didn't it what you mean?

kprokofi:
tests are passed

sungmanc (dismissed):

tests/unit/cli/manager/test_config_manager.py
eunwoosh: I mean a single test includes multiple test cases. For example, `test_auto_semisl_detection` tests `Incremental`  case and `Semisupervised` case also.

kprokofi: I divided tests one more time to test__get_train_type_selfsupervised(), test__get_train_type_semisuprvised, test__get_train_type_incremental to check test cases related to those train type only

eunwoosh: Thanks!

eunwoosh (approved):

kprokofi:
@harimkang @sungmanc could you approve one more time?

sovrasov (approved):

PR #2205, opened 2023-05-30
Introduce channel last to improve the OTX tasks

sungmanc:
### Summary

* Introduce channel_last parameter to all tasks (cls, det, i-seg, s-seg, action cls/det)
* Add unit-tests
* (TODO)(DRAFT) Benchmark table will be updated soon. total epoch is set to 10.

|   | Classification   | Detection  | Instance Segmentation  | Semantic Segmentation   | Action Classification | Action Detection |
|---|---|---|---|---|---|---|
| Accuracy (before)   | 0.59  | 0.46  | 0.35  | 0.82  | 0.60  | 0.61 |
| Accuracy (after)  | 0.59  | 0.46  | 0.35  | 0.80   | 0.62  | 0.62  |
| E2E time (before)  | 0:13:02  | 0:07:21  | 0:30:24  | 0:02:26  | 0:15:11  | 1:29:36  |
| E2E time (after)  | 0:12:52  | 0:07:23  | 0:30:33   | 0:02:28   | 0:15:04  |  1:28:45 |

### Checklist

- [X] I have added the description of my changes into CHANGELOG in my target branch (e.g., [CHANGELOG](https://github.com/openvinotoolkit/training_extensions/blob/develop/CHANGELOG.md) in develop).​

### License

- [X] I submit _my code changes_ under the same [Apache License](https://github.com/openvinotoolkit/training_extensions/blob/develop/LICENSE) that covers the project.

```python
```


goodsong81 (dismissed):

otx/recipes/stages/classification/incremental.yaml
```diff
@@ -22,3 +22,5 @@ evaluation:
 task_adapt:
   type: "mpa"
   op: "REPLACE"
+
+channel_last: True
```
jaegukhyun: Why don't we add this more general recipe so that selfsl and semisl also can be faster

jaegukhyun (commented):
I left a comment

eunwoosh (changes requested):
As far as I know, input data also should be transformed to channel last format. But this pr doesn't do that. Is it ok?
Here is what I read https://pytorch.org/tutorials/intermediate/memory_format_tutorial.html

sungmanc:
> As far as I know, input data also should be transformed to channel last format. But this pr doesn't do that. Is it ok? Here is what I read https://pytorch.org/tutorials/intermediate/memory_format_tutorial.html

No, converting channels for all input shapes is also kind of another method.
We can choose the one in both of two ways. 
1. by using model
2. by using every input
image

otx/recipes/stages/classification/incremental.yaml
sungmanc: Good suggestion, I'll apply

eunwoosh:
> > As far as I know, input data also should be transformed to channel last format. But this pr doesn't do that. Is it ok? Here is what I read https://pytorch.org/tutorials/intermediate/memory_format_tutorial.html
> 
> No, converting channels for all input shapes is also kind of another method. We can choose the one in both of two ways.
> 
> 1. by using model
> 2. by using every input
>    image

Yes, maybe I'm wrong. But I see below things also
image
it comes from here https://pytorch.org/blog/accelerating-pytorch-vision-models-with-channels-last-on-cpu/
Sorry for bothering, but could you check again and let me know if I'm wrong?

sungmanc:
> Yes, maybe I'm wrong. But I see below things also  it comes from here https://pytorch.org/blog/accelerating-pytorch-vision-models-with-channels-last-on-cpu/ Sorry for bothering, but could you check again and let me know if I'm wrong?

I'm not sure why pytorch document use both ways. However, we could check that the stride of output tensors were successfully converted in unit tests.

And here is another reference, https://github.com/mosaicml/composer/blob/dev/composer/algorithms/channels_last/channels_last.py

otx/recipes/stages/classification/incremental.yaml
sungmanc: 7b1445821eaf7bbd0c2ea71bb8803665f9e7e24c

eunwoosh (approved):
Thanks for your kind reply. I think torch just wants to do it certainly.

jaegukhyun (approved):

PR #2206, opened 2023-05-30
Change _MMInitializeDataLoader.get_inputs() output format

kprokofi:
### Checklist

### License

```python
```

### DIFF
--- a/otx/algorithms/classification/adapters/mmcls/nncf/builder.py
+++ b/otx/algorithms/classification/adapters/mmcls/nncf/builder.py
@@ -80,7 +80,7 @@ def build_nncf_classifier(  # pylint: disable=too-many-locals,too-many-statement
                 dataset_builder=mmcls_build_dataset,
             ),
             config,
-            subset="train",
+            subset="val",
             dataloader_builder=mmcls_build_dataloader,
             distributed=distributed,
         )
--- a/otx/algorithms/detection/adapters/mmdet/nncf/builder.py
+++ b/otx/algorithms/detection/adapters/mmdet/nncf/builder.py
@@ -91,7 +91,7 @@ def build_nncf_detector(  # pylint: disable=too-many-locals,too-many-statements
                 dataset_builder=mmdet_build_dataset,
             ),
             config,
-            subset="train",
+            subset="val",
             dataloader_builder=mmdet_build_dataloader,
             distributed=distributed,
         )
--- a/otx/algorithms/segmentation/adapters/mmseg/nncf/builder.py
+++ b/otx/algorithms/segmentation/adapters/mmseg/nncf/builder.py
@@ -92,7 +92,7 @@ def build_nncf_segmentor(  # noqa: C901  # pylint: disable=too-many-locals,too-m
                 dataset_builder=mmseg_build_dataset,
             ),
             config,
-            subset="train",
+            subset="val",
             dataloader_builder=mmseg_build_dataloader,
             distributed=distributed,
         )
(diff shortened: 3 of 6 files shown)


kprokofi:
CVS-112024
Need to change dataloader format from train to validation and use test functions on mmcv side

otx/algorithms/classification/adapters/mmcls/nncf/builder.py
```diff
@@ -80,7 +80,7 @@ def build_nncf_classifier(  # pylint: disable=too-many-locals,too-many-statement
                 dataset_builder=mmcls_build_dataset,
             ),
             config,
-            subset="train",
+            subset="val",
```
cih9088, line 83:
```suggestion
            subset="train",
```
The quantisation parameter init dataset should be a train subset not a validation subset.

otx/algorithms/detection/adapters/mmdet/nncf/builder.py
```diff
@@ -91,7 +91,7 @@ def build_nncf_detector(  # pylint: disable=too-many-locals,too-many-statements
                 dataset_builder=mmdet_build_dataset,
             ),
             config,
-            subset="train",
+            subset="val",
```
cih9088, line 94:
```suggestion
            subset="train",
```

otx/algorithms/segmentation/adapters/mmseg/nncf/builder.py
```diff
@@ -92,7 +92,7 @@ def build_nncf_segmentor(  # noqa: C901  # pylint: disable=too-many-locals,too-m
                 dataset_builder=mmseg_build_dataset,
             ),
             config,
-            subset="train",
+            subset="val",
```
cih9088, line 95:
```suggestion
            subset="train",
```

cih9088 (changes requested):

cih9088:
@nikita-savelyevv, Could you please double-check if it is a bug or not since it was implemented in this way when we maintained a forked mm repositories like following.
https://github.com/openvinotoolkit/mmsegmentation/blob/ote/mmseg/integration/nncf/compression.py#L125-L130
https://github.com/openvinotoolkit/mmdetection/blob/ote/mmdet/integration/nncf/compression.py#L127-L132

@kprokofi, It is not in hurry but we should investigate `ignore_scopes` in each `compression_config.json` as this change would make different NNCF internal graph for some models. We might not need some of them in `ignore_scopes`. If it is not included in this PR, at least we should add `TODO` comment somewhere.

kprokofi:
@cih9088 , I use not val, but train. It just constructs images differently (1 image per batch) for NNCF initialization. 
Moreover, I discussed with NNCF what subset is better to use for init procedure and it was suggested to me using validation subset instead of train, but I left train for now.
If we use train subset in dataloader - it will lead a problem in simple_test() evaluation function

I will add TODO for future ignore_scope investigation

otx/algorithms/classification/adapters/mmcls/nncf/builder.py
kprokofi: We use train images, with "val" parameter  we construct batch differently, in the way we need for NNCF initialization

cih9088: ahh, sorry, my bad. it was for dataloader not dataset.

kprokofi:
Locally, tests are passed (for cls). Waiting for CI. Some problem with connection also observed for detection tests
image

kprokofi:
Tests are passed

cih9088 (approved):

sungmanc (approved):

PR #2211, opened 2023-05-31
[Anomaly] 🐞 NNCF e2e bugfix

ashwinvaidya17:
### Summary

- Fixes e2e issue in nncf

### License

```python
```

## Tests

> _Note:_ This is tested only for Padim but STFPM should work

image


samet-akcay (approved):

samet-akcay (dismissed):

otx/algorithms/anomaly/tasks/nncf.py
```diff
@@ -144,9 +144,17 @@ def load_model(self, otx_model: Optional[ModelEntity]) -> AnomalyModule:
                     pl_modules[key] = model_data["model"][key]
             model_data["model"] = nncf_modules
 
+            dataloader = None
```
jaegukhyun: I think we can add a type hint for this line

otx/algorithms/anomaly/tasks/nncf.py
```diff
@@ -144,9 +144,17 @@ def load_model(self, otx_model: Optional[ModelEntity]) -> AnomalyModule:
...
+            if hasattr(self, "trainer") and hasattr(self.trainer, "datamodule"):
+                if self.trainer.datamodule.train_dataset is not None:
+                    dataloader = self.trainer.datamodule.train_dataloader()
+                elif self.trainer.datamodule.test_dataset is not None:
+                    dataloader = self.trainer.datamodule.test_dataloader()
+
             self.compression_ctrl, model.model = wrap_nncf_model(
                 model.model,
                 self.optimization_config["nncf_config"],
+                dataloader=dataloader,  # type:ignore
```
jaegukhyun, line 160: Is it okay dataloader is None? If it is okay then what case could be?

jaegukhyun (commented):
I have some minor comments. Could you take a look?

otx/algorithms/anomaly/tasks/nncf.py
ashwinvaidya17: If I understand correctly, dataloader will always be None in this case as `wrap_nncf_model` requires either a trained nncf model or a dataloader. In this case we are loading weights from a trained model into nncf. But @AlexanderDokuchaev can clarify this better as he is the owner of `wrap_nncf_model` code

otx/algorithms/anomaly/tasks/nncf.py
ashwinvaidya17: Done!

jaegukhyun (approved):

samet-akcay (approved):
thanks @ashwinvaidya17

PR #2219, opened 2023-06-05
Added Self-SL adapter for classification to handle case when only images for train data provided + Documentation update

kprokofi:
### Checklist

### License

```python
```


kprokofi:
The idea is to add fake label to "image dir" dataset format for classification in case of Self-SL when no annotations are provided. This will let us to construct OTX dataset and proceed with Byol unsupervised training providing just folder with any images

otx/core/data/adapter/classification_dataset_adapter.py
```diff
@@ -120,3 +127,31 @@ def _generate_classification_label_schema(
 
     def _select_data_type(self, data_candidates: Union[list, str]) -> str:
         return "imagenet" if "imagenet" in data_candidates else data_candidates[0]
+
+
+class SelfSLClassificationDatasetAdapter(ClassificationDatasetAdapter):
```
sungmanc, line 132: Could you add unit-test for this new class?

sungmanc (commented):
Overall, LGTM. BTW, how about setting the default training_type as `incremental`?
And triggering auto mode by inserting `--train-type auto`?

I know this is just my preference, so this is a minor comment.
In my opinion, middle~high level developer assumes the `supervised` manner is default.

sungchul1 (commented):
LGTM, but I have a question.
Do users should set only folder with images to use self-sl?
What about if users want self-sl w/ class folders?

kprokofi:
> LGTM, but I have a question. Do users should set only folder with images to use self-sl? What about if users want self-sl w/ class folders?

Sure, users can run Self-SL with image folder dataset as before. However, in that case we need to add `--train-type` option as I described in docs

otx/core/data/adapter/classification_dataset_adapter.py
kprokofi: Ok

kprokofi: Done

kprokofi:
> Overall, LGTM. BTW, how about setting the default training_type as `incremental`? And triggering auto mode by inserting `--train-type auto`?
> 
> I know this is just my preference, so this is a minor comment. In my opinion, middle~high level developer assumes the `supervised` manner is default.

Setting train_type as None by default is solution how to distinguish between actual incremental passed by user in CLI and just default value. As we discussed on the meeting it is the most simple and working solution for that.
IMO, setting new value for train_type "--auto" doesn't make sense since the main purpose of this feature is to set training type automatically without typing "--train type"

Now, there is no overlap with corner cases and using auto train type detection is safe. I documented also one corner case, when we pass image folder format to SelfSL task instead of images only. In that case we have overlap with auto split for incremental. I also added this note in documentation both in explanation and tutorial sections.

sungchul1:
@kprokofi Thanks for answering. Let's take a look at test issues.

sungmanc (commented):
Thanks for the explanation, before merging, we need to resolve test failures.

kprokofi:
I think it was a CI error
Locally everything is fine with OV unit tests
image

sungchul1 (approved):

sungmanc (approved):

PR #2224, opened 2023-06-07
Set persistent_workers and pin_memory as True

eunwoosh:
### Summary

Set persistent_workers and pin_memory as True in all tasks, which can decrease data loader time.
When training with persistent_workers in nncf, nncf saves dataloaders in cfg. It makes an error when deepcopy `cfg` in `train_model` function of each tasks because datalaoder with persistent_workers aren't pickleable.
So, I set `persistent_workers` as False in nncf.

### Checklist

### License

```python
```

### DIFF
--- a/otx/algorithms/detection/adapters/mmdet/configurer.py
+++ b/otx/algorithms/detection/adapters/mmdet/configurer.py
@@ -99,6 +99,7 @@ class DetectionConfigurer:
         patch_fp16(cfg)
         patch_adaptive_interval_training(cfg)
         patch_early_stopping(cfg)
+        patch_persistent_workers(cfg)
 
         if data_cfg is not None:
             align_data_config_with_recipe(data_cfg, cfg)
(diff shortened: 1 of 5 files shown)


otx/algorithms/detection/adapters/mmdet/configurer.py
```diff
@@ -219,6 +218,17 @@ def update_config(src, pipeline_options):
             cfg.data.train.org_type = cfg.data.train.type
             cfg.data.train.type = super_type
 
+        for subset in ["train", "val", "test", "unlabeled"]:
```
sungmanc: Couldn't we move these parts inside of `patch_persistent_workers`? It seems that we could do.

eunwoosh: It can affect to all mmXX tasks. But I think that's not bad idea to set their default value as True. I'll update it.

sungmanc (dismissed):

sungmanc (approved):

jaegukhyun (approved):

PR #2227, opened 2023-06-07
Per-class saliency maps for M-RCNN

negvet:
### Summary
Generate per-class saliency maps for maskrcnn model, by aggregating raw segmentation masks.

For OV IR model maps are getting generated from filtered raw model predictions of mask head (implemented as postprocessing in `OTXMaskRCNNModel` wrapper). 
For Torch model maps are getting from unfiltered raw model predictions of mask head (implemented as `RecordingForwardHook`).

Resolves CVS-111726.

To be done in a separate PR: add e2e tests for maskrcnn

Some results (OV iR model, coco dataset):
image

### Checklist

### License

```python
```

### DIFF
--- a/otx/algorithms/common/adapters/mmcv/hooks/recording_forward_hook.py
+++ b/otx/algorithms/common/adapters/mmcv/hooks/recording_forward_hook.py
@@ -82,12 +82,11 @@ class BaseRecordingForwardHook(ABC):
             self._records.append(tensor)
 
     def _torch_to_numpy_from_list(self, tensor_list: List[Optional[torch.Tensor]]):
-        if isinstance(tensor_list[0], list):
-            self._torch_to_numpy_from_list(tensor_list[0])
-        else:
-            for i in range(len(tensor_list)):
-                if isinstance(tensor_list[i], torch.Tensor):
-                    tensor_list[i] = tensor_list[i].detach().cpu().numpy()
+        for i in range(len(tensor_list)):
+            if isinstance(tensor_list[i], list):
+                self._torch_to_numpy_from_list(tensor_list[i])
+            elif isinstance(tensor_list[i], torch.Tensor):
+                tensor_list[i] = tensor_list[i].detach().cpu().numpy()
 
     def __enter__(self) -> BaseRecordingForwardHook:
         """Enter."""
--- a/otx/algorithms/common/adapters/mmcv/hooks/recording_forward_hook.py
+++ b/otx/algorithms/common/adapters/mmcv/hooks/recording_forward_hook.py
@@ -16,8 +16,9 @@
 from __future__ import annotations
 
 from abc import ABC
-from typing import List, Sequence, Union
+from typing import List, Optional, Sequence, Union
 
+import numpy as np
 import torch
 
 from otx.algorithms.classification import MMCLS_AVAILABLE
@@ -69,9 +70,24 @@ class BaseRecordingForwardHook(ABC):
         self, _: torch.nn.Module, x: torch.Tensor, output: torch.Tensor
     ):  # pylint: disable=unused-argument
         tensors = self.func(output)
-        tensors = tensors.detach().cpu().numpy()
-        for tensor in tensors:
+        if isinstance(tensors, torch.Tensor):
+            tensors_np = tensors.detach().cpu().numpy()
+        elif isinstance(tensors, np.ndarray):
+            tensors_np = tensors
+        else:
+            self._torch_to_numpy_from_list(tensors)
+            tensors_np = tensors
+
+        for tensor in tensors_np:
             self._records.append(tensor)
+
+    def _torch_to_numpy_from_list(self, tensor_list: List[Optional[torch.Tensor]]):
+        if isinstance(tensor_list[0], list):
+            self._torch_to_numpy_from_list(tensor_list[0])
+        else:
+            for i in range(len(tensor_list)):
+                if isinstance(tensor_list[i], torch.Tensor):
+                    tensor_list[i] = tensor_list[i].detach().cpu().numpy()
 
     def __enter__(self) -> BaseRecordingForwardHook:
         """Enter."""
--- a/otx/algorithms/detection/adapters/mmdet/hooks/det_class_probability_map_hook.py
+++ b/otx/algorithms/detection/adapters/mmdet/hooks/det_class_probability_map_hook.py
@@ -178,7 +178,7 @@ class MaskRCNNRecordingForwardHook(BaseRecordingForwardHook):
         batch_size = x[0].shape[0]
         img_metas = [
             {
-                "scale_factor": [1, 1, 1, 1],
+                "scale_factor": [1, 1, 1, 1],  # dummy scale_factor, not used
                 "img_shape": self._input_img_shape,
             }
         ]
--- a/otx/algorithms/detection/adapters/mmdet/hooks/det_class_probability_map_hook.py
+++ b/otx/algorithms/detection/adapters/mmdet/hooks/det_class_probability_map_hook.py
@@ -126,13 +126,15 @@ class DetClassProbabilityMapHook(BaseRecordingForwardHook):
         return cls_scores
 
 
-class MaskRCNNHook(BaseRecordingForwardHook):
+class MaskRCNNRecordingForwardHook(BaseRecordingForwardHook):
     """Saliency map hook for Mask R-CNN model. Only for torch model, does not support OV IR model.
 
     Args:
         module (torch.nn.Module): Mask R-CNN model.
         input_img_shape (Tuple[int]): Resolution of the model input image.
         saliency_map_shape (Tuple[int]): Resolution of the output saliency map.
+        max_detections_per_img (int): Upper limit of the number of detections
+            from which soft mask predictions are getting aggregated.
         normalize (bool): Flag that defines if the output saliency map will be normalized.
             Although, partial normalization is anyway done by segmentation mask head.
     """
@@ -142,12 +144,14 @@ class MaskRCNNHook(BaseRecordingForwardHook):
         module: torch.nn.Module,
         input_img_shape: Tuple[int, int],
         saliency_map_shape: Tuple[int, int] = (224, 224),
+        max_detections_per_img: int = 300,
         normalize: bool = True,
     ) -> None:
         super().__init__(module)
         self._neck = module.neck if module.with_neck else None
         self._input_img_shape = input_img_shape
         self._saliency_map_shape = saliency_map_shape
+        self._max_detections_per_img = max_detections_per_img
         self._norm_saliency_maps = normalize
 
     def func(
@@ -181,9 +185,9 @@ class MaskRCNNHook(BaseRecordingForwardHook):
         img_metas *= batch_size
         proposals = self._module.rpn_head.simple_test_rpn(x, img_metas)
         test_cfg = copy.deepcopy(self._module.roi_head.test_cfg)
-        test_cfg["max_per_img"] = 300
+        test_cfg["max_per_img"] = self._max_detections_per_img
         test_cfg["nms"]["iou_threshold"] = 1
-        test_cfg["nms"]["max_num"] = 300
+        test_cfg["nms"]["max_num"] = self._max_detections_per_img
         det_bboxes, det_labels = self._module.roi_head.simple_test_bboxes(
             x, img_metas, proposals, test_cfg, rescale=False
         )
--- a/otx/algorithms/detection/adapters/mmdet/hooks/det_class_probability_map_hook.py
+++ b/otx/algorithms/detection/adapters/mmdet/hooks/det_class_probability_map_hook.py
@@ -127,7 +127,7 @@ class DetClassProbabilityMapHook(BaseRecordingForwardHook):
 
 
 class MaskRCNNRecordingForwardHook(BaseRecordingForwardHook):
-    """Saliency map hook for Mask R-CNN model. Only for torch model, does not support OV IR model.
+    """Saliency map hook for Mask R-CNN model. Only for torch model, does not support OpenVINO IR model.
 
     Args:
         module (torch.nn.Module): Mask R-CNN model.
--- a/otx/algorithms/detection/adapters/mmdet/hooks/det_class_probability_map_hook.py
+++ b/otx/algorithms/detection/adapters/mmdet/hooks/det_class_probability_map_hook.py
@@ -2,10 +2,13 @@
 # Copyright (C) 2023 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 #
-from typing import List, Tuple, Union
+import copy
+from typing import List, Optional, Tuple, Union
 
+import numpy as np
 import torch
 import torch.nn.functional as F
+from mmdet.core import bbox2roi
 
 from otx.algorithms.common.adapters.mmcv.hooks.recording_forward_hook import (
     BaseRecordingForwardHook,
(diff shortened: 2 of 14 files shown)


otx/algorithms/detection/adapters/mmdet/hooks/det_class_probability_map_hook.py
```diff
@@ -121,3 +124,123 @@ def forward_single(x, cls_convs, conv_cls):
...
+        img_metas = [
+            {
+                "scale_factor": [1, 1, 1, 1],
+                "img_shape": self._input_img_shape,
+            }
+        ]
+        img_metas *= batch_size
+        proposals = self._module.rpn_head.simple_test_rpn(x, img_metas)
+        test_cfg = copy.deepcopy(self._module.roi_head.test_cfg)
+        test_cfg["max_per_img"] = 300
```
sungmanc: Will we use the fixed constant for this? Does it need to be considered as a configurable parameter? For those who wants to check the saliency map, they might want to control the number of objects per image.

otx/algorithms/detection/adapters/mmdet/hooks/det_class_probability_map_hook.py
```diff
@@ -121,3 +124,123 @@ def forward_single(x, cls_convs, conv_cls):
                     "YOLOXHead, ATSSHead, SSDHead, VFNetHead."
                 )
         return cls_scores
+
+
+class MaskRCNNHook(BaseRecordingForwardHook):
```
sungmanc: * How about changing the name of this class? `MaskRCNNRecordingForwardHook`? `MaskRCNNHook` itself is not proper naming for only the saliency map. 

* We could set the additional parent class for the detection task (couldn't we use the `DetClassProbabilityMapHook`?) for further implementations if we have the plan to implement other hooks such as `ATSSRecordingForwardHook`. And in this case, we also move `normalize` function to parent class for further use.

otx/algorithms/detection/adapters/mmdet/task.py
```diff
@@ -399,7 +402,13 @@ def hook(module, inp, outp):
...
                 raw_model = raw_model.get_nncf_wrapped_model()
             if isinstance(raw_model, TwoStageDetector):
-                saliency_hook = ActivationMapHook(feature_model)
+                test_pipeline = cfg.data.test.pipeline
+                width, height = None, None
+                for pipeline in test_pipeline:
+                    width, height = pipeline.get("img_scale", (None, None))
+                if height is None:
+                    raise ValueError("img_scale has to be defined in the test pipeline.")
+                saliency_hook = MaskRCNNHook(feature_model, input_img_shape=(height, width))
```
sungmanc: Will we use `MaskRCNNHook` for all of the two-stage detection models? If so, this name should be changed.

otx/algorithms/detection/adapters/mmdet/hooks/det_class_probability_map_hook.py
negvet: Thanks for the comment, I agree. Fixed

otx/algorithms/detection/adapters/mmdet/hooks/det_class_probability_map_hook.py
negvet: I agree that `MaskRCNNRecordingForwardHook` might be more concreate. Updated

Regarding the OOP refactoring .
`DetClassProbabilityMapHook` is used for all single-stage detectors.
`normalize` is pretty general for all task types, so it can be moved to the upper-most level - `BaseRecordingForwardHook`. But I propose to do it in a separate PR, after @GalyaZalesskaya will also finish with sal map merge for tiling.
Except normalization, I do not see common code between `DetClassProbabilityMapHook` and `MaskRCNNRecordingForwardHook`, so probbaly for now base class for detectors is not needed.

otx/algorithms/detection/adapters/mmdet/task.py
negvet: `MaskRCNNRecordingForwardHook` is designed for MaskRCNN. It will not work with other two-stage detector.
But, afaik so far we do not have plans to support any other two-stage detector rather than m-rcnn, so, to keep it simple and clear, I propose to keep it as is (rather that support two cases: 1. maskrcnn and 2. some other general two-stage detector). What do you think?

otx/algorithms/common/adapters/mmcv/hooks/recording_forward_hook.py
```diff
@@ -69,10 +70,25 @@ def _recording_forward(
...
+            tensors_np = tensors
+        else:
+            self._torch_to_numpy_from_list(tensors)
+            tensors_np = tensors
+
+        for tensor in tensors_np:
             self._records.append(tensor)
 
+    def _torch_to_numpy_from_list(self, tensor_list: List[Optional[torch.Tensor]]):
+        if isinstance(tensor_list[0], list):
```
eunwoosh: It's just a question. Is there any reason to check that only first element is list type?

otx/algorithms/detection/adapters/mmdet/hooks/det_class_probability_map_hook.py
```diff
@@ -121,3 +124,127 @@ def forward_single(x, cls_convs, conv_cls):
...
+            saliency_maps = self._get_saliency_maps_from_mask_predictions(feature_map, det_bboxes, det_labels)
+            if self._norm_saliency_maps:
+                saliency_maps = self._normalize(saliency_maps)
+        return saliency_maps
+
+    def _get_detections(self, x: torch.Tensor) -> Tuple[List[torch.Tensor], List[torch.Tensor]]:
+        batch_size = x[0].shape[0]
+        img_metas = [
+            {
+                "scale_factor": [1, 1, 1, 1],
```
sungmanc: In my understanding, if the default `scale_factor` of Mask-RCNN is changed, this value is also needed to chage, right? Then, couldn't we use the `img_meta` field to make it more adaptive?

otx/algorithms/detection/adapters/mmdet/task.py
```diff
@@ -610,7 +630,7 @@ def hook(module, inp, outp):
 
         # Class-wise Saliency map for Single-Stage Detector, otherwise use class-ignore saliency map.
```
eunwoosh: Isn't it necessary to change a comment?

otx/algorithms/detection/adapters/mmdet/task.py
```diff
@@ -14,6 +14,7 @@
 # See the License for the specific language governing permissions
 # and limitations under the License.
 
+import functools
```
eunwoosh:
```suggestion
```
Is it necessary to import?

otx/algorithms/detection/adapters/mmdet/task.py
```diff
@@ -61,7 +62,9 @@
 from otx.algorithms.detection.adapters.mmdet.datasets import ImageTilingDataset
 from otx.algorithms.detection.adapters.mmdet.hooks.det_class_probability_map_hook import (
     DetClassProbabilityMapHook,
+    MaskRCNNRecordingForwardHook,
 )
+from otx.algorithms.detection.adapters.mmdet.models.detectors import CustomMaskRCNN
```
eunwoosh:
```suggestion
```
I think it isn't used

otx/algorithms/detection/adapters/mmdet/task.py
```diff
@@ -399,7 +402,13 @@ def hook(module, inp, outp):
             if raw_model.__class__.__name__ == "NNCFNetwork":
                 raw_model = raw_model.get_nncf_wrapped_model()
             if isinstance(raw_model, TwoStageDetector):
-                saliency_hook = ActivationMapHook(feature_model)
+                test_pipeline = cfg.data.test.pipeline
+                width, height = None, None
+                for pipeline in test_pipeline:
+                    width, height = pipeline.get("img_scale", (None, None))
```
negvet: Is there a better way to get input image size?

otx/algorithms/detection/adapters/mmdet/hooks/det_class_probability_map_hook.py
negvet: Here, `img_metas` define `scale_factor` just to make detection head not breaking when calling it, in fact it is not used so it is dummy here. Actually only `img_shape` is used for rescaling.
I would use `img_meta` if it would be available at the hook level, but unfortunately it is not, the same as input image.

otx/algorithms/common/adapters/mmcv/hooks/recording_forward_hook.py
negvet: Assuming that this is a recursive function that can handle nested lists, `isinstance(tensor_list[0], list)` defines a recursive case when the elements in the list are still lists rather than tensors -> then we go to the next level down.

otx/algorithms/detection/adapters/mmdet/task.py
negvet: Thanks! Probably there is no more need for this comment

otx/algorithms/detection/adapters/mmdet/task.py
negvet: Corrected!

otx/algorithms/detection/adapters/mmdet/task.py
negvet: It was us nut I replaced it with `isinstance(feature_model, TwoStageDetector)`. Thanks!

otx/algorithms/detection/adapters/mmdet/hooks/det_class_probability_map_hook.py
sungmanc: Thanks for kind reply

otx/algorithms/common/adapters/mmcv/hooks/recording_forward_hook.py
eunwoosh: Thanks for your reply! As my understand, this code transforms only first element from list to numpy and leaves others be. I wonder it's your intention.

otx/algorithms/detection/adapters/mmdet/task.py
eunwoosh: `mm_dataset[0]['img_metas']` can be one of options

jaegukhyun: Yes, mm_dataset is the actual dataset for inference. You can use it to find input tensor's size

otx/algorithms/common/adapters/mmcv/hooks/recording_forward_hook.py
negvet: Actually this function convert all tensors to numpy from the nested list. The first element is just used to check if we reached the lower-level list.

eunwoosh: I may think wrongly, but I think it doesn't work when second element is also list. Isn't code need to be changed as below to transform all list?
```
    def _torch_to_numpy_from_list(self, tensor_list: List[Optional[torch.Tensor]]):
        for i in range(len(tensor_list)):
            if isinstance(tensor_list[i], list):
                self._torch_to_numpy_from_list(tensor_list[i])
            elif isinstance(tensor_list[i], torch.Tensor):
                tensor_list[i] = tensor_list[i].detach().cpu().numpy()
```
It could be wrong, so please let me know when it is :)

otx/algorithms/detection/adapters/mmdet/hooks/det_class_probability_map_hook.py
GalyaZalesskaya: I moved `normalize` parameter to `BaseRecordingForwardHook` in my [PR#2240](https://github.com/openvinotoolkit/training_extensions/pull/2240) related to XAI for tiling. After merging the current PR I'll rebase my PR onto this, and finish refactoring of saliency map hooks

otx/algorithms/detection/adapters/mmdet/hooks/det_class_probability_map_hook.py
```diff
@@ -121,3 +124,127 @@ def forward_single(x, cls_convs, conv_cls):
...
+    @staticmethod
+    def _normalize(saliency_maps: List[List[Optional[np.ndarray]]]) -> List[List[Optional[np.ndarray]]]:
+        batch_size = len(saliency_maps)
+        num_classes = len(saliency_maps[0])
+        for i in range(batch_size):
+            for class_id in range(num_classes):
+                per_class_map = saliency_maps[i][class_id]
+                if per_class_map is not None:
+                    max_values = np.max(per_class_map)
+                    per_class_map = 255 * (per_class_map) / (max_values + 1e-12)
```
GalyaZalesskaya, line 247: I don't completely understand, in which cases `per_class_map ` could be None, should we check it?
And also,  we know that saliency map min_values are 0 in any case, that's why we don't need subtract them during normalization?
```
max_values = np.max(per_class_map)
min_values = np.min(per_class_map)
per_class_map = 255 * (per_class_map- **min_values** ) / (max_values + 1e-12)
```

otx/algorithms/detection/adapters/mmdet/task.py
```diff
@@ -608,9 +626,8 @@ def hook(module, inp, outp):
             raise NotImplementedError(f"Explainer algorithm {explainer} not supported!")
         logger.info(f"Explainer algorithm: {explainer}")
 
-        # Class-wise Saliency map for Single-Stage Detector, otherwise use class-ignore saliency map.
         eval_predictions = []
-        with explainer_hook(feature_model) as saliency_hook:
+        with explainer_hook(feature_model) as saliency_hook:  # type: ignore
```
GalyaZalesskaya, line 620: The pylance issue with incompatible type can be solved if we'd define the type of saliency_hook above as Union[MaskRCNNRecordingForwardHook, BaseRecordingForwardHook, DetClassProbabilityMapHook].

If that's the reason why you put ignore flag here.
The same is applicable to `# type: ignore` above

otx/algorithms/detection/adapters/mmdet/hooks/det_class_probability_map_hook.py
```diff
@@ -121,3 +124,127 @@ def forward_single(x, cls_convs, conv_cls):
                     "YOLOXHead, ATSSHead, SSDHead, VFNetHead."
                 )
         return cls_scores
+
+
+class MaskRCNNRecordingForwardHook(BaseRecordingForwardHook):
+    """Saliency map hook for Mask R-CNN model. Only for torch model, does not support OV IR model.
```
GalyaZalesskaya: If you don't mind, can you substitute OV with OpenVINO IR model?

otx/algorithms/detection/adapters/mmdet/hooks/det_class_probability_map_hook.py
negvet: `per_class_map = None` covers the cases, when we cannot generate any saliency map for this particular target class. This can be the case for e.g. `MaskRCNNRecordingForwardHook`, because it actually generate sal maps from model predictions ( it is rare the case when all possible classes are getting predicted for each image)

negvet: > we know that saliency map min_values are 0 in any case

This is not true in the general case. For example if you remove softmax activation from `DetClassProbabilityMapHook`, there will be negative values (because there can be negative values in the detection activations). So I would propose to keep it for the seek of generality, especially if we will unify the normalization for all methods.

otx/algorithms/detection/adapters/mmdet/task.py
negvet: `MaskRCNNRecordingForwardHook` and `DetClassProbabilityMapHook` are both subclasses of `BaseRecordingForwardHook`.
The issue actually happening due to `functools.partial(MaskRCNNRecordingForwardHook, ...)`, removing it do not raise the `error: "object" not callable  [operator]`.
Frankly, I do not know the fast way to fix it.

otx/algorithms/detection/adapters/mmdet/hooks/det_class_probability_map_hook.py
negvet: Sure

otx/algorithms/common/adapters/mmcv/hooks/recording_forward_hook.py
negvet: Your proposal is more general, handles case when the list can contain tensor and another list, while current approach is limited, although still working. I adopt your proposal. Thanks!

otx/algorithms/detection/adapters/mmdet/hooks/det_class_probability_map_hook.py
GalyaZalesskaya: I'm sorry if I asked not really clearly, but it was a question: why don't we calculate and subtract min_values during normalization?

GalyaZalesskaya: `per_class_map = None covers the cases, when we cannot generate any saliency map for this particular target class.`
Got it, thank you. I thought, in that case, we'll get a saliency map with all zeros, so we don't need to make a special way to postprocess them

otx/algorithms/detection/adapters/openvino/model_wrappers/openvino_models.py
```diff
@@ -107,6 +108,85 @@ def postprocess(self, outputs, meta):
...
+        boxes[:, 0::2] /= scale_x
+        boxes[:, 1::2] /= scale_y
+
+        saliency_maps = [None for _ in range(num_classes)]
+        for box, score, cls, raw_mask in zip(boxes, scores, classes, masks):
+            resized_mask = self._resize_mask(box, raw_mask * score, *meta["original_shape"][:-1])
+            if saliency_maps[cls] is None:
+                saliency_maps[cls] = [resized_mask]
+            else:
+                saliency_maps[cls].append(resized_mask)
```
GalyaZalesskaya, line 136:
```suggestion
    saliency_maps = [[] for _ in range(num_classes)]
    for box, score, cls, raw_mask in zip(boxes, scores, classes, masks):
        resized_mask = self._resize_mask(box, raw_mask * score, *meta["original_shape"][:-1])
        saliency_maps[cls].append(resized_mask)
```

GalyaZalesskaya: Oh, nevermind. Some saliency maps for undetected classes should stay as a None to fit the rest of the processing logic

otx/algorithms/detection/adapters/openvino/task.py
```diff
@@ -527,7 +543,7 @@ def add_prediction(id: int, predicted_scene: AnnotationSceneEntity, aux_data: tu
 
             if add_saliency_map and saliency_map is not None:
                 labels = self.task_environment.get_labels().copy()
-                if saliency_map.shape[0] == len(labels) + 1:
+                if len(saliency_map) == len(labels) + 1:
```
GalyaZalesskaya, line 546: Can I ask one more question to check if I understood correctly? 
Does it mean that before saliency map was an array with `(num_classes, feat_w, feat_h)` shape, 
and now it's a list of `num_classes` np.arrays with `(feat_w, feat_h)` size or `None`, if this class wasn't detected?

otx/algorithms/detection/adapters/mmdet/hooks/det_class_probability_map_hook.py
negvet: > I'm sorry if I asked not really clearly, but it was a question: why don't we calculate and subtract min_values during normalization?

We should in general case - when/if we move normalization to the common part for all methods. In this particular example, for `MaskRCNNRecordingForwardHook` map combined from segmentation masks, which cannot be negative (min value is zero). So it did not subtracted min_values

negvet: > I thought, in that case, we'll get a saliency map with all zeros, so we don't need to make a special way to postprocess them

This is an option. But the downside of it is memory overhead

otx/algorithms/detection/adapters/openvino/task.py
negvet: Yeah, that is correct. 
For now, these data strictures can handle all the cases that we have.

But if/when we will support per-box as well, we will need to update it probably.

tests/integration/api/detection/api_detection.py
```diff
@@ -0,0 +1,98 @@
+"""API Tests for detection training"""
```
sovrasov, line 1: Is it launched on CI?

sovrasov (commented):
Overall LGTM, though a bit hacky

otx/algorithms/detection/adapters/mmdet/hooks/det_class_probability_map_hook.py
GalyaZalesskaya: Thank you for the explanation. It's clear now

GalyaZalesskaya (approved):

eunwoosh (approved):

sovrasov (approved):

PR #2228, opened 2023-06-08
Add custom max iou assigner to prevent CPU OOM in training phase

jaegukhyun:
### Summary
When a training image contain too many gt(over > 10000), current assigner, matching anchor bbox and gt, raises CPU OOM.
This is because assigner makes too large cpu tensor matrix. To fix this problem, this PR add a new custom max iou assigner.
This new custom max iou assigner splits gt bboxes into reasonable size when gt bboxes is too large. This may lead time increase but prevent sudden cpu oom. 

### How to test
In my desktop which has 64GB RAM cannot train Kiemgetal dataset. However after this fix, my desktop can train Kiemgetal dataset.

### Checklist

### License

```python
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to this project will be documented in this file.
 ### New features
 
 - Support encrypted dataset training (<https://github.com/openvinotoolkit/training_extensions/pull/2209>)
-- Add custom max iou assigner (<https://github.com/openvinotoolkit/training_extensions/pull/2228>)
+- Add custom max iou assigner to prevent CPU OOM when large annotations are used (<https://github.com/openvinotoolkit/training_extensions/pull/2228>)
 
 ### Enhancements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file.
 ### New features
 
 - Support encrypted dataset training (<https://github.com/openvinotoolkit/training_extensions/pull/2209>)
+- Add custom max iou assigner (<https://github.com/openvinotoolkit/training_extensions/pull/2228>)
 
 ### Enhancements
 
--- a/otx/algorithms/detection/adapters/mmdet/models/assigners/custom_max_iou_assigner.py
+++ b/otx/algorithms/detection/adapters/mmdet/models/assigners/custom_max_iou_assigner.py
@@ -73,14 +73,11 @@ class CustomMaxIoUAssigner(MaxIoUAssigner):
 
         if assign_on_cpu and gt_bboxes.shape[0] > self.cpu_assign_thr:
             split_length = gt_bboxes.shape[0] // self.cpu_assign_thr + 1
-            overlaps = None
+            overlaps = []
             for i in range(split_length):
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file.
 ### New features
 
 - Support encrypted dataset training (<https://github.com/openvinotoolkit/training_extensions/pull/2209>)
+- Add custom max iou assigner (<https://github.com/openvinotoolkit/training_extensions/pull/2228>)
```
sungmanc:
```suggestion
- Add custom max iou assigner to prevent CPU OOM when large annotations are used (<https://github.com/openvinotoolkit/training_extensions/pull/2228>)
```

otx/algorithms/detection/adapters/mmdet/models/assigners/custom_max_iou_assigner.py
```diff
@@ -0,0 +1,107 @@
...
+            overlaps = None
+            for i in range(split_length):
+                gt_bboxes_split = gt_bboxes[i * self.cpu_assign_thr : (i + 1) * self.cpu_assign_thr]
+                if overlaps is None:
+                    overlaps = self.iou_calculator(gt_bboxes_split, bboxes)
+                else:
+                    overlaps = torch.concat((overlaps, self.iou_calculator(gt_bboxes_split, bboxes)), dim=0)
+
+        else:
+            overlaps = self.iou_calculator(gt_bboxes, bboxes)
```
goodsong81: The size of `overlaps` tensor is same regardless of looping. Then, does it mean that some internal buffers used in `self.iou_calculater()` had been causing the CPU OOM issue?

Additionally,
```suggestion
        if assign_on_cpu and gt_bboxes.shape[0] > self.cpu_assign_thr:
            split_length = gt_bboxes.shape[0] // self.cpu_assign_thr + 1
            overlaps = []
            for i in range(split_length):
                gt_bboxes_split = gt_bboxes[i * self.cpu_assign_thr : (i + 1) * self.cpu_assign_thr]
                overlaps.append(self.iou_calculator(gt_bboxes_split, bboxes))
            overlaps = torch.concat(overlaps, dim=0)
        else:
            overlaps = self.iou_calculator(gt_bboxes, bboxes)
```

sungmanc (dismissed):
Thanks for the nice work. BTW, how about using our own docstring instead of the original one for `CustomMaxIoUAssigner`?

goodsong81 (dismissed):
Thank you for the great work. Minor comment and question from my side:

otx/algorithms/detection/adapters/mmdet/models/assigners/custom_max_iou_assigner.py
jaegukhyun: For your question, yes self.iou_calculator generates (9 x N x M + N + M) matrix, where N is number of gt and M is number of anchor, to return NxM output matrix. Therefore you can think this may prevent OOM from 9 x N x M + N + M matrix but can't prevent from very large N x M matrix

goodsong81 (approved):

sungmanc (approved):

PR #2247, opened 2023-06-15
Mask Evaluaiton Optimization

eugene123tw:
### Checklist

### License

```python
```


otx/algorithms/detection/adapters/mmdet/datasets/dataset.py
```diff
@@ -362,14 +342,17 @@ def __init__(
...
+            filter_empty_gt=filter_empty_gt if data_subset != Subset.TESTING else False,
+            sampling_ratio=sampling_ratio if data_subset != Subset.TESTING else 1.0,
+            include_full_img=include_full_img if data_subset != Subset.TESTING else True,
         )
         self.flag = np.zeros(len(self), dtype=np.uint8)
         self.pipeline = Compose(pipeline)
         self.test_mode = test_mode
         self.num_samples = len(self.dataset)  # number of original samples
+        annotation = [self.get_ann_info(i) for i in range(len(self))]
+        if data_subset == Subset.VALIDATION:
```
eunwoosh: It's just a queestion. I think `self.evaluator` is initialized only if `data_subset == Subset.VALIDATION` is `True` but don't other cases  something like `Subset.TESTING` need `self.evaluator`?

otx/algorithms/detection/adapters/mmdet/evaluation/evaluator.py
```diff
@@ -0,0 +1,393 @@
...
+
+    if len(gt_masks) == 0:
+        fp[...] = 1
+        return tp, fp, 0.0
+    if num_dets == 0:
+        return tp, fp, 0.0
+
+    ious = mask_iou(det, gt_masks)  # (M, N)
+
+    assert ious.shape[0] == num_dets, f"{ious.shape[0]} != {num_dets}"
```
eunwoosh, line 200: Please change assert to raise. We have decided to use raise rather than assert.
And Could you explain what is wrong in more detail?

eunwoosh: Could you explain what is wrong in more detail?

otx/algorithms/detection/adapters/mmdet/evaluation/evaluator.py
```diff
@@ -0,0 +1,393 @@
...
+            det_results (list(tuple)): detection results including bboxes and masks
+            class_id (int): class index
+
+        Returns:
+            cls_dets: per-class detection results including bboxes and decoded masks
+            cls_scores: class scores
+        """
+        cls_scores = [img_res[0][class_id][..., -1] for img_res in det_results]
+        cls_dets: List[Tuple] = []
+        for i, det in enumerate(det_results):
```
eunwoosh:
```suggestion
        for det in det_results:
```
I think i isn't used here

otx/algorithms/detection/adapters/mmdet/models/heads/custom_fcn_mask_head.py
```diff
@@ -0,0 +1,107 @@
...
+from mmdet.models.roi_heads.mask_heads.fcn_mask_head import FCNMaskHead
+
+from otx.algorithms.common.adapters.mmdeploy.utils import is_mmdeploy_enabled
+
+
+@HEADS.register_module()
+class CustomFCNMaskHead(FCNMaskHead):
+    """Custom FCN Mask Head for fast mask evaluation."""
+
+    def __init__(self, *args, **kwargs):
```
eunwoosh: If it just use parent's function, then it can be removed, I think.

otx/algorithms/detection/utils/utils.py
eunwoosh, line 1: Could you implement unit test for added function?

otx/algorithms/detection/utils/utils.py
```diff
@@ -118,3 +125,160 @@ def get_det_model_api_configuration(label_schema: LabelSchemaEntity, task_type:
...
+    scale_h = mask.shape[0] / (mask.shape[0] - 2.0)
+    scale_w = mask.shape[1] / (mask.shape[1] - 2.0)
+    extended_box = expand_box(box, scale_h=scale_h, scale_w=scale_w).astype(int)
+    w, h = np.maximum(extended_box[2:] - extended_box[:2] + 1, 1)
+    mask = cv2.resize(mask.astype(np.float32), (w, h))
+    mask = mask.astype(np.uint8)
+    return mask
+
+
+def create_detection_shapes(pred_results, width, height, confidence_threshold, use_ellipse_shapes, labels):
```
eunwoosh: Could you add type hint?

otx/algorithms/detection/adapters/mmdet/models/heads/custom_fcn_mask_head.py
eunwoosh: If you just use parent's function, then it can be removed, I think.

otx/algorithms/detection/adapters/mmdet/models/heads/custom_fcn_mask_head.py
```diff
@@ -0,0 +1,107 @@
...
+        super().__init__(*args, **kwargs)
+
+    def get_seg_masks(self, mask_pred, det_bboxes, det_labels, rcnn_test_cfg, ori_shape, scale_factor, rescale):
+        """Get segmentation masks from mask_pred and bboxes.
```
eunwoosh, line 18: I think this function is changed to make it faster, right?
How about to explain which one is changed briefly to make maintenance easier later?

jaegukhyun:
Generally LGTM. Could you reflect @eunwoosh's comments?

otx/algorithms/detection/adapters/mmdet/models/heads/custom_fcn_mask_head.py
eugene123tw: I added this to the method: 

> The original `FCNMaskHead.get_seg_masks` grid sampled 28 x 28 masks to the original image resolution.
> As a result, the resized masks occupy a large amount of memory and slow down the inference.
> This method directly returns 28 x 28 masks and resizes to bounding boxes size in the post-processing steps.
> Doing so can save memory and speed up the inference.

otx/algorithms/detection/adapters/mmdet/evaluation/evaluator.py
eugene123tw: I added these additional checks while implementing. Removed now.

otx/algorithms/detection/adapters/mmdet/datasets/dataset.py
eugene123tw: The evaluation function had some issues when given empty ground truth annotation. I removed this condition and changed a bit in the evaluator so now it takes empty gt without outputting errors

otx/algorithms/detection/utils/utils.py
eugene123tw: Added to `tests/unit/algorithms/detection/adapters/mmdet/datasets/test_detection_dataset.py`

eunwoosh (approved):
LGTM. Thanks for your effort :)

sungmanc (approved):

PR #2268, opened 2023-06-24
Introducing adaptive LR schedule to Classification task

sungmanc:
### Summary

* Introduce the `patch_from_hparams` function to the Classification task. 
* Edit the doc-string to some functions
* Remove adaptive patience 

With this PR, we could improve the previous OTX's performance and convert the fixed LR schedule to an adaptive one that is more appropriate for an early-stopping scenario.

image

### Checklist

### License

```python
```


src/otx/algorithms/classification/configs/base/data/data_pipeline.py
```diff
@@ -20,19 +20,17 @@
 __resize_target_size = 224
 
 __train_pipeline = [
-    dict(type="Resize", size=__resize_target_size),
+    dict(type="RandomResizedCrop", size=224, efficientnet_style=True, interpolation="bicubic"),
```
goodsong81: Isn't "bicubic" too slow compared to "bilinear"?

goodsong81 (dismissed):
Thank you for the enhancement. I have just one minor question.

src/otx/algorithms/classification/configs/base/data/data_pipeline.py
sovrasov: By default linear interpolation is used everywhere (pytoprch, openvino). Why do we need to use bicubic or even bilinear? Will it cause an accuracy drop if we evaluate model with linear resize?

sungmanc: No special reason for this, I just followed the default MMCLS setting. Do you think we should change this one?
https://github.com/open-mmlab/mmpretrain/blob/748ab7aa7d76dc35939e09a59632a8c3f073c8bc/configs/efficientnet/efficientnet-b0_8xb32_in1k.py#L18

sovrasov: Their test pipeline also contains bicubic interpolation, and I think that combination provides the best accuracy, but that type of resize is slower than linear. Therefore I have the following thoughts:
- now we have the following combination: bicubic resize on train, linear resized on test & OV inference. From my experience, changing the resize type in inference results in significant drop (2-3% of top1), so the current test pipeline doesn't reveal the true performance of the trained models.
- probably, bicubic provides better top1 on low-res data or even in general, so mmpretrain use it to obtain SOTA-level pretrained models.
- currently training in OTX is slower than it could be, because of bicubic resize, however, on inference we probably don't see any benefits of that.

sungmanc: I changed to the default interpolation:), https://github.com/openvinotoolkit/training_extensions/pull/2268/commits/e15a7adf8c44bbaeb3e66bbe9e49acd2a263afaf

eunwoosh (approved):

sungmanc:
FYI, this change could affects the other tasks, however, we decided to keep this PR AS-IS.
And if there is a degradation issue at other tasks, I'll add the condition to enable this change for only classification.

goodsong81 (approved):

PR #2273, opened 2023-06-28
Remove randomness in DINO unit tests

jaegukhyun:
### Checklist

### License

```python
```

### DIFF
--- a/tests/unit/algorithms/detection/adapters/mmdet/models/heads/test_custom_dino_head.py
+++ b/tests/unit/algorithms/detection/adapters/mmdet/models/heads/test_custom_dino_head.py
@@ -18,6 +18,7 @@ from tests.test_suite.e2e_test_system import e2e_pytest_unit
 class TestCustomDINOHead:
     @pytest.fixture(autouse=True)
     def setup(self):
+        torch.manual_seed(555)
         cfg = ConfigDict(
             dict(
                 type="CustomDINOHead",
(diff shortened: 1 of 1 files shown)


jaegukhyun:
I tested 3 times using our CI, and their results are same.

sungmanc:
How about applying this manual seed logic to all other tests? to remove further uncertainty

jaegukhyun:
I'm doubting to control overall test randomness in one file. I believe test should control its own randomness.

sungmanc:
> I'm doubting to control overall test randomness in one file. I believe test should control its own randomness.

Hmm,, I don't mind the current way. it seems to be a little bit tricky.
If the user inputs another seed, then could the `len(losses)` be different? If so, then we need to consider randomness in our test case. I'm not sure that adding deterministic for the test to resolve the test error is the right direction. 

If the changed value of `len(losses)` is okay, then changing the `assert` line would be a great option.

sungmanc (approved):
Although I left a concern, I think that this is not important factor. So, want to review as approve.

jaegukhyun:
I understand your concern. Errors occur when some tensors have nan value during forward function. It is highly related initialized weights for bbox_head, so I set random seed to control it. The best way to solve this issue is mock inner forward function. However, the calculation which give NaN value changes randomly, therefore I had to set random seed to control those randomness.

sovrasov (approved):
I propose merging this to unblock other PRs, and make an issue for further discussion if required.

PR #2274, opened 2023-06-28
OTX export for visual prompting task

sungchul1:
### Summary

This PR includes:
- enable `otx export`
    - export dividing SAM into image encoder and decoder (prompt encoder + mask decoder)
- enable `otx eval` with OpenVINO IR
- refactoring & bug fix

I followed a flow of `OpenVINOSegmentationTask` for `OpenVINOVisualPromptingTask`, so they look so similar.
But, for visual prompting task, there are two models, image encoder and decoder.
I think this point is just difference between both tasks.

### TODOs
- [x] otx export
    - [x] ONNX
    image
    - [x] IR
- [x] OpenVINO inference
    - [x] sync inference (as default)
        ```shell
        $ otx eval
        2023-07-06 10:52:00,838 | INFO : Avg time per image: 4.619890505930892 secs
        2023-07-06 10:52:00,838 | INFO : Total time: 124.73704366013408 secs
        2023-07-06 10:52:00,838 | INFO : Visual Prompting OpenVINO inference completed
        2023-07-06 10:52:00,839 | INFO : Computing mDice
        2023-07-06 10:52:03,808 | INFO : mDice after evaluation: 0.9487857719406433
        ```
    - [ ] async inference
        - It showed performance degradation, need to check
- [x] Unit test
- [x] Integration test
- [x] rebase

#### Unit test (local)
```shell
$ tox -vv -e unittest-all-py310 -- tests/unit/algorithms/visual_prompting
====================== 165 passed, 110 warnings in 33.36s ======================
```

#### Integration test (local)
```shell
$ tox -vv -e tests-all-py310 -- tests/integration/cli/visual_prompting/test_visual_prompting.py
=================== 8 passed, 1 warning in 122.65s (0:02:02) ===================
```

### Checklist

### License

```python
```


src/otx/algorithms/visual_prompting/configs/base/configuration_enums.py
```diff
@@ -0,0 +1,29 @@
...
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions
+# and limitations under the License.
+
+import logging
+
+from otx.api.configuration import ConfigurableEnum
+
+logger = logging.getLogger(__name__)
```
eunwoosh: I think logger isn't used in this file.

src/otx/algorithms/visual_prompting/configs/sam_vit_b/config.yaml
```diff
@@ -17,7 +26,12 @@ model:
   freeze_image_encoder: true
   freeze_prompt_encoder: true
   freeze_mask_decoder: false
-  checkpoint: https://dl.fbaipublicfiles.com/segment_anything/sam_vit_b_01ec64.pth
+  checkpoint: pretrained
```
eunwoosh, line 29: It's just a question. if it is set as `pretrained` then proper weight is loaded automatically?

src/otx/algorithms/visual_prompting/tasks/inference.py
```diff
@@ -111,33 +119,29 @@ def get_config(self) -> Union[DictConfig, ListConfig]:
...
-                logger.info("[*] Pytorch checkpoint cannot be used for resuming. It will be supported.")
-                resume_from_checkpoint = None
-            elif not self.task_environment.model.model_adapters.get("resume", False):
+        model_checkpoint: Optional[str] = None
+        resume_from_checkpoint: Optional[str] = None
+        if self.mode == "train" and self.task_environment.model is not None:
+            # when args.load_weights or args.resume_from is set
+            resume_from_checkpoint = model_checkpoint = self.task_environment.model.model_adapters.get("path", None)  # type: ignore  # noqa: E501
+            if self.task_environment.model.model_adapters.get("resume", False):
+                if resume_from_checkpoint.endswith(".pth"):  # type: ignore
```
eunwoosh, line 128: It's just a question. In other mmXX tasks, weight saved in the middle of training also can be used for resume. Isn't pytorch lightning model possible?

src/otx/algorithms/visual_prompting/adapters/pytorch_lightning/models/visual_prompters/segment_anything.py
```diff
@@ -159,8 +159,14 @@ def replace_state_dict_keys(state_dict, revise_keys):
             # state_dict from args.load_from
             state_dict = replace_state_dict_keys(state_dict, revise_keys)
             self.load_state_dict(state_dict)
+        elif self.config.model.checkpoint == "pretrained":
```
eunwoosh, line 162: How about to write a checkpoint path  in each model config file statically. I think it's better than patching.

src/otx/algorithms/visual_prompting/tasks/inference.py
sungchul1: It's possible. Lightning saves checkpoint in the middle of training and we can use these checkpoints to resume training.
A different thing from mmx is "just using path". Lightning trainer loads checkpoint and meta and enables to resume training.
There are too many considerations to manually load & apply checkpoint and meta.

src/otx/algorithms/visual_prompting/configs/base/configuration_enums.py
sungchul1: Done.

src/otx/algorithms/visual_prompting/adapters/pytorch_lightning/models/visual_prompters/segment_anything.py
sungchul1: I just worried some problems when applying configs in `inference` and loading checkpoint in here.
But, I think it doesn't matter but changing them seems to take time.
If it is not critical, can I change them in other PR (https://github.com/openvinotoolkit/training_extensions/pull/2311)?

eunwoosh: yes, then let's change it in next PR. It would be better to reduce dynamic patching, I think.

src/otx/algorithms/visual_prompting/tasks/inference.py
eunwoosh: I understand. As you said, it'll be supported soon right?

sungchul1: Resuming w/ lightning checkpoint is already available.
https://github.com/openvinotoolkit/training_extensions/blob/ee7610e592f59d83ad173d8d16820df7e29610aa/src/otx/cli/tools/train.py#L219-L225
Get path and set `resume=True` like the above, then path is set into `resume_from_checkpoint` right there and is set into `trainer` through `resume_from_checkpoint`.

eunwoosh: Ah yes but it doesn't resume with `weights.pth` because file with "pth" suffix doesn't support resume. Then resume with "weights.pth" will be supported. Do I correctly understand it?

sungchul1: Ah, sorry for my misunderstanding.
That will be supported, and metadata and suffix might be changed just for pytorch lightning.

eunwoosh (approved):

src/otx/algorithms/visual_prompting/configs/base/configuration.py
```diff
@@ -20,15 +20,47 @@
...
+        blur_strength = configurable_integer(
+            header="Blur strength",
+            description="With a higher value, the segmentation output will be smoother, but less accurate.",
+            default_value=1,
+            min_value=1,
+            max_value=25,
+            affects_outcome_of=ModelLifecycle.INFERENCE,
+        )
+        soft_threshold = configurable_float(
+            default_value=0.5,
```
jaegukhyun, line 56: Does this variable change depending on trained model?

src/otx/algorithms/visual_prompting/tasks/inference.py
```diff
@@ -86,9 +90,13 @@ def __init__(self, task_environment: TaskEnvironment, output_path: Optional[str]
         # Hyperparameters.
         self._work_dir_is_temp = False
         self.output_path = output_path
+        self.mode = "train"
+        if task_environment.model is not None and task_environment.model.train_dataset is None:
+            self.mode = "export"
         if self.output_path is None:
             self.output_path = tempfile.mkdtemp(prefix="otx-visual_prompting")
             self._work_dir_is_temp = True
+            self.mode = "inference"
```
jaegukhyun, line 99: I think this kind of branch looks risky. Can this task get mode from cli? and why inference task care train mode?

src/otx/algorithms/visual_prompting/tasks/inference.py
```diff
@@ -258,7 +310,76 @@ def export(  # noqa: D102
...
+                    "--model_name",
+                    module,
+                ]
+                if precision == ModelPrecision.FP16:
+                    optimize_command.append("--compress_to_fp16")
+                subprocess.run(optimize_command, check=True)
+                with open(path.replace(".onnx", ".bin"), "rb") as file:
+                    output_model.set_data(f"{module}.bin", file.read())
+                with open(path.replace(".onnx", ".xml"), "rb") as file:
+                    output_model.set_data(f"{module}.xml", file.read())
```
jaegukhyun, line 376: I think if we set name as openvino.xml and openvino.bin, we can test without modifying test. Correct me if I'm wrong

jaegukhyun (commented):
I left some comments. Frankly there are many codes that is not fully understood. 
I think we should revisit this after FF. Before merging this PR I want to ensure there's no big accuracy drop before and after export. Do you have some data for this? I want to make sure of this since preprocessing and postprocessing of OV model is implemented in this PR, not from ModelAPI.

wonjuleee (approved):

src/otx/algorithms/visual_prompting/tasks/inference.py
sungchul1: You're right. I think we can use `mode` from `train.py` like below.
```python
task = task_class(task_environment=environment, output_path=str(config_manager.output_path / "logs"), mode=config_manager.mode)
```
But this change will affect any other tasks so I'm not sure if it is okay to handle this part.

jaegukhyun: I think directly set mode from cli is more safe, but leave it as it is.

src/otx/algorithms/visual_prompting/configs/base/configuration.py
sungchul1: Original trained model uses `threshold=0` without sigmoid.
I apply sigmoid to `soft_prediction` because of `hard_prediction`, so `soft_threshold=0.5` is the same.

jaegukhyun: I think this value is a little bit high for some dataset, especially small dataset. However if you think this value is okay then please resolve this conversation.

src/otx/algorithms/visual_prompting/tasks/inference.py
sungchul1: You're right. But visual prompting model should be divided into image encoder and decoder (prompt encoder + mask decoder) because they run independently.
Image encoder runs just once per image, but decoder runs the number of prompts times and this number of times is various.
If you can recommend any idea to combine these two models, I'll update it in other PR.

jaegukhyun: Ah, I didn't know about that. It will be better if we can combine them together, but it seems hard job to do.

sungchul1:
> I left some comments. Frankly there are many codes that is not fully understood. I think we should revisit this after FF. Before merging this PR I want to ensure there's no big accuracy drop before and after export. Do you have some data for this? I want to make sure of this since preprocessing and postprocessing of OV model is implemented in this PR, not from ModelAPI.

From this version, exported model showed slight mDice performance drop from 0.9503 to 0.9488.
As you said, it may be due to not using modules from modelapi.
I'm proceeding to change preprocessing and postprocessing to using modelapi in the next PR (https://github.com/openvinotoolkit/training_extensions/pull/2311).

jaegukhyun (approved):

src/otx/algorithms/visual_prompting/configs/base/configuration.py
sungchul1: Thanks for your comment :)

PR #2283, opened 2023-06-30
Fix incorrect matching of device in semisl_head

harimkang:
### Summary

Find some strange code. -> CPU training is not working in Semi-SL
Remove cuda() things in semisl_cls_head -> Now Semi-SL with CPU is working

### Checklist

### License

```python
```


sungmanc (commented):
Is this case occurs when users have GPUs but don't want to use?

harimkang:
> Is this case occurs when users have GPUs but don't want to use?

Exactly.The reason we didn't find out for a long time is because there is no cpu test. :)

sungmanc (approved):
Sounds great, we should need that for the future dev.

jaegukhyun (commented):
Is this for cpu training while machine contain gpu? So the root cause is training is run by cpu but torch.cuda.is_available() returns True.

harimkang:
> Is this for cpu training while machine contain gpu? So the root cause is training is run by cpu but torch.cuda.is_available() returns True.

Yes, that's right.

eunwoosh (approved):

jaegukhyun (approved):

PR #2286, opened 2023-06-30
Updated to use pytest for all testing execpt unittest

yunchu:
### Summary

* updated to use `pytest` instead of `coverage` for all testing execpt unittest.
  because `coverage` is slower than `pytest` due to the additional context to trace coverage
* added a step into the workflow to upload coverage.xml as artifact
* added separate coverage upload workflow for testing `workflow_run` completed event
* removed `use_develop = true` configuration to the tox testenv to test with package installation
* changed project structure to 'src-layout'

### Checklist

### License

```python
```


setup.py
```diff
@@ -195,8 +198,11 @@ def find_yaml_recipes():
         "Programming Language :: Cython",
     ],
     license="Apache License 2.0",
-    packages=find_packages(exclude=("tests",)),
+    # packages=find_packages(exclude=["tests"]),
```
eunwoosh: How about to delete this line?

sungmanc (dismissed):

tox.ini
```diff
@@ -58,10 +59,18 @@ commands =
...
-use_develop = true
+commands =
+    python -m pytest -ra --showlocals --csv={toxworkdir}/{envname}.csv {posargs:tests/integration/{[testenv]test_dir}}
+
+
+[testenv:unittest-all-{py38,py39,py310}]
+deps =
+    {[testenv]deps}
+    -r{toxinidir}/requirements/dev.txt
+extras = full
```
eunwoosh, line 70: It's just a question. `extras` added below has `all` value and here is `full`. Doesn't it matter?

eunwoosh (commented):
Thanks for your effort. It's quite huge change. I left some comments. Please take a look.

tox.ini
yunchu: > It's just a question. `extras` added below has `all` value and here is `full`. Doesn't it matter?

oops, thank you very much for your finding! the 'full' is correct one for the `extras` setting. it will be used to install the otx using pip like `pip install otx[<extras>]`

setup.py
yunchu: will remove it. thanks.

src/otx/algorithms/detection/configs/detection/resnet50_dino/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""Initialization of Deformable DETR for OTX Detection."""
```
jaegukhyun:
```suggestion
"""Initialization of DINO for OTX Detection."""
```

eunwoosh (approved):

sungmanc (approved):

PR #2305, opened 2023-07-05
Update openvino export

ashwinvaidya17:
### Summary
- This PR adds the metadata to the model IR so that the exported models can be used from modelAPI.
- The model in OTX still generates metadata.json but is not required by modelAPI. Other parts of the code are not touched so as to keep the tests intact and not introduce the breaking change.
- A PR will follow to modelAPI that uses the `.xml` and `.bin` files to generate result.

### New IR XML
image

### modelAPI
WIP modelAPI results without post-processing
image


src/otx/algorithms/anomaly/tasks/inference.py
```diff
@@ -302,6 +303,9 @@ def export(
             subprocess.run(optimize_command, check=True)
             bin_file = glob(os.path.join(self.config.project.path, "*.bin"))[0]
             xml_file = glob(os.path.join(self.config.project.path, "*.xml"))[0]
+
+            self._convert_to_new_format(xml_file)
```
samet-akcay: Even though this is a temporary solution, I'm not sure about the naming as `new` could be subjective and become old quickly. 

We could perhaps have the following alternative namings:

1. `self.convert_to_v*_format`
2. `self.add_metadata_to_model_ir`
3. `self.add_metadata_to_ir`
4. `self.append_metadata_to_model_ir`
5. `self.append_metadata_to_ir`

samet-akcay (changes requested):
Thanks for the quick fix @ashwinvaidya17. I only have a single comment in regards to the name of the private method. The rest looks good.

src/otx/algorithms/anomaly/tasks/inference.py
```diff
@@ -313,6 +317,32 @@ def export(
...
+                continue
+            model.set_rt_info(value, ["model_info", key])
+        # Add transforms
+        if "transform" in metadata:
+            for transform_dict in metadata["transform"]["transform"]["transforms"]:
+                transform = transform_dict.pop("__class_fullname__")
+                for key, value in transform_dict.items():
+                    model.set_rt_info(value, ["model_info", "transforms", transform, key])
+        model.set_rt_info("AnomalyDetection", ["model_info", "model_type"])
+        serialize(model, xml_file)
```
sovrasov: In OV 2023 we can't read and then serialize to the same file: possible workaround is here https://github.com/sovrasov/openvino_training_extensions/blob/vs/upd_ov/src/otx/algorithms/common/utils/ir.py#L36

samet-akcay (approved):
Thanks

sovrasov (approved):

ashwinvaidya17:
The checks have passed. Can someone merge this? @openvinotoolkit/ote-maintainers

PR #2315, opened 2023-07-06
Resolve label deletion issue

sungmanc:
### Summary

If user add/delete the label, the circumstances below could be happen.

* Multi-class --> Multi-class
* Multi-class --> Multi-label
* Multi-class --> H-label
* Multi-label --> Multi-class
* Multi-label --> Multi-label
* Multi-label --> H-label
* H-label --> Multi-class
* H-label --> Multi-label
* H-label --> H-label

To consider the above cases, I edited the test datasets and added integration tests to check the class decremental case (label deletion). It is not perfect and I don't want to change many things since it is a quick workaround. 

image

Need more precise test cases to consider the label deletion, for further use.

### Checklist

- [X] I have added integration tests to cover my changes.​

### License

```python
```


sungmanc:
http://validationreports.sclab.intel.com:8004/reports/build_number_report?test_session_build_number=IMPTOps-66139&environment=mclx33

otx/algorithms/classification/task.py
```diff
@@ -127,16 +129,22 @@ def __init__(self, task_environment: TaskEnvironment, output_path: Optional[str]
         if self._task_environment.model is not None:
             self._load_model()
 
+    def _is_multi_label(self, label_groups: List[LabelGroup], all_labels: List[LabelEntity]):
+        """Check whether the current training mode is multi-label or not."""
+        # NOTE: In the current Geti, multi-label should have `___` symbol for all group names.
+        find_multilabel_symbol = ["___" in getattr(i, "name", "") for i in label_groups]
```
eunwoosh, line 135: It's just a question. Does it consider both Geti case and OTX case? Will `find_multilabel_symbol` get a list whose all elements are True in OTX alone case?

otx/algorithms/classification/task.py
```diff
@@ -127,16 +129,22 @@ def __init__(self, task_environment: TaskEnvironment, output_path: Optional[str]
...
-        )  # noqa:E127
+        label_groups = self._task_environment.label_schema.get_groups(include_empty=False)
+        all_labels = self._task_environment.label_schema.get_labels(include_empty=False)
+
+        self._multilabel = self._is_multi_label(label_groups, all_labels)
         if self._multilabel:
             logger.info("Classification mode: multilabel")
-
-        if not self._multilabel and len(self._task_environment.label_schema.get_groups(False)) > 1:
+        if not self._multilabel and len(label_groups) > 1:
```
eunwoosh:
```suggestion
        elif len(label_groups) > 1:
```
It's a minor one but how about to change like this?

tests/integration/cli/classification/test_classification.py
```diff
@@ -296,6 +296,18 @@ def test_otx_train(self, template, tmp_dir_path):
         tmp_dir_path = tmp_dir_path / "multi_label_cls"
         otx_train_testing(template, tmp_dir_path, otx_dir, args_m)
 
+    @e2e_pytest_component
+    @pytest.mark.parametrize("template", templates, ids=templates_ids)
```
eunwoosh, line 300: How about to run this test only on default template to reduce test time?

tests/integration/cli/classification/test_classification.py
```diff
@@ -411,6 +423,18 @@ def test_otx_train(self, template, tmp_dir_path):
         tmp_dir_path = tmp_dir_path / "h_label_cls"
         otx_train_testing(template, tmp_dir_path, otx_dir, args_h)
 
+    @e2e_pytest_component
+    @pytest.mark.parametrize("template", templates, ids=templates_ids)
```
eunwoosh, line 427: Same as above

eunwoosh (commented):
Thanks for your work! Doesn't this change also have to apply to other versions?

otx/algorithms/classification/task.py
sungmanc: Nice point. In the current OTX, we only consider the exported Geti dataset for multi/h label classification. So, Geti = OTX for multi-label and h-label cases.

tests/integration/cli/classification/test_classification.py
sungmanc: In the state_dict_mixing algorithm, the logic could be different to model architectures. So I think checking all architectures could be safe. https://github.com/sungmanc/training_extensions/blob/312d51ed6de8182a9f3449ea65721bacc353ca9b/otx/algorithms/classification/adapters/mmcls/models/classifiers/sam_classifier.py#L244

goodsong81 (approved):

eunwoosh (approved):

PR #2320, opened 2023-07-07
Added publish workflow

yunchu:
### Summary

Added publish workflow and it will publish package to pypi when a release created.


yunchu:
successfully built package here https://github.com/openvinotoolkit/training_extensions/actions/runs/5484678149
only uploading TestPyPI has failed due the the invalid API token

PR #2340, opened 2023-07-11
added fix-cpu numbers option

yunchu:
### Summary

add an option to specify number of CPUs for the actions-runner CI container


eunwoosh (approved):

sungmanc (commented):
Is this option required to avoid the CPU bottleneck..?

yunchu:
> Is this option required to avoid the CPU bottleneck..?

I thing so. actually, I hope this option to use ci instance for the regression testing. by fixing number of CPUs for each container, I think we can get the constant performance from every runs of regression testing.

sungmanc (approved):

PR #2364, opened 2023-07-17
Enable multi-gpus for weekly

yunchu:
### Summary

Enable multi-gpus for weekly


.github/workflows/weekly.yml
```diff
@@ -12,31 +12,43 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - task: "iseg"
-            test_dir: "tests/regression/instance_segmentation"
+          - toxenv_task: "iseg"
```
eunwoosh, line 15: It's just a question. Is there a reason why ing-seg tilling and int-seg is separated unlike other tasks??

eunwoosh (dismissed):
LGTM. I have one question. please take a look.

.github/workflows/weekly.yml
yunchu: @eunwoosh thanks for your question. actually iseg tests took too much time compared with other tasks so I broken down them into small pieces so they can be ran parallel if there is any idle CI resources for reducing total execution time of the regression tests.

chuneuny-emily (approved):

sungmanc (approved):

PR #2369, opened 2023-07-18
Align label order between Geti and OTX

eunwoosh:
### Summary
When model is deployed from Geti and then run it on local, order of label is wrong.
It's because Geti makes label using hexadecimal and then sorts it by str order as below
image
But OTX sorts it by integer in string. What 'sorts it by integer in string.' means is that label is split by integer and integer format string is converted to real integer and then all split partt are wrapped by list. And that list is used as key for sorting. For example, if label id is "123abc123" then key is [123, "abc", 123].
This difference makes misalignment between Geti and OTX, so I changed function making sorting key to return original id if id isn't decimal format.
Some latent problem from change may exists, so please let me know if you have suspicious part.

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/api/entities/label_schema.py
+++ b/src/otx/api/entities/label_schema.py
@@ -41,9 +41,9 @@ def natural_sort_label_id(target: Union[ID, LabelEntity, ScoredLabel]) -> List:
 
     if isinstance(target, (LabelEntity, ScoredLabel)):
         target = target.id_
-    if isinstance(target, int):
-        return [target]
-    return [int(t) if t.isdigit() else t for t in re.split(r"(\d+)", target)]
+    if target.isdecimal():
+        return [int(target)]
+    return [target]
 
 
 class LabelGroupExistsException(ValueError):
--- a/src/otx/api/entities/label_schema.py
+++ b/src/otx/api/entities/label_schema.py
@@ -41,8 +41,8 @@ def natural_sort_label_id(target: Union[ID, LabelEntity, ScoredLabel]) -> List:
 
     if isinstance(target, (LabelEntity, ScoredLabel)):
         target = target.id_
-    if target.isdecimal():
-        return [int(target)]
+    if isinstance(target, str) and target.isdecimal():
+        return ["", int(target)]  # "" is added for the case where id of some lables is None
     return [target]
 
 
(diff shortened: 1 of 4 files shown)


goodsong81 (dismissed):
LGTM. Thank you for the quick fix!

src/otx/api/entities/label_schema.py
```diff
@@ -41,9 +41,9 @@ def natural_sort_label_id(target: Union[ID, LabelEntity, ScoredLabel]) -> List:
 
     if isinstance(target, (LabelEntity, ScoredLabel)):
         target = target.id_
-    if isinstance(target, int):
-        return [target]
-    return [int(t) if t.isdigit() else t for t in re.split(r"(\d+)", target)]
+    if isinstance(target, str) and target.isdecimal():
+        return ["", int(target)]  # "" is added for the case where id of some lables is None
```
goodsong81, line 45: Could you help me understand? If `target.id_` is `None`, the output will be `[None]`. Then, will it be compared against other cases like `["", 1234]` or `["a1234b1234"]`?

goodsong81 (dismissed):
Minor question below:

src/otx/api/entities/label_schema.py
sungmanc, line 45: Minor comment, don't we need to change the type hint? 
L#24: List --> List[Union[int, str]]
L#31: List[int] --> List[Union[int, str]]

eunwoosh: For the first question, if `target.id_` is None, then it's converted to `""`. Then it can make error when comparing `""` and `integer` because they have different type. 
For the second question, yes. Let's assume that we get two labels each of which has `1234` and `a1234b1234`. Then keys for sorting are `["", 1234]` and `["a1234b1234"]`. Because `["", 1234]` <  `["a1234b1234"]`, `1234` is prior to `a1234b1234`. But this case is not the case we consider as right case. I suppose that two cases can be given to us. First case, all labels have decimal format string or None (OTX) and second case all labels have string format or None (Geti).

sungmanc (dismissed):

src/otx/api/entities/label_schema.py
eunwoosh: Updated it.

goodsong81 (approved):

sungmanc (approved):

PR #2373, opened 2023-07-18
Fix NNCF training on CPU

sovrasov:
### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algorithms/common/adapters/mmcv/nncf/runners.py
+++ b/src/otx/algorithms/common/adapters/mmcv/nncf/runners.py
@@ -111,7 +111,11 @@ class AccuracyAwareRunner(EpochRunnerWithCancel):  # pylint: disable=too-many-in
             self._eval_hook._save_ckpt(self, nncf_runner.best_val_metric_value)
             return self._eval_hook.best_ckpt_path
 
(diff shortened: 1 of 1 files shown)


src/otx/algorithms/common/adapters/mmcv/nncf/runners.py
```diff
@@ -111,7 +111,11 @@ def dump_checkpoint_fn(model, compression_ctrl, nncf_runner, save_dir):
             self._eval_hook._save_ckpt(self, nncf_runner.best_val_metric_value)
             return self._eval_hook.best_ckpt_path
 
-        uncompressed_model_accuracy = self.model.module.nncf._uncompressed_model_accuracy
+        if hasattr(self.model, "module"):
```
sungmanc, line 114: It's interesting, if we use only CPU then `self.model` doesn't contain the `module`?

sungmanc (approved):

src/otx/algorithms/common/adapters/mmcv/nncf/runners.py
sovrasov: When we use GPU, we actually use a wrapped model. The original one is accessible via .module attribute

eunwoosh (approved):

PR #2400, opened 2023-07-25
Update Anomaly OpenVINO export for ModelAPI

ashwinvaidya17:
### Summary

Updates keys in `rt_info` for modelAPI. Since min and max values are used for normalization, both are fused into one. This also reverses the input channles so that RGB image is passed to the model during inference by modelAPI


src/otx/algorithms/anomaly/tasks/inference.py
```diff
@@ -333,21 +333,29 @@ def _add_metadata_to_ir(self, xml_file: str) -> None:
...
+                    )
                 elif transform == "Resize":
                     model.set_rt_info(transform_dict["height"], ["model_info", "orig_height"])
                     model.set_rt_info(transform_dict["width"], ["model_info", "orig_width"])
                 else:
                     warn(f"Transform {transform} is not supported currently")
+        # Since we only need the diff of max and min, we fuse the min and max into one op
+        if "min" in metadata and "max" in metadata:
+            model.set_rt_info(metadata["max"] - metadata["min"], ["model_info", "normalization_scale"])
+        model.set_rt_info(True, ["model_info", "reverse_input_channels"])
```
sovrasov: I propose using `embed_ir_model_data` here to unify implementation with other tasks.

sovrasov (changes requested):

src/otx/algorithms/anomaly/tasks/inference.py
ashwinvaidya17: Thanks. Didn't know it existed. That's pretty neat!

ashwinvaidya17: Done :)

sovrasov (commented):
Could you also enable flag `check_ir_meta` in `otx_export_testing` for anomaly models? 
`otx_export_testing` checks that `labels` field exists, and you may add labels=[] to metadata, if we don't need them in case of anomaly.

goodsong81:
@ashwinvaidya17 I've refined the title according to the convention.

sovrasov (approved):

goodsong81 (approved):

PR #2420, opened 2023-08-04
Configurable input size

eunwoosh:
### Summary
This PR includes configurable input size implementation.
Currently, this feature supports train, eval and export with classification, detection, instance segmentation and semantic segmentation tasks.
In eval and export, input size is automatically aligned if model weight was trained with other input size.
For refactoring purpose, code to patch input shape when exporting is changed to use newly added class which can estimate input size from data pipeline.

### Implementation detail
When configuring input size, each input size related value in data pipeline is updated.
And `InputSizeManager` class is in charge of whole of this job.
When changing values, original values are scaled down or up based on base input size and new input size.
For example, base input size is 512 and new input size is 256 then all related values are divided by 2.
Therefore, `InputSizeManager` needs base input size before updating.
User can set it explicitly but if not given, it estimates that value from data pipeline.
When updating or estimating, only pipelines types below are considered now:

- resize
- pad
- crop
- mosiac
- randomaffine
- multiscaleflipaug
- MultiScaleFlipAug
- AutoAugment

If new data pipelines need to be considered, code should be updated.

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algorithms/classification/adapters/mmcls/task.py
+++ b/src/otx/algorithms/classification/adapters/mmcls/task.py
@@ -209,6 +209,7 @@ class MMClassificationTask(OTXClassificationTask):
             ir_options,
             data_classes,
             model_classes,
+            self._hyperparams.learning_parameters.input_size,
             options_for_patch_datasets=options_for_patch_datasets,
             options_for_patch_evaluation=options_for_patch_evaluation,
         )
--- a/src/otx/algorithms/classification/adapters/mmcls/task.py
+++ b/src/otx/algorithms/classification/adapters/mmcls/task.py
@@ -58,6 +58,7 @@ from otx.algorithms.common.adapters.mmcv.utils import (
 from otx.algorithms.common.adapters.mmcv.utils.config_utils import (
     MPAConfig,
     update_or_add_custom_hook,
+    InputSizeManager,
 )
 from otx.algorithms.common.adapters.torch.utils import convert_sync_batchnorm
 from otx.algorithms.common.configs.configuration_enums import BatchSizeAdaptType
@@ -675,23 +676,15 @@ class MMClassificationTask(OTXClassificationTask):
                 mo_options.flags = list(set(mo_options.flags))
 
             def patch_input_shape(deploy_cfg):
-                resize_cfg = get_configs_by_pairs(
-                    cfg.data.test.pipeline,
-                    dict(type="Resize"),
-                )
-                assert len(resize_cfg) == 1
-                resize_cfg = resize_cfg[0]
-                size = resize_cfg.size
-                if isinstance(size, int):
-                    size = (size, size)
+                input_size_manager = InputSizeManager(cfg.data)
+                size = input_size_manager.get_input_size_from_cfg("test")
                 assert all(isinstance(i, int) and i > 0 for i in size)
                 # default is static shape to prevent an unexpected error
                 # when converting to OpenVINO IR
                 deploy_cfg.backend_config.model_inputs = [ConfigDict(opt_shapes=ConfigDict(input=[1, 3, *size]))]
 
             patch_input_preprocessing(deploy_cfg)
-            if not deploy_cfg.backend_config.get("model_inputs", []):
-                patch_input_shape(deploy_cfg)
+            patch_input_shape(deploy_cfg)
 
         return deploy_cfg
 
--- a/src/otx/algorithms/classification/adapters/mmcls/task.py
+++ b/src/otx/algorithms/classification/adapters/mmcls/task.py
@@ -53,8 +53,8 @@ from otx.algorithms.common.adapters.mmcv.utils import (
 )
 from otx.algorithms.common.adapters.mmcv.utils import (
     build_dataloader as otx_build_dataloader,
+    build_dataset as otx_build_dataset,
 )
-from otx.algorithms.common.adapters.mmcv.utils import build_dataset as otx_build_dataset
 from otx.algorithms.common.adapters.mmcv.utils.config_utils import (
     MPAConfig,
     update_or_add_custom_hook,
--- a/src/otx/algorithms/classification/adapters/mmcls/task.py
+++ b/src/otx/algorithms/classification/adapters/mmcls/task.py
@@ -58,7 +58,6 @@ from otx.algorithms.common.adapters.mmcv.utils import (
 from otx.algorithms.common.adapters.mmcv.utils.config_utils import (
     MPAConfig,
     update_or_add_custom_hook,
-    InputSizeScaler,
 )
 from otx.algorithms.common.adapters.torch.utils import convert_sync_batchnorm
 from otx.algorithms.common.configs.configuration_enums import BatchSizeAdaptType
@@ -211,11 +210,6 @@ class MMClassificationTask(OTXClassificationTask):
             options_for_patch_evaluation=options_for_patch_evaluation,
         )
         self._config = cfg
-
-        if self._hyperparams.learning_parameters.input_size != 0:
-            logger.warning("Classification models don't support to change input size."
-                           "Original input size will be used.")
-
         return cfg
 
     def build_model(
@@ -564,7 +558,7 @@ class MMClassificationTask(OTXClassificationTask):
 
         return eval_predictions, saliency_maps
 
-    def _export_model(self, precision: ModelPrecision, export_format: ExportType, dump_features: bool, input_size: Optional[int] = None):
+    def _export_model(self, precision: ModelPrecision, export_format: ExportType, dump_features: bool):
         self._data_cfg = ConfigDict(
             data=ConfigDict(
                 train=ConfigDict(
@@ -584,7 +578,7 @@ class MMClassificationTask(OTXClassificationTask):
         self._precision[0] = precision
         assert len(self._precision) == 1
         export_options: Dict[str, Any] = {}
-        export_options["deploy_cfg"] = self._init_deploy_cfg(cfg, input_size)
+        export_options["deploy_cfg"] = self._init_deploy_cfg(cfg)
 
         export_options["precision"] = str(precision)
         export_options["type"] = str(export_format)
@@ -630,7 +624,7 @@ class MMClassificationTask(OTXClassificationTask):
 
         return results
 
-    def _init_deploy_cfg(self, cfg: Config, input_size: Optional[int] = None) -> Union[Config, None]:
+    def _init_deploy_cfg(self, cfg: Config) -> Union[Config, None]:
         base_dir = os.path.abspath(os.path.dirname(self._task_environment.model_template.model_template_path))
         deploy_cfg_path = os.path.join(base_dir, "deployment.py")
         deploy_cfg = None
@@ -680,17 +674,14 @@ class MMClassificationTask(OTXClassificationTask):
                 mo_options.flags.extend(options["flags"])
                 mo_options.flags = list(set(mo_options.flags))
 
-            def patch_input_shape(deploy_cfg, input_size=None):
-                if input_size is None:
-                    resize_cfg = get_configs_by_pairs(
-                        cfg.data.test.pipeline,
-                        dict(type="Resize"),
-                    )
-                    assert len(resize_cfg) == 1
-                    resize_cfg = resize_cfg[0]
-                    size = resize_cfg.size
-                else:
-                    size = input_size
+            def patch_input_shape(deploy_cfg):
+                resize_cfg = get_configs_by_pairs(
+                    cfg.data.test.pipeline,
+                    dict(type="Resize"),
+                )
+                assert len(resize_cfg) == 1
+                resize_cfg = resize_cfg[0]
+                size = resize_cfg.size
                 if isinstance(size, int):
                     size = (size, size)
                 assert all(isinstance(i, int) and i > 0 for i in size)
@@ -699,8 +690,8 @@ class MMClassificationTask(OTXClassificationTask):
                 deploy_cfg.backend_config.model_inputs = [ConfigDict(opt_shapes=ConfigDict(input=[1, 3, *size]))]
 
             patch_input_preprocessing(deploy_cfg)
-            if not deploy_cfg.backend_config.get("model_inputs", []) or input_size is not None:
-                patch_input_shape(deploy_cfg, input_size)
+            if not deploy_cfg.backend_config.get("model_inputs", []):
+                patch_input_shape(deploy_cfg)
 
         return deploy_cfg
 
--- a/src/otx/algorithms/classification/adapters/mmcls/task.py
+++ b/src/otx/algorithms/classification/adapters/mmcls/task.py
@@ -58,6 +58,7 @@ from otx.algorithms.common.adapters.mmcv.utils import (
 from otx.algorithms.common.adapters.mmcv.utils.config_utils import (
     MPAConfig,
     update_or_add_custom_hook,
+    InputSizeScaler,
 )
 from otx.algorithms.common.adapters.torch.utils import convert_sync_batchnorm
 from otx.algorithms.common.configs.configuration_enums import BatchSizeAdaptType
@@ -210,6 +211,11 @@ class MMClassificationTask(OTXClassificationTask):
             options_for_patch_evaluation=options_for_patch_evaluation,
         )
         self._config = cfg
+
+        if self._hyperparams.learning_parameters.input_size != 0:
+            logger.warning("Classification models don't support to change input size."
+                           "Original input size will'be used.")
+
         return cfg
 
     def build_model(
--- a/src/otx/algorithms/classification/adapters/mmcls/task.py
+++ b/src/otx/algorithms/classification/adapters/mmcls/task.py
@@ -558,7 +558,7 @@ class MMClassificationTask(OTXClassificationTask):
 
         return eval_predictions, saliency_maps
 
-    def _export_model(self, precision: ModelPrecision, export_format: ExportType, dump_features: bool):
+    def _export_model(self, precision: ModelPrecision, export_format: ExportType, dump_features: bool, input_size: Optional[int] = None):
         self._data_cfg = ConfigDict(
             data=ConfigDict(
                 train=ConfigDict(
@@ -578,7 +578,7 @@ class MMClassificationTask(OTXClassificationTask):
         self._precision[0] = precision
         assert len(self._precision) == 1
         export_options: Dict[str, Any] = {}
-        export_options["deploy_cfg"] = self._init_deploy_cfg(cfg)
+        export_options["deploy_cfg"] = self._init_deploy_cfg(cfg, input_size)
 
         export_options["precision"] = str(precision)
         export_options["type"] = str(export_format)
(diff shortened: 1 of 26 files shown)


src/otx/algorithms/classification/configs/configuration.yaml
```diff
@@ -275,6 +275,29 @@ learning_parameters:
...
+      _1024x1024: "1024x1024"
+    type: SELECTABLE
+    ui_rules:
+      action: DISABLE_EDITING
+      operator: AND
+      rules: []
+      type: UI_RULES
+    value: Default
+    visible_in_ui: false
+    warning: Small input size may decrease model performance.
```
goodsong81:
```suggestion
    warning: Small input size may decrease model accuracy at the price of better train and inference speed.
```

src/otx/algorithms/classification/configs/configuration.yaml
```diff
@@ -275,6 +275,29 @@ learning_parameters:
...
+    editable: true
+    enum_name: InputSizePreset
+    header: Change input image size.
+    options:
+      DEFAULT: "Default"
+      _256x256: "256x256"
+      _384x384: "384x384"
+      _512x512: "512x512"
+      _768x768: "768x768"
+      _1024x1024: "1024x1024"
```
goodsong81: If we are to enable config input param to classification task in at least OTX, what do you think of adding 64x64, 128x128?

src/otx/algorithms/classification/configs/configuration.yaml
```diff
@@ -275,6 +275,29 @@ learning_parameters:
       type: UI_RULES
     visible_in_ui: true
     warning: null
+  input_size:
+    affects_outcome_of: NONE
+    default_value: Default
+    description: You can reduce training and inference time by using small input size.
+    editable: true
+    enum_name: InputSizePreset
+    header: Change input image size.
```
goodsong81:
```suggestion
    header: Configure model input size
```

src/otx/algorithms/classification/configs/configuration.yaml
```diff
@@ -275,6 +275,29 @@ learning_parameters:
       type: UI_RULES
     visible_in_ui: true
     warning: null
+  input_size:
+    affects_outcome_of: NONE
+    default_value: Default
+    description: You can reduce training and inference time by using small input size.
```
goodsong81:
```suggestion
    description: The input size of the given model could be configured to one of the predefined resolutions. Reduced training and inference time could be expected by using smaller input size. Defaults to per-model default resolution.
```

src/otx/algorithms/common/adapters/mmcv/utils/config_utils.py
goodsong81, line 1: Why don't you use `subset` rather than `task` for items in ["test", "val", "train"]?

src/otx/algorithms/common/adapters/mmcv/utils/config_utils.py
```diff
@@ -698,3 +700,243 @@ def get_data_cfg(config: Union[Config, ConfigDict], subset: str = "train") -> Co
...
+            for idx in range(len(pipeline[attr])):
+                pipeline[attr][idx] = (
+                    round(pipeline[attr][idx][0] * scale[0]),
+                    round(pipeline[attr][idx][1] * scale[1]),
+                )
+        else:
+            pipeline[attr] = (round(pipeline[attr][0] * scale[0]), round(pipeline[attr][1] * scale[1]))
+
+
+def get_configurable_input_size(
```
goodsong81: How about `get_model_input_size_from_ckpt` or `get_trained_model_input_size`?

src/otx/algorithms/common/configs/configuration_enums.py
```diff
@@ -46,3 +46,14 @@ class BatchSizeAdaptType(ConfigurableEnum):
...
+
+class InputSizePreset(ConfigurableEnum):
+    """Configurable input size preset."""
+
+    DEFAULT = "Default"
+    _256x256 = "256x256"
+    _384x384 = "384x384"
+    _512x512 = "512x512"
+    _768x768 = "768x768"
+    _1024x1024 = "1024x1024"
```
goodsong81: Why don't you implement `__str__()` method to eliminate the under_bar from UX?

goodsong81: In addition, we could add more options in this preset enum as I noted in other comments :)

src/otx/algorithms/common/configs/training_base.py
```diff
@@ -198,6 +198,15 @@ class BaseLearningParameters(ParameterGroup):
             affects_outcome_of=ModelLifecycle.TRAINING,
         )
 
+        input_size = selectable(
+            default_value=InputSizePreset.DEFAULT,
+            header="Change input image size.",
+            description="You can reduce training and inference time by using small input size.",
+            warning="Small input size may decrease model performance.",
```
goodsong81: Just a question: should we specify these duplicated info here again? Can we just inherit somehow? BTW, please align the descriptions.

src/otx/algorithms/detection/configs/rotated_detection/configuration.yaml
goodsong81, line 1: Specific reason not to provide selections?

src/otx/algorithms/segmentation/adapters/mmseg/configurer.py
```diff
@@ -534,6 +539,25 @@ def _configure_dataloader(cfg: Config) -> None:
...
+        input_size = get_configurable_input_size(input_size_config, model_ckpt)
+        if input_size is None:
+            return
+
+        # segmentation models have different input size in train and val data pipeline
+        base_input_size = {
+            "train": 512,
+            "val": 544,
+            "test": 544,
+        }
```
goodsong81, line 556: Wow... same for SegNext, too?

goodsong81 (changes requested):
What a great work! I've suggested some changes. Thanks :)

src/otx/algorithms/common/adapters/mmcv/utils/config_utils.py
eunwoosh: Because it's just data configuration dictionary, I need to refer data pipeline in cfg.data.test, val or train.
If your question is about why not initializing dataset class and using it, I considered it but there are some of limitations from code flow or different format.

src/otx/algorithms/common/adapters/mmcv/utils/config_utils.py
eunwoosh: Actually, it doesn't refer not only model weight but also `input_size_config` user set. If `input_size_config` is set, then model weight is ignored. I make this function just to avoid code duplication in detection and segmentation task.

src/otx/algorithms/common/configs/configuration_enums.py
eunwoosh: As far as I know, `ConfigurableEnum`, parent class of this, already have `__str()__` function, so string value is shown in UI.
I may be wrong, so please let me know if it is :)

src/otx/algorithms/common/configs/training_base.py
eunwoosh: oh maybe it is already inherited, so it isn't necessary. I'll check it.

src/otx/algorithms/segmentation/adapters/mmseg/configurer.py
eunwoosh: yes, all segmentation models refer same data pipeline file.

src/otx/algorithms/classification/configs/configuration.yaml
eunwoosh: Yes I agree. I'll update it.

src/otx/algorithms/detection/configs/rotated_detection/configuration.yaml
eunwoosh: I missed it. I'll update it. Thanks for letting me know.

src/otx/algorithms/common/configs/training_base.py
eunwoosh: I tried, but it doesn't work. I'll align all sentences with updated one.

src/otx/algorithms/classification/configs/configuration.yaml
eunwoosh: I updated it.

src/otx/algorithms/classification/configs/configuration.yaml
eunwoosh: I updated it.

src/otx/algorithms/classification/configs/configuration.yaml
```diff
@@ -275,6 +275,34 @@ learning_parameters:
...
+      _1024x1024: "1024x1024"
+    type: SELECTABLE
+    ui_rules:
+      action: DISABLE_EDITING
+      operator: AND
+      rules: []
+      type: UI_RULES
+    value: Default
+    visible_in_ui: false
+    warning: Small input size may decrease model accuracy at the price of better train and inference speed.
```
sungmanc:
```suggestion
    warning: Modifying input size may decrease model performance
```

Some users might use `larger` resolution instead of `smaller`

src/otx/algorithms/classification/adapters/mmcls/task.py
```diff
@@ -675,23 +678,15 @@ def patch_input_preprocessing(deploy_cfg):
...
+                size = input_size_manager.get_input_size_from_cfg("test")
                 assert all(isinstance(i, int) and i > 0 for i in size)
                 # default is static shape to prevent an unexpected error
                 # when converting to OpenVINO IR
                 deploy_cfg.backend_config.model_inputs = [ConfigDict(opt_shapes=ConfigDict(input=[1, 3, *size]))]
 
             patch_input_preprocessing(deploy_cfg)
-            if not deploy_cfg.backend_config.get("model_inputs", []):
-                patch_input_shape(deploy_cfg)
+            patch_input_shape(deploy_cfg)
```
sungmanc, line 690: How about integrating `patch_input_shape` into the `patch_input_preprocessing`? I think `preprocessing` include the concept of the shape

src/otx/algorithms/common/adapters/mmcv/utils/config_utils.py
```diff
@@ -698,3 +700,243 @@ def get_data_cfg(config: Union[Config, ConfigDict], subset: str = "train") -> Co
...
+        base_input_size: Optional[Union[int, Tuple[int, int], Dict[str, int], Dict[str, Tuple[int, int]]]] = None,
+    ):
+        self._data_config = data_config
+        if isinstance(base_input_size, int):
+            base_input_size = (base_input_size, base_input_size)
+        elif isinstance(base_input_size, dict):
+            for task in base_input_size.keys():
+                if isinstance(base_input_size[task], int):
+                    base_input_size[task] = (base_input_size[task], base_input_size[task])  # type: ignore[assignment]
+            for data_type in ["train", "val", "test"]:
```
sungmanc: Making the variable `DATA_TYPE`=["train", "val", "test"] and reusing this variable looks better to me since ["train", "val", "test"] used repeatedly

src/otx/algorithms/common/adapters/mmcv/utils/config_utils.py
```diff
@@ -698,3 +700,243 @@ def get_data_cfg(config: Union[Config, ConfigDict], subset: str = "train") -> Co
...
+        for data_type in ["train", "val", "test"]:
+            if data_type in self._data_config:
+                if isinstance(self.base_input_size, dict):
+                    resize_ratio = (
+                        input_size[0] / self.base_input_size[data_type][0],
+                        input_size[1] / self.base_input_size[data_type][1],
+                    )
+                pipelines = self._get_pipelines(data_type)
+                for pipeline in pipelines:
+                    self._set_pipeline_size_vlaue(pipeline, resize_ratio)
```
sungmanc:
```suggestion
                    self._set_pipeline_size_value(pipeline, resize_ratio)
```

sungmanc (commented):
Thanks for the huge work, please check my comments.

src/otx/algorithms/common/adapters/mmcv/utils/config_utils.py
eunwoosh: Thanks for finding it. I'll fix it.

src/otx/algorithms/common/adapters/mmcv/utils/config_utils.py
eunwoosh: I see. I'll reflect it.

src/otx/algorithms/classification/adapters/mmcls/task.py
eunwoosh: I agree in some point but there is ambiguous part. `patch_input_preprocessing` changes only data pipeline but `patch_input_shape` changes input size in model part. In my thought, the function name `patch_input_preprocessing` is a little confusing and maybe something like `patch_data_pipeline` is more readable. Anyway changing this code flow seems out of range. How about changing it in other PR after some conversation?

src/otx/algorithms/classification/configs/configuration.yaml
eunwoosh: Right. I'll update it.

src/otx/algorithms/classification/adapters/mmcls/task.py
sungmanc: Sounds good :)

sungmanc (dismissed):

jaegukhyun (dismissed):
Codes are look good to me. Do you have a plan for updating docs for this new feature? I think explanation for usage during train, eval, export would be needed.

eunwoosh:
> Codes are look good to me. Do you have a plan for updating docs for this new feature? I think explanation for usage during train, eval, export would be needed.

yes, I'll make a PR for updating docs after merging this PR.

src/otx/algorithms/common/adapters/mmcv/utils/config_utils.py
goodsong81: I mean the terminology. Is "train" is a task or a subset?

src/otx/algorithms/classification/configs/configuration.yaml
goodsong81: Could you update other yaml files, too?

src/otx/algorithms/classification/configs/configuration.yaml
goodsong81: No need to accept this suggestion directly, but would you decide explicitly and apply to all other yaml files? (py files might need update, too.

src/otx/algorithms/common/adapters/mmcv/utils/config_utils.py
goodsong81: Then, what about `get_configured_input_size()`?

src/otx/algorithms/classification/configs/configuration.yaml
```diff
@@ -275,6 +275,34 @@ learning_parameters:
...
+      _1024x1024: "1024x1024"
+    type: SELECTABLE
+    ui_rules:
+      action: DISABLE_EDITING
+      operator: AND
+      rules: []
+      type: UI_RULES
+    value: Default
+    visible_in_ui: false
+    warning: Modifying input size may decrease model performance.
```
goodsong81, line 305: Please align this warning.

src/otx/algorithms/common/configs/training_base.py
```diff
@@ -198,6 +198,17 @@ class BaseLearningParameters(ParameterGroup):
             affects_outcome_of=ModelLifecycle.TRAINING,
         )
 
+        input_size = selectable(
+            default_value=InputSizePreset.DEFAULT,
+            header="Configure model input size.",
+            description="The input size of the given model could be configured to one of the predefined resolutions."
+            "Reduced training and inference time could be expected by using smaller input size."
+            "Defaults to per-model default resolution.",
+            warning="Modifying input size may decrease model performance.",
```
goodsong81, line 207: Same here :)

src/otx/algorithms/detection/configs/detection/configuration.yaml
```diff
@@ -221,6 +221,32 @@ learning_parameters:
...
+      _1024x1024: "1024x1024"
+    type: SELECTABLE
+    ui_rules:
+      action: DISABLE_EDITING
+      operator: AND
+      rules: []
+      type: UI_RULES
+    value: Default
+    visible_in_ui: false
+    warning: Modifying input size may decrease model performance.
```
goodsong81, line 249: Same here~

src/otx/algorithms/detection/configs/instance_segmentation/configuration.yaml
```diff
@@ -221,6 +221,32 @@ learning_parameters:
...
+      _1024x1024: "1024x1024"
+    type: SELECTABLE
+    ui_rules:
+      action: DISABLE_EDITING
+      operator: AND
+      rules: []
+      type: UI_RULES
+    value: Default
+    visible_in_ui: false
+    warning: Modifying input size may decrease model performance.
```
goodsong81, line 249: Same here~

src/otx/algorithms/detection/configs/rotated_detection/configuration.yaml
```diff
@@ -221,8 +221,32 @@ learning_parameters:
...
+      _1024x1024: "1024x1024"
+    type: SELECTABLE
+    ui_rules:
+      action: DISABLE_EDITING
+      operator: AND
+      rules: []
+      type: UI_RULES
+    value: Default
+    visible_in_ui: false
+    warning: Modifying input size may decrease model performance.
```
goodsong81, line 249: Same here~

src/otx/algorithms/segmentation/configs/configuration.yaml
```diff
@@ -230,6 +230,32 @@ learning_parameters:
...
+      _1024x1024: "1024x1024"
+    type: SELECTABLE
+    ui_rules:
+      action: DISABLE_EDITING
+      operator: AND
+      rules: []
+      type: UI_RULES
+    value: Default
+    visible_in_ui: false
+    warning: Modifying input size may decrease model performance.
```
goodsong81, line 258: Same here~

goodsong81 (commented):
I left some minor comments. Could you check them?

src/otx/algorithms/common/configs/training_base.py
eunwoosh: Actually I changed it as Sungman said after applying your feedback. He said that user not just decrease input size but also increase and I agree on that. How do you think about that?

src/otx/algorithms/common/adapters/mmcv/utils/config_utils.py
eunwoosh: ah I understand. I'll change it.

src/otx/algorithms/common/adapters/mmcv/utils/config_utils.py
eunwoosh: I'll change it.

src/otx/algorithms/classification/configs/configuration.yaml
eunwoosh: I'm sorry but I don't know which file needs to be changed. I already changed "training_base.py". Are there any other files I need to change?

goodsong81: I get it, thanks for the update!

goodsong81 (approved):
LGTM! Thank you for the great feature.

jaegukhyun (approved):

Issue #2427, opened 2023-08-09
Weired score reporting at Classification


sungmanc:
**Describe the bug**

At the final validation phase during training, the represented score is incorrect. 
![image](https://github.com/openvinotoolkit/training_extensions/assets/80735344/01f8f064-a070-4062-b923-1c2d253191b7)

sungmanc:
Resolved at PR#2428

PR #2428, opened 2023-08-09
Introduce AdaptiveRepeatDataHook for the classification

sungmanc:
### Summary

This PR introduces the `AdaptiveRepeatDataHook` and `OTXSampler` for the classification now, later I will consider it for other tasks. It seems that it's not huge work that applies to other tasks. 

By changing the number of iterations per epoch, we could significantly reduce the training time at the small data regime.
For the classification, we could check the improvements by checking the table below.

Number of training data: 16
Number of repeats: 5

mode | performance | gpu_mem | iter_time | data_time | e2e_time
-- | -- | -- | -- | -- | --
develop | 0.870 | 3081 | 0.380 | 0.312 | 0:00:17
sampler | 0.881 | 3084 | 0.215 (-43.3%) | 0.147 (-53.1%) | 0:00:13 (-25.4%)

When we considered to use the `AdaptiveRepeatDataHook`, `AdaptiveSchedulingHook` should be disabled. So, I changed the default value at the `configuration.yaml`

This commit(https://github.com/openvinotoolkit/training_extensions/pull/2428/commits/fb40feadb20b333ef1eed45afd1e502055936034) added to resolve the issue(https://github.com/openvinotoolkit/training_extensions/issues/2427)

For now, this sampler only affects to the multi-class classification.

### Checklist

- [X] I have updated the documentation in my target branch accordingly (e.g., [documentation](https://github.com/openvinotoolkit/training_extensions/tree/develop/docs) in develop).

### License

- [X] I submit _my code changes_ under the same [Apache License](https://github.com/openvinotoolkit/training_extensions/blob/develop/LICENSE) that covers the project.
- [X] I have updated the license header for each file (see an example below).

```python
```


src/otx/algorithms/classification/adapters/mmcls/configurer.py
```diff
@@ -648,6 +650,7 @@ def configure_task_adapt_hook(self, cfg, training):
             sampler_flag=sampler_flag,
             sampler_type=sampler_type,
             efficient_mode=efficient_mode,
+            use_adaptive_repeat=True if is_multiclass else False,
```
eunwoosh:
```suggestion
            use_adaptive_repeat=is_multiclass
```
It's a minor one, but how about changing like this?

src/otx/algorithms/classification/configs/deit_tiny/model.py
```diff
@@ -18,3 +18,5 @@
 
 optimizer = dict(_delete_=True, type="AdamW", lr=0.01, weight_decay=0.05)
 optimizer_config = dict(_delete_=True)
+
+custom_hooks = [dict(type="AdaptiveRepeatDataHook")]
```
eunwoosh: As far as I know, custom_hooks is already defined in 'recipes/stages/classification/incremental.yaml'.
Then, isn't it ok to override custom_hooks in that file?

src/otx/algorithms/classification/adapters/mmcls/configurer.py
```diff
@@ -639,6 +640,7 @@ def configure_task_adapt_hook(self, cfg, training):
             sampler_flag = False
         else:
             sampler_flag = True
+
         # Update Task Adapt Hook
         task_adapt_hook = ConfigDict(
             type="TaskAdaptHook",
```
eunwoosh, line 258: I have a question. In class incremental case, `TaskAdaptHook` change the sampler. Then, in that case, it seems that both `TaskAdaptHook` and `AdaptiveRepeatDataHook` try to change sampler. Am I right? If yes, is it ok?

src/otx/algorithms/common/adapters/torch/dataloaders/samplers/balanced_sampler.py
```diff
@@ -21,24 +27,40 @@ class BalancedSampler(Sampler):  # pylint: disable=too-many-instance-attributes
...
+            distributed training. By default, :attr:`world_size` is retrieved from the
+            current distributed group.
+        rank (int, optional): Rank of the current process within :attr:`num_replicas`.
+            By default, :attr:`rank` is retrieved from the current distributed
+            group.
+        drop_last (bool, optional): if ``True``, then the sampler will drop the
+            tail of the data to make it evenly divisible across the number of
+            replicas. If ``False``, the sampler will add extra indices to make
+            the data evenly divisible across the replicas. Default: ``False``.
+        use_adaptive_repeats (bool): Flag about using adaptive repeats
```
eunwoosh:
```suggestion
        use_adaptive_repeats (bool, optional): Flag about using adaptive repeats
```
I think it's missed.

src/otx/algorithms/common/adapters/torch/dataloaders/samplers/otx_sampler.py
```diff
@@ -0,0 +1,98 @@
...
+
+class OTXSampler(Sampler):  # pylint: disable=too-many-instance-attributes
+    """Sampler that easily adapts to the dataset statistics.
+
+    In the exterme small dataset, the iteration per epoch could be set to 1 and then it could make slow training
+    since DataLoader reinitialized at every epoch. So, in the small dataset case,
+    OTXSampler repeats the dataset to enlarge the iterations per epoch.
+
+    In the large dataset, the useful information is not totally linear relationship with the number of datasets.
+    It closes to the log scale relationship, rather.
```
eunwoosh:
```suggestion
    It is close to the log scale relationship, rather.
```

src/otx/algorithms/common/adapters/torch/dataloaders/samplers/otx_sampler.py
```diff
@@ -0,0 +1,98 @@
...
+
+    So, this sampler samples or repeats the datasets acoording to the statistics of dataset.
+
+    Args:
```
eunwoosh, line 29: It seems that argument comment is a little different with real arguments. Could you check it?

src/otx/algorithms/common/adapters/torch/dataloaders/samplers/otx_sampler.py
```diff
@@ -0,0 +1,98 @@
...
+    ):
+
+        self.dataset, _ = unwrap_dataset(dataset)
+        self.samples_per_gpu = samples_per_gpu
+        self.num_replicas = num_replicas
+        self.rank = rank
+        self.shuffle = shuffle
+        self.repeat = self._get_proper_repeats(use_adaptive_repeats, coef, min_repeat)
+
+        self.num_samples = int(math.ceil(len(self.dataset) * self.repeat / self.num_replicas))
```
eunwoosh:
```suggestion
        self.num_samples = math.ceil(len(self.dataset) * self.repeat / self.num_replicas)
```
It's a minor one, math.ceil returns int since python3

src/otx/algorithms/common/adapters/torch/dataloaders/samplers/cls_incr_sampler.py
```diff
@@ -25,16 +25,35 @@ class ClsIncrSampler(Sampler):  # pylint: disable=too-many-instance-attributes
...
+            distributed training. By default, :attr:`world_size` is retrieved from the
+            current distributed group.
+        rank (int, optional): Rank of the current process within :attr:`num_replicas`.
+            By default, :attr:`rank` is retrieved from the current distributed
+            group.
+        drop_last (bool, optional): if ``True``, then the sampler will drop the
+            tail of the data to make it evenly divisible across the number of
+            replicas. If ``False``, the sampler will add extra indices to make
+            the data evenly divisible across the replicas. Default: ``False``.
+        use_adaptive_repeats (bool): Flag about using adaptive repeats
```
eunwoosh:
```suggestion
        use_adaptive_repeats (bool, optional): Flag about using adaptive repeats
```

src/otx/algorithms/common/adapters/mmcv/hooks/adaptive_repeat_data_hook.py
```diff
@@ -0,0 +1,57 @@
...
+
+logger = get_logger()
+
+
+@HOOKS.register_module()
+class AdaptiveRepeatDataHook(Hook):
+    """Hook that adaptively repeats the dataset to control the number of iterations.
+
+    Args:
+        coef (float) : coefficient that effects to number of repeats
```
eunwoosh:
```suggestion
        coef (float, optional) : coefficient that effects to number of repeats
```

src/otx/algorithms/common/adapters/mmcv/hooks/adaptive_repeat_data_hook.py
```diff
@@ -0,0 +1,57 @@
...
+                       (coef * math.sqrt(num_iters-1)) +5
+        min_repeat (float) : minimum repeats
```
eunwoosh:
```suggestion
        min_repeat (float, optional) : minimum repeats
```

src/otx/algorithms/common/adapters/torch/dataloaders/samplers/otx_sampler.py
```diff
@@ -0,0 +1,98 @@
...
+from torch.utils.data import Dataset
+from torch.utils.data.sampler import Sampler
+
+from otx.algorithms.common.utils.logger import get_logger
+from otx.algorithms.common.utils.task_adapt import unwrap_dataset
+
+logger = get_logger()
+
+
+class OTXSampler(Sampler):  # pylint: disable=too-many-instance-attributes
```
eunwoosh, line 17: Could you add unit tests for this class?

eunwoosh, line 17: In my thought, `OTXSampler` is for adapting `repeat` values in sampler. So how about changing it's name to provide more detailed info? For example `RepeatSampler` can be one of candidates.

eunwoosh (commented):
Thanks for your work! It can reduce much overhead, I think. I left some comments, please take a look.

src/otx/algorithms/classification/configs/deit_tiny/model.py
sungmanc: Both of hooks still alive at the training phase. LazyEarlystop, AdaptiveRepeatData
image

src/otx/algorithms/classification/adapters/mmcls/configurer.py
sungmanc: It could be, it should be corrected. Thanks :)

src/otx/algorithms/common/adapters/torch/dataloaders/samplers/otx_sampler.py
sungmanc: I agree with you, this is the personal preference. Since we'll handle the 1 > repeats for the large dataset or we might add some sampling logic at the future, I want to represent this Sampler as a functionality of the OTX as like adaptive dataset control.

src/otx/algorithms/classification/configs/deit_tiny/model.py
eunwoosh: I checked that `LazyEarlystop` configuration is override but it's newly patched at `patch_early_stopping()`. It can be refined later but anyway it's ok now.

src/otx/algorithms/common/adapters/torch/dataloaders/samplers/otx_sampler.py
sungmanc: Thanks for information. I just set the int casting since `self.repeat` will be float to handle the large dataset. 
For the large dataset, we don't need to use all of dataset, so will sampling by controlling the repeat (i.e. self.repeat=0.7)

eunwoosh: Thanks for information. Nevertheless, `self.num_samples` will be integer right? then it seems that result whether there is int casting or not is same.

sungmanc: Ah, you mean that `int(9) * float(0.7) / int(1)` will be integer ?

eunwoosh: I mean `math.ceil(int(9) * float(0.7) / int(1))` is integer as `int(math.ceil(int(9) * float(0.7) / int(1)))` is.

sungmanc: !! I mis-recognized the location of the `)`, thanks :)

src/otx/algorithms/common/adapters/torch/dataloaders/samplers/otx_sampler.py
sungmanc: done @https://github.com/openvinotoolkit/training_extensions/pull/2428/commits/a496a82861ea094fca937a3c06caff9aac85f3fd

src/otx/algorithms/common/adapters/mmcv/hooks/adaptive_repeat_data_hook.py
sungmanc: done https://github.com/openvinotoolkit/training_extensions/pull/2428/commits/a496a82861ea094fca937a3c06caff9aac85f3fd

src/otx/algorithms/common/adapters/mmcv/hooks/adaptive_repeat_data_hook.py
sungmanc: done https://github.com/openvinotoolkit/training_extensions/pull/2428/commits/a496a82861ea094fca937a3c06caff9aac85f3fd

src/otx/algorithms/common/adapters/torch/dataloaders/samplers/cls_incr_sampler.py
sungmanc: done https://github.com/openvinotoolkit/training_extensions/pull/2428/commits/a496a82861ea094fca937a3c06caff9aac85f3fd

src/otx/algorithms/common/adapters/torch/dataloaders/samplers/otx_sampler.py
sungmanc: done https://github.com/openvinotoolkit/training_extensions/pull/2428/commits/a496a82861ea094fca937a3c06caff9aac85f3fd

src/otx/algorithms/common/adapters/torch/dataloaders/samplers/otx_sampler.py
sungmanc: done https://github.com/openvinotoolkit/training_extensions/pull/2428/commits/a496a82861ea094fca937a3c06caff9aac85f3fd

src/otx/algorithms/common/adapters/torch/dataloaders/samplers/otx_sampler.py
sungmanc: done https://github.com/openvinotoolkit/training_extensions/pull/2428/commits/a496a82861ea094fca937a3c06caff9aac85f3fd

src/otx/algorithms/common/adapters/torch/dataloaders/samplers/balanced_sampler.py
sungmanc: done https://github.com/openvinotoolkit/training_extensions/pull/2428/commits/a496a82861ea094fca937a3c06caff9aac85f3fd

src/otx/algorithms/classification/adapters/mmcls/configurer.py
sungmanc: I added the priority for each hooks. The default training should start with `AdaptiveRepeatDataHook` and then check the event whether `TaskAdaptHook` needs to change the `Sampler` or not. So, if OTX needs to change the `ClsIncrSampler` or  `BalancedSampler`, it'll overwrite the previous `OTXSampler`.

sungmanc:
> Thanks for your work! It can reduce much overhead, I think. I left some comments, please take a look.

Thanks for the detailed review, it helps a lot :), I finished to reflect your review

src/otx/algorithms/classification/adapters/mmcls/configurer.py
eunwoosh: Thanks for change. But how about removing hook or set some flag to decide whether making a new DataLoader to reduce data time?

sungmanc: At first, I also think like what you think. The disadvantage of that way is that we need to another pathcing inside of the code for the hook deletion.  I'll consider the clean way while implementing `AdaptiveRepeatDataHook` for the det. / seg. / iseg. So, how about skipping at this PR? If you don't mind

eunwoosh (dismissed):
LGTM. Thanks for applying lots of comment :)

src/otx/algorithms/classification/configs/deit_tiny/model.py
```diff
@@ -18,3 +18,5 @@
 
 optimizer = dict(_delete_=True, type="AdamW", lr=0.01, weight_decay=0.05)
 optimizer_config = dict(_delete_=True)
+
+custom_hooks = [dict(type="AdaptiveRepeatDataHook", priority="ABOVE_NORMAL")]
```
jaegukhyun: Can we move this to recipe?

jaegukhyun (dismissed):
LGTM, I have one comment

src/otx/algorithms/classification/configs/deit_tiny/model.py
sungmanc: Good idea, done.

eunwoosh (dismissed):

eunwoosh (approved):

src/otx/algorithms/common/adapters/torch/dataloaders/samplers/otx_sampler.py
```diff
@@ -0,0 +1,99 @@
...
+
+        self.dataset, _ = unwrap_dataset(dataset)
+        self.samples_per_gpu = samples_per_gpu
+        self.num_replicas = num_replicas
+        self.rank = rank
+        self.shuffle = shuffle
+        self.repeat = self._get_proper_repeats(use_adaptive_repeats, coef, min_repeat)
+
+        self.num_samples = math.ceil(len(self.dataset) * self.repeat / self.num_replicas)
+        self.total_size = self.num_samples * self.num_replicas
```
goodsong81, line 64: Meaning num_samples_per_worker?

src/otx/algorithms/common/adapters/torch/dataloaders/samplers/otx_sampler.py
```diff
@@ -0,0 +1,99 @@
...
+        """Calculate the proper repeats with considering the number of iterations."""
+        n_repeats = 1
+        if use_adaptive_repeats:
+            # NOTE
+            # Currently, only support the integer type repeats.
+            # Will support the floating point repeats and large dataset cases.
+            n_iters_per_epoch = math.ceil(len(self.dataset) / self.samples_per_gpu)
+            n_repeats = math.floor(max(coef * math.sqrt(n_iters_per_epoch - 1) + 5, min_repeat))
+            logger.info("OTX Sampler: adaptive repeats enabled")
+            logger.info(f"OTX will use {n_repeats} times larger dataset made by repeated sampling")
```
goodsong81, line 76: Should we limit the `n_repeat` as integer? Rather, I thought this sampler would adapt `num_iters_per_epoch` more continuously which is effectively len(inidces) in __iter__().

src/otx/algorithms/common/adapters/torch/dataloaders/samplers/otx_sampler.py
```diff
@@ -0,0 +1,99 @@
...
+        return n_repeats
+
+    def __iter__(self):
+        """Iter."""
+        indices = list(range(len(self.dataset)))
+
+        # produce repeats e.g. [0, 0, 0, 1, 1, 1, 2, 2, 2....]
+        indices = [x for x in indices for _ in range(self.repeat)]
```
goodsong81, line 84: Would this be better in performance? I'm not sure.
indices = indices*self.repeat
or
indices = np.repeat(indices, self.repeat)

goodsong81 (commented):
Thank you for the update. I have a few questions.

src/otx/algorithms/common/adapters/torch/dataloaders/samplers/otx_sampler.py
sungmanc: AFAIK, this doesn't work in per worker, it works in per rank (distributed environment)

src/otx/algorithms/common/adapters/torch/dataloaders/samplers/otx_sampler.py
sungmanc: We don't need to. I just set this value as integer since I didn't implement yet for the continuous values.

src/otx/algorithms/common/adapters/torch/dataloaders/samplers/otx_sampler.py
sungmanc: Hmm, `np.repeat` seems good. I just used the pre-existed one. https://github.com/open-mmlab/mmpretrain/blob/732b0f4c9869d952eca7d66cb82ad0be8d83cc10/mmpretrain/datasets/samplers/repeat_aug.py#L74

goodsong81 (approved):

PR #2442, opened 2023-08-21
Refactor configurer - Phase 3

jaegukhyun:
### Summary
This PR includes final refactoring configurer procedures.
1. Remove unused parameters from model config to prevent user from confusion
2. Create new configurer mixins for incremental learning and semi supervised learning to reduce duplicated codes across algos.

### Checklist

### License

```python
```


src/otx/algorithms/classification/adapters/mmcls/configurer.py
```diff
@@ -284,49 +222,25 @@ def configure_loss(self, cfg):
             )
             update_or_add_custom_hook(cfg, ib_loss_hook)
 
+    @staticmethod
+    def get_sampler_type(cfg):
+        """Return sampler type."""
+        if not cfg.model.get("multilabel", False) and not cfg.model.get("hierarchical", False):
```
eunwoosh: How about to make this line as a function (such as `_is_multiclass()`?) to increase readability and reusability?

src/otx/algorithms/classification/adapters/mmcls/configurer.py
```diff
@@ -284,49 +222,25 @@ def configure_loss(self, cfg):
...
+            cfg["task_adapt"].get("efficient_mode", True)
+            sampler_type = "balanced"
+        else:
+            cfg["task_adapt"].get("efficient_mode", False)
+            sampler_type = "cls_incr"
+        return sampler_type
```
goodsong81: What are the `get()` in line#229 & 232 doing?

src/otx/algorithms/common/adapters/mmcv/clsincr_mixin.py
```diff
@@ -0,0 +1,64 @@
...
+class IncrConfigurerMixin:
+    """Patch config to support incremental learning for object detection."""
+
+    org_model_classes: List = []
+    model_classes: List = []
+
+    def configure_task(self, cfg):
+        """Patch config to support incremental learning."""
+        super().configure_task(cfg)
+        if "task_adapt" in cfg and self.task_adapt_type == "mpa":
```
goodsong81: Why don't you take the change to change the name "mpa"? :)

goodsong81 (dismissed):
Thanks for the neat refactoring. A few comments of mine:

src/otx/algorithms/common/adapters/mmcv/semisl_mixin.py
```diff
@@ -0,0 +1,62 @@
...
+class SemiSLConfigurerMixin:
+    """Patch config to support semi supervised learning."""
+
+    def configure_data(self, cfg, data_cfg):
+        """Patch cfg.data."""
+        super().configure_data(cfg, data_cfg)
+        # Set unlabeled data hook
+        if self.training:
+            if cfg.data.get("unlabeled", False) and cfg.data.unlabeled.get("otx_dataset", False):
+                if len(cfg.data.unlabeled.get("pipeline", [])) == 0:
```
eunwoosh, line 25:
```suggestion
                if cfg.data.unlabeled.get("pipeline", []):
```
How about to align with PEP recommended way to check a empty list?

src/otx/algorithms/common/adapters/mmcv/clsincr_mixin.py
```diff
@@ -0,0 +1,64 @@
...
+        )
+
+    def is_incremental(self) -> bool:
+        """Return whether current model classes is increased from original model classes."""
+        sampler_flag = True
+        if len(set(self.org_model_classes) & set(self.model_classes)) == 0 or set(self.org_model_classes) == set(
+            self.model_classes
+        ):
+            sampler_flag = False
+        return sampler_flag
```
eunwoosh:
```suggestion
        return (
            set(self.org_model_classes) & set(self.model_classes)
            and set(self.org_model_classes) != set(self.model_classes)
        )
```
How about to change like this?

src/otx/algorithms/segmentation/adapters/mmseg/configurer.py
```diff
@@ -202,89 +197,27 @@ def configure_input_size(
...
-        task_adapt_hook = ConfigDict(
-            type="TaskAdaptHook",
-            src_classes=self.org_model_classes,
-            dst_classes=self.model_classes,
-            model_type=cfg.model.type,
-            sampler_flag=is_cls_incr,
-            efficient_mode=cfg["task_adapt"].get("efficient_mode", False),
-        )
-        update_or_add_custom_hook(cfg, task_adapt_hook)
+        return is_cls_incr
```
eunwoosh, line 212:
```suggestion
        # TODO: Revisit this part when removing bg label -> it should be 1 because of 'background' label
        return (
            len(set(self.org_model_classes) & set(self.model_classes)) > 1
            and set(self.org_model_classes) != set(self.model_classes)
        )
```
How about to change code like this?

src/otx/algorithms/classification/adapters/mmcls/configurer.py
```diff
@@ -200,78 +196,20 @@ def configure_input_size(
...
-
-            if cfg.model.type in WEIGHT_MIX_CLASSIFIER:
-                cfg.model.task_adapt = ConfigDict(
-                    src_classes=self.org_model_classes,
-                    dst_classes=self.model_classes,
-                )
-            else:
-                logger.warning(f"Weight mixing for {cfg.model.type} is not yet supported!")
-
-            train_data_cfg.classes = self.model_classes
```
eunwoosh, line 238: It seems that these lines of code disappear. Is it ok?

eunwoosh (commented):
Thanks for your effort. Almost LGTM, but I have some comments. Please take a look.

src/otx/algorithms/classification/adapters/mmcls/configurer.py
eunwoosh, line 238: It seems that these lines of code disappear. Is it ok?

src/otx/algorithms/classification/adapters/mmcls/configurer.py
jaegukhyun: They are unused codes, so I deleted them

src/otx/algorithms/common/adapters/mmcv/clsincr_mixin.py
jaegukhyun: I changed it to default_task_adapt to make easy to search. If you have any opinion for name, please notice me.

src/otx/algorithms/classification/adapters/mmcls/configurer.py
jaegukhyun: Good idea, I changed them

src/otx/algorithms/common/adapters/mmcv/clsincr_mixin.py
jaegukhyun: I changed with your way, but I have to modify a little due to some pre-commit issues

src/otx/algorithms/classification/adapters/mmcls/configurer.py
jaegukhyun: They just print warning without any changes, so I just deleted them for simpler codes.

eunwoosh (dismissed):

sungmanc (dismissed):

src/otx/algorithms/classification/adapters/mmcls/configurer.py
```diff
@@ -284,49 +222,25 @@ def configure_loss(self, cfg):
...
-    def configure_hook(self, cfg):
-        """Update cfg.custom_hooks."""
-        super().configure_hook(cfg)
-        # Set unlabeled data hook
-        if self.training:
-            if cfg.data.get("unlabeled", False) and cfg.data.unlabeled.get("otx_dataset", False):
-                self.configure_unlabeled_dataloader(cfg)
+        Currently, only multi class classification supports adaptive repeat.
+        """
+        return self._is_multiclass(cfg)
```
harimkang, line 238: Will use_adaptive_repeat always be true if it's a multi class?
What if the user has a multi-class task and doesn't want to use adaptive_repeat?

src/otx/recipes/stages/classification/incremental.yaml
```diff
@@ -29,7 +29,7 @@ lr_config:
   warmup_ratio: 0.333
 
 task_adapt:
-  type: "mpa"
+  type: "default_task_adapt"
```
harimkang, line 32: Wouldn't something like `enable: True` be better in this case, rather than using type separately?
What do you think?
It doesn't seem to fit with type naming.

tests/unit/algorithms/segmentation/adapters/mmseg/test_mmseg_configurer.py
```diff
@@ -280,40 +272,23 @@ def setup(self) -> None:
...
-    @e2e_pytest_unit
-    def test_configure_unlabeled_dataloader(self, mocker):
-        mocker_build_dataset = mocker.patch("otx.algorithms.segmentation.adapters.mmseg.configurer.build_dataset")
-        mocker_build_dataloader = mocker.patch("otx.algorithms.segmentation.adapters.mmseg.configurer.build_dataloader")
-        self.configurer.configure_hook(self.cfg)
-
-        mocker_build_dataset.assert_called_once()
-        mocker_build_dataloader.assert_called_once()
-        assert "ComposedDataLoadersHook" in [hook["type"] for hook in self.cfg.custom_hooks]
+        data_cfg = MPAConfig(
```
harimkang, line 281: What about renamed MPAConfig?
I don't think it's necessary to do this in this PR, but I thought I'd mention that MPA shouldn't be visible in OTX anymore.

src/otx/algorithms/classification/adapters/mmcls/configurer.py
sungmanc: Same with the `AdaptiveTrainingSchedule`, it will be automatically triggered. If users don't want to use, they could do by modifying recipe at the current status to remove the `CustomHook` named `AdaptiveRepeatDataHook`.

jaegukhyun: Current only multi class classification has two methods for adaptive training. First is adaptive training with adaptive epoch validation, which is common for the other tasks. The other is adaptive training with repeat sampler. This method enlarges iterations per epoch. Currently we all agreed that default operation of adaptive training for multi class classification is repeat sampler. Therefore, if user don't want this, then user should turn off adaptive_training flag.

harimkang: But shouldn't this function always come out as True, even if the user changes adaptive repeat to disable?

tests/unit/algorithms/segmentation/adapters/mmseg/test_mmseg_configurer.py
jaegukhyun: Yes there are some mpa legacy in config and logger. I think they are beyond this PR. Let's handle this in another PR, but I'm not sure when it will be ready.

src/otx/algorithms/classification/adapters/mmcls/configurer.py
harimkang: What I mean is that I don't think it's semantically correct for the `use_adaptive_repeat` function to `return is_multiclass` as it is. I think this function needs to be renamed, what do you think?

jaegukhyun: > But shouldn't this function always come out as True, even if the user changes adaptive repeat to disable?

Yes, but it does not effect training when users disable adaptive training

harimkang: > > But shouldn't this function always come out as True, even if the user changes adaptive repeat to disable?
> 
> Yes, but it does not effect training when users disable adaptive training

Yeah I was worried that from an API perspective, this is a public function and it would be confusing for users to see it always be true even when disabled.

jaegukhyun: > What I mean is that I don't think it's semantically correct for the use_adaptive_repeat function to return is_multiclass as it is.

I understand your concern, but it is refactoring PR, and this PR didn't change any logic for this. Therefore we need to discuss this and patch in anather PR

Previous code is
        task_adapt_hook = ConfigDict(
            type="TaskAdaptHook",
            src_classes=self.org_model_classes,
            dst_classes=self.model_classes,
            model_type=cfg.model.type,
            sampler_flag=sampler_flag,
            sampler_type=sampler_type,
            efficient_mode=efficient_mode,
            use_adaptive_repeat=is_multiclass,
            priority="NORMAL",
        )

sungmanc: I agree with @jaegukhyun and @harimkang . Current implementation is still temporary since we need to more implementation for multi-label and h-label and other tasks (det., seg.). So, it'll be refactored at my later PRs.

harimkang: Yes thanks for the clarification. I guess it's a function that will be removed in the near future.

harimkang:
Thank you for your hard work. I have a few comments, but they are not critical. :)

src/otx/recipes/stages/classification/incremental.yaml
jaegukhyun: I'm not sure, since I don't have any background that why previous task adapt type is mpa. I just guess there could be another option for task_adapt. If there's not, maybe we can change type to enable. But it requires lots of changes, I want to make another PR for this and other things we discuss

harimkang: yeah, no problem

goodsong81 (dismissed):

harimkang (dismissed):

sungmanc (dismissed):

src/otx/algorithms/common/adapters/mmcv/semisl_mixin.py
jaegukhyun: I revert this change since it raises errors for segmentation semisl, and hangs for detection semisl.

goodsong81 (approved):

sungmanc (approved):

PR #2450, opened 2023-08-23
Fix OpenVINO inference for legacy models

ashwinvaidya17:
### Summary

- Addresses : [CVS-118551](https://jira.devtools.intel.com/browse/CVS-118551)

## How to test

- Checkout releases/1.2.2
- Train and export openvino model
- Checkout 1.4.0
- Call eval

### License

```python
```

### DIFF
--- a/src/otx/algorithms/anomaly/tasks/openvino.py
+++ b/src/otx/algorithms/anomaly/tasks/openvino.py
@@ -26,6 +26,7 @@ import nncf
 import numpy as np
 import openvino.runtime as ov
 from addict import Dict as ADDict
+from anomalib.data.utils.transform import get_transforms
 from anomalib.deploy import OpenVINOInferencer
 from nncf.common.quantization.structs import QuantizationPreset
 from omegaconf import OmegaConf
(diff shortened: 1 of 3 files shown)


goodsong81 (commented):
Looks generally OK but why don't you add 1 unit test that simulates this scenario?

ashwinvaidya17:
> Looks generally OK but why don't you add 1 unit test that simulates this scenario?

How should that look? Should I upload the openvino model somewhere? The problem is that only the models trained using a previous version do not work the current version.

goodsong81:
> > Looks generally OK but why don't you add 1 unit test that simulates this scenario?
> 
> How should that look? Should I upload the openvino model somewhere? The problem is that only the models trained using a previous version do not work the current version.

Ah, I didn't mean that realistic scenario with actual file I/O, but we can raise exception using mocks to check if the `_populate_metadata()` or `_populate_metadata_legacy()` is called correctly, and also the outputs are as expected.
It's just suggestion. Up to you :)

ashwinvaidya17:
> Ah, I didn't mean that realistic scenario with actual file I/O, but we can raise exception using mocks to check if the `_populate_metadata()` or `_populate_metadata_legacy()` is called correctly, and also the outputs are as expected. It's just suggestion. Up to you :)

I've added the tests now.

src/otx/algorithms/anomaly/tasks/openvino.py
```diff
@@ -216,16 +217,47 @@ def get_metadata(self) -> Dict:
...
-            metadata["min"] = np.array(metadata["min"], dtype=np.float32).item()
-            metadata["max"] = np.array(metadata["max"], dtype=np.float32).item()
+            try:
+                metadata = json.loads(self.task_environment.model.get_data("metadata").decode())
+                self._populate_metadata(metadata)
+                logger.info("Metadata loaded from model v1.4.")
+            except Exception:
+                # model is from version 1.2.x
+                metadata = self._populate_metadata_legacy(self.task_environment.model)
+                logger.info("Metadata loaded from model v1.2.x.")
```
djdameln: Would there be a more explicit way to check the model version? If not, I think it would at least be good to catch a more specific exception.

djdameln (changes requested):
Thanks. One minor comment

src/otx/algorithms/anomaly/tasks/openvino.py
```diff
@@ -216,16 +217,47 @@ def get_metadata(self) -> Dict:
...
         if self.task_environment.model is not None:
-            metadata = json.loads(self.task_environment.model.get_data("metadata").decode())
-            metadata["image_threshold"] = np.array(metadata["image_threshold"], dtype=np.float32).item()
-            metadata["pixel_threshold"] = np.array(metadata["pixel_threshold"], dtype=np.float32).item()
-            metadata["min"] = np.array(metadata["min"], dtype=np.float32).item()
-            metadata["max"] = np.array(metadata["max"], dtype=np.float32).item()
+            try:
+                metadata = json.loads(self.task_environment.model.get_data("metadata").decode())
+                self._populate_metadata(metadata)
+                logger.info("Metadata loaded from model v1.4.")
```
samet-akcay, line 223: Will this logging message require an update once otx is upgraded to 1.5+

samet-akcay (commented):
thanks, looking good to me. I only have a minor comment.

src/otx/algorithms/anomaly/tasks/openvino.py
ashwinvaidya17: If we update the task then we might have to add another check that can then mention 1.5. But if we don't change the task then the model format will still be 1.4.x right?

samet-akcay (approved):

goodsong81 (approved):

djdameln (approved):

PR #2454, opened 2023-08-25
Make MaskRCNN-SwinT from experimental to production

eugene123tw:
### Summary

- Experimental template to production template
- Keep aspect ratio and add paddings for all MaskRCNN data pipeline

### Checklist

### License

```python
```


src/otx/algorithms/detection/adapters/openvino/task.py
```diff
@@ -453,7 +453,9 @@ def load_inferencer(
             inferencer: BaseInferencerWithConverter = OpenVINODetectionInferencer(*args)
         if self.task_type == TaskType.INSTANCE_SEGMENTATION:
             if self.config.tiling_parameters.enable_tiling:
-                args.append({"resize_type": "fit_to_window_letterbox"})
+                args.append({"resize_type": "fit_to_window_letterbox", "pad_value": 114})
+            else:
+                args.append({"resize_type": "fit_to_window"})
```
jaegukhyun: I think logic should be changed. if tiling -> fit to window, else -> fit to window letterbox + padding

eugene123tw: @jaegukhyun I think it should be changed to `fit to window` for Instance Segmentation's case. Padding works differently for these 2 types, `fit_to_window` pads only bottom and right, but letterbox pads both left, right and top, bottom equally. Letterbox padding is mainly used in YOLO-based model, but I don't see it used in other architecture. 

I also observed 2% accuracy drop when I changed to letterbox padding for MaskRCNN.

sovrasov: Originally, I added `"resize_type": "fit_to_window_letterbox", "pad_value": 114` for YOLOX only, so I don't think this setup is correct for any other model (expecially pad_value=114). resize_cfg=dict(type="Resize", img_scale=__img_size, keep_ratio=True), corresponds to "fit_to_window". At the same time I don't see how to reproduce dict(type="Pad", size_divisor=32) in MAPI, therefore we may decrease accuracy by adding it.

eugene123tw: @sovrasov Instead of padding like a letterbox, could we add this `np.pad(a, ((0, pad_bottom), (0, pad_right), (0, 0)))` to `resize_image_with_aspect`?

sovrasov: In theory, we could implement the same resize type in several ways via numpy, but that's not the case for current inference architecture. We can't use numpy there and have to implement the preprocessing by means of openvino opset.
If existing resize options in MAPI don't cover all the desired cases, I propose the following process: gather information about all resize/padding types in one document and start a discussion with MAPI owners if it's possible to implement this via current ov opset and schedule corresponding works on MAPI side.

src/otx/algorithms/detection/configs/base/data/iseg_efficientnet_data_pipeline.py
jaegukhyun, line 1: Is there reason for not adding Padding pipeline?

src/otx/algorithms/detection/adapters/openvino/task.py
jaegukhyun: @eugene123tw Thanks for the reply, I have a misunderstand for "fit-to-window" ops. However I have a another question. Why iseg and tiling use different ops while they are use same pipeline(Resize -> padding). Is it because Iseg use size_divisor?

jaegukhyun: @eugene123tw Please check https://github.com/openvinotoolkit/model_api/blob/master/model_api/python/openvino/model_api/models/utils.py#L289
It seems fit_to_window means "Resize with aspected ratio and no additional padding" and 
fit_to_window_letterbox means "Resize with aspected ratio and additional padding to right and bottom"
You changed iseg's data pipeline to Resize and Pad, so it should be "fit_to_window_letterbox"
And most of tiling pipeline use simple resize(Resize(keep_ratio=False)), therefore it should be "standard"

eugene123tw: @jaegukhyun `fit_to_window_letterbox` pads 4 sides (either top+bottom or left+right) https://github.com/openvinotoolkit/model_api/blob/master/model_api/python/openvino/model_api/models/utils.py#L298
image

```
return np.pad(
    image,
    ((dy, h - nh - dy), (dx, w - nw - dx), (0, 0)),
    mode="constant",
    constant_values=pad_value,
)
```
`fit_to_window` does not pad atm, but it should behave like mmcv padding where padding is done only bottom and right. 

`fit_to_window_letterbox` is not used due to OpenVINO doesn't take np ops but instead https://github.com/openvinotoolkit/model_api/blob/master/model_api/python/openvino/model_api/adapters/utils.py#L91

src/otx/algorithms/detection/configs/base/data/iseg_efficientnet_data_pipeline.py
eugene123tw: [its here, no?](https://github.com/openvinotoolkit/training_extensions/blob/develop/src/otx/algorithms/detection/configs/base/data/iseg_efficientnet_data_pipeline.py#L66)

jaegukhyun (approved):

goodsong81 (approved):
Thank you for the update!

PR #2489, opened 2023-09-08
Set mem-cache-1G as a default training

sungmanc:
### Summary

It enables the mem-cache logic as a default for OTX classification, detection, i-seg, s-seg.
Note applied to rotated det., visual prompting which couldn't support the mem cache functionality.

### Checklist

### License

```python
```


goodsong81 (commented):
LGTM, but why rotated detection is not available? They use the same model & data pipelines with instance segmentation (by far).

sungmanc:
> LGTM, but why rotated detection is not available? They use the same model & data pipelines with instance segmentation (by far).

I'm not sure. However, there is no `mem_cache_size` in the configuration for rotated detection. @eugene123tw is there any reason?

goodsong81:
> > LGTM, but why rotated detection is not available? They use the same model & data pipelines with instance segmentation (by far).
> 
> I'm not sure. However, there is no `mem_cache_size` in the configuration for rotated detection. @eugene123tw is there any reason?

There must be no reason. We just often forget the rotated detection update for common features. ;)
Could you apply in this PR?

sungmanc:
> > > LGTM, but why rotated detection is not available? They use the same model & data pipelines with instance segmentation (by far).
> > 
> > 
> > I'm not sure. However, there is no `mem_cache_size` in the configuration for rotated detection. @eugene123tw is there any reason?
> 
> There must be no reason. We just often forget the rotated detection update for common features. ;) Could you apply in this PR?

Sure, I'll add it to this PR :)

sungmanc:
For the rotated detection, we need to fix some parts. It didn't work.
When I executed with develop branch, I got error below
image

After copying the `configuration.yaml` from the instance segmentation. It works at first, however, occurs an error like below
image

So, how about diving the PR for fixing the rotated detection ? I'll make it . Anyway, I added the `mem_cache_size` to the rotated detection configuration.

eugene123tw:
> For the rotated detection, we need to fix some parts. It didn't work. When I executed with develop branch, I got error below image
> 
> After copying the `configuration.yaml` from the instance segmentation. It works at first, however, occurs an error like below image
> 
> So, how about diving the PR for fixing the rotated detection ? I'll make it . Anyway, I added the `mem_cache_size` to the rotated detection configuration.

@sungmanc Could you copy configuration.xml in this PR: https://github.com/openvinotoolkit/training_extensions/pull/2246

Let me know if you still have issue

goodsong81 (dismissed):

sungmanc:
Added a logic to check the CPU status and then control the mem cache size. https://github.com/openvinotoolkit/training_extensions/pull/2489/commits/d20983858a91483041aa4fe240203822bcb6c875

src/otx/core/data/caching/mem_cache_handler.py
```diff
@@ -188,10 +190,20 @@ def create(cls, mode: str, mem_size: int) -> MemCacheHandlerBase:
...
+        available_cpu_mem = memory_info.available / GIB
+
         if world_size > 1:
             mem_size = mem_size // world_size
+            available_cpu_mem = available_cpu_mem // world_size
             logger.info(f"Since world_size={world_size} > 1, each worker a {mem_size} size memory pool.")
 
+        logger.info(f"Try to create a {mem_size} size memory pool.")
+        if available_cpu_mem < ((mem_size / GIB) + cls.CPU_MEM_LIMITS_GIB):
+            logger.info("No available CPU memory left, mem_size will be set to 0.")
```
eunwoosh:
```suggestion
            logger.warning("No available CPU memory left, mem_size will be set to 0.")
```
How about changing it to `warning`?
And Current code sets mem cache size as 0 if there isn't sufficient memory, but how about adapting it according to memory size?
For example, if available memory is 16GiB, then set mem cache size as `16GiB / 30`.

src/otx/algorithms/common/adapters/mmcv/pipelines/load_image_from_otx_dataset.py
```diff
@@ -162,10 +169,13 @@ def _resize_img_ann_if_any(self, results: Dict[str, Any]) -> Dict[str, Any]:
 
     def _load_cache(self, results: Dict[str, Any]) -> Optional[Dict[str, Any]]:
         """Try to load pre-computed results from cache."""
+
         if not self._enable_outer_memcache:
             return None
         key = self._get_unique_key(results)
-        img, meta = self._mem_cache_handler.get(key)
+
+        mem_cache_handler = self._get_memcache_handler()
```
eunwoosh, line 177: It's just a question. It seems that handler is gotten every times this function is called, Is overhead for that neglectable?

eunwoosh (dismissed):
LGTM. I left some comments. Please take a look.

src/otx/algorithms/common/adapters/mmcv/pipelines/load_image_from_otx_dataset.py
sungmanc: Yes, no overhead exists.

src/otx/core/data/caching/mem_cache_handler.py
sungmanc: It's a good option. However, this memory checking is conducted before the construction of the DataLoader. So, after this checking, large CPU memory will be consumed by Dataset, DataLoader, and some functions (i.e. evaluation).

I just left 30GiB as a minimum bound since I saw the OTX using almost 30GiB when the large images were used at instance segmentation or rotated detection tasks. 

Since this logic was added to guarantee stability, more room could be safe :) 
Let's change the logic as you said after we get some experience.

eunwoosh (approved):

goodsong81 (approved):

PR #2495, opened 2023-09-12
Fix unit tests for XAI

GalyaZalesskaya:
### Summary

This PR is an addition to [PR#2491](https://github.com/openvinotoolkit/training_extensions/pull/2491)
It fixes a floating error on CI where reference values of saliency maps can differ in 1 with actual values.

### Checklist

### License

```python
```

### DIFF
--- a/tests/unit/algorithms/classification/test_xai_classification_validity.py
+++ b/tests/unit/algorithms/classification/test_xai_classification_validity.py
@@ -65,7 +65,9 @@ class TestExplainMethods:
         assert saliency_maps[0].shape == saliency_map_ref_shape
         actual_sal_vals = saliency_maps[0][0][0].astype(np.uint8)
         ref_sal_vals = self.ref_saliency_vals_cls[template.name].astype(np.uint8)
-        assert np.all(np.abs(actual_sal_vals - ref_sal_vals) <= 1)
+        # convert to int8 in case of negative delta
+        delta_sal_map = np.abs((actual_sal_vals - ref_sal_vals).astype(np.int8))
+        assert np.all(delta_sal_map) <= 1
 
 
 class TestViTExplain:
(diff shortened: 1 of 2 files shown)


tests/unit/algorithms/classification/test_xai_classification_validity.py
```diff
@@ -65,7 +65,9 @@ def test_saliency_map_cls(self, template):
         assert saliency_maps[0].shape == saliency_map_ref_shape
         actual_sal_vals = saliency_maps[0][0][0].astype(np.uint8)
         ref_sal_vals = self.ref_saliency_vals_cls[template.name].astype(np.uint8)
-        assert np.all(np.abs(actual_sal_vals - ref_sal_vals) <= 1)
+        # convert to int8 in case of negative delta
+        delta_sal_map = np.abs((actual_sal_vals - ref_sal_vals).astype(np.int8))
```
sovrasov: Should we convert one of the operands to signed int first to let "-" operation avoid overflow?

GalyaZalesskaya: Actually, if we convert the difference to int8 like here, it converts the negative difference correctly, without the overflow

sovrasov: Overflow occurs , but later these numbers overflow astype as well:
>>> a = np.array([0,0,0]).astype(np.uint8)
>>> b = np.array([1,2,3]).astype(np.uint8)
>>> a - b
array([255, 254, 253], dtype=uint8)
>>> (a - b).astype(np.int8)
array([-1, -2, -3], dtype=int8)

Exploiting this behavior is a quite thin ice, since conversion to int32 leads to expected, but not desirable result (the second overflow doesn't cancel the effect of the first one):
>>> (a - b).astype(np.int32)
array([255, 254, 253], dtype=int32)

GalyaZalesskaya: Yes, you're right. I'll remove this uncertainty not to shoot ourselves in the foot

sovrasov (approved):

eunwoosh (approved):

PR #2497, opened 2023-09-12
Add RTM-Det Inst Tiny

eugene123tw:
### Checklist

### License

```python
```


src/otx/algorithms/detection/configs/instance_segmentation/rtmdet_ins_tiny/template.yaml
```diff
@@ -0,0 +1,64 @@
...
+application: ~
+
+# Algo backend.
+framework: OTXDetection v2.9.1
+
+# Task implementations.
+entrypoints:
+  base: otx.algorithms.detection.adapters.mmdet.task.MMDetectionTask
+  openvino: otx.algorithms.detection.adapters.openvino.task.OpenVINODetectionTask
+  nncf: otx.algorithms.detection.adapters.mmdet.nncf.task.DetectionNNCFTask
```
goodsong81, line 17: We don't need to support NNCF QAT for new models. Let's remove this line.

goodsong81 (changes requested):
Thank you for the great work!
I'd like to request a small change, and a little bit more tests.
Unit tests for modules copied from mmlab are not needed as it will be removed in mmdet3.0 era, but we need to implement tests for our own custom modules. In addition, it might be good to have integration / e2e tests and documentation update. Not necessarily in this PR, though.

src/otx/algorithms/detection/adapters/mmdet/models/assigners/dynamic_soft_label_assigner.py
```diff
@@ -0,0 +1,224 @@
+"""Dynamic Soft Label Assigner for RTMDet."""
+# Copyright (c) OpenMMLab. All rights reserved.
```
sungmanc, line 2: @yunchu , is there no problem use this copyright?

src/otx/algorithms/detection/configs/instance_segmentation/rtmdet_ins_tiny/data_pipeline.py
```diff
@@ -0,0 +1,101 @@
...
+train_pipeline = [
+    dict(type="LoadImageFromOTXDataset", enable_memcache=True),
+    dict(
+        type="LoadAnnotationFromOTXDataset",
+        domain="instance_segmentation",
+        with_bbox=True,
+        with_mask=True,
+        poly2mask=False,
+    ),
+    dict(type="CachedMosaic", img_scale=img_size, pad_val=114.0, max_cached_images=20, random_pop=False),
```
sungmanc, line 46: This is the extreme case.

What would happen if there is no memory for `CachedMosaic` or `CachedMixUp`? It seems that it will raise an error if there are < 4 cached_images, is it safe to use?

sungmanc (commented):
Thanks for the introducing nice architecture. 
I left 2 comments for this PR.

src/otx/algorithms/detection/adapters/mmdet/datasets/pipelines/transforms.py
jaegukhyun, line 1: Is this file going to remove when we migrate to mmdet3.x?

src/otx/api/usecases/exportable_code/prediction_to_annotation_converter.py
jaegukhyun, line 1: Could you explain about this change? It seems affecting to all instance segmentation models

jaegukhyun (dismissed):
Generally LGTM, I added some questions

src/otx/algorithms/detection/adapters/mmdet/models/assigners/dynamic_soft_label_assigner.py
yunchu: if the code was copied from the OpenMMLab without any modification it should keep the same header and it's better to left a comment about the original code URL.

PR #2498, opened 2023-09-13
Move configuration of configs in task.py to configurer.py

jaegukhyun:
### Checklist

### License

```python
```


src/otx/algorithms/common/adapters/mmcv/configurer.py
```diff
@@ -345,6 +380,69 @@ def configure_compat_cfg(cfg: Config):
...
+                cfg,
+                ConfigDict(
+                    type="OTXProgressHook",
+                    time_monitor=self.time_monitor,
+                    verbose=True,
+                    priority=71,
+                ),
+            )
+        cfg.log_config.hooks.append({"type": "OTXLoggerHook", "curves": self.learning_curves})
+        if hasattr(cfg, "algo_backend"):
```
eunwoosh, line 414: It's just a question. Why is it executed only if cfg has a `algo_backend`?

src/otx/algorithms/common/adapters/mmcv/configurer.py
```diff
@@ -59,15 +80,46 @@ def configure(
...
+        self.configure_hooks(
+            cfg,
+        )
         self.configure_compat_cfg(cfg)
         return cfg
 
+    def merge_configs(self, cfg, data_cfg, data_pipeline_path, hyperparams_from_otx, **kwargs):
+        """Merge model cfg, data_pipeline cfg, data_cfg, and hyperparams from otx cli."""
+
+        logger.info("merge_configs()")
```
eunwoosh:
```suggestion
        logger.debug("merge_configs()")
```
How about to change it as `debug`? It's just my opinion, so if you think `info` is more appropriate, just resolve it.

src/otx/algorithms/common/adapters/mmcv/configurer.py
jaegukhyun: It's just for safe. Updating caching modules require attributes from algo_backend in configuration.yaml file

eunwoosh (approved):

src/otx/algorithms/segmentation/adapters/mmseg/task.py
```diff
@@ -162,13 +124,39 @@ def configure(
...
-            configurer = IncrSegmentationConfigurer("segmentation", training)
+            configurer = IncrSegmentationConfigurer(
+                "segmentation",
+                training,
+                export,
+                self.override_configs,
+                self.on_hook_initialized,
+                self._time_monitor,
+                self._learning_curves,
+            )
```
goodsong81, line 135: Just nitpick: why don't we use kwarg names in case of many argument? At least in product code, not in test code. I cannot see which is which.

goodsong81 (approved):
Generally, looks good.
One question: what is conceptual difference you think of
configure() vs init_task()?

jaegukhyun:
> Generally, looks good. One question: what is conceptual difference you think of configure() vs init_task()?

_init_task is designed for initial settings before configuring config files. At that time, there are bunch of things between load configs and configure configs, but it seems now we can combine _init_task and configure to load_config. I'll make another PR for this.

src/otx/algorithms/segmentation/adapters/mmseg/task.py
jaegukhyun: I'll treat this in another following PR

PR #2499, opened 2023-09-13
Adapt input size automatically based on dataset statistics

goodsong81:
### Summary

* Classification / Segmentation
  - Set input size based on typical large image size in dataset
  - If larger than the default size, keep default size
  - ex) There's no much impact as most of datasets have images > 224x224.

* Detection / Instance segmentation
  - Set input size based on typical large image size in dataset
  - Adjust input size based on typical small shape annotation size in dataset
  - If larger than the default size, keep default size
  - ex) "Peanut" dataset for object detection
  
image

* Set "Auto" as default for all models

### How to test

[Example CMD]

```bash
otx train src/otx/algorithms/detection/configs/detection/cspdarknet_yolox/template.yaml --train-data-roots ./data/peanut --workspace outputs/load-resize-det/feat/load-resize-cache-detiseg/peanut_cspdarknet_yolox_256x256_cache2g params --learning_parameters.input_size Auto --algo_backend.mem_cache_size 1000000000
```

### Checklist

- [x] I have updated the documentation in my target branch accordingly (e.g., [documentation](https://github.com/openvinotoolkit/training_extensions/tree/develop/docs) in develop).

### License

```python
```


src/otx/algorithms/common/adapters/mmcv/utils/config_utils.py
```diff
@@ -853,37 +871,109 @@ def _set_size_value(pipeline: Dict, attr: str, scale: Tuple[Union[int, float], U
...
+            input_size (Tuple[int, int]): Query input size
+            preset_sizes (List[Tuple[int, int]]): List of preset input sizes
+
+        Returns:
+            Tuple[int, int]: Best matching size out of preset. Returns input_size if preset is empty.
+        """
+        if len(preset_sizes) == 0:
+            return input_size
+
+        def scale_of(x):
```
sungmanc:
```suggestion
        def to_log_scale(x):
```
This might be more clear

src/otx/algorithms/common/adapters/mmcv/utils/config_utils.py
```diff
@@ -853,37 +871,109 @@ def _set_size_value(pipeline: Dict, attr: str, scale: Tuple[Union[int, float], U
...
+        """
+
+        logger.info("Adapting model input size based on dataset stat")
+
+        base_input_size = self.base_input_size
+        if isinstance(base_input_size, Dict):
+            base_input_size = base_input_size.get("test", None)
+        logger.info(f"-> Current base input size: {base_input_size}")
+
+        if max_image_size <= 0:
```
sungmanc, line 953: It seems that this case is an error. Don't we need to raise an error if `max_image_size` was calculated < 0? Or want to check `max_image_size > base_input_size` ?

sungmanc (commented):
I left minor comments, BTW, this PR needs the documentation. Will it be done at the upcoming PR? And just wondering about the performance change between before (develop) and after (this patch), no big differences?

src/otx/algorithms/common/adapters/mmcv/utils/config_utils.py
```diff
@@ -853,37 +871,109 @@ def _set_size_value(pipeline: Dict, attr: str, scale: Tuple[Union[int, float], U
...
+        def scale_of(x):
+            return np.log(np.sqrt(x[0] * x[1]))
+
+        input_scale = scale_of(input_size)
+        preset_scales = np.array(list(map(scale_of, preset_sizes)))
+        abs_diff = np.abs(preset_scales - input_scale)
+        return preset_sizes[np.argmin(abs_diff)]
+
+    def adapt_input_size_to_dataset(
+        self, max_image_size: float, min_object_size: float = None, downscale_only: bool = True
```
eunwoosh:
```suggestion
        self, max_image_size: int, min_object_size: Optional[int] = None, downscale_only: bool = True
```
They seem typo.

src/otx/algorithms/common/adapters/mmcv/utils/config_utils.py
```diff
@@ -853,37 +871,109 @@ def _set_size_value(pipeline: Dict, attr: str, scale: Tuple[Union[int, float], U
...
+    ) -> Tuple[int, int]:
+        """Compute appropriate model input size w.r.t. dataset statistics.
+
+        Args:
+            max_image_size (int): Typical large image size of dataset in pixels.
+            min_object_size (int): Typical small object size of dataset in pixels.
```
eunwoosh:
```suggestion
            min_object_size (int, optional): Typical small object size of dataset in pixels.
```
Same as above.

src/otx/algorithms/common/adapters/mmcv/utils/config_utils.py
```diff
@@ -853,37 +871,109 @@ def _set_size_value(pipeline: Dict, attr: str, scale: Tuple[Union[int, float], U
...
+
+        Returns:
+            Tuple[int, int]: (width, height)
+        """
+
+        logger.info("Adapting model input size based on dataset stat")
+
+        base_input_size = self.base_input_size
+        if isinstance(base_input_size, Dict):
+            base_input_size = base_input_size.get("test", None)
```
eunwoosh: Should we consider a case where there is no "test" in `base_input_size` such as semi or self SL?

src/otx/algorithms/common/utils/data.py
```diff
@@ -230,3 +218,115 @@ def __len__(self):
...
+    """Computes robust statistics of image & annotation sizes.
+
+    Args:
+        dataset (DatasetEntity): Input dataset.
+        ann_stat (bool, optional): Whether to compute annotation size statistics. Defaults to False.
+        max_samples (int, optional): Maximum number of dataset subsamples to analyze. Defaults to 1000.
+
+    Returns:
+        Dict[str, Any]: Robust avg, min, max values for images, and annotations optionally.
+            e.x) stat = {
```
eunwoosh:
```suggestion
            ex) stat = {
```
Maybe, it's a typo?

src/otx/algorithms/common/adapters/mmcv/configurer.py
```diff
@@ -492,3 +495,41 @@ def get_data_cfg(cfg, subset):
                 dataset = dataset.dataset
             return dataset
         return cfg.data[subset]
+
+    @staticmethod
+    def adapt_input_size_to_dataset(
```
eunwoosh, line 500: IMHO, it is more suitable to move this function in `InputSizeManager`. How do you think about?

eunwoosh (commented):
Thanks for your work! It's a nice feature. I left some comments, please take a look.

tests/integration/cli/detection/test_detection.py
```diff
@@ -96,20 +96,37 @@ class TestDetectionCLI:
     @pytest.mark.parametrize("template", templates_w_experimental, ids=templates_ids_w_experimental)
     def test_otx_train(self, template, tmp_dir_path):
         tmp_dir_path = tmp_dir_path / "detection"
-        otx_train_testing(template, tmp_dir_path, otx_dir, args)
+        _args = args.copy()
+        # FIXME: remove this block once Issue#2054 resolved
```
eunwoosh:
```suggestion
        # FIXME: remove this block once Issue#2504 resolved
```
I think it's a typo

tests/integration/cli/detection/test_detection.py
```diff
@@ -96,20 +96,37 @@ class TestDetectionCLI:
...
+        otx_train_testing(template, tmp_dir_path, otx_dir, _args)
 
     @e2e_pytest_component
     @pytest.mark.parametrize("template", templates_w_experimental, ids=templates_ids_w_experimental)
     def test_otx_resume(self, template, tmp_dir_path):
         tmp_dir_path = tmp_dir_path / "detection/test_resume"
-        otx_resume_testing(template, tmp_dir_path, otx_dir, args)
+        _args = args.copy()
+        _resume_params = resume_params.copy()
+        # FIXME: remove this block once Issue#2054 resolved
```
eunwoosh:
```suggestion
        # FIXME: remove this block once Issue#2504 resolved
```
Same as above

goodsong81:
> I left minor comments, BTW, this PR needs the documentation. Will it be done at the upcoming PR? And just wondering about the performance change between before (develop) and after (this patch), no big differences?

Thank you for the feedback. I would add documentation in a dedicated PR.
In addition, I'm thinking of reverting the default mode of "Auto" but enable this option in integration test. Turning it will be also handled in other PR if decided.

src/otx/algorithms/common/adapters/mmcv/utils/config_utils.py
goodsong81: `max_image_size` is given input from this method's perspective,
and we don't need to raise error as we can stick to the default value
whenever the image size is not acceptable.

src/otx/algorithms/common/adapters/mmcv/utils/config_utils.py
goodsong81: Will fix in next commit.

src/otx/algorithms/common/adapters/mmcv/utils/config_utils.py
goodsong81: "train" might be more safe. Thanks.

src/otx/algorithms/common/adapters/mmcv/configurer.py
goodsong81: Good point, but I didn't want to let `InputSizeManager` know about `DatasetEntity` and others.
Maybe, it will directly handle `dm.Dataset` in the near future.

goodsong81:
> > I left minor comments, BTW, this PR needs the documentation. Will it be done at the upcoming PR? And just wondering about the performance change between before (develop) and after (this patch), no big differences?
> 
> Thank you for the feedback. I would add documentation in a dedicated PR. In addition, I'm thinking of reverting the default mode of "Auto" but enable this option in integration test. Turning it will be also handled in other PR if decided.

Ah, I've just updated the document.

sungmanc (approved):

eunwoosh (approved):

PR #2500, opened 2023-09-13
Add new argument to track CPU&GPU utilization and memory usage

eunwoosh:
### Summary
Add new argument with which user can easily know CPU and GPU utilization and max memory usage.
Currently, CPU and GPU are supported.
Tracking resource usages starts right before and stops right before `task.train()`.
To get more accurate GPU average utilization, recorded GPU utilizations before it becomes more than max_util/5 are ignored.

### How to test
otx train --track-resource-usage all

### Checklist

### License

```python
```

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -10,3 +10,4 @@ scipy>=1.8
 bayesian-optimization>=1.2.0
 tensorboard>=2.11.0
 multiprocess
+pynvml
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -10,4 +10,4 @@ scipy>=1.8
 bayesian-optimization>=1.2.0
 tensorboard>=2.11.0
 multiprocess
-pynvml
+pynvml>=11.0.0
--- a/src/otx/cli/utils/experiment.py
+++ b/src/otx/cli/utils/experiment.py
@@ -215,6 +215,9 @@ class GpuUsageRecorder(ResourceRecorder):
 
     def report(self) -> Dict[str, str]:
         """Aggregate GPU usage."""
+        if not list(self._record.values())[0]["util_record"]:  # record isn't called
+            return {}
+
         total_max_mem = 0
         total_avg_util = 0
         gpus_record = self._record.copy()
--- a/src/otx/cli/utils/experiment.py
+++ b/src/otx/cli/utils/experiment.py
@@ -69,7 +69,7 @@ def _check_resource(
     trackers: Dict[str, ResourceRecorder] = {}
     for resource_type in resource_types:
         if resource_type == "cpu":
-            trackers[resource_type] = CpuUsageRecorder(psutil.Process().parent())
+            trackers[resource_type] = CpuUsageRecorder()
         elif resource_type == "gpu":
             if pynvml is None:
                 logger.warning("GPU can't be found. Tracking GPU usage is skipped.")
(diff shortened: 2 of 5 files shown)


requirements/base.txt
```diff
@@ -10,3 +10,4 @@ scipy>=1.8
 bayesian-optimization>=1.2.0
 tensorboard>=2.11.0
 multiprocess
+pynvml
```
sungmanc: How about making it more specific by adding the version? and I think `requirements/dev.txt` is more proper location.

src/otx/cli/utils/experiment.py
```diff
@@ -0,0 +1,253 @@
...
+        resource_type (str, optional): Which resource to track. Available values are cpu, gpu or all now.
+                                        Defaults to "all".
+        gpu_ids (Optional[str]): GPU indices to record.
+    """
+
+    def __init__(self, resource_type: str = "all", gpu_ids: Optional[str] = None):
+        if resource_type == "all":
+            self._resource_type = AVAILABLE_RESOURCE_TYPE
+        else:
+            self._resource_type = [val for val in resource_type.split(",")]
```
sungmanc, line 41: Making proper error raise is a good option to users. (i.e. if users use `xpu` or something)

sungmanc (dismissed):
LGTM, left minor comments

src/otx/cli/utils/experiment.py
eunwoosh: You're right. But it seems that line 99 is better place to raise the error. I'll update it.

requirements/base.txt
eunwoosh: IHMO, it seems that using more robust version is better to avoid a conflict with other packages. But it's better to set major version. I'll update it.
And in terms of `dev.txt`, because it's `otx train` argument, it should be placed in `base.txt`. If not, it raises an error in case where user tries to use this feature after installing otx using `pip install otx[classification]`

eunwoosh: version is updated at [22ad518](https://github.com/openvinotoolkit/training_extensions/pull/2500/commits/22ad518e43478297aafd7a5c9cf47d7adf1593c4)

src/otx/cli/utils/experiment.py
eunwoosh: updated at [22ad518](https://github.com/openvinotoolkit/training_extensions/pull/2500/commits/22ad518e43478297aafd7a5c9cf47d7adf1593c4)

sungmanc (approved):

src/otx/cli/utils/experiment.py
```diff
@@ -0,0 +1,253 @@
...
+
+        gpu_ids_arr = None
+        if gpu_ids is not None:
+            gpu_ids_arr = [int(idx) for idx in gpu_ids.split(",")]
+            gpu_ids_arr[0] = 0
```
jaegukhyun, line 46: Why do we need to specify the first element of gpu id array?

eunwoosh: In multi GPU training case, main process always uses `0` index GPU. it's applied to that code.

jaegukhyun: What if user gives "1,2" as a gpu_ids? Should gpu_ids_arr be [0,2]?

eunwoosh: Yes you're right. I tried to change it before, it's quite hard, because origin solution is setting `CUDA_VISIBLE_DEVICES` before executing python file. I'll try to find other ways to make it run as we expect. After than I can change this code also.

jaegukhyun: If this is intended behavior, then it is okay to leave it as it is.

jaegukhyun (approved):

PR #2503, opened 2023-09-18
Fix reg summary bug

yunchu:
### Summary

fix regression test result summary bug

### Checklist

### License

```python
```

### DIFF
--- a/tests/regression/classification/test_classification.py
+++ b/tests/regression/classification/test_classification.py
@@ -46,7 +46,6 @@ class TestRegressionMultiClassClassification:
 
     TRAIN_PARAMS = ["--learning_parameters.num_iters", REGRESSION_TEST_EPOCHS]
     templates = Registry(f"src/otx/algorithms/{REG_CATEGORY}").filter(task_type=TASK_TYPE.upper()).templates
-    templates = templates[:1]
     templates_ids = [template.model_template_id for template in templates]
 
     reg_cfg: RegressionTestConfig
--- a/tests/regression/classification/test_classification.py
+++ b/tests/regression/classification/test_classification.py
@@ -46,6 +46,7 @@ class TestRegressionMultiClassClassification:
 
     TRAIN_PARAMS = ["--learning_parameters.num_iters", REGRESSION_TEST_EPOCHS]
     templates = Registry(f"src/otx/algorithms/{REG_CATEGORY}").filter(task_type=TASK_TYPE.upper()).templates
+    templates = templates[:1]
     templates_ids = [template.model_template_id for template in templates]
 
     reg_cfg: RegressionTestConfig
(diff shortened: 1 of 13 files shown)


sungmanc (dismissed):

tests/regression/classification/test_classification.py
```diff
@@ -46,6 +46,7 @@ class TestRegressionMultiClassClassification:
 
     TRAIN_PARAMS = ["--learning_parameters.num_iters", REGRESSION_TEST_EPOCHS]
     templates = Registry(f"src/otx/algorithms/{REG_CATEGORY}").filter(task_type=TASK_TYPE.upper()).templates
+    templates = templates[:1]
```
eunwoosh: It seems that only one template will be tested. Is it intended?

tests/regression/summarize_test_results.py
```diff
@@ -211,18 +211,45 @@ def save_file(result_data: dict, output_path: str, file_name: str):
     df.to_csv(os.path.join(output_path, file_name))
 
 
+def merge_dict(target, source, overwrite=False):
```
eunwoosh: It seems that if `target` is `{1:2}` and `source` is `{3:4}`, `merge_dict(target, source)` doesn't change `target`. Is it ok?

tests/regression/classification/test_classification.py
yunchu: @eunwoosh thanks for your finding! i have to remove that line :)

tests/regression/summarize_test_results.py
yunchu: yes, it will do. i think it's better to rename the function of it due to it would not cover the general `dict` merge operation. it's only worthy to the regression result `dict` generated with different train types or label types.

tests/regression/summarize_test_results.py
```diff
@@ -211,18 +211,44 @@ def save_file(result_data: dict, output_path: str, file_name: str):
...
+        entity_path = os.path.join(input_path, entity)
+        if os.path.isdir(entity_path):
+            task_key, task = filter_task(entity_path)
+            results_list = []
+            for result_json in os.listdir(entity_path):
+                result_json_path = os.path.join(entity_path, result_json)
+                if os.path.isfile(result_json_path) and result_json_path.split(".")[-1] == "json":
+                    with open(result_json_path, "r") as f:
+                        results_list.append(json.load(f))
+            json_data = merge_results_list(results_list)
```
eunwoosh, line 251: I think that there is a possibility that `json_data` is used before initialization. How about to add `json_data = {}` at the front of this function?

eunwoosh (commented):
Sorry for not leaving comments at once. I left a minor comment. Please take a look.

tests/regression/summarize_test_results.py
yunchu: I don't think so. The `json_data` initialization and re-initialization is placed in the same scope with the usages as well as before the usage. so it cannot be used before initialization.

eunwoosh: Sorry for confusing. I mean `json_data` is used before initialization at line 254 or 257. If it can't happen or doesn't matter, please just resolve it :)

goodsong81 (approved):

eunwoosh (approved):

PR #2553, opened 2023-10-12
POC: xpu support

sovrasov:
### Summary

- More files are copied from mm*
- ATSSAssigner is temporally replaced with XPUATSSAssigner, which also works in CUDA env

### Checklist

### License

```python
```


harimkang:
I'm looking at this PR from a v2.0 design perspective.
Overall, it seems like a lot of the dataset, trainer (training api functions) stuff is changing (for mmcv).

We need to support more devices besides XPU.
How about we create something like a `DeviceManager` and collect functions like is_xpu_available, is_cuda_available, is_hpu_available (+...), and functions for supporting device that are agnostic to frameworks? 

(Of course, there will be things that each framework will need to handle depending on the device, but I think the DeviceManager can prevent unnecessary code duplication up front as much as possible. The way it is now, we have to check is_{device}_available with an if statement every time we add a device.)

We have to consider that anomalib is using lightning. Of course, it is expected that lightning already supports many devices, but it would be worth investigating, and it seems like something like this could be done in mmengine (mmcv 2.X).

Actually, for 1.X, the current PR's suggestion might be a good way to go, but I think we need a more general device management design as more frameworks might be added.
**If have any ideas or suggestions, please let me know**, I think I can do a PoC on 2.0 using this code as a reference.

@sovrasov @kprokofi I think your experience can help me a lot. :)

sovrasov:
Now every mm* component has it's own training launcher & model wrapper, so we have to manage devices right in those custom training launchers. In fact, we rely on built-in mm* device management. If this is improved somehow in mmcv 2.0, this can be adjusted on our side as well. Perhaps, a device manager could be created, but first we need to review all the cases including HPU.

sovrasov:
For further considerations it's wroth mentioning that device support consists of 3 basic components:
1. We need to move all the input tensors to the target device -> it's done by wrapping the target model
2. We need to move model's weights to the target device
3. We may need to call a special operator to initialize device-specific optimizer patching (usually that operator requires model+optimizer)

eunwoosh:
I think all of us agree with that we need to make better design. Because Harim already experienced using mmcv 2.0, after we implement HPU in some degree, it's better to discuss about that together.

eunwoosh (approved):

docs/source/guide/explanation/algorithms/visual_prompting/index.rst
kprokofi, line 1: Can you update target branch and this one for latest develop? I think it is not related to XPU implementation as well as some other changes

kprokofi:
> I think all of us agree with that we need to make better design. Because Harim already experienced using mmcv 2.0, after we implement HPU in some degree, it's better to discuss about that together.

I guess it is more like PoC -> let's not focus on architecture and code design on this step

docs/source/guide/explanation/algorithms/visual_prompting/index.rst
sovrasov: Done

src/otx/algorithms/classification/configs/deit_tiny/model.py
```diff
@@ -14,7 +14,7 @@
     backbone=dict(arch="deit-tiny", init_cfg=dict(type="Pretrained", checkpoint=ckpt_url, prefix="backbone")),
 )
 
-fp16 = dict(loss_scale=512.0)
+# fp16 = dict(loss_scale=512.0)
```
kprokofi: Let's make this change in train function, not in templates. Depending on the given device we can set fp16 in config to None (if it is XPU devices).

sovrasov: Done

kprokofi (approved):
LGTM, if tests pass

PR #2578, opened 2023-10-25
Split torch.anomalib into torch.lightning and torch.lightning.anomalib [v2]

harimkang:
### Summary

- Move adapters.torch.anomalib -> adapters.torch.lightning.anomalib
- Move common code used by lightning to torch.lightning for future scalability.
- Add lightning/engine.py, dataset.py, registry.py
- Add Unit-Tests for lightning/engine.py, dataset.py, registry.py

image

### How to test

`python -m pytest --cov=otx.v2.adapters.torch.lightning tests/v2/unit/adapters/torch/lightning --cov-report term-missing`
image

### Checklist

### License

```python
```


goodsong81 (approved):
LGTM.

samet-akcay:
it logically makes sense to separate lightning since anomalib uses lightning as the backend. One concern I'm having, though, is the level of depth when importing. For example, if one wants to import anomalib dataset, the following is needed:

```python
from otx.v2.adapters.torch.lightning.anomalib.dataset import AnomalibDataset
```

It is getting a bit verbose. I think one of the agenda in one of the forums would be to discuss relative imports to make this simpler.

samet-akcay:
Ideally, a developer should be able to quickly import it via 

```python
from otx.data import AnomalibDataset
```

or something similar

src/otx/v2/adapters/torch/lightning/anomalib/dataset.py
```diff
@@ -0,0 +1,167 @@
...
+from otx.v2.adapters.torch.lightning.dataset import LightningDataset
+from otx.v2.api.entities.task_type import TaskType, TrainType
+from otx.v2.api.entities.utils.dataset_utils import (
+    contains_anomalous_images,
+    split_local_global_dataset,
+)
+from otx.v2.api.utils.decorators import add_subset_dataloader
+from otx.v2.api.utils.type_utils import str_to_subset_type, str_to_task_type
+
+if TYPE_CHECKING:  # pragma: no cover
```
samet-akcay: can you explain the no-cover comment here

src/otx/v2/adapters/torch/lightning/anomalib/dataset.py
```diff
@@ -0,0 +1,167 @@
...
+    import albumentations as al
```
samet-akcay, line 24: Is `al` used in some other repos too? `albumentations` documentation recommends `A`. I don't like that either, but it would be something to consider to follow the standard

src/otx/v2/adapters/torch/lightning/anomalib/engine.py
```diff
@@ -0,0 +1,220 @@
...
+
+        Returns:
+            list[Logger] | Logger | None: Updated loggers.
+        """
+        self.trainer_config.pop("logger", None)
+        if logger is not None:
+            if isinstance(logger, list):
+                return logger
+            return [logger]
+        return [AnomalibTensorBoardLogger(save_dir=self.work_dir, name=target_path)]
```
samet-akcay, line 79: Can you explain why tensorboard logger is used as default?

samet-akcay (commented):
I'm not going to review the content of the anomalib code as it will significantly change after anomalib v1.0

src/otx/v2/adapters/torch/lightning/anomalib/dataset.py
harimkang: This is a comment to ignore that TYPE_CHECKING syntax when measuring unit test coverage. What do you think?

src/otx/v2/adapters/torch/lightning/anomalib/dataset.py
harimkang: Yes, let's fix this.

src/otx/v2/adapters/torch/lightning/anomalib/engine.py
harimkang: I've configured Anomalib to use the same logic it used in the old OTX v1. Shouldn't we have this?

samet-akcay: not sure :) It is not important though since we will work on this for anomalib v1.0

src/otx/v2/adapters/torch/lightning/anomalib/dataset.py
samet-akcay: Maybe we could add this to our `pyproject.toml` file. Otherwise, we will have to ignore this every time we have `if TYPE_CHECKING` statement. The following link might help 
https://github.com/nedbat/coveragepy/issues/831

Thoughts?

harimkang: I see, I'll fix this.

samet-akcay: Great, thanks!

harimkang: I'll make this work with a PR that removes black.

src/otx/v2/adapters/torch/lightning/anomalib/dataset.py
harimkang: This conflicts with RUFF, so I'm going to leave it as is for now.

samet-akcay (approved):
thanks, let's merge this and address the remaining stuff on the ruff pr

ashwinvaidya17 (approved):
Thanks for the efforts

Issue #2559, opened 2023-10-16
Mobilenet V3 Large -  Classification training with batch size=1 is not working


harimkang:
**Describe the bug**

When training the classification of Mobilenet-V3-Large model, an error occurs when the batch size is 1.
```
Traceback (most recent call last):
  File "/home/harimkan/workspace/otx-v1/venv/bin/otx", line 8, in <module>
    sys.exit(main())
  File "/home/harimkan/workspace/otx-v1/src/otx/cli/tools/cli.py", line 77, in main
    results = globals()[f"otx_{name}"]()
  File "/home/harimkan/workspace/otx-v1/src/otx/cli/tools/train.py", line 192, in main
    return train(exit_stack)
  File "/home/harimkan/workspace/otx-v1/src/otx/cli/tools/train.py", line 290, in train
    task.train(
  File "/home/harimkan/workspace/otx-v1/src/otx/algorithms/classification/task.py", line 216, in train
    results = self._train_model(dataset)
  File "/home/harimkan/workspace/otx-v1/src/otx/algorithms/classification/adapters/mmcls/task.py", line 420, in _train_model
    train_model(
  File "/home/harimkan/workspace/otx-v1/venv/lib/python3.10/site-packages/mmcls/apis/train.py", line 233, in train_model
    runner.run(data_loaders, cfg.workflow)
  File "/home/harimkan/workspace/otx-v1/venv/lib/python3.10/site-packages/mmcv/runner/epoch_based_runner.py", line 136, in run
    epoch_runner(data_loaders[i], **kwargs)
  File "/home/harimkan/workspace/otx-v1/src/otx/algorithms/common/adapters/mmcv/runner.py", line 81, in train
    self.run_iter(data_batch, train_mode=True, **kwargs)
  File "/home/harimkan/workspace/otx-v1/venv/lib/python3.10/site-packages/mmcv/runner/epoch_based_runner.py", line 31, in run_iter
    outputs = self.model.train_step(data_batch, self.optimizer,
  File "/home/harimkan/workspace/otx-v1/src/otx/algorithms/classification/adapters/mmcls/models/classifiers/mixin.py", line 29, in train_step
    return super().train_step(data, optimizer, **kwargs)
  File "/home/harimkan/workspace/otx-v1/src/otx/algorithms/classification/adapters/mmcls/models/classifiers/mixin.py", line 105, in train_step
    return super().train_step(data, optimizer, **kwargs)
[...]

PR #2593, opened 2023-10-31
Refactor Dataset API structure

harimkang:
### Summary

image

- Changed the` mmpretrain.dataset.py::Dataset` class by splitting its structure into `torch.dataset.py::TorchBaseDataset`, `torch.mmengine.dataset.py::MMXDataset`, and `torch.mmengine.mmpretrain.dataset.py::MMPretrainDataset`.
- TorchBaseDataset inherit from BaseDataset (otx.api.core.dataset).
- LightningDataset and MMXDataset inherit from TorchBaseDataset.
- MMPretrainDataset inherit from MMXDataset.

This is to reuse commonly available code.

### Checklist

### License

```python
```


src/otx/v2/adapters/torch/mmengine/dataset.py
```diff
@@ -0,0 +1,309 @@
...
+        self.base_dataset = self._get_sub_task_dataset()
+        self.initialize = True
+
+    def _get_sub_task_dataset(self) -> TorchDataset:
+        """Returns a TorchDataset object containing the data for the current sub-task.
+
+        Returns:
+            TorchDataset: A TorchDataset object containing the data for the current sub-task.
+        """
+
```
eugene123tw, line 101: AFAIU, this would be used as an interface? could we add `raise NotImplementedError`?

src/otx/v2/adapters/torch/dataset.py
```diff
@@ -0,0 +1,198 @@
...
+            subset (str): The subset to build the dataset for.
+            pipeline (Optional[Union[list, dict]]): The pipeline to use for the dataset.
+                Defaults to None.
+            config (Optional[Union[str, dict]]): The configuration to use for the dataset.
+                Defaults to None.
+
+        Returns:
+            Optional[TorchDataset]: The built TorchDataset object, or None if the dataset is empty.
+        """
+
```
eugene123tw, line 54: AFAIU, this would be used as an interface? could we add raise NotImplementedError?

src/otx/v2/adapters/torch/mmengine/dataset.py
```diff
@@ -0,0 +1,309 @@
...
+        self.set_datumaro_adapters()  # Set self.dataset_entity & self.label_schema
+        self.base_dataset = self._get_sub_task_dataset()
+        self.initialize = True
+
+    def _get_sub_task_dataset(self) -> TorchDataset:
+        """Returns a TorchDataset object containing the data for the current sub-task.
+
+        Returns:
+            TorchDataset: A TorchDataset object containing the data for the current sub-task.
+        """
```
eugene123tw, line 104: So we must implement this for each mmlab task, right? If it's so, could we add an NotImplemented Error?

eugene123tw (commented):
LGTM. Basically, each mmlab task inherits `MMXDataset` and implements `_get_sub_task_dataset`, right?

src/otx/v2/adapters/torch/dataset.py
harimkang: Thank you. I have fixed this.

src/otx/v2/adapters/torch/mmengine/dataset.py
harimkang: Thank you. I have fixed this.

harimkang:
> LGTM. Basically, each mmlab task inherits `MMXDataset` and implements `_get_sub_task_dataset`, right?

Yes for now, but if you can think of a better way, please share.

sungmanc (approved):
Thanks for the nice change :) It looks good to me

src/otx/v2/adapters/torch/dataset.py
```diff
@@ -0,0 +1,199 @@
...
+            batch_size=batch_size,
+            sampler=sampler,
+            num_workers=num_workers,
+            pin_memory=pin_memory,
+            shuffle=shuffle,
+            drop_last=drop_last,
+            persistent_workers=persistent_workers,
+            **kwargs,
+        )
+        sampler_cfg = sampler if isinstance(sampler, dict) else {"type": f"{sampler.__class__.__qualname__}"}
```
wonjuleee:
```suggestion
        if isinstance(sampler, dict):
    sampler_cfg = sampler
else:
    sampler_class_name = getattr(sampler.__class__, '__qualname__', None)
    sampler_cfg = {"type": sampler_class_name} if sampler_class_name else {"type": str(sampler.__class__)}
```

src/otx/v2/adapters/torch/dataset.py
```diff
@@ -0,0 +1,199 @@
...
+            "num_workers": num_workers,
+            "pin_memory": pin_memory,
+            "shuffle": shuffle,
+            "persistent_workers": persistent_workers,
+            "dataset": dataset_cfg,
+            **kwargs,
+        }
+        return dataloader
+
+    def subset_dataloader(
```
wonjuleee, line 137: maybe `build_subset_dataloader`?

src/otx/v2/adapters/torch/mmengine/dataset.py
```diff
@@ -0,0 +1,310 @@
...
+            sampler=sampler,
+            num_workers=num_workers,
+            pin_memory=pin_memory,
+            shuffle=shuffle,
+            drop_last=drop_last,
+            persistent_workers=persistent_workers,
+            **kwargs,
+        )
+
+    def subset_dataloader(
```
wonjuleee, line 221: here too.

src/otx/v2/adapters/torch/dataset.py
```diff
@@ -0,0 +1,199 @@
...
+            config (Optional[Union[str, dict]]): The configuration to use for the dataset.
+                Defaults to None.
+
+        Returns:
+            Optional[TorchDataset]: The built TorchDataset object, or None if the dataset is empty.
+        """
+        raise NotImplementedError
+
+
+    def build_dataloader(
```
wonjuleee: I am not sure why this is required. This is because, dataset is built per each subset level and each subset is going to generate its own dataloader. So, from my guess, I don't think this is used ever. So my suggestion is removing this and rename `subset_dataloader` to `build_dataloader`. What do you think?

tests/v2/unit/adapters/torch/lightning/test_dataset.py
```diff
@@ -73,14 +73,14 @@ def test_build_dataloader(self, mocker: MockerFixture) -> None:
             num_workers=2,
             pin_memory=False,
             shuffle=False,
-            drop_last=False,
+            drop_last=True,
```
wonjuleee: why this should be turned to True?

src/otx/v2/adapters/torch/dataset.py
harimkang: Thanks for the comment. I will fix this.

src/otx/v2/adapters/torch/dataset.py
harimkang: Hi @wonjuleee, I'll answer this with a comment on the subset_dataloader function above.

Currently, `otx.v2.api.core.dataset::BaseDataset` consists of the functions `build_dataset`, `build_dataloader`, and `subset_dataloader`.
`build_dataset` is a function that creates a torch dataset based on the input,
`build_dataloader` creates a torch Dataloader based on the torch Dataset.
`subset_dataloader` is a convenience function that calls build_dataset and build_dataloader internally.
Using the subset_dataloader function and a decorator function called add_subset_dataloader, we can provide a call to .train_dataloader(), val_dataloader(), test_dataloader() from this dataset object, which behaves internally as subset_dataloader(subset="train").

We might consider renaming subset_dataloader to build_subset_dataloader. @samet-akcay Any thoughts?

tests/v2/unit/adapters/torch/lightning/test_dataset.py
harimkang: I think this should only apply to classification MV3 model, but it looks like I made a mistake. I'll revert. For reference, the issue on the classification side is https://github.com/openvinotoolkit/training_extensions/issues/2559, which has been fixed in develop but not yet reflected in feature/v2.

Thanks for the detailed review!

src/otx/v2/adapters/torch/dataset.py
vinnamkim: It's a common sense spread out widely for those who use `torch.data.Dataset` and `torch.data.DataLoader` that `torch.data.Dataset` includes only a single subset and there exists a 1:1 corresponded `torch.data.DataLoader` to it. In this view, `build_dataset` implies `build_subset_of_dataset` and `build_dataloader` does too for many practitioners. For me, this seems some duplications.

src/otx/v2/adapters/torch/dataset.py
```diff
@@ -0,0 +1,210 @@
...
+
+from otx.v2.api.core.dataset import BaseDataset
+from otx.v2.api.utils import set_tuple_constructor
+from otx.v2.api.utils.decorators import add_subset_dataloader
+
+SUBSET_LIST = ["train", "val", "test"]
+
+
+@add_subset_dataloader(SUBSET_LIST)
+class TorchBaseDataset(BaseDataset):
```
vinnamkim: It seems strange to me that `Dataset` builds `Dataset` (itself?) and `DataLoader`. Pytorch Lightning's `DataModule` design looks much more comprehensible to users https://lightning.ai/docs/pytorch/stable/data/datamodule.html.

Since we decided to go with Pytorch Lightning based framework, why not introduce the `DataModule` concept here?

src/otx/v2/adapters/torch/dataset.py
harimkang: @wonjuleee @vinnamkim 
Yes, that's right, I agree.
`build_dataset` returns a torch Dataset with a subset as input,
`build_dataloader` returns a torch Dataloader with torch Dataset as input.
`subset_dataloader` returns a torch Datalaoder with subset as input.

We can use 
```
train_dataset = dataset.build_dataset(subset="train")
train_dataloader = dataset.build_dataloader(dataset=train_dataset)
```
but If we have a subset_dataloader function, then we can also get a dataloader in the same way as
```
train_dataloader = dataset.train_dataloader()
# OR train_dataloader = dataset.subset_dataloader(subset="train")
```
and so on.

It seems redundant, but I think it would be a nice function for users, what do you think?

src/otx/v2/adapters/torch/dataset.py
harimkang: Yes, if we change to a lightning based framework I agree. but for now i'm still not sure about the decision.

Anyway, @samet-akcay As we feared, the name Dataset seems to cause a lot of confusion for developers.
Maybe we should rename it to something like DatasetBuilder. What do you think?

jaegukhyun: DataModule seems fine

src/otx/v2/adapters/torch/mmengine/dataset.py
```diff
@@ -0,0 +1,310 @@
...
+    from mmengine.registry import Registry
+    from torch.utils.data import DataLoader as TorchDataLoader
+    from torch.utils.data import Dataset as TorchDataset
+    from torch.utils.data import Sampler
+
+SUBSET_LIST = ["train", "val", "test"]
+
+
+@add_subset_dataloader(SUBSET_LIST)
+class MMXDataset(TorchBaseDataset):
```
vinnamkim: I don't know why there are many different `Dataset` for each framework. What's the reason? Is it for the input type transformation for the model developed by each framework?

src/otx/v2/adapters/torch/dataset.py
wonjuleee: I am still not sure. I think users may want to have build_dataloader for each subset, and hence subset_dataloader can bring confusions. What do you think @vinnamkim?

tests/v2/unit/adapters/torch/lightning/test_dataset.py
wonjuleee: Please attach the commit when you fix :)

src/otx/v2/adapters/torch/dataset.py
wonjuleee: Vote for `DataModule`.

src/otx/v2/adapters/torch/mmengine/dataset.py
harimkang: In the case of MMXDataset, yes, it's meant to build by converting inputs, and we try to keep the common code together as much as possible. And each of mmpretrain, mmdet, mmseg, etc. have their own overrides that are needed because they all use modules that turn a DatasetEntity into a Dataset that can be used in mmlab. I'm hoping that eliminating DatasetEntity will reduce the number of unnecessary classes.

src/otx/v2/adapters/torch/dataset.py
vinnamkim: It is same story with my comment below.

1. `dataset.build_dataset()` returns `dataset` but it is actually the subset of `dataset`.
2. `dataset.build_dataloader()` can take any `dataset` (not just for itself) but anyway return `dataloader`, so that `dataloader` can have no relation with the `dataset` which created itself.
3. On the other hand, `dataset.subset_dataloader()` can only produce `dataloader` which should produce a subset of `dataset`.

Let's get back to the step 2. Why do we need this kind of interface as a member function?

tests/v2/unit/adapters/torch/lightning/test_dataset.py
harimkang: Yes, sure. https://github.com/openvinotoolkit/training_extensions/pull/2593/commits/f2a40cb1824809b649c15af7ae7bb00a3dfaa980 here is fixed commit :)

src/otx/v2/adapters/torch/mmengine/dataset.py
vinnamkim: My suggestion for this is to move all that type conversions to `torch.data.DataLoader`'s `collate_fn` or `torch.data.Dataset`'s `transform`. I don't see why there should be inheritances for the conversion of `otx.api.entities.*` -> data entity for the model developed by the specific framework.

harimkang:
@wonjuleee @vinnamkim @sungmanc @jaegukhyun @samet-akcay 
Thanks for the comments, everyone. I think some of the comments are topics that should be discussed elsewhere, not in this PR. (Design aspects) We could probably have a separate call to discuss this (or use the OTX2.0 forum).

I think the design is a done deal for Phase1, and good suggestions are always welcome and I think we can improve through discussion. Currently, We're focusing on the goals of Phase 1 first. so please focus your review on comments about code quality or bugs. 

Once Again, Thanks for all the comments. :)

src/otx/v2/adapters/torch/mmengine/dataset.py
harimkang: Yes, I think that's a good suggestion but there are a few things to check. Currently, we utilize the registry that each mmlab library uses to build the dataset, so this could be a conflict from there, but It would be great to discuss and think about this a bit more.

goodsong81 (approved):
Regarding the design, I've given similar feedback that @wonjuleee and @vinnamkim say. I hope this could be enhanced in the next round of design refinement. Others look good to me in terms of your intension. Thanks!

jaegukhyun (approved):

eugene123tw (approved):

src/otx/v2/adapters/torch/dataset.py
```diff
@@ -0,0 +1,210 @@
...
+from torch.utils.data import Sampler
+
+from otx.v2.api.core.dataset import BaseDataset
+from otx.v2.api.utils import set_tuple_constructor
+from otx.v2.api.utils.decorators import add_subset_dataloader
+
+SUBSET_LIST = ["train", "val", "test"]
+
+
+@add_subset_dataloader(SUBSET_LIST)
```
djdameln, line 23: What's the motivation for using a decorator here instead of hard-coding the individual dataloader methods? It makes the class a bit cleaner but at the same time it might make it a bit hard to navigate the code.

src/otx/v2/adapters/torch/dataset.py
djdameln: I agree that `build_dataloader` feels redundant. I don't  think we need 3 different ways to obtain the dataloader of a given subset (in fact, I'm concerned it could cause confusion). To keep the terminology consistent I would be in favor of @wonjuleee's proposal to rename `subset_dataloader` to `build_dataloader` and remove the current `build_dataloader`. This way, both `build_dataset` and `build_dataloader` require passing the subset.

We could consider making `build_dataset` and `build_dataloader` private methods, to signal to the user that the only recommended way to obtain the dataloader of a given subset is the `<subset>_dataloader` method.

src/otx/v2/adapters/torch/dataset.py
djdameln: I agree that the behaviour of this class is very similar to a Lightning DataModule. The only problem I see with calling it this way is that we would then have a class that is called `adapters.torch.Lightning.DataModule` but which is not actually an instance of a Lightning DataModule. Not sure how problematic this would be

src/otx/v2/adapters/torch/dataset.py
harimkang: Yes, I agree. This was already feedback from the design proposal. There are pros and cons to this, but we decided the pros outweigh the cons at this point.

src/otx/v2/adapters/torch/dataset.py
harimkang: We'll continue the discussion and refactoring of this after this PR is merged.

src/otx/v2/adapters/torch/dataset.py
harimkang: Yes I think everyone's comments make sense. Thanks for the comment, I will proceed to replace `build_dataset` and `build_dataloader` with private functions.
And we'll discuss renamings for `subset_dataloader` a bit later. https://github.com/openvinotoolkit/training_extensions/pull/2593/commits/021e865281b9dbe6fcee9865bb0288f065e0a918

src/otx/v2/adapters/torch/dataset.py
samet-akcay: How about?
```suggestion
class BaseTorchDataset(BaseDataset):
```

src/otx/v2/adapters/torch/dataset.py
```diff
@@ -0,0 +1,210 @@
...
+
+    def _initialize(self) -> None:
+        self.set_datumaro_adapters()  # Set self.dataset_entity & self.label_schema
+        self.initialize = True
+
+    def _build_dataset(
+        self,
+        subset: str,
+        pipeline: list | dict | None = None,
+        config: str | dict | None = None,
```
samet-akcay, line 35: what exactly the difference between `pipeline` and `config`? It is a bit unclear and could be difficult to users to distinguish

src/otx/v2/adapters/torch/dataset.py
```diff
@@ -0,0 +1,210 @@
...
+        Args:
+            dataset (Optional[TorchDataset]): The dataset to load.
+            batch_size (int): The batch size to use.
+            num_workers (int): The number of worker processes to use for data loading.
+            shuffle (bool): Whether to shuffle the data.
+            pin_memory (bool): Whether to pin memory for faster GPU transfer.
+            drop_last (bool): Whether to drop the last incomplete batch.
+            sampler (Optional[Union[Sampler, Iterable, Dict]]): The sampler to use for data loading.
+            persistent_workers (bool): Whether to keep the worker processes alive between iterations.
+            **kwargs: Additional arguments to pass to the DataLoader constructor.
```
samet-akcay: What could these `kwargs` be? Can you add more description here about what we support here?

src/otx/v2/adapters/torch/dataset.py
```diff
@@ -0,0 +1,210 @@
...
+            batch_size=batch_size,
+            sampler=sampler,
+            num_workers=num_workers,
+            pin_memory=pin_memory,
+            shuffle=shuffle,
+            drop_last=drop_last,
+            persistent_workers=persistent_workers,
+            **kwargs,
+        )
+        sampler_cfg: dict | list | None = None
```
samet-akcay: it's a minor and possibly an opinionated comment, but `sampler_cfg` and `dataset_cfg` are the only variables that uses an abbreviation (cfg) as variable naming. The rest of the code does not use any abbreviation. It might be an idea to rename them to `sampler_config` and `dataset_config` in terms of consistency

src/otx/v2/adapters/torch/dataset.py
```diff
@@ -0,0 +1,210 @@
...
+        num_workers: int | None = None,
+        config: str | dict | None = None,
+        shuffle: bool = True,
+        pin_memory: bool = False,
+        drop_last: bool = False,
+        sampler: Sampler | (Iterable | dict) | None = None,
+        persistent_workers: bool = False,
+        **kwargs,
+    ) -> TorchDataLoader:
+        r"""Torch Based Dataset.subset_dataloader.
```
samet-akcay, line 151: Can you add a bit more description? Without looking at the code, the description is not sufficient to understand what it does.

src/otx/v2/adapters/torch/dataset.py
```diff
@@ -0,0 +1,210 @@
...
+            drop_last (bool, optional): value for whether to drop the last data when the batch is not divided up.
+                Defaults to False.
+            sampler (Optional[Union[Sampler, Iterable, Dict]], optional): Defines the strategy to draw
+                samples from the dataset. Can be any ``Iterable`` with ``__len__``
+                implemented. If specified, :attr:`shuffle` must not be specified.. Defaults to None.
+            persistent_workers (bool, optional): If ``True``, the data loader will not shutdown
+                the worker processes after a dataset has been consumed once. This allows to
+                maintain the workers `Dataset` instances alive. Defaults to False.
+            **kwargs (Any): Additional arguments to pass to the DataLoader constructor.
+
```
samet-akcay, line 186: Also would it be possible to add an `Example` section here?

src/otx/v2/adapters/torch/dataset.py
```diff
@@ -0,0 +1,210 @@
...
+        subset_pipeline = pipeline
+        if isinstance(subset_pipeline, dict):
+            subset_pipeline = subset_pipeline[subset]
+        subset_dataset = self._build_dataset(subset=subset, pipeline=subset_pipeline, config=_config)
+        if batch_size is None:
+            batch_size = _config.get("batch_size", 1)
+        if num_workers is None:
+            num_workers = _config.get("num_workers", 0)
+
+        # kwargs conflict
```
samet-akcay: What does this mean?

src/otx/v2/adapters/torch/mmengine/dataset.py
```diff
@@ -0,0 +1,310 @@
...
+                to be used for validation. Defaults to None.
+            test_data_roots (Optional[str], optional): The root address of the dataset
+                to be used for testing. Defaults to None.
+            test_ann_files (Optional[str], optional): Location of the annotation file for the dataset
+                to be used for testing. Defaults to None.
+            unlabeled_data_roots (Optional[str], optional): The root address of the unlabeled dataset
+                to be used for training. Defaults to None.
+            unlabeled_file_list (Optional[str], optional): The file where the list of unlabeled images is declared.
+                Defaults to None.
+            data_format (Optional[str], optional): The format of the dataset. Defaults to None.
```
samet-akcay, line 72: An example section here would be useful maybe?

src/otx/v2/adapters/torch/mmengine/dataset.py
```diff
@@ -0,0 +1,310 @@
...
+            return dataset
+
+        # Config Setting
+        if isinstance(config, str):
+            _config = Config.fromfile(filename=config)
+        elif isinstance(config, dict):
+            _config = Config(cfg_dict=config)
+        else:
+            _config = config
+        # Case with Config
```
samet-akcay: Can you elaborate these comments for future reference?

src/otx/v2/adapters/torch/mmengine/dataset.py
```diff
@@ -0,0 +1,310 @@
...
+        init_config = dataset_config.copy()
+        dataset_config["otx_dataset"] = otx_dataset
+        dataset_config["labels"] = labels
+        if pipeline is not None:
+            dataset_config["pipeline"] = pipeline
+        dataset_config.pop("data_roots", None)
+        dataset_config.pop("ann_files", None)
+        dataset_config.pop("file_list", None)
+        dataset_config["_scope_"] = self.scope
+        # Valid inputs
```
samet-akcay: Same as above. What should we understand from this comment when we read this 6 months later

samet-akcay (changes requested):
Thanks @harimkang, some minor comments..

src/otx/v2/adapters/torch/dataset.py
harimkang: Thanks for the suggestion. renamed it.

src/otx/v2/adapters/torch/dataset.py
harimkang: pipeline refers to dataset-transforms only, while config can hold the overall config of the dataset.
I think we need to modify pipeline to a word like transform, as Sungman suggested.
However, the rename to transform also involves AutoRunner and CLI, so I think it should be done in a different PR.

src/otx/v2/adapters/torch/dataset.py
harimkang: This can provide additional input to the torch dataloader in addition to the arguments we provide. Corrected description.

src/otx/v2/adapters/torch/dataset.py
harimkang: Thank you. I've renamed those words.

src/otx/v2/adapters/torch/dataset.py
harimkang: I've added a more specific description for this.

src/otx/v2/adapters/torch/dataset.py
harimkang: I also added an Example section.

src/otx/v2/adapters/torch/dataset.py
harimkang: This is an unnecessary comment. Cleared.

src/otx/v2/adapters/torch/mmengine/dataset.py
harimkang: This is an unnecessary comment. Cleared.

src/otx/v2/adapters/torch/mmengine/dataset.py
harimkang: This is an unnecessary comment. Cleared.

src/otx/v2/adapters/torch/mmengine/dataset.py
harimkang: I added an Example section. :)

src/otx/v2/adapters/torch/dataset.py
samet-akcay: we could discuss it during the next phase, but I agree with @djdameln. I feel that `DataModule` would be confusing as `DataModule` is known to be a lightning component. Having our own `DataModule` in Lightning adapters might potentially confuse users. 

Based on this argument, `Dataset` might also be confusing as torch already has this :) 

Anyway, leaving these here for a future reference. When the time comes, we could discuss this again.

src/otx/v2/adapters/torch/dataset.py
samet-akcay: Yes, I agree that `pipeline` should be renamed to `transforms` in another PR

samet-akcay (approved):
lgtm for the first phase. thanks

PR #2599, opened 2023-11-03
Fix issues in mmengine adapter & CLI side

harimkang:
### Summary

Thanks to Jaeguk, we found a bug that not update dataset.pipeline properly.

- Modify `_build_dataset` to accept config input as dict | None // pipeline input as list | None
- Enable pipeline to update to default only if config.dataset.pipeline does not exist
- Fixed an issue dropping pipelines from the CLI

Thanks to Jaeguk once more, we found a bug that not update mmengine.Runner scope properly.

- Add one line for setting default scope to Runner.default_config

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/v2/adapters/torch/dataset.py
+++ b/src/otx/v2/adapters/torch/dataset.py
@@ -49,6 +49,11 @@ class BaseTorchDataset(BaseDataset):
         Example:
         >>> dataset._build_dataet(subset="train")
         torch.utils.data.Dataset()
+        >>> dataset._build_dataset(
+            subset="train",
+            pipeline=[dict(type="Resize", scale=[224, 224])],
+        )
+        Dataset with Resize pipeline
         """
         raise NotImplementedError
 
@@ -162,7 +167,14 @@ class BaseTorchDataset(BaseDataset):
 
         Args:
             subset (str): Enter an available subset of that dataset.
-            pipeline (dict[str, list] | list | None, optional):
+            pipeline (dict[str, list] | list | None, optional): This can take a dict or a list.
+                Case with Dict: In the Semi-SL case,
+                    may want to apply a different pipeline between labeled and unlabeled.
+                    then we can give a dictionary of the form below.
+                    pipeline = {
+                        "train": [...],
+                        "unlabeled": [...]
+                    }
                 Dataset Pipeline. Defaults to None.
             batch_size (int | None, optional): How many samples per batch to load. Defaults to None.
             num_workers (int | None, optional): How many subprocesses to use for data loading.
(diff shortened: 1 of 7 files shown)


src/otx/v2/adapters/torch/dataset.py
```diff
@@ -31,20 +31,20 @@ def _initialize(self) -> None:
...
-        config: str | dict | None = None,
+        pipeline: list | None = None,
+        config: dict | None = None,
     ) -> TorchDataset | None:
         """Builds a TorchDataset object for the given subset using the specified pipeline and configuration.
 
         Args:
             subset (str): The subset to build the dataset for.
-            pipeline (Optional[Union[list, dict]]): The pipeline to use for the dataset.
+            pipeline (list | None, optional): The pipeline to use for the dataset.
```
goodsong81, line 41: Isn't this a duplicate? I'm not sure about the convention, but it would be better, IMHO, to be either of

- pipeline (list, optional)
- pipeline (list | None)

Please ignore if it's well accepted convention. :)

harimkang: I'm not sure of the exact convention, but I understand the meaning of optional to be whether it has a default value or not.

harimkang: I'll have to look into the convention a bit more and try to make it consistent across the entire code (in the near future..?)

jaegukhyun (approved):

goodsong81 (approved):

sungmanc (approved):

src/otx/v2/adapters/torch/dataset.py
samet-akcay, line 41: Can you explain why `pipeline` cannot be `dict` but only a `list`? I have the feeling that we do need to re-visit this pipeline approach.

harimkang: Sure. The reason the pipeline used to be a dictionary was for more flexibility (for example, to get labeled and unlabeled together in semi-sl), but build_dataset doesn't need to get them both at the same time, it can get them as a list, and subset_dataloader's pipeline can gets them as dict[str, list], but `build_dataset` need just one pipeline. so subset_dataloader just passes it to `buil_dataset` as list (If it's a dictionary, it will pull a list and pass it). This meant that `build_dataset` was accepting both `dict | list` but not actually using them properly, which Jaeguk found and I fixed.

harimkang: subset_dataloader(..., pipeline: dict | list | None) -> Pick a list from a dictionary (when semi-sl) and pass it -> build_dataset(..., pipeline: list | None)

samet-akcay: Thanks for clariying this. Now, another ask then :). Would it be possible to add an `Example` section here, so that reader/user would easily get a clearer picture when they read the docstring?

harimkang: Sure, I'll add it.

samet-akcay: Thanks!

harimkang: I added https://github.com/openvinotoolkit/training_extensions/pull/2599/commits/3eaa8f3bb7446279b4136dc871cc2fe38777b96d

src/otx/v2/adapters/torch/dataset.py
```diff
@@ -31,24 +31,29 @@ def _initialize(self) -> None:
...
+            config (dict | None, optional): The configuration to use for the dataset.
                 Defaults to None.
 
         Returns:
-            Optional[TorchDataset]: The built TorchDataset object, or None if the dataset is empty.
+            TorchDataset | None: The built TorchDataset object, or None if the dataset is empty.
 
         Example:
         >>> dataset._build_dataet(subset="train")
         torch.utils.data.Dataset()
```
samet-akcay:
```suggestion
        Examples:
        >>> dataset._build_dataset(subset="train")
        torch.utils.data.Dataset()
```

src/otx/v2/adapters/torch/dataset.py
samet-akcay: Thanks, just a typo suggestion

samet-akcay (approved):
Lgtm, thanks

PR #2605, opened 2023-11-06
Refactor MMXEngine._update_config

harimkang:
### Summary

- Modify MMXEngine to not accept config parameter.
- Create DEFAULT_CONFIG as a variable
- Fix _update_config to return config
- The MMXEngine class no longer uses config files. (AutoRunner, the CLI still use Config files.)

### Checklist

### License

```python
```


src/otx/v2/adapters/torch/mmengine/engine.py
```diff
@@ -32,6 +32,39 @@
 
 logger = get_logger()
 MMENGINE_DTYPE = ("float16", "bfloat16", "float32", "float64")
+# NOTE: We need to discuss where to declare the default value.
+DEFAULT_CONFIG: dict[str, dict | list | str | int | float | None] = {
```
goodsong81: Just a question. Can't we just use `dict` or `dict[str, Any]`?
Looks too verbose.

goodsong81 (approved):
Changes look generally OK to me.
By the way, files and functions are too large in overall. Maybe we can keep them small in the next phase of refactoring.

src/otx/v2/adapters/torch/mmengine/engine.py
samet-akcay: I think we may not need an annotation here as the values are already assigned.

src/otx/v2/adapters/torch/mmengine/engine.py
```diff
@@ -44,131 +77,108 @@ class MMXEngine(Engine):
...
         update_check = not all(value is None for value in func_args.values()) or not all(
             value is None for value in kwargs.values()
         )
 
-        # Update Model & Dataloaders & Custom hooks
+        runner_config = Config({})
+        for key, value in kwargs.items():
+            if value is None:
+                continue
+            runner_config[key] = value
```
samet-akcay, line 194: is this always safe? Would there be cases that some unwanted keys passed to this dictionary? Or it does not matter?

src/otx/v2/adapters/torch/mmengine/utils/runner_config.py
```diff
@@ -4,146 +4,93 @@
...
+        arg_key="max_epochs",
+        positional_args={"max_epochs": 10},
+        config=Config({"max_epochs": 20}),
+    )
+    10
+    >>> get_value_from_config(
+        arg_key="max_epochs",
+        positional_args={},
+        config=Config({"max_epochs": 20}),
+    )
```
samet-akcay:
```suggestion
    >>> get_value_from_config(
    ...    arg_key="max_epochs",
    ...    positional_args={"max_epochs": 10},
    ...    config=Config({"max_epochs": 20}),
    ...)
    10
    >>> get_value_from_config(
    ...    arg_key="max_epochs",
    ...    positional_args={},
    ...    config=Config({"max_epochs": 20}),
    ...)
```

samet-akcay (changes requested):
thanks, just minor comments.

src/otx/v2/adapters/torch/mmengine/engine.py
```diff
@@ -44,131 +77,108 @@ class MMXEngine(Engine):
...
         Args:
             work_dir (Optional[Union[str, Path]], optional): The working directory for the engine. Defaults to None.
-            config (Optional[Union[dict, Config, str]], optional): The configuration for the engine. Defaults to None.
         """
         super().__init__(work_dir=work_dir)
         self.runner: Runner
         self.latest_model = {"model": None, "checkpoint": None}
         self.registry = MMEngineRegistry()
-        self._initial_config(config)
+        self.default_config = Config(DEFAULT_CONFIG)
```
wonjuleee: How about setting this as a class static variable instead of `self.default_config`? This may not be updated and fed into `get_value_from_config` as default.

src/otx/v2/adapters/torch/mmengine/engine.py
```diff
@@ -264,15 +274,15 @@ def train(
                 work_dir=str(target_folder),
                 experiment_name="otx_train",
                 cfg=Config({}),  # To prevent unnecessary dumps.
-                **self.config,
+                **config,
             )
         if checkpoint is not None:
             self.runner.load_checkpoint(checkpoint)
-        self.dumped_config = dump_lazy_config(config=self.config, scope=self.registry.name)
+        self.dumped_config = dump_lazy_config(config=config, scope=self.registry.name)
```
wonjuleee, line 359: when `dumped_config` is used? This looks like used and updated during export operation.

src/otx/v2/adapters/torch/mmengine/utils/runner_config.py
```diff
@@ -4,146 +4,93 @@
...
-    return config, kwargs
+    # Priority 1: Positional Args
+    result = positional_args.get(arg_key, None)
+    # Priority 2: Input Config Value
+    return config.get(arg_key, None) if result is None else result
 
 
-def update_val_test_config(
-    func_args: dict,
+def update_train_config(
```
wonjuleee: can I ask why this is declared here as a global method, while `_update_config` and `_update_eval_config` are defined at framework engine?

src/otx/v2/adapters/torch/mmengine/engine.py
harimkang: Thanks for the comment. I've changed the variable to declare it as a class static variable and use it, and also changed `get_value_from_config` to `MMXEngine._get_value_from_config`.

src/otx/v2/adapters/torch/mmengine/engine.py
harimkang: This is a temporary way of storing the model's config, which we'll fix later in a cleaner way.

src/otx/v2/adapters/torch/mmengine/utils/runner_config.py
harimkang: Yes, now that I think about it, your comment makes sense. I've modified this to the `MMXEngine._update_train_config` function. Thanks for the nice comment. :)

src/otx/v2/adapters/torch/mmengine/utils/runner_config.py
harimkang: Fixed this, thank you.

src/otx/v2/adapters/torch/mmengine/engine.py
harimkang: There is code afterward to filter out unwanted keys.
https://github.com/harimkang/training_extensions/blob/fcc8a8036173bc2f90d8a6caf73f9c717af6c802/src/otx/v2/adapters/torch/mmengine/engine.py#L251
This is probably a bit tricky, but this check the __init__ args of the Runner to filter it out one last time.

src/otx/v2/adapters/torch/mmengine/engine.py
harimkang: Any is now disabled by ruff, and I've also annotated this because mypy wants it.

src/otx/v2/adapters/torch/mmengine/engine.py
```diff
@@ -40,135 +73,190 @@ class MMXEngine(Engine):
...
-        removed_key = [config_key for config_key in self.config if config_key not in runner_arg_list]
+        removed_key = [config_key for config_key in runner_config if config_key not in runner_arg_list]
         if removed_key:
             msg = f"In Engine.config, remove {removed_key} that are unavailable to the Runner."
             logger.warning(msg, stacklevel=2)
             for config_key in removed_key:
-                self.config.pop(config_key)
-        return update_check
+                runner_config.pop(config_key)
+        return runner_config, update_check
```
jaegukhyun, line 259: I want to suggest other approach for this function. 
We now have self.default_config, func_args, kwargs. 
kwargs can be directly merge with self.default_config 
And func_args need some change to merge with self.default_config directly.
At this stage, I think only difference is randomness. So how about this kind of implementation?
`runner_config = deepcopy(self.default_config)`
 `_config_from_kwargs = Config(kwargs)`
 `runner_config.merge_from_dict(_config_from_kwargs)`
 `_config_from_func_args = Config(func_args)`
 `randomness = {`
 `    "seed": _config_from_func_args.pop("seed", runner_config.get("seed", 1234),` 
 `    "deterministic": _config_from_func_args.pop("deterministic", runner_config.get("determinisitc", False)`
 `}`
`_config_from_func_args.randomness = randomness`
`runner_config.merge_from_dict(_config_from_func_args)`

I might forget details, so my work cannot run properly. 
Anyway could you share your opinion for this kind of implementation?

src/otx/v2/adapters/torch/mmengine/engine.py
```diff
@@ -334,17 +422,17 @@ def validate(
...
-                **self.config,
+                **config,
             )
         elif update_check:
-            self.runner._val_dataloader = self.config["val_dataloader"]  # noqa: SLF001
-            self.runner._val_loop = self.config["val_cfg"]  # noqa: SLF001
-            self.runner._val_evaluator = self.config["val_evaluator"]  # noqa: SLF001
+            self.runner._val_dataloader = config["val_dataloader"]  # noqa: SLF001
+            self.runner._val_loop = config["val_cfg"]  # noqa: SLF001
+            self.runner._val_evaluator = config["val_evaluator"]  # noqa: SLF001
```
jaegukhyun, line 430: Could you explain why should we set this value? It seems they should be set when runner is created

src/otx/v2/adapters/torch/mmengine/engine.py
harimkang: Let me check this out.

src/otx/v2/adapters/torch/mmengine/engine.py
harimkang: This is for reusing the runner.  
If we reuse the runner, we only need to change those three variables in the runner to use the new val_dataloader.

src/otx/v2/adapters/torch/mmengine/engine.py
harimkang: I checked and our function (update_config) change a total of six variables in the args for the mmengine.Runner.
`seed`, `deterministic`, `optimizer`, `precision`, `max_iters`, `max_epochs`
We should pop all of these variables from func_args (_config_from_func_args), modify them appropriately, and then merge them. But that doesn't change the current code size or flow.

If you have a better idea, please let me know.

src/otx/v2/adapters/torch/mmengine/engine.py
sungmanc: Then, how about ignoring the option that forces to avoid `Any` ? I also have similar opinion with Songki. `dict | list | str | int | float | None` seems verbose and almost the same with `Any`

jaegukhyun (approved):

src/otx/v2/adapters/torch/mmengine/engine.py
```diff
@@ -40,135 +73,190 @@ class MMXEngine(Engine):
...
-        if config is not None:
-            if isinstance(config, str):
-                self.config = Config.fromfile(config)
-            elif isinstance(config, Config):
-                self.config = copy.deepcopy(config)
-            elif isinstance(config, dict):
-                self.config = Config(config)
-        else:
-            self.config = Config({})
+    def _get_value_from_config(
```
sungmanc, line 93:
```suggestion
    def _get_value_from_updated_config(
```

I want to make sure that this function will return the final value extracted from the `positional_args + default_config`

src/otx/v2/adapters/torch/mmengine/engine.py
```diff
@@ -40,135 +73,190 @@ class MMXEngine(Engine):
...
+        if max_epochs is not None:
+            config["train_cfg"]["by_epoch"] = True
+            config["train_cfg"]["max_epochs"] = max_epochs
+        elif max_iters is not None:
+            config["train_cfg"]["by_epoch"] = False
+            config["train_cfg"]["max_iters"] = max_iters
+        # Update Optimizer
+        if "optim_wrapper" not in config or config["optim_wrapper"] is None:
+            optimizer = self._get_value_from_config("optimizer", arguments)
+            if get_device() not in ("cuda", "gpu", "npu", "mlu"):
```
sungmanc, line 164: Is it correct to wrap the optimizer for the mixed precision? Is there no `cuda + fp16` case ?

harimkang: The opposite is true. The inside of that if statement targets the CPU.

src/otx/v2/adapters/torch/mmengine/engine.py
harimkang: I'm not sure if this renaming makes sense, wouldn't it be more confusing? Of course the word config is confusing, but updated_config doesn't seem to be the correct word.

src/otx/v2/adapters/torch/mmengine/engine.py
harimkang: We should probably use Config directly instead of this, I'll fix it.

harimkang: I modified this from a dict to a Config.

src/otx/v2/adapters/torch/mmengine/engine.py
wonjuleee: alright.

src/otx/v2/adapters/torch/mmengine/utils/runner_config.py
wonjuleee: thank you!

src/otx/v2/adapters/torch/mmengine/engine.py
wonjuleee: thank you.

wonjuleee (approved):
LGTM

samet-akcay (approved):

PR #2574, opened 2023-10-24
POC: HPU support

sungchul1:
### Checklist

- [ ] I have added unit tests to cover my changes.​
- [ ] I have added integration tests to cover my changes.​
- [ ] I have added e2e tests for validation.
- [ ] I have added the description of my changes into CHANGELOG in my target branch (e.g., [CHANGELOG](https://github.com/openvinotoolkit/training_extensions/blob/develop/CHANGELOG.md) in develop).​
- [ ] I have updated the documentation in my target branch accordingly (e.g., [documentation](https://github.com/openvinotoolkit/training_extensions/tree/develop/docs) in develop).
- [ ] I have [linked related issues](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

### License

- [ ] I submit _my code changes_ under the same [Apache License](https://github.com/openvinotoolkit/training_extensions/blob/develop/LICENSE) that covers the project.
      Feel free to contact the maintainers if that's a concern.
- [ ] I have updated the license header for each file (see an example below).

```python
# Copyright (C) 2023 Intel Corporation
# SPDX-License-Identifier: Apache-2.0
```

PR #2608, opened 2023-11-06
Object Detection with Gaudi2

kprokofi:
### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algorithms/common/adapters/mmcv/utils/_builder_build_data_parallel.py
+++ b/src/otx/algorithms/common/adapters/mmcv/utils/_builder_build_data_parallel.py
@@ -140,9 +140,10 @@ class XPUDataParallel(MMDataParallel):
 
 
 class HPUDataParallel(MMDataParallel):
-    def __init__(self, *args, enable_autocast: bool = False, **kwargs):
+    def __init__(self, *args, enable_autocast: bool = False, put_gt_on_device=True, **kwargs):
         super().__init__(*args, **kwargs)
         self.enable_autocast = enable_autocast
+        self.put_gt_on_device = put_gt_on_device
         self.src_device_obj = torch.device("hpu", self.device_ids[0])
 
     def scatter(self, inputs, kwargs, device_ids):
@@ -153,6 +154,8 @@ class HPUDataParallel(MMDataParallel):
                 for val in x:
                     if isinstance(val, dict):
                         for k in val:
+                            if not self.put_gt_on_device and k.startswith("gt_"):
+                                continue
                             if isinstance(val[k], torch.Tensor):
                                 val[k] = val[k].to(self.src_device_obj)
                             elif isinstance(val[k], list):
--- a/src/otx/algorithms/common/adapters/mmcv/utils/_builder_build_data_parallel.py
+++ b/src/otx/algorithms/common/adapters/mmcv/utils/_builder_build_data_parallel.py
@@ -12,7 +12,13 @@ import torch
 from mmcv import Config
 from mmcv.parallel import MMDataParallel, MMDistributedDataParallel
 
+<<<<<<< HEAD
 from otx.algorithms.common.utils import is_hpu_available, is_xpu_available
+=======
+from otx.algorithms.common.utils import is_xpu_available, is_hpu_available
+import habana_frameworks.torch as htorch
+from torch._utils import _get_device_index
+>>>>>>> added support for OD on habana
 
 
 @overload
@@ -138,12 +144,63 @@ class XPUDataParallel(MMDataParallel):
         with torch.autocast(device_type="xpu", dtype=torch.bfloat16, enabled=self.enable_autocast):
             return super().val_step(*inputs, **kwargs)
 
+<<<<<<< HEAD
 
 class HPUDataParallel(MMDataParallel):
     def __init__(self, *args, enable_autocast: bool = False, **kwargs):
         super().__init__(*args, **kwargs)
         self.enable_autocast = enable_autocast
         self.src_device_obj = torch.device("hpu", self.device_ids[0])
+=======
+def _get_available_device_type():
+    if torch.cuda.is_available():
+        return "cuda"
+    if hasattr(torch, "xpu") and torch.xpu.is_available():  # type: ignore[attr-defined]
+        return "xpu"
+    if is_hpu_available():
+        return "hpu"
+    # add more available device types here
+    return None
+
+
+def _get_device_attr(get_member):
+    device_type = _get_available_device_type()
+    if device_type and device_type.lower() == "cuda":
+        return get_member(torch.cuda)
+    if device_type and device_type.lower() == "xpu":
+        return get_member(torch.xpu)  # type: ignore[attr-defined]
+    if device_type and device_type.lower() == "hpu":
+        return get_member(htorch.hpu)
+    # add more available device types here
+    return None
+
+
+def _get_all_device_indices():
+    # all device index
+    return _get_device_attr(lambda m: list(range(m.device_count())))
+
+
+class HPUDataParallel(MMDataParallel):
+    def __init__(self, module, device_ids=None, output_device=None, dim=0, is_autocast=True):
+        super().__init__(module=module)
+        device_type = _get_available_device_type()
+        if device_type is None:
+            self.module = module
+            self.device_ids = []
+            return
+
+        if device_ids is None:
+            device_ids = _get_all_device_indices()
+
+        if output_device is None:
+            output_device = device_ids[0]
+
+        self.dim = dim
+        self.device_ids = [_get_device_index(x, True) for x in device_ids]
+        self.output_device = _get_device_index(output_device, True)
+        self.src_device_obj = torch.device(device_type, self.device_ids[0])
+        self.is_autocast = is_autocast
+>>>>>>> added support for OD on habana
 
     def scatter(self, inputs, kwargs, device_ids):
         inputs, kwargs = super().scatter(inputs, kwargs, [-1])
@@ -154,25 +211,42 @@ class HPUDataParallel(MMDataParallel):
                     if isinstance(val, dict):
                         for k in val:
                             if isinstance(val[k], torch.Tensor):
+<<<<<<< HEAD
                                 val[k] = val[k].to(self.src_device_obj)
                             elif isinstance(val[k], list):
                                 for i, item in enumerate(val[k]):
                                     if isinstance(item, torch.Tensor):
                                         val[k][i] = item.to(self.src_device_obj)
+=======
+                                val[k] = val[k].to(torch.device(f"hpu:{device_ids[0]}"))
+                            elif isinstance(val[k], list):
+                                for i, item in enumerate(val[k]):
+                                    if isinstance(item, torch.Tensor):
+                                        val[k][i] = item.to(torch.device(f"hpu:{device_ids[0]}"))
+>>>>>>> added support for OD on habana
 
         for x in kwargs:
             if isinstance(x, dict):
                 for k in x:
                     if isinstance(x[k], torch.Tensor):
+<<<<<<< HEAD
                         x[k] = x[k].to(f"hpu:{device_ids[0]}")
                     elif isinstance(x[k], list):
                         for i, item in enumerate(x[k]):
                             if isinstance(item, torch.Tensor):
                                 x[k][i] = item.to(self.src_device_obj)
+=======
+                        x[k] = x[k].to("hpu")
+                    elif isinstance(x[k], list):
+                        for i, item in enumerate(x[k]):
+                            if isinstance(item, torch.Tensor):
+                                x[k][i] = item.to(torch.device(f"hpu:{device_ids[0]}"))
+>>>>>>> added support for OD on habana
 
         return inputs, kwargs
 
     def forward(self, *inputs, **kwargs):
+<<<<<<< HEAD
         with torch.cuda.amp.autocast(dtype=torch.bfloat16, enabled=self.enable_autocast):
             return super().forward(*inputs, **kwargs)
 
@@ -182,4 +256,15 @@ class HPUDataParallel(MMDataParallel):
 
     def val_step(self, *inputs, **kwargs):
         with torch.cuda.amp.autocast(dtype=torch.bfloat16, enabled=self.enable_autocast):
+=======
+        with torch.autocast(device_type="hpu", dtype=torch.bfloat16, enabled=self.is_autocast):
+            return super().forward(*inputs, **kwargs)
+
+    def train_step(self, *inputs, **kwargs):
+        with torch.autocast(device_type="hpu", dtype=torch.bfloat16, enabled=self.is_autocast):
+            return super().train_step(*inputs, **kwargs)
+
+    def val_step(self, *inputs, **kwargs):
+        with torch.autocast(device_type="hpu", dtype=torch.bfloat16, enabled=self.is_autocast):
+>>>>>>> added support for OD on habana
             return super().val_step(*inputs, **kwargs)
--- a/src/otx/algorithms/common/adapters/mmcv/utils/_builder_build_data_parallel.py
+++ b/src/otx/algorithms/common/adapters/mmcv/utils/_builder_build_data_parallel.py
@@ -154,6 +154,8 @@ class HPUDataParallel(MMDataParallel):
                 for val in x:
                     if isinstance(val, dict):
                         for k in val:
+                            # don't put annotations on the HPU to proceed
+                            # post-processing on the CPU
                             if not self.put_gt_on_device and k.startswith("gt_"):
                                 continue
                             if isinstance(val[k], torch.Tensor):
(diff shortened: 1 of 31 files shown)


sungchul1:
@kprokofi I and @eunwoosh opened https://github.com/openvinotoolkit/training_extensions/pull/2574, could you update this PR after merging https://github.com/openvinotoolkit/training_extensions/pull/2574?

src/otx/algorithms/detection/adapters/mmdet/apis/train.py
```diff
@@ -198,6 +223,18 @@ def train_detector(model, dataset, cfg, distributed=False, validate=False, times
...
+        return cfg_optim
+
+    # Only SGD for OD and IS supported by now on HPU
+    cfg_optim["type"] = "SGD"
+    if "betas" in cfg_optim:
+        del cfg_optim["betas"]
+    return cfg_optim
+
+
 def monkey_patched_xpu_nms(ctx, bboxes, scores, iou_threshold, offset, score_threshold, max_num):
```
sungchul1: I think `xpu` in the function name can be deleted.

src/otx/algorithms/detection/adapters/mmdet/apis/train.py
```diff
@@ -137,6 +154,14 @@ def train_detector(model, dataset, cfg, distributed=False, validate=False, times
         model.train()
         model, optimizer = torch.xpu.optimize(model, optimizer=optimizer, dtype=dtype)
 
+    if cfg.device == "hpu":
+        NMSop.forward = monkey_patched_xpu_nms
+        RoIAlign.forward = monkey_patched_xpu_roi_align
```
sungchul1: `xpu` in these function names can be deleted.

src/otx/algorithms/common/adapters/mmcv/utils/_builder_build_data_parallel.py
```diff
@@ -64,7 +64,7 @@ def build_data_parallel(
         model = model.xpu()
         model = XPUDataParallel(model, device_ids=config.gpu_ids)
     elif is_hpu_available() and config.get("gpu_ids", []):
-        model = model.hpu()
+        model = model.to("hpu")
```
eunwoosh, line 67: It's just a question. Are they different?

src/otx/algorithms/common/adapters/mmcv/utils/_builder_build_data_parallel.py
```diff
@@ -153,6 +154,8 @@ def scatter(self, inputs, kwargs, device_ids):
                 for val in x:
                     if isinstance(val, dict):
                         for k in val:
+                            if not self.put_gt_on_device and k.startswith("gt_"):
```
eunwoosh, line 159: Could you add a comment explaning what it's for?

src/otx/algorithms/detection/adapters/mmdet/apis/train.py
```diff
@@ -119,11 +121,26 @@ def train_detector(model, dataset, cfg, distributed=False, validate=False, times
     elif cfg.device == "xpu":
         model = build_dp(model, cfg.device, device_ids=cfg.gpu_ids, enable_autocast=bool(fp16_cfg))
         model.to(f"xpu:{cfg.gpu_ids[0]}")
+    elif cfg.device == "hpu":
+        import habana_frameworks.torch.core as htcore
+        from habana_frameworks.torch.utils.library_loader import load_habana_module
```
eunwoosh: Could you remove code using `load_habana_module`? It isn't required according to https://docs.habana.ai/en/latest/Release_Notes/Release%20Notes%20v1.5.0.html?highlight=load_habana_module#id6

src/otx/algorithms/detection/adapters/mmdet/apis/train.py
```diff
@@ -119,11 +121,26 @@ def train_detector(model, dataset, cfg, distributed=False, validate=False, times
...
+
+        load_habana_module()
+        os.environ["PT_HPU_LAZY_MODE"] = "1"
```
eunwoosh: As far as I know, `PT_HPU_LAZY_MODE` is set as 1 in default setting. How about removing this code?

src/otx/algorithms/detection/adapters/mmdet/apis/train.py
```diff
@@ -119,11 +121,26 @@ def train_detector(model, dataset, cfg, distributed=False, validate=False, times
...
+        assert len(cfg.gpu_ids) == 1
+        model = build_dp(
+            model, cfg.device, device_ids=cfg.gpu_ids, dim=0, enable_autocast=bool(fp16_cfg), put_gt_on_device=False
+        )
+        model.to(model.src_device_obj)
+        htcore.mark_step()
+        model.zero_grad()
```
sungchul1: Could you explain why they are here? AFAIK, they can be in train/eval loop.

src/otx/algorithms/detection/adapters/mmdet/apis/train.py
```diff
@@ -119,11 +121,26 @@ def train_detector(model, dataset, cfg, distributed=False, validate=False, times
...
+        from habana_frameworks.torch.utils.library_loader import load_habana_module
+
+        load_habana_module()
+        os.environ["PT_HPU_LAZY_MODE"] = "1"
+        assert len(cfg.gpu_ids) == 1
+        model = build_dp(
+            model, cfg.device, device_ids=cfg.gpu_ids, dim=0, enable_autocast=bool(fp16_cfg), put_gt_on_device=False
+        )
+        model.to(model.src_device_obj)
+        htcore.mark_step()
```
eunwoosh: It's just a question. Is it necessary for `htcore.mark_step()` to exist at this point?

src/otx/algorithms/detection/adapters/mmdet/apis/train.py
sungchul1: It would be better to move here if possible.
https://github.com/openvinotoolkit/training_extensions/blob/575b4ec0740d1512af8826eb075085a4dbef0640/src/otx/algorithms/common/utils/__init__.py#L61-L62

src/otx/algorithms/detection/adapters/mmdet/apis/train.py
```diff
@@ -119,11 +121,26 @@ def train_detector(model, dataset, cfg, distributed=False, validate=False, times
...
+        import habana_frameworks.torch.core as htcore
+        from habana_frameworks.torch.utils.library_loader import load_habana_module
+
+        load_habana_module()
+        os.environ["PT_HPU_LAZY_MODE"] = "1"
+        assert len(cfg.gpu_ids) == 1
+        model = build_dp(
+            model, cfg.device, device_ids=cfg.gpu_ids, dim=0, enable_autocast=bool(fp16_cfg), put_gt_on_device=False
+        )
+        model.to(model.src_device_obj)
```
sungchul1: @eunwoosh Does it seem same with segmentation? If not, `model` can be `model.cuda()` in `build_dp`.

src/otx/algorithms/detection/adapters/mmdet/apis/train.py
```diff
@@ -137,6 +154,14 @@ def train_detector(model, dataset, cfg, distributed=False, validate=False, times
         model.train()
         model, optimizer = torch.xpu.optimize(model, optimizer=optimizer, dtype=dtype)
 
+    if cfg.device == "hpu":
+        NMSop.forward = monkey_patched_xpu_nms
+        RoIAlign.forward = monkey_patched_xpu_roi_align
+        # build runner
+        if cfg.device == "hpu":
+            if (new_type := "Fused" + cfg.optimizer.get("type", "SGD")) in HABANA_OPTIMIZERS:
+                cfg.optimizer["type"] = new_type
```
sungchul1: It should be set before building optimizer (L144).

src/otx/algorithms/detection/adapters/mmdet/evaluation/evaluator.py
```diff
@@ -94,12 +100,19 @@ def print_map_summary(  # pylint: disable=too-many-locals,too-many-branches
...
-        table_data.append(["mAP", "", "", "", f"{mean_ap[i]:.3f}", f"{np.mean(mious[i]):.3f}"])
+        table_ = (
+            ["mAP", "", "", "", f"{mean_ap[i]:.3f}", f"{np.mean(mious[i]):.3f}"]
+            if segmentation
+            else ["mAP", "", "", "", f"{mean_ap[i]:.3f}"]
+        )
+        table_data.append(table_)
         table = AsciiTable(table_data)
         table.inner_footing_row_border = True
+        time.sleep(0.1)  # prevent segmentation fault
```
eunwoosh: I think this code don't have to be excuted in a case where HPU device isn't used. How about change this code to be executed only in HPU case?

src/otx/algorithms/detection/adapters/mmdet/evaluation/evaluator.py
```diff
@@ -59,6 +61,7 @@ def print_map_summary(  # pylint: disable=too-many-locals,too-many-branches
     if scale_ranges is not None:
         assert len(scale_ranges) == num_scales
 
+    segmentation = "miou" in results
```
eunwoosh, line 65: It's just a question. Does `miou` make an error? What is this change for?

src/otx/algorithms/common/adapters/mmcv/utils/_builder_build_data_parallel.py
```diff
@@ -153,6 +154,8 @@ def scatter(self, inputs, kwargs, device_ids):
...
+                                continue
```
sungchul1, line 160: It is just to use post processing on CPU, right?

sungchul1 (commented):
I left comments. And I'm just curious if habana's OD models also use post-processing on CPU.

src/otx/algorithms/detection/adapters/mmdet/apis/train.py
eunwoosh: Because `cfg.device` is `hpu`, `model.cuda()` isn't executed. So this code is necessary, I think.

sungchul1: I double checked in segmentation, with `model.cuda()` on gpu migration, model's weights are automatically converted to `hpu`.
`cfg.device` is not important in this situation because it is just for using `HPUDataParallel` in `dp_factory`.

```python
# before model.cuda()
for i, (name, params) in enumerate(model.named_parameters()):
    print(name, params.device)
    if i > 4: break

>>> backbone.stem.conv1.conv.weight cpu
>>> backbone.stem.conv1.bn.weight cpu
>>> backbone.stem.conv1.bn.bias cpu
>>> backbone.stem.branch1.0.conv.weight cpu
>>> backbone.stem.branch1.0.bn.weight cpu

model.cuda()
for i, (name, params) in enumerate(model.named_parameters()):
    print(name, params.device)
    if i > 4: break 

>>> backbone.stem.conv1.conv.weight hpu:0
>>> backbone.stem.conv1.bn.weight hpu:0
>>> backbone.stem.conv1.bn.bias hpu:0
>>> backbone.stem.branch1.0.conv.weight hpu:0
>>> backbone.stem.branch1.0.bn.weight hpu:0
>>> backbone.stem.branch1.0.bn.bias hpu:0
```

eunwoosh: Below code is `bulid_dp` function in mmseg. As you can see, because device is `hpu`, `model.cuda()` isn't executed. Due to same reason, `build_dp` in mmdet doesn't execute `model.cuda()` in `build_dp` function. This is mmdet [code](https://github.com/open-mmlab/mmdetection/ddddddddblob/v2.28.2/mmdet/utils/util_distribution.py#L10).
image

sungchul1: Yes, you're right. So, I suggest `model.cuda()` as `build_dp` args.
As you said, all if conditions will be skipped.
```python
model = build_dp(model.cuda(), cfg.device, ...) # cfg.device="hpu"
```

eunwoosh: I don't think so, because some of users want to use on CPU.

eunwoosh: Do you mean code like below?
```
elif cfg.device == "hpu":
    model = build_dp(
            model.cuda(), cfg.device, device_ids=cfg.gpu_ids, dim=0, enable_autocast=bool(fp16_cfg), put_gt_on_device=False
        )
```
Then it should work well. I think it's design choice. Do you think it's better than code below?
```
elif cfg.device == "hpu":
    model = build_dp(
            model, cfg.device, device_ids=cfg.gpu_ids, dim=0, enable_autocast=bool(fp16_cfg), put_gt_on_device=False
        )
    model.to(model.src_device_obj)
```

sungchul1: Oh, it seems that we don't need both of them because `DataParallel` casts model's device ([link](https://github.com/pytorch/pytorch/blob/e9ebda29d87ce0916ab08c06ab26fd3766a870e5/torch/nn/parallel/data_parallel.py#L147-L148)).
I checked `model` was converted to `hpu` after `model = build_dp(model, cfg.device, ...)`.
```python
# before model = build_dp(model, cfg.device, ...) in apis/train.py
for i, (name, params) in enumerate(model.named_parameters()):
    print(name, params.device)
    if i == 2: break
>>> backbone.stem.conv1.conv.weight cpu
>>> backbone.stem.conv1.bn.weight cpu
>>> backbone.stem.conv1.bn.bias cpu

# before super().__init__(*args, **kwargs) in _builder_build_data_parallel.py
for i, (name, params) in enumerate(args[0].named_parameters()):
    print(name, params.device)
    if i == 2: break
>>> backbone.stem.conv1.conv.weight cpu
>>> backbone.stem.conv1.bn.weight cpu
>>> backbone.stem.conv1.bn.bias cpu

super().__init__(*args, **kwargs)
for i, (name, params) in enumerate(args[0].named_parameters()):
    print(name, params.device)
    if i == 2: break
>>> backbone.stem.conv1.conv.weight hpu:0
>>> backbone.stem.conv1.bn.weight hpu:0
>>> backbone.stem.conv1.bn.bias hpu:0
```

Let's discuss it later because we need more tests, or escalate it to habana team.

eunwoosh: oh ok then we don't have to call explicitly. That's good news.

src/otx/algorithms/detection/adapters/mmdet/apis/train.py
kprokofi: I just followed some code examples in Habana documentation. I don't think we need it either. I will remove this

src/otx/algorithms/detection/adapters/mmdet/evaluation/evaluator.py
kprokofi: No, it is just needed to recognize instance segmentation task and object detection. If "miou" in results, then we deal with segmentation task.

src/otx/algorithms/detection/adapters/mmdet/apis/train.py
kprokofi: I just followed documentation form Habana and they add this in tutorials

src/otx/algorithms/common/adapters/mmcv/utils/_builder_build_data_parallel.py
kprokofi: Since post-processing is computed on CPU -> we don't need to put annotations on HPU

src/otx/algorithms/common/adapters/mmcv/utils/_builder_build_data_parallel.py
kprokofi: model = model.hpu() throws an error

src/otx/algorithms/common/adapters/mmcv/utils/_builder_build_data_parallel.py
eunwoosh: Ah I mean that could you add some comment nearby code to make us easier to understand code later.

src/otx/algorithms/detection/adapters/mmdet/apis/train.py
eunwoosh: yes, I also saw each documentation has different `mark_step` usage. It seems that as Habana version increases, where it should be added is changed referring this [link](https://forum.habana.ai/t/when-to-use-htcode-mark-step/446/2).
I think `mark_step` after model initialization isn't necesaary. How about removing it?

src/otx/algorithms/detection/adapters/mmdet/evaluation/evaluator.py
eunwoosh: I see. It seems that some condition branch depending on `segmentation ` variable are added. Is it for removing unnecessary metric according to task?

src/otx/algorithms/common/adapters/mmcv/utils/_builder_build_data_parallel.py
kprokofi: Yes, we don't need to put annotations on HPU if we compute loss on CPU. 
YOLOX from Habana team do the same -> they put post-processing on CPU. But I think it is not a good way to boost a perfomance, but other thing didn't work for OD

sungchul1 (approved):
LGTM. Please check pre-commit.

eunwoosh (approved):

PR #2612, opened 2023-11-07
Unify logger usage

eunwoosh:
### Summary
Current OTX uses different logger per files. This PR unified it to use same OTX logger.
Changes this PR made are as below:

- Change part using other logging method to use OTX logging method.
- To avoid circular import error, logger file is moved to `otx/utils`.
- After config manager instance is made, `config_logger` is called in each cli files.
 

### Checklist

### License

```python
```


src/otx/algorithms/classification/adapters/mmcls/utils/builder.py
```diff
@@ -10,7 +10,7 @@
 from mmcv.runner import load_checkpoint
 from mmcv.utils import Config, ConfigDict, get_logger
 
-from otx.algorithms.common.utils.logger import LEVEL
+from otx.utils.logger import LEVEL
 
 logger = get_logger("mmcls")
```
jaegukhyun: Is it okay to use get_logger function like this?

eunwoosh: Thanks for finding. I missed it. I'll update.

eunwoosh: fixed at [4051d39](https://github.com/openvinotoolkit/training_extensions/pull/2612/commits/4051d395412953637667133bfea8f918151161c9)

jaegukhyun (approved):

src/otx/cli/tools/train.py
```diff
@@ -199,6 +200,7 @@ def train(exit_stack: Optional[ExitStack] = None):  # pylint: disable=too-many-b
     args, override_param = get_args()
 
     config_manager = ConfigManager(args, workspace_root=args.workspace, mode=mode)
+    config_logger(config_manager.output_path / "otx.log", "INFO")
```
sungmanc, line 203: How about adding the detailed path? If user conduct `train->eval->export->eval`, then it seems that user can't check the train log`. Also same opinion for the eval, explain, ...
```suggestion
    config_logger(config_manager.output_path / "otx_train.log", "INFO")
```

eunwoosh: log file is made in each output directory as below. I think it's unnecessary to have different name. How do you think about?
image

sungmanc: Ah, if so I don't have any concerns :)

sungmanc (approved):

PR #2614, opened 2023-11-07
Migrate default segmentation model to OTXv2 

eugene123tw:
### How to test

```
unit/cli/utils/test_install.py::test_mim_installation
unit/cli/utils/test_install.py::test_mim_installation
  /home/yuchunli/git/otx-2x/venv/lib/python3.10/site-packages/pip/_vendor/pkg_resources/__init__.py:2871: DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('zope')`.
  Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
    declare_namespace(pkg)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
============================================================================================== 636 passed, 1 xfailed, 882 warnings in 435.27s (0:07:15) ==============================================================================================
```

### Checklist

### License

```python
```


jaegukhyun:
Could you remove added datasets for test? It seems for debugging purpose, but it makes hard to look your PR. 
And also I highly recommend to cherry-pick your commit to feature/v2 branch since Harim's previous commit is merged to one commit.

src/otx/v2/adapters/torch/mmengine/engine.py
```diff
@@ -608,7 +607,12 @@ def export(
         data_preprocessor = self.dumped_config.get("model", {}).get("data_preprocessor", None)
         mean = data_preprocessor["mean"] if data_preprocessor is not None else [123.675, 116.28, 103.53]
         std = data_preprocessor["std"] if data_preprocessor is not None else [58.395, 57.12, 57.375]
-        to_rgb = data_preprocessor["to_rgb"] if data_preprocessor is not None else False
+        to_rgb = False
+        if data_preprocessor is not None:
+            for rgb_keyword in ("bgr_to_rgb", "to_rgb"):
+                if rgb_keyword in data_preprocessor:
+                    to_rgb = data_preprocessor[rgb_keyword]
+                    break
```
wonjuleee:
```suggestion                    
to_rgb = data_preprocessor.get("bgr_to_rgb", data_preprocessor.get("to_rgb", False))
```

src/otx/v2/adapters/torch/mmengine/mmseg/dataset.py
```diff
@@ -0,0 +1,252 @@
...
+        # build sampler
+        if sampler is None:
+            sampler = DATA_SAMPLERS.build(
+                {"type": "DefaultSampler", "shuffle": shuffle},
+                default_args={"dataset": dataset, "seed": seed},
+            )
+
+        init_fn = partial(worker_init_fn, num_workers=num_workers, rank=rank, seed=seed) if seed is not None else None
+        if digit_version(torch.__version__) >= digit_version("1.8.0"):
+            kwargs["persistent_workers"] = persistent_workers
```
wonjuleee, line 172: basically, good to have this. But aren't we going to enforce to install pytorch >= 1.13?

src/otx/v2/adapters/torch/mmengine/mmseg/dataset.py
```diff
@@ -0,0 +1,252 @@
...
+            subset (str): The subset to build the dataset for.
+            pipeline (list | None, optional): The pipeline to use for the dataset.
+                Defaults to None.
+            config (dict | None, optional): The configuration to use for the dataset.
+                Defaults to None.
+
+        Returns:
+            Optional[TorchDataset]: The built TorchDataset object, or None if the dataset is empty.
+        """
+        def check_and_convert_to_tuple(pipeline: list[dict] | None) -> list[dict] | None:
```
wonjuleee: don't we need this conversion for other frameworks when building dataset?

wonjuleee: If it is required only for mmseg, could you clarify what difference is between others?

src/otx/v2/adapters/torch/mmengine/mmseg/dataset.py
```diff
@@ -0,0 +1,252 @@
...
+
+        Returns:
+            The number of classes in the dataset.
+        """
+        # TODO (Eugene): add test cases
+        # CVS-124394
+        if not self.initialize:
+            self._initialize()
+        label_names = [lbs.name for lbs in self.label_schema.get_labels(include_empty=False)]
+        if "background" in label_names:
```
wonjuleee, line 235: `num_classes` is basically implemented in torch/dataset and other frameworks didn't take care of this. I think it could be handled when creating `label_schema` in mmseg data adapter. What do you think?

wonjuleee: don't we need to add background class in label_schema too?

src/otx/v2/adapters/torch/mmengine/mmseg/dataset.py
eugene123tw: I guess for now it's fine in case we wanna support 1.8.0 in the future.

src/otx/v2/adapters/torch/mmengine/mmseg/dataset.py
eugene123tw: Each mmX has its own data pipeline transformation, so even mmdet might have a `RandomCrop` supports list, mmseg have a `RandomCrop` only allows `crop_size` being tuple.

src/otx/v2/adapters/torch/mmengine/mmseg/dataset.py
eugene123tw: My concern is that the label schema or label entity should be independent and not specific to any particular task. IMO, label schema should be universally applicable and not tailored exclusively for segmentation.

src/otx/v2/adapters/torch/mmengine/mmseg/dataset.py
wonjuleee: got it.

src/otx/v2/adapters/torch/mmengine/mmseg/dataset.py
wonjuleee: If so, how about doing this at src/otx/v2/adapters/torch/mmengine/mmseg/modules/datasets/otx_datasets.py? Some data preparation before augmentation may be performed at her. What do you think?

src/otx/v2/adapters/torch/mmengine/mmseg/dataset.py
wonjuleee: What I suggest is implementing `get_label_schema` method in task specific adapter. Please refer to https://github.com/openvinotoolkit/training_extensions/blob/e6c7325c506e93dec7c1b6a5ab9e63bc80a9d50c/src/otx/v2/adapters/datumaro/adapter/classification_dataset_adapter.py#L76. We may be able to update a label schema according to tasks.

src/otx/v2/adapters/torch/mmengine/mmseg/engine.py
```diff
@@ -0,0 +1,168 @@
...
+        """
+        config, update_check = super()._update_config(func_args, **kwargs)
+        if getattr(config, "val_dataloader", None) and not hasattr(config.val_evaluator, "type"):
+            config.val_evaluator = self.evaluator
+            config.val_cfg = {"type": "ValLoop"}
+        if getattr(config, "test_dataloader", None) and not hasattr(config.test_evaluator, "type"):
+            config.test_evaluator = self.evaluator
+            config.test_cfg = {"type": "TestLoop"}
+        if hasattr(config, "visualizer") and config.visualizer.type not in VISUALIZERS:
+            config.visualizer = self.visualizer
```
wonjuleee: I have a similar comment on PR for mmdet in v2 https://github.com/openvinotoolkit/training_extensions/pull/2598/files#r1387937126. Isn't is possible to do this with adequate `func_args`?

src/otx/v2/adapters/torch/mmengine/mmseg/engine.py
```diff
@@ -0,0 +1,168 @@
...
+
+        inferencer = MMSegInferencer(
+            model=config,
+            weights=checkpoint,
+            device=device,
+        )
+
+        return inferencer(img, batch_size=batch_size, **kwargs)
+
+    def export(
```
wonjuleee, line 118: why do we have this?

src/otx/v2/adapters/torch/mmengine/mmseg/dataset.py
```diff
@@ -0,0 +1,252 @@
...
+        dataloader.configs = {
+            "batch_size": batch_size,
+            "sampler": sampler_cfg,
+            "num_workers": num_workers,
+            "collate_fn": {"type": "pseudo_collate"},
+            "pin_memory": pin_memory,
+            "shuffle": shuffle,
+            "dataset": dataset_cfg,
+        }
+        return dataloader
```
jaegukhyun: Could you explain why mmseg cannot use MMXDataset's build_dataloader?

src/otx/v2/adapters/torch/mmengine/mmseg/dataset.py
```diff
@@ -0,0 +1,252 @@
...
+            Returns:
+                list[dict] | None: The converted pipeline, or None if the input was None.
+            """
+            if pipeline is None:
+                return None
+            for step in pipeline:
+                for key, value in step.items():
+                    if isinstance(value, list):
+                        step[key] = tuple(value)
+            return pipeline
```
jaegukhyun: Could you explain example scenario that pipeline contains unsupported type?
If we set cli config and api's default pipeline properly, then I think this can be removed

tests/v2/integration/cli/test_cli.py
```diff
@@ -29,6 +29,7 @@ def test_otx_cli(task: str, tmp_dir_path: Path) -> None:
     """
     # 1. Training
     tmp_dir_path = tmp_dir_path / f"{task}_e2e_test"
+    deterministic = "False" if task == "segmentation" else "True"
```
wonjuleee: Is there any special reason to set False only for segmentation?

src/otx/v2/adapters/torch/mmengine/mmseg/dataset.py
```diff
@@ -0,0 +1,252 @@
...
+            The number of classes in the dataset.
+        """
+        # TODO (Eugene): add test cases
+        # CVS-124394
+        if not self.initialize:
+            self._initialize()
+        label_names = [lbs.name for lbs in self.label_schema.get_labels(include_empty=False)]
+        if "background" in label_names:
+            return len(label_names)
+        return len(label_names) + 1
```
jaegukhyun, line 237: This is because len(self.label_schema.get_labels(include_empty=False)) is not num_classes when segmentation task considers background. Right?

src/otx/v2/adapters/torch/mmengine/mmseg/modules/datasets/otx_datasets.py
```diff
@@ -0,0 +1,156 @@
...
+        """Callback function of LoadAnnotationFromOTXDataset."""
+        dataset_item = results.pop("dataset_item")  # Prevent unnessary deepcopy
+        labels = results["ann_info"]["labels"]
+
+        ann_info = get_annotation_mmseg_format(dataset_item, labels, self.use_otx_adapter)
+
+        results["gt_seg_map"] = ann_info["gt_seg_map"]
+        results["seg_fields"].append("gt_seg_map")
+
+        return results
```
jaegukhyun: How about moving this and get_annotation_mmseg_format to modules/datasets/pipeline?

tests/v2/integration/cli/test_cli.py
```diff
@@ -37,7 +38,7 @@ def test_otx_cli(task: str, tmp_dir_path: Path) -> None:
         "--data.val_data_roots", TASK_CONFIGURATION[task]["val_data_roots"],
         "--max_epochs", "1",
         "--seed", "1234",
-        "--deterministic", "True",
+        "--deterministic", deterministic,
```
jaegukhyun: I think we can set deterministic False for all tasks for a while

tests/v2/integration/test_helper.py
```diff
@@ -25,4 +25,11 @@
         "sample": "tests/assets/car_tree_bug/images/train/Slide6.PNG",
         "models": ["otx_tiny_vit"],
     },
+    "segmentation": {
+        "train_data_roots": "tests/assets/common_semantic_segmentation_dataset/train",
+        "val_data_roots": "tests/assets/common_semantic_segmentation_dataset/val",
+        "test_data_roots": "tests/assets/common_semantic_segmentation_dataset/val",
+        "sample": "tests/assets/common_semantic_segmentation_dataset/train/images/0001.png",
+        "models": ["ocr_lite_hrnet_s_mod2"],
```
jaegukhyun:
```suggestion
        "models": ["otx_lite_hrnet_s_mod2"],
```

tests/v2/integration/cli/test_cli.py
wonjuleee: quick question: isn't setting deterministic to True more safer than False?

jaegukhyun: For perspective of test, it is reasonable to set this True since test should produce same output without code changes.
However sometime deterministic raise numerical error. I met this situation in my task. I guess Eugene also met some numerical errors during this PR. I think we should set this value as True eventually, but it is okay to leave as it is in this phase.

src/otx/v2/adapters/torch/mmengine/mmseg/dataset.py
eugene123tw: default_collate does not work. Has to use pseudo_collate (mmengine also uses pseudo_collate as default)

src/otx/v2/adapters/torch/mmengine/mmseg/dataset.py
eugene123tw: yes, that's right.

tests/v2/integration/cli/test_cli.py
eugene123tw: I don't think it's possible at the moment. https://github.com/pytorch/pytorch/issues/107999 

`RuntimeError: upsample_bilinear2d_backward_out_cuda does not have a deterministic implementation, but you set \'torch.use_deterministic_algorithms(True)\'. You can turn off determinism just for this operation, or you can use the \'warn_only=True\' option, if \nthat\'s acceptable for your application. You can also file an issue at https://github.com/pytorch/pytorch/issues to help us prioritize adding deterministic support for this operation.\n'))`

tests/v2/integration/cli/test_cli.py
eugene123tw: upsample_bilinear2d_backward_out_cuda does not support deterministic
`RuntimeError: upsample_bilinear2d_backward_out_cuda does not have a deterministic implementation, but you set \'torch.use_deterministic_algorithms(True)\'. You can turn off determinism just for this operation, or you can use the \'warn_only=True\' option, if \nthat\'s acceptable for your application. You can also file an issue at https://github.com/pytorch/pytorch/issues to help us prioritize adding deterministic support for this operation.\n'))`

src/otx/v2/adapters/torch/mmengine/mmseg/engine.py
eugene123tw: Some arguments are task-specific. This part needs to be refactored as a whole.
```
task: str | None = "Segmentation",
codebase: str | None = "mmseg",
```

src/otx/v2/adapters/torch/mmengine/mmseg/dataset.py
```diff
@@ -0,0 +1,252 @@
...
+        # CVS-124394
+
+        if dataset is None:
+            return None
+        rank, _ = get_dist_info()
+
+        # mmengine build dataloader in runner where we build in dataset.
+        # So, I'm basically copying things from runner.build_dataloader to here
+        # which does not make sense. why do we not use dataloader_cfg and build dataloader
+        # in runner.build_dataloader?
```
goodsong81, line 160: That's what we've been talking about the location of `build_data_loader()`. In the next phase, hopefully this could be moved to the right place.

src/otx/v2/adapters/torch/mmengine/mmseg/engine.py
eugene123tw: your suggestion sounded good, but when I tried it out, it didn't quite do the trick. I don't see much point in sticking with the current `_update_config` design — it's kind of messy, and updating parameters isn't clear.

 Instead of tweaking the config here, why not just load a YAML file that spells out everything we need? I think @sungmanc already proposed refactoring this part.

src/otx/v2/adapters/torch/mmengine/mmseg/engine.py
```diff
@@ -0,0 +1,168 @@
...
+from otx.v2.api.utils.logger import get_logger
+
+if TYPE_CHECKING:
+    import numpy as np
+
+logger = get_logger()
+
+
+class MMSegEngine(MMXEngine):
+    """The MMPretrainEngine class is responsible for running inference on pre-trained models."""
```
goodsong81: Minor: MMPretrain -> MMSeg in this file and others
```suggestion
    """The MMSegEngine class is responsible for running inference on pre-trained models."""
```

src/otx/v2/adapters/torch/mmengine/mmseg/engine.py
```diff
@@ -0,0 +1,168 @@
...
+        """Initialize a new instance of the MMPretrainEngine class.
+
+        Args:
+            work_dir (Optional[Union[str, Path]], optional): The working directory for the engine. Defaults to None.
+            config (Optional[Union[Dict, Config, str]], optional): The configuration for the engine. Defaults to None.
+        """
+        super().__init__(work_dir=work_dir)
+        self.registry = MMSegmentationRegistry()
+        self.visualizer = {"name": "visualizer", "type": "SegLocalVisualizer"}
+        self.evaluator = {"type": "IoUMetric", "iou_metrics": ["mDice"]}
```
goodsong81: Nitpick: these are not instance but configurations, right?
`evaluator_config` might be better not to mislead.

src/otx/v2/adapters/torch/mmengine/mmseg/dataset.py
eugene123tw: When `config.yaml` gets dumped into the workspace, `OmegaConf` turns Python tuples into YAML list style - _the issue is actually not tuple-specific; the same problem occurs with any other type that is not supported by OmegaConf_. Because tuples were converted to lists during the dump, upon loading, they're treated as lists of values. Essentially, here's where we dump config.yaml: https://github.com/openvinotoolkit/training_extensions/blob/feature/v2/src/otx/v2/cli/utils/workspace.py#L91

We have a couple of options to fix this. One is to convert the list back to a tuple on our end. Alternatively, we can explore a workaround like the one mentioned here: https://github.com/omry/omegaconf/issues/392. However, based on the attached issue, it's not the most ideal solution IMO.

src/otx/v2/adapters/torch/mmengine/mmseg/dataset.py
eugene123tw: no problem I will move this function there.

src/otx/v2/adapters/torch/mmengine/mmseg/modules/datasets/otx_datasets.py
eugene123tw: sure

src/otx/v2/adapters/torch/mmengine/mmseg/dataset.py
eugene123tw: I will add in a TODO in `segmentation_dataset_adapter.py ` about thinking of adding a background label to the label schema for segmentation tasks down the road. My concern is that different segmentation frameworks might not work well with a one-size-fits-all background label.

src/otx/v2/adapters/torch/mmengine/mmseg/dataset.py
wonjuleee: thank you!

src/otx/v2/adapters/torch/mmengine/mmseg/dataset.py
wonjuleee: alright, please just add TODO tag for now. thanks.

src/otx/v2/adapters/torch/mmengine/mmseg/engine.py
wonjuleee: sounds great. let's merge as it is for now first and try to redesign configurations overall later.

tests/v2/integration/cli/test_cli.py
wonjuleee: thank you. let's keep thread in Jaeguk's comment.

pyproject.toml
```diff
@@ -226,6 +238,7 @@ ignore = [
     "ANN003",   # Missing type annotation for **kwargs
     # flake8-type-checking
     "TCH001",   # typing-only-first-party-import, Sometimes this causes an incorrect error.
+    "FIX002"    # line-contains-todo
```
samet-akcay, line 250:
```suggestion
    # flake8-fixme
    "FIX002"    # line-contains-todo
```

src/otx/v2/adapters/torch/mmengine/engine.py
```diff
@@ -627,3 +626,15 @@ def export(
         )
 
         return exporter.export()
+
+    def _update_codebase_config(self, codebase: str | None, task: str | None, deploy_config_dict: dict) -> None:
```
samet-akcay: i have seen this in mmdetection pr too. what does codebase refer to? mmx? is this mmx-specific funtion?

src/otx/v2/adapters/torch/mmengine/mmseg/dataset.py
```diff
@@ -0,0 +1,235 @@
...
+        val_ann_files: str | None = None,
+        test_data_roots: str | None = None,
+        test_ann_files: str | None = None,
+        unlabeled_data_roots: str | None = None,
+        unlabeled_file_list: str | None = None,
+        data_format: str | None = None,
+    ) -> None:
+        r"""MMSeg's Dataset class.
+
+        Args:
```
samet-akcay, line 80: can you make sure if all the annotations in the docstrings follow the convention

src/otx/v2/adapters/torch/mmengine/mmseg/dataset.py
```diff
@@ -0,0 +1,235 @@
...
+            task (Optional[Union[TaskType, str]], optional): The task type of the dataset want to load.
```
samet-akcay:
```suggestion
            task (TaskType | str, optional): The task type of the dataset want to load.
```

src/otx/v2/adapters/torch/mmengine/mmseg/dataset.py
```diff
@@ -0,0 +1,235 @@
...
+            pin_memory (bool): Whether to pin memory for faster GPU transfer.
+            drop_last (bool): Whether to drop the last incomplete batch.
+            sampler (Optional[Union[Sampler, Iterable, Dict]]): The sampler to use for data loading.
+            persistent_workers (bool): Whether to keep the worker processes alive between iterations.
+            **kwargs: Additional arguments to pass to the DataLoader constructor.
+
+        Returns:
+            Optional[TorchDataLoader]: The DataLoader for the given dataset.
+        """
+        # TODO (Eugene): almost identical as mmengine.runner. Need to rethink about the design.
```
samet-akcay, line 150: it's good to mention it here, We could hopefully refactor to a simpler form iin the next phase

src/otx/v2/adapters/torch/mmengine/mmseg/engine.py
```diff
@@ -0,0 +1,168 @@
...
+logger = get_logger()
+
+
+class MMSegEngine(MMXEngine):
+    """The MMPretrainEngine class is responsible for running inference on pre-trained models."""
+
+    def __init__(
+        self,
+        work_dir: str | Path | None = None,
+    ) -> None:
```
samet-akcay:
```suggestion
    def __init__(self, work_dir: str | Path | None = None) -> None:
```

src/otx/v2/adapters/torch/mmengine/mmseg/engine.py
```diff
@@ -0,0 +1,168 @@
...
+            config = model._config  # noqa: SLF001
+        elif isinstance(model, dict) and "_config" in model:
+            config = model["_config"]
+        config["test_dataloader"] = {"dataset": {"pipeline": pipeline}}
+
+        # Check if the model can use mmseg's inference api.
+        if isinstance(checkpoint, Path):
+            checkpoint = str(checkpoint)
+
+        inferencer = MMSegInferencer(
```
samet-akcay: this could fit into a single line?

src/otx/v2/adapters/torch/mmengine/mmseg/model.py
```diff
@@ -0,0 +1,172 @@
...
+    import torch
+
+from otx.v2.adapters.torch.mmengine.mmseg.registry import MODELS
+from otx.v2.adapters.torch.mmengine.modules.utils.config_utils import CustomConfig as Config
+from otx.v2.api.utils.importing import get_files_dict, get_otx_root_path
+from otx.v2.api.utils.logger import get_logger
+
+logger = get_logger()
+
+MODEL_CONFIG_PATH = Path(get_otx_root_path()) / "v2/configs/segmentation/models"
```
samet-akcay, line 25: we will need to double check this to ensure if it works for pypi too

src/otx/v2/adapters/torch/mmengine/mmseg/model.py
```diff
@@ -0,0 +1,172 @@
...
+def list_models(pattern: str | None = None) -> list[str]:
+    """Returns a list of available models for training.
+
+    Args:
+        pattern (Optional[str]): A string pattern to filter the list of available models. Defaults to None.
+
+    Returns:
+        List[str]: A sorted list of available models for pretraining.
+    """
+    model_list = []
```
samet-akcay:
```suggestion
    model_list: list[str] = []
```

src/otx/v2/adapters/torch/mmengine/mmseg/modules/datasets/__init__.py
```diff
@@ -0,0 +1,24 @@
+"""OTX Algorithms - Segmentation Dataset."""
+
+# Copyright (C) 2023 Intel Corporation
```
samet-akcay, line 3: maybe we coul use the shorter license header for consistency

src/otx/v2/adapters/torch/mmengine/mmseg/modules/datasets/otx_datasets.py
```diff
@@ -0,0 +1,125 @@
+"""Base MMDataset for Segmentation Task."""
+
+# Copyright (C) 2023 Intel Corporation
```
samet-akcay, line 3: same here, and below if there are any

src/otx/v2/adapters/torch/mmengine/mmseg/modules/datasets/otx_datasets.py
```diff
@@ -0,0 +1,125 @@
...
+        """Dataset class for OTX datasets.
+
+        Args:
+            otx_dataset (DatasetEntity): The OTX dataset to use.
+            labels (list[LabelEntity]): List of label entities.
+            empty_label (list | None, optional): Empty label. Defaults to None.
+            pipeline (list | None, optional): Data processing pipeline. Defaults to None.
+            **kwargs: Additional keyword arguments.
+
+        Attributes:
```
samet-akcay: we may not need this section

src/otx/v2/adapters/torch/mmengine/mmseg/modules/datasets/pipeline.py
```diff
@@ -0,0 +1,76 @@
...
+    dataset_item: DatasetItemEntity,
+    labels: list[LabelEntity],
+    use_otx_adapter: bool = True,
+) -> dict:
+    """Function to convert a OTX annotation to mmsegmentation format.
+
+    This is used both in the OTXDataset class defined in this file
+    as in the custom pipeline element 'LoadAnnotationFromOTXDataset'
+
+    :param dataset_item: DatasetItem for which to get annotations
```
samet-akcay: can you convert this docstring to google format?

src/otx/v2/adapters/torch/mmengine/mmseg/modules/datasets/pipeline.py
```diff
@@ -0,0 +1,76 @@
...
+        results['dataset_item']: dataset_item from which to load the annotation
+        results['ann_info']['label_list']: list of all labels in the project
+
+    """
+
+    def __init__(self, use_otx_adapter: bool = True) -> None:
+        """Initializes the pipeline element.
+
+        Args:
+            use_otx_adapter (bool): Whether to use the OTX adapter or not.
```
samet-akcay: this should be in the class docstring. init does not need a docstring. It is to prperly parse the socumentation via sphinx

src/otx/v2/adapters/torch/mmengine/mmseg/modules/heads/otx_fcn_head.py
```diff
@@ -0,0 +1,120 @@
...
+    def __init__(
+        self,
+        enable_aggregator: bool = False,
+        aggregator_min_channels: int | None = None,
+        aggregator_merge_norm: str | None = None,
+        aggregator_use_concat: bool = False,
+        *args,
+        **kwargs,
+    ) -> None:
+        """Initializes OTXFCNHead.
```
samet-akcay, line 32: docstring not needed here, already introduced in class docstring

src/otx/v2/adapters/torch/mmengine/mmseg/modules/models/backbones/litehrnet.py
```diff
@@ -0,0 +1,1606 @@
...
+
+# Copyright (C) 2023 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
+
+from __future__ import annotations
+
+import torch
+import torch.utils.checkpoint as cp
+from mmcv.cnn import (
```
samet-akcay: could this fit into a single line

src/otx/v2/adapters/torch/mmengine/mmseg/modules/models/backbones/litehrnet.py
```diff
@@ -0,0 +1,1606 @@
...
+    """Conditional channel weighting module."""
+
+    def __init__(
+        self,
+        in_channels: list[int],
+        stride: int,
+        reduce_ratio: int,
+        conv_cfg: dict | None = None,
+        norm_cfg: dict | None = None,
+        with_cp: bool = False,
```
samet-akcay, line 401: can we use a more explicit arg name. it was not clear what cp was. Had to go to the docstring

src/otx/v2/adapters/torch/mmengine/mmseg/modules/models/backbones/litehrnet.py
```diff
@@ -0,0 +1,1606 @@
...
+        dropout: float | None = None,
+        weighting_module_version: str = "v1",
+        neighbour_weighting: bool = False,
+        dw_ksize: int = 3,
```
samet-akcay, line 405: same here, more explicit arg name would be great.

src/otx/v2/adapters/torch/mmengine/mmseg/modules/models/backbones/litehrnet.py
```diff
@@ -0,0 +1,1606 @@
...
+
+    def __init__(
+        self,
+        in_channels: int,
+        stem_channels: int,
+        out_channels: int,
+        expand_ratio: int,
+        conv_cfg: dict | None = None,
+        norm_cfg: dict | None = None,
+        with_cp: bool = False,
```
samet-akcay, line 537: how about `use_checkpointing`?

src/otx/v2/adapters/torch/mmengine/mmseg/modules/models/backbones/litehrnet.py
```diff
@@ -0,0 +1,1606 @@
...
```
samet-akcay, line 714: same here

src/otx/v2/adapters/torch/mmengine/mmseg/modules/models/backbones/litehrnet.py
```diff
@@ -0,0 +1,1606 @@
...
+            out_channels=stem_channels,
+            kernel_size=3,
+            stride=strides[0],
+            padding=1,
+            conv_cfg=self.conv_cfg,
+            norm_cfg=self.norm_cfg,
+            act_cfg={"type": "ReLU"},
+        )
+
+        self.conv2 = None
```
samet-akcay, line 777: why this?

src/otx/v2/adapters/torch/mmengine/mmseg/modules/models/backbones/litehrnet.py
samet-akcay: or `with_checkpointing`?

src/otx/v2/adapters/torch/mmengine/mmseg/modules/models/utils/__init__.py
```diff
@@ -0,0 +1,23 @@
+"""Utils used for mmseg model."""
+# Copyright (C) 2023 Intel Corporation
```
samet-akcay, line 3:
```suggestion

# Copyright (C) 2023 Intel Corporation
```

src/otx/v2/adapters/torch/mmengine/mmseg/modules/models/utils/aggregator.py
```diff
@@ -0,0 +1,169 @@
+"""Aggregators."""
+# Copyright (C) 2023 Intel Corporation
```
samet-akcay, line 3:
```suggestion

# Copyright (C) 2023 Intel Corporation
```

src/otx/v2/adapters/torch/mmengine/mmseg/modules/models/utils/aggregator.py
```diff
@@ -0,0 +1,169 @@
...
+
+import torch
+import torch.nn.functional
+from mmcv.cnn import ConvModule, DepthwiseSeparableConvModule
+from torch import nn
+
+from otx.v2.adapters.torch.mmengine.mmseg.modules.models.utils.normalize import normalize
+
+
+# pylint: disable=invalid-name
```
samet-akcay: pylint not needed anymore

src/otx/v2/adapters/torch/mmengine/mmseg/modules/models/utils/aggregator.py
```diff
@@ -0,0 +1,169 @@
...
+            y = normalize(y, dim=2, p=2)
+            out = y.view(-1, c, h, w)
+
+        return out
+
+    def forward(self, x: torch.Tensor) -> list[torch.Tensor]:
+        """Forward."""
+        x = x[::-1]
+
+        y_list = []
```
samet-akcay: can you specify the type here too

src/otx/v2/adapters/torch/mmengine/mmseg/modules/models/utils/asymmetric_position_attention.py
```diff
@@ -0,0 +1,122 @@
+"""Asymmetric position attention module."""
+# Copyright (c) 2019 MendelXu
```
samet-akcay, line 3:
```suggestion

# Copyright (c) 2019 MendelXu
```

samet-akcay: can you check the rest of the files

src/otx/v2/adapters/torch/mmengine/mmseg/modules/models/utils/asymmetric_position_attention.py
```diff
@@ -0,0 +1,122 @@
...
+
+import torch
+import torch.nn.functional
+from mmcv.cnn import ConvModule
+from torch import nn
+
+from .psp_layer import PSPModule
+
+
+# pylint: disable=too-many-instance-attributes
```
samet-akcay:
```suggestion
```

src/otx/v2/adapters/torch/mmengine/mmseg/modules/models/utils/asymmetric_position_attention.py
```diff
@@ -0,0 +1,122 @@
...
+
+    Reference: https://github.com/MendelXu/ANN.
+    """
+
+    def __init__(
+        self,
+        in_channels: int,
+        key_channels: int,
+        value_channels: int | None = None,
+        psp_size: tuple[int, ...] | None = None,
```
samet-akcay, line 30: more explicit naming possible?

samet-akcay: or is it the name of the module/model?

src/otx/v2/adapters/torch/mmengine/mmseg/modules/models/utils/channel_shuffle.py
```diff
@@ -0,0 +1,36 @@
+"""Channel shuffle method."""
+# Copyright (c) 2018-2020 Open-MMLab.
```
samet-akcay:
```suggestion
"""Channel shuffle method."""

# Copyright (c) 2018-2020 Open-MMLab.
```

src/otx/v2/adapters/torch/mmengine/mmseg/modules/models/utils/local_attention.py
```diff
@@ -0,0 +1,89 @@
+"""Local attention module."""
+# Copyright (C) 2019-2021 Xiangtai Lee
```
samet-akcay:
```suggestion
"""Local attention module."""

# Copyright (C) 2019-2021 Xiangtai Lee
```

src/otx/v2/adapters/torch/mmengine/mmseg/modules/models/utils/normalize.py
```diff
@@ -0,0 +1,66 @@
+"""Normalization."""
+# Copyright (C) 2023 Intel Corporation
```
samet-akcay:
```suggestion
"""Normalization."""

# Copyright (C) 2023 Intel Corporation
```

src/otx/v2/adapters/torch/mmengine/mmseg/modules/models/utils/normalize.py
```diff
@@ -0,0 +1,66 @@
...
+import torch.nn.functional
+from torch import nn
+
+
+class OnnxLpNormalization(torch.autograd.Function):
+    """OnnxLpNormalization."""
+
+    @staticmethod
+    def forward(
+        ctx: Callable,  # noqa: ARG004
```
samet-akcay: Instead of ignoring the unused args and methods, the suggested way is to delete them at the beginning of the function. Something like

```python
    @staticmethod
    def forward(
        ctx: Callable,
        x: torch.Tensor,
        axis: int = 0,
        p: int = 2,
        eps: float = 1e-12,
    ) -> torch.Tensor:
        """Forward."""
        del ctx, p    # These args are not used.

        denom = x.norm(2, axis, True).clamp_min(eps).expand_as(x)
        return x / denom
```

src/otx/v2/adapters/torch/mmengine/mmseg/modules/models/utils/psp_layer.py
```diff
@@ -0,0 +1,51 @@
+"""PSP module."""
+# Copyright (c) 2019 MendelXu
```
samet-akcay:
```suggestion
"""PSP module."""

# Copyright (c) 2019 MendelXu
```

src/otx/v2/adapters/torch/mmengine/mmseg/modules/models/utils/psp_layer.py
```diff
@@ -0,0 +1,51 @@
...
+#
+from __future__ import annotations
+
+from typing import ClassVar
+
+import torch
+from torch import nn
+
+
+class PSPModule(nn.Module):
```
samet-akcay: Is PSP a well-known module in the recent literature? If not would it be an idea to make the naming a bit more explicit?

tests/v2/integration/api/segmentation/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Copyright (C) 2023 Intel Corporation
```
samet-akcay, line 3: can you add a docstring here?

tests/v2/integration/api/segmentation/test_mmseg.py
```diff
@@ -0,0 +1,137 @@
+# Copyright (C) 2023 Intel Corporation
```
samet-akcay, line 3: docstring here too

tests/v2/integration/api/segmentation/test_mmseg.py
```diff
@@ -0,0 +1,137 @@
...
+        6. Export Openvino IR Model
+
+        Returns:
+            None
+        """
+        # Setup Engine
+        engine = Engine(work_dir=tmp_dir_path)
+        built_model = get_model(model=model, num_classes=dataset.num_classes)
+
+        # Train (1 epochs)
```
samet-akcay, line 96:
```suggestion
        # Train (1 epoch)
```

samet-akcay (changes requested):
Thanks @eugene123tw, it was a really clean implementation! Really enjoyed reading the code!

I've got some minor comments only.

src/otx/v2/adapters/torch/mmengine/mmseg/dataset.py
jaegukhyun: I understand, we should apply global solution for this in next phase. In detection case, the only problem is LoadResizeDataFromOTXDataset pipeline, so I just fix like this
image

tests/v2/integration/cli/test_cli.py
jaegukhyun: Okay, then how about setting deterministic False for a while? Detection task also couldn't support deterministic

jaegukhyun:
@eugene123tw Will you update unit tests in this PR or another separate PR?

src/otx/v2/adapters/torch/mmengine/mmseg/dataset.py
```diff
@@ -0,0 +1,235 @@
...
+    Args:
+        subset (str): The subset of the dataset. Can be "train", "val", or "test".
+
+    Returns:
+        list: The list of processing steps to be applied to the data.
+    """
+    if subset == "train":
+        return [
+            {"type": "LoadImageFromOTXDataset"},
+            {"type": "LoadAnnotationFromOTXDataset", "_scope_": "mmseg"},
```
goodsong81, line 43: Why not `LoadResizeDataFromOTXDataset`?
Aligned with detection & classification? @jaegukhyun @harimkang

src/otx/v2/adapters/torch/mmengine/mmseg/model.py
goodsong81: Similar approach worked in OTX1.x PyPI sdist.

goodsong81 (commented):
Thank you for work! I've left a few comments.
Besides, I would also recommend to replace the license notice by shorter version as Samet suggested.

src/otx/v2/adapters/torch/mmengine/engine.py
eugene123tw: yes, this is related to mmX deployment (export-related) function. I think we need a better mechanism loading YAML instead of patching things in `_update_config` as it's quite messy and error-prone patching things in functions.

tests/v2/integration/cli/test_cli.py
eugene123tw: @jaegukhyun lets' change it to False.

src/otx/v2/adapters/torch/mmengine/mmseg/modules/models/utils/asymmetric_position_attention.py
eugene123tw: This is one of variables for PyramidSceneParsing (PSP Module). I'd like to keep the name, but I changed `class PSPModule` to `class PyramidSceneParsing`, so I think it should be quite clear to call it `psp_size` instead of `pyramid_scene_parsing_pooling_size ` (as it's a bit verbose I think)

samet-akcay: fair comment

samet-akcay (commented):
Thanks @eugene123tw. I have 1-2 super-minor comments that have not been resolved. Will you be able to address them or do you want to leave them to the next phase?

src/otx/v2/adapters/torch/mmengine/mmseg/dataset.py
eugene123tw: I added a TODO and will implement it in optimization phase (second phase.)

src/otx/v2/adapters/torch/mmengine/mmseg/modules/models/utils/psp_layer.py
eugene123tw: I renamed it to `PyramidScenePooling`

samet-akcay (approved):
Looking good to me! Thanks a lot @eugene123tw

samet-akcay (approved):
I'm happy with the changes @eugene123tw. Thanks again!

wonjuleee (approved):
LGTM!

src/otx/v2/adapters/datumaro/adapter/segmentation_dataset_adapter.py
```diff
@@ -152,22 +144,31 @@ def _remove_labels(self, label_names: list) -> bool:
...
 
+    def get_label_schema(self) -> LabelSchemaEntity:
+        """Get Label Schema."""
+        # TODO (Eugene): move this to base class
+        # CVS-124394
+        if getattr(self, "label_entities", None) is None:
+            label_information = self._prepare_label_information(self.dataset)
+            self.label_entities = label_information["label_entities"]
+        return super()._generate_default_label_schema(self.label_entities)
+
```
jaegukhyun: I think this could be removed.

src/otx/v2/api/core/dataset.py
```diff
@@ -315,8 +315,10 @@ def set_datumaro_adapters(self, data_roots: str | None = None) -> None:
             unlabeled_data_roots=self.unlabeled_data_roots,
             unlabeled_file_list=self.unlabeled_file_list,
         )
-        self.label_schema: LabelSchemaEntity = self.dataset_adapter.get_label_schema()
         self.dataset_entity: dict[Subset, DatumDataset] = self.dataset_adapter.get_otx_dataset()
+        # NOTE: sequence matters as label_entities is changed in get_otx_dataset()
+        # I'd suggest not to tamper with labels in get_otx_dataset() as unexpected behavior may occur
+        self.label_schema: LabelSchemaEntity = self.dataset_adapter.get_label_schema()
```
jaegukhyun: Order should be get_label_schema() -> get_otx_dataset()

When get_lable_schema() -> get_otx_dataset() scenario, there is a bug that OTX will show unused categories in validation metric. 
image

When get_otx_dataset() -> get_label_schema() scenario, there is a bug that OTX(Datumaro adapters) will delete used categories.
image
I think the second bug looks more severe, so let's keep the procedure as it is and fix this bug in later phase

jaegukhyun (commented):
I left some comments. Please take a look

goodsong81 (approved):
Sorry for late review. LGTM from functionality perspective.
Could you rename the PR title according to the convention, by the way?

sungmanc (approved):
LGTM

src/otx/v2/api/core/dataset.py
wonjuleee: @eugene123tw, one more vote for @jaegukhyun's opinion. Please remove `get_otx_dataset()` -> `get_label_schema()` scenario.

eugene123tw: DONE

jaegukhyun (approved):

PR #2615, opened 2023-11-08
Add config_dict attribute to mmengine based model

jaegukhyun:
### Summary
Unlike current visual prompting and anomaly models, mmengine based models need some patch to model config during building model with the model config. 
Therefore, cli need to get patched config from model api. 
This PR is for adding config_dict attribute to mmengine based model so that mmengine based model api can give proper model config to cli for re-building same model.

Since model api should be refactored in near future, this PR only patch mmengine based model api.

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/v2/adapters/torch/mmengine/mmpretrain/model.py
+++ b/src/otx/v2/adapters/torch/mmengine/mmpretrain/model.py
@@ -6,6 +6,7 @@
 from __future__ import annotations
 
 import fnmatch
+from copy import deepcopy
 from pathlib import Path
 
 import torch
@@ -147,11 +148,15 @@ def get_model(
             if head and hasattr(head, "num_classes"):
                 model["model"]["head"]["num_classes"] = num_classes
 
-    model = get_mmpretrain_model(model, pretrained=pretrained, **kwargs)
+    torch_model = get_mmpretrain_model(model, pretrained=pretrained, **kwargs)
 
-    if channel_last and isinstance(model, torch.nn.Module):
-        model = model.to(memory_format=torch.channels_last)
-    return model
+    source_config = deepcopy(torch_model._config.model)  # noqa: SLF001
+    source_config.name = model_name
+    torch_model.config_dict = Config.to_dict(source_config)
+
+    if channel_last and isinstance(torch_model, torch.nn.Module):
+        torch_model = torch_model.to(memory_format=torch.channels_last)
+    return torch_model
 
 
 def list_models(pattern: str | None = None, **kwargs) -> list[str]:
--- a/src/otx/v2/adapters/torch/mmengine/mmpretrain/model.py
+++ b/src/otx/v2/adapters/torch/mmengine/mmpretrain/model.py
@@ -150,6 +150,7 @@ def get_model(
 
     torch_model = get_mmpretrain_model(model, pretrained=pretrained, **kwargs)
 
+    # Config modification below should not affect original model config.
     source_config = deepcopy(torch_model._config.model)  # noqa: SLF001
     source_config.name = model_name
     torch_model.config_dict = Config.to_dict(source_config)
--- a/src/otx/v2/cli/cli.py
+++ b/src/otx/v2/cli/cli.py
@@ -349,8 +349,8 @@ class OTXCLIv2:
         self.model = self.auto_runner.get_model(model={**model_cfg}, num_classes=num_classes)
         # For prediction class
         # NOTE: We'll need to move this somewhere other than here in the next phase.
-        if num_classes is not None and "num_classes" in model_cfg.get("head", {}):
-            model_cfg["head"]["num_classes"] = num_classes
+        if getattr(self.model, "config_dict", None):
+            model_cfg = self.model.config_dict
         workspace_config["model"] = {**model_cfg}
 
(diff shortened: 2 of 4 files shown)


src/otx/v2/cli/cli.py
```diff
@@ -349,8 +349,8 @@ def instantiate_classes(self, subcommand: str) -> None:
         self.model = self.auto_runner.get_model(model={**model_cfg}, num_classes=num_classes)
         # For prediction class
         # NOTE: We'll need to move this somewhere other than here in the next phase.
-        if num_classes is not None and "num_classes" in model_cfg.get("head", {}):
-            model_cfg["head"]["num_classes"] = num_classes
+        if getattr(self.model, "config_dict", None):
+            model_cfg = self.model.config_dict
```
goodsong81, line 352: Can this be more explicit?
It's weird to see CLI checks if some attribute exists.
For example,
```suggestion
        model_cfg = self.model.get_config()
```

goodsong81 (changes requested):
Tried to minimize design change request, but exposing implementation details to outside of APIs would make things complicated.
Why don't you make an abstract method for it? CLI should work only with the abstract APIs.

src/otx/v2/cli/cli.py
goodsong81: If it's temporary, let's keep it as is. Let me know your opinion. :)

sungmanc: It sounds great, however, currently we don't have a proper model API now. So, `self.model` is a `nn.Module` itself. 
We might need wrapper or APIs for the Models. How about applying this opinion when we bring the Model API?

goodsong81: Sure.

goodsong81 (approved):

src/otx/v2/cli/cli.py
jaegukhyun: Thank you for your opinion. I totally agree with you that we do not expose implementation details like this. However, this is temporary solution before we refactor our model api. Currently self.model is just nn.Module, not our model api. So this PR try to add temporary attribute to nn.Module if they change input model api. What do you think?

sungmanc (approved):

src/otx/v2/adapters/torch/mmengine/mmpretrain/model.py
```diff
@@ -147,11 +148,15 @@ def get_model(
             if head and hasattr(head, "num_classes"):
                 model["model"]["head"]["num_classes"] = num_classes
 
-    model = get_mmpretrain_model(model, pretrained=pretrained, **kwargs)
+    torch_model = get_mmpretrain_model(model, pretrained=pretrained, **kwargs)
 
-    if channel_last and isinstance(model, torch.nn.Module):
-        model = model.to(memory_format=torch.channels_last)
-    return model
+    source_config = deepcopy(torch_model._config.model)  # noqa: SLF001
```
samet-akcay, line 154: Can you add a comment here why this is needed? This would be for a future reference

src/otx/v2/cli/cli.py
```diff
@@ -349,8 +349,8 @@ def instantiate_classes(self, subcommand: str) -> None:
         self.model = self.auto_runner.get_model(model={**model_cfg}, num_classes=num_classes)
         # For prediction class
         # NOTE: We'll need to move this somewhere other than here in the next phase.
```
samet-akcay: Is the comment above still valid?

src/otx/v2/cli/cli.py
samet-akcay, line 352: How about something like this to make it easier to follow?
```suggestion
       model_cfg = self.model.config_dict if hasattr(self.model, 'config_dict') else None
```

samet-akcay (changes requested):
Thanks, I've got some minor comments only.

src/otx/v2/adapters/torch/mmengine/mmpretrain/model.py
jaegukhyun: Done by [cafc1eb](https://github.com/openvinotoolkit/training_extensions/pull/2615/commits/cafc1ebd39299426c03172d9ed443265257bfa16)

src/otx/v2/cli/cli.py
jaegukhyun: Remove first line. I think we should remove this kind of implementation in next phase

src/otx/v2/cli/cli.py
jaegukhyun: I modified it, but someone may think it is a little bit awkward. Please check this.

samet-akcay: yeah, this does not look good. I think I misunderstand the code

samet-akcay: so we assign `self.model.config_dict` even if it is `None`?

jaegukhyun: if `self.model.config_dict` is `None` then code will pass the if branch. So model_cfg will not be changed.

samet-akcay: thanks for the clarification. My bad, let's keep the old version

samet-akcay (approved):

PR #2625, opened 2023-11-10
Update publish workflow

yunchu:
* update publish workflow to push whl to internal pypi
* remove 'rc' suffix from the version string

### DIFF
--- a/src/otx/__init__.py
+++ b/src/otx/__init__.py
@@ -3,5 +3,5 @@
 # Copyright (C) 2021-2023 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 
-__version__ = "1.4.4rc1"
+__version__ = "1.4.4"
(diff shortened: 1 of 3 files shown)


chuneuny-emily (approved):

src/otx/__init__.py
```diff
@@ -3,5 +3,5 @@
 # Copyright (C) 2021-2023 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 
-__version__ = "1.4.4rc1"
+__version__ = "1.4.4"
```
goodsong81, line 6: Can we upload same versions multiple times on internal PyPI?

yunchu: no, but I think we can delete uploaded files in the internal PyPI. I've asked the way to do that to the internal team but did not get the answer yet.

goodsong81 (approved):

PR #2626, opened 2023-11-10
Fix h-label bug of missing parent labels in output

### LINKED ISSUE #2579: training accuracy reported as 0.0 with CUB hierarchical label for the classification
**Describe the bug**
Training accuracy reported as 0.0 with CUB hierarchical label for the classification
here is the part of regression_config.json for this testing
image

**Steps to Reproduce**

1. run classification regression TC 
```
$ CI_DATA_ROOT=/home/yunchule/data/ci_datasets/ tox -vvv -e tests-cls-py310-pt1 -- tests/regression/classification/test_classification.py::TestRegressionHierarchicalLabelClassification::test_otx_train[Custom_Image_Classification_EfficientNet-V2-S]
```
2. check the log
```
>       assert test_result["passed"] is True, test_result["log"]
E       AssertionError: [EfficientNet-V2-S] Performance: (0.0) < Criteria: (0.5418), threshold: 0.1.
E       assert False is True
```

**Environment:**

- OS: Ubuntu 20.04
- Framework version: Pytorch
- Python version: 3.10
- OpenVINO version: 2023.0
- CUDA/cuDNN version: 11.7
- GPU model and memory: 3090/24G

goodsong81:
### Summary

* In hierarchical classification, the parent labels might not be annotated explicitly.
In this case, the higher level classifier cannot learn how to classify the abstract semantic
so that the losses could be NaN, and the final leaf outputs are filtered out in post-processing.
(e.g. Datumaro dataset exported from Geti contains only leaf label annotations)
* This PR fills the parent label annotation if missing based on label schema.
As a result, the output would look like [ancestor(80%), ..., parent(60%), child(70%), ... other(60%)]. 
(No harm in case of full annotation)
* Additionally, added the top level label exclusive group for h-label test data.
Without this, models would not classify the highest level semantic.
* This PR would resolve: https://github.com/openvinotoolkit/training_extensions/issues/2579

### How to test

* E2e / Regression test
```bash
tox -e tests-all-py310 -- tests/e2e/cli/classification/test_classification.py::TestToolsHierarchicalClassification::test_otx_train[Custom_Image_Classification_EfficinetNet-B0]
```

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algorithms/classification/adapters/mmcls/datasets/otx_datasets.py
+++ b/src/otx/algorithms/classification/adapters/mmcls/datasets/otx_datasets.py
@@ -300,6 +300,7 @@ class OTXHierarchicalClsDataset(OTXMultilabelClsDataset):
 
     def __init__(self, **kwargs):
         self.hierarchical_info = kwargs.pop("hierarchical_info", None)
+        self.label_schema = kwargs.pop("label_schema", None)
         super().__init__(**kwargs)
 
     def load_annotations(self):
@@ -308,6 +309,13 @@ class OTXHierarchicalClsDataset(OTXMultilabelClsDataset):
         for i, _ in enumerate(self.otx_dataset):
             class_indices = []
             item_labels = self.otx_dataset[i].get_roi_labels(self.labels, include_empty=include_empty)
+            if self.label_schema:
+                # NOTE: Parent labels might be missing in annotations.
+                # This code fills the gap just in case.
+                full_item_labels = set()
+                for label in item_labels:
+                    full_item_labels.update(self.label_schema.get_ancestors(label))
+                item_labels = full_item_labels
             ignored_labels = self.otx_dataset[i].ignored_labels
             if item_labels:
                 num_cls_heads = self.hierarchical_info["num_multiclass_heads"]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ All notable changes to this project will be documented in this file.
 - Fix h-label loss normalization issue w/ exclusive label group of singe label (<https://github.com/openvinotoolkit/training_extensions/pull/2604>)
 - Fix division by zero in class incremental learning for classification (<https://github.com/openvinotoolkit/training_extensions/pull/2606>)
 - Fix saliency maps calculation issue for detection models (<https://github.com/openvinotoolkit/training_extensions/pull/2609>)
+- Fix h-label bug of missing parent labels in output (<https://github.com/openvinotoolkit/training_extensions/pull/2626>)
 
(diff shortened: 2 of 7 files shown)


src/otx/algorithms/classification/adapters/mmcls/datasets/otx_datasets.py
```diff
@@ -308,6 +309,13 @@ def load_annotations(self):
         for i, _ in enumerate(self.otx_dataset):
             class_indices = []
             item_labels = self.otx_dataset[i].get_roi_labels(self.labels, include_empty=include_empty)
+            if self.label_schema:
+                # NOTE: Parent labels might be missing in annotations.
+                # This code fills the gap just in case.
+                full_item_labels = set()
+                for label in item_labels:
+                    full_item_labels.update(self.label_schema.get_ancestors(label))
+                item_labels = full_item_labels
```
sungmanc, line 318: What `label_schema.get_ancestors()` returns ? is it okay to overwrite the `item_labels` with the `full_item_labels` ?

sungmanc (commented):
Thanks for the fix ! I left a question :)

src/otx/algorithms/classification/adapters/mmcls/datasets/otx_datasets.py
goodsong81: It returns [child, parent, grand parent, ...] which is needed for h-label classifier training & evaluation by definition. Geti gives such annotation, but Datumaro does not. So we need this patch.
In case of below hierarchy,
* human
  * male
    * boy
    * man 
  * female
    * girl
    * woman

We need not only leaf label annotations like [boy], [woman], but also full annotations like [male, man], [female, girl]
to train "male" vs "female" high-level classifier.

So, yes, it should be added if absence.

jaegukhyun (approved):

src/otx/algorithms/classification/adapters/mmcls/datasets/otx_datasets.py
sungmanc: > It returns [child, parent, grand parent, ...] which is needed for h-label classifier training & evaluation by definition. Geti gives such annotation, but Datumaro does not. So we need this patch. In case of below hierarchy,
> 
> * human
>   
>   * male
>     
>     * boy
>     * man
>   * female
>     
>     * girl
>     * woman
> 
> We need labels like not only [boy], [woman], but also [male, man], [female, girl] to train "male" vs "femail" high-level classifier.
> 
> So, yes, it should be added if absence.

Thanks for the detailed reply.

sungmanc (approved):

PR #2631, opened 2023-11-14
Refactor the datumaro adapter

sungmanc:
### Summary

After this PR, `self.label_entities` will be generated at the `get_label_schema()` function instead of `get_otx_dataset()`.
So, the only function we should inherit is the `get_otx_dataset()` at each child class. 

Recommended step is `get_label_schema()` --> `get_otx_dataset()`. However, if there is a counter-example, it raises an error since it couldn't find the `self.label_entities`. That's the reason why I added the codes below.
```python
def get_otx_dataset(self) -> DatasetEntity:
        if not hasattr(self, "label_entities"):
            super().get_label_schema()
```

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/v2/adapters/datumaro/adapter/action_dataset_adapter.py
+++ b/src/otx/v2/adapters/datumaro/adapter/action_dataset_adapter.py
@@ -118,9 +118,8 @@ class ActionClassificationDatasetAdapter(ActionBaseDatasetAdapter):
 
     def get_otx_dataset(self) -> DatasetEntity:
         """Convert DatumaroDataset to DatasetEntity for Acion Classification."""
-        label_information = self._prepare_label_information(self.dataset)
-        self.label_entities = label_information["label_entities"]
-
+        if not hasattr(self, "label_entities"):
+            super().get_label_schema()
         dataset_items: List[DatasetItemEntity] = []
         for subset, subset_data in self.dataset.items():
             for _, datumaro_items in subset_data.subsets().items():
@@ -156,9 +155,8 @@ class ActionDetectionDatasetAdapter(ActionBaseDatasetAdapter):
 
     def get_otx_dataset(self) -> DatasetEntity:
         """Convert DatumaroDataset to DatasetEntity for Acion Detection."""
-        label_information = self._prepare_label_information(self.dataset)
-        self.label_entities = label_information["label_entities"]
-
+        if not hasattr(self, "label_entities"):
+            super().get_label_schema()
         # Detection use index 0 as a background category
         for label_entity in self.label_entities:
             label_entity.id = ID(int(label_entity.id) + 1)
--- a/src/otx/v2/adapters/datumaro/adapter/action_dataset_adapter.py
+++ b/src/otx/v2/adapters/datumaro/adapter/action_dataset_adapter.py
@@ -153,8 +153,6 @@ class ActionDetectionDatasetAdapter(ActionBaseDatasetAdapter):
 
     def get_otx_dataset(self) -> DatasetEntity:
         """Convert DatumaroDataset to DatasetEntity for Acion Detection."""
-        if not hasattr(self, "label_entities"):
-            super().get_label_schema()
         # Detection use index 0 as a background category
         for label_entity in self.label_entities:
             label_entity.id = ID(int(label_entity.id) + 1)
--- a/src/otx/v2/adapters/datumaro/adapter/action_dataset_adapter.py
+++ b/src/otx/v2/adapters/datumaro/adapter/action_dataset_adapter.py
@@ -118,8 +118,6 @@ class ActionClassificationDatasetAdapter(ActionBaseDatasetAdapter):
 
     def get_otx_dataset(self) -> DatasetEntity:
         """Convert DatumaroDataset to DatasetEntity for Acion Classification."""
-        if not hasattr(self, "label_entities"):
-            super().get_label_schema()
         dataset_items: List[DatasetItemEntity] = []
         for subset, subset_data in self.dataset.items():
             for _, datumaro_items in subset_data.subsets().items():
--- a/src/otx/v2/adapters/datumaro/adapter/anomaly_dataset_adapter.py
+++ b/src/otx/v2/adapters/datumaro/adapter/anomaly_dataset_adapter.py
@@ -7,7 +7,7 @@
 from __future__ import annotations
 
 import os
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, Union
 
 import cv2
 import numpy as np
@@ -15,17 +15,17 @@ from datumaro.components.annotation import Bbox as DatumBbox
 from datumaro.components.annotation import Label as DatumLabel
 from datumaro.components.annotation import Mask as DatumMask
 from datumaro.components.dataset import Dataset as DatumDataset
+from datumaro.components.annotation import Categories as DatumCategories
 
 from otx.v2.adapters.torch.modules.utils.mask_to_bbox import mask2bbox
-from otx.v2.api.entities.datasets import DatasetEntity
 from otx.v2.api.entities.id import ID
 from otx.v2.api.entities.label import LabelEntity
 from otx.v2.api.entities.label_schema import LabelSchemaEntity
 from otx.v2.api.entities.subset import Subset
-from otx.v2.api.entities.utils.segmentation_utils import create_annotation_from_segmentation_map
 
 from .datumaro_dataset_adapter import DatumaroDatasetAdapter
 
+LabelInformationType = Dict[str, Union[List[LabelEntity], List[DatumCategories]]]
 
 class AnomalyBaseDatasetAdapter(DatumaroDatasetAdapter):
     """BaseDataset Adpater for Anomaly tasks inherited from DatumaroDatasetAdapter."""
@@ -72,7 +72,7 @@ class AnomalyBaseDatasetAdapter(DatumaroDatasetAdapter):
             dataset[Subset.TESTING] = DatumDataset.import_from(test_data_roots, format="image_dir")
         return dataset
 
-    def _prepare_anomaly_label_information(self) -> List[LabelEntity]:
+    def _prepare_label_information(self) -> LabelInformationType:
         """Prepare LabelEntity List."""
         normal_label = LabelEntity(id=ID(0), name="Normal", domain=self.domain)
         abnormal_label = LabelEntity(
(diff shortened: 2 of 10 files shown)


src/otx/v2/adapters/datumaro/adapter/action_dataset_adapter.py
```diff
@@ -118,9 +118,8 @@ class ActionClassificationDatasetAdapter(ActionBaseDatasetAdapter):
 
     def get_otx_dataset(self) -> DatasetEntity:
         """Convert DatumaroDataset to DatasetEntity for Acion Classification."""
-        label_information = self._prepare_label_information(self.dataset)
-        self.label_entities = label_information["label_entities"]
-
+        if not hasattr(self, "label_entities"):
+            super().get_label_schema()
```
wonjuleee: `get_label_schema` is always called prior to get dataset as following https://github.com/openvinotoolkit/training_extensions/blob/376ce9c155a4afe9e72baa87a6f6a693e3108939/src/otx/v2/api/core/dataset.py#L318. Accordingly, `label_entities` should be computed within `get_label_schema`. Could you check the order of function calls during this operation?

src/otx/v2/adapters/datumaro/adapter/action_dataset_adapter.py
```diff
@@ -156,9 +155,8 @@ class ActionDetectionDatasetAdapter(ActionBaseDatasetAdapter):
 
     def get_otx_dataset(self) -> DatasetEntity:
         """Convert DatumaroDataset to DatasetEntity for Acion Detection."""
-        label_information = self._prepare_label_information(self.dataset)
-        self.label_entities = label_information["label_entities"]
-
+        if not hasattr(self, "label_entities"):
+            super().get_label_schema()
```
wonjuleee: Same as above :)

src/otx/v2/adapters/datumaro/adapter/detection_dataset_adapter.py
```diff
@@ -24,9 +24,8 @@ class DetectionDatasetAdapter(DatumaroDatasetAdapter):
 
     def get_otx_dataset(self) -> DatasetEntity:
         """Convert DatumaroDataset to DatasetEntity for Detection."""
-        # Prepare label information
-        label_information = self._prepare_label_information(self.dataset)
-        self.label_entities = label_information["label_entities"]
+        if not hasattr(self, "label_entities"):
+            super().get_label_schema()
```
wonjuleee, line 28: Same as above.

src/otx/v2/adapters/datumaro/adapter/datumaro_dataset_adapter.py
```diff
@@ -219,6 +219,8 @@ def get_otx_dataset(self) -> dict[Subset, DatumDataset]:
 
     def get_label_schema(self) -> LabelSchemaEntity:
         """Get Label Schema."""
+        label_information = self._prepare_label_information(self.dataset)
+        self.label_entities = label_information["label_entities"]
```
goodsong81: In my opinion, `get_something()` method should give something, but nothing else. Setting `self.label_entities` in this method is a side-effect, which is counterintuitive.
Why don't you make it more explicit?

src/otx/v2/adapters/datumaro/adapter/action_dataset_adapter.py
sungmanc: Yeah, I also think that. However, some unit tests are not written in this order, that's the reason why I changed the unit test code for the detection. I think changing unit tests is the better way. Thanks for pointing out.


image

src/otx/v2/adapters/datumaro/adapter/detection_dataset_adapter.py
goodsong81: Calling get_somthing() but not actually getting something, which mean it's just being called for the internal side effect. I don't think this is quite readable code.

wonjuleee:
Thank you so much for your quick execution. I prefer to create `self.label_entities` in `get_label_schema()` rather than adding below. What do you think?
```python
        if not hasattr(self, "label_entities"):
            super().get_label_schema()
 ```

**Review by goodsong81 — CHANGES_REQUESTED** [review]
Could you consider to refine the function signatures or names?
It's not that good practice to depend on some side-effect of a method.
Please see my following comments.

File: src/otx/v2/adapters/datumaro/adapter/action_dataset_adapter.py (hunk 1 above)
Comment (jaegukhyun) [reply src/otx/v2/adapters/datumaro/adapter/action_dataset_adapter.py]: I also prefer removing this part

Comment (wonjuleee) [reply src/otx/v2/adapters/datumaro/adapter/action_dataset_adapter.py]: Please make a commit for updating unit tests and attach the commit number here.

**Comment by samet-akcay** [comment]
Just to understand this better, this is a temporary workaround, right? At the end, we won't need this as this will be provided by Datumaro, correct?

**Comment by sungmanc** [comment]
> Just to understand this better, this is a temporary workaround, right? At the end, we won't need this as this will be provided by Datumaro, correct?

Yeah, this is not the final one :)

File: src/otx/v2/adapters/datumaro/adapter/datumaro_dataset_adapter.py (hunk 4 above)
Comment (sungmanc) [reply src/otx/v2/adapters/datumaro/adapter/datumaro_dataset_adapter.py]: Done. https://github.com/openvinotoolkit/training_extensions/pull/2631/commits/3ca5154cc9fd8172865590ad46d03db01b772a91

File: src/otx/v2/adapters/datumaro/adapter/detection_dataset_adapter.py (hunk 3 above)
Comment (sungmanc) [reply src/otx/v2/adapters/datumaro/adapter/detection_dataset_adapter.py]: Done. https://github.com/openvinotoolkit/training_extensions/pull/2631/commits/3ca5154cc9fd8172865590ad46d03db01b772a91

**Comment by sungmanc** [comment]
> Thank you so much for your quick execution. I prefer to create `self.label_entities` in `get_label_schema()` rather than adding below. What do you think?
> 
> ```python
>         if not hasattr(self, "label_entities"):
>             super().get_label_schema()
> ```

I removed all code lines regarding this. https://github.com/openvinotoolkit/training_extensions/pull/2631/commits/3ca5154cc9fd8172865590ad46d03db01b772a91

File: src/otx/v2/adapters/datumaro/adapter/action_dataset_adapter.py (hunk 1 above)
Comment (sungmanc) [reply src/otx/v2/adapters/datumaro/adapter/action_dataset_adapter.py]: Done. https://github.com/openvinotoolkit/training_extensions/pull/2631/commits/3ca5154cc9fd8172865590ad46d03db01b772a91

**Comment by sungmanc** [comment]
@wonjuleee , @goodsong81  Please review again this PR, I reflected your reviews. 
So, from now, `self._prepare_label_information()` is called inside of the `__init__` and `get_label_schema` just generates the `label_schema` by using `self.label_entities`

File: src/otx/v2/adapters/datumaro/adapter/action_dataset_adapter.py [hunk 5]
```diff
@@ -156,9 +153,8 @@ class ActionDetectionDatasetAdapter(ActionBaseDatasetAdapter):
...
```
jaegukhyun: We need to remove this

sungmanc: Thanks ! https://github.com/openvinotoolkit/training_extensions/pull/2631/commits/27bd09929221077fede025123dda1cee8d28f69f

jaegukhyun (approved):

src/otx/v2/adapters/datumaro/adapter/anomaly_dataset_adapter.py
```diff
@@ -7,25 +7,25 @@
...
-from otx.v2.api.entities.datasets import DatasetEntity
 from otx.v2.api.entities.id import ID
 from otx.v2.api.entities.label import LabelEntity
 from otx.v2.api.entities.label_schema import LabelSchemaEntity
 from otx.v2.api.entities.subset import Subset
-from otx.v2.api.entities.utils.segmentation_utils import create_annotation_from_segmentation_map
 
 from .datumaro_dataset_adapter import DatumaroDatasetAdapter
 
+LabelInformationType = Dict[str, Union[List[LabelEntity], List[DatumCategories]]]
```
wonjuleee: This can be imported from datumaro dataset adapter.

src/otx/v2/adapters/datumaro/adapter/classification_dataset_adapter.py
```diff
@@ -32,11 +34,6 @@ def get_otx_dataset(self) -> dict[Subset, DatumDataset]:
 
     def get_label_schema(self) -> LabelSchemaEntity:
         """Get Label Schema."""
-        label_information = self._prepare_label_information(self.dataset)
-        self.category_items = label_information["category_items"]
-        self.label_groups = label_information["label_groups"]
-        self.label_entities = label_information["label_entities"]
-
```
wonjuleee, line 39: could you clarify where these are computed at your PR?

src/otx/v2/adapters/datumaro/adapter/datumaro_dataset_adapter.py
```diff
@@ -410,7 +413,7 @@ def _filter_unlabeled_data(self, unlabeled_dataset: DatumDataset, unlabeled_file
         for item in copy_dataset:
             if item.id not in file_list:
                 unlabeled_dataset.remove(item.id, item.subset)
-
+
```
wonjuleee: typo?

wonjuleee (approved):
Looks good to me overall. Please fix minor typos.

src/otx/v2/adapters/datumaro/adapter/anomaly_dataset_adapter.py
sungmanc: Done. https://github.com/openvinotoolkit/training_extensions/pull/2631/commits/1cbcdcd953906dadb225f5d365962c2dc9ba8467

goodsong81 (approved):
Thank you for the refinement. Much more clear now.

src/otx/v2/adapters/datumaro/adapter/datumaro_dataset_adapter.py
sungmanc: Ah, this comes from my VIM plugin :) , done -> https://github.com/openvinotoolkit/training_extensions/pull/2631/commits/1cbcdcd953906dadb225f5d365962c2dc9ba8467

PR #2647, opened 2023-11-17
Make `max_num_detections` configurable

goodsong81:
### Summary

* Fix hard-coded maximum number of detections regarding NMS threshold both for torch & OpenVINO models
* Add a configurable parameter: `postprocessing.max_num_detections` for detection, instance segmenatation, rotated_detection tasks.
* Defaults to 0, which means using the per-model default (hard-coded) setting

image

### How to test

```bash
otx train MobileNetV2-ATSS --train-data-roots /path/to/dataset params --postprocessing.max_num_detections 500
```

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algorithms/common/configs/training_base.py
+++ b/src/otx/algorithms/common/configs/training_base.py
@@ -1,18 +1,7 @@
 """Base Configuration of OTX Common Algorithms."""
 
-# Copyright (C) 2022 Intel Corporation
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-# http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing,
-# software distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions
-# and limitations under the License.
+# Copyright (C) 2022-2023 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
 
 from sys import maxsize
 
@@ -224,6 +213,16 @@ class BaseConfig(ConfigurableParameters):
             max_value=1,
             header="Confidence threshold",
             description="This threshold only takes effect if the threshold is not set based on the result.",
+            affects_outcome_of=ModelLifecycle.INFERENCE,
+        )
+
+        max_num_detections = configurable_integer(
+            header="Maximum number of detection per image",
+            description="Extra detection outputs will be discared in non-maximum suppression process. "
+            "Defaults to 0, which means per-model default value.",
+            default_value=0,
+            min_value=0,
+            max_value=10000,
             affects_outcome_of=ModelLifecycle.INFERENCE,
         )
 
--- a/src/otx/algorithms/detection/adapters/mmdet/configurer.py
+++ b/src/otx/algorithms/detection/adapters/mmdet/configurer.py
@@ -114,7 +114,7 @@ class DetectionConfigurer:
             new_classes = np.setdiff1d(data_classes, model_classes).tolist()
             train_data_cfg["new_classes"] = new_classes
 
-    def configure_model(self, cfg, ir_options, max_num_detections):  # noqa: C901
+    def configure_model(self, cfg, ir_options, max_num_detections=0):  # noqa: C901
         """Patch config's model.
 
         Change model type to super type
--- a/src/otx/algorithms/detection/adapters/mmdet/configurer.py
+++ b/src/otx/algorithms/detection/adapters/mmdet/configurer.py
@@ -158,10 +158,14 @@ class DetectionConfigurer:
             test_cfg.nms_pre = max_num_detections * 10
             # Special cases for 2-stage detectors (e.g. MaskRCNN)
             if hasattr(test_cfg, "rpn"):
-                test_cfg.rpn.nms_pre = max_num_detections * 10
+                test_cfg.rpn.nms_pre = max_num_detections * 20
                 test_cfg.rpn.max_per_img = max_num_detections * 10
             if hasattr(test_cfg, "rcnn"):
                 test_cfg.rcnn.max_per_img = max_num_detections
+            train_cfg = cfg.model.train_cfg
+            if hasattr(train_cfg, "rpn_proposal"):
+                train_cfg.rpn_proposal.nms_pre = max_num_detections * 20
+                train_cfg.rpn_proposal.max_per_img = max_num_detections * 10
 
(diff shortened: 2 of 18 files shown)


src/otx/algorithms/common/configs/training_base.py
```diff
@@ -227,6 +216,16 @@ class BasePostprocessing(ParameterGroup):
...
+        max_num_detections = configurable_integer(
+            header="Maximum number of detection per image",
+            description="Extra detection outputs will be discared in non-maximum suppression process. "
+            "Defaults to 0, which means per-model default value.",
+            default_value=0,
+            min_value=0,
+            max_value=10000,
+            affects_outcome_of=ModelLifecycle.INFERENCE,
+        )
+
```
jaegukhyun, line 228: Don't we have a parameter to control maximum outputs of object detector?

src/otx/algorithms/detection/adapters/mmdet/configurer.py
```diff
@@ -149,6 +150,13 @@ def is_mmov_model(key, value):
                 {"model_path": ir_model_path, "weight_path": ir_weight_path, "init_weight": ir_weight_init},
             )
 
+        # Test config
+        if max_num_detections > 0:
+            logger.info(f"Model max_num_detections: {max_num_detections}")
+            test_cfg = cfg.model.test_cfg
+            test_cfg.max_per_img = max_num_detections
+            test_cfg.nms_pre = max_num_detections * 10
+
```
jaegukhyun: Two stage detector, such as maskrcnn, has parameters to control rpn output. Shouldn't we change that parameters too?

src/otx/algorithms/detection/adapters/mmdet/task.py
```diff
@@ -513,6 +503,12 @@ def _export_model(
         assert len(self._precision) == 1
         export_options["precision"] = str(self._precision[0])
         export_options["type"] = str(export_format)
+        if self.max_num_detections > 0:
+            logger.info(f"Export max_num_detections: {self.max_num_detections}")
+            post_proc_cfg = export_options["deploy_cfg"]["codebase_config"]["post_processing"]
+            post_proc_cfg["max_output_boxes_per_class"] = self.max_num_detections
+            post_proc_cfg["keep_top_k"] = self.max_num_detections
+            post_proc_cfg["pre_top_k"] = self.max_num_detections * 10
```
jaegukhyun, line 511: I'm not sure this may change postprocessing of exported model. How did you verify this?

src/otx/algorithms/common/configs/training_base.py
goodsong81: I didn't know that, except for the `--tiling_parameters.tile_max_number`.
Could you let me know what it is?

jaegukhyun: I think that's the parameter I ask. What is the relation between tile_max_number and max_num_detections?

src/otx/algorithms/detection/adapters/mmdet/task.py
goodsong81: I've looked into ModelAPI & MMDeploy source code. The former had no such filtering, but MMDeploy was using those parameters for OpenVINO NMS layer parameters.
I've tested train -> export -> OV eval to verify if this approach worked.
Do you have any good idea for us to make this a test case?

src/otx/algorithms/detection/adapters/mmdet/configurer.py
goodsong81: Hmm, I forgot them. Thanks for the catch.

src/otx/algorithms/detection/adapters/mmdet/task.py
jaegukhyun: It's hard to test by unit test because it requires exported openvino.xml. Anyway, I just want to check whether this really changes behaviors of mmdeploy. Sometimes mmdeply ignores parameters in config file. If your check this e2e way then I think it's fine at the current stage.

src/otx/algorithms/common/configs/training_base.py
goodsong81: It's only for tile-merging NMS, which means it will restrict the merged output number of all tiles.
`max_num_detections` in this case controls per-tile detections. So in fact, it had been limited to detect maximum 100 or 200 objects for each tile so far.

Anyway, I think I need to align this 2 somehow... let me think.
@eugene123tw Do you have any suggestion?

goodsong81: `--tiling_parameters.tile_max_number` is tiling-specific and there is an own adaptation mechanism so that I think we don't need to merge it with `--postprocessing.max_num_detections` for now because the latter would still control per-time maximum even in case of tiling.
I'd like to hear from @eugene123tw before resolving this conversation.

jaegukhyun: One more question. If max_num_detections is 500 and a image has 4 tiles. What is the max number of outputs for this image if tile_max_number is 500. I think it should be 2000 but it would be 500

src/otx/algorithms/detection/adapters/mmdet/configurer.py
goodsong81: Addressed in https://github.com/openvinotoolkit/training_extensions/pull/2647/commits/1627d4c0a064ebc4144ebb354c1e6226969e0dcd

jaegukhyun: I think we should change train_cfg also. What do you think?

jaegukhyun: If max_dets param is only for inference then, we don't need

src/otx/algorithms/common/configs/training_base.py
goodsong81: Yes, that's the behavior. Currently, users need to set both values properly in case of "manual" tiling setting. In auto tiling, we set the value using dataset statistics.

Do you think the tile_max_number should be automatically adjusted to # tiles x max_num_detections? I'm a little bit against it.

jaegukhyun: I don't think we should do it. It will raise complexity very much. Thanks

src/otx/algorithms/detection/adapters/mmdet/configurer.py
goodsong81: Well... I've been also thinking about it; whether it would have been affecting the loss flow?
Anyway, let's be consistent at least. There might be edge cases, but currently those are all we have regarding the `max_per_img`.
https://github.com/openvinotoolkit/training_extensions/pull/2647/commits/5979c3f94ad96c17d7058abf4728e735ab104240

jaegukhyun (approved):

chuneuny-emily (approved):

PR #2649, opened 2023-11-17
Unify mmX engine's export function

jaegukhyun:
### Summary
In this phase, we want to integrate each algo frameworks' engine.py into a single engine.py
The first step of it is integrate adapters/torch/mmengine/***/engine.py into adapters/torch/mmengine/engine.py

Currently each adapters/torch/mmengine/mmX/engine.py have __init__, _update_config, predict and export function.
Therefore we need to migrate all of these function into adapters/torch/mmengine/engine.py
This PR is going to migrate export function of each adapters/torch/mmengine/mmX/engine.py

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/v2/adapters/torch/lightning/anomalib/engine.py
+++ b/src/otx/v2/adapters/torch/lightning/anomalib/engine.py
@@ -20,6 +20,7 @@ from pytorch_lightning.loggers.logger import Logger
 from torch.utils.data import DataLoader
 
 from otx.v2.adapters.torch.lightning.engine import LightningEngine
+from otx.v2.api.entities.task_type import TaskType
 
 from .registry import AnomalibRegistry
 
@@ -38,16 +39,16 @@ class AnomalibEngine(LightningEngine):
 
     def __init__(
         self,
+        task: TaskType,
         work_dir: str | Path | None = None,
         config: str | dict | None = None,
-        task: str = "anomaly_classification",
     ) -> None:
         """Initialize the Anomalib engine.
 
         Args:
+            task (TaskType): The task to perform.
             work_dir (str | Path | None, optional): The working directory for the engine. Defaults to None.
             config (str | dict | None, optional): The configuration for the engine. Defaults to None.
-            task (str, optional): The task to perform. Defaults to "anomaly_classification".
         """
         super().__init__(work_dir=work_dir, config=config, task=task)
         self.registry = AnomalibRegistry()
--- a/src/otx/v2/adapters/torch/mmengine/engine.py
+++ b/src/otx/v2/adapters/torch/mmengine/engine.py
@@ -20,6 +20,7 @@ from otx.v2.adapters.torch.mmengine.modules.utils.config_utils import CustomConf
 from otx.v2.adapters.torch.mmengine.modules.utils.config_utils import dump_lazy_config
 from otx.v2.adapters.torch.mmengine.registry import MMEngineRegistry
 from otx.v2.api.core.engine import Engine
+from otx.v2.api.entities.task_type import TaskType
 from otx.v2.api.utils.importing import get_all_args, get_default_args
 from otx.v2.api.utils.logger import get_logger
 
@@ -76,14 +77,16 @@ class MMXEngine(Engine):
     """
     def __init__(
         self,
+        task: TaskType,
         work_dir: str | Path | None = None,
     ) -> None:
         """Initialize a new instance of the MMEngine class.
 
         Args:
+            task (TaskType): Task type of engine.
             work_dir (Optional[Union[str, Path]], optional): The working directory for the engine. Defaults to None.
         """
-        super().__init__(work_dir=work_dir)
+        super().__init__(task=task, work_dir=work_dir)
         # Engine's configuration should not affect to DEFAULT_CONFIG
         self.default_config = deepcopy(DEFAULT_CONFIG)
         self.runner: Runner
@@ -503,8 +506,6 @@ class MMXEngine(Engine):
         model: torch.nn.Module | (str | Config) | None = None,  # Module with _config OR Model Config OR config-file
         checkpoint: str | Path | None = None,
         precision: str | None = "float32",  # ["float16", "fp16", "float32", "fp32"]
-        task: str | None = None,
-        codebase: str | None = None,
         export_type: str = "OPENVINO",  # "ONNX" or "OPENVINO"
         deploy_config: str | dict | None = None,
         device: str = "cpu",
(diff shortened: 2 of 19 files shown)


src/otx/v2/adapters/torch/mmengine/engine.py
```diff
@@ -649,6 +644,24 @@ def _update_codebase_config(
...
+            # This should be handled model api
+            codebase_config["post_processing"] = {
+                "score_threshold": 0.05,
+                "confidence_threshold": 0.005,  # for YOLOv3
+                "iou_threshold": 0.5,
+                "max_output_boxes_per_class": 200,
+                "pre_top_k": 5000,
+                "keep_top_k": 100,
+                "background_label_id": -1,
+            }
```
wonjuleee, line 666: All `codebase_config` per task is going to be defined here in the future?

wonjuleee: Aha, I just saw your comment about model API. Do you have any idea about how to handle this kind of configurations in model API?

src/otx/v2/api/core/engine.py
```diff
@@ -23,15 +24,17 @@ class Engine:
     )
     """
 
-    def __init__(self, work_dir: str | Path | None) -> None:
+    def __init__(self, task: TaskType, work_dir: str | Path | None = None) -> None:
```
wonjuleee, line 27: Personally, I am good to have `task` when creating engine. But, I just want to know the consensus behind. Do we have any agreement from others?

src/otx/v2/adapters/torch/mmengine/engine.py
jaegukhyun: I think postprocessing would be different on each model. So each model should have own codebase_config or postprocess_config to export itself. However current otxv2 do not have class for model api. Therefore I use unify config for all model, but in later phase model api should provide it's own config for export to engine. And I don't have precise plan for this

src/otx/v2/api/core/engine.py
jaegukhyun: We don't have agreement of this and I also cautious to add parameter to global api. However, I believe engine should have task attribute if our design is a single engine across all tasks.  
How about discuss about this in this PR or execution forum?

src/otx/v2/adapters/torch/mmengine/engine.py
```diff
@@ -649,6 +644,24 @@ def _update_codebase_config(
...
+        mmdeploy_task_dict = {
+            TaskType.SEGMENTATION: "Segmentation",
+            TaskType.CLASSIFICATION: "Classification",
+            TaskType.DETECTION: "ObjectDetection",
+            TaskType.INSTANCE_SEGMENTATION: "InstanceSegmentation",
+        }
+        codebase_config: dict[str, str | dict] = {"type": codebase, "task": mmdeploy_task_dict[self.task]}
+        if self.task == TaskType.DETECTION:
+            # Detection task require post_processing config
+            # This should be handled model api
```
samet-akcay, line 657:
```suggestion
            # This should be handled by model api
```

src/otx/v2/adapters/torch/mmengine/engine.py
```diff
@@ -649,6 +644,24 @@ def _update_codebase_config(
...
         """
-        codebase = codebase if codebase is not None else self.registry.name
-        codebase_config = {"type": codebase, "task": task}
+        codebase = self.registry.name
+        mmdeploy_task_dict = {
+            TaskType.SEGMENTATION: "Segmentation",
+            TaskType.CLASSIFICATION: "Classification",
+            TaskType.DETECTION: "ObjectDetection",
+            TaskType.INSTANCE_SEGMENTATION: "InstanceSegmentation",
+        }
```
samet-akcay, line 653: would there be another way to store this task info? `TaskType` already has these tasks, right? Why manual adding?

src/otx/v2/api/core/engine.py
harimkang: If I remember correctly, we had already agreed to use the `MMXEngine` as one (If we clean up our API of Exporting & Prediction later), and this could be part of that process. + Also, `LightningEngine` already uses this approach.
For now, while the actual Engine usage example doesn't change, but it would be nice to have a discussion in the future about whether we should always include a task when creating engine.

harimkang (approved):
LGTM, Thank you for your efforts.

wonjuleee (approved):

tests/v2/integration/api/auto_runner/test_auto_runner.py
```diff
@@ -5,12 +5,23 @@
...
+from otx.v2.api.entities.task_type import TaskType
 
 from tests.v2.integration.test_helper import TASK_CONFIGURATION
 
 # NOTE: This test currently only checks the basic pipeline and doesn't do much checking on the result, which will be fixed in the future.
 
 class TestAutoRunnerAPI:
+    @pytest.fixture(autouse=True)
+    def setup(self):
+        self.task_type_dict = {
```
samet-akcay, line 17: again, not sure if i understand why we need this manual mapping. would it be possible to get this directly from the enum?

samet-akcay (commented):
Thanks @jaegukhyun

src/otx/v2/adapters/torch/lightning/anomalib/engine.py
```diff
@@ -38,16 +39,16 @@ class AnomalibEngine(LightningEngine):
 
     def __init__(
         self,
+        task: TaskType,
```
harimkang, line 42: I'm not sure if having an Enum called TaskType is really convenient, because the user has to import the TaskType once to use the API.
What about taking a str as well and converting it?

src/otx/v2/adapters/torch/mmengine/engine.py
harimkang: This may not be directly related to this comment,
I think the `TaskType` is a necessary type internally, but not a good idea for users to use directly. `TaskType` should only be used internally and the user should be able to send and receive it as an input like str. What do you think?

src/otx/v2/adapters/torch/lightning/anomalib/engine.py
sungmanc: I agree with this.

harimkang: It's probably best to get both, and convert the str to TaskType in the case of str.
we can use below
`from otx.v2.api.utils.type_utils import str_to_task_type, str_to_train_type`

PR #2361, opened 2023-07-14
Enable `deterministic` for anomalib and visual prompting

sungchul1:
### Summary

Those tasks are implemented based on pytorch lightning.
In order to use `deterministic=True`, it should be set as `deterministic=warn` for `Trainer` (because otx supports `torch==1.13`), but otx only gets `deterministic` argument as `bool`.
So, inner patching in `task.train` is required to use `deterministic=True`.

Reference:
https://github.com/Lightning-AI/lightning/blob/a020506a81a671e380872aa5e435647b549e14e6/src/pytorch_lightning/trainer/connectors/accelerator_connector.py#L222-L227

- [x] Update `deterministic` from `False` to `True`
- [x] Remove unused pot test modules in visual prompting e2e test

### How to test

```shell
$ tox -e tests-ano-py310 -- tests/integration/cli/anomaly
================================================================= short test summary info ==================================================================FAILED tests/integration/cli/anomaly/test_anomaly_classification.py::TestToolsAnomalyClassification::test_nncf_optimize[ote_anomaly_classification_stfpm] - AssertionError: Traceback (most recent call last):
FAILED tests/integration/cli/anomaly/test_anomaly_classification.py::TestToolsAnomalyClassification::test_nncf_optimize[ote_anomaly_classification_padim] - AssertionError: Traceback (most recent call last):
FAILED tests/integration/cli/anomaly/test_anomaly_detection.py::TestToolsAnomalyDetection::test_nncf_optimize[ote_anomaly_detection_stfpm] - AssertionError: Traceback (most recent call last):
FAILED tests/integration/cli/anomaly/test_anomaly_detection.py::TestToolsAnomalyDetection::test_nncf_optimize[ote_anomaly_detection_padim] - AssertionError: Traceback (most recent call last):
FAILED tests/integration/cli/anomaly/test_anomaly_segmentation.py::TestToolsAnomalySegmentation::test_nncf_optimize[ote_anomaly_segmentation_stfpm] - AssertionError: Traceback (most recent call last):
FAILED tests/integration/cli/anomaly/test_anomaly_segmentation.py::TestToolsAnomalySegmentation::test_nncf_optimize[ote_anomaly_segmentation_padim] - AssertionError: Traceback (most recent call last):
=================================================== 6 failed, 57 passed, 1 warning in 723.59s (0:12:03) ====================================================
-> All fails occurred because nvcc is unavailable in ikvensx010

$ tox -e tests-visprompt-py310 -- tests/integration/cli/visual_prompting
[...]

PR #2664, opened 2023-11-21
Fix determinsticness for anomaly tasks

djdameln:
### Summary

This addresses problems with the `--determinstic` flag in anomaly tasks. 

The issue was caused by the use of `torch.cumsum` in the computation of a validation metric.
Since the validation metric was not critical to the model pipeline, we could safely remove this computation.

### Checklist

### License

```python
```


src/otx/algorithms/anomaly/adapters/anomalib/config/anomalib_config.py
```diff
@@ -41,6 +44,13 @@ def get_anomalib_config(task_name: str, otx_config: ConfigurableParameters) -> U
     """
     config_path = Path(anomalib.__file__).parent / "models" / task_name.lower() / "config.yaml"
     anomalib_config = get_configurable_parameters(model_name=task_name.lower(), config_path=config_path)
+
+    try:
+        anomalib_config.metrics.image.remove("AUROC")
+        anomalib_config.metrics.pixel.remove("AUROC")
+    except ValueError:
+        logger.info("Tried removing AUROC metric from anomalib config, but metric was not found.")
```
sungmanc: Is the `AUROC`  the mandatory thing? 
If `torch.cumsum` included in the `AUROC` really matter, how about setting `deterministic=False` instead of `ValueError` if there is an `AUROC` metric  ?

eunwoosh: Is it ok to remove "AUROC" metric? When I checked configuration.yaml in anomaly config directory, AUROC can be used as early stopping metric. Does early stopping work if user decides to use AUROC after this change is applied?

src/otx/algorithms/anomaly/tasks/train.py
```diff
@@ -67,7 +67,7 @@ def train(
         if seed:
             logger.info(f"Setting seed to {seed}")
             seed_everything(seed, workers=True)
-        config.trainer.deterministic = "warn" if deterministic else deterministic
+        config.trainer.deterministic = deterministic
```
sungchul1: I modified this at https://github.com/openvinotoolkit/training_extensions/pull/2361 to avoid errors about operations that deterministic doesn't support.
Could you explain why you reverted this part?

yunchu:
@djdameln could you check whether this PR is still valid or not? then if it's not feasible, please close this PR.

PR #2676, opened 2023-11-29
Fix unsupported dtype in ov graph constant converter

### LINKED ISSUE #2675: Failures on unittests for otx.core.ov modules
**Describe the bug**

After upgrading openvino version to 2023.1, downloaded omz models were compressed with option of mo "--compress_to_fp16 true" as default. this leads some unittests failure.
ref. [openvino 2023.1 release note](https://www.intel.com/content/www/us/en/developer/articles/release-notes/openvino/2023-1.html)
image

**Steps to Reproduce**

1. delete cached omz model file if it exists 

```bash
$ rm -rf ~/.cache/otx/omz
```

2. run unittests

```bash
$ tox -vvv -e unittest-all-py310-pt1 -- tests/unit/core/ov/graph/test_ov_graph_utils.py::test_handle_reshape
```

3. TypeError will be occurred

```
========================================================================== short test summary info ==========================================================================
FAILED tests/unit/core/ov/graph/test_ov_graph_utils.py::test_handle_reshape - TypeError: can't convert np.ndarray of type numpy.uint16. The only supported types are: float64, float32, float16, complex64, complex128, int64, int32, int16, int8, uin...
```

**Environment:**

- OS: Ubuntu 20.04
- Framework version: Pytorch
- Python version: 3.9
- OpenVINO version: 2023.1
- CUDA/cuDNN version: 11.7
- GPU model and memory:

sovrasov:
### Summary

workarounds #2675

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/core/ov/ops/infrastructures.py
+++ b/src/otx/core/ov/ops/infrastructures.py
@@ -233,6 +233,8 @@ class ConstantV0(Operation[ConstantV0Attribute]):
             if not np.array_equal(data, data_):
                 logger.warning(f"Overflow detected in {op_name}")
             data = torch.from_numpy(data_)
+        elif data.dtype == np.uint16:
+            data = torch.from_numpy(data.astype(np.int32))
         else:
             data = torch.from_numpy(data)
 
--- a/src/otx/core/ov/ops/type_conversions.py
+++ b/src/otx/core/ov/ops/type_conversions.py
@@ -25,6 +25,7 @@ _ov_to_torch = {
     "u1": torch.uint8,  # no type in torch
     "u4": torch.uint8,  # no type in torch
     "u8": torch.uint8,
+    "u16": torch.int32,  # no type in torch
     "u32": torch.int32,  # no type in torch
     "u64": torch.int64,  # no type in torch
     "i4": torch.int8,  # no type in torch
--- a/tests/unit/core/ov/graph/test_ov_graph_utils.py
+++ b/tests/unit/core/ov/graph/test_ov_graph_utils.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
+import pytest
 from otx.core.ov.graph.graph import Graph
 from otx.core.ov.graph.utils import (
     get_constant_input_nodes,
@@ -38,6 +39,7 @@ def test_handle_merging_into_batchnorm():
 
 
 @e2e_pytest_unit
+@pytest.mark.skip(reason="Updated models are not compatible with the paired batchnorm converter")
 def test_handle_paired_batchnorm():
     graph = get_graph()
     handle_paired_batchnorm(graph)


goodsong81 (dismissed):

sungmanc (dismissed):
Minor question, Is it okay to assign more bit ?

src/otx/core/ov/ops/infrastructures.py
```diff
@@ -233,6 +233,8 @@ def from_ov(cls, ov_op):
             if not np.array_equal(data, data_):
                 logger.warning(f"Overflow detected in {op_name}")
             data = torch.from_numpy(data_)
+        elif data.dtype == np.uint16:
+            data = torch.from_numpy(data.astype(np.int32))
```
yunchu, line 237:
```suggestion
            data = torch.from_numpy(data.astype(np.int32))
```

src/otx/core/ov/ops/infrastructures.py
```diff
@@ -233,6 +233,8 @@ def from_ov(cls, ov_op):
             if not np.array_equal(data, data_):
                 logger.warning(f"Overflow detected in {op_name}")
             data = torch.from_numpy(data_)
+        elif data.dtype == np.uint16:
```
yunchu, line 236: @sovrasov I have little concern about this changes to resolving https://github.com/openvinotoolkit/training_extensions/issues/2675 because it caused by updating the default behavior of the model optimizer (mo) in the openvino 2023.1 (turned on as true to "--compress_to_fp16" option). I think the intention to the core module's unitests seemed to expecting uncompressed model as input but this change will not fix this thing.
```suggestion
        elif data.dtype == np.uint16:
```

yunchu (changes requested):

src/otx/core/ov/ops/infrastructures.py
sovrasov: Do you mean that with these changes some other tests are expected to fail?

yunchu: no. we can check that all tests were passed with you change but what i meant is that the data type 'unintXX' is the artifact of the compression by model optimizer. so the way that i expect is that should be omitting compression option from the graph generation flow.
but as i've mentioned before, i don't have enough idea about the things in the otx.core.ov modules so if you think this is the correct changes, you can merge it. these comments from me are only my point of view to see the root cause of this issue.

yunchu (dismissed):

src/otx/core/ov/ops/infrastructures.py
sovrasov: Yeah, I see your point. In my opinion, dtype compatibility problem is not a big deal, since we can always substitute a wider type when building a torch model.
I discovered one more failing test, and it raises even a bigger concern: graph patterns that were hardcoded in the module are not valid for the newly generated IRs. I'll skip one more test, because graph transformation carried out there is no longer applicable.

sovrasov:
@yunchu thanks for your comments, I removed "resolves" from the PR description, since graph parser is not actually fully compatible with the graph structure generated by the new versions of model converter tool in OV

yunchu (approved):

yunchu:
@sovrasov if there is no more discussion onto this PR, i think you can ask to get review from the peer again for merging. how do you think?

sovrasov:
> @sovrasov if there is no more discussion onto this PR, i think you can ask to get review from the peer again for merging. how do you think?

I forgot about this PR, sorry

@goodsong81 could you have a look?

goodsong81 (approved):

PR #2677, opened 2023-11-29
Add XPU support to anomaly task

sovrasov:
### Summary
- Training is done in FP32, as on CUDA
- This solution is temporal: device selection for batch should be done on lightning side, but apparently the lighting version we use can't do that.

### Checklist

### License

```python
```


src/otx/algorithms/anomaly/adapters/anomalib/callbacks/xpu.py
```diff
@@ -0,0 +1,39 @@
...
+
+    Applies IPEX optimization before training, moves data to XPU.
+    """
+
+    def __init__(self, device_idx=0):
+        self.device = torch.device(f"xpu:{device_idx}")
+
+    def on_fit_start(self, trainer, pl_module):
+        """Applies IPEX optimization before training."""
+        if is_xpu_available():
```
eunwoosh: I think this code isn't necessary, because this callback will be used only if xpu is available. How do you think about?

wonjuleee (approved):
LGTM

src/otx/algorithms/anomaly/adapters/anomalib/callbacks/xpu.py
```diff
@@ -0,0 +1,39 @@
...
+            pl_module.to(self.device)
+            model, optimizer = torch.xpu.optimize(trainer.model, optimizer=trainer.optimizers[0], dtype=torch.float32)
+            trainer.optimizers = [optimizer]
+            trainer.model = model
+
+    def on_train_batch_start(self, trainer, pl_module, batch, batch_idx):
+        """Moves train batch tensors to XPU."""
+        for k in batch:
+            if not isinstance(batch[k], list):
```
sungchul1, line 29: I'm not sure if there is a case where `batch[k]` is `list`, but if so, wouldn't we consider that too?

src/otx/algorithms/anomaly/adapters/anomalib/callbacks/xpu.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+
+    def __init__(self, device_idx=0):
+        self.device = torch.device(f"xpu:{device_idx}")
+
+    def on_fit_start(self, trainer, pl_module):
+        """Applies IPEX optimization before training."""
+        if is_xpu_available():
+            pl_module.to(self.device)
+            model, optimizer = torch.xpu.optimize(trainer.model, optimizer=trainer.optimizers[0], dtype=torch.float32)
```
sungchul1: (minor) If we use fp32 model, we don't need to assign `dtype` arg.

sungchul1 (commented):
LGTM. I left simple comments.

src/otx/algorithms/anomaly/adapters/anomalib/callbacks/xpu.py
sovrasov: That happens for some of the models (detection in particular), otherwise I wouldn't have considered that case. This callback works only with anomaly models which we have right now, so this is not a general solution.

src/otx/algorithms/anomaly/adapters/anomalib/callbacks/xpu.py
sovrasov: I removed extra if

src/otx/algorithms/anomaly/adapters/anomalib/callbacks/xpu.py
sovrasov: It's a big question if we should always rely on default arg values or not in general. Implicit dtype could be useful for understanding that this callback works only with fp32 for those who're not that closely familiar to IPEX API.

sungchul1 (approved):

eunwoosh (approved):

PR #2683, opened 2023-12-01
Disable mixed precision training on XPU

eunwoosh:
### Summary
As @sungchul2 said, Current IPEX doesn't have GradScaler implementation for XPU.
For this reason, this PR disables mixed precision and bfp16 training on XPU.
And current `XPUDataParallel` class is using `torch.autocast` but it's duplication because `auto_fp16` is already using it.
So, I remove it in the class.
And I updated minor one related to 'otx optimize'

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algorithms/common/adapters/mmcv/configurer.py
+++ b/src/otx/algorithms/common/adapters/mmcv/configurer.py
@@ -264,18 +264,11 @@ class BaseConfigurer:
                 opts["type"] = "HPUOptimizerHook"
                 cfg.optimizer_config.update(opts)
             elif is_xpu_available():
-                opts.update({"distributed": distributed, **fp16_config})
                 if optim_type == "SAMOptimizerHook":
                     logger.warning("SAMOptimizerHook is not supported on XPU yet, changed to OptimizerHook.")
                     opts["type"] = "OptimizerHook"
-                if optim_type == "OptimizerHook":
-                    opts["type"] = "BFp16XPUOptimizerHook"
-                else:
-                    # does not support optimizerhook type
-                    # let mm library handle it
-                    cfg.fp16 = fp16_config
-                    opts = dict()
                 cfg.optimizer_config.update(opts)
+                logger.warning("XPU doesn't support mixed precision training currently.")
             elif torch.cuda.is_available():
                 opts.update({"distributed": distributed, **fp16_config})
                 if optim_type == "SAMOptimizerHook":
--- a/src/otx/algorithms/segmentation/adapters/mmseg/apis/train.py
+++ b/src/otx/algorithms/segmentation/adapters/mmseg/apis/train.py
@@ -51,6 +51,9 @@ def train_segmentor(model, dataset, cfg, distributed=False, validate=False, time
     train_loader_cfg = {**loader_cfg, **cfg.data.get("train_dataloader", {})}
     data_loaders = [build_dataloader(ds, **train_loader_cfg) for ds in dataset]
 
+    if cfg.device == "xpu":
+        model.to(f"xpu:{cfg.gpu_ids[0]}")
+
     # put model on devices
     if distributed:
         find_unused_parameters = cfg.get("find_unused_parameters", False)
@@ -68,11 +71,6 @@ def train_segmentor(model, dataset, cfg, distributed=False, validate=False, time
             assert digit_version(mmcv.__version__) >= digit_version(
                 "1.4.4"
             ), "Please use MMCV >= 1.4.4 for CPU training!"
-
-        if cfg.device == "xpu":
-            use_autocast = bool(cfg.get("fp16_", False))
-            model = build_dp(model, cfg.device, device_ids=cfg.gpu_ids, enable_autocast=use_autocast)
-            model.to(f"xpu:{cfg.gpu_ids[0]}")
         elif cfg.device == "hpu":
             use_autocast = bool(cfg.get("fp16_", False))
             model = build_dp(model, cfg.device, device_ids=cfg.gpu_ids, enable_autocast=use_autocast)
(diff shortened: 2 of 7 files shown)


src/otx/algorithms/common/adapters/mmcv/configurer.py
```diff
@@ -264,18 +264,11 @@ def configure_fp16(cfg: Config):
                 opts["type"] = "HPUOptimizerHook"
                 cfg.optimizer_config.update(opts)
             elif is_xpu_available():
-                opts.update({"distributed": distributed, **fp16_config})
                 if optim_type == "SAMOptimizerHook":
                     logger.warning("SAMOptimizerHook is not supported on XPU yet, changed to OptimizerHook.")
                     opts["type"] = "OptimizerHook"
-                if optim_type == "OptimizerHook":
-                    opts["type"] = "BFp16XPUOptimizerHook"
```
sungchul1, line 272: I think removing `cfg.fp16_` is required as well because of setting `enable_autocast` to `False`.

For example,
https://github.com/openvinotoolkit/training_extensions/blob/c1beb0549e58f18c3e9f17097d2a02e52bd81231/src/otx/algorithms/detection/adapters/mmdet/apis/train.py#L121-L123

sungchul1 (commented):
I left a comment.
And I didn't check if `autocast` in `XPUDataParallel` is really necessary because I customizably patched `auto_fp16` to `custom_auto_fp16`.
Could you check it? If it is unnecessary, we can remove this part.

https://github.com/openvinotoolkit/training_extensions/blob/c1beb0549e58f18c3e9f17097d2a02e52bd81231/src/otx/algorithms/common/adapters/mmcv/utils/_builder_build_data_parallel.py#L127-L139

src/otx/algorithms/common/adapters/mmcv/configurer.py
eunwoosh: I think `torch.autocast` in `XPUDataParallel` should be removed. When I checked `custom_auto_fp16` code it already uses `torch.autocast`. So, it's duplicated. After removing it, we don't have to remove `cfg.fp16_` because `enable_autocast` disappear after update.

sungchul1: I agree with that.

eunwoosh: done at [@48f463a](https://github.com/openvinotoolkit/training_extensions/pull/2683/commits/48f463ac9682ce6a0f550420d33c0201d4f52291)

sungchul1 (approved):

src/otx/algorithms/segmentation/adapters/mmseg/apis/train.py
```diff
@@ -64,15 +67,10 @@ def train_segmentor(model, dataset, cfg, distributed=False, validate=False, time
...
     else:
-        if not torch.cuda.is_available():
+        if not torch.cuda.is_available():  # noqa
             assert digit_version(mmcv.__version__) >= digit_version(
                 "1.4.4"
             ), "Please use MMCV >= 1.4.4 for CPU training!"
-
-        if cfg.device == "xpu":
-            use_autocast = bool(cfg.get("fp16_", False))
-            model = build_dp(model, cfg.device, device_ids=cfg.gpu_ids, enable_autocast=use_autocast)
```
sovrasov, line 74: Why was building of XPU dp deleted?

sovrasov (commented):
@sungchul2 do I understand it correctly, that from your experiments it's easier to disable BFP16 at all because it completely doesn't work in all the cases?

src/otx/algorithms/segmentation/adapters/mmseg/apis/train.py
eunwoosh: `build_dp` is executed at L79. I removed it because after `enable_autocast` argument is removed, code is exactly same as L79.

eunwoosh: `build_dp` is executed at L79. I removed it because after `enable_autocast` argument is removed, code is exactly same as L79.

eunwoosh:
> @sungchul2 do I understand it correctly, that from your experiments it's easier to disable BFP16 at all because it completely doesn't work in all the cases?

As you know, to train a model with fp16 or bfp16, 'autocast' and 'GradScaler' are needed. But current torch implemented GradScaler only for cpu and cuda and also IPEX implemented it only for cpu. So, there is no GradScaler for XPU currently.

sungchul1:
> @sungchul2 do I understand it correctly, that from your experiments it's easier to disable BFP16 at all because it completely doesn't work in all the cases?

As Eunwoo said, current official GradScaler for XPU is not implemented but it may be required for stable training.
So, I temporarily implemented it and saw that it made training more stable.
But as we discussed before, we decided to check FP32 results only, so as far as I know this PR's purpose is not to use any other precision training for now.

sungchul1:
> > @sungchul2 do I understand it correctly, that from your experiments it's easier to disable BFP16 at all because it completely doesn't work in all the cases?
> 
> As Eunwoo said, current official GradScaler for XPU is not implemented but it may be required for stable training. So, I temporarily implemented it and saw that it made training more stable. But as we discussed before, we decided to check FP32 results only, so as far as I know this PR's purpose is not to use any other precision training for now.

Of course, we can enable FP32 training by setting `fp16` to `False` or `None` in `model.py`, but it is so much hassle as you know.

kprokofi:
Why do we want to remove even the possibility of training in fp16? Maybe it is OK to run fp32 training by default, but I don't like removing all the functionality related to fp16 from the code + we don't delete custom GradScaler  and all stuff related to it.
So, we either delete everything or keep everything

eunwoosh:
> Why do we want to remove even the possibility of training in fp16? Maybe it is OK to run fp32 training by default, but I don't like removing all the functionality related to fp16 from the code + we don't delete custom GradScaler and all stuff related to it. So, we either delete everything or keep everything

The only removed part regarding bfp16 training is 'src/otx/algorithms/common/adapters/mmcv/configurer.py'. bfp16 training is configured in configurer, so I removed some code to prevent user from enabling bfp16 training. 'api/trian.py' is just about refactoring and 'src/otx/algorithms/common/adapters/mmcv/utils/_builder_build_data_parallel.py' is just about removing duplicated code.

src/otx/algorithms/segmentation/adapters/mmseg/apis/train.py
eunwoosh: Sorry there was a wrong implementation. I fixed at https://github.com/openvinotoolkit/training_extensions/pull/2683/commits/5f691522378f41650531e846bbe087dd4f86faa6

kprokofi (approved):

sovrasov (approved):

PR #2695, opened 2023-12-05
Add Simple Integration test (CLI - 'otx train')

harimkang:
### Summary

- Add CLI Integration Test using recipes
    - 1 epochs training with all recipes
    - Currently, only classification and detection work
- Add Classification test assets dataset: `tests/assets/classification_dataset`
- Add `trainer.device=auto` setting
- Add tox integration-test
- Add integration test into CI
- Fix install issue of case without 'version.json' in '/cuda/'

### How to test

- TOX test
`tox -vv -e integration-test`

- Local Test
`python -m pytest tests/integration/cli/test_cli.py`
image

### Checklist

### License

```python
```


jaegukhyun (dismissed):

tests/integration/cli/test_cli.py
```diff
@@ -0,0 +1,67 @@
...
+# This assumes have OTX installed in environment.
+otx_module = importlib.import_module("otx")
+RECIPE_PATH = Path(inspect.getfile(otx_module)).parent / "recipe"
+RECIPE_LIST = [str(_.relative_to(RECIPE_PATH)) for _ in RECIPE_PATH.glob("**/*.yaml")]
+
+# [TODO]: This is a temporary approach.
+DATASET = {
+    "classification": {
+        "data_dir": "tests/assets/classification_dataset",
+        "overrides": ["model.otx_model.config.head.num_classes=2"],
```
sungmanc:
```suggestion
        "overrides": ["model.otx_model.config.head.num_classes=2", "model.otx_model.config.data_preprocessor.num_classes=2"],
```

harimkang: Thanks! - https://github.com/openvinotoolkit/training_extensions/pull/2695/commits/b5399f75f1f726cbae28339c8a83e455864acb02

sungmanc (approved):
LGTM

jaegukhyun (approved):

kprokofi (approved):
In general, LGTM. But it might be worth extending the tests by 2 epochs instead of 1 to make sure that any processes between epochs also work.
What do you think?

harimkang:
> In general, LGTM. But it might be worth extending the tests by 2 epochs instead of 1 to make sure that any processes between epochs also work. What do you think?

I think it's a good idea, but for now I'll merge this PR for another task model recipe integration test and create a new PR after we discuss it together.

tests/integration/cli/test_helper.py
```diff
@@ -0,0 +1,88 @@
...
+    async def run_and_capture(cmd: list) -> tuple:
+        environ = os.environ.copy()
+        environ["PYTHONUNBUFFERED"] = "1"
+        process = await asyncio.create_subprocess_exec(
+            *cmd,
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.PIPE,
+            env=environ,
+            **kwargs,
+        )
```
vinnamkim, line 55: I do not want to spawn a separate process for the integration test. This is because it makes a developer harder to debug if a problem is found during the integration test. Please replace it with executing the CLI Python function in the main process.

tests/integration/cli/test_cli.py
```diff
@@ -0,0 +1,70 @@
...
+    task = recipe.split("/")[0]
+    model_name = recipe.split("/")[1].split(".")[0]
+    tmp_path = tmp_path / f"otx_train_{model_name}"
+    command_line = [
+        "otx", "train",
+        f"+recipe={recipe}",
+        f"base.data_dir={DATASET[task]['data_dir']}",
+        "trainer.max_epochs=1",
+        "trainer.check_val_every_n_epoch=1",
+        "trainer=auto",
```
vinnamkim, line 55: Rather than manually overriding at the variable level, consider to override this https://github.com/openvinotoolkit/training_extensions/blob/v2/src/otx/config/debug/default.yaml

src/otx/config/trainer/auto.yaml
vinnamkim, line 1: Why is this needed?

tests/integration/cli/test_cli.py
harimkang: Thanks for the comment. I will create a PR to reflect your comments. I have a question regarding this comment. I want to create a config file called `debug/intg_test.yaml` and want to apply it to test. 
Can you tell me what args should be added to override the config in this debug folder in the cli?

src/otx/config/trainer/auto.yaml
harimkang: I wanted to make sure that CI would work whether the test was using cpu or gpu, so I thought using the `auto` provided by lightning for the test would be appropriate. What do you think? This too can be modified as per your suggestion.

tests/integration/cli/test_cli.py
vinnamkim: I guess that you need to add `debug` optional field to `TrainConfig` as same as `recipe`.
image
Then, it might be working by adding argument `+debug=intg_test`. This is exactly same with what we do with `+recipe=???`.

src/otx/config/trainer/auto.yaml
vinnamkim: Is that possible by giving `trainer=cpu` or `trainer=gpu` explicitly outside? If you set it as `auto`, then it seems not traceable which device we used for the tests.

tests/integration/cli/test_cli.py
harimkang: Thank you. I'll give it a try.

src/otx/config/trainer/auto.yaml
harimkang: For test, yes i understand, I'm going to modify it in intg-test to gpu for now. Do you want to remove this file?
I guess `auto` might be a good option for the user as well.

vinnamkim: Then, let it leave as is. However, the decision for which device will be used should be explicit in the tests. This is what I only concern.

harimkang:
@vinnamkim I create next PR https://github.com/openvinotoolkit/training_extensions/pull/2701

PR #2698, opened 2023-12-05
Add Instance Segmentation models to new design

eugene123tw:
### Checklist

### License

```python
```


harimkang:
Thanks for the good work. I have added an integration test (https://github.com/openvinotoolkit/training_extensions/pull/2695), to make the recipe work as well, please see https://github.com/openvinotoolkit/training_extensions/blob/ef7fdfb163c8ee06654e987485a56c0650b14e5b/tests/integration/cli/test_cli.py#L19 to add and set up a dataset for the task. Let me know if you need any help.

src/otx/core/data/dataset/instance_segmentation.py
```diff
@@ -0,0 +1,76 @@
...
+
+        gt_bboxes = np.zeros(shape=(0, 4), dtype=np.float32)
+        gt_labels = np.zeros(shape=(0, ), dtype=int)
+        gt_masks = np.zeros(shape=(0, *img_shape), dtype=bool)
+        gt_polygons = []
+        for annotation in item.annotations:
+            if isinstance(annotation, Polygon):
+                points = np.array(annotation.points).reshape(-1, 2).astype(np.int32)
+                x1, y1 = np.min(points, axis=0)
+                x2, y2 = np.max(points, axis=0)
```
vinnamkim:
```suggestion
                bbox = np.array(annotation.get_bbox(), dtype=np.int32) # xywh
```

1. Please reuse `Datumaro`'s API to make it looking clean
2. Please `np.array(..., dtype=...)` rather than `np.array(...).astype(...)` because there is an overhead to convert the Numpy array's dtype after construction.

src/otx/core/data/dataset/instance_segmentation.py
```diff
@@ -0,0 +1,76 @@
...
+                gt_bboxes = np.vstack((gt_bboxes, np.array([x1, y1, x2, y2])))
+                gt_labels = np.append(gt_labels, annotation.label)
```
vinnamkim: Please do not `np.vstack` and `np.append` in the for-loop. This will have `O(n^2)` complexity for copy operations.

src/otx/core/data/dataset/instance_segmentation.py
```diff
@@ -0,0 +1,76 @@
...
+
+from .base import OTXDataset, Transforms
+
+
+class OTXInstanceSegDataset(OTXDataset[InstanceSegDataEntity]):
+    """OTXDataset class for instance segmentation."""
+
+    def __init__(self, dm_subset: DatasetSubset, transforms: Transforms) -> None:
+        super().__init__(dm_subset, transforms)
+        self.poly2mask = 'polygon' not in dm_subset.get_annotated_type()
```
vinnamkim: Why is this needed?

src/otx/core/data/dataset/instance_segmentation.py
```diff
@@ -0,0 +1,76 @@
...
+                x1, y1 = np.min(points, axis=0)
+                x2, y2 = np.max(points, axis=0)
+                gt_bboxes = np.vstack((gt_bboxes, np.array([x1, y1, x2, y2])))
+                gt_labels = np.append(gt_labels, annotation.label)
+                if self.poly2mask:
+                    mask = polygon_to_bitmap([annotation], *img_shape)
+                    gt_masks = np.vstack((gt_masks, mask))
+                else:
+                    gt_polygons.append(annotation)
+
```
vinnamkim, line 61:
```suggestion
      # From xywh to xyxy
      gt_bboxes[:, 2] += gt_bboxes[:, 0] 
      gt_bboxes[:, 3] += gt_bboxes[:, 1]
```

src/otx/core/data/dataset/instance_segmentation.py
```diff
@@ -0,0 +1,76 @@
...
+        for annotation in item.annotations:
+            if isinstance(annotation, Polygon):
+                points = np.array(annotation.points).reshape(-1, 2).astype(np.int32)
+                x1, y1 = np.min(points, axis=0)
+                x2, y2 = np.max(points, axis=0)
+                gt_bboxes = np.vstack((gt_bboxes, np.array([x1, y1, x2, y2])))
+                gt_labels = np.append(gt_labels, annotation.label)
+                if self.poly2mask:
+                    mask = polygon_to_bitmap([annotation], *img_shape)
+                    gt_masks = np.vstack((gt_masks, mask))
```
vinnamkim: Same here. You are trying to invoke copy operation for 3d tensor in the for-loop => `O(m^2*h*w)`, `m` is the number of masks and `h*w` is the image size. This is enormously heavy since we are manipulating images.

src/otx/core/data/entity/instance_segmentation.py
```diff
@@ -0,0 +1,93 @@
...
+
+    @property
+    def task(self) -> OTXTaskType:
+        """OTX Task type definition."""
+        return OTXTaskType.INSTANCE_SEGMENTATION
+
+    bboxes: tv_tensors.BoundingBoxes
+    masks: tv_tensors.Mask
+    labels: LongTensor
+    polygons: list[Polygon]
```
vinnamkim, line 40: It is not desired to include non-`torch.Tensor` or `tv_tensors.*` here. Would you please change it to `list[torch.FloatTensor]` if needed?

src/otx/core/data/entity/instance_segmentation.py
```diff
@@ -0,0 +1,93 @@
...
+
+    bboxes: list[tv_tensors.BoundingBoxes]
+    masks: list[tv_tensors.Mask]
+    labels: list[LongTensor]
+    polygons: list[list[Polygon]]
+
+    @property
+    def task(self) -> OTXTaskType:
+        """OTX Task type definition."""
+        return OTXTaskType.DETECTION
```
vinnamkim:
```suggestion
        return OTXTaskType.INSTANCE_SEGMENTATION
```

src/otx/core/data/transform_libs/mmdet_inst_seg.py
vinnamkim, line 1: Please change here as well according to the entity changes which I left as comments above.

src/otx/core/model/entity/instance_segmentation.py
```diff
@@ -0,0 +1,149 @@
...
+            entity.imgs_info,
+            entity.bboxes,
+            entity.masks,
+            entity.polygons,
+            entity.labels,
+        ):
+            height, width = img_info.ori_shape
+            if len(masks):
+                mmdet_masks = BitmapMasks(
+                    masks.data.cpu().numpy(), height, width)
```
vinnamkim, line 67: Why do you send `masks` back from the GPU memory to the host memory? It will make this code blocking and bring undesired latency (and this is huge please remember `masks` is a `h*w` tensor).

src/otx/core/model/module/instance_segmentation.py
```diff
@@ -0,0 +1,152 @@
...
+            })
+
+        for imgs_info, bboxes, masks, polygons, labels in zip(
+            inputs.imgs_info,
+            inputs.bboxes,
+            inputs.masks,
+            inputs.polygons,
+            inputs.labels,
+        ):
+            bit_masks = masks if len(masks) else polygon_to_bitmap(polygons, *imgs_info.ori_shape)
```
vinnamkim, line 121: We cannot execute `polygon_to_bitmap` in the main thread (training loop). This is a time-consuming job and should be done in the data preparation threads (not main thread) in parallel and asynchronously. Please do this at the data preparation stage.

src/otx/core/data/dataset/instance_segmentation.py
vinnamkim: According to this change https://github.com/openvinotoolkit/training_extensions/pull/2698#discussion_r1418198970

harimkang:
@eugene123tw Some callbacks have been hooked up to the latest commit.
https://github.com/openvinotoolkit/training_extensions/pull/2703

Since we need to set a metric for each task, something similar to the below should be done for that PR as well.
https://github.com/openvinotoolkit/training_extensions/commit/e549e2643da0e94a43e8138b0d891f5cdcd49148
You can either write a callback for instance_segmentation or use detection (val/map) as is.
Instead, you'll need to add the callbacks override in all recipe files.

src/otx/core/data/dataset/instance_segmentation.py
eugene123tw: `self.poly2mask` serves the purpose of facilitating direct mask training. When processing item annotations, there are two possible scenarios: if a datumaro mask exists in annotated format, it will be utilized; otherwise, the polygon information will be employed for training.

In the provided code snippet:
```python
for annotation in item.annotations:
    if isinstance(annotation, Polygon):
        bbox = np.array(annotation.get_bbox(), dtype=np.int32)
        gt_bboxes.append(bbox)
        gt_labels.append(annotation.label)

        if self.poly2mask:
            gt_masks.append(polygon_to_bitmap([annotation], *img_shape)[0])
        else:
            gt_polygons.append(annotation.points)
```

src/otx/core/model/entity/instance_segmentation.py
eugene123tw: I'm aware of this issue. The current structure of MMDet masks, as outlined in their implementation :https://github.com/open-mmlab/mmdetection/blob/main/mmdet/structures/mask/structures.py#L643 , doesn't accommodate `torch.tensor`. The structure is designed to handle only numpy arrays. Consequently, I'm required to detach and convert torch.tensor to a numpy array for compatibility.

We have two potential courses of action: invest time in implementing a customized BitmapMasks that supports torch.tensor, or continue with the existing approach of detaching to CPU and converting to numpy.

It's crucial to note that this issue extends to polygons as well. If we opt to switch from `list[datumaro.Polygon]` to `list[torch.FloatTensor]` in `InstanceSegDataEntity`, we will encounter the same challenge of detaching from GPU to CPU and converting to numpy.

src/otx/core/data/entity/instance_segmentation.py
eugene123tw: I'm considering replacing `list[datumaro.Polygon]` with `list[torch.FloatTensor]`. But, it's essential to note that in the `MMDetInstanceSegCompatibleModel._customize_inputs` function, there's a requirement to convert torch CPU tensors to GPU tensors due to the limitation in mmdet.PolygonMasks, which exclusively accepts numpy arrays (refer to the implementation [link](https://github.com/open-mmlab/mmdetection/blob/main/mmdet/structures/mask/structures.py#L643)). As previously mentioned, this conversion may have implications for processing speed.

One critical consideration is that the data loading process will automatically convert any CPU torch tensor to a GPU torch tensor if a GPU device is in use. However, using `list[datumaro.Polygon]` would prevent this automatic conversion, as the collate function is designed to handle only torch tensors.

If you believe transitioning to `list[torch.FloatTensor]` is the preferred approach, and I will make changes accordingly. Alternatively, if you have any alternative suggestions or considerations, I'm open to exploring them.

src/otx/core/data/transform_libs/mmdet_inst_seg.py
eugene123tw: Please check my other comments related to this issue.

src/otx/core/data/dataset/instance_segmentation.py
eugene123tw: I changed this part accordingly.

src/otx/core/data/dataset/instance_segmentation.py
eugene123tw: Thanks for the suggestion. I changed this part.

src/otx/core/data/entity/instance_segmentation.py
vinnamkim: I studied about this. I launched the debugger on the `FCNMaskHead` to see how it is getting loss and `BitmapMasks`is used.

`BitmapMasks` turns into `mask_targets` which is `torch.Tensor` in `FCNMaskHead.get_targets()`:
image

The conversion actually happens in `mask_target.py`'s `mask_target_single` (`get_targets()` -> `mask_target()` -> `mask_target_single()`:
image

It means that we have two options:
1) We can customize `FCNMaskHead.get_targets()` and re-register it as same as we did in the transform part.
2) We can implement a custom `BitmapMasks` with customizing `crop_and_resize` as you said.

However, I agree that this can be handled in this PR and let it leave as the future work.

src/otx/core/data/dataset/instance_segmentation.py
vinnamkim:
```python
self.poly2mask = 'polygon' not in dm_subset.get_annotated_type()
```
However, from this line, `self.poly2mask=True` means that there is no `datumaro.Polygon` in the subset.
So, it seems that it cannot reach to L47-48 because of L40.

https://github.com/openvinotoolkit/training_extensions/blob/e86f6e433a5e8cdb421234806ba02c10c7d964a1/src/otx/core/data/dataset/instance_segmentation.py#L39-L50

Is it correct?

src/otx/core/model/module/instance_segmentation.py
vinnamkim: Can we move this to `PackInputs`? I didn't actually do it but it seems more degenerate to decode polygons to 2d masks in the main thread. I guess that we can endure `Tensor cpu->gpu->cpu` to move them away from the main thread.

src/otx/core/data/transform_libs/mmdet_inst_seg.py
```diff
@@ -0,0 +1,128 @@
...
+
+from .mmcv import MMCVTransformLib
+
+if TYPE_CHECKING:
+    from mmengine.registry import Registry
+
+    from otx.core.config.data import SubsetConfig
+
+
+@TRANSFORMS.register_module(force=True)
```
vinnamkim: This registration can affect to the detection task's behavior, which uses the `mmdet` data pipeline. I think that you need to unify this with `src/otx/core/data/transform_libs/mmdet.py`.

eugene123tw: Could you provide more details on how the registration of `LoadAnnotations` might impact the detection task? As per the current design, Instance Segmentation has introduced a distinct `MMDetInstSegTransformLib` separate from `detection.MMDetTransformLib`, indicating a deliberate divergence, and should work separately.

My concern is that unifying these processes might result in the nesting of `LoadAnnotation` and `PackDetInputs` with two different data entity checks (InstanceSegDataEntity and DetDataEntity). This distinction is crucial as detection tasks don't need masks, while instance segmentation tasks do. If we proceed with unification, PackDetInputs would then need to return two different data entities.

src/otx/core/data/dataset/instance_segmentation.py
eugene123tw: I agree. Currently, there's no clearly defined method for passing this parameter through the data configuration. We could consider two potential approaches: introducing a new parameter (`polygon2mask`) in `DataModuleConfig` or opting for a more straightforward solution by exclusively using `gt_polygons` and removing support for direct masks (or using gt_masks not gt_polygons)

An alternative approach is incorporating both `gt_masks` and `gt_polygons`, as demonstrated in the following snippet. However, it's essential to note that saving `gt_masks` in CUDA memory (num_objects * img_input_h * img_input_w) is more resource-intensive compared to polygons (num_objects * n_points * 2).

```python
for annotation in item.annotations:
    if isinstance(annotation, Polygon):
        bbox = np.array(annotation.get_bbox(), dtype=np.int32)
        gt_bboxes.append(bbox)
        gt_labels.append(annotation.label)
        gt_masks.append(polygon_to_bitmap([annotation], *img_shape)[0])
        gt_polygons.append(annotation)
```

Please let me know your thoughts.

src/otx/core/data/entity/instance_segmentation.py
eugene123tw: Thank you for your response.  Just to confirm, are you suggesting that we proceed with handling the optimization in the current pull request, or would you prefer to leave it as future work?

src/otx/core/model/module/instance_segmentation.py
eugene123tw: As highlighted in the previous comment, storing bitmap tensors in CUDA memory incurs a significant cost. Due to that, I've chosen to utilize polygons as an alternative. However, torchmetric requires bitmap masks, necessitating the conversion of polygons to meet this requirement. Consider that converting batch ground truth polygons consumes less memory than caching batch ground truth masks but may introduce a potential slowdown in the validation process.

If we opt to handle this conversion in `PackDetInputs`, it involves working with BitmapMasks and necessitates a sequence of conversions between CPU and GPU (CPU -> GPU -> CPU (`detach`) -> GPU). The dilemma lies in identifying the optimal approach, considering the trade-offs between memory usage and processing speed. Currently, I'm unsure which path is more favorable in this scenario.

eugene123tw:
@vinnamkim, please review my latest commit at https://github.com/openvinotoolkit/training_extensions/pull/2698/commits/30e5a179d399df2d131b1f8c4a9f9f5a1902e972.

In this commit, I introduced `self.use_polygon` parameter to `OTXInstanceSegDataset` class. This parameter allows developers to choose between using polygons or masks for training. Ideally, I propose adding this option to the `DataModuleConfig` to enable configuration changes through either the config or CLI. If you have alternative suggestions on where to place this parameter, please provide guidance.

I conducted profiling tests on both polygon and mask cases, running 10 epochs on the WGISD and Vitens Coliform datasets.

The first scenario employs polygons on WGISD (`self.use_polygon=True`) resulting in a more efficient training time of 153.91 seconds for 10 epochs.
polygon on wgisd

The second scenario uses direct masks on WGISD (self.use_polygon=False), leading to a longer training time of 181.13 seconds for 10 epochs.

mask on wgisd

However, the difference on WGISD seems negligible. On the other hand, for the Vitens dataset, which primarily consists of 4k x 4k images, the distinction is significant. The end-to-end training time using polygons is 159.1 seconds.

image

And using masks prolongs the training time to 402 seconds for 10 epochs.

mask on coliform

As anticipated, the use of direct masks introduces significant slowness due to device conversion and extensive collation.

I recommend retaining polygons as the default option and exposing the direct masks choice in the Hydra config. Please share your thoughts on this approach.

src/otx/core/data/dataset/instance_segmentation.py
eugene123tw: Please review my latest commit at https://github.com/openvinotoolkit/training_extensions/pull/2698/commits/30e5a179d399df2d131b1f8c4a9f9f5a1902e972

I introduced self.use_polygon parameter to OTXInstanceSegDataset class. This parameter allows developers to choose between using polygons or masks for training. Ideally, I propose adding this option to the DataModuleConfig to enable configuration changes through either the config or CLI. If you have alternative suggestions on where to place this parameter, please provide guidance.

src/otx/core/data/entity/instance_segmentation.py
vinnamkim: Sorry, I missed "not" here, "However, I agree that this can 'not' be handled in this PR and let it leave as the future work.". Absolutely, it should be left as future work.

src/otx/core/data/transform_libs/mmdet_inst_seg.py
vinnamkim: > Could you provide more details on how the registration of LoadAnnotations might impact the detection task? As per the current design, Instance Segmentation has introduced a distinct MMDetInstSegTransformLib separate from detection.MMDetTransformLib, indicating a deliberate divergence, and should work separately.

Let assume that a user uses OTX as Python API, not CLI. Than, the order of importing `src/otx/core/data/transform_libs/mmdet.py` and `src/otx/core/data/transform_libs/mmdet_inst_seg.py` can affect to their runtime behavior. This is hidden and hard to find.

> My concern is that unifying these processes might result in the nesting of LoadAnnotation and PackDetInputs with two different data entity checks (InstanceSegDataEntity and DetDataEntity). This distinction is crucial as detection tasks don't need masks, while instance segmentation tasks do. If we proceed with unification, PackDetInputs would then need to return two different data entities.

I think we put our data entity in `results["__otx__"]` at the very first pipeline (`LoadImageFromFile`). We can make branches by investigating the type of `results["__otx__"]`. This seems easy to modify such as

```python
class PackInputs():
     def transform(self, results)
          if isinstance(results["__otx__"], DetDataEntity):
               pack_inputs_det(...)
          elif isinstance(results["__otx__"], InstSegDataEntity):
               pack_inputs_inst_seg(...)
```

vinnamkim:
> @vinnamkim, please review my latest commit at [30e5a17](https://github.com/openvinotoolkit/training_extensions/commit/30e5a179d399df2d131b1f8c4a9f9f5a1902e972).
> ...
> I recommend retaining polygons as the default option and exposing the direct masks choice in the Hydra config. Please share your thoughts on this approach.

Is this degeneration coming from that the cost of handling the 2D masks than 1D Polygons during transforms? I mean that this is because we change it to 2D masks at the earliest step. Is it correct? 

My suggestion here https://github.com/openvinotoolkit/training_extensions/pull/2698#discussion_r1418898670 was doing transforms with 1D Polygons themselves (not 2D masks if possible) and change 1D Polygons to 2D Masks at the last step (`PackInputs`).

src/otx/core/model/module/instance_segmentation.py
vinnamkim: We might be able to adjust `pl.LightningModule._apply_batch_transfer_handler()` to get our own strategy for data transfer between devices for this problem.
image

Is it too cost to keep the masks in the CUDA memory? I guess we are using 1024x1024 resolution. It takes 1MB for each mask. Therefore, the memory usage for batch of masks would be `batch_size * average_number_of_masks_in_sample  * 1MB`.

vinnamkim (dismissed):
LGTM. We found that there are many points that can be researched further and might have a possibility to be improved in the future. Let them leaved as a future work.

src/otx/cli/train.py
```diff
@@ -42,6 +42,8 @@ def otx_train(overrides: list[str]) -> None:
     # apply extra utilities
     # (e.g. ask for tags if none are provided in cfg, print cfg tree, etc.)
     # utils.extras(cfg)
+    if core.global_hydra.GlobalHydra().is_initialized():
```
kprokofi: We can remove if -> since calling core.global_hydra.GlobalHydra().clear() is safe even when not GlobalHydra().is_initialized()

src/otx/core/data/dataset/instance_segmentation.py
```diff
@@ -0,0 +1,78 @@
...
+
+        gt_bboxes, gt_labels, gt_masks, gt_polygons = [], [], [], []
+
+        for annotation in item.annotations:
+            if isinstance(annotation, Polygon):
+                bbox = np.array(annotation.get_bbox(), dtype=np.float32)
+                gt_bboxes.append(bbox)
+                gt_labels.append(annotation.label)
+
+                if self.use_polygon:
```
kprokofi: Why do we need this condition if you have set self.use_polygon always true? Maybe we will use polygons every time if it is better?

src/otx/core/data/transform_libs/mmdet_inst_seg.py
```diff
@@ -0,0 +1,145 @@
...
+
+        Args:
+            otx_data_entity (OTXDataEntity): The data entity containing the masks or polygons.
+            height (int): The height of the masks.
+            width (int): The width of the masks.
+
+        Returns:
+            gt_masks (BitmapMasks or PolygonMasks): The generated ground truth masks.
+        """
+        if len(otx_data_entity.masks):
```
kprokofi: Same here. As I understand it always will be Polygons. Can we stick to Polygons and disregard BitmapMasks?

kprokofi (commented):
My main comment: can we stick to the one representation of the mask (Polygons) and delete everything related to BitMasks? Do you have concerns about that?

harimkang (dismissed):

src/otx/cli/train.py
harimkang: @eugene123tw @kprokofi 
Vinnam and I found a solution to this. - https://github.com/openvinotoolkit/training_extensions/commit/c5ac6a30b96c5400a1c9f88d3f4d2ca5e50ea2e5

If that PR(https://github.com/openvinotoolkit/training_extensions/pull/2707) is merged first, I think we can remove this code.

harimkang:
This commit (https://github.com/openvinotoolkit/training_extensions/pull/2707/commits/c5ac6a30b96c5400a1c9f88d3f4d2ca5e50ea2e5) was merged and I fixed the conflicts in this PR. - https://github.com/openvinotoolkit/training_extensions/pull/2698/commits/b9db4d2689a2e688d18b4207da97ce90e7daf408

harimkang (dismissed):

src/otx/cli/train.py
eugene123tw: Removed! Thanks!

src/otx/core/data/dataset/instance_segmentation.py
eugene123tw: I've implemented a data module configuration specifically for Instance Segmentation, accessible https://github.com/eugene123tw/training_extensions/blob/eugene/v2-inst-seg/src/otx/core/config/data.py#L45

This addition anticipates the possibility of integrating a future instance segmentation algorithm that handles masks more effectively than the current polygon method. Users may prefer training with direct masks instead of polygons in such instances.

Providing users with these two options offers flexibility and accommodating potential improvement in instance segmentation techniques.

src/otx/core/data/transform_libs/mmdet_inst_seg.py
eugene123tw: Kindly refer to my previous comment for context. Additionally, please review the following file for modification: https://github.com/eugene123tw/training_extensions/blob/eugene/v2-inst-seg/src/otx/core/data/transform_libs/mmdet.py. This file combines MMDetection instance segmentation transforms with the MMDetection framework due to this comment: https://github.com/openvinotoolkit/training_extensions/pull/2698/#discussion_r1419727385

src/otx/core/config/data.py
```diff
@@ -35,4 +35,11 @@ class DataModuleConfig:
...
     mem_cache_size: str = "1GB"
-    mem_cache_img_max_size: tuple[int, int] | None = None
+    mem_cache_img_max_size: Optional[Tuple[int, int]] = None
+
+
+@dataclass
+class InstSegDataModuleConfig(DataModuleConfig):
+    """DTO for instance segmentation data module configuration."""
+
+    include_polygons: bool = True
```
eugene123tw, line 45: @harimkang @vinnamkim @kprokofi

I introduced this modification to provide flexibility with two training options: utilizing masks or polygons. Please also review the changes in src/otx/core/config/__init__.py

src/otx/core/data/factory.py
```diff
@@ -85,6 +91,17 @@ def create(
                 mem_cache_img_max_size=cfg_data_module.mem_cache_img_max_size,
             )
 
+        if task == OTXTaskType.INSTANCE_SEGMENTATION:
+            from .dataset.instance_segmentation import OTXInstanceSegDataset
+            # NOTE: DataModuleConfig does not have include_polygons attribute
+            include_polygons = getattr(cfg_data_module, 'include_polygons', False)
```
eugene123tw, line 97: @harimkang @vinnamkim @kprokofi

In mypy checks, it checks if this attribute include_polygons exist.

src/otx/core/data/transform_libs/mmdet.py
```diff
@@ -8,21 +8,26 @@
 from typing import TYPE_CHECKING, Callable
```
eugene123tw, line 8: @harimkang @vinnamkim @kprokofi

Please review the specified file, as I've consolidated the instance segmentation MMDet pipeline with the detection's pipeline based on Vinnam's feedback.

src/otx/core/config/data.py
```diff
@@ -2,10 +2,10 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 """Config data type objects for data."""
-from __future__ import annotations
-
+# NOTE: omegaconf would fail to parse dataclass with `from __future__ import annotations` in Python 3.8, 3.9
+# ruff: noqa: FA100
```
eugene123tw, line 6: @harimkang @vinnamkim

Python 3.10 CI test runs fine with `from __future__ import annotations`. However, for Python 3.8 and 3.9, the usage of this import statement triggers an error, potentially related to PEP 604. To address this, I have temporarily modified the file to employ Python 3.8 type-checking conventions.

Please take a look at the provided error traceback, and if you have a more optimal solution or insights into the compatibility issue, I'd be greatly appreciated.
```
  File "/home/yuchunli/git/otx-2x/venv_39/lib/python3.8/site-packages/hydra/core/config_store.py", line 85, in store
    cfg = OmegaConf.structured(node)
  File "/home/yuchunli/git/otx-2x/src/otx/core/config/__init__.py", line 118, in register_configs
    cs.store(group="data", name="base_inst_seg", node=InstSegDataModuleConfig)
  ...
  omegaconf.errors.ConfigTypeError: 'type' object is not subscriptable
    full_key: 
    object_type=None
```

harimkang (approved):

vinnamkim (approved):
LGTM. However, please keep in your mind the comments here about data pipeline optimizations and improve them in the future.

sungmanc (approved):

jaegukhyun (approved):

kprokofi (approved):

PR #2701, opened 2023-12-06
Fix integration test config settings

harimkang:
### Summary

- Add debug/intg_test cfg for integration testing
- Modify epochs to 2 & accelerator to gpu
    - 2 epochs training: to use 2 epochs and also examine what happens between each epoch (@kprokofi)
    - trainer.accelerator=gpu: device will be used should be explicit in the tests. (@vinnamkim)
    - Currently only uses GPU, but consider adding CPU testing in the future
- Replace subprocess runner with os.system

### How to test

`python -m pytest tests/integration/cli/test_cli.py`

### Checklist

### License

```python
```


tests/integration/cli/test_cli.py
```diff
@@ -46,18 +45,18 @@ def test_otx_train(recipe: str, tmp_path: Path) -> None:
...
-        "trainer=auto",
         f"base.work_dir={tmp_path}",
         f"base.output_dir={tmp_path / 'outputs'}",
+        "+debug=intg_test",
         *DATASET[task]['overrides'],
     ]
-    rc, stdout, _ = check_run(command_line)
+    command = " ".join(command_cfg)
+
+    rc = os.system(command=command)
```
vinnamkim: Can this execute the CLI process in the main process without spawning another subprocess? I mean, if I execute `pdb -m pytest tests/integration/cli/test_cli.py`, is it possible to break at the model training code? My suggestion was that developers should be easy to debug when there is problem on the intergration tests.

sungmanc (dismissed):

tests/integration/cli/test_cli.py
harimkang: https://github.com/openvinotoolkit/training_extensions/pull/2701/commits/a39b283ddfcb36314a320572d1b692166cc2eaa9
-> I followed your advice and modified it to call the `main` function directly, but had a problem in this case: I noticed that  if the module is registered in the mmengine.registry in the first test, it affects the next test. (test case with first detection recipe is passed, but others are failed) So for now, I couldn't think of any other good way to solve this, so I changed the registry registration to force in `src/otx/core/model/entity/detection.py`, 

I'm sure there's a way to prevent this in tests only without touching existing code, but I can't think of it at the moment - if you know of one, please let me know.

vinnamkim: I'll take a look after this PR.

vinnamkim (dismissed):
LGTM.

harimkang:
I rechecked the integration test as the segmentation PR was merged. had a similar issue with mmdetection and fixed it in mmseg side.

vinnamkim (approved):

sungmanc (approved):

jaegukhyun (approved):

PR #2703, opened 2023-12-06
Organize and Enable callbacks per task

harimkang:
### Summary

- Re-enable callbacks that were not applied
    - Model Checkpoint Callback & Early-Stopping
- Organize callbacks by task

### Checklist

### License

```python
```


src/otx/config/callbacks/default.yaml
```diff
@@ -1,22 +1,47 @@
...
 
+# https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.callbacks.RichModelSummary.html
 model_summary:
-  max_depth: -1
+  _target_: lightning.pytorch.callbacks.RichModelSummary
+  max_depth: 1 # the maximum depth of layer nesting that the summary will include
+
+# https://lightning.ai/docs/pytorch/latest/api/lightning.pytorch.callbacks.RichProgressBar.html
+rich_progress_bar:
+  _target_: lightning.pytorch.callbacks.RichProgressBar
```
vinnamkim, line 47: @jaegukhyun 
It seems that `RichProgressBar` can automatically add new lines if the terminal is too narrow.
image
Please revert your change for detection metric logging after this PR.

vinnamkim (dismissed):
LGTM.

src/otx/config/callbacks/default.yaml
jaegukhyun: Sure, Thanks

src/otx/config/callbacks/default.yaml
```diff
@@ -1,22 +1,47 @@
...
-  - rich_progress_bar
-  - _self_
+# https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.callbacks.EarlyStopping.html
+early_stopping:
+  _target_: lightning.pytorch.callbacks.EarlyStopping
+  monitor: ??? # quantity to be monitored, must be specified !!!
+  min_delta: 0. # minimum change in the monitored quantity to qualify as an improvement
+  patience: 3 # number of checks with no improvement after which training will be stopped
+  verbose: False # verbosity mode
+  mode: "min" # "max" means higher metric value is better, can be also "min"
```
sungmanc, line 8: How about changing this default value to "max" ? I think most of our case, we'll use the "max" Acc, IoU, mAP, etc.

sungmanc (dismissed):
LGTM, minor comment

src/otx/config/callbacks/default.yaml
harimkang: As I understand it, this is the default value for lightning.callbacks.EarlyStopping. Changing the value would require discussion. (https://github.com/Lightning-AI/pytorch-lightning/blob/master/src/lightning/pytorch/callbacks/early_stopping.py)

harimkang:
As merged the segmentation PR, segmentation also added callbacks. https://github.com/openvinotoolkit/training_extensions/pull/2703/commits/e549e2643da0e94a43e8138b0d891f5cdcd49148

src/otx/config/callbacks/default.yaml
sungmanc: Yeah, I also know that this is the lightning's default value. However, we need to change the `mode` for every new recipe or other configs since "min" `mode` is useless for our cases.

sungmanc (approved):

jaegukhyun (approved):

src/otx/config/callbacks/default.yaml
harimkang: We'll discuss this further, but for now, default alone doesn't currently work without overriding.

vinnamkim (approved):

PR #2708, opened 2023-12-07
Bring the SOTA(DINOv2) model to the classification task

sungmanc:
### Summary

This PR includes:
* DINO-v2-register-small model implementation and experimental results.
* Wrap the common thing of configs/models/*.yaml
* Used the MMpretrain's data pipeline, Torchvision transforms will be introduced at the further PR.
* Add tests for the DINO-v2 model

Experimental results: (iter_time_epoch) will be noted at OneNote.
* To experiment with the DINO-v2-register-small model, **I freeze the backbone**. If not, the model couldn't training well.
image

### Checklist

### License

```python
```


src/otx/recipe/classification/otx_dino_v2.yaml
```diff
@@ -0,0 +1,68 @@
...
+          - 58.395
+          - 57.12
+          - 57.375
+        to_rgb: True
+  optimizer:
+    _target_: torch.optim.SGD
+    _partial_: true
+    lr: 0.007
+    momentum: 0.9
+    weight_decay: 0.0001
```
vinnamkim: To finetune Transformers, we normally use 1) AdamW optimizer and 2) small lr ~ 1e-5 or 1e-4. It comes from my experiences.

Please see the following articles for more details:
1. https://www.montevideolabs.com/fine-tuning-a-vision-transformer/
2. https://colab.research.google.com/github/NielsRogge/Transformers-Tutorials/blob/master/VisionTransformer/Fine_tuning_the_Vision_Transformer_on_CIFAR_10_with_the_%F0%9F%A4%97_Trainer.ipynb#scrollTo=8wTw4RXe73vA
3. https://arxiv.org/pdf/2106.10270.pdf

tests/unit/algo/classification/test_otx_dino_v2.py
```diff
@@ -0,0 +1,45 @@
...
+from otx.algo.classification import DINOv2RegisterClassifier
+from otx.core.data.entity.base import OTXBatchLossEntity
+from otx.core.data.entity.classification import MulticlassClsBatchDataEntity, MulticlassClsBatchPredEntity
+from torch import nn
+
+if TYPE_CHECKING:
+    from omegaconf import DictConfig
+
+class TestDINOv2RegisterClassifier:
+    def test__init(self, fxt_config_mock: DictConfig) -> None:
```
vinnamkim: What's the rationale behind this double underscore?

tests/unit/algo/classification/conftest.py
```diff
@@ -0,0 +1,51 @@
...
+    return MulticlassClsBatchDataEntity(
+        batch_size=2,
+        images=tv_tensor,
+        imgs_info=img_infos,
+        labels=[torch.tensor([0]), torch.tensor([1])],
+    )
+
+@pytest.fixture()
+def fxt_config_mock() -> DictConfig:
+    config_mock = MagicMock(spec=DictConfig)
```
vinnamkim: Can it be just `DictConfig`? Is there any reason that it should be `MagicMock`?

src/otx/algo/classification/backbones/otx_efficientnet_v2.py
```diff
@@ -1,6 +1,6 @@
 # Copyright (C) 2023 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
-#
```
vinnamkim, line 3: @harimkang, It seems that we don't have code formatter in our pre-commit rules. I have been able to see this kind of code clutters from other guys' PR (including me). If this is true, please introduce it.

src/otx/algo/classification/otx_dino_v2.py
```diff
@@ -0,0 +1,138 @@
...
+            losses = OTXBatchLossEntity()
+            losses["loss"] = outputs
+            return losses
+
+        batch_size = outputs["pred_labels"].shape[0]
+        scores = []
+        labels = []
+        for b in range(batch_size):
+            scores.append(outputs["pred_scores"][b])
+            labels.append(outputs["pred_labels"][b])
```
vinnamkim: Please use https://pytorch.org/docs/stable/generated/torch.unbind.html

src/otx/algo/classification/otx_dino_v2.py
```diff
@@ -0,0 +1,138 @@
...
+            }
+        return self._customize_outputs(outputs, inputs)
+
+    def _customize_inputs(self, entity: MulticlassClsBatchDataEntity) -> dict[str, Any]:
+        """Customize the inputs for the model."""
+        inputs: dict[str, Any] = {}
+        inputs["x"] = self._preprocess_img(
+            torch.stack([torch.as_tensor(image) for image in entity.images]),
+        )
+        inputs["labels"] = torch.cat([torch.as_tensor(label) for label in entity.labels])
```
vinnamkim: Guess that we don't need `torch.as_tensor()`.

src/otx/algo/classification/backbones/otx_efficientnet_v2.py
harimkang: Yes, I get it, pre-commit doesn't seem to catch this stuff (file docstring).
I'll see if there's a tool that fixes it all at once, and if so, I'll do it on another PR.

src/otx/algo/classification/otx_dino_v2.py
```diff
@@ -0,0 +1,138 @@
...
+            v.requires_grad = False
+
+    def _preprocess_img(self, imgs: torch.Tensor) -> torch.Tensor:
+        """Control normalize and BGR/RGB conversion."""
+        # BGR -> RGB
+        if self.data_preprocess_cfg.to_rgb and imgs.size(1) == 3:
+            imgs = imgs.flip(1)
+
+        # Normalization
+        imgs = imgs.float()
```
vinnamkim: Please do this at initialization.

src/otx/algo/classification/otx_dino_v2.py
```diff
@@ -0,0 +1,138 @@
...
+        """
+        customized_inputs = self._customize_inputs(inputs)
+        feats = self.model(customized_inputs["x"])
+        if self.training:
+            outputs = self.loss(feats, customized_inputs["labels"])
+        else:
+            pred_scores = nn.functional.softmax(feats, dim=1)
+            max_pred_elements, max_pred_idxs = torch.max(pred_scores, dim=1)
+            pred_scores = max_pred_elements.cpu().detach().numpy()
+            pred_labels = max_pred_idxs.cpu().detach().numpy()
```
vinnamkim: Do not call blocking API as possible in the main loop such as `.cpu()`. See https://lightning.ai/docs/pytorch/stable/advanced/speed.html#item-numpy-cpu

tests/unit/algo/classification/conftest.py
sungmanc: No special reason in here, is there advantage to use the `DictConfig`?

src/otx/algo/classification/otx_dino_v2.py
```diff
@@ -0,0 +1,138 @@
...
+            self.config.head.in_channels,
+            self.config.head.num_classes,
+        )
+
+        return nn.Sequential(
+            OrderedDict([
+                ("backbone", self.backbone),
+                ("head", self.head),
+            ]),
+        )
```
vinnamkim:
```suggestion
class DinoV2:
    def __init__(self, name, freeze_backbone, in_channels, num_classes):
        self.backbone = torch.hub.load(
            repo_or_dir="facebookresearch/dinov2",
            model=name,
        )
        self.loss = nn.CrossEntropyLoss()
        self.softmax = nn.Softmax()
        if freeze_backbone:
            self._freeze_backbone(self.backbone)
    
        self.head = nn.Linear(
            in_channels,
            num_classes,
        )
    def forward(self, imgs, labels=None):
        feats = self.backbone(imgs)
        logits = self.head(feats)
        if self.training:
             return self.loss(logits, labels)
        return self.softmax(logits) 


....


class DINOv2RegisterClassifier:
...        
            
    def _create_model(self) -> nn.Module:
        """Create the model."""
        return DinoV2(...)
```

src/otx/algo/classification/otx_dino_v2.py
```diff
@@ -0,0 +1,138 @@
...
+        """Control normalize and BGR/RGB conversion."""
+        # BGR -> RGB
+        if self.data_preprocess_cfg.to_rgb and imgs.size(1) == 3:
+            imgs = imgs.flip(1)
+
+        # Normalization
+        imgs = imgs.float()
+        return (imgs - self.mean) / self.std
+
+    def forward(
```
vinnamkim: Please remove this custom forward function according to my above comment to create separate `nn.Module`.

tests/unit/algo/classification/conftest.py
vinnamkim: It looks better to put in the genuine input in as possible. We need mocking if it is impossible or too costly.

sungmanc: Do we need genuine input for the unit test? We could check with the genuine data at integration tests.

src/otx/algo/classification/otx_dino_v2.py
sungmanc: Do you know the alternative? If so, please let me know. I was also trying to remove this statement, however, I failed to find it. BTW, I'll looking for this

src/otx/algo/classification/otx_dino_v2.py
sungmanc: Could you explain more? You want to avoid using this `_preprocess_img` function at the loop?

tests/unit/algo/classification/conftest.py
vinnamkim: I mean there is no reason not to use actual `DictConfig` because there is no additional cost compared to creating `MagicMock` with complete parameters.

src/otx/algo/classification/otx_dino_v2.py
sungmanc: Done. https://github.com/openvinotoolkit/training_extensions/pull/2708/commits/7c8db3a54e84a4e183a48378fdd428e047288441

src/otx/algo/classification/otx_dino_v2.py
sungmanc: Done. https://github.com/openvinotoolkit/training_extensions/pull/2708/commits/7c8db3a54e84a4e183a48378fdd428e047288441

src/otx/algo/classification/otx_dino_v2.py
sungmanc: Done. https://github.com/openvinotoolkit/training_extensions/pull/2708/commits/7c8db3a54e84a4e183a48378fdd428e047288441

src/otx/algo/classification/otx_dino_v2.py
sungmanc: Done. https://github.com/openvinotoolkit/training_extensions/pull/2708/commits/7c8db3a54e84a4e183a48378fdd428e047288441

src/otx/algo/classification/otx_dino_v2.py
sungmanc: Done. https://github.com/openvinotoolkit/training_extensions/pull/2708/commits/7c8db3a54e84a4e183a48378fdd428e047288441

src/otx/algo/classification/otx_dino_v2.py
sungmanc: Done. https://github.com/openvinotoolkit/training_extensions/pull/2708/commits/7c8db3a54e84a4e183a48378fdd428e047288441

tests/unit/algo/classification/conftest.py
sungmanc: Done. https://github.com/openvinotoolkit/training_extensions/pull/2708/commits/7c8db3a54e84a4e183a48378fdd428e047288441

tests/unit/algo/classification/test_otx_dino_v2.py
sungmanc: Done. https://github.com/openvinotoolkit/training_extensions/pull/2708/commits/7c8db3a54e84a4e183a48378fdd428e047288441

src/otx/recipe/classification/otx_dino_v2.yaml
sungmanc: Done. https://github.com/openvinotoolkit/training_extensions/pull/2708/commits/7c8db3a54e84a4e183a48378fdd428e047288441

src/otx/algo/classification/otx_dino_v2.py
```diff
@@ -0,0 +1,132 @@
...
+            if not isinstance(outputs, torch.Tensor):
+                raise TypeError(outputs)
+
+            losses = OTXBatchLossEntity()
+            losses["loss"] = outputs
+            return losses
+
+        max_pred_elements, max_pred_idxs = torch.max(outputs, dim=1)
+        pred_scores = max_pred_elements.detach()
+        pred_labels = max_pred_idxs.detach()
```
vinnamkim: What's the reason that `detach` is remained still?

src/otx/recipe/classification/otx_dino_v2.yaml
vinnamkim: `AdamW` and `Adam` are different each other. Is https://github.com/openvinotoolkit/training_extensions/commit/7c8db3a54e84a4e183a48378fdd428e047288441 using `Adam`?

src/otx/algo/classification/otx_dino_v2.py
sungmanc: I think we don't need to sustain the tensors at the CUDA memory anymore. That's the reason why I remained `detach`. Do you think this is also blocker ?

src/otx/recipe/classification/otx_dino_v2.yaml
sungmanc: Ah, sorry. I mis-thinked that the default optimizer is the `AdamW`, thkanks for noticing

src/otx/algo/classification/otx_dino_v2.py
vinnamkim: `detach()` does not change the device memory location of `Tensor`. It just clones a new `Tensor` which is detached from the current computation graph for `backward()`. Please see https://pytorch.org/docs/stable/generated/torch.Tensor.detach.html#torch.Tensor.detach. Cloning `Tensor` incurs unnecessary costs. We must avoid unnecessary costs if there is no special reason.

sungmanc: Thanks for the comment, https://github.com/openvinotoolkit/training_extensions/pull/2708/commits/686c15d2c9d2d025f01b02286cde6963a9aca7d9

src/otx/recipe/classification/otx_dino_v2.yaml
sungmanc: Done. https://github.com/openvinotoolkit/training_extensions/pull/2708/commits/686c15d2c9d2d025f01b02286cde6963a9aca7d9

vinnamkim (dismissed):
LGTM.

src/otx/config/model/default.yaml
```diff
@@ -0,0 +1,12 @@
+optimizer:
+  _target_: torch.optim.Adam
```
harimkang, line 2: Why is Adam the default optimizer on a model default?

harimkang (dismissed):
LGTM

tests/integration/cli/test_cli.py
```diff
@@ -21,7 +21,6 @@
         "data_dir": "tests/assets/classification_dataset",
         "overrides": [
             "model.otx_model.config.head.num_classes=2",
-            "model.otx_model.config.data_preprocessor.num_classes=2",
```
harimkang, line 24: I think it would be better to add `data_preprocessor.num_classes` to the dinov2 recipe yaml file instead of this way, what do you think?

sungmanc: Actually, DINOv2 doesn't use the data_preprocessor.num_classes..

src/otx/config/model/default.yaml
sungmanc: Hmm,, I'm not sure about the default. may be it comes from the pytorchlightning?

tests/integration/cli/test_cli.py
sungmanc: Since we discussed before, num_classes for the data_preprocessor will not affects to the other models. So, I think removing this line is better. We couldn't use the `data_preprocessor.num_classes` field if we want to use non mmbased model.

harimkang: However, since this override is used by other models as well, I don't think it makes sense to remove it for DINOv2 only. Even if it doesn't use in DINOv2 actually, it would be nice to keep the `data_preprocessor.num_classes` in to config format it with other classification configs.

harimkang: Or shouldn't we be clearing the unused data_preprocessor config parts from the other models as well? This seems to be confusing.

src/otx/config/model/default.yaml
harimkang: I asked because I know most of our models use SGD, but we can discuss that later.

tests/integration/cli/test_cli.py
sungmanc: Done. https://github.com/openvinotoolkit/training_extensions/pull/2708/commits/94282b0015747affa828f916615f4f9d0bd36de3

jaegukhyun (dismissed):

tests/unit/algo/classification/test_otx_dino_v2.py
```diff
@@ -0,0 +1,32 @@
...
+
+from __future__ import annotations
+
+import torch
+from otx.algo.classification import DINOv2
+
+
+class TestDINOv2:
+    def setup(self) -> None:
+        self.model = DINOv2(
```
vinnamkim: From what we observed in https://github.com/openvinotoolkit/training_extensions/actions/runs/7137703253/job/19438115260?pr=2708, we cannot allow the model download from the external URL in the **"unit test"**. Please mock `torch.hub.load()`.

src/otx/algo/classification/otx_dino_v2.py
```diff
@@ -0,0 +1,132 @@
...
+
+        self.head = nn.Linear(
+            head_in_channels,
+            num_classes,
+        )
+
+        self.loss = nn.CrossEntropyLoss()
+        self.softmax = nn.Softmax()
+
+        self.training = training
```
vinnamkim: Please remove it. This variable is already in the super class, `nn.Module`.

tests/unit/algo/classification/test_otx_dino_v2.py
vinnamkim:
```suggestion
    @pytest.fixture
    def model(self) -> None:
        return DINOv2(
```

I have not been noticed but I guess there were so many anti-patterns which are far away from the pytest common  practice, not only this PR, but also https://github.com/openvinotoolkit/training_extensions/pull/2711.

Please do not use member variable, use pytest fixture.

sungmanc: Thanks, https://github.com/openvinotoolkit/training_extensions/pull/2708/commits/00153bb018437a0ca6120ac610f8aaf3501d9bd9

src/otx/algo/classification/otx_dino_v2.py
sungmanc: Done. https://github.com/openvinotoolkit/training_extensions/pull/2708/commits/00153bb018437a0ca6120ac610f8aaf3501d9bd9

tests/unit/algo/classification/test_otx_dino_v2.py
sungmanc: Done. https://github.com/openvinotoolkit/training_extensions/pull/2708/commits/2abae5c112ecbcc3ce0a012136c33e043e877e1f

vinnamkim (approved):

jaegukhyun (approved):

harimkang (approved):

PR #2712, opened 2023-12-08
Add custom model example

vinnamkim:
### Summary

- This should be merged after https://github.com/openvinotoolkit/training_extensions/pull/2709

### Checklist

### License

```python
```


sungmanc (approved):

src/otx/core/engine/train.py
```diff
@@ -32,20 +32,27 @@
...
+    from lightning import Callback, Trainer
     from lightning.pytorch.loggers import Logger
 
+    from otx.core.model.module.base import OTXLitModule
 
-def train(cfg: TrainConfig) -> tuple[Trainer, dict[str, Any]]:
+
+def train(
+    cfg: TrainConfig,
+    otx_model: OTXModel | None = None,
```
harimkang, line 46: Actually, I understand your intention to make it so that only custom models receive it separately.
However, I'm not sure if adding this is intended to mean that other things will eventually get it as well.
For example, if user write an `OTXDataModule` outside of the function, shouldn't it also receive it later?

I'm not saying this is wrong, but i just wonder,
- Is it better to add arguments one by one? Before we add this one by one, shouldn't we define what we should get?
- Are we going to go in the direction of providing both config (build from config) & argument object patching case?
- Or is this a workaround, and in the future we will only receive TrainConfig?

Looking at this, I'm confused about which direction the entry point should go.

vinnamkim: See my Teams chat message. It is not concrete one. It is a temporary one for the Jupyter notebook example. We will develop further.

harimkang: I've talked to Mark and suggested that this is something we should discuss more, and I'll be working on a PoC to improve this on v2 branch. If I come back with a proposal, I'd love to discuss it with you and other group members.

wonjuleee (approved):
LGTM!

harimkang:
@eugene123tw @kprokofi @sungmanc @jaegukhyun 
Hi all,
Through discussion with Mark,
I decided to do a PoC on the current code regarding the entry point design.
this PR changes the train function, but I think it could be confusing, so I'd rather not have to merge it, but instead fix it once we decide on a future direction.
What do you guys think? I'll go with the majority of the group. Let me know what you think.

sungmanc:
> @eugene123tw @kprokofi @sungmanc @jaegukhyun Hi all, Through discussion with Mark, I decided to do a PoC on the current code regarding the entry point design. this PR changes the train function, but I think it could be confusing, so I'd rather not have to merge it, but instead fix it once we decide on a future direction. What do you guys think? I'll go with the majority of the group. Let me know what you think.

Is there any reason to merge this PR urgently? Why do we need a vote although there was a decision from Mark and you?

jaegukhyun:
> @eugene123tw @kprokofi @sungmanc @jaegukhyun Hi all, Through discussion with Mark, I decided to do a PoC on the current code regarding the entry point design. this PR changes the train function, but I think it could be confusing, so I'd rather not have to merge it, but instead fix it once we decide on a future direction. What do you guys think? I'll go with the majority of the group. Let me know what you think.

I agree on Sungman. Unless we have urgent requests for documentation about adding custom modules. we don't have to hurry to prepare documentation.

vinnamkim:
> @eugene123tw @kprokofi @sungmanc @jaegukhyun Hi all, Through discussion with Mark, I decided to do a PoC on the current code regarding the entry point design. this PR changes the train function, but I think it could be confusing, so I'd rather not have to merge it, but instead fix it once we decide on a future direction. What do you guys think? I'll go with the majority of the group. Let me know what you think.

I don't think that you raised a reasonable concern. This is to show an example of how to add custom model in OTX 2.0. To achieve this objection, it is inevitable to change the `train()` API slightly to explain the e2e flow of adding custom model. Do you really think that this example should just end up before showing the actual model training? You raised the potential confusions on the codebase and it can be harmful for the current developer's work. However, I don't think so. This is because:

1. This change has no effect on the previous `train()` API behavior.
2. This change is a super simple change just to add argument to override the OTX model object:
https://github.com/openvinotoolkit/training_extensions/blob/1be65ca8c1009817db910fa4f9fa1d3ddc335f6c/src/otx/core/engine/train.py#L50-L52
- I don't think that this description and the code are too confusion not to be understandable by the person of average intelligence.

On the other hand, there are many reasons that this example should be merged.

1. I'd like to ask our stakeholders for the current shape of custom model adding processes. We have a mission to improve our design if it is not fully adequate for our mission. And, this should be done continuously.
2. There is a new developer who should be participated at this moment, @sungchul2. He should refer to this material to integrate visual prompting task into OTX 2.0.

Now, I expect you to give me valid and rational reasons that you have in your mind.

harimkang:
> Now, I expect you to give me valid and rational reasons that you have in your mind.

I understand your intentions through your code, and I understand why we need this example, but there are a few things I'd like to point out.

- This is a change to the `train` function, and it would have been less confusing if you had told the group beforehand. I thought this was a pretty important decision for the future direction.
- I understand that Sungchul works on things after the entry point is in place (not for now), so I don't think this PR is urgent. But if it's a rescheduling, I get it.

Anyway, I fully understand your intentions and the reason for this PR, and now my position is that it doesn't matter if we merge the PR.

vinnamkim:
> * This is a change to the `train` function, and it would have been less confusing if you had told the group beforehand. I thought this was a pretty important decision for the future direction.

I think the main problem is that you guys did not read anything and think carefully about my PR. 
I already specified two times (and one with bold style at the end of example) to try to remove any of misunderstanding here.
image
image
It is your duty as a reviewer to read the work carefully.

If you guys read this correctly and have average of intelligence, I don't think there will be any confusion. This is proved that you changed your position now that this PR has no major concern. If you have concern, you can request enough discussions. I requested you guys reviewing on this and put my messages in the example as you seen. Not reading this carefully and not trying to request discussion on it with me are entirely your problem.

Without thinking carefully, you guys made such a drama from the last week. You guys are doing like a gang without any valid reason and discussion. I think you are using violence by taking advantage of the majority vote taking advantages you have. I believe that this is not healthy team culture and you made a deep mistake on it. I want you to make a public apology for this.

harimkang:
> I want you to make a public apology for this.

I don't think you understood me, I was questioning whether this PR needs to be merged in its current state where The training from Python API design is undefined, so I don't think I'm not fulfilling my reviewer duties by commenting. I didn't say the code doesn't or won't work, I just expressed concern and said there's no good reason for it to be merged. I don't think the fact that you added that comment to the article establishes a reason for this PR to be merged.

vinnamkim:
> I don't think you understood me, I was questioning whether this PR needs to be merged in its current state where The training from Python API design is undefined, so I don't think I'm not fulfilling my reviewer duties by commenting. I didn't say the code doesn't or won't work, I just expressed concern and said there's no good reason for it to be merged. I don't think the fact that you added that comment to the article establishes a reason for this PR to be merged.

Until now, you haven't been able to raise any valid reason. You have kept saying just again and again that it is not looking good and confusing just in your eyes. You haven't been able to get any valid reason why it is confusing compared to my grounds. Let me clear this again. What do you expect the additional cost to improve `train()` if this PR is merged? Is there any difference before this PR?

I see that you think that you had no mistake and you had no intention of apologizing.

harimkang:
> I see that you think that you had no mistake and you had no intention of apologizing.

I think the design direction of using Config only and using Config + object are very different directions, but as you say, this may be temporary, and as a flow behavior, it's not a problem. If it was a mistake to comment on this directly in the PR, I apologize for that.

MarkByun:
As Vinnam mentioned, this PR is one sample for helping new user of OTX v.2.0 like me and I requested this manual so this is not a any final design related document or code so we can merge this as a sample code even though this contains slightly different parameter usage.

MarkByun (approved):
This PR has its own value and it will be good for understanding how to use current OTX v.2.0.

PR #2716, opened 2023-12-11
Change the cls head to make them same with the otx1.5

sungmanc:
### Summary

To ensure backward compatibility, we need to use the same head architecture.
So, I changed the head architecture to make the same head with OTX 1.5.

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/recipe/classification/efficientnet_b0_light.yaml
+++ b/src/otx/recipe/classification/efficientnet_b0_light.yaml
@@ -12,9 +12,6 @@ data:
       - backend: cv2
         scale: 224
         type: RandomResizedCrop
-      - direction: horizontal
-        prob: 0.5
-        type: RandomFlip
       - type: PackInputs
   val_subset:
     batch_size: 64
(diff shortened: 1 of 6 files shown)


vinnamkim (commented):
Where were the previous configs coming from?

jaegukhyun (dismissed):

sungmanc:
> Where were the previous configs coming from?

It comes from the default ImageClassifier of MobileNet-v3 from MMPretrain

vinnamkim:
> > Where were the previous configs coming from?
> 
> It comes from the default ImageClassifier of MobileNet-v3 from MMPretrain

Did you compare the model accuracy results between the previous configs and this PRs? It is important to keep in mind that if some of legacy are found to be not good, we should throw it away rather than maintaining it blindly.

sungmanc:
> Did you compare the model accuracy results between the previous configs and this PRs? It is important to keep in mind that if some of legacy are found to be not good, we should throw it away rather than maintaining it blindly.

Now I'm checking the results. In my opinion, we should keep the head architecture with OTX 1.5. 
If there was a user who used the Geti classification with OTX1.5, we should keep the accuracy if the user used the same dataset.
However, if the head architecture is different, we can't guarantee the previous model's accuracy due to the difference in the head.

That's the reason why I reverted the head architecture to the OTX1.5. 

I'm not sure we should change the head architecture although some backward compatibilities could be broken.

sungmanc:
I discussed with the Mark, and we're going to support the backward compatible model (***_light.yaml) and otx model (otx_****).
Since I don't want to use the `Geti` word into the recipe, I used the `light` word. cc. @vinnamkim

eugene123tw (approved):

vinnamkim (approved):
LGTM.

harimkang (approved):

PR #2725, opened 2023-12-13
Fix DINOv2 data transform, add linear probe recipe

sungmanc:
### Summary

image

As you can see in the figure above, DINOv2 with linear probe showed faster & better performance in our 3 datasets including small, medium, large datasets. So, I added the recipe for the linear probing of the DINO v2 Classification.

Since linear probing is not mandatory recipe to us, I didn't added the linear probing recipe to the Regression tests

### Checklist

### License

```python
```


jaegukhyun:
If dinov2 works better on linear probe, how about move src/otx/recipe/classification/otx_dino_v2_linear_probe.yaml to src/otx/recipe/classification/otx_dino_v2.yaml. In my opinion, a recipe would be the best config for each model.

sungmanc:
> If dinov2 works better on linear probe, how about move src/otx/recipe/classification/otx_dino_v2_linear_probe.yaml to src/otx/recipe/classification/otx_dino_v2.yaml. In my opinion, a recipe would be the best config for each model.

I also agree with your opinion. Want to hear other guys opinions also. Please share your thoughts.
@vinnamkim @harimkang

vinnamkim:
> > If dinov2 works better on linear probe, how about move src/otx/recipe/classification/otx_dino_v2_linear_probe.yaml to src/otx/recipe/classification/otx_dino_v2.yaml. In my opinion, a recipe would be the best config for each model.
> 
> I also agree with your opinion. Want to hear other guys opinions also. Please share your thoughts. @vinnamkim @harimkang

From my experience, in almost all cases, the finetuning cannot be worse than the linear probing. If the target dataset is too small and has significant domain similarity with the pretrained dataset, the linear probing may have better generalization than the finetuning.

Keeping this in mind, I don't like `multiclass-CUB-*` benchmarks. I personally think that we should not use only these datasets to decide our decision. However, for now, we have to use them because we should compare OTX 2.0 with OTX 1.0.

Therefore, I recommend to include both of them. However, it seems that we should invent a new clever way to concise our configurations to cover those model variants.

sungmanc:
> it seems that we should invent a new clever way to concise our configurations to cover those model variants.

I prefer this way, So, how about considering the better way to organize our recipes in the future ? @jaegukhyun

vinnamkim (approved):

jaegukhyun (approved):
Let's organize recipes later.

harimkang (approved):

PR #2729, opened 2023-12-14
Make mean teacher algorithm consider distributed training

### LINKED ISSUE #2705: Multi-GPU training integration test failure
The failure occurs in daily tests for at least a week in a row :
https://github.com/openvinotoolkit/training_extensions/actions/runs/7105336612
FAILED tests/e2e/cli/detection/test_detection.py::TestToolsOTXSemiSLDetection::test_otx_multi_gpu_train_semisl[Object_Detection_ResNeXt101_ATSS]

@eunwoosh could you have a look?

eunwoosh:
### Summary
This PR fixes #2705 
The reason of bug is that when distributed training, one process may train a model w/ pseudo label but other processes doesn't  in some case. In this case, each process executes different code flow, which makes the error.
This PR changes code as below

-  If any process will train a model w/ pseudo label, then all process train a model w/ pseudo label.
- Not to affect backprop, process, which doesn't have to train w/ pseudo label, get 0 loss from pseudo label.

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algorithms/detection/adapters/mmdet/models/detectors/mean_teacher.py
+++ b/src/otx/algorithms/detection/adapters/mmdet/models/detectors/mean_teacher.py
@@ -10,11 +10,11 @@ import cv2
 import mmcv
 import numpy as np
 import torch
-from torch import distributed as dist
 from mmdet.core import bbox2result, bbox2roi
 from mmdet.core.mask.structures import BitmapMasks
 from mmdet.models import DETECTORS, build_detector
 from mmdet.models.detectors import BaseDetector
+from torch import distributed as dist
 
 from otx.utils.logger import get_logger
 
--- a/src/otx/algorithms/detection/adapters/mmdet/models/detectors/mean_teacher.py
+++ b/src/otx/algorithms/detection/adapters/mmdet/models/detectors/mean_teacher.py
@@ -193,19 +193,19 @@ class MeanTeacher(SAMDetectorMixin, BaseDetector):
                 get_unlabeled_loss = True
                 non_empty[0] = True
 
-        if self.filter_empty_annotations:
-            pseudo_bboxes = [pb for i, pb in enumerate(pseudo_bboxes) if non_empty[i]]
-            pseudo_labels = [pl for i, pl in enumerate(pseudo_labels) if non_empty[i]]
-            pseudo_masks = [pm for i, pm in enumerate(pseudo_masks) if non_empty[i]]
-            ul_img_metas = [im for i, im in enumerate(ul_img_metas) if non_empty[i]]
-            ul_img = ul_img[non_empty]
-        if self.visualize:
-            self._visual_online(ul_img, pseudo_bboxes, pseudo_labels)
         losses.update(ps_ratio=torch.tensor([pseudo_ratio], device=current_device))
 
         # Unsupervised loss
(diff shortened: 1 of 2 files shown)


eunwoosh:
Hi @kprokofi. I changed mean teacher algorithm code to make it compatible with distributed training.
It seems that those code is written by you, so could you check this change is ok?
I checked that the error doesn't occur w/ this change but I'm not sure it doesn't affect to training.

src/otx/algorithms/detection/adapters/mmdet/models/detectors/mean_teacher.py
```diff
@@ -182,22 +183,29 @@ def forward_train(
         pseudo_bboxes, pseudo_labels, pseudo_masks, pseudo_ratio = self.generate_pseudo_labels(
             teacher_outputs, device=current_device, img_meta=ul_img_metas, **kwargs
         )
+        non_empty = [bool(len(i)) for i in pseudo_labels] if self.filter_empty_annotations else [True]
+        get_unlabeled_loss = pseudo_ratio >= self.min_pseudo_label_ratio and any(non_empty)
+
+        if dist.is_initialized():
+            reduced_get_unlabeled_loss = torch.tensor(int(get_unlabeled_loss)).to(current_device)
```
kprokofi, line 190: Why do we need to wrap the bool variable to tensor and set the device?

src/otx/algorithms/detection/adapters/mmdet/models/detectors/mean_teacher.py
```diff
@@ -182,22 +183,29 @@ def forward_train(
...
+            dist.all_reduce(reduced_get_unlabeled_loss)
+            if dont_have_to_train := not get_unlabeled_loss and reduced_get_unlabeled_loss > 0:
```
kprokofi, line 192: If dont_have_to_train is True -> then we don't need to compute unlabeled loss. Why do we set get_unlabeled_loss to True if it was False before? Can we avoid such behavior, it is too complicated to understand when read the code.

src/otx/algorithms/detection/adapters/mmdet/models/detectors/mean_teacher.py
```diff
@@ -182,22 +183,29 @@ def forward_train(
...
+                get_unlabeled_loss = True
+                non_empty[0] = True
+
         if self.filter_empty_annotations:
```
kprokofi: Please, put this block inside "if get_unalbeled_loss" statement, since we don't need to compute this when pseudo_ratio < self.min_pseudo_label_ratio. I know it wasn't written by you, but it just catches my eye.

eunwoosh: Thanks for comment. But it seems that if this statement is moved into `if get_unlabeled_loss`, which argumentr are given to `self._visual_online()` can be different, when `get_unlabeled_loss` is `True` and `self.visualize` is `True`. Is it ok?

kprokofi:
@eunwoosh Let's merge it as is with this workaround. I think we can return to the refactoring later, when adding this feature to the new OTX2.0 design

kprokofi (approved):

sovrasov (approved):

PR #2739, opened 2023-12-19
Enable multilabel classification

sungmanc:
### Summary

Currently, although I've made separate classes for the multi-label, multi-class, it'll be integrated after H-label classfication integration. 

Experimental results
image

### Checklist

### License

```python
```


src/otx/algo/classification/heads/custom_multilabel_non_linear_cls_head.py
```diff
@@ -0,0 +1,124 @@
...
+from torch import nn
+
+from .custom_multilabel_linear_cls_head import AnglularLinear
+
+if TYPE_CHECKING:
+    from mmpretrain.structures import DataSample
+
+
+@MODELS.register_module()
+class CustomMultiLabelNonLinearClsHead(MultiLabelClsHead):
```
kprokofi, line 24: In OTX1.X I thought that these two heads (NonLinear and Linear) can be merged into one class and initialized by a config.  The main obstacle was that we have one template for both tasks. In OTX2.0 we have a freedom to divide these two tasks into two templates, so I think now we can merge this into one class MultiLabelClsHead. 

It seems like the _init_layers() the main difference between two classification heads
image

Could you consider that?

src/otx/recipe/multilabel_classification/efficientnet_b0_light.yaml
```diff
@@ -0,0 +1,75 @@
+# @package _global_
```
kprokofi, line 1: Why are all models called "light"?

kprokofi (commented):
Overall LGTM
As we have already discussed, I am glad that we can divide the classification into 3 different tasks now and modify them independently!
However, I have a proposal for combining MultilabelClsHead:

src/otx/algo/classification/heads/custom_multilabel_non_linear_cls_head.py
sungmanc: I could consider, however, is there no problem of the backward compatibility ? We've been using the `NonLinearHead` to the `MobileNet-v3-large-1.0` template for a long time. So, I have a concern about the backward compatibility

src/otx/recipe/multilabel_classification/efficientnet_b0_light.yaml
sungmanc: image
`light` represents the model what we used before at the OTX1.X and Geti. Other models are represents the new model had new head.

MarkByun (commented):
Could you please share your experimental result?

src/otx/core/data/entity/classification.py
```diff
@@ -74,3 +74,58 @@ def collate_fn(
...
+
+    :param labels: labels as integer indices
+    """
+
+    @property
+    def task(self) -> OTXTaskType:
+        """OTX Task type definition."""
+        return OTXTaskType.MULTI_LABEL_CLS
+
+    labels: LongTensor
```
vinnamkim:
```suggestion
    :param labels: Multi labels represented as an one-hot vector.
    """

    @property
    def task(self) -> OTXTaskType:
        """OTX Task type definition."""
        return OTXTaskType.MULTI_LABEL_CLS

    labels: LongTensor
```
Can we make this as an one-hot vector?

tests/regression/test_regression.py
```diff
@@ -87,11 +87,11 @@ def _test_regression(
 class TestMultiClassCls(BaseTest):
     # Test case parametrization for model
     MODEL_TEST_CASES = [  # noqa: RUF012
-        ModelTestCase(task="classification", name="otx_deit_tiny"),
-        ModelTestCase(task="classification", name="otx_dino_v2"),
-        ModelTestCase(task="classification", name="otx_efficientnet_b0"),
-        ModelTestCase(task="classification", name="otx_efficientnet_v2"),
-        ModelTestCase(task="classification", name="otx_mobilenet_v3_large"),
+        ModelTestCase(task="multi_class_classification", name="otx_deit_tiny"),
```
vinnamkim: image
You've not changed this directory in this PR.

tests/regression/test_regression.py
```diff
@@ -132,3 +132,51 @@ def test_regression(
             fxt_num_repeat=fxt_num_repeat,
             tmpdir=tmpdir,
         )
+
+class TestMultilabelCls(BaseTest):
+    # Test case parametrization for model
+    MODEL_TEST_CASES = [  # noqa: RUF012
+        ModelTestCase(task="multi_label_classification", name="efficientnet_b0_light"),
```
vinnamkim: It seems not aligned with the real path `recipe/multilabel_classification`.

tests/regression/test_regression.py
sungmanc: Ah, thanks :)

src/otx/core/data/entity/classification.py
sungmanc: Could I ask you why we need to use the one-hot vector in this phase ? Since the `MultiLabelClsHead` conducts the label convert process, I think we don't need to do that from our side. https://github.com/open-mmlab/mmpretrain/blob/e95d9acb8919c59f7f1fab24a7701bdc23d25337/mmpretrain/models/heads/multi_label_cls_head.py#L99

vinnamkim: First of all, I saw that one-hot vector conversion is done in the evaluation step
https://github.com/openvinotoolkit/training_extensions/blob/6dfe06b9b28de932183bb70786a5efdf95f84ae0/src/otx/core/model/module/classification.py#L167
I think it would be better to be moved to the data pipeline for two perspectives: 1) reducing computation in the main thread and 2) clean code

About the link you gave, I can see [the very close line](https://github.com/open-mmlab/mmpretrain/blob/e95d9acb8919c59f7f1fab24a7701bdc23d25337/mmpretrain/models/heads/multi_label_cls_head.py#L96C16-L96C16) above you pointed seeks a one hot vector from `gt_score`. It seems no problem for me.

sungmanc: Okay, then, it seems that we need to implement a function that calculates the number of classes to make a one-hot vector. A user already gives the `num_classes` to the model.head information, it's hard to use at the `Dataset` part. 

In my thoughts, this line https://github.com/openvinotoolkit/training_extensions/blob/6dfe06b9b28de932183bb70786a5efdf95f84ae0/src/otx/core/data/dataset/classification.py#L57 would be changed like below,

```python3
class OTXMultilabelClsDataset(OTXDataset[MultilabelClsDataEntity]):
    """OTXDataset class for multi-label classification task."""
    def __init__(***):
        super().__init__(***)
        self.num_classes = self._get_num_classes()

    def _get_num_classes(self):
        # find the max index of the labels by iterating the datumaro dataset
        # Or could we use the datumaro function to get the max label of the `Categories`?

    def _to_one_hot(self, label: Label):
        # Convert the label to the one-hot vector

    def _get_item_impl(self, index: int) -> MultilabelClsDataEntity | None:
        item = self.dm_subset.get(id=self.ids[index], subset=self.dm_subset.name)
        img = item.media_as(Image)
        img_data, img_shape = self._get_img_data_and_shape(img)

        label_anns = [ann for ann in item.annotations if isinstance(ann, Label)]

        entity = MultilabelClsDataEntity(
            image=img_data,
            img_info=ImageInfo(
                img_idx=index,
                img_shape=img_shape,
                ori_shape=img_shape,
                pad_shape=img_shape,
                scale_factor=(1.0, 1.0),
            ),
            labels=torch.as_tensor([self._to_one_hot(ann.label) for ann in label_anns]),
        )

        return self._apply_transforms(entity)

    @property
        """Collection function to collect MultilabelClsDataEntity into MultilabelClsBatchDataEntity in data loader."""
        return MultilabelClsBatchDataEntity.collate_fn
```

Or do we have any other option? Please correct me if I was wrong

vinnamkim: You may access `self.dm_subset.categories()` to get the label categories in Datumaro. It will allow you to get the number of total labels in the dataset I think.

src/otx/algo/classification/heads/custom_multilabel_non_linear_cls_head.py
kprokofi: Do we have there a problem with loading pre-trained weights if we change Class name? If it is the same -> then I think we can consider such change.

sungmanc: Could I consider what you suggested at the later phase? I think integrating H-label classification is more important item.

sungmanc:
> Could you please share your experimental result?

I attached it at the summary. Similar with multi-class classification, there were performance drops at the small dataset and maybe it comes from the LR scheduling. I'll analyze more at the later phase.

tests/regression/test_regression.py
sungmanc: Done, thanks. https://github.com/openvinotoolkit/training_extensions/pull/2739/commits/b6beac6aeb95277e4a8a95eb2026ea78f43f8687

tests/regression/test_regression.py
sungmanc: Done. https://github.com/openvinotoolkit/training_extensions/pull/2739/commits/b6beac6aeb95277e4a8a95eb2026ea78f43f8687

src/otx/core/data/entity/classification.py
sungmanc: Done. https://github.com/openvinotoolkit/training_extensions/pull/2739/commits/b6beac6aeb95277e4a8a95eb2026ea78f43f8687

MarkByun (dismissed):
We need to check the accuray and performance again when current pipeline is stabled.

sungmanc:
> We need to check the accuray and performance again when current pipeline is stabled.

Sure, I'll definitely do that

vinnamkim (dismissed):
LGTM.

jaegukhyun (approved):

MarkByun (approved):

vinnamkim (approved):

PR #2747, opened 2023-12-22
Enable action detection task for v2

jaegukhyun:
### Checklist

### License

```python
```


src/otx/core/data/transform_libs/mmaction.py
```diff
@@ -37,7 +46,162 @@ def __call__(self, entity: ActionClsDataEntity) -> dict:
...
+        frame_dir = Path(frame_path).parent
+        extension = Path(frame_path).suffix[1:]
+        shot_info = (1, len(os.listdir(frame_dir)))
+        timestamp = int(Path(frame_path).stem.split("_")[-1])
+
+        return frame_dir, extension, shot_info, timestamp
+
+
+@TRANSFORMS.register_module(force=True)
+class LoadAnnotations:
```
vinnamkim: If this class overrides the existing one from mmaction, please notify it explicitly such as

```suggestion
@TRANSFORMS.register_module(force=True)
class LoadAnnotations(_LoadAnnotations):
```

src/otx/core/data/transform_libs/mmaction.py
```diff
@@ -70,6 +234,51 @@ def transform(self, results: dict) -> ActionClsDataEntity:
         )
 
 
+@TRANSFORMS.register_module(force=True)
+class PackActionDetInputs(MMPackActionInputs):
```
vinnamkim: Is it possible to merge `PackActionClsInputs` and `PackActionDetInputs` into `PackActionInputs` as same as 
https://github.com/openvinotoolkit/training_extensions/blob/6f8e3078e12a55fc6375f0d8aba7e91a603e65aa/src/otx/core/data/transform_libs/mmdet.py#L46

We want to minimize the change for the mm* configurations for the better compatibility. It seems that we have to change `PackActionInputs` to `PackActionDetInputs` in the mm* configuration.

src/otx/core/data/dataset/action_detection.py
```diff
@@ -0,0 +1,67 @@
...
+                bboxes,
+                format=tv_tensors.BoundingBoxFormat.XYXY,
+                canvas_size=img_shape,
+            ),
+            labels=torch.nn.functional.one_hot(
+                torch.as_tensor([ann.label for ann in bbox_anns]),
+                self.num_classes,
+            ).to(torch.float),
+            frame_path=item.media.path,
+            proposal_file=self.dm_subset.infos().get(f"{self.dm_subset.name}_proposals", None),
```
sungmanc: Do we need to put the proposal_file information to every Entity? It might be needed at the `Model` forward, however, I think putting the same information to the every entity is not the optimal.

src/otx/core/model/entity/action_detection.py
```diff
@@ -0,0 +1,141 @@
...
+        for output in outputs:
+            if not isinstance(output, ActionDataSample):
+                raise TypeError(output)
+
+            output_scores, output_labels = output.pred_instances.scores.max(-1)
+            output.pred_instances.bboxes[:, 0::2] *= output.img_shape[1]
+            output.pred_instances.bboxes[:, 1::2] *= output.img_shape[0]
+            scores.append(output_scores)
+            bboxes.append(
+                tv_tensors.BoundingBoxes(
```
vinnamkim, line 127: Please use `tv_tensors.wrap()` to follow the best practices

https://pytorch.org/vision/0.16/auto_examples/transforms/plot_tv_tensors.html#using-tv-tensors-wrap

src/otx/core/data/entity/action_detection.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+@register_pytree_node
+@dataclass
+class ActionDetDataEntity(OTXDataEntity):
+    """Data entity for action classification task.
+
+    Args:
+        bboxes: 2D bounding boxes for actors.
+        labels: Video's action labels.
```
vinnamkim: Is this an one-hot vector? If so, please update it with more details.

src/otx/core/data/entity/action_detection.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+
+    bboxes: tv_tensors.BoundingBoxes
+    labels: LongTensor
+    frame_path: str | None = None
+    proposal_file: str | None = None
+    proposals: tv_tensors.BoundingBoxes | None = None
```
vinnamkim: Please add docstrings for this member variables and why it is nullable.

src/otx/core/data/entity/action_detection.py
```diff
@@ -0,0 +1,86 @@
...
+    def collate_fn(cls, entities: list[ActionDetDataEntity]) -> ActionDetBatchDataEntity:
+        """Collection function to collect `ActionClsDataEntity` into `ActionClsBatchDataEntity`."""
+        batch_data = super().collate_fn(entities)
+        return ActionDetBatchDataEntity(
+            batch_size=batch_data.batch_size,
+            images=batch_data.images,
+            imgs_info=batch_data.imgs_info,
+            bboxes=[entity.bboxes for entity in entities],
+            labels=[entity.labels for entity in entities],
+            proposals=[entity.proposals for entity in entities],
```
vinnamkim, line 81: Is it true that `proposals` can be `None`?

MarkByun:
In the logic migration, one of the important things is checking the result of the regression test so we need your experimental results to review this code. Could you please share the result if you have it?

src/otx/core/data/dataset/action_detection.py
jaegukhyun: 9c6a904 I moved adding proposal function from data pipeline to dataset. So now entity will have proposal info not proposal file info

src/otx/core/data/transform_libs/mmaction.py
jaegukhyun: [d14d790](https://github.com/openvinotoolkit/training_extensions/pull/2747/commits/d14d7905ba067e75b84a1441e2bac302ab299ae8) LoadAnnotation is a new function in OTX, so I removed force=True parameters

src/otx/core/data/transform_libs/mmaction.py
jaegukhyun: [d14d790](https://github.com/openvinotoolkit/training_extensions/pull/2747/commits/d14d7905ba067e75b84a1441e2bac302ab299ae8) I merged PackActionClsInputs and PackAcionDetInputs

src/otx/core/model/entity/action_detection.py
jaegukhyun: I'm not sure we can use that function here. tv_tensros.wrap() requires torch tensor and reference tv_tensors, but here we only have torch tensors. Could you explain how can we use wrap function here?

src/otx/core/data/entity/action_detection.py
jaegukhyun: Yes it is, I modify docstring

src/otx/core/data/entity/action_detection.py
jaegukhyun: 9c6a904 I removed nullable variables

src/otx/core/data/entity/action_detection.py
jaegukhyun: 9c6a904 No it can't be None, I modified the codes.

jaegukhyun:
> In the logic migration, one of the important things is checking the result of the regression test so we need your experimental results to review this code. Could you please share the result if you have it?

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/jaegukhy/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/jaegukhy/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">


VERSION | ACCURACY | ITER_TIME
-- | -- | --
OTX1.5 | 0.2864 | 1.073
OTX2.0 | 0.2775 | 0.311



</body>

</html>
Here is an experimental results for 5% of JHMDB dataset. I noticed that OTX2.0 have higher training loss than OTX1.5. I'll look into this issue in next phase. I think we can merge this PR, because OTX2.0 is able to train action detection dataset.

src/otx/core/data/transform_libs/mmaction.py
vinnamkim: Then, why do we need this? How can `mmaction` originally import annotations? It is not desired to register a new own transform. This is because we mandatorily require an user who wants to migrate `mmaction` based workflow to OTX 2.0 to include that transform. We want to avoid it as much as possible.

MarkByun:
I wonder whether 2.0 score is normal or not because the score is very low even though the score is better than 1.5. Could you please let me know full training dataset's score also?

src/otx/core/model/entity/action_detection.py
vinnamkim: Sorry it was my miss

jaegukhyun:
> I wonder whether 2.0 score is normal or not because the score is very low even though the score is better than 1.5. Could you please let me know full training dataset's score also?

The accuracy table was wrong, I changed the table

src/otx/core/data/entity/action_detection.py
vinnamkim: Is it correct?
https://github.com/openvinotoolkit/training_extensions/blob/9c6a9046ae5fee2c248d7aa82fa0dc75a1bd12d5/src/otx/core/data/entity/action_detection.py#L62

src/otx/core/data/transform_libs/mmaction.py
jaegukhyun: mmaction adds gt_bboxes and gt_labels in Dataset class. (https://github.com/open-mmlab/mmaction2/blob/4d6c93474730cad2f25e51109adcf96824efc7a3/mmaction/datasets/ava_dataset.py#L295)
In OTX we need to add this function since we need to add bboxes and labels info to data dictionary from otx entity

MarkByun:
The accuracy issue should be corrected before merged in this version because the difference is very huge. Could you please search the root cause?

src/otx/core/data/entity/action_detection.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+@register_pytree_node
+@dataclass
+class ActionDetDataEntity(OTXDataEntity):
+    """Data entity for action classification task.
+
+    Args:
+        bboxes: 2D bounding boxes for actors.
+        labels: One-hot vector of video's action labels.
```
vinnamkim, line 32: What is `frame_path` and `proposals`?

src/otx/core/data/dataset/action_detection.py
```diff
@@ -0,0 +1,105 @@
...
+            with Path.open(proposal_file_path, "rb") as f:
+                info = pickle.load(f)  # noqa: S301
+                if ",".join(Path(frame_path).stem.rsplit("_", 1)) in info:
+                    proposals = info[",".join(Path(frame_path).stem.rsplit("_", 1))][:, :4]
+                else:
+                    proposals = np.array([[0, 0, 1, 1]], dtype=np.float64)
+        else:
+            proposals = np.array([[0, 0, 1, 1]], dtype=np.float64)
+
+        return proposals
```
vinnamkim:
```suggestion
        if not proposal_file_path.exists():
            return np.array([[0, 0, 1, 1]], dtype=np.float64)
        with Path.open(proposal_file_path, "rb") as f:
            info = pickle.load(f)  # noqa: S301

        return (
            info[",".join(Path(frame_path).stem.rsplit("_", 1))][:, :4]
            if ",".join(Path(frame_path).stem.rsplit("_", 1)) in info
            else np.array([[0, 0, 1, 1]], dtype=np.float64)
        )

        return proposals
```

Please reduce code nesting as possible for the better readability. In addition, why should weuse `np.float64` rather than `np.float32`?

src/otx/core/data/entity/action_detection.py
jaegukhyun: frame_path is data media's file path. We need this to get proper meta information of data entity.
Proposals are pre-calculated actor bounding box proposals. Current algorithm saves training time using this kind of pre-calculated actor proposals.

vinnamkim: You should update them in the docstrings

src/otx/core/data/dataset/action_detection.py
jaegukhyun: ddb54d7 Updated

src/otx/core/data/entity/action_detection.py
jaegukhyun: ddb54d7 Updated

jaegukhyun:
> The accuracy issue should be corrected before merged in this version because the difference is very huge. Could you please search the root cause?

There is a missed variable in loading video pipeline. I updated the table

vinnamkim (approved):
LGTM.

MarkByun (approved):
I approve this but I wonder whether the accuracy is normal or not because the value is very low so the model cannot be used for practical use cases. Could you please share the accuracy for the full dataset?

sungmanc (approved):

PR #2748, opened 2023-12-22
Add OpenVINO IR inference for CLS, OD, SSEG, IS tasks + OMZ OV models validation support  

kprokofi:
### Checklist

### License

```python
```


src/otx/core/engine/test.py
```diff
@@ -33,7 +34,7 @@ def test(cfg: TrainConfig) -> tuple[Trainer, dict[str, Any]]:
     trainer: Trainer = hydra.utils.instantiate(cfg.trainer)
 
     log.info("Starting testing!")
-    if cfg.checkpoint is None:
+    if not isinstance(model.model.model, Model) and cfg.checkpoint is None:
```
vinnamkim: Why do we need this change?

src/otx/core/data/entity/base.py
```diff
@@ -157,10 +157,6 @@ def collate_fn(cls, entities: list[T_OTXDataEntity]) -> OTXBatchDataEntity:
             msg = "collate_fn() input should include a single OTX task"
             raise RuntimeError(msg)
 
-        if not all(entity.image_type == ImageType.TV_IMAGE for entity in entities):
```
vinnamkim, line 175: Why do we need this change?

src/otx/core/model/entity/classification.py
```diff
@@ -217,3 +217,55 @@ def _customize_outputs(
...
+    def forward(
+        self,
+        inputs: MulticlassClsBatchDataEntity,
+    ) -> MulticlassClsBatchPredEntity | OTXBatchLossEntity:
+        """Model forward function."""
+        # If customize_inputs is overrided
+        if self.training:
+            msg = "OV model cannot be used for training"
+            raise RuntimeError(msg)
+        if len(inputs.images) > 1:
```
vinnamkim:
```suggestion
        if inputs.batch_size > 1:
```

src/otx/core/model/entity/classification.py
```diff
@@ -217,3 +217,55 @@ def _customize_outputs(
...
+
+    def _customize_outputs(
+        self,
+        outputs: Any,  # noqa: ANN401
+        inputs: MulticlassClsBatchDataEntity,
+    ) -> MulticlassClsBatchPredEntity:
+        # add label index
+        labels = [outputs.top_labels[0][0]]
+        # add probability
+        scores = [outputs.top_labels[0][2]]
```
vinnamkim: Are these `torch.Tensor`?

src/otx/core/model/entity/detection.py
```diff
@@ -143,3 +142,67 @@ def _customize_outputs(
...
+        return DetectionModel.create_model(self.model_name)
+
+    def _customize_outputs(
+        self,
+        outputs: Any,  # noqa: ANN401
+        inputs: DetBatchDataEntity,
+    ) -> DetBatchPredEntity | OTXBatchLossEntity:
+        # add label index
+        outputs_objects = outputs.objects
+        bboxes = [
```
vinnamkim, line 186: This should be `tv_tensor.BoundingBoxes`.

src/otx/core/model/entity/detection.py
```diff
@@ -143,3 +142,67 @@ def _customize_outputs(
...
+            torch.tensor(
+                [[output.xmin, output.ymin, output.xmax, output.ymax] for output in outputs_objects],
+                dtype=torch.float32,
+            ),
+        ]
+        scores = [torch.tensor([output.score for output in outputs_objects])]
+
+        if self.model.get_label_name(0) == "background":
+            # some OMZ models require labels shift to match target labels
```
vinnamkim: Please add warning for this special case.

src/otx/core/model/module/segmentation.py
```diff
@@ -31,7 +31,7 @@ def __init__(
         super().__init__(otx_model, optimizer, scheduler, torch_compile)
         metric_params = {
             "task": "multiclass",
-            "num_classes": otx_model.model.decode_head.num_classes,
+            "num_classes": otx_model.config.get("decode_head", {}).get("num_classes", None),
```
vinnamkim: Is this can be nullable?

https://github.com/Lightning-AI/torchmetrics/blob/3514d715ef42b8dc0c8d96cb3d581caba512fbd7/src/torchmetrics/classification/jaccard.py#L462

It seems that we cannot pass `None` for `multiclass`. Please try to fail here.

src/otx/recipe/detection/atss_mobilenetv2.yaml
```diff
@@ -132,7 +132,7 @@ model:
         relu_before_extra_convs: true
       bbox_head:
         type: ATSSHead
-        num_classes: 2
+        num_classes: 20
```
vinnamkim: Why do we need this?

tests/unit/core/model/module/test_segmentation.py
```diff
@@ -33,9 +33,7 @@ class TestOTXSegmentationModel:
     def model(self, mocker, fxt_seg_data_entity) -> OTXSegmentationLitModule:
         # define otx model
         otx_model = mocker.MagicMock(spec=MMSegCompatibleModel)
-        otx_model.model = mocker.MagicMock(spec=torch.nn.Module)
-        otx_model.model.decode_head = mocker.MagicMock(spec=torch.nn.Module)
-        otx_model.model.decode_head.num_classes = 2
+        otx_model.config = {"decode_head": {"num_classes": 2}}
```
vinnamkim, line 36: Is it necessary?

src/otx/core/data/transform_libs/torchvision.py
```diff
@@ -9,39 +9,63 @@
...
 
             if type(transform) not in availables:
                 msg = f"transform={transform} is not a valid TorchVision V2 transform"
                 raise ValueError(msg)
 
             transforms.append(transform)
 
+        if return_as_list:
+            return transforms
+
```
vinnamkim: I'm against of adding if-else branches agglomeratively to the existing code. Why don't you make a new transform such as `transform = Identity()`?

src/otx/core/engine/test.py
kprokofi: Because Openvino model runs without any checkpoints, it is a final representation of a model. So, with this condition we couldn't run "otx test" without checkpoint

src/otx/core/data/entity/base.py
kprokofi: Because OV IR should receive Numpy images, not torch vision tensors. This condition raised an error

src/otx/core/model/entity/detection.py
kprokofi: It can work perfectly without tv_tensor.BoundingBoxes. However, for the types consistency I can change it

src/otx/core/model/module/segmentation.py
kprokofi: We can raise an exception in case of "num_classes" is None

tests/unit/core/model/module/test_segmentation.py
kprokofi: Yes, to pass the unit tests since I've changed the logic of how to get the number of classes.

src/otx/core/data/transform_libs/torchvision.py
kprokofi: I've already added a new transform Identity(). However, to use it we need register it in "list_available_transforms" and then, we should return not the torchvision Compose, but a list. So, how do you propose to change "generate" method not using "if-else" statements. 
Maybe we could create separate TransformLib like TorchVisionTransformLib, but OVTransformLib. What do you think?

vinnamkim: > Maybe we could create separate TransformLib like TorchVisionTransformLib, but OVTransformLib. What do you think?

Yes, that looks good

src/otx/core/model/entity/detection.py
vinnamkim: > However, for the types consistency I can change it

Yes, please keep it consistent and care about the runtime type mismatching.

src/otx/core/data/entity/base.py
vinnamkim: There is no `np.ndarray` type in `OTXBatchDataEntity.images`. Please make it consistent.

src/otx/core/engine/test.py
vinnamkim: For a new comer who sees this first time, it is really hard to understand its underlying behavior. Can we enhance this code clearer? We may utilize `cfg.checkpoint` to load the OV IR file. Even if it would be not exactly this idea, I think it should be improved.

src/otx/core/model/module/segmentation.py
vinnamkim: Please raise an exception in that case with a comprehensible message such as `msg = 'if num_classes=None, cannot get JaccardIndex("multiclass") metric'`.

src/otx/core/engine/test.py
kprokofi: I see two options here:
1) instead of model_name we can use checkpoint to load IR model
2) remove raising error in case when checkpoint is None. In some cases even for PyTorch models we may want to use pure model with pre-trained weights to check the score on dataset

src/otx/core/data/entity/base.py
kprokofi: I will add np.array to OTXBatchDataEntity.images, bacause it is essential to use numpy arrays for OV model

src/otx/config/data/openvino.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+train_subset:
+  transform_lib_type: TORCHVISION
+  transforms:
+    - _target_: otx.core.data.transform_libs.torchvision.Identity
+  batch_size: 1
+val_subset:
+  transform_lib_type: TORCHVISION
+  transforms:
+    - _target_: otx.core.data.transform_libs.torchvision.Identity
```
sovrasov: Why do we have to use underscores here? It looks as some kind of irregularity

kprokofi: It comes from Hydra syntaxis. With "__ target __ " we initialize target class
https://hydra.cc/docs/advanced/instantiate_objects/overview/#instantiation-of-builtins

src/otx/core/data/entity/base.py
vinnamkim: Can we maintain the type of `OTXBatchDataEntity.images` as `tv_tensor.Image` only?
I confirmed that we can avoid data copying costs between `np.ndarray` and `tv_tensors.Image` conversion.
Please see the following code. It shows that they share the same pointer to the internal data buffer.

```python
import numpy as np
import torch
from torchvision.tv_tensors import Image

np_tensor = np.zeros([2, 2])
tv_image = Image(data=torch.from_numpy(np_tensor))
back_to_np = tv_image.numpy()
np_tensor += 1

print("np_tensor", np_tensor)
print("tv_image", tv_image)
print("back_to_np", back_to_np)
```

```console
np_tensor [[1. 1.]
 [1. 1.]]
tv_image Image([[[1., 1.],
        [1., 1.]]], dtype=torch.float64, )
back_to_np [[[1. 1.]
  [1. 1.]]]
```

src/otx/core/engine/test.py
vinnamkim: Let's remove this check

kprokofi: 1st option seems to raise problems inside lightning, OV model can't be loaded in this way

src/otx/core/model/module/segmentation.py
kprokofi: [done](https://github.com/openvinotoolkit/training_extensions/pull/2748/commits/fa36076ef72b7ad990a68b70db33ed4eb007a889)

src/otx/core/model/entity/detection.py
kprokofi: [done](https://github.com/openvinotoolkit/training_extensions/pull/2748/commits/fa36076ef72b7ad990a68b70db33ed4eb007a889) 
I added it in model creation block to avoid showing warning every time during validation

src/otx/core/model/entity/detection.py
kprokofi: [done](https://github.com/openvinotoolkit/training_extensions/pull/2748/commits/fa36076ef72b7ad990a68b70db33ed4eb007a889)

src/otx/core/model/entity/classification.py
kprokofi: [done](https://github.com/openvinotoolkit/training_extensions/pull/2748/commits/fa36076ef72b7ad990a68b70db33ed4eb007a889)

src/otx/core/model/entity/classification.py
kprokofi: [done](https://github.com/openvinotoolkit/training_extensions/pull/2748/commits/fa36076ef72b7ad990a68b70db33ed4eb007a889)

src/otx/core/data/entity/base.py
kprokofi: [done](https://github.com/openvinotoolkit/training_extensions/pull/2748/commits/fa36076ef72b7ad990a68b70db33ed4eb007a889) 
I used ToImage transform from torchvision. I removed Identity since it is not needed anymore. In "_customize_inputs" I added conversion from torch.tensort to np.array

src/otx/core/engine/test.py
kprokofi: [done](https://github.com/openvinotoolkit/training_extensions/pull/2748/commits/fa36076ef72b7ad990a68b70db33ed4eb007a889)

src/otx/core/data/transform_libs/torchvision.py
kprokofi: [done](https://github.com/openvinotoolkit/training_extensions/pull/2748/commits/fa36076ef72b7ad990a68b70db33ed4eb007a889) 
It isn't required anymore

src/otx/core/engine/test.py
```diff
@@ -33,9 +33,6 @@ def test(cfg: TrainConfig) -> tuple[Trainer, dict[str, Any]]:
     trainer: Trainer = hydra.utils.instantiate(cfg.trainer)
 
     log.info("Starting testing!")
-    if cfg.checkpoint is None:
```
sungmanc, line 36: So, in this case, users can `test` without checkpoint?

src/otx/core/model/entity/detection.py
```diff
@@ -143,3 +144,65 @@ def _customize_outputs(
...
+    def __init__(self, config: DictConfig) -> None:
+        self.model_name = config.pop("model_name")
+        self.config = config
+        super().__init__()
+
+    def _create_model(self) -> nn.Module:
+        from openvino.model_api.models import DetectionModel
+
+        model = DetectionModel.create_model(self.model_name)
+        if model.get_label_name(0) == "background":
```
sungmanc:
```suggestion
        if model.get_label_name(0).lower() == "background":
```
I suggest to use `lower()` function to cover the "Background" case.

vinnamkim (dismissed):
LGTM

src/otx/core/engine/test.py
kprokofi: Yes, just with pretrained weights. It is useful sometimes to check model and what the random score or just to test pre-trained weights

vinnamkim (approved):

sungmanc (approved):

jaegukhyun (approved):

src/otx/core/model/module/segmentation.py
```diff
@@ -29,9 +29,15 @@ def __init__(
         torch_compile: bool,
     ):
         super().__init__(otx_model, optimizer, scheduler, torch_compile)
+        num_classes = otx_model.config.get("decode_head", {}).get("num_classes", None)
```
harimkang, line 32: Here's a question: as far as I know, an OTXModel doesn't always hold a config, does it?

MarkByun (approved):

src/otx/core/model/module/segmentation.py
kprokofi: Hmm, as far as I know, we always have  "config" in our recipes and rely on it heavily. What cases do you mean?

PR #2749, opened 2023-12-26
Enhance classification task's model performance

vinnamkim:
### Summary

1. Memory caching
    - Fix the bug that the memory cache handler tries to resize and cache an image although there is no space on its memory pool.
    - Fix the bug that `multiprocessing.Manager` can be shutdown not in the main thread (dataloader's worker thread).
    - Reuse the resized image during caching

2. Change image decoder from OpenCV to PIL
    - Change the default image color channel format as RGB and remove the BGR to RGB color channel format conversion in the data preprocessor.

3.  Set `pin_memory=True` in `DataLoader`.

4. Utilize a stacked tensor for a batch of images rather than a list of individual images.

   - This approach is necessary to eliminate the need for stacking image tensors in the main thread.

***
- **This is the performance comparison according to this PR**
image
***

### How to test
I updated our unit tests related to this change and the integration tests must guard the correct training pipeline.

### Checklist

### License

```python
```


harimkang (dismissed):
LGTM

sungmanc (dismissed):

vinnamkim:
<del>Converted to draft to fix https://github.com/openvinotoolkit/training_extensions/actions/runs/7384134398/attempts/1?pr=2749</del>

src/otx/recipe/detection/atss_mobilenetv2.yaml
```diff
@@ -115,7 +115,7 @@ model:
           - 255
           - 255
           - 255
-        bgr_to_rgb: true
+        bgr_to_rgb: false
```
jaegukhyun, line 118: Should this value be always false or follow original mmX config?

src/otx/core/data/entity/action_classification.py
```diff
@@ -72,6 +72,12 @@ def collate_fn(cls, entities: list[ActionClsDataEntity]) -> ActionClsBatchDataEn
             labels=[entity.labels for entity in entities],
         )
 
+    def pin_memory(self) -> ActionClsBatchDataEntity:
+        """Pin memory for member tensor variables."""
+        super().pin_memory()
+        self.labels = [label.pin_memory() for label in self.labels]
+        return self
+
```
jaegukhyun, line 80: It seems ActionDetBatchDataEntity doesn't have this part

src/otx/recipe/detection/atss_mobilenetv2.yaml
sungmanc: After this PR, our data pipeline used the PIL.load(RGB) to load the image, however, mmX used the openCV(BGR). So we need to change the value according to the loading pipeline.

vinnamkim: > After this PR, our data pipeline used the PIL.load(RGB) to load the image, however, mmX used the openCV(BGR). So we need to change the value according to the loading pipeline.

True

jaegukhyun: So it is reverse of original mmX config right?

sungmanc: Yes, if mmX will use the openCV loading.

src/otx/core/data/entity/action_classification.py
vinnamkim: Fixed in https://github.com/openvinotoolkit/training_extensions/pull/2749/commits/072f92dc6c8f36d67828a7cc5efe1ddbb74eacfe

src/otx/recipe/detection/atss_mobilenetv2.yaml
vinnamkim: > So it is reverse of original mmX config right?

Yes

jaegukhyun (approved):

sungmanc (approved):

harimkang (approved):

PR #2751, opened 2023-12-29
Update tools/expreiment.py

eunwoosh:
### Summary

This PR updates tools/expreiment.py as below.

- [x] Make experiment directory name same always
- [x] Parse all metric values in `otx eval` output file.
- [x] Fix minor unit test bug

#### Why is experiment name different in each run before?
When parsing variables in YAML recipe file, keys of variables are saved in a set.
Using the set, not only which values to use but also order of dict (python 3.6+ keeps order in default dict) is decided.
In contrast to my expectation, `set` doesn't have always same order.
So, I added a code to sort the set, which makes constant order in every runs.

#### How is `otx eval` ouptut parsed now?
After this PR is merged, all metrics in `otx eval` output file are parsed and saved in final experiment result file.
For example, if we assume that `eval` is executed after `train`, `export` and `optimize` in detection task,
metrics as below are saved

test_score => f-measure(train)
Recall(train) is newly added.
Precision(train) is newly added.
export_model_score => f-measure(export)
Recall(export) is newly added.
Precision(export) is newly added.
optimize_model_score => f-measure(optimize)
Recall(optimize) is newly added.
Precision(optimize) is newly added.
And other metrics are same as previous

### Checklist

### License

```python
```

### DIFF
--- a/tests/unit/algorithms/segmentation/adapters/mmseg/datasets/pipelines/test_transforms.py
+++ b/tests/unit/algorithms/segmentation/adapters/mmseg/datasets/pipelines/test_transforms.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict
+from typing import Any, Dict, List
 
 import numpy as np
 import pytest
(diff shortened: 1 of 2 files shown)


goodsong81 (dismissed):
LGTM

tests/unit/algorithms/segmentation/adapters/mmseg/datasets/pipelines/test_transforms.py
```diff
@@ -110,11 +110,11 @@ class TestNormalize:
     @pytest.mark.parametrize(
         "mean,std,to_rgb,expected",
         [
-            (1.0, 1.0, True, np.array([[[1.0, 0.0, 0.0]]], dtype=np.float32)),
-            (1.0, 1.0, False, np.array([[[-1.0, 0.0, 0.0]]], dtype=np.float32)),
+            ([1.0 for _ in range(3)], [1.0 for _ in range(3)], True, np.array([[[1.0, 0.0, -1.0]]], dtype=np.float32)),
+            ([1.0 for _ in range(3)], [1.0 for _ in range(3)], False, np.array([[[-1.0, 0.0, 1.0]]], dtype=np.float32)),
```
goodsong81, line 114: Nitpick: [1.0, 1.0, 1.0] is better than [1.0 for _ in range(3)] for readability. But it's all OK.

goodsong81 (approved):
Thanks for the fix.

tests/unit/algorithms/segmentation/adapters/mmseg/datasets/pipelines/test_transforms.py
sungmanc: I also prefer [1.0, 1.0, 1.0]

sungmanc (approved):

tests/unit/algorithms/segmentation/adapters/mmseg/datasets/pipelines/test_transforms.py
eunwoosh: I agree with you. I'll use that format next time.

PR #2758, opened 2024-01-04
Enable smart weight loading 

jaegukhyun:
### Summary
Enable smart weight loading for all task.

### **Note**
**There are some exceptions in copying algorithms for two stage detectors, and ssd-mobilenetv2
They are resolved by next PR**

### How to test
I added unit tests for smart weight loading algo for general cases

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/core/data/module.py
+++ b/src/otx/core/data/module.py
@@ -5,7 +5,8 @@
 from __future__ import annotations
 
 import logging as log
-from typing import TYPE_CHECKING, Any
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
 
 from datumaro import Dataset as DmDataset
 from datumaro.components.annotation import AnnotationType
@@ -23,6 +24,21 @@ if TYPE_CHECKING:
     from otx.core.config.data import DataModuleConfig, InstSegDataModuleConfig
 
     from .dataset.base import OTXDataset
+
+
+@dataclass
+class DataMetaInfo:
+    """Meta information of OTXDataModule.
+
+    This meta information will be used by OTXLitModule.
+    """
+
+    class_names: list[str]
+
+    @property
+    def num_classes(self) -> int:
+        """Return number of classes."""
+        return len(self.class_names)
 
 
 class OTXDataModule(LightningDataModule):
@@ -48,7 +64,9 @@ class OTXDataModule(LightningDataModule):
         VIDEO_EXTENSIONS.append(".mp4")
 
         dataset = DmDataset.import_from(self.config.data_root, format=self.config.data_format)
-        self.classes = [category.name for category in dataset.categories()[AnnotationType.label]]
+        self.meta_info = DataMetaInfo(
+            class_names=[category.name for category in dataset.categories()[AnnotationType.label]],
+        )
 
         config_mapping = {
             self.config.train_subset.subset_name: self.config.train_subset,
@@ -139,12 +157,3 @@ class OTXDataModule(LightningDataModule):
         """Teardown for each stage."""
         # clean up after fit or test
         # called on every process in DDP
-
-    def state_dict(self) -> dict[str, Any]:
-        """Called when saving a checkpoint, implement to generate and save datamodule state.
-
-        Returns:
-            A dictionary containing datamodule state.
-
-        """
-        return {"classes": self.classes}
--- a/src/otx/core/data/module.py
+++ b/src/otx/core/data/module.py
@@ -5,9 +5,10 @@
 from __future__ import annotations
 
 import logging as log
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 from datumaro import Dataset as DmDataset
+from datumaro.components.annotation import AnnotationType
 from lightning import LightningDataModule
 from torch.utils.data import DataLoader
 
@@ -46,10 +47,8 @@ class OTXDataModule(LightningDataModule):
 
         VIDEO_EXTENSIONS.append(".mp4")
 
-        dataset = DmDataset.import_from(
-            self.config.data_root,
-            format=self.config.data_format,
-        )
+        dataset = DmDataset.import_from(self.config.data_root, format=self.config.data_format)
+        self.classes = [category.name for category in dataset.categories()[AnnotationType.label]]
 
         config_mapping = {
             self.config.train_subset.subset_name: self.config.train_subset,
@@ -140,3 +139,12 @@ class OTXDataModule(LightningDataModule):
         """Teardown for each stage."""
         # clean up after fit or test
         # called on every process in DDP
+
+    def state_dict(self) -> dict[str, Any]:
+        """Called when saving a checkpoint, implement to generate and save datamodule state.
+
+        Returns:
+            A dictionary containing datamodule state.
+
+        """
+        return {"classes": self.classes}
--- a/src/otx/core/engine/train.py
+++ b/src/otx/core/engine/train.py
@@ -70,9 +70,11 @@ def train(
 
     log.info(f"Instantiating datamodule <{cfg.data}>")
     datamodule = OTXDataModule(task=cfg.base.task, config=cfg.data)
+    data_meta_info = datamodule.meta_info
 
     log.info(f"Instantiating model <{cfg.model}>")
     model: OTXLitModule = hydra.utils.instantiate(cfg.model)
+    model.meta_info = data_meta_info
 
     if otx_model is not None:
         if not isinstance(otx_model, OTXModel):
@@ -115,10 +117,6 @@ def train(
             # load weight
             if cfg.checkpoint is not None:
                 loaded_checkpoint = torch.load(cfg.checkpoint)
-                model.register_load_state_dict_pre_hook(
-                    datamodule.classes,
-                    loaded_checkpoint["OTXDataModule"]["classes"],
-                )
                 model.load_state_dict(loaded_checkpoint["state_dict"])
             # train
             trainer.fit(model=model, datamodule=datamodule)
--- a/src/otx/core/engine/train.py
+++ b/src/otx/core/engine/train.py
@@ -70,11 +70,9 @@ def train(
 
     log.info(f"Instantiating datamodule <{cfg.data}>")
     datamodule = OTXDataModule(task=cfg.base.task, config=cfg.data)
-    data_meta_info = datamodule.meta_info
-
     log.info(f"Instantiating model <{cfg.model}>")
     model: OTXLitModule = hydra.utils.instantiate(cfg.model)
-    model.meta_info = data_meta_info
+    model.meta_info = datamodule.meta_info
 
     if otx_model is not None:
         if not isinstance(otx_model, OTXModel):
@@ -114,7 +112,7 @@ def train(
         if cfg.resume:
             trainer.fit(model=model, datamodule=datamodule, ckpt_path=cfg.checkpoint)
         else:
-            # load weight
+            # load weight to finetune the model
             if cfg.checkpoint is not None:
                 loaded_checkpoint = torch.load(cfg.checkpoint)
                 model.load_state_dict(loaded_checkpoint["state_dict"])
--- a/src/otx/core/engine/train.py
+++ b/src/otx/core/engine/train.py
@@ -54,6 +54,7 @@ def train(
     Returns:
         A tuple with Pytorch Lightning Trainer and Python dict of metrics
     """
+    import torch
     from lightning import seed_everything
 
     from otx.core.data.module import OTXDataModule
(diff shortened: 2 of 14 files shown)


vinnamkim (commented):
I have questions:
1. What is the smart weight loading?
2. Which does an user scenario this smart weight loading cover?

jaegukhyun:
> I have questions:
> 
> 1. What is the smart weight loading?
> 2. Which does an user scenario this smart weight loading cover?

Smart weight loading is weight loading method for class incremental learning. 
When user trained their model on dataset whose classes are ["car", "bus"]
And user wants to train their trained model with the new dataset whose classes are ["car", "bus", "truck"]
In this case, without smart weight loading, the final classification layer will be initialized since dimension of final classification layer between checkpoint and model are different.
Here smart weight loading will load weights from checkpoint for previous classes and initialize weights for newly added classes.

vinnamkim:
> Smart weight loading is weight loading method for class incremental learning. When user trained their model on dataset whose classes are ["car", "bus"] And user wants to train their trained model with the new dataset whose classes are ["car", "bus", "truck"] 

It seems that we should add a dataset-level meta info to `OTXModel`. Let's call it as `PredictionMetaInfo` (you can name a more appropriate one for it, shorter one) such as

```python
class PredictionMetaInfo:
     pass

@dataclasses
class MultiClassClsPredictionMetaInfo(PredictionMetaInfo):
     num_classes: int
     class_names: list[str]
```

Then, assume `OTXModel` should be created by `OTXModel(prediction_meta_info=PredictionMetaInfo)`.

Now, you can handle weight loading in class incremental learning scenarios by

```python
ckpt = old_model.state_dict()

assert "prediction_meta_info" in ckpt

new_model = OTXModel(prediction_meta_info=new_prediction_meta_info)

new_model.load_state_dict(ckpt) # Smart weight loading can be done

class OTXModel(nn.Module):
     ...

     def load_state_dict(self, ckpt):
           old_info = ckpt["prediction_meta_info"]
           new_info = self.prediction_meta_info
           
           if old_info != new_info:
               self._load_state_dict_incrementally(ckpt)
           else:
               super().load_state_dict(ckpt)

     def _load_state_dict_incrementally(self, ckpt):
           ...

     def state_dict(self):
           ckpt = super().state_dict()
           ckpt["prediction_meta_info"] = asdict(self.prediction_meta_info)
```

This requirement is not only observed in this PR, but also @sungmanc's today presentation,
image
This `HLabelInfo` includes the dataset-level meta information, but is passed in each data sample. In addition, it is related to ModelAPI, so that it is not just required for H-label but for all other tasks.


p.s. Is it fundamentally possible to solve the following problem by implementing own `_load_state_dict_incrementally()` for each `OTXModel`? We can have some general solutions at some model classes, but it is not possible to infer what is the head layer of the model by machine automatically.
> There are some exceptions in copying algorithms for two stage detectors, and ssd-mobilenetv2. They are resolved by next PR

sungmanc:
> It seems that we should add a dataset-level meta info to `OTXModel`. Let's call it as `PredictionMetaInfo` (you can name a more appropriate one for it, shorter one)

Sounds good to me, if we can feed a dataset-lele meta info to the `ModelEntity`, it's great

jaegukhyun:
@vinnamkim @sungmanc 
I'll update codes from updating meta info from dataset to updating meta info from model

src/otx/core/engine/train.py
```diff
@@ -108,7 +111,15 @@ def train(
...
-        trainer.fit(model=model, datamodule=datamodule, ckpt_path=cfg.checkpoint)
+        if cfg.resume:
+            trainer.fit(model=model, datamodule=datamodule, ckpt_path=cfg.checkpoint)
+        else:
+            # load weight
+            if cfg.checkpoint is not None:
+                loaded_checkpoint = torch.load(cfg.checkpoint)
+                model.load_state_dict(loaded_checkpoint["state_dict"])
+            # train
+            trainer.fit(model=model, datamodule=datamodule)
```
vinnamkim:
```suggestion
       # load weight
       if cfg.checkpoint is not None:
           loaded_checkpoint = torch.load(cfg.checkpoint)
           model.load_state_dict(loaded_checkpoint["state_dict"])
       # train
       trainer.fit(model=model, datamodule=datamodule)
```

Is it impossible to drop `cfg.resume`?

src/otx/core/engine/train.py
```diff
@@ -69,9 +70,11 @@ def train(
 
     log.info(f"Instantiating datamodule <{cfg.data}>")
     datamodule = OTXDataModule(task=cfg.base.task, config=cfg.data)
+    data_meta_info = datamodule.meta_info
 
     log.info(f"Instantiating model <{cfg.model}>")
     model: OTXLitModule = hydra.utils.instantiate(cfg.model)
+    model.meta_info = data_meta_info
```
sungmanc:
```suggestion
    log.info(f"Instantiating model <{cfg.model}>")
    model: OTXLitModule = hydra.utils.instantiate(cfg.model)
    model.meta_info = datamodule.meta_info
```

src/otx/core/engine/train.py
```diff
@@ -108,7 +111,15 @@ def train(
 
     if cfg.train:
         log.info("Starting training!")
-        trainer.fit(model=model, datamodule=datamodule, ckpt_path=cfg.checkpoint)
+        if cfg.resume:
+            trainer.fit(model=model, datamodule=datamodule, ckpt_path=cfg.checkpoint)
+        else:
+            # load weight
```
sungmanc:
```suggestion
            # load weight
```
```suggestion
            # load weight to finetune the model
```

src/otx/core/model/module/base.py
```diff
@@ -105,7 +108,58 @@ def configure_optimizers(self) -> dict[str, Any]:
...
+        return state_dict
+
+    def load_state_dict(self, state_dict: dict[str, Any]) -> None:
+        """Load state dictionary from checkpoint state dictionary.
+
+        If checkpoint's meta_info and OTXLitModule's meta_info are different,
+        load_state_pre_hook for smart weight loading will be registered.
+        """
+        ckpt_meta_info = state_dict.pop("meta_info", None)
+        if ckpt_meta_info and self.meta_info and ckpt_meta_info != self.meta_info:
```
vinnamkim, line 145:
```suggestion
        # ckpt_meta_info exists but self.meta_info does not exists
        if ckpt_meta_info and self.meta_info is None:
            msg = "`state_dict` to load has `meta_info`, but the current model has no `meta_info`. " \ 
                       "It is recommended to set proper `meta_info` for the incremental learning case."
            warn(msg)
            return super().load_state_dict(state_dict)
        # This is the case
        if ckpt_meta_info and self.meta_info and ckpt_meta_info == self.meta_info:
            ...
            return ...
        return super().load_state_dict()
```
Please add an user warning for there is `meta_info` in `state_dict` but not in `self`. In addition, there is a return object in `load_state_dict()`. Please see https://pytorch.org/docs/stable/_modules/torch/nn/modules/module.html#Module.load_state_dict

sungmanc (dismissed):

src/otx/core/data/module.py
```diff
@@ -24,6 +26,21 @@
     from .dataset.base import OTXDataset
 
 
+@dataclass
+class DataMetaInfo:
```
sungmanc, line 30: It seems that we need to consider the `BaseClass` (maybe `BaseDataMeta`?) to broaden the tasks. What do you think ? In my case, I need to add additional `HLabelInfo` dataclass and it would be better if there is a baseclass to inherit.

src/otx/core/model/entity/base.py
```diff
@@ -57,3 +60,49 @@ def forward(
...
+        self._register_load_state_dict_pre_hook(self.load_state_dict_pre_hook)
+
+    def load_state_dict_pre_hook(self, state_dict: dict[str, torch.Tensor], prefix: str, *args, **kwargs) -> None:
+        """Modify input state_dict according to class name matching before weight loading."""
+        model2ckpt = self.map_class_names(self.model_classes, self.ckpt_classes)
+
+        for param_name in self.state_dict():
+            model_param = self.state_dict()[param_name].clone()
+            ckpt_param = state_dict[prefix + param_name]
+            if model_param.shape != ckpt_param.shape:
```
vinnamkim: How can we cover the following case?

```python
          src_classes = ['person', 'car', 'tree']
          dst_classes = ['sky', 'ball', 'person']
```

vinnamkim: I believe that this can be fundamentally solved by implementing a logic to search the prediction layer for each `OTXModel`. Like

> p.s. Is it fundamentally possible to solve the following problem by implementing own _load_state_dict_incrementally() for each OTXModel? We can have some general solutions at some model classes, but it is not possible to infer what is the head layer of the model by machine automatically.

vinnamkim: I believe that `meta_info` decides the shape of `OTXModel`, so that `meta_info` should be a member of `OTXModel`. In addition, it means that `OTXModel.load_state_dict()` should be able to handle the increamental learning case by itself.

src/otx/core/data/module.py
jaegukhyun: How about HLabelDataMetaInfo(DataMetaInfo)?

vinnamkim: I told @jaegukhyun to share `DataMetaInfo` across all tasks. `HLabelInfo` can be inherited from `DataMetaInfo`. However, `DataMetaInfo` seems not appealing name to me. We might have a better name in the future or this time by someone.

src/otx/core/engine/train.py
vinnamkim: Furthermore, if `model` can handle `cfg.checkpoint` having `meta_info`, do we need this change?

jaegukhyun: Lightening trainer will load other module's hyperparameters such as optimization, lr scheduling when `trainer.fit(model=model, datamodule=datamodule, ckpt_path=cfg.checkpoint)`
This is useful feature when user wants resume, but this could be unwanted behavior those who wants retrain with previous checkpoint

src/otx/core/model/module/base.py
jaegukhyun: Done by 5bfff3084cfdc99a987f9d3b516c488219b48ca9

src/otx/core/model/entity/base.py
jaegukhyun: Done by 4517f09 and 486a2e3
I added auto detection of classification layers by comparing two models with different num_classes

src/otx/core/model/entity/action_classification.py
```diff
@@ -13,7 +13,7 @@
 )
 from otx.core.data.entity.base import OTXBatchLossEntity
 from otx.core.model.entity.base import OTXModel
-from otx.core.utils.build import build_mm_model
+from otx.core.utils.build import build_mm_model, get_classification_layers
```
sungmanc, line 16: I'm not sure about the naming of `get_classification_layers`, however, I also have no better idea. I hope we could consider the better naming later.

sungmanc (dismissed):

src/otx/core/model/entity/classification.py
```diff
@@ -46,7 +46,8 @@ def device(self) -> device:
             else:
                 return buf.device
 
-    return build_mm_model(config, MODELS, load_from)
+    classification_layers = get_classification_layers(config, MODELS, "model.")
+    return build_mm_model(config, MODELS, load_from), classification_layers
```
vinnamkim, line 50: Is this possible to return the tuple although the return type is `nn.Module` at the line 32?

src/otx/core/model/entity/classification.py
```diff
@@ -63,7 +64,9 @@ def __init__(self, config: DictConfig) -> None:
         super().__init__()
 
     def _create_model(self) -> nn.Module:
-        return _create_mmpretrain_model(self.config, self.load_from)
+        model, classification_layers = _create_mmpretrain_model(self.config, self.load_from)
+        self.classification_layers = classification_layers
```
vinnamkim, line 68: Why don't we create them line by line?
```suggestion
        model = _create_mmpretrain_model(self.config, self.load_from)
        self.classification_layers = get_classification_layers(self.config)
```

jaegukhyun: get_classification_layers funciton require model registry to build model with different num classes

src/otx/core/model/entity/classification.py
jaegukhyun: Changed by 
b095a26 It seems mypy do not test return type strictly

vinnamkim (approved):

sungmanc (approved):

harimkang (approved):

PR #2761, opened 2024-01-05
Enable h-label classification

sungmanc:
### Summary

This PR introduces the H-label Classification for OTX2.0. The final benchmark table will be uploaded soon.
Also, this PR will be edited after migrating https://github.com/openvinotoolkit/training_extensions/pull/2758.

# Notes,
* In this PR, the model metric is different from the OTX1.5. In the OTX1.5, the final metric is [top-1-acc. + mAP], however, I edited the metric to [top-1-acc + multilabel-acc] since it is more reasonable for me. **_Changing the model metric at the OTX side doesn't affect to the Geti._**

The figure below represents the simple design of H-label data part
image

### Checklist

- [X] I have added integration tests to cover my changes.​

### License

```python
```


src/otx/algo/classification/metrics/hlabel_accuracy.py
```diff
@@ -0,0 +1,117 @@
...
+        self.head_idx_to_logits_range = head_idx_to_logits_range
+
+        # Multiclass classification accuracy will be defined later
+        self.multiclass_head_accuracy: list[Accuracy] | None = None
+
+        # Multilabel classification accuracy metrics
+        if num_multilabel_classes > 0:
+            # self.multilabel_accuracy = MultilabelAveragePrecision(
+            #     num_labels=self.num_multilabel_classes, average='macro'
+            # )
```
vinnamkim: Please check again that there are commented lines

src/otx/algo/classification/metrics/hlabel_accuracy.py
```diff
@@ -0,0 +1,117 @@
...
+                for head_range in self.head_idx_to_logits_range.values()
+            ]
+        else:
+            msg = "head_idx_to_logits_range should be set."
+            raise TypeError(msg)
+
+    def update(self, preds: torch.Tensor, target: torch.Tensor) -> None:
+        """Update state with predictions and targets."""
+        if not self.multiclass_head_accuracy:
+            self._set_multiclass_head_accuracy(preds.device)
```
vinnamkim: Why should this be created at the update time?

src/otx/core/data/dataset/classification.py
```diff
@@ -89,3 +93,171 @@ def _convert_to_onehot(self, labels: torch.tensor) -> torch.tensor:
...
+            for i, group in enumerate(exclusive_groups):
+                head_idx_to_logits_range[str(i)] = (last_logits_pos, last_logits_pos + len(group))
+                last_logits_pos += len(group)
+                for j, c in enumerate(group):
+                    class_to_idx[c] = (i, j)
+                    num_single_label_classes += 1
+
+            return {
+                "num_multiclass_heads": len(exclusive_groups),
+                "exclusive_groups": exclusive_groups,
```
vinnamkim: Is this field used in the next step?

src/otx/core/data/dataset/classification.py
```diff
@@ -89,3 +93,171 @@ def _convert_to_onehot(self, labels: torch.tensor) -> torch.tensor:
...
+                "head_idx_to_logits_range": head_idx_to_logits_range,
+                "class_to_idx": class_to_idx,
+                "num_single_label_classes": num_single_label_classes,
+            }
```
vinnamkim: It is cumbersome but can we stick to static type approaches if possible rather than Python dictionary such as dataclass or namedtuple?

src/otx/core/data/dataset/classification.py
```diff
@@ -89,3 +93,171 @@ def _convert_to_onehot(self, labels: torch.tensor) -> torch.tensor:
...
+
+        exclusive_group_info = get_exclusive_group_info(all_groups)
+        single_label_group_info = get_single_label_group_info(all_groups, exclusive_group_info["num_multiclass_heads"])
+
+        merged_class_to_idx = merge_class_to_idx(
+            exclusive_group_info["class_to_idx"],
+            single_label_group_info["class_to_idx"],
+        )
+
+        return HLabelInfo(
```
vinnamkim: Can we wrap this function into
```suggestion
        return HLabelInfo.from_dm_label_groups(self.dm_categories.label_groups)
```

src/otx/core/data/entity/classification.py
```diff
@@ -144,3 +144,126 @@ def pin_memory(self) -> MultilabelClsBatchDataEntity:
...
+@dataclass
+class HlabelClsBatchDataEntity(OTXBatchDataEntity[HlabelClsDataEntity]):
+    """Data entity for H-label classification task.
+
+    :param labels: A list of labels as integer indices
+    :param label_groups: A list of label group
+    """
+
+    labels: list[LongTensor]
+    hlabel_info: list[HLabelInfo]
```
vinnamkim: Is it should be (identical to all samples)?
```suggestion
    hlabel_info: HLabelInfo
```

tests/unit/algo/classification/heads/test_custom_hlabel_cls_head.py
```diff
@@ -0,0 +1,69 @@
...
+                    "0": 0,
+                    "1": 1,
+                    "2": 2,
+                    "3": 3,
+                    "4": 4,
+                    "5": 5,
+                },
+            ),
+        )
+        return [data_sample, data_sample]
```
vinnamkim: Would you move this to more common place so that can be used by other tests? For example,
https://github.com/openvinotoolkit/training_extensions/blob/3441111af702831c836d070ae5d62a4b7932209a/tests/conftest.py#L12-L23

sungmanc:
@vinnamkim , thanks for the comments, I'll update this PR after merging this PR (https://github.com/openvinotoolkit/training_extensions/pull/2758) since it brings the `DataMetaInfo` what I need.

sovrasov:
When label resolving functionality is going to be added? As I see now, we have only metric computation, and therefore we don't need to produce human-readable predictions of the model, but in the future obtaining postprocessed results is also good to add (at least for Geti).

tests/unit/algo/classification/heads/test_custom_hlabel_cls_head.py
sungmanc: Done. https://github.com/openvinotoolkit/training_extensions/pull/2761/commits/fdc033ec6b1fdc6615705ddd75af960a49af9f1d

src/otx/core/data/entity/classification.py
sungmanc: I removed hlabel_info field from HLabel related entities since we could use the `HLabelMetaInfo`. https://github.com/openvinotoolkit/training_extensions/pull/2761/commits/fdc033ec6b1fdc6615705ddd75af960a49af9f1d

src/otx/core/data/dataset/classification.py
sungmanc: Done. https://github.com/openvinotoolkit/training_extensions/pull/2761/commits/fdc033ec6b1fdc6615705ddd75af960a49af9f1d

src/otx/core/data/dataset/classification.py
sungmanc: I removed it, thanks. https://github.com/openvinotoolkit/training_extensions/pull/2761/commits/fdc033ec6b1fdc6615705ddd75af960a49af9f1d

src/otx/algo/classification/metrics/hlabel_accuracy.py
sungmanc: In the previous implementation, this was needed because there is no h-label related information at the initialized phase. 
However, by using the `HLabelMetaInfo` made by `DataModule`, I removed this logic. https://github.com/openvinotoolkit/training_extensions/pull/2761/commits/fdc033ec6b1fdc6615705ddd75af960a49af9f1d

src/otx/algo/classification/metrics/hlabel_accuracy.py
sungmanc: Thanks, removed. https://github.com/openvinotoolkit/training_extensions/pull/2761/commits/fdc033ec6b1fdc6615705ddd75af960a49af9f1d

sungmanc:
@jaegukhyun , @vinnamkim 

This commit(https://github.com/openvinotoolkit/training_extensions/pull/2761/commits/fdc033ec6b1fdc6615705ddd75af960a49af9f1d) 
* introduces the `HLabelMetaInfo` and changed the location of `DataMetaInfo`. For the h-label classification, `HLabelClsDataset` need the information of h-label, however, in the previous implementation it is located at the `DataModule` side. That's the reason why I changed the location.  
* Also I added the `SubsetDataMetaInfo` and `DataMetaInfo` to represent the information for all subsets (train, val, test). Currently, there are no cases that have different data classes between train and val. However, we need to care about the `semi` or `self` manner training.

sungmanc:
> When label resolving functionality is going to be added? As I see now, we have only metric computation, and therefore we don't need to produce human-readable predictions of the model, but in the future obtaining postprocessed results is also good to add (at least for Geti).

It seems great, I'll add that functionality at the later phase.

src/otx/core/data/dataset/base.py
```diff
@@ -27,6 +29,28 @@
...
+@dataclass
+class SubsetDataMetaInfo:
+    """Meta information of each subset datasets."""
+
+    class_names: list[str]
+
+    @property
+    def num_classes(self) -> int:
+        """Return number of classes."""
+        return len(self.class_names)
```
vinnamkim: What's the reason to introduce this?
Is there any use-case where a dataset can have different `num_classes` across its subsets?

src/otx/core/data/entity/classification.py
```diff
@@ -144,3 +146,210 @@ def pin_memory(self) -> MultilabelClsBatchDataEntity:
...
+    @classmethod
+    def collate_fn(
+        cls,
+        entities: list[HlabelClsDataEntity],
+    ) -> HlabelClsBatchDataEntity:
+        """Collection function to collect `OTXDataEntity` into `OTXBatchDataEntity` in data loader."""
+        batch_data = super().collate_fn(entities)
+        return HlabelClsBatchDataEntity(
+            batch_size=batch_data.batch_size,
+            images=batch_data.images,
```
vinnamkim: Please use a batch-wise stacked image tensor such as
https://github.com/openvinotoolkit/training_extensions/blob/fdc033ec6b1fdc6615705ddd75af960a49af9f1d/src/otx/core/data/entity/classification.py#L134

src/otx/core/data/transform_libs/mmpretrain.py
```diff
@@ -43,19 +55,70 @@ def transform(self, results: dict) -> MulticlassClsDataEntity:
...
+        data_entity.img_info.pad_shape = packed_common_inputs["pad_shape"]
+        return data_entity
+
+    def _pack_multilabel_inputs(self, results: dict) -> MultilabelClsDataEntity:
+        """Pack multilabel classification inputs.
+
+        NOTE,
+        Currently, this function is the same with multiclass case.
+        However, it should have different functionality if we consider the ignore_label.
+        That's the reason why I reamined same function.
```
vinnamkim, line 90: I'm still doubt that we should split it. If there is `ignore_label` information, should it be considered when creating `OTXDataEntity`? Why do we need to handle at `PackInputs` which is effective only for `mmx` transforms.

src/otx/core/data/dataset/base.py
sungmanc: In the current status, I think there are no cases that you mentioned. I introduced this class as reasons below
1. At the H-label Classification, HLabelInfo should be generated at the `OTXDataset` level, not `OTXDataModule` level. So, I need to insert the `MetaInfo` from `OTXDataset` to `OTXDataModule`. In this situation, we need to construct the dictionary based `MetaInfo` since `OTXDataset` is the subset of datasets. Since current `OTXLitModule` communicates with the `OTXDataModule`, the overall flow is `OTXDataset generates the subset metainfo` --> `deliver to OTXDataModule` --> `deliver to OTXLitModule`.
2. I think some users might want to use different classes at the train and val, however, AFAIK, it is prohibited now from Datumaro.import_from. So, I want to make additional room for this case. 
3. I'm not 100% sure about the case, however, maybe Self-SL scenario looks similar with number 2.

src/otx/core/data/transform_libs/mmpretrain.py
sungmanc: After some consideration, I think we can integrate all parts here. BTW, I want to do that at the next PR since now I'm focusing on checking the backward compatibility, what do you think?

src/otx/core/data/entity/classification.py
sungmanc: Done, https://github.com/openvinotoolkit/training_extensions/pull/2761/commits/d2fd2089fe44c8c81a573760634ea0992750fa48

src/otx/core/data/transform_libs/mmpretrain.py
vinnamkim: LGTM

src/otx/core/data/dataset/base.py
vinnamkim: > 2. I think some users might want to use different classes at the train and val, however, AFAIK, it is prohibited now from Datumaro.import_from. So, I want to make additional room for this case.
> 3. I'm not 100% sure about the case, however, maybe Self-SL scenario looks similar with number 2.

I don't think so. Rather, if there is a mismatch in labels between subsets, we should notify to the user with an appropriate exception to fail as fast as it can. In addition, in the current flow, it is impossible that they are different between subsets. AFAIK, it is always `dm_dataset.subsets().categories == dm_dataset.categories`.

> At the H-label Classification, HLabelInfo should be generated at the OTXDataset level, not OTXDataModule level.

How `HLabelInfo` can be different between subsets unless `dm_dataset.subsets().categories == dm_dataset.categories`? Even if they are different, so that `OTXDataModule` can have different `HLabelInfo` as its member variables, in that case, you should raise an exception to notify it to an user.

sungmanc: > I don't think so. Rather, if there is a mismatch in labels between subsets, we should notify to the user with an appropriate exception to fail as fast as it can. In addition, in the current flow, it is impossible that they are different between subsets. AFAIK, it is always dm_dataset.subsets().categories == dm_dataset.categories.

Okay, make sense to me for this agenda. 

> How HLabelInfo can be different between subsets unless dm_dataset.subsets().categories == dm_dataset.categories? Even if they are different, so that OTXDataModule can have different HLabelInfo as its member variables, in that case, you should raise an exception to notify it to an user.

What I said is not about the label difference. To make H-label annotation, Datumaro label should be converted to the h-label format label. (https://github.com/sungmanc/training_extensions/blob/d2fd2089fe44c8c81a573760634ea0992750fa48/src/otx/core/data/dataset/classification.py#L142) and `HLabelInfo` is needed in this phase. So, if we follow the current implementation, we might need the code below. I added my thoughts at the comment lines, please check them.
```python3
        dataset = DmDataset.import_from(self.config.data_root, format=self.config.data_format)

        config_mapping = {
            self.config.train_subset.subset_name: self.config.train_subset,
            self.config.val_subset.subset_name: self.config.val_subset,
            self.config.test_subset.subset_name: self.config.test_subset,
        }

        mem_size = parse_mem_cache_size_to_int(config.mem_cache_size)
        mem_cache_mode = (
            "singleprocessing"
            if all(config.num_workers == 0 for config in config_mapping.values())
            else "multiprocessing"
        )
        mem_cache_handler = MemCacheHandlerSingleton.create(
            mode=mem_cache_mode,
            mem_size=mem_size,
        )

        meta_info_dict: dict[str, SubsetDataMetaInfo] = {}
        for name, dm_subset in dataset.subsets().items():
            if name not in config_mapping:
                log.warning(f"{name} is not available. Skip it")
                continue

            self.subsets[name] = OTXDatasetFactory.create(
                task=self.task,
                dm_subset=dm_subset,
                mem_cache_handler=mem_cache_handler,
                cfg_subset=config_mapping[name],
                cfg_data_module=config,
            )
            
            ### If `MetaInfo` is generated from the `OTXDataModule` level,
            ### Need to check the task type to deliver the MetaInfo to the `OTXDataset`
            ### I want to avoid this if/else sentence to avoid the complexity.
            ### That's the reason why I introduced the `SubsetMetaInfo`
            
            if self.task == OTXTaskType.H_LABEL_CLS:
                self.subsets[name].meta_info = self.meta_info
```

If the code flow above looks okay or have other design, please let me know. I'm not a big fan of `SubsetMetaInfo`.

vinnamkim: As I said, we have to exclude that case and notify it to user if there is a mismatch. Therefore,

```python
        meta_infos = []
        for name, dm_subset in dataset.subsets().items():
            ...
            meta_infos += [self.subsets[name].meta_info]
        
        if has_mismatch(meta_infos):
            msg = "..."
            raise ValueError(msg)
         self.meta_info = next(iter(meta_infos))
```

sungmanc: I got your points. https://github.com/openvinotoolkit/training_extensions/pull/2761/commits/a0f493ce8502ca45645315da89367d2a48879017

vinnamkim (approved):
LGTM.

jaegukhyun (approved):

harimkang (approved):

PR #2764, opened 2024-01-05
Tiling v2

eugene123tw:
### Summary
- Introduce tiling for both detection and instance segmentation tasks.
- Implement functions to unpack tiles in Detection Model and Instance Segmentation Model.
- Introduce new entities for handling tiling operations.
- Develop task-specific tiling datasets that generate tile entities.
- [Work in Progress] Ongoing work on unit tests and a CLI test for testing tiling functionality.

### How to test

`otx train +recipe=detection/atss_mobilenetv2_tile.yaml base.data_dir=dataset model.otx_model.num_classes=1`

### Checklist

### License

```python
```


vinnamkim:
Hi @eugene123tw,
Is this reviewable or not?

eugene123tw:
> Hi @eugene123tw, Is this reviewable or not?

@vinnamkim no, I'm still working on it.

src/otx/core/model/entity/detection.py
```diff
@@ -27,6 +29,47 @@
...
+        ):
+            tile_preds: list[DetBatchPredEntity] = []
+            for tile, tile_info in zip(tiles, tile_infos):
+                tile_input = DetBatchDataEntity(
+                    batch_size=1,
+                    images=[tile],
+                    imgs_info=[tile_info],
+                    bboxes=[bboxes],
+                    labels=[labels],
+                )
```
vinnamkim: Can we introduce a function such as `unbind()` to `TileBatchDetDataEntity` to clean this?
```suggestion
        for det_batch_data_entity in inputs.unbind():
            prediction = self.forward(det_batch_data_entity)
            pred_entities.append(merge_detection_tiles(tile_preds))

...

class TileBatchDetDataEntity:
    def unbind(self) -> list[BatchDetDataEntity]:
        ...
```

src/otx/core/utils/tile_merge.py
```diff
@@ -0,0 +1,215 @@
...
+    dataset_items = []
+    anno_id = 0
+
+    for tile_pred in tile_preds:
+        annotations = []
+        for batch_bboxes, batch_labels, batch_scores in zip(tile_pred.bboxes, tile_pred.labels, tile_pred.scores):
+            for bbox, label, score in zip(batch_bboxes.data, batch_labels, batch_scores):
+                _bbox = bbox.detach().cpu().numpy()
+                _label = label.detach().cpu().numpy()
+                _score = score.detach().cpu().numpy()
```
vinnamkim: Can we optimize it by having `torch.Tensor` operation-based un-tiling methods in the future (not in this PR)?

src/otx/core/model/entity/base.py
```diff
@@ -87,6 +88,11 @@ def forward(
     ) -> T_OTXBatchPredEntity | OTXBatchLossEntity:
         """Model forward function."""
         # If customize_inputs is overrided
+        if isinstance(inputs, TileBatchDetDataEntity):
+            return self.unpack_det_tiles(inputs)
+        if isinstance(inputs, TileBatchInstSegDataEntity):
+            return self.unpack_inst_seg_tiles(inputs)
```
vinnamkim: It feels difficult to me. This is because `OTXModel` is the base class. However, to deal with this if-branching, `OTXModel` is doing like that it has knowledge about the specifics of the derived classes. It is not desired in the perspective of object oriented programming.

Can we fix this problem such as

```python
    def forward(
        self,
        inputs: T_OTXBatchDataEntity,
    ) -> T_OTXBatchPredEntity | OTXBatchLossEntity:
        """Model forward function."""
        # If customize_inputs is overrided
        if isinstance(inputs, TileBatchDataEntity): # Common tile batch data entity
            return self.forward_tiles(inputs)

class DerivedOTXModel:
     def forward_tiles(...):
           ...
```

src/otx/core/config/data.py
```diff
@@ -34,6 +43,8 @@ class DataModuleConfig:
     val_subset: SubsetConfig
     test_subset: SubsetConfig
 
+    tile_config: TilerConfig
```
vinnamkim, line 47: You added the tiling configuration to datamodule only. However, I'm feeling that we need to add a tiling flag to model as well. This is because, when we export the model, we should know whether it is tiled model or not, right?

src/otx/config/data/default.yaml
```diff
@@ -4,6 +4,9 @@ data_root: ${base.data_dir}
 mem_cache_size: 1GB
 mem_cache_img_max_size: null
 
+tile_config:
+  enable_tiler: false
+
```
jaegukhyun: Is this mean we can use tiling for all task?

src/otx/recipe/detection/atss_mobilenetv2_tile.yaml
```diff
@@ -0,0 +1,186 @@
+# @package _global_
```
jaegukhyun: recipe/detection/atss_mobilenetv2.yaml is updated. Could you check it?

jaegukhyun (commented):
Could you explain a plan for testing tiling?

src/otx/config/data/default.yaml
samet-akcay: Since this is already a config file, would `_config` suffix add extra description? Would it be an idea to name this section as `tiling`, `tiler` or `tile`?

src/otx/core/data/entity/base.py
```diff
@@ -27,6 +27,7 @@ class ImageInfo(tv_tensors.TVTensor):
         img_id: Image id
         img_shape: Image shape (heigth, width) after preprocessing
         ori_shape: Image shape (heigth, width) right after loading it
+        attributes: Image attributes
```
samet-akcay: Would it be possible to elaborate `attributes`. To me image shape is also an attribute of an image, no?

src/otx/config/data/default.yaml
eugene123tw: Ideally, the implementation of tiling should extend to all tasks. This feature is now available for detection and instance segmentation, as OTX 1.5 introduced support for both tasks. 

Looking ahead, it'd be good for OTX 2.0 to expand tiling to include semantic segmentation and classification tasks, given the use cases (e.g. segmenting/classifying tiny cells on large medical images, etc). 

It's not clear to me if we can use tiling for action tasks, but it doesn't seem like any obvious problems are preventing it from being supported.

src/otx/recipe/detection/atss_mobilenetv2_tile.yaml
eugene123tw: I updated the recipe. Thanks!

src/otx/core/model/entity/detection.py
eugene123tw: @vinnamkim Thanks for the input. I added an unbind method to tile-related entities. It looks neater now.

src/otx/core/config/data.py
eugene123tw: No, the model doesn't need to know if tiling is enabled, this is because tiling feature is activated at the dataset level. It prompts the dataset to break down an image into tiles, and after the model predicts all the tiles, it unfolds the predictions. This way, the model doesn't need to be aware of whether it was trained with tiled images or not, as it remains consistent with the original model.

src/otx/core/model/entity/base.py
eugene123tw: Thanks for the suggestion, I added a common `OTXTileBatchDataEntity` and made changes accordingly
https://github.com/openvinotoolkit/training_extensions/pull/2764/files#diff-cb5e8239808876a102af23ebf579364e24835dd1e12cf6970f6da70e8c27be51R48

src/otx/core/utils/tile_merge.py
eugene123tw: yea, it would be ideal to have a tile merge function that is compatible with PyTorch tensors. Currently, tile merge relies on datumaro, so there are two potential approaches: either create custom tile merge utilities in OTX or request support for torch.tensor in datumaro tile merge. I will make a note for future reference to address this.

eugene123tw:
> Could you explain a plan for testing tiling?

@jaegukhyun I intend to include tiling CLI tests into `tests/integration/cli/test_cli.py` and additionally include various unit tests to cover the implemented changes.

samet-akcay (commented):
I'll continue later on, too many PRs to review :)

MarkByun:
Could you please share your test results?

src/otx/core/data/entity/base.py
vinnamkim: This seems a variable only for tiling to me. Can we move this to `Tile*DataEntity`? We don't want to increase the size of variables in the base object indefinitely.

src/otx/core/config/data.py
vinnamkim: Yea, but I meant that we need it at **the model exportation**. This is because the exported model (OV IR or ONNX) should be able to

1) take a high resolution image input 
2) tile a high resolution image into low resolution tiled images
3) merge predictions from the tiled images

in standalone. In the current way, we do not distinguish whether it is a tiling model or not after the model exportation. I believe that you have to indicate it to the model exportation pipeline at least, https://github.com/openvinotoolkit/training_extensions/pull/2774

If you expect that it is not solvable in this PR, please pile this on the backlog, would you?

src/otx/core/data/entity/base.py
eugene123tw: I removed attributes from ImageInfo and moved tile attributes to `OTXTileBatchDataEntity`.

eugene123tw: I removed attributes from ImageInfo and moved tile attributes to OTXTileBatchDataEntity.

src/otx/core/config/data.py
eugene123tw: @vinnamkim I added a todo to model.py: https://github.com/openvinotoolkit/training_extensions/pull/2764/files#diff-d6eede8717d7452da4455f96f12096f179062a7610b8b7c9502dcecc32d5ff8cR8

vinnamkim (dismissed):

harimkang (dismissed):

MarkByun (dismissed):

src/otx/recipe/detection/atss_mobilenetv2_tile.yaml
```diff
@@ -0,0 +1,189 @@
...
+        meta_keys:
+          - ori_filename
+          - scale_factor
+          - ori_shape
+          - filename
+          - img_shape
+          - pad_shape
+model:
+  otx_model:
+    config:
```
harimkang: Most of the `otx_models` now don't have a `config` argument, which seems to be an old-version model config. This seems to be causing integration tests to fail.

src/otx/recipe/instance_segmentation/maskrcnn_efficientnetb2b_tile.yaml
```diff
@@ -0,0 +1,275 @@
...
+        meta_keys:
+          - img_id
+          - img_path
+          - ori_shape
+          - img_shape
+          - scale_factor
+
+model:
+  otx_model:
+    config:
```
harimkang: Same here.

harimkang:
https://github.com/openvinotoolkit/training_extensions/pull/2805 - Due to integration testing, Configuration Refactoring PR was merged first. This is a PR that brings a lot of refactoring to the configuration side, so I think you should rewrite the recipe.

harimkang (approved):

MarkByun (approved):

vinnamkim (approved):

PR #2769, opened 2024-01-09
Add asynchronous OV IR inference 

kprokofi:
### Checklist

### License

```python
```


src/otx/core/model/entity/base.py
```diff
@@ -106,3 +109,62 @@ def map_class_names(src_classes: list[str], dst_classes: list[str]) -> list[int]
...
+
+class OVModel(OTXModel, Generic[T_OTXBatchDataEntity, T_OTXBatchPredEntity]):
+    """Base class for the OpenVINO model."""
+
+    def __init__(self, config: DictConfig) -> None:
+        self.model_name = config.pop("model_name")
+        self.model_type = config.pop("model_type")
+        self.async_inference = config.pop("async_inference", False)
+        self.num_requests = config.pop("max_num_requests", get_default_async_reqs_num())
+        self.use_tp_mode = config.pop("use_tp_mode", False)
```
sungmanc: For me, `use_tp_mode` is unreadable. How about `use_throughput_mode`? It's a little bit long, though.

src/otx/core/model/entity/base.py
```diff
@@ -106,3 +109,62 @@ def map_class_names(src_classes: list[str], dst_classes: list[str]) -> list[int]
...
+            self.model_name,
+            max_num_requests=self.num_requests,
+            plugin_config=plugin_config,
+        )
+
+        return Model.create_model(model_adapter, model_type=self.model_type)
+
+    def _customize_inputs(self, entity: T_OTXBatchDataEntity) -> dict[str, Any]:
+        # restore original numpy image
+        images = [np.transpose(im.numpy(), (1, 2, 0)) for im in entity.images]
```
sungmanc, line 258: So, it means RGB -> BGR? Do OV_API needs the BGR images?

src/otx/core/model/entity/classification.py
```diff
@@ -227,43 +226,25 @@ def _customize_outputs(
...
-            raise RuntimeError(msg)
-        # restore original numpy image
-        img = np.transpose(entity.images[-1].numpy(), (1, 2, 0))
-        return {"inputs": img}
-
     def _customize_outputs(
         self,
         outputs: Any,  # noqa: ANN401
-        inputs: MulticlassClsBatchDataEntity,
+        inputs: MultilabelClsBatchDataEntity,
```
sungmanc: It seems wrong correction since it's classname is `OVMulticlassClassificationModel`

src/otx/core/model/entity/classification.py
```diff
@@ -227,43 +226,25 @@ def _customize_outputs(
...
     ) -> MulticlassClsBatchPredEntity:
         # add label index
-        labels = [torch.tensor(outputs.top_labels[0][0], dtype=torch.long)]
+        labels = [torch.tensor(out.top_labels[0][0], dtype=torch.long) for out in outputs]
         # add probability
-        scores = [torch.tensor(outputs.top_labels[0][2])]
+        scores = [torch.tensor(out.top_labels[0][2]) for out in outputs]
```
sungmanc: I prefer below since it reduces the comment line.
```python3
pred_labels = [torch.tensor(out.top_labels[0][0], dtype=torch.long) for out in outputs]
pred_scores = [torch.tensor(out.top_labels[0][2]) for out in outputs]
```

tests/integration/cli/test_cli.py
```diff
@@ -138,6 +138,11 @@ def test_otx_ov_test(recipe: str, tmp_path: Path) -> None:
     Returns:
         None
     """
+    if recipe == "instance_segmentation/openvino_model.yaml":
```
sungmanc, line 148: How about maintaining this issue more explicitly ? by using JIRA or Github issue

src/otx/core/data/entity/classification.py
```diff
@@ -68,7 +68,7 @@ def collate_fn(
         batch_data = super().collate_fn(entities)
         return MulticlassClsBatchDataEntity(
             batch_size=batch_data.batch_size,
-            images=tv_tensors.Image(data=torch.stack(batch_data.images, dim=0)),
```
vinnamkim, line 73: Did you check conflicts properly?

src/otx/core/model/entity/base.py
```diff
@@ -106,3 +109,62 @@ def map_class_names(src_classes: list[str], dst_classes: list[str]) -> list[int]
             else:
                 src2dst.append(-1)
         return src2dst
+
+
+class OVModel(OTXModel, Generic[T_OTXBatchDataEntity, T_OTXBatchPredEntity]):
```
vinnamkim, line 215: Why don't we just
```suggestion
class OVModel(OTXModel):
```

src/otx/core/model/entity/base.py
```diff
@@ -106,3 +109,62 @@ def map_class_names(src_classes: list[str], dst_classes: list[str]) -> list[int]
...
+            self.model.set_callback(_callback)
+            for im in numpy_inputs:
+                if not self.model.is_ready():
+                    self.model.await_any()
+                self.model.infer_async(im, user_data=outputs)
+            self.model.await_all()
+        else:
+            outputs = [self.model(im) for im in numpy_inputs]
+
+        return self._customize_outputs(outputs, inputs)
```
vinnamkim, line 282: Is it still synchronous for each batch? I'd rather to make `self.async_inference=True` as the default (batch-wise synchronized inference) for `forward()`. Then, create `def forward_async()` which is truly asynchronous, returning a future of model output. If it is hard to develop `def forward_async()` this time. Please do it later.

src/otx/core/utils/build.py
```diff
@@ -40,6 +40,16 @@ def build_mm_model(config: DictConfig, model_registry: Registry, load_from: str
...
 
 
+def get_default_async_reqs_num() -> int:
+    """Returns a default number of infer request for OV models."""
+    import os
+
+    reqs_num = os.cpu_count()
+    if reqs_num is not None:
+        return max(1, int(reqs_num / 2))
+    return 1
```
vinnamkim: Can we notify this behavior to an user such as
```suggestion
    reqs_num = os.cpu_count()
    reqs_num = max(1, int(reqs_num / 2)) if reqs_num is not None else 1
    msg = f"Set the number of OpenVINO inference requests to {reqs_num} because you did not configure 'max_num_requests'"
    warnings.warn(msg)
    return reqs_num
```

src/otx/core/model/entity/base.py
vinnamkim: It's not about changing color channel. It is about changing an order of tensor dimensions,
`C x H x W => H x W x C`.

src/otx/core/model/entity/base.py
kprokofi: Ok

src/otx/core/model/entity/base.py
kprokofi: Correct, to infer OV model, we need to run np.arrays with H x W x C dimension. 
I checked that BGR is by default used in our OTX. Answering to your question, OV requires BGR images, not RGB

src/otx/core/model/entity/classification.py
kprokofi: Oh, you are right, thanks

tests/integration/cli/test_cli.py
kprokofi: We don't need it. When we add export we change integration tests for OV to e2e regime. It is just to not break inference in the whole. OMZ is not developed anymore, only support current models.

src/otx/core/data/entity/classification.py
kprokofi: I checked train and test with torch model. We have same interface for other tasks, I am not sure why we added torch.stack here. @sungmanc do you have comments/concerns here?

src/otx/core/model/entity/base.py
kprokofi: I did it at first time, however, it leads to mypy issues, because of the types override mismatch with the base class. Adding Generic type helps. Other solution is to use #noqa

src/otx/core/model/entity/base.py
kprokofi: why should we create forward_async()? It works OK in forward() as well. self.async_inference depends on parameter in recipe.yaml. By default it is True for all tasks.

src/otx/core/utils/build.py
kprokofi: We can change it similar to this, yes

src/otx/core/data/entity/classification.py
kprokofi: Ok, I see where can be a problem during validation. I will apply fix accordingly

src/otx/core/utils/build.py
kprokofi: [done](https://github.com/openvinotoolkit/training_extensions/pull/2769/commits/1da85964b12868bb793297fab408ea395fe88500)

src/otx/core/model/entity/classification.py
kprokofi: [done](https://github.com/openvinotoolkit/training_extensions/pull/2769/commits/1da85964b12868bb793297fab408ea395fe88500)

src/otx/core/model/entity/classification.py
kprokofi: [done](https://github.com/openvinotoolkit/training_extensions/pull/2769/commits/1da85964b12868bb793297fab408ea395fe88500)

src/otx/core/model/entity/base.py
kprokofi: [done](https://github.com/openvinotoolkit/training_extensions/pull/2769/commits/1da85964b12868bb793297fab408ea395fe88500)

src/otx/core/data/entity/classification.py
vinnamkim: It came from https://github.com/openvinotoolkit/training_extensions/pull/2749

src/otx/core/model/entity/base.py
vinnamkim: You are doing this
image
But I was asking you change it as a default (implementation) of `forward()` and create this one
image

jaegukhyun (dismissed):

src/otx/core/data/entity/classification.py
sungmanc: I'm not sure how it works without error. `OTXModel(nn.Module)` might raise the error since `batch_data.images` is `List[tv_tensors.Image]` while it expects the `Tensor`. After the PR what @vinnamkim pointed out, there is no `torch.stack` in the `_customized_input`.

kprokofi: I will merge and take a look one more time

src/otx/core/model/entity/base.py
kprokofi: Let's discuss it later, we now need to focus on export

src/otx/core/data/entity/classification.py
kprokofi: OTXBatchDataEntity can receive list of tensors as well. There won't be problems. Training and validation is working
image

kprokofi: We can't use torch stack for OV IR, since it will require resizing operation, but Model_API do it inside. We shouldn't resize it before. Other tasks are also use list[tensor] in batch

kprokofi: Please, let me know if it will raise any issues. For now, it seems to work OK, integration tests prove it

vinnamkim (dismissed):

src/otx/core/model/entity/base.py
```diff
@@ -206,3 +210,63 @@ def _reset_prediction_layer(self, num_classes: int) -> None:
...
         """
         raise NotImplementedError
+
+
+class OVModel(OTXModel, Generic[T_OTXBatchDataEntity, T_OTXBatchPredEntity]):
+    """Base class for the OpenVINO model."""
+
+    def __init__(self, num_classes: int, config: DictConfig) -> None:
+        config = inplace_num_classes(cfg=config, num_classes=num_classes)
+        self.model_name = config.pop("model_name")
```
sovrasov, line 230: These optional config members look like a workaround. I suggest conducting a refactoring in the future to make it regular

src/otx/core/model/entity/base.py
sovrasov: Previously, we were using bgr-rgb swap in OTX 1.x, since the original torch models are all trained in RGB, and model API doesn't color channels order.
Does OTX 2.0 train models in BRG?

sovrasov (dismissed):

jaegukhyun (dismissed):
> You are doing this image But I was asking you change it as a default (implementation) of `forward()` and create this one image


It is worth to think about this

src/otx/core/model/entity/base.py
```diff
@@ -206,3 +210,63 @@ def _reset_prediction_layer(self, num_classes: int) -> None:
             num_classes: Number of classes
         """
         raise NotImplementedError
+
+
+class OVModel(OTXModel, Generic[T_OTXBatchDataEntity, T_OTXBatchPredEntity]):
+    """Base class for the OpenVINO model."""
+
+    def __init__(self, num_classes: int, config: DictConfig) -> None:
```
samet-akcay, line 228: Can you explain the motivation why `num_classes` defined as a separate argument, while the rest of the args are defined within the config?

samet-akcay: Is it possible to explicitly define these args here. `config` sort of hides them, and it is hard to know what is inside  this `config`

src/otx/core/model/entity/base.py
```diff
@@ -206,3 +210,63 @@ def _reset_prediction_layer(self, num_classes: int) -> None:
             num_classes: Number of classes
         """
         raise NotImplementedError
+
+
+class OVModel(OTXModel, Generic[T_OTXBatchDataEntity, T_OTXBatchPredEntity]):
+    """Base class for the OpenVINO model."""
```
samet-akcay: can you provide more description/context here,
- what is OVModel?
- why do we need it?
- what context we would use the `OVModel`

src/otx/core/model/entity/classification.py
```diff
@@ -319,46 +318,25 @@ def _customize_outputs(
...
     def _customize_outputs(
         self,
         outputs: Any,  # noqa: ANN401
         inputs: MulticlassClsBatchDataEntity,
     ) -> MulticlassClsBatchPredEntity:
-        # add label index
-        labels = [torch.tensor(outputs.top_labels[0][0], dtype=torch.long)]
-        # add probability
-        scores = [torch.tensor(outputs.top_labels[0][2])]
+        pred_labels = [torch.tensor(out.top_labels[0][0], dtype=torch.long) for out in outputs]
```
samet-akcay, line 334: can `outputs` be `Any`? I see that `out` has `top_labels` attribute. can you explain what `outputs` actually contain?

src/otx/core/model/entity/detection.py
```diff
@@ -149,60 +147,45 @@ def _customize_outputs(
...
-                format="XYXY",
-                canvas_size=inputs.imgs_info[-1].img_shape,
-            ),
-        ]
-        scores = [torch.tensor([output.score for output in outputs_objects])]
+        bboxes = []
+        scores = []
+        labels = []
+        for obj in outputs:
+            output_objects = obj.objects
```
samet-akcay:
```suggestion
        for output in outputs:
            output_objects = output.objects
```

src/otx/core/model/entity/instance_segmentation.py
```diff
@@ -171,52 +171,42 @@ def _customize_outputs(
...
-                canvas_size=inputs.imgs_info[-1].img_shape,
-            ),
-        ]
-        masks = [tv_tensors.Mask(outputs[3])]
+        bboxes = []
+        scores = []
+        labels = []
+        masks = []
+        for obj in outputs:
+            output_objects = obj.segmentedObjects
```
samet-akcay:
```suggestion
        for output in outputs:
            output_objects = output.segmentedObjects
```

src/otx/core/model/entity/instance_segmentation.py
```diff
@@ -171,52 +171,42 @@ def _customize_outputs(
         )
 
 
-class OVInstanceSegCompatibleModel(OTXInstanceSegModel):
+class OVInstanceSegModel(OVModel):
```
samet-akcay: In here `src/otx/core/model/entity/segmentation.py` we define the class name as `OVSegmentationModel`. In terms of consistency, can we follow the same convention and use `OVInstanceSegmentationModel`. It might be a bit longer, but looks a lot cleaner and readable

src/otx/core/model/module/segmentation.py
```diff
@@ -88,19 +88,22 @@ def validation_step(self, inputs: SegBatchDataEntity, batch_idx: int) -> None:
         if not isinstance(preds, SegBatchPredEntity):
             raise TypeError(preds)
 
-        self.val_metric.update(
-            **self._convert_pred_entity_to_compute_metric(preds, inputs),
-        )
+        pred_list = self._convert_pred_entity_to_compute_metric(preds, inputs)
```
samet-akcay, line 91: Not sure if this function is clear. Why is it not possible to pass `preds` and `inputs` to `self.test_metrics.udpate()`? 
I'm aware that it is not the scope of this PR, maybe we could address this in the future.

samet-akcay: If this conversion is really needed, test_metric should ideally handle this conversion within update. Am I missing anything here?

samet-akcay: Also is it possible to rename `pred_list` to `predictions` to improve the readability

src/otx/core/utils/build.py
```diff
@@ -40,6 +41,17 @@ def build_mm_model(config: DictConfig, model_registry: Registry, load_from: str
     return model
 
 
+def get_default_async_reqs_num() -> int:
```
samet-akcay: How about 
```suggestion
def get_default_num_async_infer_requests() -> int:
```
or `get_default_num_async_infer_reqs`?

a bit longer but easier to read

src/otx/core/utils/build.py
```diff
@@ -40,6 +41,17 @@ def build_mm_model(config: DictConfig, model_registry: Registry, load_from: str
...
+    """Returns a default number of infer request for OV models."""
+    import os
+
+    reqs_num = os.cpu_count()
```
samet-akcay: ?
```suggestion
    num_requests = os.cpu_count()
```
or `num_reqs`?

src/otx/recipe/detection/openvino_model.yaml
```diff
@@ -7,10 +7,15 @@ defaults:
...
   data_format: coco_instances
+  test_subset:
+    batch_size: 64
 
 model:
   otx_model:
-    _target_: otx.core.model.entity.detection.OVDetectionCompatibleModel
+    _target_: otx.core.model.entity.detection.OVDetectionModel
     config:
       model_name: ssd300
```
samet-akcay, line 17: If these are the only args that comes within `config`, I propose to explicitly define them in the base model signature

tests/integration/cli/test_cli.py
```diff
@@ -145,6 +145,11 @@ def test_otx_ov_test(recipe: str, tmp_path: Path) -> None:
...
+        # OMZ doesn't have proper model for Pytorch MaskRCNN interface
+        # Need to change this test when export enabled
+        pytest.skip()
```
samet-akcay: Maybe something like this to see this reason in the logs too?

```python
pytest.skip("OMZ doesn't have proper model for Pytorch MaskRCNN interface. Need to change this test when export enabled")
```
ofc, we might need to divide the line, as it is a bit long :)

tests/unit/core/model/entity/test_base.py
```diff
@@ -43,3 +47,28 @@ def test_smart_weight_loading(self, mocker) -> None:
             current_model.state_dict()["model.head.bias"].index_select(0, indices)
             == prev_model.state_dict()["model.head.bias"],
         )
+
+
+class TestOVModel:
+    @pytest.fixture()
+    def entity(self) -> OTXBatchDataEntity:
```
samet-akcay: how about?
```suggestion
    def batch(self) -> OTXBatchDataEntity:
```
 or `dataset_batch` or `input_batch`, but not `entity` :)

samet-akcay: the reason is entity is not clear and too generic. In the future, I'll propose removing `Entity` from all of these objects :)

samet-akcay (changes requested):
Thanks @kprokofi! I have added some comments in the code base and also have a more general simple questions:
- How many OVModel is an OTX developer expected to maintain now. The ones in here and also ones in modelapi repo?
- If a developer is expected to maintain both of the openvino models, will this add extra maintenance effort?

Thanks again!

src/otx/core/model/entity/base.py
kprokofi: Actually, if model_name is not in config -> the error will be raised. I agree with you that we may refactor config parameters and maybe add some of them in constructor directly

src/otx/core/model/entity/base.py
kprokofi: Actually, num_classes was added by @vinnamkim in OTXModel for unifying representation of the number of classes. As we quickly discussed with Vlad above, we indeed could refactor it latter and add some parameters explicitly in constructor. However, it is not for that PR, I guess.
Let's address this issue a bit later

src/otx/core/model/entity/base.py
kprokofi: Ok

tests/unit/core/model/entity/test_base.py
kprokofi: input_batch looks good for me

src/otx/recipe/detection/openvino_model.yaml
kprokofi: someone can define more parameters here. I will add them there.

src/otx/core/model/module/segmentation.py
kprokofi: We should prepare our outputs from the model in format suitable for metric computation, that is why we use self._convert_pred_entity_to_compute_metric(). Changing this logic requires more discussions, probably
I will rename `pred_list` to `predictions`

src/otx/core/model/entity/instance_segmentation.py
kprokofi: Ok

kprokofi:
> Thanks @kprokofi! I have added some comments in the code base and also have a more general simple questions:
> 
> * How many OVModel is an OTX developer expected to maintain now. The ones in here and also ones in modelapi repo?
> * If a developer is expected to maintain both of the openvino models, will this add extra maintenance effort?
> 
> Thanks again!

I guess, we will maintain only our models come from OTX. Those models from OMZ should be maintained by according team. We only support its loading and inference via ModelAPI.
We will also add OVModel for multilabel and hierarhical classification. Also, OVMOdel for action task should be implemented, but there is a bit harder situation, since we don't have ModelAPI for Action Classification

src/otx/recipe/detection/openvino_model.yaml
kprokofi: image
Actually we have all required parameters here

tests/unit/core/model/entity/test_base.py
kprokofi: [done](https://github.com/openvinotoolkit/training_extensions/pull/2769/commits/0c846b7d0e0e44feaf1f1e42bfd5202b122102a3)

tests/integration/cli/test_cli.py
kprokofi: [done](https://github.com/openvinotoolkit/training_extensions/pull/2769/commits/0c846b7d0e0e44feaf1f1e42bfd5202b122102a3)

src/otx/core/utils/build.py
kprokofi: [done](https://github.com/openvinotoolkit/training_extensions/pull/2769/commits/0c846b7d0e0e44feaf1f1e42bfd5202b122102a3)

src/otx/core/utils/build.py
kprokofi: [done](https://github.com/openvinotoolkit/training_extensions/pull/2769/commits/0c846b7d0e0e44feaf1f1e42bfd5202b122102a3)

src/otx/core/model/entity/instance_segmentation.py
kprokofi: [done](https://github.com/openvinotoolkit/training_extensions/pull/2769/commits/0c846b7d0e0e44feaf1f1e42bfd5202b122102a3)

src/otx/core/model/module/segmentation.py
kprokofi: [Done](https://github.com/openvinotoolkit/training_extensions/pull/2769/commits/5fd6f80cf57fcaa7d8685169944bb8a28a801563)

src/otx/core/model/entity/classification.py
kprokofi: [I added proper types](https://github.com/openvinotoolkit/training_extensions/pull/2769/commits/0c846b7d0e0e44feaf1f1e42bfd5202b122102a3)

src/otx/core/model/entity/base.py
kprokofi: [Done](https://github.com/openvinotoolkit/training_extensions/pull/2769/commits/0c846b7d0e0e44feaf1f1e42bfd5202b122102a3)

kprokofi:
@samet-akcay , could you, please, check changes? It is important to merge it today/early tomorrow. We should start to check export - inference IR e2e results.

src/otx/core/model/entity/base.py
samet-akcay: alright. I see that `config` is passed to many other objects, so we could think about a wider refactor in the future addressing those too. Thanks!

src/otx/recipe/detection/openvino_model.yaml
samet-akcay: thanks, yeah, this is relevant to what we discussed above. We could address it later on then

samet-akcay (approved):
thanks for the effort!

kprokofi:
@vinnamkim @jaegukhyun Could you approve once again and merge, please? Thanks!

jaegukhyun (approved):

vinnamkim (approved):

PR #2770, opened 2024-01-10
Add medium and large reg tests for multiclass and multilabel classifications

sungmanc:
### Checklist

### License

```python
```

### DIFF
--- a/tests/regression/test_regression.py
+++ b/tests/regression/test_regression.py
@@ -107,8 +107,7 @@ class TestMultiClassCls(BaseTest):
             extra_overrides={"trainer.max_epochs": "20"},
         )
         for idx in range(1, 4)
-    ]
-    DATASET_TEST_CASES.extend([
+    ] + [
         DatasetTestCase(
             name=f"multiclass_CUB_medium",
             data_root=Path("multiclass_CUB_medium"),
@@ -123,7 +122,7 @@ class TestMultiClassCls(BaseTest):
             num_classes=20,
             extra_overrides={"trainer.max_epochs": "20"},
         )
-    ])
+    ]
 
     @pytest.mark.parametrize(
         "model_test_case",
@@ -175,8 +174,7 @@ class TestMultilabelCls(BaseTest):
             extra_overrides={"trainer.max_epochs": "20"},
         )
         for idx in range(1, 4)
-    ]
-    DATASET_TEST_CASES.extend([
+    ] + [
         DatasetTestCase(
             name=f"multilabel_CUB_medium",
             data_root=Path("multilabel_CUB_medium"),
@@ -191,7 +189,7 @@ class TestMultilabelCls(BaseTest):
             num_classes=21,
             extra_overrides={"trainer.max_epochs": "20"},
         )
-    ])
+    ]
 
     @pytest.mark.parametrize(
         "model_test_case",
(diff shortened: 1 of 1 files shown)


tests/regression/test_regression.py
```diff
@@ -108,6 +108,22 @@ class TestMultiClassCls(BaseTest):
         )
         for idx in range(1, 4)
     ]
+    DATASET_TEST_CASES.extend([
```
vinnamkim: Can we abbreviate it such as

```python
DATASET_TEST_CASES = [  # noqa: RUF012
        DatasetTestCase(
            name=f"multiclass_CUB_small_{idx}",
            data_root=Path("multiclass_CUB_small") / f"{idx}",
            data_format="imagenet_with_subset_dirs",
            num_classes=2,
            extra_overrides={"trainer.max_epochs": "20"},
        )
        for idx in range(1, 4)
    ] + [
        ...
    ]
```

sungmanc: Done, https://github.com/openvinotoolkit/training_extensions/pull/2770/commits/532fe304bdfe0759b5c34befd10a0be8c0daa10f. BTW, just curiosity: is it better to read?

vinnamkim: The difference between the previous one and this one is that

1) The previous one makes `DATASET_TEST_CASES` look like a **mutable variable**. At the runtime, it is expanded by the next line `.expand()`.
2) For this one, there is no change of `DATASET_TEST_CASES` after its creation, so that it looks like an **immutable variable**.

It is important to distinguish between mutable and immutable variables.

vinnamkim (approved):
LGTM.

tests/regression/test_regression.py
sungmanc: Thanks for the comment :)

jaegukhyun (approved):

tests/regression/test_regression.py
```diff
@@ -107,6 +107,21 @@ class TestMultiClassCls(BaseTest):
             extra_overrides={"trainer.max_epochs": "20"},
         )
         for idx in range(1, 4)
+    ] + [
+        DatasetTestCase(
+            name=f"multiclass_CUB_medium",
+            data_root=Path("multiclass_CUB_medium"),
+            data_format="imagenet_with_subset_dirs",
+            num_classes=67,
+            extra_overrides={"trainer.max_epochs": "20"},
```
chuneuny-emily, line 116: Each small, medium, and large dataset for each task was created with an adjusted volume considering the training time for each task. Is there any particular reason to limit max_epoch to 20 epochs for performance tests? Is it because the purpose of the test is only to see whether regression occurs compared to before?

tests/regression/test_regression.py
```diff
@@ -214,25 +244,22 @@ class TestObjectDetection(BaseTest):
             extra_overrides={"trainer.max_epochs": "10"},
         )
         for idx in range(1, 4)
+    ] + [
+        DatasetTestCase(
+            name="pothole_medium",
+            data_root="pothole_medium",
+            data_format="coco",
+            num_classes=1,
+            extra_overrides={"trainer.max_epochs": "10"}
```
chuneuny-emily, line 253: Same question here as above. Is there a reason why max_epochs should be different for each task if it's checking simply to check whether there is regression?

tests/regression/test_regression.py
sungmanc: In this phase, the main purpose of sustaining regression test is to check the model performance and iter_time. So, we don't have any problems when we use the fixed epochs.

tests/regression/test_regression.py
sungmanc: Same with the above.

chuneuny-emily (approved):

PR #2772, opened 2024-01-10
Implementation of Engine class & CLI for OTX Entry Point

harimkang:
### Summary

- Engine Class for OTX Entry Point
- CLI Class for OTX CLI
- hydra -> jsonargparse

It's okay to review it, noting that it's a draft for now. (left below task)
[WIP] Organize data config files
[WIP] Docstring Align
[WIP] Unit-Test

### Checklist

### License

```python
```


vinnamkim (commented):
To make code-review possible, break this huge PR into smaller PRs as possible, For example,

1. PR for Python API
2. PR for CLI
3. PR for configuration

harimkang:
> To make code-review possible, break this huge PR into smaller PRs as possible, For example,
> 
> 1. PR for Python API
> 2. PR for CLI
> 3. PR for configuration

Good point, thanks. I'll take a look and see if can split this PR as much as possible.

for_developers/entry_point_examples/test_config.yaml
```diff
@@ -0,0 +1,210 @@
+data_root: ../../tests/assets/classification_dataset
```
ashwinvaidya17, line 1: Can we avoid `data_root` if we already have a `data` section in the config file? If someone changes this parameter, does the data_root in the data class also get overridden?

src/otx/cli/cli.py
```diff
@@ -0,0 +1,337 @@
...
+                "engine",
+                fail_untyped=False,
+                sub_configs=True,
+                instantiate=False,
+                skip=engine_skip,
+            )
+            sub_parser.link_arguments("data_root", "engine.data_root")
+
+            # Model Settings
+            from otx.core.model.entity.base import OTXModel
```
ashwinvaidya17, line 153: What's the reason for having an import statement here?

src/otx/cli/cli.py
```diff
@@ -0,0 +1,337 @@
...
+
+    def run(self) -> None:
+        """Executes the specified subcommand.
+
+        Raises:
+            ValueError: If the subcommand is not recognized.
+        """
+        self.console.print(f"[blue]{OTX_LOGO}[/blue] ver.{__version__}", justify="center")
+        if self.subcommand == "install":
+            from otx.cli.install import otx_install
```
ashwinvaidya17, line 323: Same. Is there a way to avoid this?

src/otx/cli/utils/jsonargparse.py
```diff
@@ -0,0 +1,140 @@
...
+    Returns:
+        Optional[str]: The short description
+    """
+    if component.__doc__ is None:
+        return None
+    docstring = docstring_parser.parse(component.__doc__)
+    return docstring.short_description
+
+
+def flatten_dict(d: dict, parent_key: str = "", sep: str = ".") -> dict:
```
ashwinvaidya17, line 31: Minor nitpicking and more of a personal preference but can we avoid single letter variable names?

src/otx/cli/utils/jsonargparse.py
```diff
@@ -0,0 +1,140 @@
...
+    value: Any,  # noqa: ANN401
+    key: str | None = None,
+    only_unset: bool = False,
+) -> Namespace:
+    """Sets or replaces all items from the given nested namespace.
+
+    Args:
+        value: A namespace to update multiple values or other type to set in a single key.
+        key: Branch key where to set the value. Required if value is not namespace.
+        only_unset: Whether to only set the value if not set in namespace.
```
ashwinvaidya17, line 67: Can you add an example in the docstrings?

src/otx/cli/utils/jsonargparse.py
```diff
@@ -0,0 +1,140 @@
...
+            if not only_unset or prefix + _key not in self:
+                self.update(val, prefix + _key)
+    if _dict_type and key is not None:
+        # Dict or Namespace -> Dict
+        self[key] = dict_to_namespace(self[key]).as_dict()
+    return self
+
+
+# To provide overriding of the Config file
+def apply_config(self: ActionConfigFile, parser: ArgumentParser, cfg: Namespace, dest: str, value: str) -> None:  # noqa: ARG001
```
ashwinvaidya17, line 97: `self` is not referred to in the method so maybe we can convert this to a static method. Or better, we can move this to a different file and move the import statements to the top.

src/otx/core/types/device.py
```diff
@@ -0,0 +1,21 @@
...
+
+# Copyright (C) 2023 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+from enum import Enum
+
+
+class OTXDeviceType(str, Enum):
```
ashwinvaidya17, line 11: Are we going to have any other device types as well? Maybe we can drop the `OTX` prefix.

src/otx/core/utils/instantiators.py
```diff
@@ -0,0 +1,89 @@
...
+
+def instantiate_callbacks(callbacks_cfg: list) -> list[Callback]:
+    """Instantiate a list of callbacks based on the provided configuration.
+
+    Args:
+        callbacks_cfg (list): A list of callback configurations.
+
+    Returns:
+        list[Callback]: A list of instantiated callbacks.
+    """
```
ashwinvaidya17, line 31: Can you provide an example in the docstring as well? Also, can we provide just the class name for it to work? Imagine we have a callback at `otx.callbacks.ModelCheckpoint`. Will passing only `ModelCheckpoint` to `class_path` work?

src/otx/core/utils/instantiators.py
ashwinvaidya17, line 1: Thanks for all the efforts here but can you also provide examples in the methods in this file? It would be nice to see example usages.

tests/integration/cli/test_cli.py
```diff
@@ -12,49 +12,65 @@
 
 # This assumes have OTX installed in environment.
 otx_module = importlib.import_module("otx")
-RECIPE_PATH = Path(inspect.getfile(otx_module)).parent / "recipe"
-ALL_RECIPE_LIST = [str(_.relative_to(RECIPE_PATH)) for _ in RECIPE_PATH.glob("**/*.yaml")]
-RECIPE_OV_LIST = [str(_.relative_to(RECIPE_PATH)) for _ in RECIPE_PATH.glob("**/openvino_model.yaml")]
-RECIPE_LIST = set(ALL_RECIPE_LIST) - set(RECIPE_OV_LIST)
+RECIPE_PATH = Path(inspect.getfile(otx_module)).parent / "configs"
```
ashwinvaidya17, line 15: Maybe not the scope of this PR but maybe we can use fixtures rather than global variables to get these values

src/otx/engine/engine.py
```diff
@@ -0,0 +1,486 @@
...
+                otx predict
+                    --model <CONFIG | CLASS_PATH_OR_NAME> --data_root <DATASET_PATH, str>
+                    --checkpoint <CKPT_PATH, str>
+                ```
+            2. If you have a ready configuration file, run it like this.
+                ```python
+                otx predict --config <CONFIG_PATH, str> --checkpoint <CKPT_PATH, str>
+                ```
+        """
+        lit_module = self._build_lightning_module(
```
ashwinvaidya17, line 335: Since we build a lightning module in each method we can possibly support passing the model in each method as well.
predict/train/test/... can have model as an optional argument. Since the model setter already supports OTX model or a string we can do something like this.
```python
def predict(self,
    model: OTXModel | str | None = None,
    ...
):
    if model:
        self.update_model(model)
        or
        self.model = model
```

One thing I can see is that the `get_model` method also takes in `num_classes` as a parameter. But I am sure we can design a system around this. Something like

```python
def update_model(model: OTXModel | str, datamodule: Datamodule | None = None):
    num_classes = datamodule.num_classes if datamodule else self.datamodule.num_classes
    self._auto_configurator.get_model(
                model=model,
                num_classes=num_classes,
     )
```

Unless I am missing something and model is required in `__init__` for some other initializations as well.

src/otx/engine/engine.py
```diff
@@ -0,0 +1,486 @@
...
+            datamodule = OTXDataModule.from_config(config=data_config)
+            model = instantiate_class(args=(), init=config_dict.pop("model", {}))
+            optimizer = partial_instantiate_class(init=config_dict.pop("optimizer", {}))
+            scheduler = partial_instantiate_class(init=config_dict.pop("scheduler", {}))
+            engine_args = config_dict.pop("engine", config_dict)
+            engine_config = EngineConfig(**engine_args)
+        else:
+            engine_config = config
+
+        data_root = config_dict.pop("data_root", "")
```
ashwinvaidya17, line 389: Is this correct? What happens if `config` is not of instance `str | Path`? In that case `config_dict` is unassigned.

ashwinvaidya17 (commented):
Thanks this is a tremendous effort! I have a few questions but overall the rest looks great.

for_developers/entry_point_examples/test_config.yaml
```diff
@@ -0,0 +1,210 @@
...
+      num_workers: 2
+    mem_cache_size: 1GB
+    mem_cache_img_max_size:
+      - 500
+      - 500
+    include_polygons: false
+max_epochs: 3
+deterministic: false
+precision: 32
+val_check_interval: 1
```
samet-akcay, line 125: Would it be an idea to define these and the rest of the `trainer` arguments as Engine arguments? This way, Engine can control all of the Trainer args?

samet-akcay: This might potentially further organise the config file ?

src/otx/algo/classification/otx_dino_v2.py
```diff
@@ -65,8 +63,8 @@ def forward(self, imgs: torch.Tensor, labels: torch.Tensor = None) -> torch.Tens
 class DINOv2RegisterClassifier(OTXMulticlassClsModel):
     """DINO-v2 Classification Model with register."""
 
-    def __init__(self, config: DictConfig) -> None:
-        self.config = config
+    def __init__(self, config: DictConfig | dict) -> None:
```
samet-akcay, line 66: is this `config` arg part of the mm stuff? I guess OTX user will only see `DINOv2`, which has args explicitly defined in `__init__`?

src/otx/cli/cli.py
```diff
@@ -0,0 +1,337 @@
...
+    def engine_subcommands() -> dict[str, set[str]]:
+        """Returns a dictionary of engine subcommands and their required arguments.
+
+        Returns:
+            A dictionary where the keys are the subcommands and the values are sets of required arguments.
+        """
+        return {
+            "train": set(),
+            "test": {"datamodule"},
+            "predict": {"datamodule"},
```
samet-akcay, line 110: can you explain why `datamodule` is required for `test` and `predict`? We could possibly create a datamodule that has test_dataloader and predict_dataloader, which Engine and CLI could use under the hood?

If the user passes the datamodule or dataloader to test and predict, we could then use it. Otherwise, use the one that the user initially created?

src/otx/cli/cli.py
samet-akcay: I think @harimkang wants to import it only after checking if Engine is available, meaning that torch and other packages are properly installed.

src/otx/cli/cli.py
```diff
@@ -0,0 +1,337 @@
...
+    def instantiate_classes(self) -> None:
+        """Instantiate the necessary classes based on the subcommand.
+
+        This method checks if the subcommand is one of the engine subcommands.
+        If it is, it instantiates the necessary classes such as config, datamodule, model, and engine.
+        """
+        if self.subcommand in self.engine_subcommands():
+            self.config_init = self.parser.instantiate_classes(self.config)
+            self.datamodule = self._get(self.config_init, "data")
+            self.model, optimizer, scheduler = self.instantiate_model()
```
samet-akcay, line 249: do model, optimizer and scheduler depend on each other? Can you clarify why these are returned as part of `instantiate_model`, which is a bit different than how datamodule in Line 248 handled

src/otx/cli/cli.py
```diff
@@ -0,0 +1,337 @@
...
+    def save_config(self) -> None:
+        """Save the configuration for the specified subcommand.
+
+        The configuration is saved as a YAML file in the engine's working directory.
+
+        Args:
+            None
+
+        Returns:
+            None
```
samet-akcay, line 305: maybe we could remove these?
```suggestion
```

src/otx/cli/utils/help_formatter.py
```diff
@@ -0,0 +1,224 @@
...
+}
+
+
+def get_verbosity_subcommand() -> dict:
+    """Parse command line arguments and returns a dictionary of key-value pairs.
+
+    Returns:
+        A dictionary containing the parsed command line arguments.
+
+    Examples:
```
samet-akcay, line 45: Thanks for adding this section. Always makes it easy to follow :)
```suggestion
    Example:
```

src/otx/cli/utils/help_formatter.py
```diff
@@ -0,0 +1,224 @@
...
+        if arguments["subcommand"] in REQUIRED_ARGUMENTS:
+            arguments["verbosity"] = 0
+            if "-v" in sys.argv or "--verbose" in sys.argv:
+                arguments["verbosity"] = 1
+            if "-vv" in sys.argv:
+                arguments["verbosity"] = 2
+    return arguments
+
+
+INTRO_MARKDOWN = (
```
samet-akcay, line 65: would it be an idea to use triple quotes `"""` to improve the readability in the code base as well?

src/otx/cli/utils/help_formatter.py
```diff
@@ -0,0 +1,224 @@
...
+    verbosity_dict = get_verbosity_subcommand()
+    verbose_level = verbosity_dict["verbosity"]
+    subcommand = verbosity_dict["subcommand"]
+
+    def add_usage(self, usage: str | None, actions: Iterable[argparse.Action], *args, **kwargs) -> None:
+        """Add usage information to the formatter.
+
+        Args:
+            usage (str | None): A string describing the usage of the program.
+            actions (Iterable[argparse.Action]): An list of argparse.Action objects.
```
samet-akcay, line 169:
```suggestion
            actions (Iterable[argparse.Action]): A list of argparse.Action objects.
```

src/otx/cli/utils/jsonargparse.py
```diff
@@ -0,0 +1,140 @@
+"""Functions related to jsonargparse."""
+
+# Copyright (C) 2023 Intel Corporation
```
samet-akcay, line 3: time flies :)
```suggestion
# Copyright (C) 2024 Intel Corporation
```

src/otx/cli/utils/jsonargparse.py
```diff
@@ -0,0 +1,140 @@
...
+import docstring_parser
+from jsonargparse import ActionConfigFile, ArgumentParser, Namespace, dict_to_namespace
+
+
+def get_short_docstring(component: TypeVar) -> str | None:
+    """Get the short description from the docstring.
+
+    Args:
+        component (object): The component to get the docstring from
+
```
samet-akcay, line 21: would it be too much effort to add an example section here

src/otx/core/config/device.py
samet-akcay, line 1: can you explain why we need this file

src/otx/core/config/engine.py
samet-akcay, line 1: also this?

samet-akcay: we could use the trainer args, right? `accelerator` and `devices` are already part of trainer args

src/otx/core/config/engine.py
```diff
@@ -0,0 +1,22 @@
...
+
+from otx.core.types.task import OTXTaskType
+
+
+@dataclass
+class EngineConfig:
+    """Configuration class for the engine."""
+
+    data_root: str
+    task: OTXTaskType
```
samet-akcay, line 18: not the scope of this work, but for a future reference. I think we don't need to add `OTX` prefix to this sort of classes we have. This could just be `TaskType`

src/otx/core/config/engine.py
```diff
@@ -0,0 +1,22 @@
...
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+from otx.core.types.task import OTXTaskType
+
+
+@dataclass
+class EngineConfig:
```
samet-akcay, line 14: regarding the configs, I'm not clear why we need this config. And ideally, Engine should consume all of the trainer args, I think

src/otx/core/data/module.py
```diff
@@ -44,11 +45,7 @@ def num_classes(self) -> int:
 class OTXDataModule(LightningDataModule):
     """LightningDataModule extension for OTX pipeline."""
 
-    def __init__(
-        self,
-        task: OTXTaskType,
-        config: DataModuleConfig | InstSegDataModuleConfig,
-    ) -> None:
+    def __init__(self, task: OTXTaskType, config: DataModuleConfig) -> None:
```
samet-akcay, line 48: can you explain this change? and would it be possible to have a datamodule without this `config` at all?

src/otx/core/data/module.py
```diff
@@ -99,6 +96,40 @@ def __init__(
...
             log.info(f"Add name: {name}, self.subsets: {self.subsets}")
 
+    @classmethod
+    def from_config(cls, config: dict | str | Path) -> OTXDataModule:
+        """Create an instance of OTXDataModule from a configuration.
+
+        Args:
+            task (OTXTaskType): The task type.
+            config (DataModuleConfig | str | Path): The configuration object or the path to the configuration file.
+
```
samet-akcay, line 106: can you add an example section here? It would be good to have it here to build the documentation. in addition, users/developers can get a clearer idea.

src/otx/core/data/module.py
```diff
@@ -99,6 +96,40 @@ def __init__(
...
+        Returns:
+            OTXDataModule: An instance of OTXDataModule.
+
+        """
+        if isinstance(config, (str, Path)):
+            with Path(config).open() as f:
+                config = yaml.safe_load(f)
+        if not isinstance(config, dict):
+            msg = "Please double-check data config."
```
samet-akcay, line 115: is it possible to have a more descriptive error message? Like what's wrong with the config

src/otx/core/model/entity/action_classification.py
samet-akcay, line 1: not the scope of this pr, and a future reference:

the `entity` directory is redundant here since core should contain entities anyway. It is adding an extra hierarchy and verbosity

src/otx/core/model/module/base.py
```diff
@@ -16,20 +16,24 @@
...
-        optimizer: torch.optim.Optimizer,
-        scheduler: torch.optim.lr_scheduler.LRScheduler,
         torch_compile: bool,
+        optimizer: OptimizerCallable = lambda p: torch.optim.SGD(p, lr=0.01),
+        scheduler: LRSchedulerCallable = torch.optim.lr_scheduler.ConstantLR,
     ):
         super().__init__()
 
         self.model = otx_model
+        self.torch_compile = torch_compile
```
samet-akcay, line 36: what does torch_compile do?

src/otx/core/model/module/base.py
samet-akcay, line 1: not the scope of this pr, and a future reference:

this hierarchy also does not sound right, `/model/module/base`. but we could discuss it later.

src/otx/core/types/device.py
samet-akcay: I agree, we don't need OTX prefix

src/otx/core/utils/__init__.py
samet-akcay, line 1: I think core should not have any utils, but instead should only have class interfaces. It might be controversial but I would personally prefer not having core, and place these base classes to their corresponding sub-packages as base sub-package. Because 
- it is tricky to maintain this
- it is hard to remember the location of the classes (eg., whether it is in otx.data, or otx.core.data and so on)

We could discuss this later though.

src/otx/core/utils/auto_configuration.py
samet-akcay, line 1: same as above, we should move utils to `otx.utils`

src/otx/core/utils/auto_configuration.py
```diff
@@ -0,0 +1,437 @@
...
+def load_model_configs(task: str) -> dict[str, str]:
+    """Loads the model configurations from the specified directory.
+
+    Returns:
+        A dictionary mapping file names to their corresponding file paths.
+    """
+    model_dir = CONFIG_PATH / "model" / task
+    file_path_dict = {}
+    for file_path in model_dir.iterdir():
+        if str(file_path).endswith("yaml") and file_path.is_file():
```
samet-akcay, line 181:
```suggestion
        if file_path.is_file() and file_path.suffix == ".yaml":
```

src/otx/core/utils/auto_configuration.py
```diff
@@ -0,0 +1,437 @@
...
+    return file_path_dict
+
+
+class AutoConfigurator:
+    """Class responsible for auto configuration of data and model in OTX.
+
+    Args:
+        data_root (str | Path | None, optional): The root directory for data storage. Defaults to None.
+        task (OTXTaskType | None, optional): The current task. Defaults to None.
+
```
samet-akcay, line 193: is it possible to have an `Examples` section here?

src/otx/core/utils/auto_configuration.py
```diff
@@ -0,0 +1,437 @@
...
+        """Returns the current task.
+
+        Raises:
+            RuntimeError: If there are no ready tasks.
+
+        Returns:
+            OTXTaskType: The current task.
+        """
+        if self._task is None:
+            msg = "There are no ready task"
```
samet-akcay, line 224: ?
```suggestion
            msg = "Task not initialized or not ready."
```

src/otx/core/utils/auto_configuration.py
```diff
@@ -0,0 +1,437 @@
...
+
+    def load_default_engine_config(self) -> dict:
+        """Loads the default engine configuration.
+
+        Returns:
+            dict
+        """
+        config_file = CONFIG_PATH / "engine" / f"{self.task.value}.yaml"
+        with Path(config_file).open() as f:
+            return yaml.safe_load(f)
```
samet-akcay, line 296: would it be an idea to refactor this part to a function like `safe_load_config` or something, and used for these methods?

samet-akcay: something like, may not be accurate, i'm just quickly prototyping :)

```python
    def load_default_config(config_path: Path, component: str) -> dict:
        with Path(config_file).open() as f:
            config = yaml.safe_load(f)
        return config.get(component, config)
```

Not sure about `component` name though. Feel free to use a better one :)

src/otx/core/utils/auto_configuration.py
```diff
@@ -0,0 +1,437 @@
...
+
+        Returns:
+            OTXModel: The instantiated OTXModel instance.
+
+        Raises:
+            TypeError: If the model configuration is not a dictionary or DictConfig.
+        """
+        model_config = self.load_default_model_config(model=model, num_classes=num_classes)
+        if not isinstance(model_config, (dict, DictConfig)):
+            msg = "Please double-check model config."
```
samet-akcay, line 335: What would be wrong with the config? Maybe a more descriptive message

src/otx/core/utils/auto_configuration.py
```diff
@@ -0,0 +1,437 @@
...
+        """Get the optimizer from the given optimizer name.
+
+        Returns:
+            OptimizerCallable: The instantiated optimizer object.
+        """
+        optimizer_cfg = self.load_default_optimizer_config()
+        logger.warning(f"Set Default Optimizer: {optimizer_cfg}")
+        optimizer = partial_instantiate_class(init=optimizer_cfg)
+        if optimizer is None:
+            msg = "Please double-check optimizer config."
```
samet-akcay, line 350: same as above

src/otx/core/utils/auto_configuration.py
```diff
@@ -0,0 +1,437 @@
...
+        self,
+        data_root: str | Path | None = None,
+        data_format: str | None = None,
+        mem_cache_size: str | None = None,
+        mem_cache_img_max_size: tuple[int, int] | None = None,
+        include_polygons: bool | None = None,
+        batch_size: int | dict[str, int] | None = None,
+        num_workers: int | dict[str, int] | None = None,
+        transforms: list | dict[str, list] | None = None,
+        transform_lib_type: TransformLibType | dict[str, TransformLibType] | None = None,
```
samet-akcay, line 378: can you explain the difference between the two? Why would we need both transforms and transform_lib_type?

src/otx/core/utils/auto_configuration.py
```diff
@@ -0,0 +1,437 @@
...
+        default_config = self.load_default_data_config()
+
+        # Update DataModuleConfig
+        default_config["config"]["data_root"] = data_root
+        datamodule_arguments = {
+            "data_format": data_format,
+            "mem_cache_size": mem_cache_size,
+            "mem_cache_img_max_size": mem_cache_img_max_size,
+            "include_polygons": include_polygons,
+        }
```
samet-akcay, line 413: let's say we added new args to datamodule in the future. Does this mean that we need to update the args manually here?

src/otx/core/utils/auto_configuration.py
```diff
@@ -0,0 +1,437 @@
...
+        for datamodule_key, datamodule_value in datamodule_arguments.items():
+            if datamodule_value is not None:
+                default_config["config"][datamodule_key] = datamodule_value
+
+        # Update SubsetConfig
+        subset_arguments = {
```
samet-akcay, line 419: same as above

src/otx/core/utils/build.py
samet-akcay, line 1: same as above comment regarding the utils in core

src/otx/core/utils/cache.py
samet-akcay, line 1: same as above. We could potentially keep this somewhere together with `Engine` since it is only used by `Engine`. What do you think?

src/otx/core/utils/instantiators.py
samet-akcay, line 1: is there anywhere else than cli or engine that these instantiators are used?

src/otx/engine/engine.py
```diff
@@ -0,0 +1,486 @@
...
+
+LITMODULE_PER_TASK = {
+    OTXTaskType.MULTI_CLASS_CLS: "otx.core.model.module.classification.OTXMulticlassClsLitModule",
+    OTXTaskType.MULTI_LABEL_CLS: "otx.core.model.module.classification.OTXMultilabelClsLitModule",
+    OTXTaskType.DETECTION: "otx.core.model.module.detection.OTXDetectionLitModule",
+    OTXTaskType.INSTANCE_SEGMENTATION: "otx.core.model.module.instance_segmentation.OTXInstanceSegLitModule",
+    OTXTaskType.SEMANTIC_SEGMENTATION: "otx.core.model.module.segmentation.OTXSegmentationLitModule",
+    OTXTaskType.ACTION_CLASSIFICATION: "otx.core.model.module.action_classification.OTXActionClsLitModule",
+    OTXTaskType.ACTION_DETECTION: "otx.core.model.module.action_detection.OTXActionDetLitModule",
+}
```
samet-akcay, line 49: can you clarify why this is needed? Would it be possible to have more automated approach here

src/otx/engine/engine.py
```diff
@@ -0,0 +1,486 @@
...
+class Engine:
+    """OTX Engine.
+
+    This class defines the common interface for OTX, including methods for training and testing.
+
+    Example:
+    >>> engine = Engine(
+        data_root="dataset/path",
+        task="MULTI_CLASS_CLS",
+    )
```
samet-akcay, line 82:
```suggestion
            >>> engine = Engine(
            ...     data_root="dataset/path",
            ...     task="MULTI_CLASS_CLS",
            ... )
```

samet-akcay: We would need more examples here, covering each case potentially

src/otx/engine/engine.py
```diff
@@ -0,0 +1,486 @@
...
+        Returns:
+            dict[str, Any]: A dictionary containing the callback metrics from the trainer.
+
+        Example:
+        >>> engine.train(
+            max_epochs=3,
+            seed=1234,
+            deterministic=False,
+            precision="32",
+        )
```
samet-akcay, line 185: can you indent these examples

tests/integration/cli/test_cli.py
samet-akcay: yes please

for_developers/entry_point_examples/test_config.yaml
```diff
@@ -0,0 +1,210 @@
+data_root: ../../tests/assets/classification_dataset
+engine:
+  task: MULTI_CLASS_CLS
```
samet-akcay, line 3: When `task` is defined in multiple places like engine and data, what is the behaviour

samet-akcay (changes requested):
@harimkang, thanks a lot for your effort! Great work! I've added some questions and comments here. 

I'll do another round and have a closer look at engine.py

src/otx/cli/cli.py
ashwinvaidya17: Ah makes sense

src/otx/engine/engine.py
```diff
@@ -0,0 +1,486 @@
...
+    >>> engine = Engine(
+        data_root="dataset/path",
+        task="MULTI_CLASS_CLS",
+    )
+    """
+
+    def __init__(
+        self,
+        *,
+        data_root: str | Path,
```
djdameln, line 88: Could this argument be made optional? If the user passes a datamodule object as the `datamodule` argument, we technically don't need `data_root` right?

src/otx/engine/engine.py
```diff
@@ -0,0 +1,486 @@
...
+        task: OTXTaskType | None = None,
+        work_dir: str | Path = "./otx-workspace",
+        datamodule: OTXDataModule | None = None,
```
djdameln, line 91: Is there a specific reason why we pass the model and datamodule to the constructor instead of the entrypoint methods (`fit`, `test`, `predict` etc.)? I'm asking because it deviates from the Lightning trainer functionality, where the model and datamodule are always passed to the entrypoint methods.

I guess it makes sense to use the constructor if an engine instance is always coupled to a just a single model and datamodule, but is this always the case? Or could there be a scenario where we need to use multiple models or datamodules with the same engine instance?

src/otx/engine/engine.py
```diff
@@ -0,0 +1,486 @@
...
+                otx train
+                    --data_root <DATASET_PATH>
+                    --max_epochs <EPOCHS, int> --checkpoint <CKPT_PATH, str>
+                ```
+            4. If you have a complete configuration file, run it like this.
+                ```python
+                otx train --data_root <DATASET_PATH> --config <CONFIG_PATH, str>
+                ```
+        """
+        lit_module = self._build_lightning_module(
```
djdameln, line 209: If I understand it correctly, we always pass a torch (non-lightning) model to the engine and obtain the lightning module by wrapping the torch model in a subclass of `OTXLitModule`, depending on the task type. 

I'm wondering if this would provide sufficient flexibility to implement different model behaviours within a certain task type. Let's take anomaly models for example. We have different classes of anomaly models that each require a unique implementation of the training step and the test step (e.g. feature extraction based models vs. student-teacher models). I'm having some trouble seeing how we could implement these model-specific behaviours if we wrap all anomaly models in the same `OTXLitModule` subclass.

src/otx/engine/engine.py
harimkang: This is something to think about a bit more. For one thing, it can be confusing for users when models are overridden (when and where), and it's pretty hard to modify the num_classes of an already built model.
Instead we could consider a setter function, which probably already exists in engine.py in some form.
What do you think?

for_developers/entry_point_examples/test_config.yaml
harimkang: This is the file that was output via print_config, so the arguments correspond to those belonging to Engine.train. We can also set additional Trainer arguments at the same level as these.

src/otx/algo/classification/otx_dino_v2.py
harimkang: We can do it. But I think it's beyond the scope of my PR, I've tried to keep the model, the data core as untouched as possible.

src/otx/cli/cli.py
harimkang: The CLI always expects to pass a datamodule only when an Engine is created. This line was created to avoid unnecessary parser argument collection.
(So datamodule is not required for test, predict, it meant datamodule argument skip here)


The answer to the second question is yes.

src/otx/cli/cli.py
harimkang: They can be considered partially dependent. But the reason I've separated them here is because the optimizer and scheduler should be partially instantiated (the optimizer depends on model.params, and the scheduler depends on the optimizer). Do you have any recommendations for a different way to do this?

src/otx/core/config/engine.py
harimkang: Yes, but OTX will offer devices beyond those provided by Trainer in the future, so we need to manage this up front in Engine. I would expect the device arguments to be set up in Engine first.

src/otx/core/config/device.py
harimkang: Added device-related arguments to be managed by OTX Engine. For now, only the Trainer's devices are connected, but i believe that other things will need to be managed in the future, so i created this.

src/otx/core/config/engine.py
harimkang: Yes, I agree. It would be nice to have a PR to clean up the OTX prefixes all at once.

src/otx/core/config/engine.py
harimkang: I actually created this to check the variables in the configuration file, but as you say, I think it might not be necessary. I'll remove it.

src/otx/core/data/module.py
harimkang: As mentioned earlier, I tried to avoid touching the model and data core classes as much as possible.
The change is simply that the InstSegDataModuleConfig is not necessarily separate from the DataModuleConfig.

src/otx/core/data/module.py
harimkang: I feel it's necessary, but I'm not sure if that's the scope of this PR.

src/otx/core/model/module/base.py
harimkang: My understanding is that this is a compile-related argument from torch 2.0, but I don't know the details.

src/otx/engine/engine.py
harimkang: Yeah this is something I'm going to think about more, as is Dick's comment.

src/otx/engine/engine.py
harimkang: Yes, that's right. I must have missed it, I'll fix it, thanks.

src/otx/engine/engine.py
harimkang: Yes, for data we need to think about whether we can get it from other functions as well. (First of all, `test`, `predict` can get an additional datamodule or dataloader. If they don't have any input, they use the datamodule set at engine creation.)

However, if we think about Engine based on the current workflow in OTX, it was written based on the one-model workflow case, so I think the current use case for multiple models is either non-existent or very limited in OTX, which is why model is accepted as an argument to `_init_`.

src/otx/engine/engine.py
harimkang: I understand your example. I'll think more about the case where there are multiple LitModules per Task. Thanks for the nice comment.

harimkang: Can we collect Anomalib's LitModules into a single factory class called OTXAnomalibLitModule so that we can provide them for OTXModel?

What do you think? In the current OTX 2.0, there is currently only one LitModule class per Task.

src/otx/engine/engine.py
harimkang: Done - 
Indent: https://github.com/openvinotoolkit/training_extensions/pull/2778/commits/e0d8de27fd90611e5a01194877149f25afe7300d
More Example: https://github.com/openvinotoolkit/training_extensions/pull/2778/commits/463e922766165c0b94e7f0a2a717ca4a5827d472

src/otx/engine/engine.py
harimkang: Thanks, Done - https://github.com/openvinotoolkit/training_extensions/pull/2778/commits/e0d8de27fd90611e5a01194877149f25afe7300d

harimkang:
Hi all, Thanks for reviewing PR!
I've gotten feedback that the size of the PR is too big from @vinnamkim (I also very much agree.), so I'm going to split it up and create the PRs. First of all, I created a PR for Engine first. - https://github.com/openvinotoolkit/training_extensions/pull/2778

The next PR will include Auto-Configuration, CLI, along with refactoring the Configuration files.

@samet-akcay @ashwinvaidya17 @djdameln 
I've incorporated as many of your comments as I can regarding Engine, so could you please review it once more?

Also, I will keep the comments from the current PR and refer to them in future work.

PR #2774, opened 2024-01-10
Add export entry point

sovrasov:
### Summary

Done:
 - Abstract exporter and an exporter which uses native tools: OVC + torch.onnx
 - Export for classification and segmentation
 - Basic accuracy tests
 
Problems:
- Correct label names are not loaded to the model

### Checklist

### License

```python
```


src/otx/core/model/module/base.py
```diff
@@ -170,3 +175,19 @@ def meta_info(self) -> DataMetaInfo:
     @meta_info.setter
     def meta_info(self, meta_info: DataMetaInfo) -> None:
         self._meta_info = meta_info
+
+    def export(self, output_dir: str) ->  str:
```
eunwoosh: IMHO, it's better for `OTXModel` inherited class than `OTXLitModule`, such as `MMPretrainMulticlassClsModel`, to be in charge of export,
because `OTXLitModule` is a just wrapper class for torch-lightning trainer.
How do you think about it?

src/otx/core/config/export.py
```diff
@@ -0,0 +1,25 @@
...
+    format: OTXExportFormatType
+    precision: OTXExportPrecisionType
+
+    input_height: int
+    input_width: int
+    mean: Tuple[float, float, float]
+    std: Tuple[float, float, float]
+    resize_mode: str
+    pad_value: int
+    swap_rgb: bool
```
eunwoosh: Class may need to have additional attribute for deployment config for `mmdeploy`.

src/otx/core/config/export.py
```diff
@@ -0,0 +1,25 @@
...
+"""Config data type objects for export method."""
+
+from dataclasses import dataclass
+from typing import Tuple
+
+from otx.core.types.export import OTXExportPrecisionType, OTXExportFormatType
+
+
+@dataclass
+class ExportConfig:
```
eunwoosh, line 13: In my thought, it's better to set this class at same level as `TrainConfig` and pass it to `export(cfg)` in "src/otx/core/engine/export.py" (may need to add other attribute such as `model` for that). How do you think about?

eunwoosh (commented):
Thanks for your work! I know it's still draft but I left some of my opinions. Please take a look.

src/otx/core/model/module/base.py
kprokofi: we will have both export methods in LitModule and OTXModel. It is needed to not call model.model.export(), but just model.export()

src/otx/core/config/export.py
sovrasov: Ok, let's add extra fields when required

src/otx/core/model/module/base.py
sovrasov: The initial idea is that `OTXLitModule` contains a wrapper for reference export implementation in `OTXModel`, which can be customized down the hierarchy. Mmdeploy export would be implemented where it's required

src/otx/engine/engine.py
```diff
@@ -323,9 +323,42 @@ def predict(
...
+            # self.model.label_info = self.datamodule.meta_info this doesn't work for some models yet
+            lit_module = self._build_lightning_module(
+                model=self.model,
+                optimizer=self.optimizer,
+                scheduler=self.scheduler,
+            )
+            lit_module.meta_info = self.datamodule.meta_info
+
+            loaded_checkpoint = torch.load(ckpt_path)
+            lit_module.load_state_dict(loaded_checkpoint["state_dict"])
```
sungmanc:
```suggestion
            lit_module = self._build_lightning_module(
                model=self.model,
                optimizer=self.optimizer,
                scheduler=self.scheduler,
            )
            
            loaded_checkpoint = torch.load(ckpt_path)
            
            # We need to use the information from the checkpoint when we want to export
            # In this phase, Datamodule doesn't have any real information. It also needs to be fixed.
            # At the export phase, we don't need to any input for the data.
            
            lit_module.meta_info = loaded_checkpoint["meta_info"]
            lit_module.load_state_dict(loaded_checkpoint["state_dict"])
```

sovrasov: Unfortunately, `loaded_checkpoint` doesn't contain `meta_info` key. Also that wouldn't solve problem of missing `model.label_info`

src/otx/engine/engine.py
```diff
@@ -323,9 +323,42 @@ def predict(
             return_predictions=return_predictions,
         )
 
-    def export(self, *args, **kwargs) -> None:
-        """Export the trained model to OpenVINO Intermediate Representation (IR) or ONNX formats."""
-        raise NotImplementedError
+    def export(self, checkpoint: str | Path | None = None, export_config: ExportConfig | None = None, **kwargs) -> Path:
```
eunwoosh: How about using `export_format` and `precision` as argument instead of `ExportConfig`? If then, user can set argument in CLI.

src/otx/engine/engine.py
eunwoosh: It's in loaded_checkpoint ['state_dict']['meta_info'].

tests/integration/test_export_inference.py
```diff
@@ -0,0 +1,253 @@
...
+
+    out, _ = capfd.readouterr()
+    assert TASK_NAME_TO_MAIN_METRIC_NAME[task] in out
+    torch_acc, ov_acc = tuple(re.findall(rf"{TASK_NAME_TO_MAIN_METRIC_NAME[task]}\s*│\s*(\d+[.]\d+)", out))
+    torch_acc, ov_acc = float(torch_acc), float(ov_acc)
+
+    msg = f"Recipe: {recipe}, (torch_accuracy, ov_accuracy): {torch_acc} , {ov_acc}"
+    log.info(msg)
+
+    _check_relative_metric_diff(torch_acc, ov_acc, 0.1)
```
eunwoosh, line 204: If torch_acc is almost 0 (including 0), and ov_acc is not 0, it mostly fails. How about adding criteria to deal whit this case? For example, if torch_acc - ov_acc < 0.01, then consider it ok.

sovrasov: In general, the fact that `torch_acc = 0` is not ok. We should not reflect that workaround in criteria, which is fine for normal cases. I observe an improvement of training accuracy after deterministic is enabled, could you please check once again? If that wouldn't work, we can apply simplified criterion for the faulty models, or increase the number of training epochs for them.

src/otx/engine/engine.py
sovrasov: Thanks, that field exists, but label_info of the model is still outdated even if we run `lit_module.meta_info = loaded_checkpoint['state_dict']["meta_info"]`

tests/integration/cli/test_cli.py
```diff
@@ -76,8 +76,10 @@ def test_otx_e2e(recipe: str, tmp_path: Path, fxt_accelerator: str) -> None:
     """
     Test OTX CLI e2e commands.
 
-    - 'otx train' with 2 epochs trainig
+    - 'otx train' with 2 epochs training
     - 'otx test' with output checkpoint from 'otx train'
+    - 'otx export' with output checkpoint from 'otx train'
+    - 'otx test' with the exported model
```
GalyaZalesskaya:
```suggestion
    - 'otx test' with the exported to ONNX/IR model
```

tests/integration/test_export_inference.py
```diff
@@ -0,0 +1,255 @@
...
+    fxt_accelerator: str,
+    capfd: "pytest.CaptureFixture",
+) -> None:
+    """
+    Test OTX CLI e2e commands.
+
+    - 'otx train' with 2 epochs training
+    - 'otx test' with output checkpoint from 'otx train'
+    - 'otx export' with output checkpoint from 'otx train'
+    - 'otx test' with the exported model
```
GalyaZalesskaya:
```suggestion
    - 'otx test' with the exported model to ONNX/IR model
```

src/otx/core/config/export.py
```diff
@@ -0,0 +1,17 @@
...
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+from otx.core.types.export import OTXExportFormatType, OTXExportPrecisionType
+
+
+@dataclass
+class ExportConfig:
+    """DTO for export configuration."""
```
GalyaZalesskaya:
```suggestion
    """Data Transfer Object (DTO) for export configuration."""
```

src/otx/core/config/model.py
```diff
@@ -0,0 +1,20 @@
...
+"""Config data type objects for model."""
+
+from dataclasses import dataclass
+
+from otx.core.config.export import ExportConfig
+
+
+@dataclass
+class ModelConfig:
+    """DTO for model configuration."""
```
GalyaZalesskaya:
```suggestion
    """Data Transfer Object (DTO) for model configuration."""
```

GalyaZalesskaya (commented):
Thank you for a huge update. Just a minor comment-related changes

src/otx/core/config/model.py
```diff
@@ -0,0 +1,20 @@
...
+#
+"""Config data type objects for model."""
+
+from dataclasses import dataclass
+
+from otx.core.config.export import ExportConfig
+
+
+@dataclass
+class ModelConfig:
```
kprokofi: We don't have ModelConfig now. I think it can be deleted

src/otx/core/exporter/native.py
```diff
@@ -0,0 +1,142 @@
...
+import torch
+
+from otx.core.exporter.base import OTXModelExporter
+from otx.core.types.export import OTXExportPrecisionType
+
+
+class OTXNativeModelExporter(OTXModelExporter):
+    """Exporter that uses native torch and OpenVINO conversion tools."""
+
+    def __init__(
```
kprokofi, line 23: I would rather move this constructor to the base class. For now it seems strange that base class doesn't have signature of the constructor.

src/otx/core/exporter/native.py
```diff
@@ -0,0 +1,142 @@
...
+        self,
+        input_size: tuple[int, ...],
```
kprokofi, line 25:
```suggestion
        input_size: tuple[int, int],
```
I think input size always integer

src/otx/core/model/entity/base.py
```diff
@@ -36,6 +37,8 @@ class OTXModel(nn.Module, Generic[T_OTXBatchDataEntity, T_OTXBatchPredEntity, T_
         num_classes: Number of classes this model can predict.
     """
 
+    _EXPORTED_MODEL_BASE_NAME = "exported_model"
```
kprokofi: I would rather make it as parameter in export method to have possibility to modify this parameter when calling export method.

src/otx/core/model/entity/classification.py
```diff
@@ -84,6 +87,16 @@ class OTXMulticlassClsModel(
 ):
     """Base class for the classification models used in OTX."""
 
+    def _generate_model_metadata(
+        self,
+    ) -> dict[tuple[str, str], Any]:
+        metadata = super()._generate_model_metadata()
+        metadata[("model_info", "model_type")] = "Classification"
```
kprokofi: Is there a special reason to store keys in metadata as tuples?

src/otx/core/model/entity/classification.py
```diff
@@ -485,6 +584,8 @@ def __init__(
         use_throughput_mode: bool = True,
         model_api_configuration: dict[str, Any] | None = None,
     ) -> None:
+        model_api_configuration = model_api_configuration if model_api_configuration else {}
```
kprokofi, line 599: Could you please make the same update for OVHlabelClassificationModel? I forgot that we create model in constructor

src/otx/engine/engine.py
kprokofi: Agree with Eunwoo, we can then delete ExportConfig DTO

src/otx/core/model/entity/classification.py
kprokofi: It is not convenient afterwards work with metadata

src/otx/core/exporter/native.py
eunwoosh: Maybe it can be done by my PR. I already wrote common part of both Exporters into base class.

src/otx/core/exporter/native.py
```diff
@@ -0,0 +1,142 @@
...
+                example_input=dummy_tensor,
+                input=(openvino.runtime.PartialShape(self.input_size),),
+            )
+
+        # workaround for OVC's bug: single output doesn't have a name in OV model
+        if len(exported_model.outputs) == 1 and len(exported_model.outputs[0].get_names()) == 0:
+            exported_model.outputs[0].tensor.set_names({"output1"})
+
+        metadata = {} if metadata is None else self._extend_model_metadata(metadata)
+        exported_model = OTXNativeModelExporter._embed_openvino_ir_metadata(exported_model, metadata)
```
eunwoosh:
```suggestion
        if metadata is not None:
            self._extend_model_metadata(metadata)
            exported_model = OTXNativeModelExporter._embed_openvino_ir_metadata(exported_model, metadata)
```
How about changing code like this to reduce function call?

src/otx/core/exporter/native.py
```diff
@@ -0,0 +1,142 @@
...
+        precision: OTXExportPrecisionType = OTXExportPrecisionType.FP32,
+        metadata: dict[tuple[str, str], str] | None = None,
+    ) -> Path:
+        """Export to ONNX format.
+
+        In this implementation the export is done only via standard OV/ONNX tools.
+        """
+        dummy_tensor = torch.rand(self.input_size).to(next(model.parameters()).device)
+        save_path = str(output_dir / (base_model_name + ".onnx"))
+        metadata = {} if metadata is None else self._extend_model_metadata(metadata)
```
eunwoosh: Save as above.

src/otx/core/model/entity/base.py
```diff
@@ -262,14 +291,21 @@ def _create_model(self) -> Model:
...
-        images = [np.transpose(im.numpy(), (1, 2, 0)) for im in entity.images]
+        images = [np.transpose(im.cpu().numpy(), (1, 2, 0)) for im in entity.images]
         return {"inputs": images}
 
+    def _customize_outputs(
+        self,
+        outputs: Any,  # noqa: ANN401
+        inputs: T_OTXBatchDataEntity,
+    ) -> T_OTXBatchPredEntity | OTXBatchLossEntity:
+        """Customize OTX output batch data entity if needed for you model."""
```
eunwoosh:
```suggestion
        """Customize OTX output batch data entity if needed for model."""
```
is it typo?

src/otx/algo/segmentation/dino_v2_seg.py
```diff
@@ -14,3 +16,12 @@ def __init__(self, num_classes: int) -> None:
...
         config = read_mmconfig(model_name=model_name)
         super().__init__(num_classes=num_classes, config=config)
+
+    def export(self, *args) -> Path:
+        """Export method for DinoV2Seg.
+
+        Model doesn't support export for now due to unsupported operations from xformers.
+        This method will raise an error.
+        """
+        msg = "{model_name} cannot be exported. It is not supported."
```
eunwoosh:
```suggestion
        msg = "DinoV2Seg cannot be exported. It is not supported."
```

src/otx/algo/segmentation/heads/custom_ham_head.py
```diff
@@ -0,0 +1,56 @@
...
+
+@MODELS.register_module()
+class CustomLightHamHead(LightHamHead):
+    """SegNeXt decode head.
+
+    It is modified LightHamHead from mmsegmentation.
+
+    Args:
+        ham_channels (int): input channels for Hamburger.
+            Defaults: 512.
```
eunwoosh:
```suggestion
```

src/otx/engine/engine.py
```diff
@@ -323,9 +323,41 @@ def predict(
...
```
```suggestion
    def export(self, checkpoint: str | Path | None = None, export_config: ExportConfig | None = None) -> Path:
```
How about removing `kwargs` in argument to reduce user confusion?

tests/integration/test_export_inference.py
```diff
@@ -0,0 +1,255 @@
...
+        *DATASET[task]["overrides"],
+        "--checkpoint",
+        str(ckpt_files[-1]),
+    ]
+
+    with patch("sys.argv", command_cfg):
+        main()
+
+    # 3) otx export
+    format_to_ext = {"OPENVINO": "xml"}  # [TODO](@Vlad): extend to "ONNX": "onnx"
```
eunwoosh, line 140: It's just a question. Are there any reasons not to test ONNX now?

tests/integration/test_export_inference.py
eunwoosh: I agree with you but I have a concern that it goes beyond integration test if it requires proper training. How about adding a test to compare scores in e2e and regression test when we start to add that rather than now?

src/otx/core/model/module/base.py
```diff
@@ -178,8 +175,8 @@ def load_state_dict(self, ckpt: dict[str, Any], *args, **kwargs) -> None:
                 f"Data classes from training data: {self.meta_info.label_names}",
             )
             self.register_load_state_dict_pre_hook(
-                self.meta_info.label_names,
-                ckpt_meta_info.class_names,
+                self.meta_info.class_names,
```
eunwoosh:
```suggestion
                self.meta_info.label_names,
```
As far as I know, both `self.meta_info` and `ckpt_meta_info` is `LabelInfo` class. shouldn't both of them access to `label_names`?

eunwoosh (commented):
Thanks for your work! I have some comments. Please take a look.

src/otx/core/model/entity/base.py
```diff
@@ -170,41 +173,67 @@ def map_class_names(src_classes: list[str], dst_classes: list[str]) -> list[int]
...
+        for lbl in self.label_info.label_names:
+            all_labels += lbl.replace(" ", "_") + " "
+            all_label_ids += lbl.replace(" ", "_") + " "
+
+        return {
+            ("model_info", "labels"): all_labels.strip(),
+            ("model_info", "label_ids"): all_label_ids.strip(),
+        }
+
+    def _export_to_exportable_code(self) -> Path:
```
vinnamkim: Why don't we move this to `exporter` as well.
```suggestion
    exporter.to_exportable_code(...)
```

src/otx/core/model/entity/base.py
```diff
@@ -170,41 +173,67 @@ def map_class_names(src_classes: list[str], dst_classes: list[str]) -> list[int]
...
+        metadata = self._generate_model_metadata()
 
-    def _export_to_openvino(self, output_dir: Path) -> None:
-        """Export to OpenVINO Intermediate Representation format.
+        if export_format == OTXExportFormatType.OPENVINO:
+            return exporter.to_openvino(self.model, output_dir, self._EXPORTED_MODEL_BASE_NAME, precision, metadata)
+        if export_format == OTXExportFormatType.ONNX:
+            return exporter.to_onnx(self.model, output_dir, self._EXPORTED_MODEL_BASE_NAME, precision, metadata)
+        if export_format == OTXExportFormatType.EXPORTABLE_CODE:
+            return self._export_to_exportable_code()
```
vinnamkim: Why don't we hide this branching into `exporter` as well?
```suggestion
        return exporter.export(exgport_format, ...)
```

src/otx/core/model/entity/segmentation.py
```diff
@@ -38,7 +59,9 @@ class MMSegCompatibleModel(OTXSegmentationModel):
     def __init__(self, num_classes: int, config: DictConfig) -> None:
         config = inplace_num_classes(cfg=config, num_classes=num_classes)
         self.config = config
+        self.export_params = _get_export_params_from_seg_mmconfig(config)
```
vinnamkim, line 68: Is this same with `_get_export_params_from_cls_mmconfig()`? Can we move it to a common place?

src/otx/core/model/entity/base.py
```diff
@@ -170,41 +173,67 @@ def map_class_names(src_classes: list[str], dst_classes: list[str]) -> list[int]
...
+
+    def _create_exporter(
+        self,
+    ) -> OTXModelExporter:
+        """Creates OTXModelExporter object that can export the model."""
         raise NotImplementedError
 
-    def _export_to_onnx(self) -> None:
-        """Export to ONNX format.
+    def _generate_model_metadata(
```
vinnamkim: I'm not sure why `metadata` exists apart from the export parameters, thus we have two sources (`_configure_export_parameters()` and `_generate_model_metadata()`) just for the one `export` feature. Can we bring them into one?

src/otx/core/model/entity/base.py
```diff
@@ -170,41 +173,67 @@ def map_class_names(src_classes: list[str], dst_classes: list[str]) -> list[int]
...
+        Returns:
+            Path: path to the exported model.
         """
-        if export_format == OTXExportFormat.OPENVINO:
-            self._export_to_openvino(output_dir)
-        if export_format == OTXExportFormat.ONNX:
-            self._export_to_onnx()
-        if export_format == OTXExportFormat.EXPORTABLE_CODE:
-            self._export_to_exportable_code()
+        exporter = self._create_exporter()
```
vinnamkim: I think that you decided to create a `exporter` component to grant all responsibilities for the model exportation. I think it is good and will help us to enhance the bad cohesion about everything masses up in the base class. As things goes like this, I propose you to clean up everything in the base class and redesign it.

We will leave two abstract property in the base class such as
```python
class OTXModel:
    @property
    def exporter(self):
          msg = (
                "To export this OTXModel, you should implement an appropriate exporter for it. "
                "You can try to reuse our provided `Exporter` in `otx.core.exporter.*`."
          )
          raise NotImplementError(msg)

    @property
    def export_parameters(self) -> dict[str, Any]:
          msg = (
                "To export this OTXModel, you should define an appropriate parameters. "
                "This is used in the constructor of `self.exporter`. "
                "For example, `self.exporter = SomeExporter(**self.export_parameters)`. "
                "For the detail of constructor arguments, please see `otx.core.exporter.*`."
          )
          raise NotImplementError(msg)
```

Both are mandatory requirements to enable the model exportation that should be derived in the `OTXModel` children. With this design, we might maintain the common `export()` function in the base class such as
```python
    def export(
        self,
        output_dir: Path,
        export_format: OTXExportFormatType,
        precision: OTXExportPrecisionType = OTXExportPrecisionType.FP32,
    ) -> Path:
        return self.exporter.export(output_dir, export_format, precision)
```

vinnamkim (changes requested):

vinnamkim (changes requested):
Thanks but please see https://github.com/openvinotoolkit/training_extensions/pull/2774#discussion_r1467188184 for my suggestion to reduce the bad cohesion in the base class.

src/otx/core/types/export.py
```diff
@@ -1,16 +1,23 @@
...
+class OTXExportFormatType(str, Enum):
+    """OTX export format type definition."""
 
-    OPENVINO = "OPENVINO"
     ONNX = "ONNX"
+    OPENVINO = "OPENVINO"
     EXPORTABLE_CODE = "EXPORTABLE_CODE"
+
+
+class OTXExportPrecisionType(str, Enum):
```
harimkang: What about bundling Precision separately so that it can be managed in a common way regardless of export? 
What do you think?

src/otx/core/config/model.py
harimkang: I agree.

src/otx/engine/engine.py
harimkang: I agree. Since there are only two (`export_format` & `precision`), it seems like a good idea to just use them explicitly here.

src/otx/engine/engine.py
```diff
@@ -323,9 +323,41 @@ def predict(
...
+        Returns:
+            Path: Path to the exported model.
+        """
+        ckpt_path = str(checkpoint) if checkpoint is not None else self.checkpoint
+        if export_config is None:
+            export_config = ExportConfig()
+
+        if ckpt_path is not None:
+            self.model.eval()
+            # self.model.label_info = self.datamodule.meta_info this doesn't work for some models yet
```
harimkang: Do we need this (use `self.datamodule.meta_info`) in the export?

src/otx/engine/engine.py
```diff
@@ -323,9 +323,41 @@ def predict(
...
+            lit_module = self._build_lightning_module(
```
harimkang, line 388: Is there a reason to build a `LightningModule`, because `load_state_dict` is hard to do with just an `OTXModel` here?

src/otx/engine/engine.py
```diff
@@ -323,9 +323,41 @@ def predict(
...
+        """Export the trained model to OpenVINO Intermediate Representation (IR) or ONNX formats.
+
+        Args:
+            checkpoint (str | Path | None, optional): Checkpoint to export. Defaults to None.
+            export_config (ExportConfig | None, optional): Config that allows to set export
+            format and precision. Defaults to None.
+
+        Returns:
+            Path: Path to the exported model.
+        """
```
harimkang, line 383: If `otx export` also needs a user guide in CLI, I can add it in a future PR (like `otx train --help`. User quick guide with verbose 0 level)

src/otx/core/model/entity/base.py
kprokofi: It is done in my [PR](https://github.com/openvinotoolkit/training_extensions/pull/2824) with exportable code. It will be opened after this one merged.

src/otx/core/model/entity/base.py
kprokofi: Same answer

tests/integration/test_export_inference.py
sovrasov: If we don't compare scores here at all, export and inference most likely fell apart after several PRs. Status of regression tests is almost always unknown (from the previous experience), also, they are designed to be run weekly, while we need a way to perform sanity check in ad-hoc manner.  
If train + inference fails to pass a sanity check, that looks like a problem. As I said, we can turn of this tests for the models that show 0 training accuracy.

Each case when OTX 2.0 fails to pass training sanity check should be addressed @sungmanc @vinnamkim @harimkang please share also your opinions.

src/otx/core/model/entity/classification.py
kprokofi: We discussed it offline, I resolve conversation

src/otx/core/exporter/native.py
sovrasov: `...` here denotes any number of ints

src/otx/core/exporter/native.py
sovrasov: `OTXModelExporter` serves rather as an interface, I didn't suppose it has any attributes.
Perhaps, constructor can be created for code sharing purposes

src/otx/core/exporter/native.py
sovrasov: Please try
```python
def statement(x):
    print("exec")
    return x

b = None
a = 1 if b is None else statement(True)
```

tests/integration/test_export_inference.py
sovrasov: Because onnx inference is not fully supported yet. Ability to infer to an onnx file is checked in cli e2e tests

src/otx/engine/engine.py
sovrasov: We need to find a way to populate self.model.label_info somehow. For now export happens with garbage label names. For me it doesn't matter how label_info comes to the model. Apparently, we don't need `self.datamodule.meta_info` here, but I don't know yet what do we need instead.

src/otx/core/model/entity/base.py
sovrasov: Yes

src/otx/core/model/module/base.py
sovrasov: That looks like a merge artifact, I'll revert this

src/otx/core/model/entity/base.py
sovrasov: Yes, Kirill has already done that in his PR which would be merged after this one

src/otx/core/model/entity/segmentation.py
sovrasov: Eunwoo already did that in his mmdeploy PR, I'd leave it as it is to avoid merge conflicts

src/otx/engine/engine.py
sovrasov: Done

src/otx/engine/engine.py
sovrasov: Exactly, I need to do some keys matching magic right here if I use model directly

src/otx/engine/engine.py
sovrasov: I updated the comment, see several lines below

src/otx/engine/engine.py
sovrasov: Added

src/otx/core/types/export.py
sovrasov: Extracted to a standalone file & renamed

src/otx/core/model/entity/base.py
sovrasov: Added to model.export()

src/otx/core/model/entity/base.py
sovrasov: In fact, now export method is common and it's not supposed to be maintained. Also, abstract methods similar to the mentioned properties are already implemented. 
You're proposing to rewrite this with fancy syntaxis sugar, right?

src/otx/core/model/entity/base.py
sovrasov: I extracted this to higher level, since every model has to menage it's metadata, and metadata tends to be extended when moving down the hierarchy. `_configure_export_parameters` methods are defined only for mmodels for convenience. The generic way to provide parameters for exporter now is to overload _create_exporter() method. I don't want to bring this low-level and weakly structured stuff higher in the hierarchy.

So abstract `OTX[TaskName]Model` knows only about `_generate_model_metadata` and `_create_exporter`, and then every particular child class manages the other details.
In brief, 
metadata: structured, has a clear hierarchical structure, managed from the top of the hierarchy.
export parameters -> a murky dict of anything, created at low levels if needed. Now required only because of mm models.

tests/integration/test_export_inference.py
eunwoosh: yes but don't we need to check whether model can be exported to ONNX?

src/otx/core/model/entity/base.py
vinnamkim: It's not just for the syntactic sugar. The real problem is https://github.com/openvinotoolkit/training_extensions/pull/2774#discussion_r1467185043. I saw your comment (https://github.com/openvinotoolkit/training_extensions/pull/2774#discussion_r1467741226), but I still don't know the point why they are split to the two sources although both are only used for the model exportation. 

> In brief,
metadata: structured, has a clear hierarchical structure, managed from the top of the hierarchy.
export parameters -> a murky dict of anything, created at low levels if needed. Now required only because of mm models.

Even if we make them as a single one, we still can have a hierarchical structure. Before you are telling about this in details, I could not get what the rationale behind this although I'm very familiar with this codebase. Then, how about the others? I still think that it is natural that there is one source for the model exportation configuration. Lastly, I think this variable (`input_size`) in `export_params` is very model specific and not just for mm*, so that I don't think we should generalize it:
  https://github.com/openvinotoolkit/training_extensions/blob/74be80b044bf5e95778e46d94fbb1bbaa36c5120/src/otx/core/model/entity/classification.py#L218

src/otx/engine/engine.py
harimkang: It's a bit weird as it is, we may need to find a way to fix this, label_info matching may be the answer here as well.

tests/integration/test_export_inference.py
sovrasov: It's done in `tests/integration/cli/test_cli.py`

src/otx/core/model/entity/base.py
sovrasov: The example with input size you provided is valid only for classification mm model, which is almost the lowest level in the hierarchy, so it may not be generalizable. I implemented your proposal, (since we have no time for discussion), although I don't see how parameters became more clear to add. They are mixed with metadata now in an unstructured dict, which is not great, as I said.

vinnamkim: No, I meant that the input size is specific not just to the model architecture, furthermore to the model template (otx/recipe in our codebase). For example, you can imagine that there are two templates that train resnet50 on 320x320 or 224x224 sized images. Considering this aspect, it does not seem appropriate to raise the input size to the hierarchy.

vinnamkim (dismissed):

tests/integration/test_export_inference.py
eunwoosh: I agree with your opinion. Maybe we can find a way to check it's sanity within reasonable time. Let's look in more detail later.

eunwoosh (dismissed):

src/otx/core/exporter/base.py
```diff
@@ -0,0 +1,115 @@
...
+        """
+
+    @abstractmethod
+    def to_onnx(
+        self,
+        model: torch.nn.Module,
+        output_dir: Path,
+        base_model_name: str = "exported_model",
+        precision: OTXPrecisionType = OTXPrecisionType.FP32,
+        embed_metadata: bool = True,
```
eunwoosh, line 81: Is error raised if meta data is embedded when exporting ONNX?

eunwoosh (dismissed):
LGTM. I have a minor question. Please take a look :)

eunwoosh (approved):

vinnamkim (approved):

src/otx/core/exporter/base.py
sovrasov: No, the error occurs in OV if we converted an ONNX model with metadata via OVC, and then try to execute via OV runtime

kprokofi (approved):

PR #2775, opened 2024-01-11
Pre-work for the design change: create own python class for every model

vinnamkim:
### Summary

- This is the pre-work, so that we will actually introduce own python class for every model in the next PR.
- Enrich `OTXModel` to have `export()`, `register_explain_hook()`, and `label_info` property.
- `label_info` will be automatically constructed by the input integer argument, `num_classes`.

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/core/data/dataset/base.py
+++ b/src/otx/core/data/dataset/base.py
@@ -30,15 +30,27 @@ Transforms = Union[Compose, Callable, List[Callable]]
 
 
 @dataclass
-class DataMetaInfo:
-    """Meta information of each subset datasets."""
+class LabelInfo:
+    """Object to represent label information."""
 
-    class_names: list[str]
+    label_names: list[str]
 
     @property
     def num_classes(self) -> int:
-        """Return number of classes."""
-        return len(self.class_names)
+        """Return number of labels."""
+        return len(self.label_names)
+
+    @classmethod
+    def from_num_classes(cls, num_classes: int) -> LabelInfo:
+        """Create this object from the number of classes.
+
+        Args:
+            num_classes: Number of classes
+
+        Returns:
+            LabelInfo(label_names=["label_0", ...])
+        """
+        return LabelInfo(label_names=[f"label_{idx}" for idx in range(num_classes)])
 
 
 class OTXDataset(Dataset, Generic[T_OTXDataEntity]):
@@ -59,8 +71,8 @@ class OTXDataset(Dataset, Generic[T_OTXDataEntity]):
         self.mem_cache_img_max_size = mem_cache_img_max_size
         self.max_refetch = max_refetch
 
-        self.meta_info = DataMetaInfo(
-            class_names=[category.name for category in self.dm_subset.categories()[AnnotationType.label]],
+        self.meta_info = LabelInfo(
+            label_names=[category.name for category in self.dm_subset.categories()[AnnotationType.label]],
         )
 
     def __len__(self) -> int:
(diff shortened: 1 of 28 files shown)


MarkByun (approved):
It looks good.

src/otx/core/data/dataset/base.py
```diff
@@ -30,15 +30,27 @@
 
 
 @dataclass
-class DataMetaInfo:
-    """Meta information of each subset datasets."""
+class LabelInfo:
+    """Object to represent label information."""
```
jaegukhyun, line 34: What if LitModule requires other info such as data_root, subset names? I'm not sure whether there are such use cases, but if there are, should we create DatasetMetaInfo class which have LabelInfo attribute?

src/otx/core/utils/config.py
```diff
@@ -66,3 +67,29 @@ def mmconfig_dict_to_dict(obj: MMConfig | list[MMConfig]) -> list | dict:
         return {k: mmconfig_dict_to_dict(v) for k, v in obj.to_dict().items()}
 
     return obj
+
+
+def inplace_num_classes(
```
jaegukhyun, line 72: It seems this function is duplicated with https://github.com/openvinotoolkit/training_extensions/blob/0e8bd95b41e899c01ec2fee2a6258d24f6f08598/src/otx/core/utils/build.py#L82
Can we choose one of them?

src/otx/core/model/entity/base.py
```diff
@@ -5,29 +5,66 @@
...
+
+        if old_num_classes != new_num_classes:
+            msg = (
+                f"Given LabelInfo has the different number of classes "
+                f"({old_num_classes}!={new_num_classes}). "
+                "The model prediction layer is reset to the new number of classes "
+                f"(={new_num_classes})."
+            )
+            warnings.warn(msg, stacklevel=0)
+            self._reset_prediction_layer(num_classes=label_info.num_classes)
```
jaegukhyun, line 64: It seems all of model entity do not implement their own _reset_prediciton_layer function. 
Do we have a plan for this?

sungmanc (approved):
Approved, BTW, could you please explain why do we need to insert the `num_classes` to the `OTXModel` at this PR ? I'm not opposite to inject this argument, however, just curiosity. It seems that inserting the `num_classes` have no much relation with own python class.

src/otx/core/utils/config.py
vinnamkim: Please deprecate it after merging it.
https://github.com/openvinotoolkit/training_extensions/blob/0e8bd95b41e899c01ec2fee2a6258d24f6f08598/src/otx/core/utils/build.py#L89
This should be `isinstance(value, ListConfig)`.

src/otx/core/data/dataset/base.py
vinnamkim: We can rename and expand it if that case happens. Honestly, we already have the expanded one, `HLabelInfo`.

src/otx/core/model/entity/base.py
vinnamkim: Sure we might do in the future

jaegukhyun (approved):

vinnamkim:
> Approved, BTW, could you please explain why do we need to insert the `num_classes` to the `OTXModel` at this PR ? I'm not opposite to inject this argument, however, just curiosity. It seems that inserting the `num_classes` have no much relation with own python class.

We introduce `LabelInfo` as a common property of `OTXModel`. However, this property affects the shape of the model itself. If we decide it lazily, not at the construction, we should implement this `_reset_prediction_layer()` as well this time. In addition, lazily resetting the layer seems not good in terms of the clean flow and computation cost.

PR #2776, opened 2024-01-11
Create own python class for every model for action cls, det, and classification tasks

vinnamkim:
### Summary

- Same as title. Other tasks will be resolved in the following PR.

### Checklist

### License

```python
```


src/otx/algo/action_classification/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright (C) 2023 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
+"""Module for OTX action classification models, hooks, utils, etc."""
```
samet-akcay: is hooks still used in this version?

src/otx/algo/action_classification/x3d.py
```diff
@@ -0,0 +1,15 @@
...
+"""X3D model implementation."""
+
+from otx.algo.utils.mmconfig import read_mmconfig
+from otx.core.model.entity.action_classification import MMActionCompatibleModel
+
+
+class X3D(MMActionCompatibleModel):
+    """X3D Model."""
+
+    def __init__(self, num_classes: int) -> None:
```
samet-akcay, line 13: what if a user wants to use this class via API? With this approach it would only be possible to keep an additional mmconfig. Ideally, this class should be fully configurable without the need for an external config file. After passing the args from the signature, this class could generate the mmconfig under the hood and instantiate the mm model.

src/otx/algo/action_detection/x3d_fastrcnn.py
```diff
@@ -0,0 +1,17 @@
...
+from __future__ import annotations
+
+from otx.algo.utils.mmconfig import read_mmconfig
+from otx.core.model.entity.action_detection import MMActionCompatibleModel
+
+
+class X3DFastRCNN(MMActionCompatibleModel):
+    """X3D Model."""
+
+    def __init__(self, num_classes: int, topk: int | tuple[int]):
```
samet-akcay, line 14: here `topk` is defined as an argument, which is different than the previous X3D signature. does the difference stem from the task type, or something else?

src/otx/algo/utils/mmconfig.py
```diff
@@ -0,0 +1,31 @@
...
+
+    if module is None or (mod_fpath := module.__file__) is None:
+        msg = "Cannot get valid model from stack"
+        raise RuntimeError(msg)
+
+    root_dir = Path(mod_fpath).parent / "mmconfigs" / subdir_name
+    yaml_fpath = root_dir / f"{model_name}.yaml"
+
+    if not yaml_fpath.exists():
+        raise FileNotFoundError
```
samet-akcay: can you add a message to specify which file is missing. It would be easier to debug this way.

src/otx/algo/utils/mmconfig.py
```diff
@@ -0,0 +1,31 @@
...
+    """Read MMConfig.
+
+    It try to read MMConfig from the yaml file which exists in
+    `<Directory path of __file__ who calls this function>/mmconfigs/<subdir_name>/<model_name>.yaml`
+    """
+    frame = inspect.stack()[1]
+    module = inspect.getmodule(frame[0])
+
+    if module is None or (mod_fpath := module.__file__) is None:
+        msg = "Cannot get valid model from stack"
```
samet-akcay: the error message is a bit too generic. can you add a more descriptive message so the user can understand what could be wrong.

src/otx/algo/utils/mmconfig.py
```diff
@@ -0,0 +1,31 @@
...
+def read_mmconfig(model_name: str, subdir_name: str = ".") -> DictConfig:
+    """Read MMConfig.
+
+    It try to read MMConfig from the yaml file which exists in
+    `<Directory path of __file__ who calls this function>/mmconfigs/<subdir_name>/<model_name>.yaml`
+    """
+    frame = inspect.stack()[1]
+    module = inspect.getmodule(frame[0])
+
+    if module is None or (mod_fpath := module.__file__) is None:
```
samet-akcay:
```suggestion
    if module is None or (module_file_path := module.__file__) is None:
```

samet-akcay (changes requested):
thanks for the effort!

Overall, it is a good step towards having an api-level definition for the models. However, this is still heavily rely on an external configuration file, which is hard to modify from API.

In addition to this, I've got some minor comments.

As we move on, I think we should make this more API friendly by avoiding config dependency

src/otx/algo/action_classification/__init__.py
vinnamkim: Fixed in https://github.com/openvinotoolkit/training_extensions/pull/2776/commits/669d935209ac14f857b52b25ce3957fc9ce03f15

src/otx/algo/utils/mmconfig.py
vinnamkim: Fixed in https://github.com/openvinotoolkit/training_extensions/pull/2776/commits/669d935209ac14f857b52b25ce3957fc9ce03f15

src/otx/algo/utils/mmconfig.py
vinnamkim: Fixed in https://github.com/openvinotoolkit/training_extensions/pull/2776/commits/669d935209ac14f857b52b25ce3957fc9ce03f15

src/otx/algo/utils/mmconfig.py
vinnamkim: Fixed in https://github.com/openvinotoolkit/training_extensions/pull/2776/commits/669d935209ac14f857b52b25ce3957fc9ce03f15

src/otx/algo/action_classification/x3d.py
vinnamkim: That will be done by each execution member who is responsible for each OTX task implementation.

src/otx/algo/action_detection/x3d_fastrcnn.py
vinnamkim: This PR aims to introduce a dedicated Python class for each OTX model. Therefore, the efforts have been made to apply minimal changes whenever possible. `topk` is a configurable parameter in the integration tests, so I added it here to ensure compatibility with the integration test. @jaegukhyun, please see this and clean up the input arguments of this model if possible

src/otx/algo/action_detection/x3d_fastrcnn.py
```diff
@@ -0,0 +1,17 @@
...
+# SPDX-License-Identifier: Apache-2.0
+#
+"""X3DFastRCNN model implementation."""
+from __future__ import annotations
+
+from otx.algo.utils.mmconfig import read_mmconfig
+from otx.core.model.entity.action_detection import MMActionCompatibleModel
+
+
+class X3DFastRCNN(MMActionCompatibleModel):
```
samet-akcay, line 11: by the way, is this fastrcnn or faster rcnn?

jaegukhyun: It is fastrcnn, it uses pre-computed proposals.

samet-akcay: thanks for the clarification

src/otx/algo/action_detection/x3d_fastrcnn.py
jaegukhyun: In this structure, topk can be configured by num_classes, I'll update it after merging this PR.

src/otx/config/model/mmaction_detection.yaml
```diff
@@ -5,7 +5,7 @@ _target_: otx.core.model.module.action_detection.OTXActionDetLitModule
 
 otx_model:
   _target_: otx.core.model.entity.action_detection.MMActionCompatibleModel
-  config: ???
+  num_classes: ???
```
jaegukhyun, line 8: Should TopK be added here?

jaegukhyun (approved):

src/otx/config/model/mmaction_detection.yaml
vinnamkim: If every MMAction model should configure `top_k`, then please add it there.

src/otx/algo/action_classification/x3d.py
sovrasov: @samet-akcay could you elaborate a bit what does "fully configurable" mean? The parameters located in mmconfig are quite specific, and it looks unrealistic to provide user control over all of them in a lightweight fashion. Perhaps, we can add a mechanism to override values items in the mmconfig via model constructor, but anyway that would require from a user quite intensive research.

sungmanc (approved):
Thanks for the change. LGTM

harimkang (approved):

vinnamkim:
@samet-akcay 
Do you have any concern more?

src/otx/algo/action_classification/x3d.py
samet-akcay: @sovrasov, the following is ideally what I want to have in otx. It is from torchvision library [here](https://github.com/pytorch/vision/blob/main/torchvision/models/detection/ssd.py#L126-L202). I don't need an additional config file to instantiate it. Especially when working with an API. If I have a config file, I could still instantiate this class, but both of these approaches are completely independent. 

I'm aware that it may not be easy with the mmlab ecosystem, that requires this config stuff. but ideally we should aim for the api definition, which is similar to the top frameworks such torchvision, ligthning, keras and others. 

```python
class SSD(nn.Module):
    """
    Implements SSD architecture from `"SSD: Single Shot MultiBox Detector" <https://arxiv.org/abs/1512.02325>`_.

    The input to the model is expected to be a list of tensors, each of shape [C, H, W], one for each
    image, and should be in 0-1 range. Different images can have different sizes, but they will be resized
    to a fixed size before passing it to the backbone.

    The behavior of the model changes depending on if it is in training or evaluation mode.

    During training, the model expects both the input tensors and targets (list of dictionary),
    containing:
        - boxes (``FloatTensor[N, 4]``): the ground-truth boxes in ``[x1, y1, x2, y2]`` format, with
          ``0 <= x1 < x2 <= W`` and ``0 <= y1 < y2 <= H``.
        - labels (Int64Tensor[N]): the class label for each ground-truth box

    The model returns a Dict[Tensor] during training, containing the classification and regression
    losses.

    During inference, the model requires only the input tensors, and returns the post-processed
    predictions as a List[Dict[Tensor]], one for each input image. The fields of the Dict are as
    follows, where ``N`` is the number of detections:

        - boxes (``FloatTensor[N, 4]``): the predicted boxes in ``[x1, y1, x2, y2]`` format, with
          ``0 <= x1 < x2 <= W`` and ``0 <= y1 < y2 <= H``.
        - labels (Int64Tensor[N]): the predicted labels for each detection
        - scores (Tensor[N]): the scores for each detection

    Args:
        backbone (nn.Module): the network used to compute the features for the model.
            It should contain an out_channels attribute with the list of the output channels of
            each feature map. The backbone should return a single Tensor or an OrderedDict[Tensor].
        anchor_generator (DefaultBoxGenerator): module that generates the default boxes for a
            set of feature maps.
        size (Tuple[int, int]): the width and height to which images will be rescaled before feeding them
            to the backbone.
        num_classes (int): number of output classes of the model (including the background).
        image_mean (Tuple[float, float, float]): mean values used for input normalization.
            They are generally the mean values of the dataset on which the backbone has been trained
            on
        image_std (Tuple[float, float, float]): std values used for input normalization.
            They are generally the std values of the dataset on which the backbone has been trained on
        head (nn.Module, optional): Module run on top of the backbone features. Defaults to a module containing
            a classification and regression module.
        score_thresh (float): Score threshold used for postprocessing the detections.
        nms_thresh (float): NMS threshold used for postprocessing the detections.
        detections_per_img (int): Number of best detections to keep after NMS.
        iou_thresh (float): minimum IoU between the anchor and the GT box so that they can be
            considered as positive during training.
        topk_candidates (int): Number of best detections to keep before NMS.
        positive_fraction (float): a number between 0 and 1 which indicates the proportion of positive
            proposals used during the training of the classification head. It is used to estimate the negative to
            positive ratio.
    """

    __annotations__ = {
        "box_coder": det_utils.BoxCoder,
        "proposal_matcher": det_utils.Matcher,
    }

    def __init__(
        self,
        backbone: nn.Module,
        anchor_generator: DefaultBoxGenerator,
        size: Tuple[int, int],
        num_classes: int,
        image_mean: Optional[List[float]] = None,
        image_std: Optional[List[float]] = None,
        head: Optional[nn.Module] = None,
        score_thresh: float = 0.01,
        nms_thresh: float = 0.45,
        detections_per_img: int = 200,
        iou_thresh: float = 0.5,
        topk_candidates: int = 400,
        positive_fraction: float = 0.25,
        **kwargs: Any,
    ):
```

samet-akcay:
> @samet-akcay Do you have any concern more?

The remaining comments are for future reference. 

Sorry, forgot to approve this one :)

samet-akcay (approved):

src/otx/algo/action_classification/x3d.py
sovrasov: I also was thinking about somethin like that, but then we need to carefully curate the list of parameters for each model and think of usecases. For instance, if user has no real chance to change the `anchor_generator` we might not want to expose one. Anyway, this looks like an ideal state, which we have to pursue.

PR #2778, opened 2024-01-12
Add Engine class for OTX Entry Point

harimkang:
### Summary

I split this PR (https://github.com/openvinotoolkit/training_extensions/pull/2772)
So the first thing I did was to separate the work for the Engine Class into PR.

- Add Engine class for Entry Point

The next PR will be include
- changes to the configuration file and structure
- Autoconfiguration
- CLI

### Checklist

### License

```python
```


harimkang:
We're having issues with our CI machine today, so CI testing is very slow (Queue), but I check locally first and open a PR.
and I'm going to write a unit-test while I get the review.

src/otx/cli/__init__.py
```diff
@@ -46,12 +46,16 @@ def setup_subcommands(self) -> None:
...
+        # otx train parser
+        train_parser = ArgumentParser()
+        train_parser.add_argument("overrides", help="overrides values", default=[], nargs="+")
+        parser_subcommands.add_subcommand("train", train_parser, help="Training subcommand for OTX")
 
-        add_test_parser(parser_subcommands)
+        # otx test parser
+        test_parser = ArgumentParser()
+        test_parser.add_argument("overrides", help="overrides values", default=[], nargs="+")
+        parser_subcommands.add_subcommand("test", test_parser, help="Testing subcommand for OTX")
```
samet-akcay, line 58: are these parser only available for train and test? not needed for other entrypoints?

src/otx/cli/test.py
```diff
@@ -6,30 +6,12 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
-
+import hydra
```
samet-akcay, line 9: is this temporary or permanent?

src/otx/core/config/device.py
```diff
@@ -0,0 +1,18 @@
...
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+from otx.core.types.device import DeviceType
+
+
+@dataclass
+class DeviceConfig:
```
samet-akcay, line 14: can you provide an exemplary use-case when `DeviceConfig` would be used?

src/otx/engine/engine.py
```diff
@@ -0,0 +1,407 @@
...
+
+        Args:
+            data_root (str | Path): The root directory of the data.
+            task (OTXTaskType | None, optional): The task type. Defaults to None.
+            work_dir (str | Path, optional): The working directory. Defaults to "./otx-workspace".
+            datamodule (OTXDataModule | None, optional): The data module. Defaults to None.
+            model (OTXModel | str | None, optional): The model. Defaults to None.
+            optimizer (OptimizerCallable | None, optional): The optimizer. Defaults to None.
+            scheduler (LRSchedulerCallable | None, optional): The learning rate scheduler. Defaults to None.
+            checkpoint (str | None, optional): The checkpoint. Defaults to None.
```
samet-akcay: Can you elaborate this. Path to checkpoint? it is not crystal clear from the first glance

src/otx/engine/engine.py
```diff
@@ -0,0 +1,407 @@
...
+        self,
+        *,
+        data_root: str | Path | None = None,
+        task: OTXTaskType | None = None,
+        work_dir: str | Path = "./otx-workspace",
+        datamodule: OTXDataModule | None = None,
+        model: OTXModel | str | None = None,
+        optimizer: OptimizerCallable | None = None,
+        scheduler: LRSchedulerCallable | None = None,
+        checkpoint: str | None = None,
```
samet-akcay, line 80: I think it would be ideal to specify the metrics here as well. it would make the flow more logical.

src/otx/cli/__init__.py
samet-akcay: Or I assume it is because this belongs to the previous CLI version?

src/otx/engine/engine.py
```diff
@@ -0,0 +1,407 @@
...
+        device: DeviceType = DeviceType.auto,
```
samet-akcay, line 81: lightning has `accelerator` and `devices` arguments to set the device type and number of devices. What if a user passed that instead of using `device` args here? It might be confusing to specify another device related arg here. 

To make it simpler from a user perspective, would it be possible to just use these lightning args?

src/otx/engine/engine.py
```diff
@@ -0,0 +1,407 @@
...
+        >>> engine = Engine(
+            data_root=<dataset/path>,
+            model=OTXModel(...),
+            checkpoint=<checkpoint/path>,
+        )
+        # With Custom OTXDataModule
+        >>> engine = Engine(
+            model = OTXModel(...),
+            datamodule = OTXDataModule(...),
+        )
```
samet-akcay: When there are multiple examples, the section name becomes Examples in sphinx. We also need to use `...` when we break the python code multiple lines.

```suggestion
    Examples:
    The following examples show how to use the Engine class.

        .. code-block:: python
            :caption: Auto-Configuration with data_root

            engine = Engine(data_root=<dataset/path>)

        .. code-block:: python
            :caption: With Custom OTXModel

            engine = Engine(
               data_root=<dataset/path>,
               model=OTXModel(...),
               checkpoint=<checkpoint/path>,
            )

        .. code-block:: python
            :caption: With Custom OTXDataModule

            engine = Engine(
               model = OTXModel(...),
               datamodule = OTXDataModule(...),
            )
```

src/otx/engine/engine.py
```diff
@@ -0,0 +1,407 @@
...
+    OTXTaskType.SEMANTIC_SEGMENTATION: "otx.core.model.module.segmentation.OTXSegmentationLitModule",
+    OTXTaskType.ACTION_CLASSIFICATION: "otx.core.model.module.action_classification.OTXActionClsLitModule",
+    OTXTaskType.ACTION_DETECTION: "otx.core.model.module.action_detection.OTXActionDetLitModule",
+}
+
+
+class Engine:
+    """OTX Engine.
+
+    This class defines the common interface for OTX, including methods for training and testing.
```
samet-akcay:
```suggestion
    This class defines the Engine for OTX, which governs each step of the OTX workflow.
```

src/otx/engine/engine.py
samet-akcay: or if you think that code block is too long, you could replace it with 

```python
            >>> engine = Engine(
            ...  model = OTXModel(...),
            ...   datamodule = OTXDataModule(...),
            ... )
```
we need to add `...` when multiple lines exist

src/otx/engine/engine.py
```diff
@@ -0,0 +1,407 @@
...
+        self,
+        max_epochs: int = 10,
+        seed: int | None = None,
+        deterministic: bool = False,
+        precision: _PRECISION_INPUT | None = "32",
+        val_check_interval: int | float | None = 1,
+        callbacks: list[Callback] | Callback | None = None,
+        logger: Logger | Iterable[Logger] | bool | None = None,
+        resume: bool = False,
+        **kwargs,
```
samet-akcay, line 135: so basically, is this the following idea?

When the user wants to change some of the trainer args after instantiating the Trainer/Engine, they will be able to do it here. For example,

```python
from otx.data import Datamodule
from otx.models import Dinov2
from otx.engine import Engine

data = Datamodule(..)
model = DinoV2(...)
engine = (data, model, max_epochs=2)

# Train it for 2 epochs.
engine.train()

# I now want to train for 4 more epochs
engine.train(max_epochs=4)
```

Overall, defining the lightning trainer args in entrypoints gives us more flexibility to tweak the parameters later on. That is the idea, right?

samet-akcay: If yes, that is good. However, we also need to think about the other way around. This approach does the following:
- Let the users use trainer args in both class constructor and in entrypoints. 

However, it does not do the other way around. We could also let them use entrypoint args in both class constructor and entrypoints. For example, in Lightning, we could do this;

```python 
datamodule = Datamodule(..)
model = Model(..)

trainer = Trainer(max_epochs=3, ...)
trainer.fit(datamodule=datamodule, model=model, ...)
```

this pr does not do that. We have to specify datamodule and model in the `Engine.__init__`. We could add this flexibility to make the best of both worlds. For example, with OTX Engine I should be able to do the following:

```python
from otx.data import Datamodule
from otx.models import Dinov2
from otx.engine import Engine

data = Datamodule(..)
model = DinoV2(...)
engine = (data, model, max_epochs=2)

# Train it for 2 epochs.
engine.train()
```

this should also be possible 
```python
from otx.data import Datamodule
from otx.models import Dinov2
from otx.engine import Engine

data = Datamodule(..)
model = DinoV2(...)
engine = ()

# Train it for 2 epochs.
engine.train(data, model, max_epochs=2)
```

Alternatively, let's say we define the model and datamodule in engine constructor. However, we might have a new dataset or dataloader or datamodule in the future to train/val or test. We should be able to do that. Like,

```python
data = Datamodule(..)
model = DinoV2(...)
engine = (data, model, max_epochs=2)

# Train it for 2 epochs.
engine.train()
engine.test()

# now use a new dataset to test 
new_test_dataset = Dataset(..)
engine.test(dataset)
```

To do that, we would need to also include the original lighting entrypoint args in `Engine` entrypoints. So, both trainer args + entrypoint args to get the best of both. 

Thoughts?

src/otx/engine/engine.py
```diff
@@ -0,0 +1,407 @@
...
+            seed_everything(seed, workers=True)
+
+        self._build_trainer(
+            logger=logger,
+            callbacks=callbacks,
+            precision=precision,
+            max_epochs=max_epochs,
+            deterministic=deterministic,
+            val_check_interval=val_check_interval,
+            **kwargs,
```
samet-akcay, line 200: would there be any need to check the kwargs to ensure that always correct args are being passed. Or does lightning already check it? Or we don't need it at all if we get a clear error message in case wrong arg is passed?

src/otx/engine/engine.py
```diff
@@ -0,0 +1,407 @@
...
+        )
+        fit_kwargs: dict[str, Any] = {}
+        if resume:
+            fit_kwargs["ckpt_path"] = self.checkpoint
```
samet-akcay, line 204: why not simply use `ckpt_path` instead of `checkpoint`? We already have `ckpt_path` as one of `kwargs` in the engine constructor, right?

src/otx/engine/engine.py
```diff
@@ -0,0 +1,407 @@
...
+            **fit_kwargs,
+        )
+        self.checkpoint = self.trainer.checkpoint_callback.best_model_path
+
+        return self.trainer.callback_metrics
+
+    def test(
+        self,
+        checkpoint: str | Path | None = None,
+        datamodule: EVAL_DATALOADERS | OTXDataModule | None = None,
```
samet-akcay, line 221: does this mean that `datamodule` could be a dataloader as well? Would it be an idea to follow the lightning signature here to make it less confusing?

```python
        self,
        model: AnomalyModule | None = None,
        dataloaders: EVAL_DATALOADERS | AnomalibDataModule | None = None,
        ckpt_path: str | None = None,
        verbose: bool = True,
        datamodule: OTXDataModule | None = None,
        **kwargs
```

src/otx/engine/engine.py
```diff
@@ -0,0 +1,407 @@
...
+            **kwargs: Additional keyword arguments for pl.Trainer configuration.
+
+        Returns:
+            dict: Dictionary containing the callback metrics from the trainer.
+
+        Example:
+            >>> engine.test(
+                datamodule=OTXDataModule(),
+                checkpoint="checkpoint.ckpt",
+            )
```
samet-akcay:
```suggestion
            >>> engine.test(
            ...    datamodule=OTXDataModule(),
            ...    checkpoint="checkpoint.ckpt",
            ... )
```

src/otx/engine/engine.py
```diff
@@ -0,0 +1,407 @@
...
+            dataloaders=datamodule,
+            ckpt_path=str(checkpoint) if checkpoint is not None else self.checkpoint,
+        )
+
+        return self.trainer.callback_metrics
+
+    def predict(
+        self,
+        checkpoint: str | Path | None = None,
+        datamodule: EVAL_DATALOADERS | OTXDataModule | None = None,
```
samet-akcay, line 275: again, would it be confusing to the user that `datamodule` could also be a dataloader?

src/otx/engine/engine.py
```diff
@@ -0,0 +1,407 @@
...
+        return_predictions: bool | None = None,
```
samet-akcay, line 276: also how about using the same signature as lightning as follows
```python
    def predict(
        self,
        model: Optional["pl.LightningModule"] = None,
        dataloaders: Optional[Union[EVAL_DATALOADERS, LightningDataModule]] = None,
        datamodule: Optional[OTXDataModule] = None,
        return_predictions: Optional[bool] = None,
        ckpt_path: Optional[str] = None,
```

so the model is optional. If we don't pass anything, we could use the one that is already saved in Engine.
We could also use another model to do a prediction?
Also, we could either use a dataloader or a datamodule ?

src/otx/engine/engine.py
```diff
@@ -0,0 +1,407 @@
...
+
+        Returns:
+            list | None: The predictions if `return_predictions` is True, otherwise None.
+
+        Example:
+            >>> engine.predict(
+                datamodule=OTXDataModule(),
+                checkpoint="checkpoint.ckpt",
+                return_predictions=True,
+            )
```
samet-akcay:
```suggestion
            >>> engine.predict(
            ...    datamodule=OTXDataModule(),
            ...    checkpoint="checkpoint.ckpt",
            ...    return_predictions=True,
            ... )
```

src/otx/engine/engine.py
```diff
@@ -0,0 +1,407 @@
...
+            return_predictions=return_predictions,
+        )
+
+    def export(self, *args, **kwargs) -> None:
+        """Export the trained model to OpenVINO Intermediate Representation (IR) or ONNX formats."""
+        raise NotImplementedError
+
+    """
+    Property and setter functions provided by Engine.
+    """
```
samet-akcay: Not sure if it is common to use triple quotes there. Let's stay on the safe side and stick to the conventional format :)

```suggestion
    # ------------------------------------------------------------------------ #
    # Property and setter functions provided by Engine.                        #
    # ------------------------------------------------------------------------ #
```

src/otx/engine/engine.py
```diff
@@ -0,0 +1,407 @@
...
+
+    def __init__(
+        self,
+        *,
+        data_root: str | Path | None = None,
+        task: OTXTaskType | None = None,
+        work_dir: str | Path = "./otx-workspace",
+        datamodule: OTXDataModule | None = None,
+        model: OTXModel | str | None = None,
```
samet-akcay, line 77: `OTXModel` is a `Torch` model, right? So are we always passing torch model to the Engine instead of lightning model? What if I create a LightningModel and pass it to the engine?

src/otx/engine/engine.py
```diff
@@ -0,0 +1,407 @@
...
+            raise NotImplementedError(msg)
+        self.datamodule: OTXDataModule = datamodule
+        self.task = self.datamodule.task
+
+        self._trainer: Trainer | None = None
+        self._model: OTXModel = model
+        self.optimizer: OptimizerCallable = optimizer
+        self.scheduler: LRSchedulerCallable = scheduler
+
+    """
```
samet-akcay:
```suggestion
    # ------------------------------------------------------------------------ #
    # General OTX Entry Points                                                 #
    # ------------------------------------------------------------------------ #
```

samet-akcay (changes requested):
Thanks @harimkang, it's shaping up! I've got some comments and questions. 

My main questions are mainly around improving the flexibility you are already proposing with your args approach. Let's discuss it on Monday.

src/otx/engine/engine.py
harimkang: `Engine.test`, `predict` have already configured their arguments to receive a new dataloader. I think the point we should consider here is model. I thought the current OTX, apart from lightning's trainer, is that the engine provides training to exporting (+@) for one model. In this case, wouldn't it be more confusing for the user if we made them receive model for each entry point? 
Of course, users who are familiar with lightning may be more familiar with receiving it from functions such as `fit` and `test`, but I think it may be a bit confusing to put model in when it is created and also receive model in functions such as train for entry OTX user.

We'll have to think about it more, but for now I'm thinking **1 Engine = 1 model**.
Get `model` args from `train()` may allow us to receive the model flexibly, as you intend, but we have to think about a more complicated overriding code flow than we might think (if the user keeps changing models, which model should we serve in some cases and which in others?)

Each entry point is configured to not override Engine's model and datamodule.
Instead, we provide a setter function for model in case the user wants to change the model.
```python
engine.model = NewOTXModel(...)
engine.train()
```

What do you think?

src/otx/engine/engine.py
harimkang: These kwargs will be tried in the same form as Trainer(**kwargs). This error message is expected to appear upon Trainer creation. So I think this is causing an error in lightning.Trainer, do you think there's a better way to do it? (raise error in Engine side?)

src/otx/engine/engine.py
harimkang: The `ckpt_path` is independent of the kwargs.
it is not an argument to the Trainer constructor, but to a function like `fit`.

src/otx/engine/engine.py
harimkang: In `Engine.test`, we use a combination of `dataloaders` and `datamodule` from `Trainer.test`. The reason i used `datamodule` as a word in the argument is because we use the same `datamodule` in the argument of the Engine constructor. 

for someone who doesn't know Trainer as well as I do, it's confusing to have dataloaders and datamodules in one place, and they're even used almost interchangeably inside `Trainer.test`. So I've combined them.

src/otx/engine/engine.py
harimkang: This `datamodule` actually primarily corresponds to the `datamodule` argument in `Trainer.fit`. However, `Trainer.fit` can accept `train_dataloader` and `val_dataloader` separately. Currently, OTX v2 doesn't have the ability to do this separately, so it doesn't receive a dataloader here for now (test, predict can receive a dataloader).

src/otx/engine/engine.py
harimkang: Users can use `dataloaders` or `datamodule` here.
However, model only uses model that depend on engine. As with the other entry points, the engine has one model, and each entry point uses that model.

src/otx/cli/__init__.py
harimkang: This PR includes Engine-specific changes to the existing CLI. (There are no major changes to the CLI.) Expect new CLI classes and configuration changes in next PR.

src/otx/cli/test.py
harimkang: I believe this is temporary and will be removed in the next PR.

src/otx/engine/engine.py
harimkang: The way I think about it is that OTX should automatically set up the device via Engine.
The Engine would allow the user to see an argument called device, and by specifying it, the Engine would later help set up the accelerator and devices for each device.

Actually, as you say, it's possible to use the Trainer's `accelerator` and `devices`, but for a non-lightning user like me, it's more confusing to see `accelerator`, `devices`. (Not only is it hard to tell the difference between the two, but the type is confusing to me, accelerator -> str & devices -> str or int)

I think this is a shortcoming of Trainer, so I think it's a good argument to make Engine more approachable than Trainer, what do you think?

src/otx/core/config/device.py
harimkang: My motivation was to collect the device settings in a `DeviceConfig` and make them communicate in API.
My first thought is that the user can specify the device via Engine, and by default OTX should support the auto method: (`default` = 'auto')
The basic goal is to support intel-device, hpu, etc, and currently we only support accelerators in Trainer, but I think we should be able to extend that. Also, multi-gpu should be controllable via Engine, and I think the ability to have it set automatically based on the environment will be added later.

So currently it's device input (str) -> DeviceConfig (always 1 gpu) -> Trainer(accelerator=DeviceConfig.accelerator, devices=DeviceConfig.devices)
In the future, I think we should be able to find the setting value for our environment and set it through that. 
```
if user environment -> arc gpu, 3 gpu
if device=auto -> accelerator=arc, devices=3
```

What do you think? We can certainly communicate in ways other than `DeviceConfig`. But my current implementation uses a `dataclass` called `DeviceConfig`.

src/otx/engine/engine.py
harimkang: Thanks! Done with 2nd way - https://github.com/openvinotoolkit/training_extensions/pull/2778/commits/367d9bc4dcbb549d9c065e7286abf587d1a36675

src/otx/engine/engine.py
harimkang: Thanks! Done - https://github.com/openvinotoolkit/training_extensions/pull/2778/commits/367d9bc4dcbb549d9c065e7286abf587d1a36675

src/otx/engine/engine.py
harimkang: Thanks! Done - https://github.com/openvinotoolkit/training_extensions/pull/2778/commits/367d9bc4dcbb549d9c065e7286abf587d1a36675

src/otx/engine/engine.py
harimkang: Thanks! Done - https://github.com/openvinotoolkit/training_extensions/pull/2778/commits/367d9bc4dcbb549d9c065e7286abf587d1a36675

src/otx/engine/engine.py
harimkang: Thanks! Done - https://github.com/openvinotoolkit/training_extensions/pull/2778/commits/367d9bc4dcbb549d9c065e7286abf587d1a36675

src/otx/engine/engine.py
harimkang: Thanks! Done - https://github.com/openvinotoolkit/training_extensions/pull/2778/commits/367d9bc4dcbb549d9c065e7286abf587d1a36675

src/otx/engine/engine.py
harimkang: Maybe what you mean is want a longer explanation? I've changed it to Path to checkpoint for now, but can you take a look and see if it needs a clearer explanation?
https://github.com/openvinotoolkit/training_extensions/pull/2778/commits/367d9bc4dcbb549d9c065e7286abf587d1a36675

src/otx/engine/engine.py
harimkang: As I understand it, the current metric is held by the LightningModule for each task, so I don't know the proper way to change it, so this is something that needs to be discussed.

MarkByun (approved):

src/otx/engine/engine.py
harimkang: LightningModule is not available with the current engine implementation.
Should we make it possible to receive LightningModules?

harimkang: We could make Engine.model have `OTXModel | OTXLitModule`, as shown below.

```python
def __init__(
    ...
    model: OTXModel | OTXLitModule | str | None = None,
    ...
):
    if isinstance(model, OTXLitModule):
        ??? -> optimizer, scheduler update?
        self._model = model
    ...

def train():
    lit_module = self._build_lightning_module()
    lit_module.meta_info = self.datamodule.meta_info

    ...

def _build_lightning_module(
    self,
) -> OTXLitModule:
    if isinstance(self.model, OTXLitModule):
       ??? -> optimizer, scheduler update?
       return self.model
    class_module, class_name = LITMODULE_PER_TASK[self.task].rsplit(".", 1)
    module = __import__(class_module, fromlist=[class_name])
    lightning_module = getattr(module, class_name)
    return lightning_module(
        otx_model=model,
        optimizer=optimizer,
        scheduler=scheduler,
        torch_compile=False,
    )

```
I'm actually confused by ??? part in example code (`_init_` or `_build_lightning_module`). We have a ??? part, should we swap optimizer and scheduler? Or should we use the existing ones in LitModule?
What do you think?

jaegukhyun (approved):

src/otx/engine/engine.py
samet-akcay: yes, that was what I was asking :) The description was not clear :)

src/otx/engine/engine.py
samet-akcay: Yeah, I found it out in #2788, but not sure if this is a good idea. It is good to have a default metric for each task. However, it is not good to hardcode the metrics for each task. A user might just want to try another metric for the task. For example, it is `Accuracy` for the classification task, but I might just want to check the `F1Score` as well.

samet-akcay: Since the base module is designed this way, maybe we could leave it for another PR, but I believe we should discuss this. Can you add this to your agenda to discuss please?

harimkang: Sure. I'd like to discuss that agenda in the Forum tomorrow rather than in the current PR, and hopefully it can be addressed in another PR in the future.

src/otx/engine/engine.py
samet-akcay: I think we could swap the default optimiser and schduler in LitModel if the user explicitly passes them from the Engine. Otherwise, we could use the default ones that is defined in the lightning module.

@ashwinvaidya17? Any thoughts?

src/otx/engine/engine.py
samet-akcay: not sure :) maybe we could try and break it to see what sort of error message we see. The point is if a user unintentionally passes a wrong arg, the error message should ideally be informative, so they could easily find out where the problem would be

src/otx/engine/engine.py
harimkang: The worry here is that the Engine will always have a default optimizer and scheduler through Auto-Configuration, so we need to think about the flow well if we introduce `LitModules` into engine's model.

samet-akcay: As discussed, this otx design assumes that each task would have a dedicated lightning module. However, this may not be in the case in the future. Especially when the migration from mmx is complete.

For example, anomalib has a dedicated lightning module for each model instead of task. This is because, each model might have different behaviour within the lightning hooks. Therefore, anomalib models would not work with this approach

ashwinvaidya17: Maybe we can pass the torch model as a backbone to the lightning module. Something like this
```python
model = OTXMulticlassClsLitModule(backbone=OTXMobilenet,...)
engine = Engine(model=model, ...)
```

src/otx/cli/test.py
```diff
@@ -47,6 +29,27 @@ def otx_test(overrides: list[str]) -> None:
...
+        trainer_kwargs = {**cfg.trainer}
+        engine = Engine(
+            task=cfg.base.task,
+            work_dir=cfg.base.output_dir,
+            model=model,
+            optimizer=optimizer,
+            scheduler=scheduler,
+            datamodule=datamodule,
+            checkpoint=cfg.checkpoint,
+            device=trainer_kwargs.pop("accelerator", "auto"),
```
sovrasov, line 53: Is there any plan how to avoid this workaround in the future?
I expect growing number of explicit configuration parsing and even editing when more features from 1.x would be restored in 2.0

harimkang: @sovrasov Hi, This cli file will be refactored. You can refer the future CLI in this PR.
https://github.com/openvinotoolkit/training_extensions/pull/2772

The current CLI is temporary code.

harimkang: I think your concerns can be addressed later via the CLI.

src/otx/core/types/device.py
```diff
@@ -0,0 +1,21 @@
...
+    """OTX Device type definition."""
+    # ("cpu", "gpu", "tpu", "ipu", "hpu", "mps", "auto")
+
+    auto = "auto"
+    gpu = "gpu"
+    cpu = "cpu"
+    tpu = "tpu"
+    ipu = "ipu"
+    hpu = "hpu"
+    mps = "mps"
```
sovrasov, line 21: How that list was generated, and while Intel devices (xpu) are missing?

src/otx/engine/engine.py
harimkang: I don't think I can say much because it is a work related to OTXLitModule. The current architecture has the entire model, not just the backbone.

src/otx/core/types/device.py
harimkang: This is a list of accelerators currently supported by lightning.Trainer.
Shouldn't we expand in the future?

src/otx/core/config/device.py
samet-akcay: alright, we could revisit later if needed. Thanks for the explanation

src/otx/engine/engine.py
samet-akcay: Alright, let's merge this as is, but we will need to modify this when anomalib is being implemented.

samet-akcay (approved):
Thanks @harimkang. We will have to modify the engine to accommodate lightning modules. Let's merge this as is and modify it when we need it

PR #2788, opened 2024-01-14
Hotfix for segmentation + raising an error when metric is None

kprokofi:
### Checklist

### License

```python
```

### DIFF
--- a/src/otx/core/model/entity/base.py
+++ b/src/otx/core/model/entity/base.py
@@ -264,18 +264,21 @@ class OVModel(OTXModel, Generic[T_OTXBatchDataEntity, T_OTXBatchPredEntity]):
     ) -> T_OTXBatchPredEntity | OTXBatchLossEntity:
         """Model forward function."""
 
-        def _callback(result: NamedTuple, user_data: list[NamedTuple]) -> None:
-            user_data.append(result)
+        def _callback(result: NamedTuple, idx: int) -> None:
+            outputs.append((idx, result))
 
         numpy_inputs = self._customize_inputs(inputs)["inputs"]
         if self.async_inference:
             outputs: list[Any] = []
             self.model.set_callback(_callback)
-            for im in numpy_inputs:
+            for idx, im in enumerate(numpy_inputs):
                 if not self.model.is_ready():
                     self.model.await_any()
-                self.model.infer_async(im, user_data=outputs)
+                self.model.infer_async(im, user_data=idx)
             self.model.await_all()
+            # sort outputs in ascending order and keep model output only
+            outputs.sort()
+            outputs = [out[1] for out in outputs]
         else:
             outputs = [self.model(im) for im in numpy_inputs]
 
--- a/src/otx/core/model/entity/base.py
+++ b/src/otx/core/model/entity/base.py
@@ -265,20 +265,18 @@ class OVModel(OTXModel, Generic[T_OTXBatchDataEntity, T_OTXBatchPredEntity]):
         """Model forward function."""
 
         def _callback(result: NamedTuple, idx: int) -> None:
-            outputs.append((idx, result))
+            output_dict[idx] = result
 
         numpy_inputs = self._customize_inputs(inputs)["inputs"]
         if self.async_inference:
-            outputs: list[Any] = []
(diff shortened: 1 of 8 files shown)


kprokofi:
Integration tests didn't catch this issue. Val metric was None for segmentation task since test metric was computed instead in validation step

src/otx/core/model/module/action_classification.py
```diff
@@ -49,6 +49,10 @@ def on_test_epoch_end(self) -> None:
 
     def _log_metrics(self, meter: Accuracy, key: str) -> None:
         results = meter.compute()
+        if results is None:
+            msg = f"{meter} has no data to compute metric or there is an error computing metric"
+            raise RuntimeError(msg)
```
vinnamkim, line 54: If this error raising is repeating over all tasks can we wrap it up into a common util function?

kprokofi: I am not sure whether it worth or not for 3 lines of code? We don't have proper file to put this common small func
image

kprokofi:
I also added fix for throughput mode in this PR.
Async requests can be executed in different order, we should remember id of requests and sort them accordingly.

src/otx/core/model/entity/base.py
```diff
@@ -264,18 +264,21 @@ def forward(
...
             self.model.set_callback(_callback)
-            for im in numpy_inputs:
+            for idx, im in enumerate(numpy_inputs):
                 if not self.model.is_ready():
                     self.model.await_any()
-                self.model.infer_async(im, user_data=outputs)
+                self.model.infer_async(im, user_data=idx)
             self.model.await_all()
+            # sort outputs in ascending order and keep model output only
+            outputs.sort()
```
sovrasov: I'd prefer to use a dict to avoid sorting

jaegukhyun (approved):

harimkang (approved):

src/otx/core/model/module/action_classification.py
sungmanc: How about adding this logic as a function of the `OTXLitModule` ? (i.e. `OTXLitModule.has_valid_output()` or `.has_nan_metric()`) Then, we can easily expand for all tasks.

vinnamkim: In addition, I investigated all the return types of metric functions:

1. `Accuracy`
image

2. `MeanAveragePrecision`
image

3. `JaccardIndex`
image

All of them is not `None`able. Are those what we really want?

kprokofi: If you don't have any data in metric, then .compute() return None
I just wanted to add raising an error for this case. However, I don't want  to create separate method or function in utils for this. I guess, we can check metric in integration tests rather than raising some errors to exclude such problems in the future

samet-akcay: I agree with @sungmanc. ideally `_log_metric` would be part of the model interface. i see that each task has separate metric type. not sure if this is a good idea, though. 

Users might just want to try many metrics for tasks. For example, I might just want to try f1 score for a classification task, rather than accuracy. To be on the safe side, we could annotate the metric as `Metric`, which would be the parent class anyway?

vinnamkim: > If you don't have any data in metric, then .compute() return None I just wanted to add raising an error for this case. However, I don't want to create separate method or function in utils for this. I guess, we can check metric in integration tests rather than raising some errors to exclude such problems in the future

I still believe that there is a chance of refactoring to simplify this. I'm really against of adding code lines indefinitely without carefully. If you don't want to do it in this PR. I'll do.

vinnamkim (approved):

src/otx/core/model/module/action_classification.py
kprokofi: I am not against refactoring this. I'm just thinking how to do it better. Creating separate method in LitModule with 2 lines of code to check metric is None looks strange for me

samet-akcay: there will be multiple round of refactoring stages at the end anyway. That's what I was mentioning this morning

PR #2789, opened 2024-01-15
Apply FP16 mixed precision for all tasks

jaegukhyun:
### Checklist

### License

```python
```

### DIFF
--- a/src/otx/recipe/action_detection/x3d_fastrcnn.yaml
+++ b/src/otx/recipe/action_detection/x3d_fastrcnn.yaml
@@ -1,5 +1,6 @@
 # @package _global_
 defaults:
+  - override /trainer: default
   - override /base: action_detection
   - override /callbacks: action_detection
   - override /data: mmaction_detection
@@ -124,3 +125,5 @@ model:
     lr: 0.005
     momentum: 0.9
     weight_decay: 0.00001
+trainer:
+  precision: 32
--- a/src/otx/recipe/detection/rtmdet_tiny.yaml
+++ b/src/otx/recipe/detection/rtmdet_tiny.yaml
@@ -1,5 +1,6 @@
 # @package _global_
 defaults:
+  - override /trainer: default
   - override /base: detection
   - override /callbacks: detection
   - override /data: mmdet
@@ -208,3 +209,5 @@ model:
         debug: false
         pos_weight: -1
       type: RTMDet
+trainer:
+  precision: 32
--- a/tests/integration/cli/test_cli.py
+++ b/tests/integration/cli/test_cli.py
@@ -68,7 +68,7 @@ DATASET = {
 
 
 @pytest.mark.parametrize("recipe", RECIPE_LIST)
-def test_otx_e2e(recipe: str, tmp_path: Path) -> None:
+def test_otx_e2e(recipe: str, tmp_path: Path, fxt_accelerator: str) -> None:
     """
     Test OTX CLI e2e commands.
 
@@ -119,7 +119,7 @@ def test_otx_e2e(recipe: str, tmp_path: Path) -> None:
         f"base.data_dir={DATASET[task]['data_dir']}",
         f"base.work_dir={tmp_path_test}",
         f"base.output_dir={tmp_path_test / 'outputs'}",
-        "trainer=gpu",
+        f"trainer={fxt_accelerator}",
         *DATASET[task]["overrides"],
         f"checkpoint={ckpt_files[-1]}",
     ]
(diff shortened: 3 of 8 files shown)


src/otx/recipe/detection/rtmdet_tiny.yaml
```diff
@@ -208,3 +209,5 @@ model:
         debug: false
         pos_weight: -1
       type: RTMDet
+trainer:
+  precision: 32
```
harimkang, line 213: Is this to align with OTX 1.5? Or is it because it's a model that doesn't work with mixed precision?

src/otx/recipe/action_detection/x3d_fastrcnn.yaml
```diff
@@ -124,3 +125,5 @@ model:
     lr: 0.005
     momentum: 0.9
     weight_decay: 0.00001
+trainer:
+  precision: 32
```
harimkang, line 129: Same question here.

src/otx/recipe/detection/rtmdet_tiny.yaml
samet-akcay: I actually have the same question, is there any specific motivation/decision for changing the default to 16?

harimkang: @samet-akcay That's a bit of a reverse question for me. By applying mixed precision, we can gain in terms of memory and speed. This is the approach we use in OTX1.5 as well.
I was actually wondering why this model couldn't use 16.

samet-akcay: oh, I've commented on the wrong file :)

samet-akcay: But it makes sense, thanks for the clarification

tests/integration/cli/test_cli.py
```diff
@@ -119,6 +119,7 @@ def test_otx_e2e(recipe: str, tmp_path: Path) -> None:
         f"base.data_dir={DATASET[task]['data_dir']}",
         f"base.work_dir={tmp_path_test}",
         f"base.output_dir={tmp_path_test / 'outputs'}",
+        "trainer=gpu",
```
vinnamkim: It is not desired to restrict these tests only for a GPU device. Would you please reuse and expand this parameterization
https://github.com/openvinotoolkit/training_extensions/blob/e563354601a734c767623091c30c80975a8635de/tests/regression/conftest.py#L158-L160
with CPU?

You should also add a marker here
https://github.com/openvinotoolkit/training_extensions/blob/e563354601a734c767623091c30c80975a8635de/pyproject.toml#L300-L302

src/otx/recipe/detection/rtmdet_tiny.yaml
jaegukhyun: This model should be trained by precision32 because in precision16, this model raises errors during training.
And this is a new model to OTX, that is we don't have RTMDet-tiny in otx1.x

src/otx/recipe/action_detection/x3d_fastrcnn.yaml
jaegukhyun: This model should be trained by precision32 because this model raises errors when the precision is 16. 
Furthermore, in OTX1.x, this model is also trained by precision 32

harimkang (dismissed):

tests/integration/cli/test_cli.py
jaegukhyun: Done by [db72f6c](https://github.com/openvinotoolkit/training_extensions/pull/2789/commits/db72f6c4abd756ec13503e41e1b424528510d143) Could you check it?

harimkang (dismissed):

MarkByun (dismissed):

sovrasov:
@jaegukhyun thanks for the PR. Does this activity include regression testing FP16 vs FP32? That transition may not be smooth for some of the models, since now we don't rely on already checked mm* FP16 infrastructure and configs.

jaegukhyun:
> @jaegukhyun thanks for the PR. Does this activity include regression testing FP16 vs FP32? That transition may not be smooth for some of the models, since now we don't rely on already checked mm* FP16 infrastructure and configs.

Through integration test we can assure that all models except for rtmdet and x3d_fastrcnn supports fp16 for training and testing.

Accuracy and speed comparison between FP32 and FP16 should be checked. We have checked regression test results by week. So we can compare them with new results after merging this PR.

sovrasov (approved):

harimkang (approved):

sungmanc (approved):

vinnamkim (approved):

PR #2792, opened 2024-01-15
Add ImageColorChannel to ImageInfo

vinnamkim:
### Summary

- Same as title
- By default, the OpenVINO data pipeline (`src/otx/config/data/openvino.yaml`) will use BGR color channel.

### How to test
Added an unit test for this change.

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/config/data/default.yaml
+++ b/src/otx/config/data/default.yaml
@@ -3,6 +3,7 @@ data_root: ${base.data_dir}
 
 mem_cache_size: 1GB
 mem_cache_img_max_size: null
+image_color_channel: RGB
 
 train_subset:
   subset_name: train
--- a/src/otx/config/data/openvino.yaml
+++ b/src/otx/config/data/openvino.yaml
@@ -1,6 +1,8 @@
 defaults:
   - default
 
+image_color_channel: BGR
+
 train_subset:
   transform_lib_type: TORCHVISION
   transforms:
--- a/src/otx/core/config/data.py
+++ b/src/otx/core/config/data.py
@@ -7,6 +7,7 @@
 from dataclasses import dataclass
 from typing import Any, Dict, List, Optional, Tuple
 
+from otx.core.types.image import ImageColorChannel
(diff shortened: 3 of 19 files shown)


sovrasov:
@vinnamkim do these changes mean we always use RGB for all torch models and all image reading methods, excepting OV models?

vinnamkim:
> @vinnamkim do these changes mean we always use RGB for all torch models and all image reading methods, excepting OV models?

It is recommended to use RGB as a default. This is because 1) our models' pretrained weights are almost based on RGB and 2) machine learning community has preferred to RGB. So I changed the default as RGB in this PR, https://github.com/openvinotoolkit/training_extensions/pull/2749.

In this PR, I wanted to introduce the color channel format which is selectable by the dataset configuration and trackable for ease of debugging. The reason why I expanded it is that I saw this comment https://github.com/openvinotoolkit/training_extensions/pull/2769#discussion_r1447229495

> I checked that BGR is by default used in our OTX. Answering to your question, OV requires BGR images, not RGB

sungmanc (approved):

jaegukhyun (approved):

samet-akcay (commented):
Would it be an idea to use one library rather than multiple ones interchangibly? PIL and cv2 are "almost" doing the same thing,  so i was wondering whether we could just use one of them? This would mean one-less-dependency. Also easier-maintainability and so on?

jaegukhyun:
https://github.com/openvinotoolkit/training_extensions/pull/2586#discussion_r1377534625
OpenCV can handle corrupted images which cannot be handled in PIL. However, I prefer to choose one framework also.

vinnamkim:
> Would it be an idea to use one library rather than multiple ones interchangibly? PIL and cv2 are "almost" doing the same thing, so i was wondering whether we could just use one of them? This would mean one-less-dependency. Also easier-maintainability and so on?

We are using PIL because of the severe performance issue on decoding JPEG.

harimkang (approved):

PR #2796, opened 2024-01-15
Fix multilabel training on non-trivial data

sovrasov:
### Summary
- Fix one-hot preparation: .sum(0) gives number of instances for each class instead of 0/1 values
- In ml-cls we suppose multiple objects to be located anywhere on an input image, therefore center crop augmentation is not suitable for test -> therefore we have also delete this on training to preserve objects scale

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/recipe/multilabel_classification/efficientnet_b0_light.yaml
+++ b/src/otx/recipe/multilabel_classification/efficientnet_b0_light.yaml
@@ -11,29 +11,23 @@ data:
       - type: LoadImageFromFile
       - backend: cv2
         scale: 224
-        type: RandomResizedCrop
+        type: Resize
       - type: PackInputs
   val_subset:
     batch_size: 64
     transforms:
       - type: LoadImageFromFile
       - backend: cv2
-        edge: short
-        scale: 256
-        type: ResizeEdge
-      - crop_size: 224
-        type: CenterCrop
+        scale: 224
+        type: Resize
       - type: PackInputs
   test_subset:
     batch_size: 64
     transforms:
       - type: LoadImageFromFile
       - backend: cv2
-        edge: short
-        scale: 256
-        type: ResizeEdge
-      - crop_size: 224
-        type: CenterCrop
+        scale: 224
+        type: Resize
       - type: PackInputs
 model:
   otx_model:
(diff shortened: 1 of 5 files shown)


src/otx/recipe/multilabel_classification/efficientnet_b0_light.yaml
```diff
@@ -11,29 +11,23 @@ data:
       - type: LoadImageFromFile
       - backend: cv2
         scale: 224
-        type: RandomResizedCrop
+        type: Resize
```
sungmanc, line 14: Why we need this change ?
We used the random resize crop at the OTX1.5 phase for the multilabel.
https://github.com/openvinotoolkit/training_extensions/blob/develop/src/otx/algorithms/classification/configs/base/data/data_pipeline.py

sovrasov: In OTX this was changed to crop for some reason: https://github.com/openvinotoolkit/training_extensions/blob/v1.0.0/otx/algorithms/classification/configs/base/data/data_pipeline.py#L23
Probably, that was done accidentally, when updating multiclass classification configs.
I conducted a fast experiment on VOC07, which indicated that omitting random crop looks beneficial. Overall, the data pipeline looks strange, because it doesn't contain actual augmentations. It's ok that we have an alignment with 1.x to some extent, but in terms of public benchmark it doesn't seem like 1.x had competitive performance. Since we target not only Geti now, we might revise all our models and compare vs public results at some later stage of OTX 2.0 development. Otherwise, that all might raise a lot of questions from external users who are familiar with the mainstream of the multilabel research.

|Model   | w/o random crop   | random crop   |
|---|---|---|
| efficientnet_b0_light  | 0.9663  | 0.9590  |

sungmanc (approved):

src/otx/recipe/multilabel_classification/efficientnet_b0_light.yaml
jaegukhyun: I think we can check this from 2.0 regression test

jaegukhyun (approved):

vinnamkim (approved):

PR #2799, opened 2024-01-16
Add adaptive validation interval

vinnamkim:
### Summary

- Request from the execution members.

### How to test
Added an unit test for this change.

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algo/callbacks/adaptive_train_scheduling.py
+++ b/src/otx/algo/callbacks/adaptive_train_scheduling.py
@@ -1,0 +1,80 @@
+# Copyright (C) 2023 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
+"""Callback to reschedule the validation interval adaptively."""
+
+from __future__ import annotations
+
+import logging as log
+import math
+from typing import TYPE_CHECKING
+
+from lightning import Callback
+
+if TYPE_CHECKING:
+    from lightning import LightningModule, Trainer
+
+
+class AdaptiveTrainScheduling(Callback):
+    """Adaptive Training Scheduling Hook.
+
+    Depending on the size of iteration per epoch, adaptively update the validation interval and related values.
+
+    Args:
+        max_interval: Maximum value of validation interval.
+            Defaults to 5.
+        decay: Parameter to control the interval. This value is set by manual manner.
+            Defaults to -0.025.
+    """
+
+    def __init__(self, max_interval: int = 5, decay: float = -0.025):
+        self.max_interval = max_interval
+        self.decay = decay
+        self._saved_check_val_every_n_epoch: int | None = None
+        self._saved_log_every_n_steps: int | None = None
+
+    def on_train_start(self, trainer: Trainer, pl_module: LightningModule) -> None:
+        """Execute this function at starting the train stage."""
+        max_interval = min(self.max_interval, trainer.max_epochs)
+        iter_per_epoch = len(trainer.train_dataloader)
+
+        adaptive_check_val_every_n_epoch = self._get_adaptive_interval(
+            iter_per_epoch=iter_per_epoch,
+            max_interval=max_interval,
+        )
+
+        if adaptive_check_val_every_n_epoch != trainer.check_val_every_n_epoch:
+            msg = (
+                "You are using AdaptiveTrainScheduling hook. "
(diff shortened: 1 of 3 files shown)


sungmanc (approved):
LGTM

src/otx/algo/callbacks/adaptive_train_scheduling.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def on_train_end(self, trainer: Trainer, pl_module: LightningModule) -> None:
+        """Execute this function at terminating the train stage."""
+        if self._saved_check_val_every_n_epoch:
+            trainer.check_val_every_n_epoch = self._saved_check_val_every_n_epoch
+            self._saved_check_val_every_n_epoch = None
+
+        if self._saved_log_every_n_steps:
+            trainer.log_every_n_steps = self._saved_log_every_n_steps
+            self._saved_log_every_n_steps = None
```
jaegukhyun, line 76: Do we need to keep original every_n_***?

jaegukhyun (approved):

src/otx/algo/callbacks/adaptive_train_scheduling.py
samet-akcay, line 1: I think callbacks also need to go to a sub-package of their own. As a user I should be able to access the callbacks via:
- `from otx.callbacks import AdaptiveTrainScheduling` or 
- `from otx.callbacks.adaptive_train_scheduling import AdaptiveTrainScheduling`

samet-akcay: I'll propose this in the future for all of the callbacks, so you could skip this comment for now :)
Just adding here for a future reference..

src/otx/algo/callbacks/adaptive_train_scheduling.py
samet-akcay: would be good to be as close to the Lightning terminology so the Lightning users would pick it up easily? Also this callback seems to be a modular one, maybe some users could even use it standalone outside otx?

harimkang (approved):
Thanks for the great feature.
One question is if we use this with Early-Stopping, shouldn't values like patience or interval of early-stop change?
What do you think?

src/otx/algo/callbacks/adaptive_train_scheduling.py
vinnamkim: > Do we need to keep original every_n_***?

We want to preserve the internal states of `pl.Trainer` immutable as possible.

vinnamkim: > would be good to be as close to the Lightning terminology so the Lightning users would pick it up easily? Also this callback seems to be a modular one, maybe some users could even use it standalone outside otx?

Sorry but what is the exactly one which is far away from the Lightning terminology?

vinnamkim:
> Thanks for the great feature. One question is if we use this with Early-Stopping, shouldn't values like patience or interval of early-stop change? What do you think?

We are using this callback with the early stopping hook as default. That's why I changed here
https://github.com/openvinotoolkit/training_extensions/blob/ddac7804dc9fe94bfa3a1876d7b53696371e4fb2/src/otx/config/callbacks/default.yaml#L13
Counting patience will be invoked only at the validation step.

src/otx/algo/callbacks/adaptive_train_scheduling.py
jaegukhyun: > > Do we need to keep original every_n_***?
> 
> We want to preserve the internal states of `pl.Trainer` immutable as possible.

Then, could every_n_epoch be different between output config from experiment and config during experiment?

vinnamkim: > Then, could every_n_epoch be different between output config from experiment and config during experiment?

Using this callback itself means that some Trainer configurations can be adaptively reconfigured at the runtime. That's why we invoke warnings here. However, we don't want to change the given configuration values affected by this callback itself.

samet-akcay: > Sorry but what is the exactly one which is far away from the Lightning terminology?

I was replying to the following:
> Do we need to keep original every_n_***?

PR #2801, opened 2024-01-16
Add register_explain_hook to OTXClsModel

negvet:
### Summary

Adding `register_explain_hook` implementation to the `OTX*ClsModel`.

Are we planning to have `OTXClsModel` as a parent class for {`OTXMulticlassClsModel`, `OTXMultilabelClsModel`, `OTXHlabelClsModel`}? `register_explain_hook` shares a common code for all cls models.

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/core/model/entity/base.py
+++ b/src/otx/core/model/entity/base.py
@@ -40,7 +40,6 @@ class OTXModel(nn.Module, Generic[T_OTXBatchDataEntity, T_OTXBatchPredEntity]):
         super().__init__()
 
         self._label_info = LabelInfo.from_num_classes(num_classes)
-        self.num_classes = num_classes
         self.classification_layers: dict[str, dict[str, Any]] = {}
         self.model = self._create_model()
         self.explain_hook = None
@@ -195,13 +194,6 @@ class OTXModel(nn.Module, Generic[T_OTXBatchDataEntity, T_OTXBatchPredEntity]):
 
         Args:
             output_dir: Directory path to save exported binary files
-        """
-        raise NotImplementedError
-
-    def register_explain_hook(self) -> None:
-        """Register explain hook.
-
-        TBD
         """
         raise NotImplementedError
 
--- a/src/otx/core/model/entity/base.py
+++ b/src/otx/core/model/entity/base.py
@@ -40,8 +40,10 @@ class OTXModel(nn.Module, Generic[T_OTXBatchDataEntity, T_OTXBatchPredEntity]):
         super().__init__()
 
         self._label_info = LabelInfo.from_num_classes(num_classes)
+        self.num_classes = num_classes
         self.classification_layers: dict[str, dict[str, Any]] = {}
         self.model = self._create_model()
+        self.explain_hook = None
 
     @property
     def label_info(self) -> LabelInfo:
--- a/src/otx/core/model/entity/classification.py
+++ b/src/otx/core/model/entity/classification.py
@@ -5,10 +5,12 @@
 
 from __future__ import annotations
 
+from abc import abstractmethod
 from typing import TYPE_CHECKING, Any
 
 import torch
 
+from otx.algo.hooks.recording_forward_hook import ReciproCAMHook
 from otx.core.data.entity.base import OTXBatchLossEntity
 from otx.core.data.entity.classification import (
     HlabelClsBatchDataEntity,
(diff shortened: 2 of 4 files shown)


src/otx/core/model/entity/base.py
```diff
@@ -40,8 +40,10 @@ def __init__(self, num_classes: int) -> None:
         super().__init__()
 
         self._label_info = LabelInfo.from_num_classes(num_classes)
+        self.num_classes = num_classes
```
vinnamkim: Can we get `num_classes` from `self.label_info`?
```suggestion
        @property
        def num_classes(self) -> int:
             return len(self.label_info)
```

src/otx/core/model/entity/classification.py
```diff
@@ -34,6 +36,33 @@ class OTXMulticlassClsModel(
...
+    @abstractmethod
+    def head_forward_fn(self, x):
+        """Defines neck (if available) and head forward function."""
+
+    @property
+    @abstractmethod
+    def has_gap(self):
+        """Defines if GAP is used right after backbone."""
+
+    def register_explain_hook(self, backbone: torch.nn.Module) -> None:
```
vinnamkim: You changed the function signature from `register_explain_hook(self)` to `register_explain_hook(self, backbone)` for this class.
However, I still prefer to `register_explain_hook(self)`. Furthermore, there is confusion from the mismatch on the way how we obtain `backbone` and `head_forward_fn`. It seems that we mandatorily need both to do the XAI algorithm, but the former comes from the argument, but the latter comes from `self`.

src/otx/core/model/entity/classification.py
```diff
@@ -34,6 +36,33 @@ class OTXMulticlassClsModel(
...
+        """Register explain hook at the model backbone output."""
+        self.explain_hook = ReciproCAMHook(
+            self.head_forward_fn,
+            num_classes=self.num_classes,
+            optimize_gap=self.has_gap,
+        )
+        self.explain_hook.handle = backbone.register_forward_hook(self.explain_hook.recording_forward)
```
vinnamkim: Can we wrap all the things for XAI algo inside of itself to clean up?

```suggestion
class ReciproCAMHook:
    def __init__(self, head_forward_fn, backbone, num_classes, optimize_gap):
        self.handle = backbone.register_forward_hook(self.recording_forward)

...

        self.explain_hook = ReciproCAMHook(
            head_forward_fn=self.head_forward_fn,
            backbone=backbone,
            num_classes=self.num_classes,
            optimize_gap=self.has_gap,
        )
```

src/otx/core/model/entity/base.py
negvet: Are we sure that there is always gonna be `self.label_info`?
Updated, take a look.

src/otx/core/model/entity/classification.py
negvet: Indeed, `backbone` and `head_forward_fn` are mandatory. I did it this way just to highlight both possible options of how we can deliver the state.
Updated to `register_explain_hook(self)`.

src/otx/core/model/entity/classification.py
negvet: I created a class method that does that.

negvet:
`ExplainClsHookMixIn` is mixed into e.g. `OTXMulticlassClsModel` and `ExplainClsHookMixIn.register_explain_hook()` is not stateless, i.e. has access to the state of the `OTXMulticlassClsModel`. 
How it is possible for the `ExplainClsHookMixIn` to request implementation of all required members (self.model, self.label_info, etc.) from `OTXMulticlassClsModel` and others? It is getting more hard also because all required members created in the parent `OTXModel` class.
Seems that property + abstractmethod does not work in this case.

src/otx/core/model/entity/base.py
vinnamkim: > Are we sure that there is always gonna be `self.label_info`? 

Yes

src/otx/core/model/entity/base.py
```diff
@@ -42,6 +42,7 @@ def __init__(self, num_classes: int) -> None:
         self._label_info = LabelInfo.from_num_classes(num_classes)
         self.classification_layers: dict[str, dict[str, Any]] = {}
         self.model = self._create_model()
+        self.explain_hook = None
```
vinnamkim, line 45: If we move `register_explain_hook()` itself to the mixin class, does it look better to move this variable to the mixin class as well?

vinnamkim (approved):
LGTM

src/otx/core/model/entity/base.py
vinnamkim: Nevermind I'll make another PR to revise this overall

vinnamkim:
> `ExplainClsHookMixIn` is mixed into e.g. `OTXMulticlassClsModel` and `ExplainClsHookMixIn.register_explain_hook()` is not stateless, i.e. has access to the state of the `OTXMulticlassClsModel`. How it is possible for the `ExplainClsHookMixIn` to request implementation of all required members (self.model, self.label_info, etc.) from `OTXMulticlassClsModel` and others? It is getting more hard also because all required members created in the parent `OTXModel` class. Seems that property + abstractmethod does not work in this case.

I slightly changed your PR in https://github.com/openvinotoolkit/training_extensions/pull/2810 to address what you said. Would you please refer to it?

negvet:
Let's merge it in https://github.com/openvinotoolkit/training_extensions/pull/2810

PR #2802, opened 2024-01-16
Update ignored scope for hrent_18_mod2

nikita-savelyevv:
### Summary

Set full ignored scope to hrnet_18_mod2 NNCF Accuracy Aware Training config. 

This is done because at the moment it's not possible to precisely control the resulting accuracy of the returned quantized model: the validation function provided for early stopping to NNCF AAT is not aligned with the evaluation function used for obtaining the final model score. That's why even if the resulting model is within 1% accuracy drop based on validation metric, it doesn't necessarily mean that it will be within 1% accuracy drop based on evaluation metric.

Ticket 129044.

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algorithms/segmentation/configs/ocr_lite_hrnet_18_mod2/compression_config.json
+++ b/src/otx/algorithms/segmentation/configs/ocr_lite_hrnet_18_mod2/compression_config.json
@@ -35,8 +35,8 @@
             }
           },
           "ignored_scopes": [
-            "{re}.*cross_resolution_weighting.*__mul__.*",
-            "{re}.*spatial_weighting.*__mul__.*"
+            "{re}.*backbone*",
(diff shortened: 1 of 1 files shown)


chuneuny-emily:
@nikita-savelyevv @sovrasov - This seems to be a fix related to issue Ticket 127047. Shouldn't the PR be created in the releases/1.4.0 branch rather than develop?

sovrasov:
It'll be backported to release once merged

nikita-savelyevv:
After internal NNCF discussion it was decided that adding the ignored scope covering the whole model does not make much sense since the performance of such model will be the same as for the original model. Because we're blocked by validation function problem in OTX, this issue should be addressed once this issue is resolved in OTX 2.0. For now, the corresponding test may be xfail-ed.

PR #2805, opened 2024-01-17
Refactor configuration & CLI

harimkang:
### Summary

I wanted to break up the PR as much as possible, but the CLI and configuration are still intertwined, so please excuse the large size of the PR. (Most files are YAML files.)

- Hydra -> Jsonargparse
- config + recipe -> recipe
- Configuration structure refactoring
- Add Jsonargparse based CLI
- LightnigModule Argument update (optimizer & scheduler argument update): https://lightning.ai/docs/pytorch/stable/cli/lightning_cli_advanced_3.html

Next PR
- Auto-configuration
- Engine.from_config

### How to test

image

```
(otx) otx train --config <Config-Path> --data_root <Data-Path> --model.num_classes 2
```

### Checklist

### License

```python
```


src/otx/cli/cli.py
```diff
@@ -0,0 +1,316 @@
...
+            action=ActionConfigFile,
+            help="Path to a configuration file in json or yaml format.",
+        )
+        parser.add_argument(
+            "--data_root",
+            type=str,
+            help="Path to dataset root.",
+        )
+        parser.add_argument(
+            "--task",
```
samet-akcay, line 100: We have also been discussing this for anomalib, which one would be clearer: task or task_type. I would like to get your opinion as well. Any thoughts?

src/otx/cli/cli.py
```diff
@@ -0,0 +1,316 @@
...
+            type=str,
+            help="Task Type.",
+        )
+        parser.add_argument(
+            "--callback_monitor",
+            type=str,
+            help="The metric for that model to monitor in the callback.",
```
samet-akcay: This description may not be clear enough to the reader/user.
Would be great if we could reword or add more description. 

It is because the user will see this description for the first time here in the cli

src/otx/cli/cli.py
```diff
@@ -0,0 +1,316 @@
...
+        and registers them with the parser.
+
+        Returns:
+            None
+        """
+        self._subcommand_parsers: dict[str, ArgumentParser] = {}
+        parser_subcommands = self.parser.add_subcommands()
+        self._set_extension_subcommands_parser(parser_subcommands)
+        if not _ENGINE_AVAILABLE:
+            return
```
samet-akcay, line 143: Would it be an idea to communicate here why engine returns? Maybe a logger.info? If that's too much, a comment maybe. Would be useful for future readers?

src/otx/cli/cli.py
```diff
@@ -0,0 +1,316 @@
...
+            description = get_short_docstring(fn)
+            added = sub_parser.add_method_arguments(
+                Engine,
+                subcommand,
+                skip=skip,
+                fail_untyped=False,
+            )
+
+            # Load default subcommand config file
+            default_config_file = Path(get_otx_root_path()) / "recipe" / "_base_" / f"{subcommand}.yaml"
```
samet-akcay: maybe `recipe` could be renamed to `recipes` as it contains multiple files?

src/otx/cli/cli.py
```diff
@@ -0,0 +1,316 @@
...
+            if default_config_file.exists():
+                with Path(default_config_file).open() as f:
+                    default_config = yaml.safe_load(f)
+                sub_parser.set_defaults(**default_config)
+
+            if "logger" in added:
```
samet-akcay: can you clarify what `added` means here? Maybe we could find a bit more descriptive variable name?

src/otx/cli/cli.py
```diff
@@ -0,0 +1,316 @@
...
+        if isinstance(optimizer_kwargs, Namespace):
+            optimizer_kwargs = namespace_to_dict(optimizer_kwargs)
+        scheduler_kwargs = self._get(self.config_init, "scheduler", {})
+        if isinstance(scheduler_kwargs, Namespace):
+            scheduler_kwargs = namespace_to_dict(scheduler_kwargs)
+        from otx.core.utils.instantiators import partial_instantiate_class
+
+        return model, partial_instantiate_class(optimizer_kwargs), partial_instantiate_class(scheduler_kwargs)
+
+    def _get(self, config: Namespace, key: str, default: Any = None) -> Any:  # noqa: ANN401
```
samet-akcay: is this the same method name from jasonargparse that we overwrite

src/otx/cli/cli.py
```diff
@@ -0,0 +1,316 @@
...
+        """Utility to get a config value which might be inside a subcommand."""
+        return config.get(str(self.subcommand), config).get(key, default)
+
+    def _parser(self, subcommand: str | None) -> ArgumentParser:
```
samet-akcay: same as this one? Is it the same name as jsonargparse. Maybe we could use a more descriptive method name?

src/otx/cli/cli.py
```diff
@@ -0,0 +1,316 @@
...
+    def save_config(self) -> None:
+        """Save the configuration for the specified subcommand.
+
+        The configuration is saved as a YAML file in the engine's working directory.
+
+        Args:
+            None
+
+        Returns:
+            None
```
samet-akcay: If these are None, these maybe removed?

src/otx/cli/cli.py
samet-akcay: how about something like `get_config_value` or `values` or something similar?

src/otx/cli/cli.py
```diff
@@ -0,0 +1,316 @@
...
+            self.instantiate_classes()
+            fn_kwargs = self._prepare_subcommand_kwargs(self.subcommand)
+            fn = getattr(self.engine, self.subcommand)
+            try:
+                fn(**fn_kwargs)
+            except Exception:
+                self.console.print_exception(width=self.console.width)
+            self.save_config()
+        else:
+            msg = f"Unrecognized subcommand: {self.subcommand}"
```
samet-akcay, line 331: is it possible to list the available subcommands as well?

src/otx/cli/utils/__init__.py
```diff
@@ -2,3 +2,24 @@
...
+        str: The root path of the otx module.
+
+    Raises:
+        ModuleNotFoundError: If the otx module is not found.
+    """
+    otx_module = importlib.import_module("otx")
+    if otx_module:
+        file_path = inspect.getfile(otx_module)
+        return str(Path(file_path).parent)
+    msg = "Cannot found otx."
```
samet-akcay, line 24: how about something liek cannot find otx in {path}, so when we see this error, we would also see what location we would look

src/otx/cli/utils/help_formatter.py
```diff
@@ -0,0 +1,222 @@
+"""Custom Help Formatters for OTX CLI."""
+
+# Copyright (C) 2023 Intel Corporation
```
samet-akcay: and for the rest :)
```suggestion
# Copyright (C) 2024 Intel Corporation
```

samet-akcay (commented):
i've reviewed until cli.py. Will continue tomorrow

src/otx/cli/cli.py
harimkang: This will return if no environment is configured to use Engine (without torch or lightning), causing the otx subcommand to only show install. I'll add a comment here.

src/otx/cli/cli.py
harimkang: I was torn between `configs` and `recipe` and `recipes`, so I left with the old name `recipe`. What do you think best?

src/otx/core/config/__init__.py
```diff
@@ -66,9 +31,10 @@ def as_torch_dtype(arg: str) -> dtype:
 
             ```yaml
             transforms:
-              - _target_: torchvision.transforms.v2.ToDtype
-                dtype: ${as_torch_dtype:torch.float32}
-                scale: True
+                - class_path: torchvision.transforms.v2.ToDtype
+                    init_args:
+                        dtype: ${as_torch_dtype:torch.float32}
+                        scale: True
```
jaegukhyun: Could you check intdentation of this?

src/otx/core/config/data.py
```diff
@@ -39,9 +39,4 @@ class DataModuleConfig:
     mem_cache_img_max_size: Optional[Tuple[int, int]] = None
     image_color_channel: ImageColorChannel = ImageColorChannel.RGB
 
-
-@dataclass
-class InstSegDataModuleConfig(DataModuleConfig):
-    """DTO for instance segmentation data module configuration."""
-
-    include_polygons: bool = True
+    include_polygons: bool = False
```
jaegukhyun, line 42: Could you explain why this bool variable should be changed?

src/otx/algo/classification/otx_dino_v2.py
```diff
@@ -65,8 +63,8 @@ def forward(self, imgs: torch.Tensor, labels: torch.Tensor = None) -> torch.Tens
 class DINOv2RegisterClassifier(OTXMulticlassClsModel):
     """DINO-v2 Classification Model with register."""
 
-    def __init__(self, num_classes: int, config: DictConfig) -> None:
-        self.config = config
+    def __init__(self, num_classes: int, config: DictConfig | dict) -> None:
```
vinnamkim: What's the reason behind of changing so many typing from `DictConfig` to `DictConfig | dict` not just for this line but many others?

src/otx/engine/engine.py
```diff
@@ -347,7 +347,8 @@ def _build_trainer(self, **kwargs) -> None:
         """Instantiate the trainer based on the model parameters."""
         if self._cache.requires_update(**kwargs) or self._trainer is None:
             self._cache.update(**kwargs)
-            self._trainer = Trainer(**self._cache.args)
+            kwargs = self._cache.args
+            self._trainer = Trainer(**kwargs)
```
jaegukhyun, line 352: It seems these changes do not chnage anything. Right?

src/otx/cli/cli.py
```diff
@@ -0,0 +1,316 @@
...
+        """Instantiate the model based on the subcommand.
+
+        This method checks if the subcommand is one of the engine subcommands.
+        If it is, it instantiates the model.
+
+        Returns:
+            tuple: The model and optimizer and scheduler.
+        """
+        model = self._get(self.config_init, "model")
+        optimizer_kwargs = self._get(self.config_init, "optimizer", {})
```
vinnamkim: What's the type of `self.config_init` and why `optimizer_kwargs` has more than two types (`Namespace, dict, ...?`)?

src/otx/cli/utils/__init__.py
```diff
@@ -2,3 +2,24 @@
 # SPDX-License-Identifier: Apache-2.0
 
 """CLI Utils."""
+
+import importlib
+import inspect
+from pathlib import Path
+
+
+def get_otx_root_path() -> str:
```
vinnamkim: Why don't we use `Path` as is?

src/otx/cli/utils/help_formatter.py
```diff
@@ -0,0 +1,222 @@
...
+            **kwargs (Any): Additional keyword arguments to pass to the superclass method.
+
+        Returns:
+            None
+        """
+        if self.subcommand in REQUIRED_ARGUMENTS:
+            if self.verbose_level == 0:
+                actions = []
+            elif self.verbose_level == 1:
+                actions = [action for action in actions if action.dest in REQUIRED_ARGUMENTS[self.subcommand]]
```
vinnamkim:
```suggestion
        actions = []
        if self.subcommand in REQUIRED_ARGUMENTS and self.verbose_level == 1:
                actions = [action for action in actions if action.dest in REQUIRED_ARGUMENTS[self.subcommand]]
```

src/otx/cli/utils/help_formatter.py
```diff
@@ -0,0 +1,222 @@
...
+        If the verbose level is set to 0, the argument is not added.
+        If the verbose level is set to 1 and the argument is not in the non-skip list, the argument is not added.
+
+        Args:
+            action (argparse.Action): The action to add to the help formatter.
+        """
+        if self.subcommand in REQUIRED_ARGUMENTS:
+            if self.verbose_level == 0:
+                return
+            if self.verbose_level == 1 and action.dest not in REQUIRED_ARGUMENTS[self.subcommand]:
```
vinnamkim: Is this safe if an user give an arbitrary subcommand such as `trani` not `train`?

src/otx/cli/utils/jsonargparse.py
```diff
@@ -0,0 +1,140 @@
...
+    """Get the short description from the docstring.
+
+    Args:
+        component (object): The component to get the docstring from
+
+    Returns:
+        Optional[str]: The short description
+    """
+    if component.__doc__ is None:
+        return None
```
vinnamkim: Can we `return ""` to make a single return type?

src/otx/cli/utils/jsonargparse.py
```diff
@@ -0,0 +1,140 @@
...
+        overrides = cfg.__dict__.pop("overrides", None)
+        if overrides is not None:
+            cfg.__dict__.update(overrides)
+        if cfg.get(dest) is None:
+            cfg[dest] = []
+        cfg[dest].append(cfg_path)
+
+
+Namespace.update = update
+ActionConfigFile.apply_config = apply_config
```
vinnamkim: Can we add a kind of contextmanager here rather than silently monkey-patching the upstream library?

src/otx/cli/cli.py
```diff
@@ -0,0 +1,316 @@
...
+from otx.cli.utils.jsonargparse import get_short_docstring
+
+if TYPE_CHECKING:
+    from jsonargparse._actions import _ActionSubCommands
+
+_ENGINE_AVAILABLE = True
+try:
+    from otx.engine import Engine
+except ImportError:
+    _ENGINE_AVAILABLE = False
```
sungmanc, line 31: Why insert `_` prefix ?

src/otx/algo/segmentation/litehrnet.py
```diff
@@ -12,7 +12,7 @@
 class LiteHRNet(MMSegCompatibleModel):
     """LiteHRNet Model."""
 
-    def __init__(self, num_classes: int, variant: Literal["18", "s", "x"]) -> None:
+    def __init__(self, num_classes: int, variant: Literal["18", 18, "s", "x"]) -> None:
```
jaegukhyun, line 15: It seems 18 is duplicated

src/otx/core/config/data.py
harimkang: Removed `InstSegDataModuleConfig` and merged it into `DataModuleConfig`. In the process, the `include_polygons` value was moved, and the default value is False, and modified to declare `include_polygons=True` in the config file of the models in the Instance-segmentation task.

src/otx/algo/segmentation/litehrnet.py
harimkang: Modified this to accept `int` as well.

src/otx/core/config/__init__.py
harimkang: Thanks, Done - https://github.com/openvinotoolkit/training_extensions/pull/2805/commits/c9cf77cb9ae825d337ee5421746f6d62093be76b

src/otx/engine/engine.py
harimkang: but because of this change, jsonargparse will know that kwargs will eventually be used in the trainer, which allows for overriding.

src/otx/cli/cli.py
harimkang: When `parse_args` in `argparser`, config is instantiated as a nested dictionary, so I modified it to accept a `dict` type.
Actually, it would be fine to modify it just for that class (the class that uses a dict argument named config in the config file// + OVModel), but for now I did the same for the model class, which uses config in the same way.

src/otx/algo/classification/otx_dino_v2.py
harimkang: When `parse_args` in `argparser`, config is instantiated as a nested dictionary, so I modified it to accept a `dict` type.
Actually, it would be fine to modify it just for that class (the class that uses a dict argument named config in the config file// + OVModel), but for now I did the same for the model class, which uses config in the same way.

src/otx/cli/utils/__init__.py
harimkang: Thanks! Done - https://github.com/openvinotoolkit/training_extensions/pull/2805/commits/1a254d9d1306b9389b89a5594eefa8faf43f2e38

src/otx/cli/utils/help_formatter.py
harimkang: Thanks! Done- https://github.com/openvinotoolkit/training_extensions/pull/2805/commits/1a254d9d1306b9389b89a5594eefa8faf43f2e38

src/otx/cli/utils/jsonargparse.py
harimkang: Thanks! Done - https://github.com/openvinotoolkit/training_extensions/pull/2805/commits/1a254d9d1306b9389b89a5594eefa8faf43f2e38

src/otx/cli/utils/jsonargparse.py
harimkang: Thanks for good suggestion! - Done https://github.com/openvinotoolkit/training_extensions/pull/2805/commits/768951e218e5b4b56ce29516c3c2dd73b44e5615

src/otx/cli/cli.py
harimkang: This looks like a mistake on my part. I've fixed it to communicate with Namespace. Thanks! - https://github.com/openvinotoolkit/training_extensions/pull/2805/commits/1a254d9d1306b9389b89a5594eefa8faf43f2e38

src/otx/cli/utils/help_formatter.py
harimkang: Thanks! - Done : https://github.com/openvinotoolkit/training_extensions/pull/2805/commits/3679a70296e06f1ed7c5fd6fab1899bf65fcf023

src/otx/cli/cli.py
harimkang: Thank! Done - https://github.com/openvinotoolkit/training_extensions/pull/2805/commits/3679a70296e06f1ed7c5fd6fab1899bf65fcf023

src/otx/cli/cli.py
harimkang: Thank! renamed it  - https://github.com/openvinotoolkit/training_extensions/pull/2805/commits/3679a70296e06f1ed7c5fd6fab1899bf65fcf023

src/otx/cli/cli.py
harimkang: Thank! renamed it - https://github.com/openvinotoolkit/training_extensions/commit/3679a70296e06f1ed7c5fd6fab1899bf65fcf023

src/otx/cli/cli.py
harimkang: Thank! This is to see what elements have been added to the parser. i've renamed this to `added_argument`.
https://github.com/openvinotoolkit/training_extensions/commit/3679a70296e06f1ed7c5fd6fab1899bf65fcf023

src/otx/cli/cli.py
harimkang: Do you have any suggestions on what I should change?

src/otx/cli/cli.py
harimkang: I think the install and engine subcommands should appear fluid, I'll try to think of a way to do this.

src/otx/algo/segmentation/litehrnet.py
sungmanc: It's unreadable for me, if we want to check whether `int` could be used or not, I think we need to add logic at another location. Maybe unit-test?

src/otx/cli/utils/__init__.py
harimkang: This is a function that finds the path based on the otx module in the environment where otx is installed, so I'm not sure which path it should show.

src/otx/cli/cli.py
```diff
@@ -0,0 +1,325 @@
...
+    def engine_subcommands() -> dict[str, set[str]]:
+        """Returns a dictionary of engine subcommands and their required arguments.
+
+        Returns:
+            A dictionary where the keys are the subcommands and the values are sets of required arguments.
+        """
+        device_kwargs = {"accelerator", "devices"}
+        return {
+            "train": device_kwargs,
+            "test": {"datamodule"}.union(device_kwargs),
```
sungmanc, line 123: Why we need {"datamodule"}.union?

src/otx/cli/utils/help_formatter.py
harimkang: Raises an error for disallowed subcommands, which are arbitrarily prefixed in argparse.
image

harimkang:
I'll be running regression tests against this branch, but until then, feel free to comment!

src/otx/core/model/module/action_classification.py
```diff
@@ -15,18 +17,26 @@
...
     """Base class for the lightning module used in OTX detection task."""
 
     def __init__(
         self,
         otx_model: OTXActionClsModel,
-        optimizer: torch.optim.Optimizer,
-        scheduler: torch.optim.lr_scheduler.LRScheduler,
         torch_compile: bool,
+        optimizer: OptimizerCallable = lambda p: torch.optim.SGD(p, lr=0.01),
+        scheduler: LRSchedulerCallable = torch.optim.lr_scheduler.ConstantLR,
```
sungmanc, line 32: Is the purpose of this to enable the auto-config?

src/otx/engine/engine.py
sungmanc, line 352: Can we convert ?
```python3
self._trainer = Trainer(**self._cache.args)
```

src/otx/algo/classification/otx_dino_v2.py
vinnamkim: Can we choose only one of either? If it is not an inevitable situation, we'd better to avoid multiple typings as possible to be developer or user friendly. In this case, it is even more so because they are used for exactly the same purpose.

> When parse_args in argparser, config is instantiated as a nested dictionary, so I modified it to accept a dict type

In addition, do we have custom resolvers for `DictConfig`? Are those correctly parsed?

src/otx/recipe/_base_/test.yaml
```diff
@@ -0,0 +1,26 @@
+callbacks:
```
sungmanc, line 1: It seems that `_base_/test` is not proper naming since it only contains the callbacks, logger.

In my opinion, the base yaml of `train` or `test` should include more close contents about the `train` or `test`. So, we might add `accerlator` or `dist_config` or such other things..
i.e. https://github.com/open-mmlab/mmdetection/blob/main/configs/_base_/default_runtime.py

sungmanc (commented):
Honestly, the CLI part is hard to understand at least for me since there are many argument related things `sub_parser` `sub_command`, `engine_subcommands`, `add_class_arguments`, etc. However, I don't have any nice suggestion. For the others, I left some comments

src/otx/cli/cli.py
harimkang: In the CLI flow, the `datamodule` comes in when the engine is created. However, engine can also get a `datamodule` in `test()`. This is a redundant case, so skip it in the CLI to disallow it.

src/otx/core/model/module/action_classification.py
harimkang: I was referring to the way Lightning provides an optimizer and scheduler.
https://lightning.ai/docs/pytorch/stable/cli/lightning_cli_advanced_3.html
It also uses `OptimizerCallable`, `LRSchedulerCallable` to cover all the optimizers and schedulers available in ligthning while making it easy to instantiate.

src/otx/engine/engine.py
harimkang: Your suggestion is the code before the modification. 
The reason I modified it is to help jsonargparse keep track that kwargs will be used for Trainer. This allows us to know that Trainer's arguments can be provided via kwargs, and helps with overriding in the CLI.

src/otx/recipe/_base_/test.yaml
harimkang: `train.yaml` and `test.yaml` are the default configuration files that correspond to the arguments to `engine.train()` and `engine.test()`, respectively.
These files correspond to `configs/train.yaml` & `configs/test.yaml` in the current v2.

For `test.yaml`, I'm thinking that things that could go in kwargs rather than arguments are declared in the file, which could be confusing. I'll think about this a bit more.

src/otx/cli/cli.py
harimkang: I used a prefix that is only used for that file, which means private. Do you think it's right to delete predix?

src/otx/algo/segmentation/litehrnet.py
harimkang: As it stands, this doesn't work with `int` 18. But shouldn't 18 be allowed to use `int` from the user's perspective?

src/otx/cli/cli.py
harimkang: I thought they meant the same thing. Wouldn't it be task? I thought it would be 'anomaly_classification', 'anomaly_detection', 'anomaly_segmentation' based on anomalib (just like in OTX 1.5).

However, I'm in favor of splitting task, train type and would like to see it discussed in the future.

src/otx/cli/cli.py
```diff
@@ -0,0 +1,325 @@
...
+            help="The metric to monitor the model performance during training callbacks.",
+        )
+        return parser
+
+    @staticmethod
+    def engine_subcommands() -> dict[str, set[str]]:
+        """Returns a dictionary of engine subcommands and their required arguments.
+
+        Returns:
+            A dictionary where the keys are the subcommands and the values are sets of required arguments.
```
ashwinvaidya17: On line 189 the values of this dictionary are used to skip the arguments. Maybe the doctoring should be updated to reflect this.

src/otx/cli/cli.py
```diff
@@ -0,0 +1,325 @@
...
+            default_config_file = get_otx_root_path() / "recipe" / "_base_" / f"{subcommand}.yaml"
+            if default_config_file.exists():
+                with Path(default_config_file).open() as f:
+                    default_config = yaml.safe_load(f)
+                sub_parser.set_defaults(**default_config)
+
+            if "logger" in added_arguments:
+                sub_parser.link_arguments("engine.work_dir", "logger.init_args.save_dir")
+            if "callbacks" in added_arguments:
+                sub_parser.link_arguments("callback_monitor", "callbacks.init_args.monitor")
```
ashwinvaidya17, line 216: Is this always true? What if the callback does not contain these two arguments?

src/otx/cli/utils/help_formatter.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+BASE_ARGUMENTS = {"config", "print_config", "help", "engine", "model", "model.help", "task"}
+ENGINE_ARGUMENTS = {"data_root", "engine.help", "engine.device", "engine.checkpoint", "engine.work_dir"}
+REQUIRED_ARGUMENTS = {
+    "train": {
+        "data",
+        "optimizer",
+        "scheduler",
+    }.union(BASE_ARGUMENTS).union(ENGINE_ARGUMENTS),
```
ashwinvaidya17: Not sure what's the convention here but I find the following easier to read
```python
{
...,
*BASE_ARGUMENTS,
*ENGINE_ARGUMENTS
}
```
But feel free to ignore this. It is a personal preference.

src/otx/cli/utils/jsonargparse.py
```diff
@@ -0,0 +1,152 @@
...
+        return ""
+    docstring = docstring_parser.parse(component.__doc__)
+    return docstring.short_description
+
+
+def flatten_dict(config: dict, parent_key: str = "", sep: str = ".") -> dict:
+    """Flatten a nested dictionary into a single-level dictionary.
+
+    Args:
+        d (dict): The dictionary to be flattened.
```
ashwinvaidya17, line 36:
```suggestion
        config (dict): The dictionary to be flattened.
```

ashwinvaidya17 (commented):
Thanks for the efforts. I don't have a lot of comments

src/otx/cli/utils/help_formatter.py
```diff
@@ -0,0 +1,220 @@
...
+        add_usage(usage, actions, *args, **kwargs)
+            Add usage information to the help output.
+        add_argument(action)
+            Add an argument to the help output.
+        format_help()
+            Format the help output.
+    """
+
+    verbosity_dict = get_verbosity_subcommand()
+    verbose_level = verbosity_dict["verbosity"]
```
samet-akcay: How about?
```suggestion
    verbosity_level = verbosity_dict["verbosity"]
```

src/otx/cli/utils/help_formatter.py
```diff
@@ -0,0 +1,220 @@
...
+
+        The help message includes information about the command's arguments and options,
+        as well as any additional information provided by the command's help guide.
+
+        Returns:
+            str: A string containing the formatted help message.
+        """
+        with self.console.capture() as capture:
+            section = self._root_section
+            _reder: RenderableType = section
```
samet-akcay: can you elaborate what `reder` means?

src/otx/cli/utils/jsonargparse.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+def get_short_docstring(component: TypeVar) -> str:
+    """Get the short description from the docstring.
+
+    Args:
+        component (object): The component to get the docstring from
+
+    Returns:
+        Optional[str]: The short description
```
samet-akcay: is this annotation correct?

src/otx/cli/utils/jsonargparse.py
```diff
@@ -0,0 +1,152 @@
...
+    only_unset: bool = False,
+) -> Namespace:
+    """Sets or replaces all items from the given nested namespace.
+
+    Args:
+        value: A namespace to update multiple values or other type to set in a single key.
+        key: Branch key where to set the value. Required if value is not namespace.
+        only_unset: Whether to only set the value if not set in namespace.
+    """
+    _dict_type = False
```
samet-akcay: How about 
```suggestion
    is_value_dict = False
```

src/otx/cli/utils/jsonargparse.py
```diff
@@ -0,0 +1,152 @@
...
+            if not only_unset or prefix + _key not in self:
+                self.update(val, prefix + _key)
+    if _dict_type and key is not None:
+        # Dict or Namespace -> Dict
+        self[key] = dict_to_namespace(self[key]).as_dict()
+    return self
+
+
+# To provide overriding of the Config file
+def apply_config(self: ActionConfigFile, parser: ArgumentParser, cfg: Namespace, dest: str, value: str) -> None:  # noqa: ARG001
```
samet-akcay, line 98: a bit nitpicky comment but, to reveal the actual intension, would it be an idea to name this as `apply_config_to_parser`? Would that be too verbose?

src/otx/core/model/module/action_classification.py
```diff
@@ -15,18 +17,26 @@
...
     ):
-        super().__init__(otx_model, optimizer, scheduler, torch_compile)
+        super().__init__(
+            otx_model=otx_model,
```
samet-akcay, line 35: optimizer and scheduler do not have otx prefix, while model does have it. Would it be an idea to just use `model` name

samet-akcay: so ideally it would be like `(..., model: ActionDetectionModel)`

Another pr though. Adding that for a future reference.

src/otx/core/utils/config.py
```diff
@@ -43,12 +43,13 @@ def to_list(dict_: dict) -> dict:
 
 
 def convert_conf_to_mmconfig_dict(
-    cfg: DictConfig,
+    cfg: DictConfig | dict,
```
samet-akcay, line 46: to follow the convention used throughout the function, would it be an idea to rename the function as `convert_config_to__mmconfig_dict` and `cfg` variable to `config`? This way it would be consistent with mmconfig and dictconfig.

It is irrelevant to your pr though. if you think this is something that could be done later, that's fine

src/otx/core/utils/instantiators.py
```diff
@@ -1,81 +1,89 @@
...
-#
-# This source code is borrowed from https://github.com/ashleve/lightning-hydra-template
-
-from typing import List
-
-import hydra
-from lightning import Callback
-from lightning.pytorch.loggers import Logger
-from omegaconf import DictConfig
+# Copyright (C) 2023 Intel Corporation
```
samet-akcay:
```suggestion
# Copyright (C) 2024 Intel Corporation
```

src/otx/core/utils/instantiators.py
```diff
@@ -1,81 +1,89 @@
...
         return callbacks
 
-    if not isinstance(callbacks_cfg, DictConfig):
-        raise TypeError("Callbacks config must be a DictConfig!")
-
-    for _, cb_conf in callbacks_cfg.items():
-        if isinstance(cb_conf, DictConfig) and "_target_" in cb_conf:
-            log.info(f"Instantiating callback <{cb_conf._target_}>")
-            callbacks.append(hydra.utils.instantiate(cb_conf))
+    for cb_conf in callbacks_cfg:
```
samet-akcay, line 38: again, conf and cfg used interchangeably here. would be good to stick to one

src/otx/recipe/_base_/train.yaml
samet-akcay, line 1: does these base callbacks mean that an api user won't have these by default? Ideally Engine should handle some of these callbacks by default such as rich progress bar, iteration timer etc. ?

src/otx/recipe/action/action_detection/x3d_fastrcnn.yaml
```diff
@@ -0,0 +1,93 @@
...
+  class_path: lightning.pytorch.cli.ReduceLROnPlateau
+  init_args:
+    mode: min
+    factor: 0.1
+    patience: 10
+    monitor: train/loss
+
+engine:
+  task: ACTION_DETECTION
+  device: auto
```
samet-akcay, line 24: we can pass the trainer args here, right?

src/otx/recipe/detection/rtmdet_tiny.yaml
```diff
@@ -1,124 +1,140 @@
...
-trainer:
+
+optimizer:
+  class_path: torch.optim.Adam
+  init_args:
+    lr: 1e-3
+    weight_decay: 0.0
+
+scheduler:
+  class_path: lightning.pytorch.cli.ReduceLROnPlateau
```
samet-akcay, line 14: is this class path correct? looks weird :)

samet-akcay (commented):
thanks again @harimkang. I've gone through the second round of my review

src/otx/algo/classification/otx_dino_v2.py
harimkang: The resolver for `DictConfig` is the same as before. (otx.core.config._init_.register_configs)
Unfortunately, jsonargparse does not provide `DictConfig` as a default parser.
(use dict by default.) -> Let's see if there's a way to do this.

So I guess there are 3 ways to do it.
Let's say we leave the configs of the mm-related models as they are, since they use only `DictConfig` (Currently they use `read_mmconfig`, so we don't need to change that.). 

The models that currently take a `DictConfig` as an argument that should be modified are `Dinov2` & `OVModel` for now.

1. leave this as `DictConfig` and specify `DictConfig` in the configuration file.
```
model:
  class_path: otx.algo.classification.otx_dino_v2.DINOv2RegisterClassifier
  init_args:
    num_classes: 1000
    config:
      class_path: omegaconf.dictconfig.DictConfig
      init_args:
        content:
          backbone:
            name: dinov2_vits14_reg
            frozen: false
          head:
            in_channels: 384
            num_classes: 1000
```
2. change this type as `dict`
3. Modifying arguments explicitly, without using config(dict, DictConfig)

For example 3, OVModel could be changed to something like this.
```python
class OVModel(OTXModel, Generic[T_OTXBatchDataEntity, T_OTXBatchPredEntity]):
    def __init__(self, num_classes: int, model_name:  str, ...) -> None:
```

Which method do you think is best?

Or maybe we should just make sure we get all of the `DictConfig | dict` and then figure out some other good way to do it.

src/otx/cli/cli.py
harimkang: If callbacks or logger come in as None, this is simply ignored.

src/otx/cli/utils/jsonargparse.py
harimkang: Thanks, but i already done - https://github.com/openvinotoolkit/training_extensions/pull/2805/commits/3679a70296e06f1ed7c5fd6fab1899bf65fcf023

src/otx/recipe/_base_/train.yaml
harimkang: Wouldn't it be better to provide this at the configuration level rather than in Engine? Of course, we could provide this in Engine if we added auto-configuration.

Which approach do you think is appropriate?

src/otx/recipe/action/action_detection/x3d_fastrcnn.yaml
harimkang: Yes, but at the same time, it can be used by declaring it in places like `train.yaml` (which is linked to `engine.train()`). In my opinion, this could be confusing for users using configuration, so I didn't add it to engine.

src/otx/recipe/detection/rtmdet_tiny.yaml
harimkang: https://github.com/Lightning-AI/pytorch-lightning/blob/97d71aba0b4739a43166acb574050f518079c5f4/src/lightning/pytorch/cli.py#L54

- Only for `ReduceLROnPlateau` are currently using overrides on the lightning.cli side, so i used this to match lightning for now.

src/otx/cli/cli.py
harimkang: Thanks! Done - https://github.com/openvinotoolkit/training_extensions/pull/2805/commits/ebc0361b443568eca751738fd744bbb8b140c59a

src/otx/cli/utils/help_formatter.py
harimkang: Thanks! Done - https://github.com/openvinotoolkit/training_extensions/commit/ebc0361b443568eca751738fd744bbb8b140c59a

src/otx/cli/utils/help_formatter.py
harimkang: Thanks! Done - https://github.com/openvinotoolkit/training_extensions/commit/ebc0361b443568eca751738fd744bbb8b140c59a

src/otx/cli/utils/jsonargparse.py
harimkang: Thanks! Fixed - https://github.com/openvinotoolkit/training_extensions/commit/ebc0361b443568eca751738fd744bbb8b140c59a

src/otx/cli/utils/jsonargparse.py
harimkang: Thanks! Done - https://github.com/openvinotoolkit/training_extensions/commit/ebc0361b443568eca751738fd744bbb8b140c59a

src/otx/core/utils/instantiators.py
harimkang: Thanks! Done - https://github.com/openvinotoolkit/training_extensions/commit/ebc0361b443568eca751738fd744bbb8b140c59a

src/otx/cli/utils/help_formatter.py
harimkang: This is what will be printed in the help message. I renamed this to `rendered_content` - https://github.com/openvinotoolkit/training_extensions/commit/ebc0361b443568eca751738fd744bbb8b140c59a

src/otx/core/utils/config.py
harimkang: Changing this will unnecessarily increase the number of file changes, which can be frustrating for reviewers, so I would suggest fixing this in one go in the near future.

src/otx/core/utils/instantiators.py
harimkang: Similar to the previous comment, I think it would be a good idea to modify the cfg to a word like config at some point in the future.

src/otx/core/model/module/action_classification.py
harimkang: This is currently associated with the OTXLitModule in core, so i expect it to change a lot to work with the current PR. It would be nice if we remembered this and renamed it to the appropriate word at once, just like the others.

src/otx/cli/utils/jsonargparse.py
harimkang: I think the name you suggested might be a bit confusing, and since it's used to replace an existing function called `apply_config`, I think it's better to keep it as it is, but what do you think?

src/otx/core/utils/instantiators.py
samet-akcay: > Similar to the previous comment, I think it would be a good idea to modify the cfg to a word like config at some point in the future.

I would love that :D

src/otx/cli/utils/jsonargparse.py
samet-akcay: oh if it is a jsonargparse replacement, then leave it as is, I agree.

src/otx/algo/classification/otx_dino_v2.py
vinnamkim: 1. It should be the option 1. This is because they were originally `MMConfig` but we integrated them into our `DictConfig` format at first. However, we did some refactoring to separate them into `algo/<task>/mmconfigs/*.yaml`, so that there is no need to keep them as `DictConfig` format. Finally, we will change again from `DictConfig` to `MMConfig` to get back them as they were. Not only here but also please revert the typing in `MM*CompatibleModel` such as
https://github.com/openvinotoolkit/training_extensions/blob/5d60742dcea63d49ad5da05deb7bbf0254bb8b4b/src/otx/core/model/entity/action_detection.py#L35

2. My second question is about the custom resolvers. How `jsonargparse` can resolve the custom Python object from the config file? The most important usage for this is
```yaml
    - _target_: torchvision.transforms.v2.ToDtype
      dtype: ${as_torch_dtype:torch.float32}
      scale: True
```
To resolve `torch.float32`, we introduced the custom resolver. I could see there remains this configuration only in this PR without the `torch.dtype` resolving:
https://github.com/openvinotoolkit/training_extensions/blob/c7ddf78d5460aee1a3095957cff14f83e7e1fa54/src/otx/recipe/_base_/data/torchvision_base.yaml#L7-L13

However, supporting TorchVision v2 transform is important. I'd like to ask you to add a test instantiating `OTXDataModule` including the `torch.dtype` resolving from the configuration file. It is no matter on how to do this. If there is a native support for this with `jsonargparse`, you can choose it, but the important thing is that we should support TorchVision v2 transform.

harimkang: Thanks for the insight.
I proceeded with number 1 as we discussed. - https://github.com/openvinotoolkit/training_extensions/pull/2805/commits/13b078582386d3562954b4e7d9aac5e3ed9514cb

I also checked out question 2. 
```
    - class_path: torchvision.transforms.v2.ToDtype
      init_args:
        dtype: ${as_torch_dtype:torch.float32}
        scale: True
```
We can use the transform above with jsonargparse.

vinnamkim (dismissed):

harimkang:
@vinnamkim Hi Vinnam,
I have an issue with this PR where the process doesn't end after the integration test, and when I looked at it locally, it seems to get stuck in the mem_cache side as shown below. Is there anything you could suspect?

image

jaegukhyun (dismissed):

src/otx/cli/cli.py
samet-akcay: We were discussing if `task` could be confusing to some users as it might indicate a full task implementation. It is instead defines the type of the task, so it might be an idea to use `task_type`. This was the idea. 

I agree with splitting `task_type`/`task` and `train_type`. Let's discuss this in a future PR

src/otx/cli/utils/jsonargparse.py
samet-akcay: the docstring seems to be showing `d` for some reason. Can you double check if the docstring is updated as well?

src/otx/recipe/action/action_detection/x3d_fastrcnn.yaml
samet-akcay: that's fair, we should make it clear  when we write the documentation that it is also possible. Let's keep that in mind when the time comes to write the documentation

src/otx/recipe/classification/multi_class_cls/openvino_model.yaml
```diff
@@ -0,0 +1,41 @@
...
+    config:
+      class_path: omegaconf.dictconfig.DictConfig
+      init_args:
+        content:
+          model_name: efficientnet-b0-pytorch
+          async_inference: True
+          use_throughput_mode: False
+          model_type: Classification
+          head:
+            num_classes: 1000
```
samet-akcay, line 14: i'm looking forward to refactor this part in the future where we don't pass configs inside configs :)

samet-akcay (dismissed):
thanks @harimkang. I think I'm happy with the changes. Great effort!

harimkang:
the process doesn't end as shown below with integration test.
image
image

It is only reproduced in the Visual Prompting model (GPU train & test) + openvino_model (CPU test only) test case.
so I disable visual_prompting SAM model in integration test case temporarily.
I have reported this to Vinnam and will revisit this at a later date.

sungmanc (approved):

vinnamkim:
Sorry for your inconvenience, but since the time zone is different, please resolve the conflict and merge after this PR #2764.

harimkang:
> Sorry for your inconvenience, but since the time zone is different, please resolve the conflict and merge after this PR #2764.

Yes, of course. If that PR passes all the tests and then gets merged, I'll fix it here.

vinnamkim (approved):
Since the integration tests of https://github.com/openvinotoolkit/training_extensions/pull/2764 were failed and seem hard to fix this time, let merge this first.

jaegukhyun (approved):

PR #2806, opened 2024-01-17
Add mmdeploy export

eunwoosh:
### Summary
This PR enables mmdeploy export to OTX2.0. Currently, detection and instance segmentation need mmdeploy for export.
This PR includes

- modify Exporter design for `mmdeploy`
- Add MMdeployExporter
- Add mmdeploy config for models
- Fix `OVInstanceSegmentationModel`

#### TBD

- This PR doesn't support exporting tiling. 

#### Difference between mmdeploy export in OTX 1.X  and OTX 2.0
When exporting a model w/ mmdeploy in OTX, model is converted to ONNX and ONNX is converted to OV.
In OTX 1.X, mmdeploy is used for both conversion but OTX 2.0 uses OpenVINO to convert ONNX model to OV.
It's for not falling behind latest update in OV because current latest mmdeploy is still using OpenVINO 22.03 but current latest version is 23.02.
I also removed unused code as mmdeploy ONNX -> OV is dropped.

#### Update integration test.
Export model score can be unstable if a model isn't fully trained.
To prevent it, some detection models get more epoch.
`maskrcnn_swint` isn't trained at all, so skip it in the test.

#### python format mmdeploy config 
Because config files given by mmdeploy has python format as default,
I leave it as it is not to give user additional burden to get mmdeploy configs for new models in OTX.

P.S. I bring some code unused now but for future (e.g. export tilling, configurable parameter related setting)

### Checklist

### License

```python
```


src/otx/config/mmdeploy/base/base_detection_dynamic.py
```diff
@@ -0,0 +1,22 @@
+"""Detection models dynamic deploy config."""
+
+_base_ = ["./base_detection_static.py"]
+
+ir_config = dict(
+    dynamic_axes={
+        "image": {
```
sovrasov: We don't need dynamic input for our models

src/otx/config/mmdeploy/base/base_detection_dynamic.py
```diff
@@ -0,0 +1,22 @@
...
+            0: "batch",
+            1: "channel",
+            2: "height",
+            3: "width",
+        },
+        "boxes": {
+            0: "batch",
+            1: "num_dets",
```
sovrasov: `num_dets` is the only dynamic dimension that's reasonable to keep here

src/otx/config/mmdeploy/base/base_detection_static.py
```diff
@@ -0,0 +1,41 @@
...
+        background_label_id=-1,
+    ),
+)
+
+backend_config = dict(
+    type="openvino",
+    mo_options=None,
+)
+
+input_data = dict(
```
sovrasov: Is that actually used?

src/otx/config/mmdeploy/base/base_detection_static.py
```diff
@@ -0,0 +1,41 @@
+"""Detection models static deploy config."""
+
+ir_config = dict(
```
sovrasov: This config can be merged with dynamic one

src/otx/config/mmdeploy/base/base_detection_dynamic.py
eunwoosh: ok I'll remove it.

src/otx/config/mmdeploy/base/base_detection_dynamic.py
eunwoosh: so you mean it need to exist right?

sovrasov: Probably that's true, because the number of output detections can vary

sovrasov (commented):
Thanks for your effort. My high-level comment is that mmdeploy export should be implemented in a standalone exporter (see the latest changes in my PR), and .py configs should be eliminated or at least minimized (1 model - 1 .py config) if possible

eunwoosh:
HI @sovrasov , thanks for your comment. I want to reply each of your comments.

> My high-level comment is that mmdeploy export should be implemented in a standalone exporter (see the latest changes in my PR).

I saw your change and it looks good to me. But I think I need to modify overall structure to apply mmdeploy because some of code doesn't consider mmdeploy way (e.g. `to_openvino` requires model instance but mmdeploy just needs model cfg). It is time consuming tasks so I want to fix accuracy issue in this design and integrate to your design later.

> .py configs should be eliminated or at least minimized (1 model - 1 .py config) if possible

Yes, actually I also thought to use another format such as yaml and merge it with hydra but I decided to use python format. It's because mmdeploy provides their configuration as python format so I don't want to make developer need to convert mmdeploy provided python cfg to other format when adding new mmdet models into OTX 2.0. And I understand your intention to reduce cfg file at least because I agonized that point. But It's against to current cfg design (reduce duplication by extracting common cfg and using it in each cfg). I'm not sure which design is better now.

So this is my thought about your comment. Maybe there are some points I haven't seen, so please let me know your opinion about that :)

src/otx/config/mmdeploy/base/base_detection_dynamic.py
eunwoosh: I have a question. If resize type is set as `fit_to_window`, I think input size will be dynamic. Then if then, are these values necessary?

src/otx/config/mmdeploy/base/base_detection_static.py
eunwoosh: yes input data for mmdeploy is depending on this value.

src/otx/core/model/entity/instance_segmentation.py
```diff
@@ -204,7 +241,7 @@ def _customize_outputs(
             )
             scores.append(torch.tensor([output.score for output in output_objects]))
             masks.append(torch.tensor([output.mask for output in output_objects]))
-            labels.append(torch.tensor([output.id for output in output_objects]))
+            labels.append(torch.tensor([output.id - 1 for output in output_objects]))
```
eunwoosh, line 287: If i-seg model is converted to OV, empty label should be added as first element. it is also aligned in this part also.

src/otx/config/mmdeploy/base/base_detection_static.py
eunwoosh: it seems unnecessary. I'll remove it.

src/otx/config/mmdeploy/base/base_detection_dynamic.py
eunwoosh: Can't `batch` be also dynamic?

src/otx/core/model/entity/base.py
```diff
@@ -186,13 +187,15 @@ def export(
...
         Returns:
             Path: path to the exported model.
         """
-        return self._exporter.export(self.model, output_dir, base_name, export_format, precision)
+        exporter = self._get_exporter(test_pipeline)
+        return exporter.export(self.model, output_dir, base_name, export_format, precision)
 
-    @property
-    def _exporter(self) -> OTXModelExporter:
+    def _get_exporter(self, test_pipeline: list[dict] | None) -> OTXModelExporter:
```
sovrasov: I'd opt for adding `test_pipeline` via `_export_parameters` (it has to be a class attribute then) rather than damaging the interface. @vinnamkim what do you think? Perhaps, new class attribute will affect snapshot structure, then that way won't work

src/otx/config/mmdeploy/base/base_detection_dynamic.py
sovrasov: In MAPI batch is always 1

sovrasov: OV prioritizes multiple infer requests over batching

src/otx/config/mmdeploy/base/base_detection_dynamic.py
sovrasov: MAPI should use a fixed input shape, as far as I know, so we definitely don't need dynamic h,w here.
If you think that's not the case, and MPAI's resize functions can output variable sizes, let's review this together

src/otx/core/model/entity/base.py
eunwoosh: I also agree with you but it's hard to find elegant way. Although I thought about various things, it seems strange for models to have a test dataset pipeline which is dataset related things. How do you think about it?

vinnamkim: > I'd opt for adding `test_pipeline` via `_export_parameters` (it has to be a class attribute then) rather than damaging the interface. @vinnamkim what do you think? Perhaps, new class attribute will affect snapshot structure, then that way won't work

Yea, it looks good to me. Please keep in mind that everything which is needed for the model exportation can be found from `_export_parameters` and please do not make a different source to avoid getting fragmented code base like OTX 1.x.

Shouldn't `test_pipeline` be configured by the model configuration? If so, I don't see the reason why it's exposed and injected from outside.

src/otx/config/mmdeploy/__init__.py
vinnamkim, line 1: Please move all the file in this directory close to their model source files in `src/otx/algo`.

src/otx/core/model/entity/base.py
eunwoosh: yes, it's impossible to get `test_pipeline` from model config.  so if we want to gather all info in single place, I think there are two options. One is passing `test_pipeline` to `OTXModel.__init__()`(maybe `OTXDetectionModel.__init__()`?) as argument. But is it ok? `test_pipeline` is required only when exporting det and i-seg models. So we should add arguments which is necessary only in limited cases. Moreover, if user decides to create model outside `Engine`, then user should pass `test_pipeline` by himself when creating model instance only for limited case.  Other option is setting `test_pipeline` after instnace is created. but I think it's bad idea. It's becase there is no hint that `test_pipeline` need to be set after instance is created for exporting det or i-seg models. Developer needs to inspect export code to infer it. These things are my concerns. If you have other ideas to avoid it, pelase let me know :) or even with this downside, if you think it's better than current design, then I'll follow it.

src/otx/config/mmdeploy/__init__.py
eunwoosh: done at https://github.com/openvinotoolkit/training_extensions/pull/2806/commits/a5ec24df81fcdf49fb9a4f2698fe1beed0d55e3a

src/otx/config/mmdeploy/base/base_detection_dynamic.py
eunwoosh: Do you mean that model will get single batch in MAPI? When I checked in the OTX sides, it can get multiple batch and output multiple batch output.

src/otx/core/model/entity/base.py
eunwoosh: I heard the idea from Vinnam in offline and trying to apply it.

eunwoosh: done at https://github.com/openvinotoolkit/training_extensions/pull/2806/commits/a3a7ec55961379f670af4cdbd8d852c4c0875772.
Vinnam's idea is to make fake `test_pipeline` referring input_size. I checked it works (IR file is exactly same).

vinnamkim (dismissed):

src/otx/config/mmdeploy/base/base_detection_dynamic.py
sovrasov: Yes, exactly, see OV adapter and ImageModel implementations in MAPI if you're interested in that. Batch != 1 is not supported there.

sovrasov (dismissed):

harimkang (dismissed):

sovrasov (dismissed):

sovrasov (dismissed):

harimkang (dismissed):

sovrasov (dismissed):

sovrasov (dismissed):

src/otx/algo/detection/mmdeploy/atss.py
```diff
@@ -0,0 +1,11 @@
+"""MMDeploy config of ATSS model for Detection Task."""
```
sungmanc:
```suggestion
# Copyright (C) 2024 Intel Corporation
# SPDX-License-Identifier: Apache-2.0
#
"""MMDeploy config of ATSS model for Detection Task."""
```

Could you please add the license to your files? many .py files located at the mmdeploy folder missed the license

src/otx/core/exporter/base.py
```diff
@@ -211,3 +210,56 @@ def _embed_openvino_ir_metadata(ov_model: openvino.Model, metadata: dict[tuple[s
...
 
         return ov_model
+
+    def _extend_model_metadata(self, metadata: dict[tuple[str, str], str]) -> dict[tuple[str, str], str]:
+        """Extends metadata coming from model with preprocessing-specific parameters.
+
+        Model's original metadata has priority over exporter's extra metadata
+
+        Args:
+            metadata (dict[tuple[str, str], str]): _description_
```
sungmanc: Please add the description

sungmanc (commented):
It seems that there are many TODO or workaround in this PR. Could we make the separate backlog for those issues to avoid the missing ?

src/otx/algo/detection/mmdeploy/atss.py
eunwoosh: This is copy from mmdeploy repository. Is it ok to add our copyright?

src/otx/core/exporter/base.py
eunwoosh: done at https://github.com/openvinotoolkit/training_extensions/pull/2806/commits/241980ac6bab566316f6e9e19b541195f2771756

src/otx/algo/detection/mmdeploy/atss.py
sungmanc: @yunchu  could you check this ? How can we deal with this case?

yunchu: if you copied the contents from the other opensource, it's better to keep the original header and recommended to add some commenst about the origination. I think this [file](https://github.com/openvinotoolkit/training_extensions/blob/7c3617ea3427c497375f80e945ff56984da45ef2/src/otx/algorithms/visual_prompting/adapters/pytorch_lightning/models/visual_prompters/segment_anything.py) would be the example of that.

yunchu:
For you guys information, currently there is an issue using the mmdeploy to the v1. that's related with the `protobuf` package that is used by mmdeploy and the `protobuf` v3 has some vulnerability issue so that required to be updated to the v4 but the latest version of mmdeploy used v3 yet.

src/otx/algo/detection/mmdeploy/atss.py
sungmanc: Thanks a lot :)

eunwoosh: I checked mmdeploy file but there is no copyright in config file. I'll just add reference in each file.

eunwoosh: done at https://github.com/openvinotoolkit/training_extensions/pull/2806/commits/7a6ac90ea226e9c1eb06f86fdba52d81c4f13f76

pyproject.toml
```diff
@@ -249,6 +250,7 @@ exclude = [
     "node_modules",
     "venv",
     "tests/assets",
+    "src/otx/algo/**/mmdeploy/",
```
harimkang, line 253: It's minor, but can you put this down to 258 lines?

harimkang (approved):

pyproject.toml
eunwoosh: I leave it as same as mmdeploy intentionally. I just don't want make developer to be confused later.

sungmanc (approved):

jaegukhyun (approved):

PR #2808, opened 2024-01-17
Added segmentation configuration for export to align new interface

kprokofi:
### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algo/classification/deit_tiny.py
+++ b/src/otx/algo/classification/deit_tiny.py
@@ -26,7 +26,6 @@ class DeitTinyForMulticlassCls(MMPretrainMulticlassClsModel):
 
     def __init__(self, num_classes: int) -> None:
         config = read_mmconfig("deit_tiny", subdir_name="multiclass_classification")
-        self.image_size = config["data_preprocessor"].get("size", (224, 224))
         super().__init__(num_classes=num_classes, config=config)
 
 
@@ -35,5 +34,4 @@ class DeitTinyForMultilabelCls(MMPretrainMultilabelClsModel):
 
     def __init__(self, num_classes: int) -> None:
         config = read_mmconfig("deit_tiny", subdir_name="multilabel_classification")
-        self.image_size = config["data_preprocessor"].get("size", (224, 224))
         super().__init__(num_classes=num_classes, config=config)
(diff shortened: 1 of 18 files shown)


src/otx/algo/segmentation/segnext.py
```diff
@@ -15,4 +15,9 @@ class SegNext(MMSegCompatibleModel):
     def __init__(self, num_classes: int, variant: Literal["b", "s", "t"]) -> None:
         model_name = f"segnext_{variant}"
         config = read_mmconfig(model_name=model_name)
+        self.image_size = config["data_preprocessor"].get("size", (512, 512))
```
sovrasov: Is 512 the final resolution for SegNext?
Is it aligned with OTX torch test?

src/otx/core/model/entity/segmentation.py
```diff
@@ -153,11 +153,12 @@ def _customize_outputs(
         )
 
     def _configure_export_parameters(self) -> None:
+        image_size = getattr(self, "image_size", (512, 512))
```
sovrasov: Since this is the entity class, it should be clean from checking attribute existence tricks, so I'd create `image_size` attribute in constructor

src/otx/algo/classification/deit_tiny.py
```diff
@@ -18,6 +18,7 @@ def __init__(self, num_classes: int, num_multiclass_heads: int, num_multilabel_c
         config = read_mmconfig(model_name="deit_tiny", subdir_name="hlabel_classification")
         config.head.num_multiclass_heads = num_multiclass_heads
         config.head.num_multilabel_classes = num_multilabel_classes
+        self.image_size = config["data_preprocessor"].get("size", (224, 224))
```
sovrasov: Barely image resolution would ever appear in cls mmconfigs, moreover, we aim to get rid of them, so, I'd avoid extra mentions of those.

sovrasov: Instead, if keeping `self.image_size` in the base class, here we can just assign it's value here

src/otx/algo/segmentation/segnext.py
kprokofi: Actually, 512 512 looks better for both models, but I need to test it more. Let's keep 544 for now

sovrasov (approved):

PR #2810, opened 2024-01-18
Et/v2 register explain hook

vinnamkim:
### Checklist

### License

```python
```

### DIFF
--- a/src/otx/core/model/entity/classification.py
+++ b/src/otx/core/model/entity/classification.py
@@ -5,10 +5,12 @@
 
 from __future__ import annotations
 
+from abc import abstractmethod
 from typing import TYPE_CHECKING, Any
 
 import torch
 
+from otx.algo.hooks.recording_forward_hook import ReciproCAMHook
 from otx.core.data.entity.base import OTXBatchLossEntity
 from otx.core.data.entity.classification import (
     HlabelClsBatchDataEntity,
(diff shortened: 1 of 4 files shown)


negvet (approved):

jaegukhyun (approved):
Overall seems nice, but I have a concern that user might confuse between callbacks and hooks. I don't have nice suggestion, but let's think about more proper naming for these classes.

vinnamkim:
> Overall seems nice, but I have a concern that user might confuse between callbacks and hooks. I don't have nice suggestion, but let's think about more proper naming for these classes.

I just wanted to keep the momentum from OTX 1.* which has been used a PyTorch Module hook for XAI. However, I think that it would be better to expose a dedicated `forward` function for XAI functionality. For example,

```python
class ExplainableOTXClsModel:
     def forward_with_xai(self, inputs: OTXBatchDataEntity) -> OTXBatchPredWithXAI:
          ...
          return prediction_with_xai
```

Could we move forward to this way in the future? @negvet

harimkang (approved):

negvet:
> > Overall seems nice, but I have a concern that user might confuse between callbacks and hooks. I don't have nice suggestion, but let's think about more proper naming for these classes.
> 
> I just wanted to keep the momentum from OTX 1.* which has been used a PyTorch Module hook for XAI. However, I think that it would be better to expose a dedicated `forward` function for XAI functionality. For example,
> 
> ```python
> class ExplainableOTXClsModel:
>      def forward_with_xai(self, inputs: OTXBatchDataEntity) -> OTXBatchPredWithXAI:
>           ...
>           return prediction_with_xai
> ```
> 
> Could we move forward to this way in the future? @negvet

Yeah, we will look into this direction

src/otx/core/model/entity/classification.py
```diff
@@ -29,9 +30,52 @@
     from torch import device, nn
 
 
-class OTXMulticlassClsModel(
-    OTXModel[MulticlassClsBatchDataEntity, MulticlassClsBatchPredEntity],
-):
+class ExplainableOTXClsModel(OTXModel[T_OTXBatchDataEntity, T_OTXBatchPredEntity]):
```
samet-akcay, line 33: Isn't it possible to attach the hook to an existing classification model? I was wondering if it is really needed to create an additional object for this? I have a concern that there is a growing list of tasks/objects we have, which will ultimately be getting a bit confusing to users.

PR #2812, opened 2024-01-18
Introduce the logic to load the OTX1.X checkpoints (Det, iSeg, sSeg, Action cls, det)

sungmanc:
### Summary

Same as title, classification backward compatibility PR will be made after this PR. 

Checked with, (not checked all models)
* Detection : ATSS
* iSeg: EfficientNetb2b-MaskRCNN
* sSEg: SegNext-b, lite_hrnet_18
* Action cls: x3d
* Action det: x3d_fast_rcnn

### Checklist

### License

```python
```


src/otx/core/model/module/segmentation.py
```diff
@@ -126,3 +126,11 @@ def test_step(self, inputs: SegBatchDataEntity, batch_idx: int) -> None:
...
         """Metric name that the learning rate scheduler monitor."""
         return "train/loss"
+
+    def _get_state_dict_from_ckpt(self, ckpt: dict) -> dict:
+        """Get the state_dict, supporting the backward compatibility."""
+        state_dict = super()._get_state_dict_from_ckpt(ckpt)
+        for key in list(state_dict.keys()):
+            if "ham.bases" in key or "decode_head.aggregator.projects" in key:
+                state_dict.pop(key)
+        return state_dict
```
sungmanc: @kprokofi , it seems that segmentation models are not compatible with the OTX1.5. Could you check from your side? If you trained the model at the OTX1.5 and then load the checkpoint in the OTX2.0, you might meet an error when you use the `litehrnet_18`.

Although the error log just showed the mis-match of the shape, however, I think there is a gap of model itself between OTX1.5 and OTX2.0. 
image

src/otx/core/model/module/base.py
```diff
@@ -133,12 +133,31 @@ def state_dict(self) -> dict[str, Any]:
...
+            return load_from_prev_otx_ckpt(model_state_dict)
+        return ckpt["state_dict"]
+
     def load_state_dict(self, state_dict: dict[str, Any], *args, **kwargs) -> None:
         """Load state dictionary from checkpoint state dictionary.
 
         If checkpoint's meta_info and OTXLitModule's meta_info are different,
         load_state_pre_hook for smart weight loading will be registered.
         """
+        state_dict = self._get_state_dict_from_ckpt(state_dict)
```
vinnamkim:
```suggestion
        if ckpt["VERSION"] == 1 and "model" in state_dict:
            model_state_dict = state_dict["model"]["state_dict"]
            state_dict = _load_from_otx_1_x_ckpt(model_state_dict)
```

src/otx/core/model/module/base.py
```diff
@@ -133,12 +133,31 @@ def state_dict(self) -> dict[str, Any]:
         state_dict["meta_info"] = self.meta_info
         return state_dict
 
+    def _get_state_dict_from_ckpt(self, ckpt: dict[str, Any]) -> dict[str, Any]:
```
vinnamkim: Why don't we expose `load_from_prev_otx_ckpt` directly? Can we introduce appropriate notifications as well?

```suggestion
    def _load_from_otx_1_x_ckpt(self, ckpt: dict[str, Any]) -> dict[str, Any]:
         msg = "You are trying to load the model checkpoint created by OTX 1.*. Please switch it to the OTX 2.0 format by saving it again. ...."
         warnings.warn(msg)
         ...
```

sungmanc: We need to consider the checkpoints which was saved from OTX1.X and OTX2.0.

In the case of OTX2.0 (lightning), there is no dict_key named `model`. model state_dict was saved at the
* OTX1.5 : ckpt['model']['state_dict']
* OTX2.0 : ckpt['state_dict']

So, I wanted to make the branch to distinguish the ckpt comes from OTX1.x or 2.0. In addition, in the case of OTX1.x, we need to add proper prefix to the ckpt. That's the reason why I implemented like this.

src/otx/core/model/module/base.py
sungmanc: Thanks for the suggestion, however, OTX2.0 checkpoint doesn't have `VERSION` key. We might have another option:  change the `state_dict` function to save the `VERSION` info at the OTX2.0.

vinnamkim: I still cannot understand why `_get_state_dict_from_ckpt()` should exist. It does something if `"model" in ckpt and ckpt["VERSION"] == 1`. 
https://github.com/openvinotoolkit/training_extensions/blob/b2b0ae0f48a7d184e130c10744a4fb1728a43526/src/otx/core/model/module/base.py#L148-L151

It means that it exists for doing something only for `version==1`, and otherwise, it does nothing. However, we cannot infer anything about its behavior from its naming.

sungmanc: I get your point, thanks. https://github.com/openvinotoolkit/training_extensions/pull/2812/commits/c1a146cc4c63d768a4adca7b670fce1d73747a5c

src/otx/core/model/module/base.py
sungmanc: Done, https://github.com/openvinotoolkit/training_extensions/pull/2812/commits/c1a146cc4c63d768a4adca7b670fce1d73747a5c

src/otx/core/model/module/detection.py
```diff
@@ -133,3 +133,11 @@ def test_step(self, inputs: DetBatchDataEntity, batch_idx: int) -> None:
     def lr_scheduler_monitor_key(self) -> str:
         """Metric name that the learning rate scheduler monitor."""
         return "train/loss"
+
+    def _get_state_dict_from_ckpt(self, ckpt: dict) -> dict:
```
vinnamkim: Do we need to change this as well? Would you please add unit tests in this PR?

src/otx/core/model/module/base.py
```diff
@@ -133,12 +134,32 @@ def state_dict(self) -> dict[str, Any]:
         state_dict["meta_info"] = self.meta_info
         return state_dict
 
-    def load_state_dict(self, state_dict: dict[str, Any], *args, **kwargs) -> None:
+    def _load_from_prev_otx_ckpt(self, state_dict: dict[str, Any]) -> dict[str, Any]:
+        """Attach the model.model prefix to load without problem."""
+        msg = "Trying to load the model checkpoint created by OTX 1.X. it will be converted to OTX2.0 format."
```
samet-akcay, line 141: Any idea, for how long are we going to support this backward compatibility? If it is temporary, it might also be an idea to add a deprecation warning for future releases?

src/otx/core/utils/utils.py
```diff
@@ -142,3 +142,25 @@ def get_metric_value(metric_dict: Dict[str, Any], metric_name: str) -> Optional[
...
+
+    Args:
+        ckpt (dict): the checkpoint file
+
+    Returns:
+        bool: True means the checkpoint comes from otx1
+    """
+    return "model" in ckpt and ckpt["VERSION"] == 1
+
+def is_ckpt_for_finetune(ckpt: dict):
```
samet-akcay: how about `is_ckpt_to_finetune` or `is_ckpt_for_finetuning`?

src/otx/core/model/module/detection.py
samet-akcay: is this method only needed for detection and segmentation tasks? If it is needed by all the tasks, I see a duplicated pattern in the methods that might potentially be refactored, something like:

```python
    def _get_state_dict_from_ckpt(self, ckpt: dict, condition: bool) -> dict:
        """Get the state_dict, supporting the backward compatibility."""
        state_dict = super()._get_state_dict_from_ckpt(ckpt)
        for key in list(state_dict.keys()):
            if condition:
                state_dict.pop(key)
        return state_dict
```

and in task-based implementation, something like

```python
self._get_state_dict_from_ckpt(ckpt=ckpt, condition=key.startswith("model.model.ema_"))
```

This is a bit minor comment though. You could skip if you disagree

samet-akcay (commented):
thanks @sungmanc, only some minor comments..

src/otx/core/model/module/base.py
sungmanc: For the backward compatibility, as far as I know, we need to wait for the decision from Geti staked holders. Maybe it isn't just temporary.

src/otx/core/model/module/detection.py
sungmanc: > Do we need to change this as well? Would you please add unit tests in this PR?

For the detection and instance segmentation tasks, there were additional chekcpoint keys for the EMA model. So, we need to eliminate theses keys to successfully load the checkpoint. I'll add the unit-test, thanks.

sungmanc: > is this method only needed for detection and segmentation tasks? If it is needed by all the tasks, I see a duplicated pattern in the methods that might potentially be refactored, something like:
> 
> ```python
>     def _get_state_dict_from_ckpt(self, ckpt: dict, condition: bool) -> dict:
>         """Get the state_dict, supporting the backward compatibility."""
>         state_dict = super()._get_state_dict_from_ckpt(ckpt)
>         for key in list(state_dict.keys()):
>             if condition:
>                 state_dict.pop(key)
>         return state_dict
> ```
> 
> and in task-based implementation, something like
> 
> ```python
> self._get_state_dict_from_ckpt(ckpt=ckpt, condition=key.startswith("model.model.ema_"))
> ```
> 
> This is a bit minor comment though. You could skip if you disagree

It would be more clear for me as well, however, I think the current is better because there were no common condition between tasks. i.e. for the segmentation models, we need to more complex rules. I'll consider the common condition at the future.

sungmanc: > Do we need to change this as well? Would you please add unit tests in this PR?

Thanks for the pointing out, I missed to edit the function name. For the tests, I think we need to add the backward compatibility tests by using the previous model's checkpoints. I'll start the discussion about this agenda in the near future

src/otx/core/utils/utils.py
sungmanc: Done, https://github.com/openvinotoolkit/training_extensions/pull/2812/commits/1f55383109a5342291c7c8b1d24d9bb3893f66c1

src/otx/core/model/module/detection.py
sungmanc: Rename the function: https://github.com/openvinotoolkit/training_extensions/pull/2812/commits/1f55383109a5342291c7c8b1d24d9bb3893f66c1

jaegukhyun (dismissed):

harimkang (dismissed):

src/otx/core/model/module/instance_segmentation.py
```diff
@@ -161,3 +161,11 @@ def test_step(self, inputs: InstanceSegBatchDataEntity, batch_idx: int) -> None:
     def lr_scheduler_monitor_key(self) -> str:
         """Metric name that the learning rate scheduler monitor."""
         return "train/loss"
+
+    def _get_state_dict_from_ckpt(self, ckpt: dict) -> dict:
```
vinnamkim:
```suggestion
    def _load_from_prev_otx_ckpt(self, ckpt: dict) -> dict:
```

Please add an unit test for each task.

sungmanc: Done, https://github.com/openvinotoolkit/training_extensions/pull/2812/commits/807b0dc82b3807baf40979aa6f704c81740e9591.

vinnamkim (approved):
LGTM.

harimkang (approved):

jaegukhyun (approved):

PR #2815, opened 2024-01-18
Added multilabel and hierarchical classification models for IR inference

kprokofi:
### Checklist

### License

```python
```


kprokofi:
I also introduced "stack_images" parameter in DataConfig to solve problem with collate function for PyTorch cls models and OV models.

kprokofi:
Hierarchical OV Model currently can't produce some meaningful metric. We should align it later, however it is essential to add interface and be able to run inference even without accuracy computation.

src/otx/core/data/entity/classification.py
```diff
@@ -65,6 +65,20 @@ def task(self) -> OTXTaskType:
...
         entities: list[MulticlassClsDataEntity],
+    ) -> MulticlassClsBatchDataEntity:
+        """Collection function to collect `OTXDataEntity` into `OTXBatchDataEntity` in data loader."""
+        batch_data = super().collate_fn(entities)
+        return MulticlassClsBatchDataEntity(
+            batch_size=batch_data.batch_size,
+            images=tv_tensors.Image(data=torch.stack(batch_data.images, dim=0)),
+            imgs_info=batch_data.imgs_info,
+            labels=[entity.labels for entity in entities],
+        )
```
vinnamkim:
```suggestion
        stack_image: bool = True
    ) -> MulticlassClsBatchDataEntity:
        """Collection function to collect `OTXDataEntity` into `OTXBatchDataEntity` in data loader."""
        batch_data = super().collate_fn(entities, stack_image)
        return MulticlassClsBatchDataEntity(
            batch_size=batch_data.batch_size,
            images=batch_data.images,
            imgs_info=batch_data.imgs_info,
            labels=[entity.labels for entity in entities],
        )
```

src/otx/core/data/dataset/classification.py
```diff
@@ -62,6 +62,8 @@ def _get_item_impl(self, index: int) -> MulticlassClsDataEntity | None:
     @property
     def collate_fn(self) -> Callable:
         """Collection function to collect MulticlassClsDataEntity into MulticlassClsBatchDataEntity in data loader."""
+        if not self.stack_images:
+            return MulticlassClsBatchDataEntity.collate_unstacked_fn
         return MulticlassClsBatchDataEntity.collate_fn
```
vinnamkim:
```suggestion
        return functools.partial(MulticlassClsBatchDataEntity.collate_fn, stack_images=self.stack_images)
```

kprokofi:
@harimkang 
image
Can it be a CI problem with env? Locally it works

src/otx/core/model/entity/classification.py
```diff
@@ -341,3 +352,79 @@ def _customize_outputs(
...
+        num_multilabel_classes: int,
+    ) -> None:
+        config.num_multiclass_heads = num_multiclass_heads
+        config.num_multilabel_classes = num_multilabel_classes
+        super().__init__(num_classes, config)
+
+    def set_hlabel_info(self, hierarchical_info: HLabelInfo) -> None:
+        """Set hierarchical information in model head.
+
+        Since OV IR model consist of all required hierarchy inforamtion,
```
sungmanc:
```suggestion
        Since OV IR model consist of all required hierarchy information,
```

sungmanc (dismissed):
LGTM with minor fix

jaegukhyun (dismissed):

src/otx/config/data/openvino.yaml
```diff
@@ -2,6 +2,7 @@ defaults:
   - default
 
 image_color_channel: BGR
+stack_images: False
```
samet-akcay, line 5: I cannot seem to find any description in any of the docstrings for `stack_images`. Would it be possible to add some description in the relevant docstring to tell the user what it is and what it does?

samet-akcay (changes requested):
Thanks @kprokofi. Overall it looks good. I have only a single comment: as a reader, I'm not sure if I understand the role of `stack_images`. Can you add a description in the docstring of the relevant module where it is defined?

src/otx/config/data/openvino.yaml
kprokofi: I added a docstring for the OTXDataset and a description for the parameters from the data config.

sovrasov (dismissed):

kprokofi:
> Thanks @kprokofi. Overall it looks good. I have only a single comment: as a reader, I'm not sure if I understand the role of `stack_images`. Can you add a description in the docstring of the relevant module where it is defined?

Done, could you take a look?

src/otx/config/data/openvino.yaml
samet-akcay: ah sorry, you had to go through the other params too :) Thanks for doing that though!

src/otx/core/data/dataset/base.py
```diff
@@ -74,7 +74,20 @@ def from_num_classes(cls, num_classes: int) -> LabelInfo:
...
+    Defines basic logic for OTX datasets.
+
+    Args:
+            dm_subset: datumaro subset of a dataset
+            transforms: transforms to apply on images
+            mem_cache_handler: handler of the images cache
+            mem_cache_img_max_size: max size of images to put in cache
+            max_refetch: maximum number of images to fetch in cache
+            image_color_channel: collor channel of images
+            stack_images: whether or not to stack images in collate function in OTXBatchData entity.
```
samet-akcay: Just one minor comment, after which I could approve :)
```suggestion
    Args:
        dm_subset: Datumaro subset of a dataset
        transforms: Transforms to apply on images
        mem_cache_handler: Handler of the images cache
        mem_cache_img_max_size: Max size of images to put in cache
        max_refetch: Maximum number of images to fetch in cache
        image_color_channel: Color channel of images
        stack_images: Whether or not to stack images in collate function in OTXBatchData entity.
```

src/otx/core/model/entity/base.py
```diff
@@ -233,10 +234,9 @@ def __init__(self, num_classes: int, config: DictConfig) -> None:
         self.config = config
         super().__init__(num_classes)
 
-    def _create_model(self) -> nn.Module:
+    def _create_model(self, configuration: dict[str, Any] | None = None) -> Model:
```
vinnamkim: 1. Please do not use `None`able typing as possible if there is no special issue. In this case, you can use `configuration: dict[str, Any] = {}`
2. In addition, please do not change the function signature of `_create_model(self)` derived from the base model if there is no special reason. In this case, I can see L253
```python
return Model.create_model(model_adapter, model_type=self.model_type, configuration=configuration)
```
One configuration is passed from the member variable (`self.model_type`), but another comes from the function argument `configuration`. Please keep one single channel for it such as
```python
def _create_model(self):
     return Model.create_model(model_adapter, model_type=self.model_type, configuration=self.configuration)
```

samet-akcay (dismissed):
thanks a lot!

kprokofi:
Since we changed our config based approach, I need to refactor base OVModel class. I propose to make it separate from OTXModel and move all parameters from config to constructor

kprokofi:
I moved OVModel from OTXModel inheritance. It is the first step of refactoring. For now, OVModel should imitate nn.Model, it is wrong logic, though.
I also want to refactor LitModule and create a separate one for OVModels. However, it is for the next PR and discussions.

src/otx/core/model/entity/base.py
kprokofi: 1. configuration: dict[str, Any] = {} raise mypy issue
2.  Please, see my new updates. OVModel isn't inherited from OTXModel now

kprokofi:
Seems like some problems with reading action dataset
image
@jaegukhyun , do you know this problem?

src/otx/core/model/entity/base.py
vinnamkim: What's the reason that we should maintain two `OTXModel` and `OVModel` base classes although they share 99% of codes?

vinnamkim (changes requested):
See https://github.com/openvinotoolkit/training_extensions/pull/2815#discussion_r1464179243

src/otx/core/model/entity/base.py
kprokofi: There are many reasons. They don't share 99% of code, unfortunately, every change now brings more and more gap between these two entitites.
1) Currently, OVModel looks like ugly Workaround to fit OTXModel. After CLI-API changes, it is really hard to work with it in in its current form. We should use config and support ugly structure for that, it is now hard to reload parameters for OVModel via CLI. We need to change constructor and put all of the configuration parameters there to have clear API and easy modification.
2) OVModel isn't nn.Module. It doesn't support schedulers, optimizers, training and validation phase at all. It doesn't need to support class incremental and all methods related to checkpoint loading.
It is completely different entity meanwhile it share "naming" of the methods, but not the nature.

Going further, I would like to divide LitModule as well and make OVModel independent and easy to modify and understand. It is required more changes and I would like to make a proposal on our OTX forum next Tuesday
For now I propose to at least divide OVModel   and modify constructor to make our work with OpenVINO functionality easier.

vinnamkim: > They don't share 99% of code

I can see that you already copied many same code pieces from `OTXModel` here
https://github.com/openvinotoolkit/training_extensions/blob/a39a7a9baee91e06a6b446e6081a1e1309c51a2d/src/otx/core/model/entity/base.py#L212-L314

> Currently, OVModel looks like ugly Workaround to fit OTXModel. After CLI-API changes, it is really hard to work with it in in its current form. We should use config and support ugly structure for that, it is now hard to reload parameters for OVModel via CLI. We need to change constructor and put all of the configuration parameters there to have clear API and easy modification.

What's the problem on
```python
class OVModel(OTXModel, Generic[T_OTXBatchDataEntity, T_OTXBatchPredEntity]):
    def __init__(
        self,
        num_classes: int,
        model_name: str,
        ...):
        super().__init__(num_classes=num_classes)
        ...
```
? You can change the constructor of `OVModel` freely.

> OVModel isn't nn.Module. It doesn't support schedulers, optimizers, training and validation phase at all. It doesn't need to support class incremental and all methods related to checkpoint loading.
It is completely different entity meanwhile it share "naming" of the methods, but not the nature.

The reason why `OVModel` should be derived from `OTXModel` is that it can shares data pipelines, evaluation loop, and metric computations in a single system. I still don't know what is the exact disadvantage of inheriting from `nn.Module` because the computational cost of deriving from `nn.Module` is ignorable. But, if you really doesn't like `nn.Module`, we might consider changing `OTXModel` as a pure Python class and introducing an additional layer such as

```python
class OTXModel:
     ...

class OTXTorchModel(OTXModel, nn.Module):
     ...

class OVModel(OTXModel):
     ...
```

> Going further, I would like to divide LitModule as well and make OVModel independent and easy to modify and understand.

I'm really against on maintaining two different systems in one product. We already had many lessons from OTX 1.x.

kprokofi: I agree with you considering maintenance of two systems. However, I want to make OVModel more independent and clear entity without any export, train functionality. 
Maybe, creating base meta class is a good option. I will review possibility to make OVModel more independent, but without creating new LitModule

For now, let's change constructor for OVModel. It is essential to avoid ugly initialization of this class and change easily via CLI arguments . I made corresponding changes.

sovrasov:
> Seems like some problems with reading action dataset image @jaegukhyun , do you know this problem?

I also faced this problem

sovrasov (dismissed):

src/otx/recipe/classification/multi_label_cls/openvino_model.yaml
```diff
@@ -0,0 +1,37 @@
...
+  device: auto
+
+callback_monitor: val/accuracy
+
+data: ../../_base_/data/torchvision_base.yaml
+
+overrides:
+  data:
+    config:
+      image_color_channel: RGB
```
sovrasov, line 36: data_format: datumaro

src/otx/recipe/classification/h_label_cls/openvino_model.yaml
```diff
@@ -0,0 +1,40 @@
...
+  device: auto
+
+callback_monitor: val/accuracy
+
+data: ../../_base_/data/torchvision_base.yaml
+
+overrides:
+  data:
+    config:
+      image_color_channel: RGB
```
sovrasov, line 39: data_format: datumaro

src/otx/recipe/classification/multi_label_cls/openvino_model.yaml
```diff
@@ -0,0 +1,37 @@
...
+scheduler:
+  class_path: lightning.pytorch.cli.ReduceLROnPlateau
+  init_args:
+    mode: min
+    factor: 0.5
+    patience: 1
+    monitor: train/loss
+
+engine:
+  task: MULTI_CLASS_CLS
```
sovrasov: MULTI_LABEL_CLS

src/otx/recipe/classification/h_label_cls/openvino_model.yaml
```diff
@@ -0,0 +1,40 @@
...
+scheduler:
+  class_path: lightning.pytorch.cli.ReduceLROnPlateau
+  init_args:
+    mode: min
+    factor: 0.5
+    patience: 1
+    monitor: train/loss
+
+engine:
+  task: MULTI_CLASS_CLS
```
sovrasov: H_LABEL_CLS

sovrasov (dismissed):

src/otx/core/model/entity/base.py
```diff
@@ -231,20 +230,25 @@ class OVModel(OTXModel, Generic[T_OTXBatchDataEntity, T_OTXBatchPredEntity]):
...
+        model_type: str,
+        async_inference: bool = True,
+        max_num_requests: int | None = None,
+        use_throughput_mode: bool = True,
+    ) -> None:
+        self.model_name = model_name
+        self.model_type = model_type
+        self.async_inference = async_inference
+        self.num_requests = max_num_requests if max_num_requests is not None else get_default_num_async_infer_requests()
+        self.use_throughput_mode = use_throughput_mode
```
vinnamkim:
```suggestion
        use_throughput_mode: bool = True,
        model_api_configuration
    ) -> None:
        self.model_name = model_name
        self.model_type = model_type
        self.async_inference = async_inference
        self.num_requests = max_num_requests if max_num_requests is not None else get_default_num_async_infer_requests()
        self.use_throughput_mode = use_throughput_mode
        self.model_api_configuration = model_api_configuration
```

src/otx/core/model/entity/base.py
```diff
@@ -231,20 +230,25 @@ class OVModel(OTXModel, Generic[T_OTXBatchDataEntity, T_OTXBatchPredEntity]):
...
         super().__init__(num_classes)
 
-    def _create_model(self) -> nn.Module:
+    def _create_model(self, model_api_configuration: dict[str, Any] | None = None) -> Model:
```
vinnamkim:
```suggestion
    def _create_model(self) -> Model:
```

src/otx/core/model/entity/base.py
```diff
@@ -257,13 +261,40 @@ def _create_model(self) -> nn.Module:
             plugin_config=plugin_config,
         )
 
-        return Model.create_model(model_adapter, model_type=self.model_type)
+        model_api_configuration = model_api_configuration if model_api_configuration is not None else {}
```
vinnamkim:
```suggestion
        model_api_configuration = self.model_api_configuration if self.model_api_configuration is not None else {}
```

src/otx/core/model/entity/base.py
```diff
@@ -257,13 +261,40 @@ def _create_model(self) -> nn.Module:
...
+        """Set this model label information."""
+        if isinstance(label_info, list):
+            label_info = LabelInfo(label_names=label_info)
+
+        self._label_info = label_info
+
+    @property
+    def num_classes(self) -> int:
+        """Returns model's number of classes. Can be redefined at the model's level."""
+        return self.label_info.num_classes
```
vinnamkim:
```suggestion
```

src/otx/core/model/entity/base.py
```diff
@@ -257,13 +261,40 @@ def _create_model(self) -> nn.Module:
...
         images = [np.transpose(im.numpy(), (1, 2, 0)) for im in entity.images]
         return {"inputs": images}
 
+    def _customize_outputs(
+        self,
+        outputs: Any,  # noqa: ANN401
+        inputs: T_OTXBatchDataEntity,
+    ) -> T_OTXBatchPredEntity | OTXBatchLossEntity:
+        """Customize OTX output batch data entity if needed for you model."""
+        raise NotImplementedError
```
vinnamkim:
```suggestion
```

src/otx/core/model/entity/classification.py
```diff
@@ -305,6 +308,14 @@ def _create_model(self) -> nn.Module:
         self.classification_layers = classification_layers
         return model
 
+    def set_hlabel_info(self, hierarchical_info: HLabelInfo) -> None:
+        """Set hierarchical information in model head.
+
+        Args:
+            hierarchical_info: the label information represents the hierarchy.
+        """
+        self.model.head.set_hlabel_info(hierarchical_info)
```
vinnamkim, line 316:
```suggestion
```

src/otx/core/model/entity/classification.py
```diff
@@ -391,3 +402,79 @@ def _customize_outputs(
...
+
+        Since OV IR model consist of all required hierarchy information,
+        this method serves as placehloder
+        """
+        return
+
+    def _create_model(self, *args) -> Model:
+        # confidence_threshold is 0.0 to return scores for all multilabel classes
+        model_api_configuration = {"hierarchical": True, "confidence_threshold": 0.0}
+        return super()._create_model(model_api_configuration)
```
vinnamkim:
```suggestion
class OVHlabelClassificationModel(OVModel[HlabelClsBatchDataEntity, HlabelClsBatchPredEntity, T_OTXTileBatchDataEntity]):
    """Hierarchical classification model compatible for OpenVINO IR inference.

    It can consume OpenVINO IR model path or model name from Intel OMZ repository
    and create the OTX classification model compatible for OTX testing pipeline.
    """

    def __init__(
        self,
        num_classes: int,
        model_name: str,
        model_type: str,
        async_inference: bool = True,
        max_num_requests: int | None = None,
        use_throughput_mode: bool = True,
        num_multiclass_heads: int = 1,
        num_multilabel_classes: int = 0,
    ) -> None:
        self.num_multiclass_heads = num_multiclass_heads
        self.num_multilabel_classes = num_multilabel_classes
        model_api_configuration = {"hierarchical": True, "confidence_threshold": 0.0}
        super().__init__(num_classes, ..., model_api_configuration)

    def set_hlabel_info(self, hierarchical_info: HLabelInfo) -> None:
        """Set hierarchical information in model head.

        Since OV IR model consist of all required hierarchy information,
        this method raises an error if the input information is not equal to the information written in the OV model
        """
        if hierarchical_info != self.hierarchical_info:
              msg = ...
              raise ValueError(msg)
```

src/otx/core/model/entity/classification.py
```diff
@@ -391,3 +402,79 @@ def _customize_outputs(
...
+        return HlabelClsBatchPredEntity(
+            batch_size=len(outputs),
+            images=inputs.images,
+            imgs_info=inputs.imgs_info,
+            scores=pred_scores,
+            labels=pred_labels,
+        )
+
+
+class OVMultilabelClassificationModel(OVModel):
```
vinnamkim:
```suggestion
class OVMultilabelClassificationModel(OVModel[MultilabelClsBatchDataEntity, MultilabelClsBatchPredEntity, T_OTXTileBatchDataEntity]):
```

vinnamkim (changes requested):

src/otx/core/model/entity/base.py
kprokofi: Why do you want to put model_api_configuration in constructor? Not every model needed it. Only multilabel and hierarchical. It is not needed to configure while creating class. Why don't you want pass it in _create_model?

src/otx/core/model/entity/classification.py
kprokofi: If you remove it -> hierarchical classification won't work

src/otx/core/model/entity/base.py
vinnamkim: Please see https://github.com/openvinotoolkit/training_extensions/pull/2815#discussion_r1462605971.

src/otx/core/model/entity/classification.py
kprokofi: Why do you remove it? without it -> hierarchical classification won't work

src/otx/core/model/entity/classification.py
kprokofi: What is self.hierarchical_info? OVModel doesn't have `hierarchical_info` attribute

src/otx/core/model/entity/classification.py
vinnamkim: Sorry for my mistake. I forgot  to remove this comment. I wanted https://github.com/openvinotoolkit/training_extensions/pull/2815#discussion_r1465744303 for `set_hlabel_info`.

src/otx/core/model/entity/classification.py
vinnamkim: Is there any interface to get information needed for hierarchical CLS task from MAPI? What I wanted to is leaving `set_hlabel_info` doing nothing, but failing fast as possible if an user tries to assign an invalid `HLabelInfo` into this class.

src/otx/core/model/entity/base.py
kprokofi: I changed it: [commit](https://github.com/openvinotoolkit/training_extensions/pull/2815/commits/224f39c4c4a5283e2f383d0ff484dc912618c34f)

src/otx/core/model/entity/classification.py
kprokofi: Yes, [added it](https://github.com/openvinotoolkit/training_extensions/pull/2815/commits/f0d0e4b3e8615ac6c911edb5c56cfa4559971a31)

vinnamkim (approved):

sovrasov (approved):

samet-akcay (approved):
thanks a lot!

PR #2818, opened 2024-01-19
Add XAI entry point

negvet:
### Checklist

### License

```python
```


src/otx/core/types/explain.py
```diff
@@ -0,0 +1,17 @@
+# Copyright (C) 2023 Intel Corporation
```
sungmanc:
```suggestion
# Copyright (C) 2024 Intel Corporation
```

src/otx/core/types/explain.py
```diff
@@ -0,0 +1,17 @@
...
+# SPDX-License-Identifier: Apache-2.0
+#
+"""OTX explain type definition."""
+
+from __future__ import annotations
+
+from enum import Enum
+
+
+class TargetExplainGroup(str, Enum):
```
sungmanc, line 11: Could you explain what is the difference between `IMAGE` and `PREDICTIONS` ?

samet-akcay: I agree with @sungmanc. Would it be possible to expand the docstring a bit? A user may not be familiar with this concept, and it would be good if we make this clear for every level of users

negvet: Extended docstring

negvet:
@sovrasov @GalyaZalesskaya @sungmanc @samet-akcay @harimkang @vinnamkim please take a look

src/otx/core/model/module/base.py
```diff
@@ -205,3 +209,7 @@ def export(self, output_dir: Path, export_format: OTXExportFormat) -> None:
             export_format: Format in which this `OTXModel` is exported.
         """
         self.model.export(output_dir, export_format)
+
+    def forward(self, *args, **kwargs) -> OTXBatchPredEntity | OTXBatchLossEntity:
+        """Model forward pass."""
+        return self.model.forward(*args, **kwargs)
```
vinnamkim, line 215: Why do we need to redefine `forward()` here?

src/otx/engine/engine.py
```diff
@@ -327,6 +328,61 @@ def export(self, *args, **kwargs) -> None:
...
+        lit_module.model.register_explain_hook()
+
+        self._build_trainer(**kwargs)
+
+        self.trainer.predict(
+            model=lit_module,
+            datamodule=datamodule,
+            ckpt_path=ckpt_path,
+        )
+        saliency_maps = self.trainer.model.model.explain_hook.records
```
vinnamkim, line 380: I hope that we have a PyTorch Lightning callback to flush `explain_hook.records` into the local disk as files periodically. The main memory cannot maintain the raw tensors which have the same size as input images.

src/otx/core/config/explain.py
```diff
@@ -0,0 +1,17 @@
...
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+from otx.core.types.explain import TargetExplainGroup
+
+
+@dataclass
+class ExplainConfig:
+    """DTO for explain configuration."""
```
GalyaZalesskaya: I see that in other files in `config` folder we use DTO abbreviation, but is it possible to write the full version here as well?

src/otx/core/types/explain.py
```diff
@@ -0,0 +1,22 @@
...
+    """OTX target explain group definition.
+
+    Enum contains the following values:
+        IMAGE - This implies that single global saliency map will be generated for input image.
```
GalyaZalesskaya, line 15: Do I understand correctly that `Image` means activation map? But can we comment `Image` mentioning here since we don't have activation map hook for now?

src/otx/engine/engine.py
GalyaZalesskaya: @vinnamkim As I understand, these returned saliency maps are much smaller than the initial image. These are grayscale images with size [7,7] for classification or approximately [23,31] for detection for each class

src/otx/engine/engine.py
```diff
@@ -327,6 +328,61 @@ def export(self, *args, **kwargs) -> None:
...
+        cv2.imwrite(str(Path(self.work_dir) / "saliency_map.tiff"), saliency_maps[0][0])
```
GalyaZalesskaya, line 382: If it is an example of saving map, can you please add a comment like `Saving saliency map for the image 0, class 0`?

src/otx/engine/engine.py
```diff
@@ -327,6 +328,61 @@ def export(self, *args, **kwargs) -> None:
...
+        return saliency_maps
```
GalyaZalesskaya, line 383: The saving saliency maps in the file system will be handled by other PR, right? Is it possible to add here the info message that saliency maps are generated successfully with the following shape?

src/otx/engine/engine.py
```diff
@@ -327,6 +328,61 @@ def export(self, *args, **kwargs) -> None:
...
+            ...     explain_config=ExplainConfig(),
+            ... )
+        """
+        from pathlib import Path
+
+        import cv2
+
+        ckpt_path = str(checkpoint) if checkpoint is not None else self.checkpoint
+        if explain_config is None:
+            explain_config = ExplainConfig()
```
GalyaZalesskaya, line 359: Do we pass  explain_config somewhere in this function?

src/otx/engine/engine.py
vinnamkim: Even if it's not as memory intensive as I thought, it doesn't seem neat to me to store many things proportional to the dataset size in main memory. Do you mind whether adding todo here to leave it as a backlog?

src/otx/engine/engine.py
```diff
@@ -327,6 +328,61 @@ def export(self, *args, **kwargs) -> None:
...
+            list: Saliency maps.
+
+        Example:
+            >>> engine.explain(
+            ...     datamodule=OTXDataModule(),
+            ...     checkpoint=<checkpoint/path>,
+            ...     explain_config=ExplainConfig(),
+            ... )
+        """
+        from pathlib import Path
```
sovrasov: Why do we need to import Path here? We already have this import in TYPE_CHECKING section. It can just be moved out of it to enable the actual usage of the package

src/otx/core/model/module/base.py
negvet: `LightningModule` requires forward to be defined for `trainer.predict`.

src/otx/engine/engine.py
negvet: Galina's comment is correct. Which one update - this is not the case for maskrcnn model.
So in the general case it make sense to manage this memory. Will add todo.

src/otx/engine/engine.py
negvet: Now - not. But we will probably when we will filter out unsessesary maps and do preprocessing.

src/otx/engine/engine.py
negvet: Right, updated

src/otx/engine/engine.py
negvet: For other engine methods we do not output any messages. I think may be it not expected to do it here.

src/otx/engine/engine.py
negvet: Added

src/otx/core/types/explain.py
negvet: I will add it in the very next PR. So let's keep it. To be honest, we do not have predictions as well yet

src/otx/core/config/explain.py
negvet: Added

GalyaZalesskaya (approved):

sovrasov (approved):

sungmanc (approved):

vinnamkim (approved):

PR #2825, opened 2024-01-22
Add PTQ optimization

sovrasov:
### Summary

This PR adds PTQ optimization to OV model
Also, all OV-related deps are updated

### Checklist

### License

```python
```


src/otx/core/model/entity/base.py
```diff
@@ -318,3 +347,89 @@ def _callback(result: NamedTuple, idx: int) -> None:
             outputs = [self.model(im) for im in numpy_inputs]
 
         return self._customize_outputs(outputs, inputs)
+
+    def optimize(
+        self,
+        output_dir: Path,
+        data_module: OTXDataModule,
+        ptq_config: dict[str, Any] | None = None,
```
vinnamkim, line 355: Should this be exposed? I can see that the user can give `max_data_subset_size` argument from `Engine`. If so, why don't we expose `max_data_subset_size` rather than `ptq_config`. I think that it would be more friendly to users who try to use this function.

sovrasov: If we don't expose this at some level, we won't allow users customize PTQ behavior from API. If a pre-defined quantization presets fail on specific data, user may refer to PTQ documentation and set the required parameters. `max_data_subset_size` was implicitly exposed only because it is set in Geti, not because it's crucially important for quantization. 
For instance, lightning follows the same approach with `.to_onnx` method: extra kwargs are forwarded to pytorch onnx converter.

sovrasov: > If we don't expose this at some level, we won't allow users customize PTQ behavior from API. If a pre-defined quantization presets fail on specific data, user may refer to PTQ documentation and set the required parameters. `max_data_subset_size` was implicitly exposed only because it is set in Geti, not because it's crucially important for quantization. For instance, lightning follows the same approach with `.to_onnx` method: extra kwargs are forwarded to pytorch onnx converter.

vinnamkim: > If we don't expose this at some level, we won't allow users customize PTQ behavior from API. If a pre-defined quantization presets fail on specific data, user may refer to PTQ documentation and set the required parameters. max_data_subset_size was implicitly exposed only because it is set in Geti, not because it's crucially important for quantization.

If so, can we improve this for better UX in the future such as
```python
    def optimize(
        self,
        output_dir: Path,
        data_module: OTXDataModule,
        (some curated configuration variables, e.g., max_data_subset_size),
        **extra_nncf_ptq_configs,
```

> For instance, lightning follows the same approach with .to_onnx method: extra kwargs are forwarded to pytorch onnx converter.

I believe [our product mission #2](https://github.com/openvinotoolkit/training_extensions/blob/v2/for_developers/product_design.md#our-product-mission) is not aim to expose everything possible to be configured to users. We need to limit our user's playground appropriately. If it is too vast, people can easily feel dizzy and escape from it. In terms of flexibility, it seems not a good idea to compete with PyTorch Lightning. There is no reason I think to use our tool rather than PyTorch Lightning if the user has to make the same effort.

vinnamkim (dismissed):

src/otx/core/model/entity/base.py
samet-akcay: it might also be an idea to expose these args with default values. So a user does not need to bother with changing them. At the same time, they can actually see those parameters there, without them being hidden in a config file. I think a config argument hides so much, and as a user that does not improve my experience.

sovrasov: The problem with PTQ configs is that they consist of multiple nested objects (mostly DTOs) declared in nncf package, and therefore we need to import them to expose to our API, or make converters on our side.
It's hard to say in advance which parameters are to be changed, and in that case exposing of rarely-used 3rd order parameters looks like overcomplicating the API.
Now we have only one parameter for 1% of advanced users (including us), others would just ignore it. Overall, I like the approach @samet-akcay is advocating. It works if we can select a reasonable set of parameters to expose.

sovrasov:
PR updated because of a merge conflict

harimkang (approved):

vinnamkim (approved):

src/otx/algo/segmentation/litehrnet.py
```diff
@@ -18,20 +18,362 @@ class LiteHRNet(MMSegCompatibleModel):
...
+                "activations_range_estimator_params": {
+                    "min": {"statistics_type": "QUANTILE", "aggregator_type": "MIN", "quantile_outlier_prob": 1e-4},
+                    "max": {"statistics_type": "QUANTILE", "aggregator_type": "MAX", "quantile_outlier_prob": 1e-4},
+                },
+            },
+        }
+        optim_config.update(ignored_scope)
+        return optim_config
+
+    def _obtain_ignored_scope(self) -> dict[str, Any]:
```
sungmanc, line 58: Just a question, why only `litehrnet` needs this complex ignored scope?

sungmanc (approved):

jaegukhyun (approved):

src/otx/algo/segmentation/litehrnet.py
sovrasov: HRNet contains a huge amount of operations and produces sparse representations, so PTQ fails to process it effectively

PR #2836, opened 2024-01-25
Add seed_everything into CLI side

harimkang:
### Summary

@jaegukhyun reported an issue that the seed was not being frozen, so I looked into it.
The seed is frozen after the OTXModel is created, so in some cases the seed was not being frozen.
OTXModel random feature: classification layer initialization

add fix seed function into the CLI.

### How to test

seed=0, deterministic=True, model: atss_mobilenetv2
Trial 1: val/map_50: 0.750 val/map_75: 0.502
Trial 2: val/map_50: 0.750 val/map_75: 0.502
Trial 3: val/map_50: 0.750 val/map_75: 0.502

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/cli/cli.py
+++ b/src/otx/cli/cli.py
@@ -102,6 +102,11 @@ class OTXCLI:
             help="Task Type.",
         )
         parser.add_argument(
+            "--seed",
+            type=int,
+            help="Sets seed for pseudo-random number generators in: pytorch, numpy, python.random.",
+        )
+        parser.add_argument(
             "--callback_monitor",
             type=str,
             help="The metric to monitor the model performance during training callbacks.",
@@ -307,6 +312,19 @@ class OTXCLI:
             skip_check=True,
         )
 
+    def set_seed(self) -> None:
+        """Set the random seed for reproducibility.
+
+        This method retrieves the seed value from the argparser and uses it to set the random seed.
+        If a seed value is provided, it will be used to set the random seed using the
+        `seed_everything` function from the `lightning` module.
+        """
+        seed = self.get_config_value(self.config, "seed", None)
+        if seed is not None:
+            from lightning import seed_everything
+
+            seed_everything(seed, workers=True)
+
     def run(self) -> None:
         """Executes the specified subcommand.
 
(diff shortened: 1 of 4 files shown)


vinnamkim (changes requested):
This will disable seed settings on the Python API side. Please find a way to maintain it rather than removing it.

harimkang:
> This will disable seed settings on the Python API side. Please find a way to maintain it rather than removing it.

Yes, I think that's a good point.
How do you think we should deliver on the API side?
I think there are three ways.
1. drive people to use `lightning.seed_everything` (-> For users who only use OTX, this may be awkward.)
2. provide the function in 1 as a function of otx like (-> It simply wraps the function that calls seed_everything once more in otx.)
3. take it as an engine's creation argument (in this case, if the user is already build a module, the seed might not be fixed in some cases).

I don't really have a preference and would love to hear your thoughts.

vinnamkim:
> > This will disable seed settings on the Python API side. Please find a way to maintain it rather than removing it.
> 
> Yes, I think that's a good point. How do you think we should deliver on the API side? I think there are three ways.
> 
> 1. drive people to use `lightning.seed_everything` (-> For users who only use OTX, this may be awkward.)
> 2. provide the function in 1 as a function of otx like (-> It simply wraps the function that calls seed_everything once more in otx.)
> 3. take it as an engine's creation argument (in this case, if the user is already build a module, the seed might not be fixed in some cases).
> 
> I don't really have a preference and would love to hear your thoughts.

I don't see any reason why the original code cannot setup the seed correctly. Is it because the `datamodule` already spawned the data pipeline workers before the engine creation?

harimkang:
> > > This will disable seed settings on the Python API side. Please find a way to maintain it rather than removing it.
> > 
> > 
> > Yes, I think that's a good point. How do you think we should deliver on the API side? I think there are three ways.
> > 
> > 1. drive people to use `lightning.seed_everything` (-> For users who only use OTX, this may be awkward.)
> > 2. provide the function in 1 as a function of otx like (-> It simply wraps the function that calls seed_everything once more in otx.)
> > 3. take it as an engine's creation argument (in this case, if the user is already build a module, the seed might not be fixed in some cases).
> > 
> > I don't really have a preference and would love to hear your thoughts.
> 
> I don't see any reason why the original code cannot setup the seed correctly. Is it because the `datamodule` already spawned the data pipeline workers before the engine creation?

Yes, that's right.

sungmanc (approved):

vinnamkim (approved):

jaegukhyun (approved):

PR #2845, opened 2024-01-29
Make objects pickleable

eunwoosh:
### Summary
This PR makes some objects pickleable. This is for preparing following HPO PR.
Except the change related to make pytorchcv model, others just change code structure so it doesn't change which code is executed.
Changes are as below:

- Code to deal with preprocessor error is moved to each file per tasks.
- `__reduce__()` is added in `MemCacheHandler`.
- `__reduce__()` is added in `OTXDataModule`.
- way to make pytorchcv model for detection is changed.

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/core/model/entity/utils/__init__.py
+++ b/src/otx/core/model/entity/utils/__init__.py
@@ -1,0 +1,4 @@
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
+"""Utility files for model entities used in OTX."""
--- a/src/otx/core/model/entity/utils/mmaction.py
+++ b/src/otx/core/model/entity/utils/mmaction.py
@@ -1,0 +1,39 @@
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
+"""Class definition for classification model entity used in OTX."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from mmaction.models.data_preprocessors import (
+    ActionDataPreprocessor as _ActionDataPreprocessor,
+)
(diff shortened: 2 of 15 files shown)


vinnamkim (approved):

src/otx/core/model/entity/utils/__init__.py
harimkang, line 1: This folder is named `utils`, but I'm not sure if that's the right name. 
I don't know if the name like `utils.mmaction.create_model` is appropriate, but I can't think of anything off the top of my head, so feel free to ignore it.

src/otx/core/model/entity/utils/mmaction.py
```diff
@@ -0,0 +1,55 @@
...
+        """Which device being used."""
+        try:
+            buf = next(self.buffers())
+        except StopIteration:
+            return super().device
+        else:
+            return buf.device
+
+
+def create_model(config: DictConfig, load_from: str | None) -> tuple[nn.Module, dict[str, dict[str, int]]]:
```
harimkang, line 44: The reason these functions can't be combined into one is because of mmX's import structure, right?

harimkang (commented):
Overall LGTM. I left a few comments.

sungmanc (approved):

src/otx/core/model/entity/utils/__init__.py
eunwoosh: yeah I just thought that it's utils for creating a model. Maybe we will be able to move it when we find better place.

src/otx/core/model/entity/utils/mmaction.py
eunwoosh: you're right. Each framework needs to use own `MODELS` registry.

harimkang (approved):

PR #2847, opened 2024-01-29
Change target ci-runner to gh-hosted runner for unittests

yunchu:
### Summary

To distribute CI workloads across all available resources, unit test for the v2 would be targeted to run on the GH-hosted runner.
For the integration tests, it required larger GPU memory (>15G) so it would keep using the dedicated ci-runner (3090) and it will be re-targeted to the runner on AWS whenever there are the GPU instances that have suitable memory size.

### Checklist

### License

```python
```


harimkang (dismissed):

sungmanc (dismissed):

vinnamkim (commented):
AFAIK, there is no unit test which needs a GPU device and I think there will not be as well in the future. So, do we need to run our unit test on the T4 node?

yunchu:
> AFAIK, there is no unit test which needs a GPU device and I think there will not be as well in the future. So, do we need to run our unit test on the T4 node?

If you can confirm that all unit tests are not required GPUs, it would be targeted to the Github hosted runner.

vinnamkim:
> > AFAIK, there is no unit test which needs a GPU device and I think there will not be as well in the future. So, do we need to run our unit test on the T4 node?
> 
> If you can confirm that all unit tests are not required GPUs, it would be targeted to the Github hosted runner.

```console
======================================================================== short test summary info ========================================================================
FAILED tests/unit/algo/segmentation/heads/test_class_incremental_mixin.py::TestClassIncrementalMixin::test_ignore_label - RuntimeError: Attempting to deserialize object on a CUDA device but torch.cuda.is_available() is False. If you are running on a CPU-only machine, please use torch.l...
FAILED tests/unit/algo/visual_prompting/encoders/test_sam_prompt_encoder.py::TestSAMPromptEncoder::test_get_device[cuda] - RuntimeError: No CUDA GPUs are available
======================================================== 2 failed, 370 passed, 1 skipped, 53 warnings in 38.90s =========================================================
(otx-v2) vinnamki@vinnamki:~/otx/training_extensions$ 
```
I found two cases. ~~I'll open a PR for them.~~ I'll stack commits on this PR.

harimkang (approved):

.github/workflows/pre_merge.yaml
```diff
@@ -32,7 +32,7 @@ jobs:
       - name: Code quality checks
         run: tox -vv -e pre-commit
   Unit-Test:
-    runs-on: [self-hosted, linux, x64, dev]
+    runs-on: ubuntu-20.04
```
sungmanc, line 35: Do we need to specify the ubuntu version ? If so, I prefer to 22.04

sungmanc (approved):

vinnamkim (approved):

.github/workflows/pre_merge.yaml
yunchu: It's just for choosing the Linux OS from the GH provide actions-runner. I think this is the first step to use GH-hosted runner for the unttests pipeline so that could be changed or added other version of OSes later.

PR #2848, opened 2024-01-29
🐞 Bugfix: Set reverse_input_channels to True in OpenVINO models

ashwinvaidya17:
### Summary

Fixes: CVS-129988

### Checklist

### License

```python
```


chuneuny-emily (approved):

sovrasov:
Do we need changes from https://github.com/openvinotoolkit/model_api/pull/152 in this PR?

ashwinvaidya17:
> Do we need changes from [openvinotoolkit/model_api#152](https://github.com/openvinotoolkit/model_api/pull/152) in this PR?

I don't think so as media pipe already passes RGB image to the inference model. But my understanding might be incorrect here.

sovrasov:
OTX also uses RGB for demos and OV tasks: https://github.com/openvinotoolkit/training_extensions/blob/releases/1.4.0/src/otx/api/entities/image.py#L99

ashwinvaidya17:
I'll leave that decision to the maintainers. I agree that OTX's OpenVINO inference is affected by the change in modelAPI but as of now since Geti does not use OTX for OpenVINO inference, the missing modelAPI upgrade should not affect Geti. I guess upgrading modelAPI version in OTX is more challenging given the current timeline of release. 
As for the actual issue, there were two. Turns out the RGB issue was not the actual reason for 0% score on normal images. That was probably a coincidental discovery during the hunt. The actual reason was that normal images also reported anomaly scores. So a good image would report anomaly score of 0. The solution implemented in modelAPI currently flips the score reported when the predicted label is Normal. So the users now see normal score of 100%

sovrasov:
> I'll leave that decision to the maintainers. I agree that OTX's OpenVINO inference is affected by the change in modelAPI but as of now since Geti does not use OTX for OpenVINO inference, the missing modelAPI upgrade should not affect Geti. I guess upgrading modelAPI version in OTX is more challenging given the current timeline of release. As for the actual issue, there were two. Turns out the RGB issue was not the actual reason for 0% score on normal images. That was probably a coincidental discovery during the hunt. The actual reason was that normal images also reported anomaly scores. So a good image would report anomaly score of 0. The solution implemented in modelAPI currently flips the score reported when the predicted label is Normal. So the users now see normal score of 100%

I see, according to the discussion in jira, let's avoid upgrading MAPI here. To be precise, Geti still uses OTX inference tasks for batch inference (see Ludo's comment in CVS-130051). If now in OTX 1.4 anomaly inference task works somehow (probably because it doesn't use MAPI), I'm ok with this solution.

sovrasov (approved):

ashwinvaidya17:
> Geti still uses OTX inference tasks for batch inference (see Ludo's comment in CVS-130051)

Hmm I wasn't aware. I'll have a closer look at this task later. For now I am merging this. It does not affect the 0% issue but it closes one loose end.

PR #2857, opened 2024-01-31
Introduce warm-up and change the LRPlateau monitor value, Enable Early stopping

sungmanc:
## What this PR introduces
1. Implemented the Warm-up Scheduler. It could be implemented by inheriting `BaseWarmupScheduler`.
2. Enable `Earlystop` by reducing the patience value. It'll  monitor the validation metric to stop the training at the proper time.
3. Update the `LRSchedule`, especially for the `PlateauLRSchedule`, to reduce the learning rate by monitoring the validation metric. 
4. The patience values of `Earlystop` and `LRSchedule` are modified according to the `AdaptiveInterval`.

## Issues
Even though the warm-up logic is successfully working, the tensorboard showing the wrong information.

If you check the tensorboard log, the LR showed the 0.0049 at the 0 step. However, when I checked the real value, it had no problem (4.9e-5).  It seems that it could happen due to the timing of the monitoring logic.

image
image

## Note
Tuning the values of each patiences, warm-up needs time. So, I'll optimize the those values and make separate PR.
Classification scores after this PR,
image

### Checklist

### License

```python
```


src/otx/core/model/module/base.py
vinnamkim, line 1: It is not desired to introduce custom codes to `LitModule` only for your custom lr scheduler. Please find another way such as combining two lr scheduler at the configuration level, https://github.com/Lightning-AI/pytorch-lightning/issues/1037#issuecomment-650618255

src/otx/algo/callbacks/adaptive_train_scheduling.py
```diff
@@ -78,3 +83,36 @@ def on_train_end(self, trainer: Trainer, pl_module: LightningModule) -> None:
...
+        should be changed according to the adaptive interval.
+        """
+        for config in lr_configs:
+            if hasattr(config, "frequency"):
+                msg = (
+                    "The frequency of LRscheduler will be changed due to the effect of adaptive interval: "
+                    f"{config.frequency} --> {adaptive_interval}."
+                )
+                log.warning(msg)
+                config.frequency = adaptive_interval
```
vinnamkim, line 115:
```suggestion
         self.revert_frequency = []
         ...
                saved_frequency = config.frequency
                config.frequency = adaptive_interval

                def _revert(config, saved_frequency):
                    config.frequency = saved_frequency
                self.revert_frequency += [functools.partial(_revert, config, saved_frequency)]

...

    def on_train_end(self, trainer: Trainer, pl_module: LightningModule) -> None:
        for revert in self.revert_frequency:
            revert()
```

vinnamkim: To maintain the same behavior as the validation interval, please add the revert action at `on_train_end()`

src/otx/algo/callbacks/adaptive_train_scheduling.py
```diff
@@ -32,6 +33,7 @@ def __init__(self, max_interval: int = 5, decay: float = -0.025):
         self.decay = decay
         self._saved_check_val_every_n_epoch: int | None = None
         self._saved_log_every_n_steps: int | None = None
+        self.is_enabled: bool = False
```
vinnamkim: Is this needed?

src/otx/core/model/module/detection.py
```diff
@@ -141,4 +141,4 @@ def test_step(self, inputs: DetBatchDataEntity, batch_idx: int) -> None:
     @property
     def lr_scheduler_monitor_key(self) -> str:
         """Metric name that the learning rate scheduler monitor."""
-        return "train/loss"
+        return "val/map_50"
```
harimkang, line 144: Maybe we need to change the recipe of the detection models as well?

sungmanc: For the detection recipes, I'll make a separate PR

goodsong81:
Not specific to this PR, but why don't we consider to set the [`min_delta` parameter](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.callbacks.EarlyStopping.html#earlystopping:~:text=to%20be%20monitored.-,min_delta,-(float)%20%E2%80%93%20minimum) (e.g. 0.001) for metric monitoring in early stoppping and learning rate plateau?
We've long been observing the delayed early(?) stop due to sporadic meaningless improvements like 0.00001.
Of course, this needs accuracy / efficiency validation.

src/otx/core/model/module/detection.py
harimkang: Is the reason for the separation an experiment? If so, I think this code change should be included in that PR so it's not confusing. What do you think?

sungmanc: > I think this code change should be included in that PR so it's not confusing. What do you think?

Agree. 

> Is the reason for the separation an experiment?

Since I need to do lots of experiments, I want to take step by step manner.

src/otx/core/model/module/base.py
sungmanc: If we follows the way what you linked, we need to modify the function located at the `LitModule` since we need to change `configure_optimizers` function. AFAIK, we couldn't input the list of scheduler configuration.
https://github.com/Lightning-AI/pytorch-lightning/issues/328#issuecomment-717085681

And I think this way is also guided from the lightning github. 
https://github.com/Lightning-AI/pytorch-lightning/blob/34a34a0754d612965b158f1f588336507554fa16/src/lightning/pytorch/core/module.py#L1281

So, I want to preserve the current method at now. And if we really need to change, I want to do that at the optimization phase.

sungmanc:
> Not specific to this PR, but why don't we consider to set the [`min_delta` parameter](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.callbacks.EarlyStopping.html#earlystopping:~:text=to%20be%20monitored.-,min_delta,-(float)%20%E2%80%93%20minimum) (e.g. 0.001) for metric monitoring in early stoppping and learning rate plateau? We've long been observing the delayed early(?) stop due to sporadic meaningless improvements like 0.00001. Of course, this needs accuracy / efficiency validation.

Sounds good, I initially added the min_delta=0.001 at now, I'll update this value after experiments. 
https://github.com/openvinotoolkit/training_extensions/pull/2857/commits/599d5a4027237be8d27e64254cb5548b63af109a

For the LRscheduler, the default threshold is set to 1e-4. I'll also check the effect of this value at the later phase.
https://github.com/openvinotoolkit/training_extensions/blob/599d5a4027237be8d27e64254cb5548b63af109a/src/otx/algo/schedulers/warmup_schedulers.py#L51

src/otx/algo/callbacks/adaptive_train_scheduling.py
sungmanc: Fixed, https://github.com/openvinotoolkit/training_extensions/pull/2857/commits/4a6cd2ba0bb1e526bd1c181b56a1db0c076685ee

src/otx/algo/callbacks/adaptive_train_scheduling.py
sungmanc: Done, https://github.com/openvinotoolkit/training_extensions/pull/2857/commits/7803e5ac6b1e0e8b3199f34ab5155b998bd76bdb

jaegukhyun (dismissed):

src/otx/recipe/_base_/train.yaml
```diff
@@ -5,8 +5,9 @@ callbacks:
     init_args:
       monitor: null
       mode: max
-      patience: 100
+      patience: 10
```
harimkang: Shouldn't we also raise `max_epochs`?

src/otx/core/model/module/base.py
sungmanc: Reflect all of your comments, thanks for supplying, https://github.com/openvinotoolkit/training_extensions/pull/2857/commits/a60d449d6f86b78b64dbc8a904b2c0dfd1a0ce7f

src/otx/recipe/_base_/train.yaml
sungmanc: At the regression tests, we've already set the proper epochs. However, set the proper epochs as a default is important. However, it needs a branch for different tasks. Let's consider this at the later PR.

vinnamkim (approved):
LGTM but please create a backlog ticket to clean up all these things by enabling `OTXLitModule` to accept a list of LR schedulers.

harimkang (approved):

jaegukhyun (approved):

PR #2859, opened 2024-01-31
Add PTQ optimization configs via embedding it into IR

kprokofi:
### Checklist

### License

```python
```


src/otx/core/model/entity/base.py
```diff
@@ -229,21 +231,22 @@ def _export_parameters(self) -> dict[str, Any]:
...
         parameters = {}
-
         all_labels = ""
         all_label_ids = ""
         for lbl in self.label_info.label_names:
             all_labels += lbl.replace(" ", "_") + " "
             all_label_ids += lbl.replace(" ", "_") + " "
 
+        # not every model requires ptq_config
+        ptq_config = self.ptq_config if hasattr(self, "ptq_config") else ""
```
sovrasov: I'd make the config empty by default instead of checking attributes here.
Also, ptq_config can be renamed into optimization_config, since we refer the whole process as "optimization"

kprokofi: Do you propose to add empty PTQ config for all the models? Like adding it as property to OTXModel?

src/otx/core/model/entity/base.py
```diff
@@ -397,3 +404,42 @@ def __iter__(self) -> Iterator:
...
+
+        Example:
+            ptq_config = _generate_ptq_config(ov_model)
+        """
+        from nncf import IgnoredScope
+        from nncf.common.quantization.structs import QuantizationPreset
+        from nncf.parameters import ModelType
+        from nncf.quantization.advanced_parameters import AdvancedQuantizationParameters
+
+        initial_ptq_config = json.loads(ov_model.rt_info["model_info"]["ptq_config"].value)
```
sovrasov: What if ["model_info"]["ptq_config"] is missing? We need a check here

src/otx/core/model/entity/base.py
```diff
@@ -397,3 +404,42 @@ def __iter__(self) -> Iterator:
         openvino.save_model(compressed_model, output_model_path)
 
         return output_model_path
+
+    def _generate_ptq_config(self, ov_model: Model) -> dict:
```
sovrasov: This function rather reads PTQ config from IR, although, in a bit sophisticated way

src/otx/core/model/entity/base.py
```diff
@@ -5,17 +5,19 @@
 
 from __future__ import annotations
 
+import json
 import warnings
 from abc import abstractmethod
 from typing import TYPE_CHECKING, Any, Generic, Iterable, Iterator, NamedTuple
 
+import nncf
```
vinnamkim: Is this needed? I don't want to import this at the top level because in my old experiece, importing `nncf` was heavy (it can be wrong).
```suggestion
```

src/otx/core/model/entity/base.py
vinnamkim: > I'd make the config empty by default instead of checking attributes here.
Also, ptq_config can be renamed into optimization_config, since we refer the whole process as "optimization"

Thanks @sovrasov, that's what I want to say.

kprokofi: Sure

src/otx/core/model/entity/base.py
kprokofi: Ok. Will move it to optimize()

src/otx/core/model/entity/base.py
sovrasov: Yes, since this config is common, we can define it at the top of models hierarchy

src/otx/core/model/entity/base.py
sovrasov: -> dict[str, Any]

kprokofi:
tests for optimize - infer - export passed for segmentation and Deit Tiny classification

PR #2864, opened 2024-02-01
Copy HPO algorithm into OTX 2.0

eunwoosh:
### Summary
This PR copies HPO algorithm code into OTX 2.0
It doesn't include code using HPO but just HPO algorithm, so it doesn't affect current OTX 2.0 at all.

### Checklist

### License

```python
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -283,6 +283,7 @@ max-returns = 10
     "PT004",    # fixture ignore type returning.
     "E501",     # Test skips lines that are too long.
     "ANN001",   # Skip annotation type hint in test codes
+    "ANN201",   # Skip return type hint in test codes
     "D",     # Test skips missing docstring argument with magic (fixture) methods.
 ]
 "src/otx/**/*.py" = [
--- a/src/otx/utils/utils.py
+++ b/src/otx/utils/utils.py
@@ -1,6 +1,6 @@
-# Copyright (C) 2023 Intel Corporation
+# Copyright (C) 2024 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
-#
+
 """OTX utility functions."""
 
 import signal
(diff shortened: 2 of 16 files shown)


src/otx/utils/utils.py
```diff
@@ -0,0 +1,28 @@
+# Copyright (C) 2023 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
```
harimkang:
```suggestion
# Copyright (C) 2024 Intel Corporation
# SPDX-License-Identifier: Apache-2.0

```

harimkang (approved):

pyproject.toml
```diff
@@ -283,6 +283,7 @@ max-returns = 10
     "PT004",    # fixture ignore type returning.
     "E501",     # Test skips lines that are too long.
     "ANN001",   # Skip annotation type hint in test codes
+    "ANN201",   # Skip return type hint in test codes
```
sungmanc, line 286: This might need the concensus since we've been implementing the test code with the type hint. @harimkang, @sovrasov  what do you think ?

sungmanc: @vinnamkim , @samet-akcay

jaegukhyun (approved):

pyproject.toml
harimkang: It would be nice to have a consensus for the test code as well, but I don't know if it's necessary for the test code. Also, the return type is mostly None in test code, so I think it's okay to ignore it.

eunwoosh: yes @harimkang is right. Number of HPO unit test function is hundreds and honestly, I don't think adding `-> None` to them is meaningful.

sungmanc: For the functions, I agree. However, I think for the variable or argument, type-hint is helpful to understand what variable means for me.

eunwoosh: yes but what I added is about return type. if we remove `ANN001`, we may need to change overall test code. I think it's better to discuss about it later.

sungmanc: Ah, sorry, I just missed the information. it just represents the return_type.

sungmanc (approved):

PR #2866, opened 2024-02-01
Transparent forwarding of deterministic parameter to trainer

sovrasov:
### Summary
Some models don't support strict deterministic mode, which causes exception when executing with deterministic True.
"warn" mode allows to avoid this, and possibly improve reproducibility for those models, while not touching others.

### Checklist

### License

```python
```


src/otx/engine/engine.py
```diff
@@ -346,7 +348,7 @@ def predict(
         return self.trainer.predict(
             model=lit_module,
             datamodule=datamodule if datamodule is not None else self.datamodule,
-            ckpt_path=str(checkpoint) if checkpoint is not None else self.checkpoint,
+            ckpt_path=str(checkpoint) if checkpoint is not None else str(self.checkpoint),
```
harimkang: Won't this cause problems if both `checkpoint` and `self.checkpoint` are `None`?

vinnamkim (dismissed):

src/otx/engine/engine.py
jaegukhyun: Also curious about why we need this change

sovrasov: It's just to satisfy type checker, but I can revert this. `trainer.predict` takes a string (by annotation), I don't know what happens if we pass a Path object. Usually, `checkpoint` is not None, so we just don't hit the second part of if. I think it'd be a file reading failure.

negvet (approved):

kprokofi (approved):

GalyaZalesskaya (approved):

PR #2823, opened 2024-01-22
Add data filtering 

jaegukhyun:
### Summary
Same as title.
Bring data filtering in OTX1.5's datumaro adapter

### Checklist

- [ ] I have added unit tests to cover my changes.​
- [ ] I have added integration tests to cover my changes.​
- [ ] I have added e2e tests for validation.
- [ ] I have added the description of my changes into CHANGELOG in my target branch (e.g., [CHANGELOG](https://github.com/openvinotoolkit/training_extensions/blob/develop/CHANGELOG.md) in develop).​
- [ ] I have updated the documentation in my target branch accordingly (e.g., [documentation](https://github.com/openvinotoolkit/training_extensions/tree/develop/docs) in develop).
- [ ] I have [linked related issues](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

### License

- [ ] I submit _my code changes_ under the same [Apache License](https://github.com/openvinotoolkit/training_extensions/blob/develop/LICENSE) that covers the project.
      Feel free to contact the maintainers if that's a concern.
- [ ] I have updated the license header for each file (see an example below).

```python
# Copyright (C) 2023 Intel Corporation
# SPDX-License-Identifier: Apache-2.0
```

PR #2867, opened 2024-02-01
Tiling Speed & Accuracy Enhancement

eugene123tw:
### Summary

* **Optimization and Bug Fix:**
  * Implemented various optimizations in tiling.
  * Addressed a memory cache bug.

* **RLE Encoder Integration:**
  * Added an RLE encoder to compute RLE directly from torch.Tensor.
  * Reduced processing time by eliminating the need for torch -> numpy -> mask_utils.encode.

* **Customized mAP Calculation:**
  * Introduced a customized OTXInstSegMeanAveragePrecision.
  * Streamlined mAP calculation directly from RLE, avoiding the slower torch -> numpy -> mask_utils.encode process.

* **Tile Merge Enhancement:**
  * Modified the tile merge process from Datumaro, transitioning from np array-based merging to direct merging of predictions from torch.Tensor.

* **Tile Recipes for Detection and Instance Segmentation:**
  * Added various tile recipes to enhance performance in both detection and instance segmentation tasks.

### Checklist

### License

```python
```


sovrasov:
@eugene123tw thanks for your effort in optimizing the tiling pipeline. 
I have a question related to backward compatibility: perfectly if we can achieve the same metrics when loading a snapshot from 1.x (which would be quite common scenario in Geti), so my question is do we have such backward compatibility (at least up to input preprocessing)? 
The second point what's already waiting around the corner is compatibility with MAPI. We have to provide it as we hit integration to Geti milestone. Tiler from 1.x was implemented in python/c++ versions, and de-facto became a standard for Geti's inference server based on MediaPipe. For now, I haven't heard about postrpocessing versioning there, which means we have to make 100% compatible tile merging in OTX 2.0 torch implementation.

src/otx/core/data/dataset/base.py
```diff
@@ -159,7 +159,7 @@ def _get_img_data_and_shape(self, img: Image) -> tuple[np.ndarray, tuple[int, in
         key = img.path if isinstance(img, ImageFromFile) else id(img)
 
         if (img_data := self.mem_cache_handler.get(key=key)[0]) is not None:
-            return img_data, img_data.shape[:2]
+            return img_data.astype(np.float32), img_data.shape[:2]
```
vinnamkim: What's the reason for this change? It is not desired to change the dtype from `uint8` to `float32`. This is because it affects severely to the following image transformation performance.

src/otx/algo/instance_segmentation/otx_instseg_evaluation.py
```diff
@@ -0,0 +1,67 @@
...
+from __future__ import annotations
+
+from typing import Any
+
+import pycocotools.mask as mask_utils
+import torch
+from torchmetrics.detection.mean_ap import MeanAveragePrecision
+
+
+class OTXInstSegMeanAveragePrecision(MeanAveragePrecision):
```
vinnamkim: Would you please add a unit test which can verify that this metric class can produce the same outputs as `torchmetrics.detection.mean_ap.MeanAveragePrecision`?

src/otx/core/utils/mask_util.py
```diff
@@ -27,4 +33,43 @@ def polygon_to_bitmap(polygons: list[Polygon], height: int, width: int) -> np.nd
     """
     polygons = [polygon.points for polygon in polygons]
     rles = mask_utils.frPyObjects(polygons, height, width)
+    if return_rle:
+        return rles
     return mask_utils.decode(rles).astype(bool).transpose((2, 0, 1))
+
+
+def encode_rle(mask: torch.Tensor) -> dict:
```
vinnamkim, line 57: Would you please add an unit test to validate whether this function is equivalent to `pycocotools.encode()`?

src/otx/core/utils/mask_util.py
```diff
@@ -8,13 +8,19 @@
...
     from datumaro import Polygon
 
 
-def polygon_to_bitmap(polygons: list[Polygon], height: int, width: int) -> np.ndarray:
+def polygon_to_bitmap(
+    polygons: list[Polygon],
+    height: int,
+    width: int,
+    return_rle: bool = False,
+) -> np.ndarray:
```
vinnamkim: I guess that this can be separated into two functions for clarity and performance
```suggestion
def polygon_to_rle(
    polygons: list[Polygon],
    height: int,
    width: int
) -> np.ndarray:
    polygons = [polygon.points for polygon in polygons]
    return mask_utils.frPyObjects(polygons, height, width)

def polygon_to_bitmap(
    polygons: list[Polygon],
    height: int,
    width: int
) -> np.ndarray:
    canvas = np.zeros([height, width], dtype=np.uint8)
    return drawing_masks(canvas, polygons)
```

src/otx/core/model/module/instance_segmentation.py
```diff
@@ -102,15 +103,27 @@ def validation_step(self, inputs: InstanceSegBatchDataEntity, batch_idx: int) ->
             raise TypeError(preds)
 
         self.val_metric.update(
-            **self._convert_pred_entity_to_compute_metric(preds, inputs),
+            **self._convert_pred_entity_to_compute_metric(batch_idx, preds, inputs),
         )
 
     def _convert_pred_entity_to_compute_metric(
         self,
+        index: int,
```
vinnamkim: Is this needed?

src/otx/core/data/dataset/tile.py
```diff
@@ -105,6 +105,44 @@ def _get_item_impl(self, index: int) -> OTXDataEntity | None:
...
+        Args:
+            image (np.ndarray): The input image.
+            item (DatasetItem): The dataset item.
+
+        Returns:
+            A tuple containing two lists:
+            - tile_entities (list[OTXDataEntity]): List of tile entities.
+            - tile_attrs (list[dict]): List of tile attributes.
+        """
+        tile_ds = DmDataset.from_iterable([item])
```
vinnamkim, line 125: I'm curious why you did apply this tile transform just to `[item]`. Is it better to apply it to `self._dataset` itself? You might also try to stack this filtering on top of it to exclude tiled images with no annotation https://github.com/openvinotoolkit/training_extensions/pull/2823. If it is not easy to handle this PR, can we create a backlog ticket for this?

src/otx/algo/instance_segmentation/otx_instseg_evaluation.py
```diff
@@ -0,0 +1,67 @@
...
+        Returns:
+            RLE encoded masks
+        """
+        if "segm" in self.iou_type:
+            masks = []
+            for rle in item["masks"]:
+                if isinstance(rle["counts"], list):
+                    rle["counts"] = mask_utils.frPyObjects(rle, *rle["size"])["counts"]
+                masks.append((tuple(rle["size"]), rle["counts"]))
+        return None, tuple(masks)
```
sungmanc, line 67: Do we need to return the None ? As following the description, returning only RLE encoded mask would be better

src/otx/core/data/dataset/base.py
eugene123tw: Thanks for the comment. I revert the change.

src/otx/core/utils/mask_util.py
eugene123tw: Thanks for the comment. I added the suggestion accordingly.

src/otx/core/model/module/instance_segmentation.py
eugene123tw: No. Removed.

src/otx/core/data/dataset/tile.py
eugene123tw: @vinnamkim 
1. Datumaro converts the initial image dataset, including full image annotations, into a tiled dataset with tiled annotations (tile annos are not used in evaluation). After tile transformation, it's not possible to get the original annotations.

2. In the testing phase, filtering should not take place as it's not possible to know which tiles have annotations and which do not.

src/otx/core/utils/mask_util.py
eugene123tw: Added in `tests/unit/core/utils/test_mask_utils.py`

src/otx/algo/instance_segmentation/otx_instseg_evaluation.py
eugene123tw: Added in `tests/unit/algo/instance_segmentation/test_evaluation.py`

eugene123tw:
> @eugene123tw thanks for your effort in optimizing the tiling pipeline. I have a question related to backward compatibility: perfectly if we can achieve the same metrics when loading a snapshot from 1.x (which would be quite common scenario in Geti), so my question is do we have such backward compatibility (at least up to input preprocessing)? The second point what's already waiting around the corner is compatibility with MAPI. We have to provide it as we hit integration to Geti milestone. Tiler from 1.x was implemented in python/c++ versions, and de-facto became a standard for Geti's inference server based on MediaPipe. For now, I haven't heard about postrpocessing versioning there, which means we have to make 100% compatible tile merging in OTX 2.0 torch implementation.


Hi @sovrasov, appreciate your feedback. Regarding the first question, the straightforward answer is currently not, but I'm working on a new PR to address this gap. To tackle your concern, I'll focus on bridging the inconsistency between 1.5 tiling and 2.x tiling. I plan to conduct a test following your suggestion, which involves acquiring a tiling model, evaluating its performance in 1.5, attempting to load it into 2.x, and verifying if the accuracy aligns.

As for the second question, the objective remains consistent with OTX 1.5. We will implement a regression test in 2.0 to ensure that OV tiling achieves the same accuracy as PyTorch tiling. However, before implementing the test, I'll work on narrowing the gap between versions 1.5 and 2.0.

Currently, there are three tickets associated with this matter, tracked under https://jira.devtools.intel.com/browse/CVS-131551.

sungmanc (approved):

src/otx/core/data/dataset/tile.py
vinnamkim: Ok let me take this problem in the future.

vinnamkim (approved):
LGTM.

harimkang (approved):

jaegukhyun (approved):

src/otx/algo/instance_segmentation/otx_instseg_evaluation.py
eugene123tw: @sungmanc It's not possible to only return encoded masks as the method is inherited from torchmetrics mAP here: https://github.com/Lightning-AI/torchmetrics/blob/v1.3.0.post0/src/torchmetrics/detection/mean_ap.py#L815 as it unfolds the tuple here: https://github.com/Lightning-AI/torchmetrics/blob/v1.3.0.post0/src/torchmetrics/detection/mean_ap.py#L492

PR #2870, opened 2024-02-02
Add XAI hook for Mask-RCNN

GalyaZalesskaya:
### Summary

Added algorithm for Mask RCNN saliency generation using predicted `masks` in model's outputs. 

Added dummy `MaskRCNNRecordingForwardHook` to save logic with hooks for all XAI algorithms, but the main saliency map generation is happening in `MaskRCNNRecordingForwardHook.get_sal_map_from_preds` class method

TODO for future PRs:
- Current masks from Mask RCNN are wrong and the reason needs to be investigated

### Checklist

### License

```python
```


src/otx/algo/hooks/recording_forward_hook.py
```diff
@@ -364,3 +366,56 @@ def func(
...
+            preds (List[InstanceSegBatchPredEntity]): Predictions of Instance Segmentation model.
+            num_classes (int): Num classes that model can predict.
+
+        Returns:
+            torch.Tensor: Class-wise Saliency Maps. One saliency map per each class - [batch, class_id, H, W]
+        """
+        # TODO(gzalessk): Add unit tests # noqa: TD003
+        batch_size = len(preds)
+        _, height, width = preds[0].masks[0].data.shape
+        batch_saliency_maps = torch.empty((batch_size, num_classes, height, width))
```
vinnamkim: There is no guarantee that `preds.masks` has the same dimensions among themselves. How about revising this this function to take `InstanceSegBatchPredEntity` instead?

src/otx/algo/utils/xai_utils.py
```diff
@@ -0,0 +1,24 @@
...
+    predictions: list | None,
+) -> list:
+    """Implement saliency map filtering and post-processing."""
+    if predictions is not None and isinstance(predictions[0], InstanceSegBatchPredEntity):
+        # Mask-RCNN case, receive saliency maps from predictions
+        raw_saliency_maps = MaskRCNNRecordingForwardHook.get_sal_map_from_preds(
+            predictions,
+            explain_config.num_classes,
+        )
+    return raw_saliency_maps
```
vinnamkim: This seems coming your explanations,
> Added dummy MaskRCNNRecordingForwardHook to save logic with hooks for all XAI algorithms, but the main saliency map generation is happening in MaskRCNNRecordingForwardHook.get_sal_map_from_preds class method

I think that things are going bad to keep momentum of using hooks for XAI. Can you add a TODO comment such as 
```
TODO: It is a temporary workaround. this function will be removed after we refactor XAI logics into `OTXModel.forward_explain()`.
```

src/otx/algo/hooks/recording_forward_hook.py
```diff
@@ -364,3 +366,56 @@ def func(
...
+    def create_and_register_hook(cls) -> BaseRecordingForwardHook:
+        """Create this object and register it to the module forward hook."""
+        return cls()
+
+    def func(self, feature_map: torch.Tensor | Sequence[torch.Tensor], _: int = -1) -> None:
+        """Return None for saliency map hook to generate saliency maps from predictions."""
+        return
+
+    @classmethod
+    def get_sal_map_from_preds(cls, preds: list[InstanceSegBatchPredEntity], num_classes: int) -> torch.Tensor:
```
negvet: In all hooks, `func` is used to generate saliency maps from some input. Should we also preserve this pattern for `MaskRCNNRecordingForwardHook`. The difference would be that `func` will get as input predictions instead of the feature maps, but to some extend feature maps are also prediction of some part of the model - so it seem to follow the same logic. What do you think?

src/otx/algo/utils/xai_utils.py
```diff
@@ -0,0 +1,24 @@
...
+    raw_saliency_maps: list,
+    explain_config: ExplainConfig,
+    predictions: list | None,
+) -> list:
+    """Implement saliency map filtering and post-processing."""
+    if predictions is not None and isinstance(predictions[0], InstanceSegBatchPredEntity):
+        # Mask-RCNN case, receive saliency maps from predictions
+        raw_saliency_maps = MaskRCNNRecordingForwardHook.get_sal_map_from_preds(
+            predictions,
+            explain_config.num_classes,
```
negvet: `explain_config` is not aware about `num_classes`, at least for now. `create_and_register_hook` might take `num_classes` as input.

GalyaZalesskaya: Yes, but this way we have to pass `self.trainer.model.model.explain_hook` into `get_processed_saliency_maps` function to be able to get `num_classes`. Is it okay to do so?

GalyaZalesskaya: Fixed

src/otx/algo/hooks/recording_forward_hook.py
GalyaZalesskaya: Thank you, fixed

src/otx/algo/utils/xai_utils.py
GalyaZalesskaya: Done, thank you

src/otx/algo/hooks/recording_forward_hook.py
GalyaZalesskaya: Great point, thank you

src/otx/engine/engine.py
```diff
@@ -514,12 +514,15 @@ def explain(
             ckpt_path=ckpt_path,
         )
 
+        explain_hook = self.trainer.model.model.explain_hook
         # Optimize for memory <- TODO(negvet)
-        raw_saliency_maps = self.trainer.model.model.explain_hook.records
+        raw_saliency_maps = explain_hook.records
+
         return get_processed_saliency_maps(
             raw_saliency_maps,
```
negvet: In this case we can get `raw_saliency_maps` inside of `get_processed_saliency_maps`, don't we?

src/otx/core/config/explain.py
```diff
@@ -15,3 +15,4 @@ class ExplainConfig:
 
     target_explain_group: TargetExplainGroup = TargetExplainGroup.ALL
     postprocess: bool = False
+    num_classes: int = 0
```
negvet: Is it still needed?

GalyaZalesskaya: Removed

negvet:
You will add tests when mask-rcnn predictions will be fixed?

src/otx/engine/engine.py
GalyaZalesskaya: Good point, moved. Didn't add if statement; 
```
if `MaskRCNN`:
  explain_hook.func(predictions)
else:
  explain_hook.records
```
to make code more simple, since `explain_hook.records` is preferable path

GalyaZalesskaya:
> You will add tests when mask-rcnn predictions will be fixed?

I wanted to add unit tests with simple shape checks in a separate PR , but I can do it now, it's even better

negvet (approved):

vinnamkim (approved):

sovrasov (approved):

PR #2897, opened 2024-02-07
Add Auto-Configuration for OVModel

harimkang:
### Summary

I was writing an API example in E2E and noticed something that bothered me.
This is the case of testing the IR Model.
I think Train -> Test -> Export -> Test is a very natural OTX user workflow.
But if we look at the current API example, it looks like this
```python
from otx.engine import Engine

config = "src/otx/recipe/classification/multi_class_cls/otx_efficientnet_b0.yaml"
data_root = "tests/assets/classification_dataset"

engine = Engine.from_config(
    config_path=config,
    data_root=data_root,
)
engine.train(max_epochs=2)
engine.test()
exported_model_path = engine.export()
```
Current Way for testing with IR model.
```python
# Current Way for Testing IR checkpoint
from otx.core.model.entity.classification import OVMulticlassClassificationModel

# User Need to create OVMulticlassClassificationModel for OV testing
ov_model = OVMulticlassClassificationModel(
    model_name=exported_model_path,
    num_classes=engine.datamodule.meta_info.num_classes,
    model_type="Classification",
)

# Need to create another engine for OV test
ov_engine = Engine(
    data_root=data_root,
    model=ov_model,
)

ov_engine.test()
```
In the above case, it can be inconvenient to have to build an OV model for IR and create a new Engine.
It's also not natural flow for me.
I found an idea to improve this with auto-configuration.
We can do below example.
```python
exported_model_path = engine.export()
engine.test(checkpoint=exported_model_path)
```
Of course, both of the above cases are possible.

CLI Usage
```shell
# Training
>>> otx train --config src/otx/recipe/detection/atss_mobilenetv2.yaml --data_root tests/assets/car_tree_bug --max_epochs 10 --seed 3

# Testing with torch
>>> otx test --config src/otx/recipe/detection/atss_mobilenetv2.yaml --data_root tests/assets/car_tree_bug --checkpoint otx-workspace/checkpoints/epoch_004.ckpt

# Export to IR
>>> otx export --config src/otx/recipe/detection/atss_mobilenetv2.yaml --data_root tests/assets/car_tree_bug --checkpoint otx-workspace/checkpoints/epoch_004.ckpt

# Testing with IR Model
>>> otx test --config src/otx/recipe/detection/atss_mobilenetv2.yaml --data_root tests/assets/car_tree_bug --engine.device cpu --engine.work_dir ./otx-workspace/atss-mobilenet --checkpoint otx-workspace/exported_model.xml
```

### Checklist

### License

```python
```


for_developers/example.ipynb
harimkang, line 1: I'm going to remove this file and switch to the API examples notebook file after discussion with Vlad.

src/otx/engine/engine.py
```diff
@@ -276,19 +276,24 @@ def test(
                 otx test --config <CONFIG_PATH, str> --checkpoint <CKPT_PATH, str>
                 ```
         """
+        model = self.model
+        checkpoint = checkpoint if checkpoint is not None else self.checkpoint
+        datamodule = datamodule if datamodule is not None else self.datamodule
+
+        is_ir_ckpt = str(checkpoint).endswith(".xml")
```
```suggestion
        is_ir_ckpt = checkpoint.suffix == "xml"
```
Minor one, but `Path` can give suffix.

src/otx/engine/engine.py
```diff
@@ -276,19 +276,24 @@ def test(
...
+        if is_ir_ckpt and not self.model.__class__.__name__.startswith("OV"):
```
```suggestion
        if is_ir_ckpt and not isinstance(self.model, OVModel):
```

src/otx/engine/utils/auto_configurator.py
```diff
@@ -257,3 +268,28 @@ def get_scheduler(self) -> LRSchedulerCallable | None:
         scheduler_config = self.config.get("scheduler", None)
         logger.warning(f"Set Default Scheduler: {scheduler_config}")
         return partial_instantiate_class(init=scheduler_config)
+
+    def get_ov_model(self, model_name: str, meta_info: LabelInfo) -> OVModel:
```
eunwoosh, line 272: How about getting num_classes instead of `meta_info`?

src/otx/engine/utils/auto_configurator.py
```diff
@@ -257,3 +268,28 @@ def get_scheduler(self) -> LRSchedulerCallable | None:
...
+
+        Returns:
+            OVModel: The OVModel instance.
+
+        Raises:
+            NotImplementedError: If the OVModel for the given task is not supported.
+        """
+        class_path = OVMODEL_PER_TASK.get(self.task, None)
+        if class_path is None:
+            msg = f"{self.task} is not support OVModel."
```
eunwoosh, line 287:
```suggestion
            msg = f"{self.task} doesn't support OVModel."
```

eunwoosh (commented):
Thanks for your work. Overall design looks good to me. But I have some comments. Please take a look.

src/otx/engine/engine.py
sovrasov: Checkpoint can be also an onnx file (MAPI allows that)

src/otx/engine/utils/auto_configurator.py
sovrasov: To my mind, `meta_info` looks better from the API side. When we call `get_ov_model` we may not bother ourselves with understanding how to obtain num classes from `Label_info`, it'll be done at the lower level. 
Also, current design prevents us from using hacks with explicit `num_classes` modification in the future.

sovrasov:
@harimkang thanks for the PR, if we find a way how to improve usability of OV models, that'd be a great progress.

The only problem with the proposed approach is data pipeline: OVModel has a different one than torch model.
For instance, see src/otx/recipe/classification/multi_label_cls/openvino_model.yaml
It's a coincidence, that for classification everything works (OVModel receives already resized to 224 images, but should receive original ones). In more complicated cases, we may define a torch pipeline, which is not suitable for OV models.

I tried the PR with CLI, and my test command now is:
```
otx test --config src/otx/recipe/classification/multi_class_cls/efficientnet_b0_light.yaml 
--engine.work_dir v2_efficientnet_b0_light
--checkpoint v2_efficientnet_b0_light/exported_model.xml
--data_root data/regression/cifar10_subset_cls_decr/
```
which is actually quite nice, since we don't need to pass a recipe with OV model. Unfortunately, that doesn't work if we pass `openvino_model.yaml` here: weights are not loaded, since they're supposed to be specified via --model.name.

From the CLI perspective, passing `openvino_model.yaml` looks as a suitable solution, since the Engine selects OV data pipeline.
From the API perspective, it looks like we need to replace data pipeline as well.

sovrasov:
I see that modifying data pipeline could complicate things for us, while in many cases mm data pipeline can work with OV models (or we can try to align it).
The main argument in favor of OV data pipeline is that we need test OV model on raw images, since it's the natural usecase for Geti and for any other users who do deployment via MAPI.

src/otx/engine/engine.py
harimkang: Thanks! Done - https://github.com/openvinotoolkit/training_extensions/pull/2897/commits/d8ab60c65a179e36732287ee3d4fec5e8e7bf0e5

src/otx/engine/engine.py
harimkang: Thanks! Done - https://github.com/openvinotoolkit/training_extensions/pull/2897/commits/d8ab60c65a179e36732287ee3d4fec5e8e7bf0e5

src/otx/engine/utils/auto_configurator.py
harimkang: Thanks! Done - https://github.com/openvinotoolkit/training_extensions/pull/2897/commits/d8ab60c65a179e36732287ee3d4fec5e8e7bf0e5

src/otx/engine/utils/auto_configurator.py
harimkang: I agree with Vlad, communicating with meta_info seems safe and I think it will be extensible in the future, so I'll leave it as is for now.

harimkang:
> @harimkang thanks for the PR, if we find a way how to improve usability of OV models, that'd be a great progress.
> 
> The only problem with the proposed approach is data pipeline: OVModel has a different one than torch model. For instance, see src/otx/recipe/classification/multi_label_cls/openvino_model.yaml It's a coincidence, that for classification everything works (OVModel receives already resized to 224 images, but should receive original ones). In more complicated cases, we may define a torch pipeline, which is not suitable for OV models.
> 
> I tried the PR with CLI, and my test command now is:
> 
> ```
> otx test --config src/otx/recipe/classification/multi_class_cls/efficientnet_b0_light.yaml 
> --engine.work_dir v2_efficientnet_b0_light
> --checkpoint v2_efficientnet_b0_light/exported_model.xml
> --data_root data/regression/cifar10_subset_cls_decr/
> ```
> 
> which is actually quite nice, since we don't need to pass a recipe with OV model. Unfortunately, that doesn't work if we pass `openvino_model.yaml` here: weights are not loaded, since they're supposed to be specified via --model.name.
> 
> From the CLI perspective, passing `openvino_model.yaml` looks as a suitable solution, since the Engine selects OV data pipeline. From the API perspective, it looks like we need to replace data pipeline as well.

Thank you for the detailed look, I'll have to think about the data pipeline and create a test for it.

src/otx/engine/utils/auto_configurator.py
eunwoosh: IMHO, giving num_classes seems better. It's because giving `LabelInfo` is actually same as giving availability to use all values in `LabelInfo`, which makes dependency between them. I think the smaller function does the better function is. How do you think about it?

harimkang:
> Unfortunately, that doesn't work if we pass `openvino_model.yaml` here: weights are not loaded, since they're supposed to be specified via --model.name.

This is something that won't change unless we improve OVModel. Since OVModel requires model.model_name, we should let users modify model.model_name. Probably, in the CLI, if it's `openvino_model.yaml`, we could add a way to link to it so that `checkpoint => model.model_name`. -> This is something I'll try.


> From the CLI perspective, passing `openvino_model.yaml` looks as a suitable solution, since the Engine selects OV data pipeline. From the API perspective, it looks like we need to replace data pipeline as well.

I personally think we can support both approaches, one using `openvino_model.yaml` and the other just using a configuration file from training + IR checkpoint.
- Case1: Using `openvino_model.yaml` -> I think this could probably be used if the user has a config they want to customize (ex. use_throughput_mode= True or False)
- Case2: Using 'configs.yaml' used for training + checkpoint:the path received as output from export -> I think it's a more natural flow for the user. It limits your ability to customize, but user don't have to change the recipe and it's more convenient.

eunwoosh:
I think it's better to align OV inference interface with torch model inference. Currently, they are different and it may confuse user. But as Vald said, we should use OV recipe data pipeline for OV models. If it's followed, I think we don't have to require user to use OV recipe for OV inference.

src/otx/engine/utils/auto_configurator.py
sovrasov: I think making the mentioned dependency is the only correct way of operating with labels-related info at the Engine level. We already have been using `num_classes` decoupled from label info in CLI auto-config, which resulted in refactoring to fix that and encapsulate `num_classes`.
Your assumption is correct at the lowest level of implementation, where we have for instance something like `self.logits_head = build_abc_head(num_classes)` .

harimkang:
Hi @sovrasov ,
I've been working on improvements on the API and CLI side.
- First, I modified the `datamodule` from `openvino_model.yaml` to be newly available on the API auto-configurator side. https://github.com/openvinotoolkit/training_extensions/pull/2897/commits/27702cbc9e00deb534147b7a5efa04604a7ca4ff
- Second, Add CLI link between --checkpoint and --model.model_name https://github.com/openvinotoolkit/training_extensions/pull/2897/commits/2aab09bdabd3c65293475c287ad5c441985107ae
- + Add API Example jupyter notebook https://github.com/openvinotoolkit/training_extensions/pull/2897/commits/39c2c03313790edf8a9bf12fda85a7e1c48a03d8
- I also added API integration tests to the export and IR Model tests.

So we can now proceed to export and test with the commands below.
```shell
# Training
>>> otx train --config src/otx/recipe/detection/atss_mobilenetv2.yaml --data_root tests/assets/car_tree_bug --max_epochs 10 --seed 3

# Testing with torch
>>> otx test --config src/otx/recipe/detection/atss_mobilenetv2.yaml --data_root tests/assets/car_tree_bug --checkpoint otx-workspace/checkpoints/epoch_004.ckpt

# Export to IR
>>> otx export --config src/otx/recipe/detection/atss_mobilenetv2.yaml --data_root tests/assets/car_tree_bug --checkpoint otx-workspace/checkpoints/epoch_004.ckpt

# Testing with IR Model
>>> otx test --config src/otx/recipe/detection/atss_mobilenetv2.yaml --data_root tests/assets/car_tree_bug --engine.device cpu --engine.work_dir ./otx-workspace/atss-mobilenet --checkpoint otx-workspace/exported_model.xml
```

We can also use traditional methods with `openvino_model.yaml`
```
# Testing with IR Model & OV configuration
>>> otx test --config src/otx/recipe/detection/openvino_model.yaml --data_root tests/assets/car_tree_bug --engine.device cpu --engine.work_dir ./otx-workspace/atss-mobilenet --checkpoint otx-workspace/exported_model.xml
```

A few limitations currently exist.
1. `otx optimize` should still use `--model.model_name`.
2. i think `model.model_name` in OVModel is still confusing to users.
3. When using API with auto-configuration, only data pipelines from fixed recipes (openvino_model.yaml) are available. It cannot be changed. -> This is only possible by modifying the configuration file or use CLI.

src/otx/engine/utils/auto_configurator.py
eunwoosh: I agree your opinion in some degree but honestly I still think it's better to give just num_classes. Dependency can makes it harder to change `LabelInfo` later(I know it barely occurs but as allusion to module), so that's why we try reducing dependency at the most. Anyway, this is about a design preference. So, I'm ok to leave it as it is.

sovrasov:
I also agree that case 2 is preferable, as soon as the issue with creating of an OV data pipeline is resolved.
Can `model.model_name` also be linked to `checkpoint` for `optimize` as well? From the code, I don't why the added link doesn't work for optimize.

harimkang:
> I also agree that case 2 is preferable, as soon as the issue with creating of an OV data pipeline is resolved. Can `model.model_name` also be linked to `checkpoint` for `optimize` as well? From the code, I don't why the added link doesn't work for optimize.

@sovrasov 
`optimize` don't have `checkpoint` argument.
image
This must have a checkpoint argument to be linkable.

sovrasov:
Ah, I see. Optimize is also implemented only for OV model now, so there's a little sense in adding a checkpoint.
If we switch to optimizing torch models, this can be resolved, but PTQ is not ready.

sovrasov (approved):

harimkang:
> Ah, I see. Optimize is also implemented only for OV model now, so there's a little sense in adding a checkpoint. If we switch to optimizing torch models, this can be resolved, but PTQ is not ready.

Yeah, so I'm just going to stop the coverage right here on that PR.

kprokofi (approved):

eunwoosh (approved):

PR #2900, opened 2024-02-07
Export with XAI branch

negvet:
### Summary

Support XAI export

Manually validated for 
- multi_class_cls/mobilenet_v3_large_light
- multi_class_cls/efficientnet_b0_light
- multi_class_cls/efficientnet_v2_light

### Checklist

### License

```python
```


src/otx/core/model/entity/classification.py
```diff
@@ -80,6 +80,33 @@ def reset_explain_hook(self) -> None:
         """Clear all history of explain records."""
         self.explain_hook.reset()
 
+    def get_custom_forward(self):
+        from otx.algo.hooks.recording_forward_hook import ReciproCAMHook
```
sovrasov, line 88: get_forward_with_explain()? Or we can add forward_with_explain() as an abstract method and suppose we always have it. If so, we don't need to pass it to the exporter as a parameter.
Does that impl work for all cls models?

src/otx/core/exporter/base.py
```diff
@@ -211,3 +220,15 @@ def _embed_openvino_ir_metadata(ov_model: openvino.Model, metadata: dict[tuple[s
...
         return ov_model
+
+    def _patch_model(self, model):
+        if self.custom_forward:
+            funcType = types.MethodType
+            model.forward = funcType(self.custom_forward, model)
+
+    def _update_output_names(self):
+        if "output_names" not in self.onnx_export_configuration:
+            self.onnx_export_configuration["output_names"] = ["logits", "saliency_map"]
```
sovrasov: Not all the model are exported to IRs via onnx, so that statement may have no effect.
As a universal way (for OVC and torch.onnx), one can return a dict with the required keys from the custom forward.

src/otx/core/exporter/base.py
```diff
@@ -64,6 +67,7 @@ def export(
         base_model_name: str = "exported_model",
         export_format: OTXExportFormatType = OTXExportFormatType.OPENVINO,
         precision: OTXPrecisionType = OTXPrecisionType.FP32,
+        dump_auxiliaries: bool = False,
```
sovrasov: We may remove this, since custom_forward can be optional, and can be applied only when it's actually passed.

src/otx/core/model/entity/base.py
```diff
@@ -248,9 +249,14 @@ def _export_parameters(self) -> dict[str, Any]:
             ("model_info", "label_ids"): all_label_ids.strip(),
             ("model_info", "optimization_config"): json.dumps(optimization_config),
         }
+        parameters["custom_forward"] = self.get_custom_forward()
 
         return parameters
 
+    def get_custom_forward(self):
```
sovrasov: I'd keep it as an private property. Can we consider renaming this to `_explain_forward` for clarity?

src/otx/core/model/entity/base.py
```diff
@@ -213,7 +214,7 @@ def export(
         Returns:
             Path: path to the exported model.
         """
-        return self._exporter.export(self.model, output_dir, base_name, export_format, precision)
+        return self._exporter.export(self.model, output_dir, base_name, export_format, precision, dump_auxiliaries)
```
sovrasov: When dump_auxiliaries is true, we need a check here, if custom forward actually exists or not 
If not -> not implemented error

src/otx/core/exporter/base.py
negvet: Exporter constructor will always get implementation of `custom_forward`, however it can be None.
It is not that handy for the user to control dumping feature vector and saliency maps (auxiliaries) by the availability of `custom_forward`. For models that support dumping auxiliaries, there always gonna be `custom_forward` implemented, but still the user should have the options to disable dumping auxiliaries (it should be by default actually).

sovrasov: I see, it looks like controlling forward is a bit painful in the exporter, so maybe we move it to the OTXModel?
This'd allow overloading exporters with model patching, and also this reduces the depth of modifications, when we need to change something related to model customization.

negvet: Ok, let me see what I can do

src/otx/core/exporter/base.py
negvet: Fixed, please take a look

src/otx/core/exporter/base.py
negvet: Moved it all to OTXModel

src/otx/core/model/entity/base.py
negvet: Fixed and renamed

src/otx/core/model/entity/base.py
negvet: Done

src/otx/core/model/entity/base.py
```diff
@@ -263,6 +280,33 @@ def _reset_prediction_layer(self, num_classes: int) -> None:
...
+            msg = "original_model_forward is None."
+            raise RuntimeError(msg)
+
+        func_type = types.MethodType
+        self.model.forward = func_type(self.original_model_forward, self.model)
+        self.original_model_forward = None
+
+    @staticmethod
+    def _update_onnx_output_names(onnx_export_configuration: dict) -> None:
+        if "output_names" not in onnx_export_configuration:
```
sovrasov: This implementation is suitable only for classification models, so it should be moved to the corresponding class.

src/otx/core/model/entity/base.py
```diff
@@ -210,10 +213,20 @@ def export(
...
+            dump_auxiliaries (bool): Flag to return "feature_vector" and "saliency_map". Defaults to False.
+
         Returns:
             Path: path to the exported model.
         """
-        return self._exporter.export(self.model, output_dir, base_name, export_format, precision)
+        exporter = self._exporter
+        if dump_auxiliaries:
+            self._reset_model_forward()
+            if hasattr(exporter, "onnx_export_configuration"):
```
sovrasov, line 224: Now this check looks as a workaround, which arises because `_exporter` is configurable only from the object state.
In the future, we can introduce explain mode, which affects the object state and allows switching to explain forward & export parameters.
For instance, `OTXModel.set_explain_mode(enabled=True/False)`.
That'd allow to use that mode with `litModule.predict` as well to dump explain results into a dataset.

src/otx/core/model/entity/base.py
negvet: Right, fixed

src/otx/core/model/entity/base.py
```diff
@@ -251,6 +266,10 @@ def _export_parameters(self) -> dict[str, Any]:
 
         return parameters
 
+    def _get_forward_with_auxiliaries(self) -> Callable | None:
```
vinnamkim, line 269: Why don't we just add `forward_explain()`  in this PR? It seems getting dirty without this I think.

sovrasov: If we go this way, we also have to provide a native mechanism for setting explain mode for the model. Now, it's done with a hook, but this mechanism has nothing to do with export and `forward_explain()`. So I believe it should be like explain_mode set -> forward is replaced with `forward_explain()`. At the same time, this should not be conflicting with the existing loss/predict modes.

sovrasov: Now it looks like custom forward immediately breaks mm's predict/loss method, while hooks allows to leave it as is.
Having this in mind, Evgeny's solution looks appropriate (replacing forward with xai one only for export), unless we change forward -> predict/loss -> _customize_outputs pipeline to support XAI natively

sovrasov:
@negvet I believe https://github.com/openvinotoolkit/training_extensions/pull/2905/commits/f846fe551ca42900bdf4322537fe9c7d67af398f is useful for the future work

sovrasov:
@negvet also please have a look at https://github.com/openvinotoolkit/training_extensions/pull/2913

src/otx/core/model/entity/base.py
vinnamkim: No, I just wanna add `forward_explain()`, not overriding `forward()`. Then, use this function for 1) `OTXModel.export(with_xai=True)` and 2) `Engine.predict(with_xai=True)`.

sovrasov:
@negvet can we close this PR?

sovrasov:
Implemented in https://github.com/openvinotoolkit/training_extensions/pull/2918

PR #2902, opened 2024-02-08
✨ Add Anomalib v1 tasks to OTX v2 

ashwinvaidya17:
# Summary

## Remaining

- [ ] Some files can be refactored and moved to separate folder to make them more readable 
- [ ] Performance numbers are lower than Anomalib. Need to investigate the cause.

### How to test

```python
import torch
from torchvision.transforms.v2 import Normalize, ToDtype

from otx.algo.anomaly.padim import Padim
from otx.algo.anomaly.stfpm import STFPM
from otx.core.data.transform_libs.torchvision import PadtoSquare, ResizetoLongestEdge
from otx.core.model.entity.anomaly import OTXAnomalyModel, OVAnomalyModel
from otx.core.types.export import OTXExportFormatType
from otx.core.types.task import OTXTaskType
from otx.data.anomaly import AnomalyDataModule
from otx.engine import Engine

transforms = [
    ResizetoLongestEdge(256, antialias=True),
    PadtoSquare(),
    ToDtype(torch.float32),
    Normalize(mean=[123.675, 116.28, 103.53], std=[58.395, 57.12, 57.375]),
]
model = Padim(input_size=(256, 256), layers=["layer1", "layer2", "layer3"], backbone="resnet18")
task = OTXTaskType.ANOMALY_CLASSIFICATION
engine = Engine(
        datamodule=datamodule,
        task=task,
        model=model,
        val_check_interval=1.0,
        num_sanity_val_steps=0,
    )
engine.train(max_epochs=1)
engine.test()
# OpenVINO
engine.export(export_format=OTXExportFormatType.OPENVINO)
model = OVAnomalyModel(
  model_name="otx-workspace/exported_model.xml",
  async_inference=True,
  use_throughput_mode=False,
)

engine = Engine(
    model=model,
    datamodule=datamodule,
    task=task,
)
engine.test()
```

### Checklist

### License

```python
```


src/otx/core/model/entity/anomaly.py
```diff
@@ -0,0 +1,257 @@
...
+            ("model_info", "image_shape"): model_info.image_shape,
+            ("model_info", "labels"): "Normal Anomaly",
+            ("model_info", "model_type"): "AnomalyDetection",
+            ("model_info", "task"): model_info.task.value,
+        }
+        super().__init__(
+            input_size=(1, 3, *model_info.image_shape),
+            mean=model_info.mean_values,
+            std=model_info.scale_values,
+            swap_rgb=True,  # set reverse input channels to True
```
sovrasov: Is this because anomaly model requires BGR as an input?

src/otx/algo/anomaly/padim.py
```diff
@@ -0,0 +1,52 @@
...
+        self.n_features = n_features
+        super().__init__()
+
+        self.anomalib_lightning_args.update(
+            input_size=self.input_size,
+            layers=self._layers,
+            backbone=self.backbone,
+            pre_trained=self.pre_trained,
+            n_features=self.n_features,
+        )
```
samet-akcay: what was the motivation of having the lightning args here again?

src/otx/core/model/entity/anomaly.py
ashwinvaidya17: In this case I would like the inputs to be in `RGB` format. If I understood the doc string in the base class correctly `(Whether to convert the image from BGR to RGB Defaults to False.)` then I should set it to True to get RGB inputs to the model. Or did I misunderstand something?

src/otx/core/data/dataset/anomaly/dataset.py
```diff
@@ -0,0 +1,146 @@
...
+            dm_subset,
+            transforms,
+            mem_cache_handler,
+            mem_cache_img_max_size,
+            max_refetch,
+            image_color_channel,
+            stack_images,
+        )
+
+    def _get_item_impl(self, index: int) -> AnomalyClassificationDataItem | AnomalySegmentationDataBatch:
```
harimkang: Doesn't the return type have a DetectionDataBatch?

src/otx/core/model/entity/anomaly.py
```diff
@@ -0,0 +1,283 @@
...
+from otx.core.model.entity.base import OTXModel, OVModel
+from otx.core.types.precision import OTXPrecisionType
+
+if TYPE_CHECKING:
+    from openvino.model_api.models import Model
+    from openvino.model_api.models.anomaly import AnomalyResult
+    from torchvision.transforms.v2 import Transform
+
+
+class _AnomalibLightningArgsCache:
```
harimkang: Is this to patch the lightning Args if it is internally an AnomalyModule?
I'm worried that this might change the value in an unintended way.
This is because Engine manages cache in a similar way, (`src/otx/core/utils/cache.py::TrainerArgumentsCache`)
and the roles look pretty much the same.
I'm not sure if this is absolutely necessary.

Can you explain a little more?

src/otx/core/model/module/anomaly.py
```diff
@@ -0,0 +1,531 @@
...
+    from anomalib.models import AnomalyModule
+    from lightning import LightningModule, Trainer
+    from lightning.pytorch.cli import LRSchedulerCallable, OptimizerCallable
+    from lightning.pytorch.core.optimizer import LightningOptimizer
+    from lightning.pytorch.utilities.types import STEP_OUTPUT
+    from openvino.model_api.models.anomaly import AnomalyResult
+    from torch.optim.optimizer import Optimizer
+
+
+class _RouteCallback(Callback):
```
harimkang: I think we need a little docstring to explain what exactly this does.

src/otx/core/model/module/anomaly.py
```diff
@@ -0,0 +1,531 @@
...
+    """Anomaly OTX Lightning model.
+
+    Used to wrap all the Anomaly models in OTX.
+    """
+
+    def __init__(
+        self,
+        otx_model: OTXAnomalyModel,
+        torch_compile: bool,
+        optimizer: OptimizerCallable,
```
harimkang: With a recent change, this can also accept `list | None`. (https://github.com/openvinotoolkit/training_extensions/pull/2910)

src/otx/recipe/anomaly/openvino.yaml
```diff
@@ -0,0 +1,21 @@
+model:
```
harimkang: The integration test assumes that the folder containing the recipe has the same name and `OTXTaskType` as the folder containing the recipe and collects it.
This may be redundant, but what about splitting the folder into `anomaly/anomaly_classification`, `anomaly/anomaly_detection`, and `anomaly/anomaly_segmentation`?

For now, I can't think of any other way than the above.

harimkang (commented):
Thanks for doing this, I have a few comments.

src/otx/core/data/dataset/anomaly/dataset.py
ashwinvaidya17: Ah I forgot to push this change.

src/otx/core/model/entity/anomaly.py
ashwinvaidya17: The problem is that in Anomalib, the parameters are passed to the lightning model. But since we create a torch (OTX) model and pass it to the engine/config.yaml, we need a way to pass parameters to the lightning model. The idea behind this is to cache all the arguments that are passed to the OTXModel and read them in the lightning model. I don't know a better workaround for this.

src/otx/core/data/dataset/anomaly/dataset.py
```diff
@@ -0,0 +1,150 @@
...
+        datumaro_item = self.dm_subset.get(id=self.ids[index], subset=self.dm_subset.name)
+        img = datumaro_item.media_as(Image)
+        # returns image in RGB format if self.image_color_channel is RGB
+        img_data, img_shape = self._get_img_data_and_shape(img)
+        # Note: This assumes that the dataset is in MVTec format.
+        # We can't use datumaro label id as it returns some number like 3 for good from which it is hard to infer
+        # whether the image is Anomalous or Normal. Because it leads to other questions like what do numbers 0,1,2 mean?
+        label: torch.LongTensor = (
+            torch.tensor(0.0, dtype=torch.long) if "good" in datumaro_item.id else torch.tensor(1.0, dtype=torch.long)
+        )
```
vinnamkim, line 70: @wonjuleee, it seems there are no binary case in `label_id` of Datumaro MVTec format.

src/otx/core/data/entity/anomaly/classification.py
```diff
@@ -0,0 +1,91 @@
...
+    """Anomaly classification Prediction item."""
+
+
+@dataclass
+class AnomalyClassificationBatchPrediction(AnomalyClassificationDataBatch, OTXBatchPredEntity):
+    """Anomaly classification batch prediction."""
+
+    anomaly_maps: torch.Tensor
+    # Note: ideally this should be anomalous_scores but it is now used to shadow the scores in OTXBatchPredEntity
+    scores: torch.bool
```
vinnamkim: What's the reason behind shadowing it?

src/otx/core/data/entity/anomaly/detection.py
```diff
@@ -0,0 +1,92 @@
...
+    ) -> AnomalyDetectionDataBatch:
+        """Collection function to collect `OTXDataEntity` into `OTXBatchDataEntity` in data loader."""
+        batch = super().collate_fn(entities)
+        images = tv_tensors.Image(data=torch.stack(batch.images, dim=0)) if stack_images else batch.images
+        return AnomalyDetectionDataBatch(
+            batch_size=batch.batch_size,
+            images=images,
+            imgs_info=batch.imgs_info,
+            labels=[entity.label for entity in entities],
+            masks=torch.vstack([resize(entity.mask, size=[images.shape[2], images.shape[3]]) for entity in entities]),
```
vinnamkim: Why do you resize here not during the `transform` pipeline?

src/otx/core/data/entity/anomaly/classification.py
ashwinvaidya17: Ah good question. I don't think I need to shadow this anymore.

src/otx/core/model/module/anomaly.py
vinnamkim, line 1: If you decided to wrap anomalib module into `OTXModel`, such as
```python
class AnomalyOTXModel(OTXModel):
      def _create_model(self):
            from anomalib import SomeModel
            return SomeModel(...)
      def forward(self, inputs):
            if self.training:
                return self.model.training_step(**customize_inputs(inputs))
            return return self.model.predict_step(**customize_inputs(inputs))
      
```
I don't know why there are so many code lines only for `OTXBaseAnomalyLitModel` unlike other `OTXLitModule`. Can you explain it?

src/otx/core/data/entity/anomaly/detection.py
ashwinvaidya17: I've updated the PR

src/otx/core/model/module/anomaly.py
ashwinvaidya17: I can try passing the anomaly lightning model to the OTX model instead of the anomaly torch model

ashwinvaidya17: I've moved the anomalib lightning model in the OTXModel. It does not solve the entire problem though. This is because some processing happens in the anomalib's lightning model. But in OTX we have another lightning model on top. So some calls need to be rerouted and we need to handle how saving and loading of model checkpoint is handled. The cleanest solution would be to have custom engine support. This way we can just call the Anomalib's engine for anomaly tasks.

vinnamkim: What if `OTXModel` promotes to `OTXModel(LitModule)` and replace `OTXLitModule` so there is no more `OTXLitModule` such as

```python
from anomalib import SomeModel

class OTXAnomalySomeModel(SomeModel, OTXAnomalyModel):
     ...

model = OTXAnomalySomeModel() # This is LitModule
Engine(model=model).train()
```
Is it helpful for your problem?

ashwinvaidya17: That might work. I'll try this on my end to see if it covers all the cases.

src/otx/core/model/entity/anomaly.py
harimkang: Then how about reuse this class (`src/otx/core/utils/cache.py::TrainerArgumentsCache`) to eliminate duplication?

src/otx/core/model/module/anomaly.py
samet-akcay: @vinnamkim, if `OTXModel` inherits `LightningModule` instead of `nn.Module`, it would solve a lot of issues on the anomalib side. Also, I believe it would reduce the overall complexity of OTX as well as coupling

ashwinvaidya17: Yes, I think this will work really well. Here is a super tiny PoC
```python
class OTXModel(LightningModule):
    def __init__(self) -> None:
        super().__init__()


class OTXToAnomalibMixin:
    def configure_callbacks(self) -> Callback | None:
        """Get all necessary callbacks required for training and post-processing on Anomalib models."""
        image_metrics = ["AUROC", "F1Score"]
        pixel_metrics = image_metrics if self.task_type != TaskType.CLASSIFICATION else None
        return [
            _PostProcessorCallback(),
            _MinMaxNormalizationCallback(),  # ModelAPI only supports min-max normalization as of now
            _ThresholdCallback(threshold="F1AdaptiveThreshold"),
            _MetricsCallback(
                task=self.task_type,
                image_metrics=image_metrics,
                pixel_metrics=pixel_metrics,
            ),
        ]

    def training_step(self, inputs, batch_idx):
        inputs = self._customize_inputs(inputs)
        return super().training_step(inputs, batch_idx)

    def validation_step(self, inputs, batch_idx):
        inputs = self._customize_inputs(inputs)
        return super().validation_step(inputs, batch_idx)

    def _customize_inputs(self, inputs):  # anomalib model inputs
        """Customize inputs for the model."""
        return {"image": inputs.images, "label": torch.vstack(inputs.labels).squeeze(), "mask": inputs.masks}

    def load_state_dict(self, ckpt, *args, **kwargs):
        ckpt = ckpt.get("state_dict", ckpt)
        return super().load_state_dict(ckpt, *args, **kwargs)


class OTXPadimLightning(OTXToAnomalibMixin, Padim, OTXModel):
    def __init__(
        self,
        input_size: tuple[int, int] = (256, 256),
        backbone: str = "resnet18",
        layers: list[str] = ["layer1", "layer2", "layer3"],
        pre_trained: bool = True,
        n_features: int | None = None,
    ) -> None:
        super().__init__(
            input_size=input_size,
            backbone=backbone,
            layers=layers,
            pre_trained=pre_trained,
            n_features=n_features,
        )
        self.task_type = TaskType.SEGMENTATION


model = OTXPadimLightning()
datamodule = AnomalyDataModule(
    task_type=OTXTaskType.ANOMALY_SEGMENTATION,
    data_dir="/home/ashwin/datasets/MVTec/bottle",
    train_transforms=transforms,
    val_transforms=transforms,
    test_transforms=transforms,
)
engine = Engine(
    datamodule=datamodule,
    task=OTXTaskType.ANOMALY_SEGMENTATION,
    model=model,
    val_check_interval=1.0,
    num_sanity_val_steps=0,
)

engine.train(max_epochs=1)
engine.test()
print(engine.predict())
```

To get this working I had to set `lit_module = self.model` in Engine

ashwinvaidya17: Of course, I don't know how it will behave once `OTXModel` is populated as both Padim and OTXModel are Lightning models.

vinnamkim: > To get this working I had to set lit_module = self.model in Engine

If so, would you please add that workaround only for anomaly tasks with `TODO` comment in this PR?

Like this comment https://github.com/openvinotoolkit/training_extensions/pull/2915#discussion_r1490505527, I'll do refactoring to deprecate `OTXLitModule`. Then, we can revisit it afterwards.

samet-akcay: Nice one! I'm excited to see this improvement! Thanks guys!

ashwinvaidya17:
I've pushed new changes. Not sure if this helps for now but maybe with the OTXModel refactor it might become better. But it would be nice to get feedback from you all.
I still have to extend changes to Anomaly OpenVINO inference.

harimkang:
> I've pushed new changes. Not sure if this helps for now but maybe with the OTXModel refactor it might become better. But it would be nice to get feedback from you all. I still have to extend changes to Anomaly OpenVINO inference.

If possible, could you please separate IR-related functionality into the next PR?

ashwinvaidya17:
> > I've pushed new changes. Not sure if this helps for now but maybe with the OTXModel refactor it might become better. But it would be nice to get feedback from you all. I still have to extend changes to Anomaly OpenVINO inference.
> 
> If possible, could you please separate IR-related functionality into the next PR?

Do you want me to remove the export stuff from this PR?

harimkang:
> > > I've pushed new changes. Not sure if this helps for now but maybe with the OTXModel refactor it might become better. But it would be nice to get feedback from you all. I still have to extend changes to Anomaly OpenVINO inference.
> > 
> > 
> > If possible, could you please separate IR-related functionality into the next PR?
> 
> Do you want me to remove the export stuff from this PR?

I'm not saying you have to do that, but if you think the IR is going to take a long time. It seems like a good idea to separate things that are going to take a long time. (It doesn't look good to have PRs open for too long.)

If you do separate them, you can skip the IR-related integration tests for now.

src/otx/algo/anomaly/padim.py
```diff
@@ -0,0 +1,43 @@
...
+
+from __future__ import annotations
+
+from anomalib.models.image import Padim as AnomalibPadim
+
+from otx.core.model.entity.base import OTXModel
+from otx.core.model.module.anomaly import OTXAnomaly
+
+
+class Padim(OTXAnomaly, OTXModel, AnomalibPadim):
```
samet-akcay, line 14: This is a temporary solution, right? Once the base interface is refactor, we would not need to inherit three classes?

src/otx/algo/anomaly/stfpm.py
```diff
@@ -0,0 +1,46 @@
...
+from anomalib.models.image.stfpm import Stfpm as AnomalibStfpm
+
+from otx.core.model.entity.base import OTXModel
+from otx.core.model.module.anomaly import OTXAnomaly
+
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+
+
+class STFPM(OTXAnomaly, OTXModel, AnomalibStfpm):
```
samet-akcay: To match the anomalib naming, maybe we could rename this to
```suggestion
class Stfpm(OTXAnomaly, OTXModel, AnomalibStfpm):
```

src/otx/algo/anomaly/stfpm.py
```diff
@@ -0,0 +1,46 @@
...
+        layers (Sequence[str]): Feature extractor layers.
+        input_size (tuple[int, int]): Input size.
+        backbone (str, optional): Feature extractor backbone. Defaults to "resnet18".
+    """
+
+    def __init__(
+        self,
+        layers: Sequence[str],
+        input_size: tuple[int, int],
+        backbone: str = "resnet18",
```
samet-akcay: Padim model has the default values for these args. Would it be possible to follow the same here too?

src/otx/core/data/dataset/anomaly/dataset.py
```diff
@@ -0,0 +1,151 @@
+"""Anomaly Classification Dataset."""
+
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+
```
samet-akcay:
```suggestion
```

src/otx/algo/anomaly/padim.py
```diff
@@ -0,0 +1,43 @@
+"""OTX Padim model."""
+
```
samet-akcay, line 2:
```suggestion
```

src/otx/core/data/dataset/anomaly/dataset.py
```diff
@@ -0,0 +1,151 @@
...
+            )
+        elif self.task_type == OTXTaskType.ANOMALY_SEGMENTATION:
+            # Note: this part of code is brittle. Ideally Datumaro should return masks
+            # Another major problem with this is that it assumes that the dataset passed is in MVTec format
+            mask_file_path = (
+                Path("/".join(datumaro_item.media.path.split("/")[:-3]))
+                / "ground_truth"
+                / f"{('/'.join(datumaro_item.media.path.split('/')[-2:])).replace('.png','_mask.png')}"
+            )
+            mask = torch.zeros(1, img_shape[0], img_shape[1], dtype=torch.uint8)
```
samet-akcay, line 91: as you mentioned, this part is to be re-visited.

src/otx/core/data/dataset/anomaly/dataset.py
```diff
@@ -0,0 +1,151 @@
...
+            msg = f"Task {self.task_type} is not supported yet."
+            raise NotImplementedError(msg)
+
+        # without ignore the following error is returned
+        # Incompatible return value type (got "Any | None", expected
+        # "AnomalyClassificationDataItem | AnomalySegmentationDataBatch | AnomalyDetectionDataBatch")
+        return self._apply_transforms(item)  # type: ignore[return-value]
+
+    @property
+    def collate_fn(self) -> Callable:
```
samet-akcay, line 141: @djdameln, would be good to get your opinion here. Would you prefer having a single `AnomalyDataset` that returns the return_val based on the task, or should we have a dataset for each task?

src/otx/core/data/entity/anomaly/classification.py
```diff
@@ -0,0 +1,89 @@
...
+    OTXDataEntity,
+    OTXPredEntity,
+)
+from otx.core.data.entity.utils import register_pytree_node
+from otx.core.types.task import OTXTaskType
+
+# ---------------------------------------------------------------------------- #
+# Data Objects.                                                                #
+# ---------------------------------------------------------------------------- #
+
```
samet-akcay:
```suggestion
```

src/otx/core/data/entity/anomaly/classification.py
```diff
@@ -0,0 +1,89 @@
...
+        """Pin memory for member tensor variables."""
+        super().pin_memory()
+        self.labels = [label.pin_memory() for label in self.labels]
+        return self
+
+
+# ---------------------------------------------------------------------------- #
+# Prediction Objects.                                                          #
+# ---------------------------------------------------------------------------- #
+
```
samet-akcay:
```suggestion
```

src/otx/core/data/entity/anomaly/classification.py
```diff
@@ -0,0 +1,89 @@
...
+
+@dataclass
+class AnomalyClassificationPrediction(AnomalyClassificationDataItem, OTXPredEntity):
```
samet-akcay, line 72: I'm aware this is related to the main interface, but is single prediction needed anywhere? Inferencers only?

src/otx/core/data/entity/anomaly/detection.py
samet-akcay, line 1: a bit related to the datamodule question: @djdameln, what do you think of the task-based data/batch items vs anomaly data/batch item? Trying to figure out which approach would be the best approach for what we will try to achieve with the new geti anomaly task.

src/otx/core/data/factory.py
samet-akcay, line 1: @harimkang @vinnamkim, ideally data factory should not exist. If this is used for the CLI only, ideally jsonargparse should get the datamodule based on the `class_path` and `init_args`.

This approach causes inconsistent config format for data and the rest of the otx components.

src/otx/core/model/module/anomaly/__init__.py
```diff
@@ -0,0 +1,9 @@
+"""Anomaly models."""
+
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+
```
samet-akcay:
```suggestion
```

src/otx/core/model/module/anomaly/anomaly_lightning.py
```diff
@@ -0,0 +1,450 @@
+"""Anomaly Lightning OTX model."""
+
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
```
samet-akcay, line 4:
```suggestion
# SPDX-License-Identifier: Apache-2.0

```

src/otx/core/model/module/anomaly/anomaly_lightning.py
```diff
@@ -0,0 +1,450 @@
...
+from __future__ import annotations
+
+from pathlib import Path
+from typing import TYPE_CHECKING, Any
+
+import onnx
+import openvino
+import torch
+from anomalib import TaskType
```
samet-akcay: would it be an idea to import this as AnomalibTaskType to avoid confusion later in the codebase

src/otx/core/model/entity/anomaly.py
```diff
@@ -0,0 +1,74 @@
...
+    AnomalySegmentationBatchPrediction,
+)
+from otx.core.model.entity.base import OVModel
+
+if TYPE_CHECKING:
+    from openvino.model_api.models import Model
+    from openvino.model_api.models.anomaly import AnomalyResult
+
+
+class OVAnomalyModel(OVModel):
```
samet-akcay: why do we need an additional ov model? What is the difference between ovmodel and exporter?

src/otx/core/model/module/anomaly/anomaly_lightning.py
```diff
@@ -0,0 +1,450 @@
...
+            input_names=["input"],
+            output_names=["output"],
+        )
+        onnx_model = onnx.load(save_path)
+        onnx_model = self._postprocess_onnx_model(onnx_model, embed_metadata, precision)
+        onnx.save(onnx_model, save_path)
+        return Path(save_path)
+
+
+class OTXAnomaly:
```
samet-akcay, line 131: Can we name this as mixin? I was initially confused that this was the actual implementation

src/otx/core/model/module/anomaly/anomaly_lightning.py
```diff
@@ -0,0 +1,450 @@
...
+            _MetricsCallback(
+                task=self.task,
+                image_metrics=image_metrics,
+                pixel_metrics=pixel_metrics,
+            ),
+        ]
+
+    def training_step(
+        self,
+        inputs: AnomalyClassificationDataBatch | AnomalySegmentationDataBatch | AnomalyDetectionDataBatch | dict,
```
samet-akcay: Would it be an idea to use a `TypeAlias` here, similar to how we do it in anomalib?

src/otx/data/__init__.py
```diff
@@ -0,0 +1,9 @@
+"""OTX Datamodules."""
+
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
```
samet-akcay, line 5:
```suggestion
```

src/otx/data/anomaly/__init__.py
```diff
@@ -0,0 +1,9 @@
+"""OTX Anomaly Datamodules."""
+
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
```
samet-akcay, line 5:
```suggestion
```

src/otx/data/anomaly/anomaly.py
```diff
@@ -0,0 +1,85 @@
+"""OTX Anomaly Datamodules."""
+
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
```
samet-akcay, line 4:
```suggestion
# SPDX-License-Identifier: Apache-2.0

```

src/otx/recipe/anomaly/anomaly_classification/stfpm.yaml
```diff
@@ -0,0 +1,89 @@
...
+    - class_path: lightning.pytorch.callbacks.EarlyStopping
+      init_args:
+        patience: 5
+  data:
+    task: ANOMALY_CLASSIFICATION
+    config:
+      data_format: mvtec
+      vpm_config:
+        use_bbox: True
+        use_point: False
```
samet-akcay: @harimkang, do you know what these params are?

samet-akcay (changes requested):
@ashwinvaidya17, thanks a lot for the great effort! Much appreciated.

I've added some minor comments, and some questions.

src/otx/recipe/anomaly/anomaly_classification/stfpm.yaml
harimkang: As far as I know, this is only used in the visual prompting model and does not need to be declared in anomaly.

src/otx/engine/engine.py
```diff
@@ -420,7 +420,7 @@ def export(
 
             lit_module.load_state_dict(loaded_checkpoint)
 
-            return self.model.export(
+            return lit_module.export(
```
harimkang: Is this behavior currently available for other tasks?

ashwinvaidya17: I've updated the base class to call self.model.export from lit_module.export. So hopefully it should not break stuff

ashwinvaidya17:
Can we skip python 3.9 checks for anomaly task? Anomalib v1 supports 3.10 and above

src/otx/algo/anomaly/padim.py
samet-akcay: Let's re-visit this once the base interface is undergone to a refactor

samet-akcay (dismissed):
Thanks for the effort!

harimkang:
> Can we skip python 3.9 checks for anomaly task? Anomalib v1 supports 3.10 and above

I talked to Mark and Emily about this, and we decided that OTX 2.0 will also not support python 3.9 for now. So I'm going to remove it from the CI test.

harimkang:
> Can we skip python 3.9 checks for anomaly task? Anomalib v1 supports 3.10 and above

i create PR (https://github.com/openvinotoolkit/training_extensions/pull/2991) - Merged

harimkang (approved):
Can you please add integration tests to CI in the next PR? I think we need to add tasks to `tox.ini` and `.github/workflows/pre_merge.yaml`.

src/otx/core/data/entity/anomaly/classification.py
ashwinvaidya17: I'll revisit this with the OpenVINO task

vinnamkim (approved):

sungmanc (approved):

PR #2903, opened 2024-02-08
Refine Tile Processing and Expand Tile Configuration

eugene123tw:
### Summary

* Introduce Tile Sampling Feature
* Implement Tile Adaptor
* Integrate OTX Tile Transform: Unfolding Tiles with Customizable Tile Size
* Filtering for Empty Val set Tiles
* Expand Tile Configuration Options
* Incorporate NMS in Tile Postprocessing
* Update Tile-Related Recipes to Accommodate Changes

### Checklist

### License

```python
```


src/otx/algo/callbacks/tile_sampling_hook.py
```diff
@@ -0,0 +1,25 @@
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
+"""Tile Sampling Hook for OTXTileTrainDataset."""
+from lightning import Callback, LightningModule, Trainer
+
+
+class TileSamplingHook(Callback):
```
sungmanc: Personally, I prefer to use the word "Callback" for the OTX2.0
```suggestion
class TileSamplingCallback(Callback):
```

vinnamkim: Why do we need this?

src/otx/core/data/dataset/tile.py
```diff
@@ -124,12 +197,16 @@ def get_tiles(self, image: np.ndarray, item: DatasetItem) -> tuple[list[OTXDataE
...
         tile_ds = tile_ds.transform(
-            Tile,
-            grid_size=self.tile_config.grid_size,
+            OTXTileTransform,
+            tile_size=self.tile_config.tile_size,
             overlap=(self.tile_config.overlap, self.tile_config.overlap),
             threshold_drop_ann=0.5,
         )
 
+        if self.dm_subset.name == "val":
```
vinnamkim, line 205: After considering https://github.com/openvinotoolkit/datumaro/pull/1264 and https://docs.ultralytics.com/yolov5/tutorials/tips_for_best_training_results/#dataset, I believe that evaluation on no label image is also important for demonstrating model's accuracy on false positives. Therefore, I think that we should not filter an image with empty annotation in validation/test subsets.

@jaegukhyun

src/otx/core/data/dataset/tile.py
```diff
@@ -152,20 +229,30 @@ class OTXTileTrainDataset(OTXTileDataset):
...
-        dataset.dm_subset = dm_subset
-        dataset.ids = [item.id for item in dm_subset]
+        self.dm_subset = DatasetSubset(dm_dataset, dataset.dm_subset.name)
+        self.tile_config = tile_config
+        sampled_subset = self.sample_subset()
+        dataset.dm_subset = sampled_subset
+        dataset.ids = [item.id for item in sampled_subset]
         super().__init__(dataset, tile_config)
 
+    def sample_subset(self) -> DatasetSubset:
```
vinnamkim: Why do we need this?

src/otx/core/data/dataset/tile.py
jaegukhyun: I agree on this. OTX1.5 filters empty annotations for train dataset, not for val or test dataset.
And OTX2.0 also only filters for train dataset
https://github.com/openvinotoolkit/training_extensions/blob/c5ad4258a29376b3f9952c9b5441b675dd6f5c8c/src/otx/core/data/pre_filtering.py#L28

jaegukhyun: I'm not sure whether tiling has same strategy with normal detection training in OTX1.5.
Could you verify this?

src/otx/algo/callbacks/tile_sampling_hook.py
eugene123tw: @vinnamkim This PR mainly brings the tiling features from OTX 1.x to here.

Tile sampling is employed to expedite training duration by selecting a portion of samples. Through extensive experimentation, we observed the efficacy (faster training while maintaining good accuracy) of tile sampling particularly for Geti dataset (Vitens). However, numerous tiles can significantly impede training speed. This enhancement primarily targets Geti users who prioritize prompt results. Tiling processes, both in training and inference, are time-consuming. We (Jauguk, Songki, Sungman and I) dedicated 2-3 weeks to optimizing tiling in OTX 1.x. For further insights, you could DM me or refer to the folks involved. I don't think it's ideal to elaborate on all previous activities in this PR.

src/otx/core/data/dataset/tile.py
eugene123tw: It's used in `TileSamplingHook` which sample tiles from training dataset when epoch starts

src/otx/core/data/dataset/tile.py
eugene123tw: @jaegukhyun We've implemented a filtering mechanism for empty images in both **train** and **val** datasets (excluding the test set). You can find more details on the implementation at this link: [GitHub - OTX Detection Adapters Dataset Module](https://github.com/openvinotoolkit/training_extensions/blob/develop/src/otx/algorithms/detection/adapters/mmdet/datasets/dataset.py#L344). This idea was originally proposed by you in OTX 1.x. Initially, I included all tiles for evaluation, but during tile optimization exercises, we discovered that this approach was too slow. Implementing the filter significantly improved evaluation speed.

@vinnamkim If you believe this filter is unnecessary, I recommend bringing it up with Mark. This feature plays a crucial role in optimizing the speed of tiling. Removing it would likely result in slower tiling processes. Ultimately, the decision regarding whether slower training is acceptable rests with Mark.

jaegukhyun: I agree on @eugene123tw, since OTX1.5 also filters validation dataset

src/otx/algo/callbacks/tile_sampling_hook.py
vinnamkim: I still don't know why it is needed. But if so, it seems better to add [RandomSampler](https://pytorch.org/docs/stable/data.html#torch.utils.data.RandomSampler) with `num_samples` argument to `torch.DataLoader` at initializing `OTXDataModule` rather than adding this callback. The current way can create a hidden constraint for the configurations (sampling prob < 1.0 should include this callback).

src/otx/core/data/dataset/tile.py
vinnamkim: I'll discuss it with Wonju and Mark

src/otx/algo/callbacks/tile_sampling_hook.py
eugene123tw: Thank you for the advice. I've implemented your suggestion by adding `RandomSampler` to train dataloader, which is a better approach. Also, I've removed the Datumaro sampling and tile sampling hook and made other necessary adjustments.

vinnamkim (approved):

harimkang (approved):

sungmanc (approved):

jaegukhyun (approved):

PR #2905, opened 2024-02-12
Refine #2900: export with XAI

sovrasov:
### Summary

This PR extends https://github.com/openvinotoolkit/training_extensions/pull/2900
The main rationale to these changes is that we need to have a single mechanism for introducing XAI into the model.
Custom forward only for export adds the second xai forward mechanism which is hard to test.

### Checklist

### License

```python
```


src/otx/core/model/entity/classification.py
```diff
@@ -75,12 +80,61 @@ def head_forward_fn(self, x: torch.Tensor) -> torch.Tensor:
...
+        if self.explain_hook is not None:
+
+            class ModelWithExplainOutput(nn.Module):
+                def __init__(self, model: nn.Module, hook: BaseRecordingForwardHook) -> None:
+                    super().__init__()
+                    self.model = model
+                    self.hook = hook
+                    self.hook.reset()
+
+                def forward(self, x: torch.Tensor) -> tuple:
```
kprokofi, line 108: Why can't we move this as global forward function for the entire ExplainableOTXClsModel?
We can call super().forward() and then check whether we have explain hook or not to return just logits or tuple consist of logits and saliency maps.
I just don't like much this wrapper class which substitute our OTXModel.

src/otx/core/model/entity/classification.py
```diff
@@ -75,12 +80,61 @@ def head_forward_fn(self, x: torch.Tensor) -> torch.Tensor:
...
+                    self.hook.reset()
+                    return logits, saliency_maps
+
+            return ModelWithExplainOutput(super()._exportable_model, self.explain_hook)
+
+        return super()._exportable_model
+
+    @staticmethod
+    def _update_onnx_output_names(onnx_export_configuration: dict) -> None:
+        if "output_names" not in onnx_export_configuration:
```
kprokofi: Where this method will be used? What if we don't want to export saliency maps. Will this OK to append both ["logits", "saliency_map"] for output_names?

src/otx/engine/engine.py
```diff
@@ -367,13 +367,15 @@ def export(
         checkpoint: str | Path | None = None,
         export_format: OTXExportFormatType = OTXExportFormatType.OPENVINO,
         export_precision: OTXPrecisionType = OTXPrecisionType.FP32,
+        dump_auxiliaries: bool = False,
```
kprokofi, line 370: I don't really like the naming. Maybe "dump_features" or "dump_saliency_maps", or just "explain"

src/otx/core/model/entity/classification.py
sovrasov: The final mechanism of obtaining the saliency maps is still not clear. Actually, we can already access the hook in `_customize_outputs`, the problem is what to do with the raw saliency maps. This wrapper doesn't look as a comprehensive replacement to forward, it's rather a workaround if batch is 1.

src/otx/core/model/entity/classification.py
sovrasov: That's a leftover, deleted.

src/otx/engine/engine.py
sovrasov: This comes from the original implementation, I also prefer *explain, since feature vector is something unclear and required only because we have Geti integration requirements.
@negvet what do you think?

src/otx/core/model/entity/classification.py
sovrasov: Also, if we look in general, the whole idea of hooks is that we don't change anything in the forward itself, while can extend it with an arbitrary number of hooks. 
If we want a custom forward, we don't need hooks, but with hooks we have to generate a custom forward for export, since torch tracer is quite limited in terms of functionality.

vinnamkim: https://github.com/openvinotoolkit/training_extensions/pull/2870#discussion_r1475416405
https://github.com/openvinotoolkit/training_extensions/pull/2900#discussion_r1487040737
Like these comments, I think that we should introduce a dedicated `forward()` function for XAI at this time.

sovrasov: @vinnamkim does that mean we have to remove the hook? If we have a standalone forward, keeping the hooks makes no sense, like I described in the previous comment.

vinnamkim: > @vinnamkim does that mean we have to remove the hook? If we have a standalone forward, keeping the hooks makes no sense, like I described in the previous comment.

Yes

I guess you guys can move forward gradually. For example,

1. Introduce `forward_explain()` with the hook

```python
def forward_explain(self, inputs) -> BatchPredictionWithXAI:
     handle = self.register_explain_hook()
     outputs = self.forward(inputs)
     augmented_outputs = augment_with_xai_outputs(outputs, handle.xai_outputs)
     handle.remove()
     return augmented_outputs
```
2. Remove using the hook from `forward_explain()`

PR #2912, opened 2024-02-13
Enable HPO in OTX 2.0

eunwoosh:
### Summary

This PR enables HPO in OTX 2.0
User can run HPO by `engine.train` argument. After HPO is done, model is trained with best hyper parameter and model weight trained w/ best hyper parameter during HPO.
Features HPO has are almost same as OTX 1.X except one thing. HPO don't have to use HPO config file in OTX 2.0. HPO automatically optimizes learning rate and batch size based on default setting if HPO config file isn't given.
HPO in OTX 2.0 supports to optimize 'data.config.train_subset.batch_size', 'optimizer' (which includes lr) and 'scheduler'.
If user want to optimize other hyper parameters in optimizer or scheduler, user can pass it via hpo_config argument (HpoConfig).
HpoConfig format is almost same as 'hpo_config.yaml' in OTX 1.X but I changed a few names to align all names in config file to HPO algorithm arguments.
Therefore, HpoConfig have same attributes as arguments of HPO algorithm class (`HyperBand` class) and users can customize HPO including which hyper parameters to optimize.

In addition to HPO, `append_signal_handler` is refined.

### How to test

- otx train --config ... --data_root ... --run_hpo true
- otx train --config ... --data_root ... --run_hpo true --hpo_config.expected_time_ratio 8

### Checklist

### License

```python
```


src/otx/engine/engine.py
```diff
@@ -155,6 +171,9 @@ def train(
         callbacks: list[Callback] | Callback | None = None,
         logger: Logger | Iterable[Logger] | bool | None = None,
         resume: bool = False,
+        run_hpo: bool = False,
```
harimkang, line 156: What do we get with `train(run_hpo=True)`?
Is this after the HPO finds the optimal parameters and then train proceeds?

If a user doesn't like the HPO results, can we get feedback on this before training?

src/otx/engine/engine.py
```diff
@@ -155,6 +171,9 @@ def train(
...
+        hpo_time_ratio: int = 4,
+        hpo_cfg_file: str | Path | None = None,
```
harimkang: What is `hpo_cfg_file` and why do we need it?
In my personal opinion, a lot of config files are a bad idea. 
Therefore, we need to put more thought into making sure that the cfg values that the user can customize are well accepted. (include `hpo_time_ratio` as well) -> I think the arguments are too much for HPO and not for train.

src/otx/engine/engine.py
```diff
@@ -240,6 +267,15 @@ def train(
         self.checkpoint = self.trainer.checkpoint_callback.best_model_path
         return self.trainer.callback_metrics
 
+    @staticmethod
+    def _prepare_hpo_args(all_args: dict[str, Any]) -> dict[str, Any]:
```
harimkang: Is there any way to remove this function?

src/otx/engine/engine.py
```diff
@@ -139,6 +135,26 @@ def __init__(
             scheduler if scheduler is not None else self._auto_configurator.get_scheduler()
         )
 
+    @property
+    def work_dir(self) -> PathLike:
+        """Work directory."""
+        return self._work_dir
+
+    @work_dir.setter
+    def work_dir(self, val: PathLike) -> None:
```
harimkang: Why do we need a setter function on `work_dir`, is it just for hpo?

src/otx/engine/utils/hpo/__init__.py
```diff
@@ -0,0 +1,9 @@
+# Copyright (C) 2024 Intel Corporation
```
harimkang: Is the difference between `otx/hpo` and `engine/utils/hpo` just to distinguish the functions used by engine?
If so, wouldn't it be better to put them all in one file called `utils/hpo.py or hpo_utils.py`?

harimkang (commented):
Thanks for doing this! i left few comments

src/otx/engine/engine.py
eunwoosh: yes but need this part in `train` functions. I think making it as separated function is better for readability.

src/otx/engine/engine.py
eunwoosh: HPO can works w/o this change. When we change `engine.work_dir`, it seems that cache isn't updated, so I added it.

src/otx/engine/utils/hpo/__init__.py
eunwoosh: I don't think so. At first, I implemented as you said but it was hard to read so I divided it into three files that have a lot in common.

src/otx/engine/engine.py
eunwoosh: hpo_cfg_file is for configuring HPO in detail. I updated this PR docs including config file explanation. Please read it.
And w/o `hpo_time_ratio` argument, how to give interface to change HPO time ratio?

src/otx/engine/engine.py
eunwoosh: yes you're right. After HPO, training proceeds automatically. Honestly, HPO must tried default hyper parameter setting, so user barely doesn't want to use it. And training resume with model weight trained during HPO, it doesn't spends much time generally. For these reasons, I think it's more natural to proceed training after HPO.

src/otx/utils/utils.py
vinnamkim, line 1: Please rename this to e.g., `utils/signal.py`

src/otx/engine/utils/hpo/__init__.py
vinnamkim, line 1: Why do you need the `src/otx/engine/utils/hpo` directory additionally and such many code lines unless already you owned the `src/otx/hpo` directory?

src/otx/engine/engine.py
```diff
@@ -205,6 +224,14 @@ def train(
                 otx train --data_root <DATASET_PATH> --config <CONFIG_PATH, str>
                 ```
         """
+        if run_hpo:
+            best_config, best_trial_weight = execute_hpo(**self._prepare_hpo_args(locals()))
```
vinnamkim:
```suggestion
            best_config, best_trial_weight = execute_hpo(engine=self, **locals()))
```

src/otx/engine/engine.py
harimkang: I think this should go in the form of `hpo_config`. I propose a way to make it easy for users to know the configuration values related to hpo. For example, I think it could be something like `src/otx/core/config/data.py::TilerConfig`.

src/otx/engine/utils/hpo/__init__.py
harimkang: Renaming hpo.py to something else seems like a good idea, at least in my opinion - it's too confusing just by looking at the name.

eunwoosh: I renamed it to hpo_api.py at https://github.com/openvinotoolkit/training_extensions/pull/2912/commits/57b7affda359b94f3eb8b9ad7cb74bca0671b195

src/otx/engine/engine.py
eunwoosh: I'll try to change as you said.

src/otx/engine/engine.py
eunwoosh: done at https://github.com/openvinotoolkit/training_extensions/pull/2912/commits/88348dfdadefa33e637267d1293b3a1432fceb9b

src/otx/utils/utils.py
eunwoosh: I moved signal related codes to signal.py at https://github.com/openvinotoolkit/training_extensions/pull/2912/commits/d594550e7ee7df9fe6d1de57592946df081aec45

src/otx/engine/utils/hpo/__init__.py
eunwoosh: `src/otx/hpo` is only about HPO algorithm. So it's completely independent to existing OTX. It means that it doesn't know about how model is trained, which hyper parameters need to be changed, etc. To use that HPO algorithm, additional code is necessary and that's why `src/otx/engine/utils/hpo` exists. In macro viewpoint, it just instantiates HPO algorithms and execute `run_hpo_loop` w/ that. But it includes some details. For example, before instantiating HPO algorithm, it adjusts some HPO algorithm configurations, such as setting missing argument and adjusting batch size if max value of batch size search space is bigger than train set size, removing hyper parameter from optimization candidates if necessary, etc. And before running `engine.train` in HPO trials, prepare registering HPO callback, which report scores to HPO algorithm, change engine's work directory to temporary directory, resume training if available, etc. This is why there is some of code size in that.

src/otx/engine/engine.py
```diff
@@ -598,6 +605,26 @@ def from_config(cls, config_path: PathLike, data_root: PathLike | None = None, *
     # Property and setter functions provided by Engine.
     # ------------------------------------------------------------------------ #
 
+    @property
+    def work_dir(self) -> PathLike:
+        """Work directory."""
+        return self._work_dir
+
+    @work_dir.setter
+    def work_dir(self, val: PathLike) -> None:
```
harimkang: An argument named `val` seems odd. Could you rename it?
The same goes for device setter functions.

eunwoosh: done at https://github.com/openvinotoolkit/training_extensions/pull/2912/commits/f209dd6e9b7127a4ea0353c8d8c0b485baa6ef6f

src/otx/engine/engine.py
eunwoosh: I changed it as you said at https://github.com/openvinotoolkit/training_extensions/pull/2912/commits/8114a6ce66feda223fc892e494b6e9c5fb984f2c

src/otx/engine/utils/hpo/__init__.py
vinnamkim: If so move `src/otx/engine/utils/hpo` to `src/otx/engine/hpo`

src/otx/core/config/hpo.py
```diff
@@ -0,0 +1,33 @@
...
+    full_dataset_size: int | None = None
+    expected_time_ratio: int | float | None = 4
+    maximum_resource: int | float | None = None
+    subset_ratio: float | int | None = None
+    min_subset_size: int | None = None
+    prior_hyper_parameters: dict | list[dict] | None = None
+    acceptable_additional_time_ratio: float | int = 1.0
+    minimum_resource: int | float | None = None
+    reduction_factor: int | None = None
+    asynchronous_sha: bool = True
```
vinnamkim: Do not insert `None` to typing if there is no special reason. Remove `None` from typing and change `None` default values to their actual default values.

src/otx/engine/utils/hpo/__init__.py
eunwoosh: done at https://github.com/openvinotoolkit/training_extensions/pull/2912/commits/0429a96cb44f33613d0f19db666aecfb8526ae1b

src/otx/core/config/hpo.py
eunwoosh: done at https://github.com/openvinotoolkit/training_extensions/pull/2912/commits/cd014b8aff92f9109cd44eeffaae365912b7aca9.
I changed all of them to use default value of `HyperBand` argument except search_space. `search_space` is required arguments for `HyperBand` but when user runs HPO w/o `hpo_config`, learning rate and batch size are set as HPO candidates based on current learning rate and batch size. So I leave it as it is.

src/otx/engine/engine.py
```diff
@@ -206,6 +207,16 @@ def train(
                 otx train --data_root <DATASET_PATH> --config <CONFIG_PATH, str>
                 ```
         """
+        if run_hpo:
+            if hpo_config is None:
+                hpo_config = HpoConfig()
```
vinnamkim, line 212:
```suggestion
```

src/otx/engine/engine.py
```diff
@@ -156,6 +153,8 @@ def train(
         callbacks: list[Callback] | Callback | None = None,
         logger: Logger | Iterable[Logger] | bool | None = None,
         resume: bool = False,
+        run_hpo: bool = False,
+        hpo_config: HpoConfig | None = None,
```
vinnamkim, line 157:
```suggestion
        hpo_config: HpoConfig = HpoConfig(),
```

eunwoosh: image

Object instantiated as initial argument is reused continuously, which has risk to lead to unexpected result as above.

vinnamkim (approved):

sungmanc (approved):

harimkang (approved):

PR #2910, opened 2024-02-13
Fix to allow Scheduler to be configured as a list and fix a bug in the AdaptiveTrainingCallback

harimkang:
### Summary

- Fix wrong update the frequency of the warmup-scheduler (This is the cause of a bug from `AdaptiveTrainingCallback`. -> Longer warmup times in small-dataset)

| Model            | Dataset         | Before (map_50) | After (map_50) | seed |
|------------------|-----------------|-----------------|----------------|------|
| ATSS_MobilenetV2 | Pothole_small_0 |             0.0 |           0.46 |    0 |

- Fixes an issue where `overrides` would throw an error if `configs.callbacks` came in as `None`.
- Modify the optimizer and scheduler to receive them as a list.
- `WarmupReduceLROnPlateau` -> `LinearWarmupScheduler + ReduceLROnPlateau`.
- `WarmupPolynomialLR` -> `LinearWarmupScheduler + PolynomialLR`.
- It also modifies it to match the configuration based on OTX1.6 (develop).
    - `max_epochs`, `warmup_iters`, `early-stopping patience`

### Checklist

- [ ] I have added unit tests to cover my changes.​
- [ ] I have added integration tests to cover my changes.​
- [ ] I have added e2e tests for validation.
- [ ] I have added the description of my changes into CHANGELOG in my target branch (e.g., [CHANGELOG](https://github.com/openvinotoolkit/training_extensions/blob/develop/CHANGELOG.md) in develop).​
- [ ] I have updated the documentation in my target branch accordingly (e.g., [documentation](https://github.com/openvinotoolkit/training_extensions/tree/develop/docs) in develop).
- [ ] I have [linked related issues](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

### License

- [ ] I submit _my code changes_ under the same [Apache License](https://github.com/openvinotoolkit/training_extensions/blob/develop/LICENSE) that covers the project.
      Feel free to contact the maintainers if that's a concern.
- [ ] I have updated the license header for each file (see an example below).

```python
# Copyright (C) 2023 Intel Corporation
# SPDX-License-Identifier: Apache-2.0
```

PR #2917, opened 2024-02-14
Enable to select the metric at the configuration level

sungmanc:
### Notes,
Still, WIP for the pre-commit, unit, and integration tests. However, if you have opinions related to the design change, please feel free to leave your comments.

### This PR introduces,
* Users can select the metrics which are inherited from the torchmetrics at the configuration level.
* Somes metrics that need num_classes information automatically updated by checking the model's `num_classes` information.
* `HLabelAccuracy` doesn't need the complex hlabel information at the `init` level.
* `val_metric` and `test_metric` are integrated --> `metric`.
* All `OTXLitModule` will monitor the learning rate by using the key name of scheduler.

###  TODOs,
* Visualprompting task use the dict format for the metric. However, this PR can't support the dict format. It'll be handled at the later PR.

### Checklist

### License

```python
```


src/otx/algo/classification/deit_tiny.py
```diff
@@ -78,6 +78,9 @@ class DeitTinyForHLabelCls(ExplainableDeit, MMPretrainHlabelClsModel):
     """DeitTiny Model for hierarchical label classification task."""
 
     def __init__(self, num_classes: int, num_multiclass_heads: int, num_multilabel_classes: int) -> None:
+        self.num_multiclass_heads = num_multiclass_heads
```
sungmanc, line 81: This information is added to use the `Model.num_multiclass_heads` and `Model.num_multilabel_classes` information at the `Metric`.

src/otx/core/model/module/base.py
```diff
@@ -203,7 +203,7 @@ def load_state_dict(self, ckpt: dict[str, Any], *args, **kwargs) -> None:
     @property
     def lr_scheduler_monitor_key(self) -> str:
         """Metric name that the learning rate scheduler monitor."""
-        return "val/loss"
+        return self.scheduler[-1].monitor
```
sungmanc: scheduler[0] is might `WarmScheduler`, so need to check the `monitor` key of last scheduler.

goodsong81 (commented):
Nice to see the metric could be configured, but why don't we enable multiple metrics? The first one might be regarded as the representative metric.

sungmanc:
> Nice to see the metric could be configured, but why don't we enable multiple metrics? The first one might be regarded as the representative metric.

Thanks for your interest, you're right. we'll support the multiple metrics at the next PRs. This is the very first step of giving the freedom of metric-related things. 

Here is the plan, 
1. Enable the chaning the metric at the configuration level
2. Support the previous OTX metric, especially OTX SDK/API metrics
3. Support multiple metric

src/otx/cli/cli.py
```diff
@@ -293,12 +294,22 @@ def instantiate_classes(self) -> None:
         If it is, it instantiates the necessary classes such as config, datamodule, model, and engine.
         """
         if self.subcommand in self.engine_subcommands():
-            # For num_classes update, Model is instantiated separately.
+            # For num_classes update, Model and Metric are instantiated separately.
             model_config = self.config[self.subcommand].pop("model")
+            metric_config = self.config[self.subcommand].pop("metric")
```
harimkang, line 300: As I understand it, `metric` is an argument that depends on `train`, `test` in the subcommand.
model is an argument that belongs to `engine.__init__`, which is fine, but metric might not be. This needs to be checked.

src/otx/cli/cli.py
```diff
@@ -415,6 +448,18 @@ def set_seed(self) -> None:
 
             seed_everything(seed, workers=True)
 
+    def _patch_metric_num_classes(self, model: Namespace, metric_config: Namespace) -> None:
```
harimkang: It doesn't seem necessary to use so many functions for the metric - couldn't this be done at the same time as updating num_classes in instantiate_model?

src/otx/cli/cli.py
```diff
@@ -308,6 +319,28 @@ def instantiate_classes(self) -> None:
                 **engine_kwargs,
             )
 
+    def instantiate_metric(self, metric_config: Namespace) -> Metric | None:
```
harimkang, line 323: I think this function could also be merged with the `instantiate_model` function, what do you think?

src/otx/core/model/module/classification.py
```diff
@@ -39,32 +40,31 @@ def __init__(
         torch_compile: bool,
         optimizer: list[OptimizerCallable] | OptimizerCallable = lambda p: torch.optim.SGD(p, lr=0.01),
         scheduler: list[LRSchedulerCallable] | LRSchedulerCallable = torch.optim.lr_scheduler.ConstantLR,
+        metric: Metric = Accuracy,
```
harimkang: In my opinion, `OTXLitModule` should also accept a metric argument.
Also, common code like `metric.reset` should probably go in an `OTXLitModule`.

samet-akcay:
> > Nice to see the metric could be configured, but why don't we enable multiple metrics? The first one might be regarded as the representative metric.
> 
> Thanks for your interest, you're right. we'll support the multiple metrics at the next PRs. This is the very first step of giving the freedom of metric-related things.
> 
> Here is the plan,
> 
> 1. Enable the chaning the metric at the configuration level
> 2. Support the previous OTX metric, especially OTX SDK/API metrics
> 3. Support multiple metric

torchmetrics has a great feature called `MetricCollection` that makes it possible to stack as many metrics you want. LightningModule handles the rest 

https://lightning.ai/docs/torchmetrics/stable/pages/overview.html#metriccollection

src/otx/recipe/classification/h_label_cls/efficientnet_b0_light.yaml
```diff
@@ -18,6 +18,12 @@ scheduler:
     patience: 1
     monitor: val/accuracy
 
+metric:
+  class_path: otx.algo.classification.metrics.hlabel_accuracy.HLabelAccuracy
```
samet-akcay, line 22: I will propose a package hierarchy once the execution is done, but a quick sneak peak. Ideally, it would be easier to access via API if metrics are be located in `otx.metrics`

sungmanc: > I will propose a package hierarchy once the execution is done, but a quick sneak peak. Ideally, it would be easier to access via API if metrics are be located in `otx.metrics`

I 100% agree with you. We need to clean the folder architecture

src/otx/core/model/module/classification.py
```diff
@@ -220,14 +229,35 @@ def __init__(
...
+            metric(
+                num_multiclass_heads=self.model.num_multiclass_heads,
+                num_multilabel_classes=self.model.num_multilabel_classes,
+            )
+            if callable(metric)
+            else metric
+        )
+
+        # Temporary, TODO (sungmanc)
+        # OTX can't support the auto-configuration for the H-label classification
```
harimkang, line 231: In near future, we should support auto-configuration for H-Labels as well. To solve this, we need to add logic to automatically set `num_multiclass_heads`, `num_multilabel_classes`. After that logic is added, we can support auto-configuration.

sungmanc: I agree, adding the logic of what you said should be done. I think that could be handled at the next PR

jaegukhyun (dismissed):

src/otx/core/model/module/action_classification.py
```diff
@@ -30,32 +32,40 @@ def __init__(
         torch_compile: bool,
         optimizer: list[OptimizerCallable] | OptimizerCallable = lambda p: torch.optim.SGD(p, lr=0.01),
         scheduler: list[LRSchedulerCallable] | LRSchedulerCallable = torch.optim.lr_scheduler.ConstantLR,
+        metric: Metric = Accuracy,
```
vinnamkim: Define `MetricCallable` type such as `LRSchedulerCallable` or `OptimizerCallable`
```suggestion
        metric: list[MetricCallable] | MetricCallable = partial(Accuracy, task="multiclass"),
```

src/otx/core/model/module/action_classification.py
```diff
@@ -30,32 +32,40 @@ def __init__(
...
+        )
+
+        self.metric = (
+            metric(
+                task="multiclass",
+                num_classes=self.model.num_classes,
+            )
+            if isinstance(metric, partial)
+            else metric
         )
```
vinnamkim:
```suggestion
import inspect

...

        sig = inspect.signature()
        if len(sig.parameters) == 0:
             self.metric = metric()
        elif len(sig.parameters) == 1 and "num_classes" in sig.parameters:
             self.metric = metric(num_classes=self.model.num_classes)
        else:
             msg = f"Given metric is not an allowed MetricCallable, {sig}"
             raise ValueError(msg)
```

src/otx/core/model/module/base.py
vinnamkim: There is no guarantee that `self.scheduler[-1]` has `monitor` attribute and there seems no use of this property after https://github.com/openvinotoolkit/training_extensions/pull/2910. Please just remove this property.

src/otx/core/model/module/base.py
```diff
@@ -53,6 +54,7 @@ def __init__(
         torch_compile: bool,
         optimizer: list[OptimizerCallable] | OptimizerCallable = lambda p: torch.optim.SGD(p, lr=0.01),
         scheduler: list[LRSchedulerCallable] | LRSchedulerCallable = torch.optim.lr_scheduler.ConstantLR,
+        metric: Metric,
```
eunwoosh: It seems that all of tasks initialize `metric` itself again. If then, how about not receiving `metric` from argument and just setting it to `None` in `OTXLitModule`?

src/otx/algo/classification/deit_tiny.py
eunwoosh, line 81: It seems that all h-label model classes need to save `num_multiclass_heads` and `num_multilabel_classes`, how about passing it to super().__init__() and making it to save them as own attribute?

eunwoosh (commented):
Thanks for your work :) I leave some comments. please take a look.

src/otx/core/model/module/base.py
sungmanc: Thanks for the suggestion, Making the `metric=None` doesn't looks good to me since metric should be delivered (So, can't be None).  As you said, all tasks initialize the `metric` again, I don't want to cause any confusion to the readers that the metric could be None.

src/otx/algo/classification/deit_tiny.py
sungmanc: For the `num_multiclass_heads` and `num_multilabel_classes`, it will be removed at the next PR. So, I don't want to enhance the logic at this PR. What do you think?

src/otx/recipe/classification/h_label_cls/efficientnet_b0_light.yaml
sovrasov: Can we import `HLabelAccuracy` in __init__.py in metrics, and then use `otx.algo.classification.metrics.HLabelAccuracy` to reduce the complexity a bit?

src/otx/recipe/classification/h_label_cls/otx_deit_tiny.yaml
```diff
@@ -22,6 +22,12 @@ scheduler:
       patience: 1
       monitor: val/accuracy
 
+metric:
+  class_path: otx.algo.classification.metrics.hlabel_accuracy.HLabelAccuracy
+  init_args:
+    num_multiclass_heads: 0
+    num_multilabel_classes: 0
```
sovrasov, line 29: Why do we need to set fictive values here?

src/otx/core/model/module/action_classification.py
sungmanc: Done, https://github.com/openvinotoolkit/training_extensions/pull/2917/commits/da532871951921f22dc38a1c9b49f1f7e5130deb

src/otx/core/model/module/action_classification.py
sungmanc: Done, https://github.com/openvinotoolkit/training_extensions/pull/2917/commits/da532871951921f22dc38a1c9b49f1f7e5130deb. Supporting the list of metric will be followed.

src/otx/recipe/classification/h_label_cls/otx_deit_tiny.yaml
sungmanc: `HLabelAccuracy` should take the `num_multiclass_heads` and `num_multilabel_classes`. However, they could be different according to the datasets. So, they will be automatically patched. That's the reason why I just set the default value as 0 for them.

src/otx/recipe/classification/h_label_cls/efficientnet_b0_light.yaml
sungmanc: I heard that @samet-akcay is planning to the restructuring for the algo folder, it could be handeled at that phase.

harimkang (approved):

jaegukhyun (approved):

eunwoosh (approved):

sovrasov (approved):

eugene123tw (approved):

src/otx/algo/metrices/__init__.py
vinnamkim, line 1: This should be moved to somewhere in `src/otx/core`

src/otx/core/model/module/action_detection.py
```diff
@@ -31,31 +34,25 @@ def __init__(
         torch_compile: bool,
         optimizer: list[OptimizerCallable] | OptimizerCallable = lambda p: torch.optim.SGD(p, lr=0.01),
         scheduler: list[LRSchedulerCallable] | LRSchedulerCallable = torch.optim.lr_scheduler.ConstantLR,
+        metric: MetricCallable | None = None,
```
vinnamkim, line 37: Why does it cannot have the default MetricCallable and should be `None` as default?

src/otx/core/model/module/detection.py
```diff
@@ -32,32 +35,25 @@ def __init__(
...
-    def on_test_epoch_start(self) -> None:
-        """Callback triggered when the test epoch starts."""
-        self.test_metric.reset()
+        param_dict = {}
+        if metric:
+            sig = inspect.signature(metric)
+            for name, param in sig.parameters.items():
+                param_dict[name] = param.default
+            param_dict.pop("kwargs")
+            metric = metric(**param_dict)  # type: ignore[call-arg]
```
vinnamkim, line 54: What's the reason that this is created at `on_test_epoch_start()`, not its `__init__()`?

src/otx/core/model/module/action_detection.py
sungmanc: For example, at the export, predict, we don't need add metric.

src/otx/core/model/module/detection.py
sungmanc: I'm sorry that I can't get your point. I think it's on the __init__ function.

src/otx/algo/metrices/__init__.py
sungmanc: About this, I heard that Samet is designing the new file architecture, so I just want to keep it as-is.

src/otx/core/model/module/action_detection.py
harimkang: It's not necessary for things like exports, but wouldn't it be more convenient for API users to at least have this as the default matric when creating a model?
I think the first case of user scenario is to try training first. However, it would be inconvenient to always include a metric.

harimkang:
We have other workers waiting for this PR, so why don't we merge this for now? Any additional comments should be addressed in the next PR.

sungmanc:
I'll reflect on the remaining reviews after merging this PR.

PR #2918, opened 2024-02-14
Enable forward_explain() for classification

negvet:
### Checklist

### License

```python
```

### DIFF
--- a/src/otx/core/exporter/native.py
+++ b/src/otx/core/exporter/native.py
@@ -32,10 +32,12 @@ class OTXNativeModelExporter(OTXModelExporter):
         metadata: dict[tuple[str, str], str] | None = None,
         via_onnx: bool = False,
         onnx_export_configuration: dict[str, Any] | None = None,
+        output_names: list[str] | None = None,
     ) -> None:
-        super().__init__(input_size, mean, std, resize_mode, pad_value, swap_rgb, metadata)
+        super().__init__(input_size, mean, std, resize_mode, pad_value, swap_rgb, metadata, output_names)
         self.via_onnx = via_onnx
         self.onnx_export_configuration = onnx_export_configuration if onnx_export_configuration is not None else {}
+        self.onnx_export_configuration.update({"output_names": output_names})
 
     def to_openvino(
         self,
--- a/src/otx/core/exporter/native.py
+++ b/src/otx/core/exporter/native.py
@@ -37,7 +37,8 @@ class OTXNativeModelExporter(OTXModelExporter):
         super().__init__(input_size, mean, std, resize_mode, pad_value, swap_rgb, metadata, output_names)
         self.via_onnx = via_onnx
         self.onnx_export_configuration = onnx_export_configuration if onnx_export_configuration is not None else {}
-        self.onnx_export_configuration.update({"output_names": output_names})
+        if output_names is not None:
+            self.onnx_export_configuration.update({"output_names": output_names})
 
     def to_openvino(
         self,
--- a/src/otx/core/model/entity/base.py
+++ b/src/otx/core/model/entity/base.py
@@ -137,7 +137,7 @@ class OTXModel(
     def forward_explain(
         self,
         inputs: T_OTXBatchDataEntity,
-    ) -> T_OTXBatchPredEntity | OTXBatchLossEntity:
+    ) -> T_OTXBatchPredEntity | T_OTXBatchPredEntityWithXAI | OTXBatchLossEntity:
         """Model forward explain function."""
         raise NotImplementedError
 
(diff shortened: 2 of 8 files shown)


negvet:
Before moving forward with these "forwards",
@vinnamkim @sovrasov please give me your opinions

src/otx/core/model/entity/classification.py
```diff
@@ -74,14 +74,66 @@ def head_forward_fn(self, x: torch.Tensor) -> torch.Tensor:
...
+    # def reset_explain_hook(self) -> None:
+    #     """Clear all history of explain records."""
+    #     self.explain_hook.reset()
+
+    def forward_explain(
+        self,
+        inputs: T_OTXBatchDataEntity,
+    ) -> T_OTXBatchPredEntity | OTXBatchLossEntity:
+        """Model forward function."""
+        self.model.explain_fn = self.get_explain_fn()
```
sovrasov, line 77: Can we pass explain fn as a parameter, rather creating one more attribute?

src/otx/core/model/entity/classification.py
```diff
@@ -74,14 +74,66 @@ def head_forward_fn(self, x: torch.Tensor) -> torch.Tensor:
...
+            else self.model(inputs)
+        )
+        return (
+            self._customize_outputs(outputs, inputs)
+            if self._customize_outputs != OTXModel._customize_outputs
+            else outputs
+        )
+
+    @staticmethod
+    def _forward_explain_image_classifier(
```
sovrasov, line 93: Which forward will we use for export? forward_explain can not be exported, so we should use a low level one

src/otx/core/model/entity/classification.py
negvet: I am afraid we cannot. You will get
`RuntimeError: Only tuples, lists and Variables are supported as JIT inputs/outputs. Dictionaries and strings are also accepted, but their usage is not recommended. Here, received an input of unsupported type: method` in this case during the export.

src/otx/core/model/entity/classification.py
negvet: `_forward_explain_image_classifier` will be used for export

src/otx/core/exporter/native.py
```diff
@@ -32,10 +32,12 @@ def __init__(
         metadata: dict[tuple[str, str], str] | None = None,
         via_onnx: bool = False,
         onnx_export_configuration: dict[str, Any] | None = None,
+        output_names: list[str] | None = None,
     ) -> None:
-        super().__init__(input_size, mean, std, resize_mode, pad_value, swap_rgb, metadata)
+        super().__init__(input_size, mean, std, resize_mode, pad_value, swap_rgb, metadata, output_names)
         self.via_onnx = via_onnx
         self.onnx_export_configuration = onnx_export_configuration if onnx_export_configuration is not None else {}
+        self.onnx_export_configuration.update({"output_names": output_names})
```
sovrasov: Check if output_names is none here?

src/otx/core/model/entity/base.py
```diff
@@ -51,6 +51,8 @@ def __init__(self, num_classes: int) -> None:
         self._label_info = LabelInfo.from_num_classes(num_classes)
         self.classification_layers: dict[str, dict[str, Any]] = {}
         self.model = self._create_model()
+        self.original_model_forward = None
+        self.explain_mode = False
```
sovrasov: This can be a public property to emphasize we have that switch in the class interface

src/otx/core/model/entity/base.py
```diff
@@ -120,6 +122,23 @@ def forward(
...
         )
 
+    def forward_explain(
+        self,
+        inputs: T_OTXBatchDataEntity,
+    ) -> T_OTXBatchPredEntity | OTXBatchLossEntity:
+        """Model forward explain function."""
+        raise NotImplementedError
+
+    def get_explain_fn(self) -> Callable:
```
sovrasov, line 154: Is that a workaround, introduced only because we have hooks now?

negvet: Yes, I propose to keep hooks as-is in this PR for simplicity and to better parallelize another tasks with @GalyaZalesskaya. We can refactor hooks (basically convert them to smth like Explainer objects) in the next PRs.

jaegukhyun (dismissed):

harimkang (dismissed):

GalyaZalesskaya (dismissed):

src/otx/engine/engine.py
```diff
@@ -364,6 +367,8 @@ def predict(
             datamodule = self.datamodule
         lit_module.meta_info = datamodule.meta_info
 
+        lit_module.model.explain_mode = explain
```
sovrasov, line 370: Should we restore the original state of `explain_mode` here? From CLI perspective I see this doesn't make sense

negvet: Currently we pass explain flag as a parameter to the engine methods, so we probably need to backtrack the state.

sovrasov (approved):

kprokofi (approved):

GalyaZalesskaya (approved):

PR #2920, opened 2024-02-15
Update weekly workflow to run perf tests

yunchu:
### Checklist

### License

```python
```

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -15,7 +15,7 @@ jobs:
     uses: ./.github/workflows/perf-speed.yml
     with:
       model-type: default
-      data-size: small
+      data-size: medium
       num-repeat: 1
       num-epoch: 3
       eval-upto: optimize
@@ -25,7 +25,7 @@ jobs:
     uses: ./.github/workflows/perf-accuracy.yml
     with:
       model-type: default
-      data-size: small
+      data-size: all
       num-repeat: 0
       num-epoch: 0
       eval-upto: optimize
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -10,41 +10,23 @@ on:
 permissions: read-all
 
 jobs:
-  Regression-Tests:
-    strategy:
-      fail-fast: false
-      matrix:
-        include:
-          - toxenv_task: "iseg"
-            test_dir: "tests/regression/instance_segmentation/test_instance_segmentation.py"
-            task: "instance_segmentation"
-          - toxenv_task: "iseg_t"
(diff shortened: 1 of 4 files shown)


.github/workflows/weekly.yml
```diff
@@ -10,41 +10,23 @@ on:
...
-      tests-dir: ${{ matrix.test_dir }}
-      runs-on: "['self-hosted', 'Linux', 'X64', 'dmount']"
-      task: ${{ matrix.task }}
-      timeout-minutes: 8640
-      upload-artifact: true
-      artifact-prefix: "weekly-test-results"
+      model-type: default
+      data-size: small
+      num-repeat: 1
+      num-epoch: 3
```
goodsong81: medium / 1 / 3 should be default for speed benchmark. Is this for just sanity check?

.github/workflows/weekly.yml
```diff
@@ -10,41 +10,23 @@ on:
...
+      eval-upto: optimize
+      artifact-prefix: weekly-perf-speed-benchmark
+  Performance-Accuracy-Tests:
+    name: Performance-Accuracy-py310
+    uses: ./.github/workflows/perf-accuracy.yml
+    with:
+      model-type: default
+      data-size: small
+      num-repeat: 0
+      num-epoch: 0
```
goodsong81: Similar here. (all / 0 / 0)

.github/workflows/weekly.yml
yunchu: I want to see that the workflow is working as expected using the small set (to reduce time) then set them with required configuration later.

goodsong81 (dismissed):

eunwoosh (dismissed):

eunwoosh:
LGTM. I have one question. I think speed test needs to use whole machines alone, but it seems not to be ensured currently. Do you have a plan to do that?

yunchu:
> LGTM. I have one question. I think speed test needs to use whole machines alone, but it seems not to be ensured currently. Do you have a plan to do that?

currently only a machine runs as a ci-runner so only one workflow would be ran at any moment. only concern to the speed tests are the h/w differences (cpu, i9-11900 vs i9-12900k) across the all ci-runners.

eunwoosh:
> > LGTM. I have one question. I think speed test needs to use whole machines alone, but it seems not to be ensured currently. Do you have a plan to do that?
> 
> currently only a machine runs as a ci-runner so only one workflow would be ran at any moment. only concern to the speed tests are the h/w differences (cpu, i9-11900 vs i9-12900k) across the all ci-runners.

That seems good, I think. we may be able to consider labelling single machine as speed test if necessary.

eunwoosh (approved):

goodsong81 (approved):

PR #2925, opened 2024-02-16
Add OV Tiler for 2.x

eugene123tw:
### Summary

* Incorporate OV tiler into OV detection and instance segmentation.
* Include tile image size in models featuring tiling (currently hardcoded; configurability of IR input size will be addressed in a separate PR).
* Note: A bug was found in openvino-model-api==0.1.9 for MaskRCNN. A hotfix PR has been merged (https://github.com/openvinotoolkit/model_api/pull/157). After upgrading to the lastest OV model API should not result in any accuracy drop.

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algo/detection/yolox.py
+++ b/src/otx/algo/detection/yolox.py
@@ -20,6 +20,7 @@ class YoloX(MMDetCompatibleModel):
         config = read_mmconfig(model_name=model_name)
         super().__init__(num_classes=num_classes, config=config)
         self.image_size = (1, 3, 640, 640)
+        self.tile_image_size = self.image_size
 
     @property
     def _export_parameters(self) -> dict[str, Any]:
@@ -44,6 +45,7 @@ class YoloXTiny(YoloX):
     def __init__(self, num_classes: int) -> None:
         super().__init__(num_classes=num_classes, variant="tiny")
         self.image_size = (1, 3, 416, 416)
+        self.tile_image_size = (1, 3, 640, 640)
 
     @property
     def _export_parameters(self) -> dict[str, Any]:
--- a/src/otx/algo/detection/yolox.py
+++ b/src/otx/algo/detection/yolox.py
@@ -45,7 +45,7 @@ class YoloXTiny(YoloX):
     def __init__(self, num_classes: int) -> None:
         super().__init__(num_classes=num_classes, variant="tiny")
         self.image_size = (1, 3, 416, 416)
-        self.tile_image_size = (1, 3, 640, 640)
+        self.tile_image_size = self.image_size
 
     @property
     def _export_parameters(self) -> dict[str, Any]:
--- a/src/otx/core/model/entity/base.py
+++ b/src/otx/core/model/entity/base.py
@@ -342,7 +342,8 @@ class OVModel(OTXModel, Generic[T_OTXBatchDataEntity, T_OTXBatchPredEntity]):
             output_dict[idx] = result
 
         numpy_inputs = self._customize_inputs(inputs)["inputs"]
-        if self.async_inference:
+        # NOTE: tiler has its own async impl
+        if self.async_inference and not self.datamodule_config.tile_config.enable_tiler:
             output_dict: dict[int, NamedTuple] = {}
             self.model.set_callback(_callback)
             for idx, im in enumerate(numpy_inputs):
--- a/src/otx/core/model/entity/base.py
+++ b/src/otx/core/model/entity/base.py
@@ -5,6 +5,7 @@
 
 from __future__ import annotations
 
+import contextlib
 import json
 import warnings
 from abc import abstractmethod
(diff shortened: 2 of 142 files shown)


src/otx/core/model/entity/detection.py
```diff
@@ -327,7 +330,19 @@ def __init__(
...
             use_throughput_mode,
             model_api_configuration,
+            tile_config,
         )
+        if self.tile_config is not None and self.tile_config.enable_tiler:
+            log.info(
+                f"Enable tiler with tile size: {self.tile_config.tile_size} and overlap: {self.tile_config.overlap}",
+            )
+            ov_tile_config = {
+                "tile_size": self.tile_config.tile_size[0],
```
sovrasov: Are these parameters also added to the model meta info on export?
Other MAPI clients can rely only on information coming from .xml

src/otx/engine/engine.py
```diff
@@ -301,7 +301,11 @@ def test(
         is_ir_ckpt = Path(str(checkpoint)).suffix in [".xml", ".onnx"]
         if is_ir_ckpt and not isinstance(model, OVModel):
             datamodule = self._auto_configurator.get_ov_datamodule()
-            model = self._auto_configurator.get_ov_model(model_name=str(checkpoint), meta_info=datamodule.meta_info)
+            model = self._auto_configurator.get_ov_model(
```
harimkang: This change does not affect users of the `openvino_model.yaml` configuration.
But in this PR, there are no changes to detection's `openvino_model.yaml`, was that taken into account?

src/otx/algo/detection/yolox.py
```diff
@@ -44,6 +45,7 @@ class YoloXTiny(YoloX):
     def __init__(self, num_classes: int) -> None:
         super().__init__(num_classes=num_classes, variant="tiny")
         self.image_size = (1, 3, 416, 416)
+        self.tile_image_size = (1, 3, 640, 640)
```
vinnamkim: Why are `image_size` and `tile_image_size` different? I know that we are almost all dealing with CNNs which are translation invariant, so that there seem not much side effects from it. However, it cannot be guaranteed 100%. I don't think that we do not need to take a risk without solid reasons.

src/otx/core/model/entity/base.py
```diff
@@ -333,13 +334,15 @@ def __init__(
         max_num_requests: int | None = None,
         use_throughput_mode: bool = True,
         model_api_configuration: dict[str, Any] | None = None,
+        tile_config: TileConfig | None = None,
```
vinnamkim: Like this comment https://github.com/openvinotoolkit/training_extensions/pull/2925/files#r1499081685, please make OpenVINO IR self-contained about tiling configuration. There should be no external injection for it like OTX 1.x.

src/otx/core/model/entity/detection.py
```diff
@@ -338,6 +353,16 @@ def _customize_outputs(
...
         labels = []
+
+        # some OMZ model requires to shift labels
+        first_label = (
+            self.model.model.get_label_name(0)
+            if isinstance(self.model, DetectionTiler)
+            else self.model.get_label_name(0)
+        )
+
+        label_shift = 1 if first_label == "background" else 0
```
vinnamkim, line 371: Please warn to an user if the label shift happens.

src/otx/core/model/entity/detection.py
eugene123tw: I added tile parameters to IR, so now OTX OV models read them from IR.

src/otx/algo/detection/yolox.py
eugene123tw: I'm adjusting tile image size for YOLOX-Tiny while keeping the modifications to image size unchanged for MaskRCNN.

The resolution of an image affects both accuracy and processing speed. Generally, smaller resolutions lead to faster processing but reduced accuracy. By employing various sizes, we aim to strike a balance between speed and accuracy, as supported by prior experimentation.

src/otx/core/model/entity/base.py
eugene123tw: I made changes accordingly.

src/otx/engine/engine.py
eugene123tw: I'll try my best to clarify, although I'm not entirely sure I understand your question.

Regarding `openvino_model` recipe, from a user perspective, enabling tiling during training is essential, as some tiling parameters are adjusted within the tile adaptor during training. When exporting the model, one can simply utilize the config.yaml saved post-training for model export, ensuring that all tile configurations are preserved within IR for OV inferencing.

Creating a separate openvino_model recipe for tiling purposes is feasible. However, it's useless if the model hasn't been trained with tiling. In such cases, exporting with this config serves no practical purpose. Furthermore, since tile adaptor dynamically modifies tile size, it poses a challenge to inject these alterations to openvino_model recipe.

Given that tiling must be enabled during both training and testing phases, I believe the optimal approach is to export using the checkpoint config yaml file.

src/otx/core/model/entity/detection.py
eugene123tw: Warning added: https://github.com/openvinotoolkit/training_extensions/pull/2925/files#diff-21305702e6b48749df9867a5a7cd30ad21f2099e1055fb1d522e42b5fe03e6e0R382

src/otx/core/model/entity/base.py
```diff
@@ -341,6 +341,17 @@ def __init__(
         self.use_throughput_mode = use_throughput_mode
         self.model_api_configuration = model_api_configuration if model_api_configuration is not None else {}
         super().__init__(num_classes)
+        try:
+            tile_config = self.model.inference_adapter.get_rt_info(["tile_config"]).astype(dict)
```
sovrasov: should this be `["model_info", "tile_config"]`, or this version works as well?

src/otx/core/model/entity/detection.py
```diff
@@ -329,6 +340,26 @@ def __init__(
...
+        """Setup tiler for tile task.
+
+        Args:
+            tile_config (dict): Tile configuration.
+        """
+        log.info(
+            f"Enable tiler with tile size: {tile_config['tile_size'].value} \
+                and overlap: {tile_config['tiles_overlap'].value}",
+        )
+        ov_tile_config = {
```
sovrasov: Apparently, we don't need this config here, at least right now.
Currently, the flow looks like: IR -> read tile config from it -> pass the tile config to tiler to override tile config from IR which is identical to the one we loaded
Is this parameter for the future usaceses?

sovrasov (dismissed):

harimkang (approved):

jaegukhyun (approved):

sungmanc (approved):

sovrasov (approved):
@eugene123tw thanks for your effort!
Do you have any comparison data for torch vs OV accuracy?

PR #2934, opened 2024-02-20
Add OTX CLI Workspace enhancements

harimkang:
### Summary

- Clean up the OTX output. (Create subfolders in chronological order. This is similar to OTX1.5.)
image

- Modify `--engine.work_dir` to link to `--work_dir`.
- Modify `work_dir` and `data_root` to use absolute paths for reusability.
- Fix `work_dir` to affect overall loggers, callbacks, etc. in CLI configuration.
- Provide a `.latest` within the workspace.
- The `.latest` stores the address of the config and checkpoint files for that training based on the most recently run training.
- Use the config and checkpoint files in the `.latest` if `--config` (all command) and `--checkpoint` (test, predict, export, explain) are omitted when running commands in the workspace.

### Examples that work with cache
Case 1: If you run a command inside `otx-workspace` that is the output of the CLI, the
```shell
cd otx-workspace

otx train
otx test
otx export
```

Case 2: If you specify a `work_dir` that already exists
```shell
otx train --work_dir otx-workspace
```

### Checklist

### License

```python
```


for_developers/cli_guide.md
```diff
@@ -313,3 +313,35 @@ overrides:
...
+otx-workspace/
+    .cache/                       # As of the last training, we have a .cache with path information for config and checkpoint.
+    20240000_000000/              # Deliverables from OTX CLI
+    20240000_000001/              # Deliverables from OTX CLI Second-Trial
+```
+
+OTX considers the folder with .cache to be the root of the entire Workspace.
+So this uses `.cache/configs.txt` as the default config, and the ckpt pass from `.cache/latest_checkpoint.txt` as a checkpoint as needed.
+
+Case 1: If a user specify an output `work_dir` (An already existing workspace)
```
eunwoosh:
```suggestion
Case 1: If a user specifies an output `work_dir` (An already existing workspace)
```

src/otx/cli/utils/workspace.py
```diff
@@ -0,0 +1,33 @@
...
+
+from datetime import datetime, timezone
+from pathlib import Path
+
+
+class Workspace:
+    """Represents a workspace for the OTX application.
+
+    Args:
+        work_dir (str): The path to the workspace directory. Defaults to "./otx-workspace".
```
eunwoosh:
```suggestion
        work_dir (Path | str | None, optional): The path to the workspace directory. Defaults to "./otx-workspace".
```

src/otx/cli/utils/workspace.py
```diff
@@ -0,0 +1,33 @@
...
+        use_sub_dir (bool): Whether to use a subdirectory within the workspace. Defaults to True.
```
eunwoosh:
```suggestion
        use_sub_dir (bool, optional): Whether to use a subdirectory within the workspace. Defaults to True.
```

src/otx/cli/cli.py
```diff
@@ -35,6 +36,19 @@
     _ENGINE_AVAILABLE = False
 
 
+def absolute_path(path: str | Path | None) -> str | None:
```
eunwoosh: Could you move this function to 'otx/utils/utils.py'?

src/otx/cli/cli.py
```diff
@@ -401,18 +451,43 @@ def _prepare_subcommand_kwargs(self, subcommand: str) -> dict[str, Any]:
...
-    def save_config(self) -> None:
+    def save_config(self, work_dir: Path) -> None:
         """Save the configuration for the specified subcommand.
 
+        Args:
+            work_dir (Path): The working directory where the configuration file will be saved.
+
         The configuration is saved as a YAML file in the engine's working directory.
         """
+        # Change all Path-related values to absolute paths.
```
eunwoosh: Isn't it wrong comment?

src/otx/cli/cli.py
```diff
@@ -401,18 +451,43 @@ def _prepare_subcommand_kwargs(self, subcommand: str) -> dict[str, Any]:
...
+
+        Args:
+            work_dir (Path): The working directory where the configurations and checkpoint files are located.
+        """
+        cache_dir = work_dir.parent / ".cache"
+        cache_dir.mkdir(exist_ok=True)
+        with (cache_dir / "configs.txt").open("w") as f:
+            f.write(str((work_dir / "configs.yaml").resolve()))
+        checkpoint_dir = work_dir / "checkpoints"
+        ckpt_list = list(checkpoint_dir.glob("epoch_*.ckpt"))  # Convert generator to list
```
eunwoosh: As far as I know, 'epoch_*.ckpt' is best model weight and 'last.ckpt' is last model weight. Is it intended to use best model weight instead of last model weight?

src/otx/cli/cli.py
```diff
@@ -249,8 +278,24 @@ def add_subcommands(self) -> None:
...
+        if (self.cache_dir / "latest_checkpoint.txt").exists():
+            with (self.cache_dir / "latest_checkpoint.txt").open("r") as f:
+                latest_checkpoint = f.read()
+            parser.set_defaults(checkpoint=latest_checkpoint)
+            if "--print_config" not in sys.argv:
+                warn(f"Load default checkpoint from {latest_checkpoint}.", stacklevel=0)
+
+    def _set_default_config(self) -> dict:
         parser_kwargs = {}
+        if self.cache_dir.exists() and (self.cache_dir / "configs.txt").exists():
```
eunwoosh:
```suggestion
        if (self.cache_dir / "configs.txt").exists():
```

src/otx/cli/utils/workspace.py
```diff
@@ -0,0 +1,33 @@
...
+            if not (Path.cwd() / ".cache").exists():
+                # Without work_dir input & no .cache directory in root
+                self.work_dir = Path.cwd() / "otx-workspace"
+            else:
+                # If Path.cwd is workspace
+                self.work_dir = Path.cwd()
+        else:
+            self.work_dir = Path(work_dir)
+        if use_sub_dir:
+            timestamp = datetime.now(tz=timezone.utc).strftime("%Y%m%d_%H%M%S")
```
eunwoosh, line 29: How about adding sub-command as suffix? It's a little hard to recognize which commands is executed with this name.

src/otx/cli/cli.py
```diff
@@ -443,15 +518,20 @@ def run(self) -> None:
...
             fn_kwargs = self._prepare_subcommand_kwargs(self.subcommand)
             fn = getattr(self.engine, self.subcommand)
             try:
-                fn(**fn_kwargs)
+                result = fn(**fn_kwargs)
             except Exception:
                 self.console.print_exception(width=self.console.width)
-            self.save_config()
+            output_dir = self.engine.work_dir
+            if isinstance(result, Path):
```
eunwoosh: It seems that `result.parent` of `export` is same as `self.engine.work_dir`. What is it for?

eunwoosh (commented):
Thanks for your work! I wait this features to be updated :) I leave some comments. Please take a look.

for_developers/cli_guide.md
harimkang: Thanks - Done!
https://github.com/openvinotoolkit/training_extensions/pull/2934/commits/c2df0ed2df2be454f153a86ba4e43373ef889719

src/otx/cli/utils/workspace.py
harimkang: Thanks - Done!
https://github.com/openvinotoolkit/training_extensions/pull/2934/commits/c2df0ed2df2be454f153a86ba4e43373ef889719

src/otx/cli/utils/workspace.py
harimkang: Thanks - Done!
https://github.com/openvinotoolkit/training_extensions/pull/2934/commits/c2df0ed2df2be454f153a86ba4e43373ef889719

src/otx/cli/cli.py
harimkang: I don't see any examples available in core, API side at the moment, so I moved this function to `cli.utils` for now.
https://github.com/openvinotoolkit/training_extensions/pull/2934/commits/c2df0ed2df2be454f153a86ba4e43373ef889719

src/otx/cli/cli.py
harimkang: Thanks - Removed
https://github.com/openvinotoolkit/training_extensions/commit/c2df0ed2df2be454f153a86ba4e43373ef889719

src/otx/cli/cli.py
harimkang: The OTX workflow is always based on `epoch_*.ckpt`. All tests also currently use that checkpoint.

src/otx/cli/cli.py
harimkang: Thanks - Done!
https://github.com/openvinotoolkit/training_extensions/commit/c2df0ed2df2be454f153a86ba4e43373ef889719

src/otx/cli/utils/workspace.py
harimkang: This is not easy with the current CLI flow, I'll think about it a bit more and add it if there's a good way to do it.

src/otx/cli/cli.py
harimkang: I chose the more obvious way, because the path that export spits out is the most obvious output folder address based on export.
Depending on API implementation, `work_dir` and `export model output path` may change in the future, so don't let them affect here.

src/otx/cli/utils/workspace.py
eunwoosh: so, do you apply it later not in this PR? If then, let's resolve this comment.

src/otx/cli/cli.py
eunwoosh: IMHO, I think `self.engine.work_dir` is more obvious because it's engine's 'work directory' as it's name already explains. `export` just makes OV IR files in 'work directory' and return it's path and it can be changed later also. I understand engine.work_dir can be changed in API use case, but isn't workspace only for CLI? If then, we don't have to consider that `engine.work_dir` is changed in the middle.

src/otx/cli/utils/workspace.py
harimkang: Yes, it's unlikely to be resolved in this PR.

src/otx/cli/cli.py
harimkang: If so, let's revert this for now. - https://github.com/openvinotoolkit/training_extensions/pull/2934/commits/234dcdede5fe1c19844f989cc2e77c8685d51e50

eunwoosh (dismissed):

sungmanc (dismissed):

for_developers/cli_guide.md
```diff
@@ -313,3 +313,35 @@ overrides:
...
         patience: 3
 ```
+
+## How to use OTX Workspace
+
+If we run a typical Training example, we'll have a folder like the one below as output.
+
+```console
+otx-workspace/
+    .cache/                       # As of the last training, we have a .cache with path information for config and checkpoint.
```
vinnamkim:
```suggestion
    .latest/                       # As of the last training, we have a .latest with path information for config and checkpoint.
```
I'm feeling that `.cache` is not an appropriate naming for it. Consider another name more suitable, e.g., `.latest_conf.`.

src/otx/cli/cli.py
```diff
@@ -250,8 +267,24 @@ def add_subcommands(self) -> None:
             self._subcommand_parsers[subcommand] = sub_parser
             parser_subcommands.add_subcommand(subcommand, sub_parser, help=description)
 
-    def _set_default_config_from_auto_configurator(self) -> dict:
+    def _load_cache_ckpt(self, parser: ArgumentParser) -> None:
+        if (self.cache_dir / "latest_checkpoint.txt").exists():
```
vinnamkim:
```suggestion
        if not (self.cache_dir / "latest_checkpoint.txt").exists():
            return
```

src/otx/cli/cli.py
```diff
@@ -250,8 +267,24 @@ def add_subcommands(self) -> None:
...
+            with (self.cache_dir / "latest_checkpoint.txt").open("r") as f:
+                latest_checkpoint = f.read()
+            parser.set_defaults(checkpoint=latest_checkpoint)
+            if "--print_config" not in sys.argv:
+                warn(f"Load default checkpoint from {latest_checkpoint}.", stacklevel=0)
+
+    def _set_default_config(self) -> dict:
         parser_kwargs = {}
+        if (self.cache_dir / "configs.txt").exists():
```
vinnamkim:
```suggestion
        if not (self.cache_dir / "configs.txt").exists():
             return {}
```

src/otx/cli/cli.py
```diff
@@ -431,18 +469,42 @@ def _prepare_subcommand_kwargs(self, subcommand: str) -> dict[str, Any]:
...
+        cache_dir = work_dir.parent / ".cache"
+        cache_dir.mkdir(exist_ok=True)
+        with (cache_dir / "configs.txt").open("w") as f:
+            f.write(str((work_dir / "configs.yaml").resolve()))
+        checkpoint_dir = work_dir / "checkpoints"
+        ckpt_list = list(checkpoint_dir.glob("epoch_*.ckpt"))  # Convert generator to list
+        if ckpt_list:
+            latest_checkpoint_file = max(ckpt_list, key=lambda p: p.stat().st_mtime)
+            with (cache_dir / "latest_checkpoint.txt").open("w") as f:
+                f.write(str(latest_checkpoint_file.resolve()))
```
vinnamkim: Why not just creating a soft link for it?

src/otx/cli/utils/workspace.py
```diff
@@ -0,0 +1,33 @@
...
+    def __init__(self, work_dir: Path | str | None = None, use_sub_dir: bool = True):
+        if work_dir is None:
+            if not (Path.cwd() / ".cache").exists():
+                # Without work_dir input & no .cache directory in root
+                self.work_dir = Path.cwd() / "otx-workspace"
+            else:
+                # If Path.cwd is workspace
+                self.work_dir = Path.cwd()
+        else:
+            self.work_dir = Path(work_dir)
```
vinnamkim:
```suggestion
    def __init__(self, work_dir: Path = Path.cwd(), use_sub_dir: bool = True):
            self.work_dir = (
                work_dir / "otx-workspace"
                # Without work_dir input & no .cache directory in root
                if work_dir == Path.cwd() and not (work_dir / ".cache").exists()            
                else work_dir
            )
```

I'm feeling a little bit tired to say this kind of comment again and again https://github.com/openvinotoolkit/training_extensions/pull/2912#discussion_r1491842254.
What's the reason that this happens again and again?

for_developers/cli_guide.md
harimkang: Thank you for the great suggestion. - https://github.com/openvinotoolkit/training_extensions/pull/2934/commits/6b560edf42ebdfdbf9bc3268df6adcb375a47f07

src/otx/cli/cli.py
harimkang: Done - https://github.com/openvinotoolkit/training_extensions/pull/2934/commits/6b560edf42ebdfdbf9bc3268df6adcb375a47f07

src/otx/cli/cli.py
harimkang: I left this as is so that if don't have the latest, CLI can load a different default config. Instead, I added a comment below

src/otx/cli/cli.py
harimkang: Thanks, Done - https://github.com/openvinotoolkit/training_extensions/pull/2934/commits/6b560edf42ebdfdbf9bc3268df6adcb375a47f07

src/otx/cli/utils/workspace.py
harimkang: I didn't pay attention to the other reviewer's comment in the other review. Anyway, I've fixed it, thanks. - https://github.com/openvinotoolkit/training_extensions/pull/2934/commits/6b560edf42ebdfdbf9bc3268df6adcb375a47f07

for_developers/cli_guide.md
```diff
@@ -313,3 +313,35 @@ overrides:
...
         patience: 3
 ```
+
+## How to use OTX Workspace
+
+If we run a typical Training example, we'll have a folder like the one below as output.
+
+```console
+otx-workspace/
+    .latest/                       # As of the last training, we have a .latest with path information for config and checkpoint.
```
jaegukhyun, line 323: Is there any reason we should hide the latest config?

vinnamkim (approved):

jaegukhyun (approved):

for_developers/cli_guide.md
harimkang: I didn't get it your comment, .latest is relative to the most recently trained work_dir.

jaegukhyun: Why should be the directory name .latest, not latest. User can't see .latest if they just type ls

harimkang: This is to omit some arguments in the CLI. I didn't think there would be many examples for users to use latest folder.
I didn't think about showing it at the same level as the other `work_dirs`.
What do you think, do you think it should be exposed via ls?

jaegukhyun: I'm not sure, but let's leave it as it is

sungmanc (approved):

PR #2955, opened 2024-02-21
Enable export for visual prompting models

sungchul1:
### Summary

This PR includes:
- [x] enable `otx export`
    - [x] finetuning
    - [x] zero-shot
- [x] enable IR inference
    - [x] finetuning
    - [x] zero-shot
- [x] enable standalone `infer` (not given reference feats, automatically execute `learn` first)
    - [x] torch
    - [x] OVIR

### Checklist

### License

```python
```


src/otx/core/exporter/mmdeploy.py
```diff
@@ -121,6 +127,7 @@ def to_onnx(
         base_model_name: str = "exported_model",
         precision: OTXPrecisionType = OTXPrecisionType.FP32,
         embed_metadata: bool = True,
+        export_args: dict[str, Any] | None = None,  # TODO (sungchul): how to apply this argument? # noqa: TD003
```
jaegukhyun: Should we link this args to 
image
?

sungchul1:
```suggestion
        export_args: dict[str, Any] | None = None,  # TODO (sungchul): how to apply this argument? # noqa: TD003
        ...
        deploy_cfg = self._prepare_onnx_cfg()
        if export_args is not None:
            deploy_cfg.merge_from_dict(export_args)
        save_path = self._cvt2onnx(model, output_dir, base_model_name, deploy_cfg)
```
Is it okay?

src/otx/algo/visual_prompting/openvino_models.py
```diff
@@ -0,0 +1,151 @@
...
+    def preprocess(self, inputs: dict[str, Any]) -> list[dict[str, Any]]:
+        """Preprocess prompts."""
+        processed_prompts: list[dict[str, Any]] = []
+        idx: int = 0
+        for prompt_name in ["bboxes", "points"]:
+            prompts = inputs.get(prompt_name, None)
+            if prompts is None:
+                continue
+            for prompt in prompts:
+                label = inputs["labels"][idx]
```
wonjuleee: does this mean all prompts have different label?

src/otx/algo/visual_prompting/openvino_models.py
```diff
@@ -0,0 +1,151 @@
...
+        coords[..., 0] = coords[..., 0] * (new_w / old_w)
+        coords[..., 1] = coords[..., 1] * (new_h / old_h)
+        return coords
+
+    def _get_preprocess_shape(self, old_h: int, old_w: int, image_size: int) -> tuple[int, int]:
+        """Compute the output size given input size and target image size."""
+        scale = image_size / max(old_h, old_w)
+        new_h, new_w = old_h * scale, old_w * scale
+        new_w = int(new_w + 0.5)
+        new_h = int(new_h + 0.5)
```
wonjuleee, line 121: why is this required?

src/otx/algo/visual_prompting/openvino_models.py
sungchul1: They can be.

src/otx/algo/visual_prompting/openvino_models.py
sungchul1: `int` does flooring floating point. If the value is 3.9, we can assume it may be 4 for integer, but it's actually 3.

src/otx/core/model/module/visual_prompting.py
```diff
@@ -243,22 +246,80 @@ def _inference_step(
...
     def on_train_epoch_end(self) -> None:
         """Skip on_train_epoch_end unused in zero-shot visual prompting."""
+        if self.model.save_outputs:
+            reference_info = {
+                "reference_feats": self.model.reference_feats,
+                "used_indices": self.model.used_indices,
+            }
+            # save reference info
+            path_reference_info: Path = (
+                self.model.root_reference_info / time.strftime("%Y%m%d_%H%M%S") / "reference_info.pt"
```
wonjuleee, line 307: It might be good for making static variable for `reference_info.pt` and `pickle` extension in the future.

src/otx/core/model/entity/visual_prompting.py
```diff
@@ -30,46 +49,920 @@ class OTXVisualPromptingModel(
...
+            assert mask2.ndim == 2  # noqa: S101
+            intersection = np.logical_and(mask1, mask2).sum().item()
+            union = np.logical_or(mask1, mask2).sum().item()
+
+            # Avoid division by zero
+            if union == 0:
+                return 0.0
+            return intersection / union
+
+        for (label, masks), (other_label, other_masks) in product(predicted_masks.items(), predicted_masks.items()):
```
wonjuleee, line 834: Is this typo? `product(predicted_masks.items(), predicted_masks.items())`

sungchul1: No, it is to check all labeled masks. It will be skipped if `other_label <= label` at the next line.

src/otx/core/model/module/visual_prompting.py
sungchul1: You mean, override them every `learn`?

src/otx/core/exporter/mmdeploy.py
sungchul1: I updated it at https://github.com/openvinotoolkit/training_extensions/pull/2955/commits/2708d7ba16465e35cd7665da3043223ecc2dde3f.
Please take a look and let me know if there is another better way.

src/otx/algo/visual_prompting/zero_shot_segment_anything.py
```diff
@@ -162,11 +172,12 @@ def _point_selection(
         # sample fg_coords_scores matched by matched_matrix
         matched_grid = fg_coords_scores.unsqueeze(1) * matched_matrix.unsqueeze(-1)
 
-        # sample the highest score one of the samples that are in the same grid
-        points_scores = matched_grid[matched_grid[..., -1].argsort(dim=0, descending=True)[0]].diagonal().T
+        matched_indices = matched_grid[..., -1].topk(k=1, dim=0, largest=True)[1][0].to(torch.int64)
```
wonjuleee, line 153: minor comment: with `topk` operation output might be already integer.

src/otx/algo/visual_prompting/zero_shot_segment_anything.py
```diff
@@ -162,11 +172,12 @@ def _point_selection(
...
+        points_scores = matched_grid[matched_indices].diagonal().T
 
         # sort by the highest score
-        points_scores = points_scores[torch.argsort(points_scores[:, -1], descending=True)]
+        sorted_points_scores_indices = torch.argsort(points_scores[:, -1], descending=True).to(torch.int64)
```
wonjuleee, line 157: minor comment: with `argsort` operation output might be already integer.

src/otx/algo/visual_prompting/zero_shot_segment_anything.py
sungchul1: Thanks, it was updated for export phase, but it is unnecessary anymore as you said. I'll update it in the next commit or PR.

src/otx/algo/visual_prompting/zero_shot_segment_anything.py
sungchul1: Thanks, it was updated for export phase, but it is unnecessary anymore as you said. I'll update it in the next commit or PR.

src/otx/algo/visual_prompting/openvino_models.py
wonjuleee: Sorry, I am asking that this is intended or not. In the code, it looks like to enforce all should be different label.

src/otx/core/exporter/base.py
```diff
@@ -68,6 +68,7 @@ def export(
         base_model_name: str = "exported_model",
         export_format: OTXExportFormatType = OTXExportFormatType.OPENVINO,
         precision: OTXPrecisionType = OTXPrecisionType.FP32,
+        export_args: dict[str, Any] | None = None,
```
wonjuleee: Personally, this should be cautious to introduce new arguments. If it is limited to the specific arguments, e.g., `input`, `input_model`, or `example_input`, it could be better to introduce them directly instead of using dict type `export_args`. Is there any example or candiate of `export_args` during `openvino.convert_model`?

src/otx/algo/visual_prompting/openvino_models.py
sungchul1: I intended. Each prompt has their own label and it can be same or different with each other.

src/otx/core/exporter/base.py
sungchul1: `self.to_onnx` was only able to support naive dummy image input for simple tasks, which was not for other complicate tasks like visual prompting, and static arguments such as `verbose` and `opset`, not inputs.
`export_args` is to support dynamic arguments like inputs in `torch.onnx.export` in `self.to_onnx`.
And it is expanded to other exports to fit argument information.
For `openvino.convert_model`, we can use this argument like below
https://github.com/openvinotoolkit/training_extensions/blob/2708d7ba16465e35cd7665da3043223ecc2dde3f/src/otx/core/exporter/native.py#L81-L87

jaegukhyun (dismissed):

src/otx/algo/visual_prompting/openvino_models.py
```diff
@@ -0,0 +1,151 @@
...
+        return "upscaled_masks"
+
+    def preprocess(self, inputs: dict[str, Any]) -> list[dict[str, Any]]:
+        """Preprocess prompts."""
+        processed_prompts: list[dict[str, Any]] = []
+        idx: int = 0
+        for prompt_name in ["bboxes", "points"]:
+            prompts = inputs.get(prompt_name, None)
+            if prompts is None:
+                continue
```
vinnamkim:
```suggestion
            if (prompts := inputs.get(prompt_name, None)) is None:
                continue
```

src/otx/algo/visual_prompting/openvino_models.py
```diff
@@ -0,0 +1,151 @@
...
+            for prompt in prompts:
```
vinnamkim:
```suggestion
            for idx, prompt in enumerate(prompts):
```

src/otx/algo/visual_prompting/openvino_models.py
```diff
@@ -0,0 +1,151 @@
...
+        return parameters
+
+    def preprocess(self, inputs: np.ndarray) -> tuple[dict[str, np.ndarray], dict[str, Any]]:
+        """Update meta for image encoder."""
+        dict_inputs, meta = super().preprocess(inputs)
+        meta["resize_type"] = self.resize_type
+        return dict_inputs, meta
+
+
+class Decoder(SegmentationModel):
```
vinnamkim:
```suggestion
class VisualPromptDecoder(SegmentationModel):
```

src/otx/algo/visual_prompting/zero_shot_segment_anything.py
```diff
@@ -47,66 +54,64 @@ def set_default_thresholds(self, default_threshold_reference: float, default_thr
...
+            threshold = torch.tensor([[0.0]], dtype=torch.float32)
+        if num_bg_points is None:
+            num_bg_points = torch.tensor([[1]], dtype=torch.int64)
+
+        ori_shape = ori_shape.squeeze()
+        threshold = threshold.squeeze()
+        num_bg_points = num_bg_points.squeeze()
 
-        target_feat = image_embedding.squeeze()
+        target_feat = image_embeddings.squeeze()
```
vinnamkim: Can you add comments and checkers to inform what is the desired tensor shape after squeeze?

src/otx/algo/visual_prompting/zero_shot_segment_anything.py
```diff
@@ -124,14 +129,19 @@ def _point_selection(
...
+        if threshold is None:
+            threshold = torch.tensor([0.0], dtype=torch.float32)
+        if num_bg_points is None:
+            num_bg_points = torch.tensor(1, dtype=torch.int64)
+
         _, w_sim = mask_sim.shape
 
         # Top-last point selection
-        bg_indices = mask_sim.flatten().topk(num_bg_points[0, 0], largest=False)[1]
+        bg_indices = mask_sim.flatten().topk(num_bg_points.item(), largest=False)[1]
```
vinnamkim: If `num_bg_points` is not a CPU tensor, calling `.item()` makes it to be transferred from accelerator to main memory and the thread will be being blocked to wait for it. Is this change necessary?

src/otx/algo/visual_prompting/zero_shot_segment_anything.py
```diff
@@ -347,21 +325,19 @@ def learn(
                     )
                     default_threshold_reference -= 0.05
 
-                self.reference_info["reference_feats"][label] = ref_feat.detach().cpu()
-                self.reference_info["used_indices"].add(int(label))
+                reference_feats[label] = ref_feat.detach().cpu()
+                used_indices = torch.cat((used_indices, torch.tensor([label])), dim=0)
```
vinnamkim: Can we do this such as 
```suggestion
                used_indices += [torch.tensor([label])]
...
used_indices = torch.cat(used_indices, dim=0).unique()
```
? Concatenating in the loop requires `O(n^2) memory allocation and copy`.

src/otx/core/exporter/native.py
```diff
@@ -46,14 +46,16 @@ def to_openvino(
...
     ) -> Path:
         """Export to OpenVINO Intermediate Representation format.
 
         In this implementation the export is done only via standard OV/ONNX tools.
         """
-        dummy_tensor = torch.rand(self.input_size).to(next(model.parameters()).device)
-
         if self.via_onnx:
+            if export_args is None:
+                export_args = {"args": torch.rand(self.input_size).to(next(model.parameters()).device)}
```
vinnamkim: Why is this needed?

src/otx/core/model/entity/base.py
```diff
@@ -328,7 +339,7 @@ class OVModel(OTXModel, Generic[T_OTXBatchDataEntity, T_OTXBatchPredEntity, T_OT
     def __init__(
         self,
         num_classes: int,
-        model_name: str,
+        model_name: str | dict[str, str],
```
vinnamkim: Why is this change needed?

src/otx/core/model/entity/visual_prompting.py
vinnamkim, line 1: Are these many functions in this file `numpy` version of the previous PyTorch dedicated codes?

vinnamkim:
You must downsize your PR size in the next work

src/otx/core/exporter/base.py
kprokofi: I think we need to extend export_args description emphasizing that we can modify sample input. Also, if it is only needed for example_input, then I propose to create `example_input` parameter instead of `export_args`. `export_args` has unclear naming and unclear purpose. "input" always can be obtained from self.input_size, no need to add it in arguments.

src/otx/core/exporter/native.py
kprokofi: I don't like this "export_args " approach everywhere. Let's consider to either remove it or change to "sample_input" if input needs to be modified. But, I can't understand in which cases we should modify dummy input for export?

src/otx/core/model/module/anomaly/anomaly_lightning.py
```diff
@@ -85,7 +85,7 @@ def __init__(
             metadata=metadata,
         )
 
-    def to_openvino(
+    def to_openvino(  # type: ignore[override]
```
kprokofi: What is the problem with overriding?

sungchul1: Due to `export_args`. I'll update it for anomaly default values.

sungchul2:
https://github.com/openvinotoolkit/training_extensions/pull/2955#discussion_r1511231706
@kprokofi For visual prompting, there are two types of inputs, one is an normal image for image encoder like other tasks and the other is image embeddings from image encoder and 4 args of prompt inputs for decoder.
https://github.com/openvinotoolkit/training_extensions/blob/2708d7ba16465e35cd7665da3043223ecc2dde3f/src/otx/core/model/entity/visual_prompting.py#L85-L104

src/otx/core/exporter/native.py
sungchul1: @vinnamkim If `export_args` is None, it is required to get ov input information for `openvino.convert_model`.
https://github.com/openvinotoolkit/training_extensions/blob/2708d7ba16465e35cd7665da3043223ecc2dde3f/src/otx/core/exporter/native.py#L71-L75

sungchul1: @kprokofi For visual prompting, there are two types of inputs, one is a normal image for image encoder like other tasks and the other is image embeddings from image encoder and 4 args of prompt inputs for decoder.
https://github.com/openvinotoolkit/training_extensions/blob/2708d7ba16465e35cd7665da3043223ecc2dde3f/src/otx/core/model/entity/visual_prompting.py#L85-L104

src/otx/core/model/entity/base.py
sungchul1: Due to `_model_name` set for visual prompting models.
https://github.com/openvinotoolkit/training_extensions/blob/2708d7ba16465e35cd7665da3043223ecc2dde3f/src/otx/core/model/entity/visual_prompting.py#L192-L204

src/otx/core/model/entity/visual_prompting.py
sungchul1: In `src/otx/algo/visual_prompting/segment_anything.py` and `src/otx/algo/visual_prompting/zero_shot_segment_anything.py`, there are so many inference functions with pytorch and they are required for ov inference.

sungchul1:
> You must downsize your PR size in the next work

Sorry for the inconvenience. I'll consider it next time.

src/otx/core/exporter/base.py
sungchul1: I'm not sure about your comment, but I just patched previous implementations.
https://github.com/openvinotoolkit/training_extensions/blob/c8c3cae11d0702e707fd17132aca2de90e61a0a2/src/otx/core/exporter/native.py#L72-L76
And if so, I'll remove `input` in `export_args`. Is it fine?

src/otx/algo/visual_prompting/openvino_models.py
sungchul1: Done at https://github.com/openvinotoolkit/training_extensions/pull/2955/commits/731447baaf2537e563504797209171b2dfa89694.

src/otx/algo/visual_prompting/openvino_models.py
sungchul1: Actually, `inputs["labels"]` had labels in that order, `bboxes` and `points`, and in order to access it, `idx` was assigned outside of for-loop.
It seems unclear, so I updated the whole pipeline to distinguish labels of each prompt with dict and removed `idx` at https://github.com/openvinotoolkit/training_extensions/pull/2955/commits/03d9795ae6de3a07c1ef2e247ae846191199d516 and https://github.com/openvinotoolkit/training_extensions/pull/2955/commits/b8d1b51b7135acfc4a2555abb188113579d2a8a9.

src/otx/algo/visual_prompting/openvino_models.py
sungchul1: I updated ov models' names as you recommended, `Decoder` -> `VisualPromptingDecoder` and `ImageEncoder` -> `VisualPromptingImageEncoder` at https://github.com/openvinotoolkit/training_extensions/pull/2955/commits/c7f5f1e611d4503b368ab506c9dd0ea86ffe4518.

src/otx/algo/visual_prompting/zero_shot_segment_anything.py
sungchul1: I removed unnecessary operations for the variables such as `ori_shape`, `threshold`, and `num_bg_points` and added the commend for `image_embeddings`' shape at https://github.com/openvinotoolkit/training_extensions/pull/2955/commits/f1aa67c97de4ce745533c2b3745d2e6770974c81.

src/otx/algo/visual_prompting/zero_shot_segment_anything.py
sungchul1: Updated `num_bg_points` to `int` at https://github.com/openvinotoolkit/training_extensions/pull/2955/commits/f1aa67c97de4ce745533c2b3745d2e6770974c81.

src/otx/algo/visual_prompting/zero_shot_segment_anything.py
sungchul1: Done at https://github.com/openvinotoolkit/training_extensions/pull/2955/commits/180990a4f88425d03f7e23f8d4a8c0c308c1eaca.

src/otx/core/exporter/base.py
sungchul1: > I think we need to extend export_args description emphasizing that we can modify sample input. Also, if it is only needed for example_input, then I propose to create `example_input` parameter instead of `export_args`. `export_args` has unclear naming and unclear purpose. "input" always can be obtained from self.input_size, no need to add it in arguments.

I updated the name from `export_args` to `example_inputs` at https://github.com/openvinotoolkit/training_extensions/pull/2955/commits/27a05cd95b0acc2e3f7db37c5a1bbaf1adc524a5.

src/otx/core/model/module/anomaly/anomaly_lightning.py
sungchul1: I updated it at https://github.com/openvinotoolkit/training_extensions/pull/2955/commits/979a01b5cfe421184ff9a21e4ab8e9ecfb84c185.

src/otx/core/exporter/native.py
vinnamkim: Can you define a new property, `VisualPromptModel._export_parameters` add a new `Exporter` for `VisualPromptModel._exporter` to solve your case (including additional input tensors not image)?

src/otx/core/model/entity/base.py
vinnamkim: I can see there is a pre-defined naming rule here
https://github.com/openvinotoolkit/training_extensions/blob/2708d7ba16465e35cd7665da3043223ecc2dde3f/src/otx/core/model/entity/visual_prompting.py#L192-L195
Is there any problem with just accepting `model_name` here, then change this
https://github.com/openvinotoolkit/training_extensions/blob/2708d7ba16465e35cd7665da3043223ecc2dde3f/src/otx/core/model/entity/visual_prompting.py#L210
to
```python
        model_names = {
            module: model_name.replace(basename, f"visual_prompting_{module}.xml")
            for module in ["image_encoder", "decoder"]
        }
```

src/otx/core/model/entity/visual_prompting.py
vinnamkim: Would you find a way to make common interfaces for them in the next work? To handle both torch and numpy

src/otx/core/exporter/native.py
```diff
@@ -99,14 +111,19 @@ def to_onnx(
...
 
         Returns:
             Path: The path to the saved ONNX model.
         """
-        dummy_tensor = torch.rand(self.input_size).to(next(model.parameters()).device)
+        if example_inputs is None:
+            example_inputs = {"args": torch.rand(self.input_size).to(next(model.parameters()).device)}
+
         save_path = str(output_dir / (base_model_name + ".onnx"))
+        example_inputs.update({"model": model, "f": save_path})
```
eunwoosh: `model` and `f` aren't example_inputs. Could you change a variable name?

src/otx/core/model/entity/base.py
```diff
@@ -345,7 +356,13 @@ def __init__(
 
         tile_enabled = False
         with contextlib.suppress(RuntimeError):
-            tile_enabled = "tile_size" in self.model.inference_adapter.get_rt_info(["model_info"]).astype(dict)
+            if isinstance(self.model, Model):
+                tile_enabled = "tile_size" in self.model.inference_adapter.get_rt_info(["model_info"]).astype(dict)
+            else:
```
eunwoosh: It's just a question. Is Visual Prompting related to tiling?

src/otx/core/model/entity/base.py
```diff
@@ -247,6 +247,7 @@ def export(
         base_name: str,
         export_format: OTXExportFormatType,
         precision: OTXPrecisionType = OTXPrecisionType.FP32,
+        export_args: dict[str, Any] | None = None,
```
eunwoosh: I think this also needs to be changed to `example_inputs`.

src/otx/core/model/module/anomaly/anomaly_lightning.py
```diff
@@ -109,19 +112,19 @@ def to_onnx(
         base_model_name: str = "exported_model",
         precision: OTXPrecisionType = OTXPrecisionType.FP32,
         embed_metadata: bool = True,
+        example_inputs: dict[str, Any] | None = None,
+        **kwargs,
```
eunwoosh:
```suggestion
```
Is it necessary?

src/otx/core/model/module/anomaly/anomaly_lightning.py
```diff
@@ -109,19 +112,19 @@ def to_onnx(
...
     ) -> Path:
+        if example_inputs is None:
+            example_inputs = {
```
eunwoosh: Same as above, except args, other things aren't 'exapmle_inputs'. Could you change a name?

src/otx/core/model/module/visual_prompting.py
```diff
@@ -243,22 +246,80 @@ def _inference_step(
 class OTXZeroShotVisualPromptingLitModule(OTXVisualPromptingLitModule):
     """Base class for the lightning module used in OTX zero-shot visual prompting task."""
 
-    def set_metrics(self) -> None:
-        """Set metrics."""
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
eunwoosh:
```suggestion
```
I think this part can be removed.

src/otx/core/model/module/visual_prompting.py
```diff
@@ -243,22 +246,80 @@ def _inference_step(
...
+
+    def configure_metric(self, cond: str = "") -> None:
```
eunwoosh: It seems that `cond` isn't used in this class and `OTXVisualPromptingLitModule` and `OTXLitModule` also. How about removing it?

src/otx/core/model/entity/visual_prompting.py
```diff
@@ -30,46 +49,918 @@ class OTXVisualPromptingModel(
...
+            },
+        }
+
+        export_paths: dict[str, Path] = {}
+        for module in ["image_encoder", "decoder"]:
+            self._export_parameters = module  # type: ignore[assignment]
+            export_paths[module] = self._exporter.export(
+                model=model[module],
+                output_dir=output_dir,
+                base_model_name=f"visual_prompting_{module}",
```
eunwoosh:
```suggestion
                base_model_name=f"{base_name}_{module}",
```
It's a minor one but how about changing like this?

src/otx/core/model/entity/visual_prompting.py
sungchul1: I'll find a way, thanks for the advice.

src/otx/core/exporter/native.py
sungchul1: As far as I understood, `_export_parameters` is for building `Exporter`, not `_exporter.export`.
https://github.com/openvinotoolkit/training_extensions/blob/2708d7ba16465e35cd7665da3043223ecc2dde3f/src/otx/core/model/entity/visual_prompting.py#L120-L156
So, I added `export_args` (current `example_inputs`) for export execution.
Could you suggest a way what you said?

src/otx/core/model/entity/base.py
sungchul1: Updated at https://github.com/openvinotoolkit/training_extensions/pull/2955/commits/05bfdfae9ef6ad8873ee1971e27f6efd26aac0a0.

src/otx/core/exporter/native.py
vinnamkim:
```python
class VisualPromptExporter(OTXExporter):
     def __init__(self, dummy_decoder_input, ...):
         ...

...

class VisualPromptModel:
    @property
    def _export_parameters(self):
         return {"dummy_decoder_input": ..., ...}
    
    @property
    def _exporter(self):
         return VisualPromptExporter(**self._export_parameters)
    
```

eunwoosh (commented):
Thanks for your work :) I left some comments. Please take a look

src/otx/core/model/entity/visual_prompting.py
sungchul1: Do you think `exported_model_{module}` seems better than `visual_prompting_{module}`?

src/otx/core/exporter/native.py
sungchul1: Done at https://github.com/openvinotoolkit/training_extensions/pull/2955/commits/73f081a634906b81345614e45df42e62dd2c10fa.

src/otx/core/model/entity/base.py
sungchul1: I just set if-else because `self.model` for visual prompting is `dict`, not `Model`.
Actually, tiling for visual prompting is not supported yet, but it can be expanded.
I removed it at https://github.com/openvinotoolkit/training_extensions/pull/2955/commits/d9335dec60ff9a5e096b1317a61042f10df00208.

src/otx/core/model/entity/base.py
sungchul1: Done at https://github.com/openvinotoolkit/training_extensions/pull/2955/commits/97d674bc75b907b27e726b5cf5f71c4d628d5bff.

src/otx/core/model/module/anomaly/anomaly_lightning.py
sungchul1: Removed at https://github.com/openvinotoolkit/training_extensions/pull/2955/commits/0a9eaa13fa0cb01c2eec8ed1358d0592c6128f05.

src/otx/core/model/module/anomaly/anomaly_lightning.py
sungchul1: Done at https://github.com/openvinotoolkit/training_extensions/pull/2955/commits/9393dd424fa1f536f7a5269d5b77dd59a39d652a.

src/otx/core/model/module/visual_prompting.py
sungchul1: Done at https://github.com/openvinotoolkit/training_extensions/pull/2955/commits/c86ddcaebcd3a4729d2e8fd9db6dc6f5dd0c3c28.

src/otx/core/model/module/visual_prompting.py
sungchul1: I'm not sure why `cond` is there, removed at https://github.com/openvinotoolkit/training_extensions/pull/2955/commits/c86ddcaebcd3a4729d2e8fd9db6dc6f5dd0c3c28.

src/otx/core/exporter/native.py
sungchul1: Is it okay to add new exporter for this task?
It seems that exporters are distinguished with frameworks, mmdeploy and torch.

vinnamkim: I cannot see any reason not to add.

src/otx/core/model/entity/visual_prompting.py
eunwoosh: I like both but my intention is for keeping unity with other tasks and not ignoring `base_name` argument.

src/otx/core/exporter/native.py
sungchul1: @harimkang @jaegukhyun @sungmanc  What do you think about adding a new exporter only for visual prompting?
As @vinnamkim said, it definitely makes setting parameters clearer, but I need more opinions because exporters seems to be distinguished with frameworks, not tasks.

vinnamkim: In the very first place, there is no such restriction that the exporters should be distinguished by framework. The reason why we make the exporters is to lower code complexity of `OTXModel`. However, in your case, serving multiple dummy input tensors during exportation is not covered by the existing exporters. To resolve this, it seems better to create a new exporter dedicated to cover your use-case, rather than overhauling a bunch of code in multiple places to extend the functionality of the existing one.

sungchul1: I agree with your opinion, but I'm just curious about others' thoughts.
For now, I'll add a new exporter for this case, and let's discuss how to handle different types of exporters and decide on policies later.

src/otx/core/model/entity/visual_prompting.py
sungchul1: I agree with that, updated at https://github.com/openvinotoolkit/training_extensions/pull/2955/commits/84ed9f4f1769453e21f59015110eee3002168d64.

src/otx/core/exporter/native.py
eunwoosh: To add my opinion, I think this is about who should be in charge of preparing export argument. When we implemented exporter classes, it's true to distinguish export class based on framework (torch + OV and mmdeploy + OV). Of course, there is no restriction that exporter should be distinguished based on framework, but it seems a little weird that det and i-seg model uses mmdeploy exporter to use mmdeploy but VP uses separated exporter even if it uses torch + OV like other tasks. Anyway, back to topic about responsibility, I think that preparing export argument by VP task model isn't wrong because it's nature to prepare necessary things for own purpose. I think making a new class for VP export also can be other way to deal with this but, as I said, it's a little awkward if we look all exports class implementation including VP exporter in my perspective.

vinnamkim: The problem is that we already prepared this mechanism to re-define these two properties in the derived class:
https://github.com/openvinotoolkit/training_extensions/blob/98fb7126085ae5766ad1bd827c555ec9cc698c29/src/otx/core/model/entity/base.py#L266-L300
It must have all the degree of freedom to adapt every model exportation use-case with this mechanism. However, this PR is ignoring this mechanism from the front and is trying to dig out a new interface by expanding `OTXModel.export()` and `OTXExporter.export()`.

sungmanc: > @harimkang @jaegukhyun @sungmanc What do you think about adding a new exporter only for visual prompting? As @vinnamkim said, it definitely makes setting parameters clearer, but I need more opinions because exporters seems to be distinguished with frameworks, not tasks.

I prefer to use the additional class for the VisualPrompting task since this task includes special cases.

eunwoosh: > It must have all the degree of freedom to adapt every model exportation use-case with this mechanism. However, this PR is ignoring this mechanism from the front and is trying to dig out a new interface by expanding OTXModel.export() and OTXExporter.export().

Yeap. In my thought, to make design more clean, we need to change overall code but it should be considered in other PR. I don't have any objection if other team member prefer new class design.

sungchul1: I reverted other exporters as removing `example_inputs` at https://github.com/openvinotoolkit/training_extensions/commit/2d756840152dd9f5dec19619fb5a8ed5e9964578 and added a new `OTXVisualPromptingModelExporter` at https://github.com/openvinotoolkit/training_extensions/commit/a80aae2dd69e06aefa477e1a7a09ddca789a70b6.

eunwoosh (approved):

vinnamkim (approved):

jaegukhyun (approved):

PR #2957, opened 2024-02-21
Enable forward_explain() for deit

negvet:
### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algo/classification/deit_tiny.py
+++ b/src/otx/algo/classification/deit_tiny.py
@@ -4,10 +4,11 @@
 """DeitTiny model implementation."""
 from __future__ import annotations
 
-from typing import Any
+from typing import TYPE_CHECKING, Any, Callable
 
 import numpy as np
 import torch
+from mmpretrain.models.utils import resize_pos_embed
 
 from otx.algo.utils.mmconfig import read_mmconfig
 from otx.algo.utils.support_otx_v1 import OTXv1Helper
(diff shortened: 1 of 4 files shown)


GalyaZalesskaya (dismissed):

sungmanc (dismissed):

src/otx/algo/classification/deit_tiny.py
```diff
@@ -68,6 +56,80 @@ def head_forward_fn(self, x: torch.Tensor) -> torch.Tensor:
...
+            raise RuntimeError(msg)
+
+        return {
+            "logits": logits,
+            "saliency_map": saliency_map,
+        }
+
+    def get_explain_fn(self) -> Callable:
+        """Returns explain function."""
+        from otx.algo.hooks.recording_forward_hook import ViTReciproCAMHook
```
vinnamkim: I think it will be safe even if we move this import to the top of this file.

src/otx/core/model/entity/classification.py
```diff
@@ -669,6 +669,7 @@ def _customize_outputs(
 
         if outputs and outputs[0].saliency_map.size != 0:
             predicted_s_maps = [out.saliency_map for out in outputs]
+            predicted_s_maps = [s_map[0] if s_map.size != 0 else s_map for s_map in predicted_s_maps]
```
vinnamkim: Is this a rule only for DEIT? If so, this should be handled in its own Python class.

sovrasov: If possible, we have to align the layout of the output saliency maps for all the classification models in torch. Other clients of MAPI may not be aware of Deit's specific.

negvet: This is a common fix, for all classification models. Each `out.saliency_map` is of `(1, num_classes, H, W)` size. OV always infers with batch size of 1 (although xai algos support batch > 1). It is better that `predicted_s_maps` is the list of explain results per image - this is also aligned with torch.
@GalyaZalesskaya Please take a look at the detection case

vinnamkim: Would you please change it so others can better understand the situation such as?
```
        def _is_not_empty(arr: np.ndarray) -> bool:
            return arr.size != 0

        if outputs:
            predicted_s_maps = [
                # Squeeze dim 4D => 3D, (1, num_classes, H, W) => (num_classes, H, W)
                s_map[0] for out in outputs 
                if _is_not_empty(s_map := out.saliency_map)
            ]
```

sovrasov: If so, should we consider `s_map.size` == 0 here? The outer check ensures the tensor is not empty. If we agree on that saliency_map is [B,C,H,W] tensor, we can omit this extra if here

vinnamkim: I don't know exactly but I guess that there is a case that `outputs` is not an empty list and `output.saliency_map == np.array([])`.

negvet: Indeed, I missed that there is already an outer check.
What about just:
```
if outputs and outputs[0].saliency_map.size != 1:
    # Squeeze dim 4D => 3D, (1, num_classes, H, W) => (num_classes, H, W)
    predicted_s_maps = [out.saliency_map[0] for out in outputs]
```

It handles the case when that outputs is not an empty list and output.saliency_map == np.array([]).

The weak point here is that we couple saliency maps with feature vector, but probably it is ok for now (the same as in OTX1.X). We can handle it later if required.

vinnamkim (dismissed):

jaegukhyun (dismissed):

sovrasov (dismissed):

harimkang:
https://github.com/openvinotoolkit/training_extensions/pull/2970 - Fixed integration test FAIL in a recent PR. That's probably where the current FAILs are coming from (update branch -> than resolved,, maybe..?), please note.

sovrasov (approved):

kprokofi (approved):

GalyaZalesskaya (approved):

PR #2973, opened 2024-02-23
Align Engine.from_config with CLI behavior

harimkang:
### Summary

- Refactoring `Engine.from_config()` & `CLI`
- Align Engine.from_config and CLI to have the same behavior as much as possible

The two examples below do the same thing.
API: from_config
```python
from otx.engine import Engine

recipe_path = "src/otx/recipe/classification/multi_class_cls/otx_efficientnet_b0.yaml"
data_root = "/home/harimkan/workspace/repo/datasets/otx_v2_dataset/multiclass_classification/multiclass_CUB_medium"


engine = Engine.from_config(
    config_path=recipe_path, data_root=data_root, work_dir="otx-workspace-api", seed=1234
)

# Metric is now engine.train, engine.test argument, So we need to declare this separately.
from functools import partial
from torchmetrics.classification.accuracy import Accuracy

metric = partial(Accuracy, task="multiclass")
engine.train(
    deterministic=True,
    max_epochs=90,
    precision=16,
    metric=metric,
)

engine.test(metric=metric)
```
CLI command
```shell
otx train \
--config src/otx/recipe/classification/multi_class_cls/otx_efficientnet_b0.yaml \
--data_root /home/harimkan/workspace/repo/datasets/otx_v2_dataset/multiclass_classification/multiclass_CUB_medium \
--seed 1234 \
--deterministic True \
--work_dir otx-workspace-cli
```
Same Test results
image

### Checklist

### License

```python
```


src/otx/cli/utils/jsonargparse.py
```diff
@@ -351,14 +353,70 @@ def get_configuration(config_path: str | Path) -> dict:
...
+        "disable_infer_num_classes",
+        "workspace",
+    ]
     logger.warning(f"The corresponding keys in config are not used.: {cli_args}")
     for arg in cli_args:
         config.pop(arg, None)
     return config
+
+
+def get_instantiated_classes(
```
harimkang, line 382: @samet-akcay @ashwinvaidya17 I think this is a good example that can be applied to `AnomalibCLI` as well. We support having the API and CLI behave the same as much as possible.

harimkang: image

samet-akcay: @harimkang, I agree

samet-akcay: API and CLI should ideally have the same behaviour

src/otx/cli/cli.py
```diff
@@ -326,41 +329,53 @@ def _set_extension_subcommands_parser(self, parser_subcommands: _ActionSubComman
...
         If it is, it instantiates the necessary classes such as config, datamodule, model, and engine.
+
+        Args:
+            instantiate_engine (bool, optional): Whether to instantiate the engine. Defaults to True.
         """
         if self.subcommand in self.engine_subcommands():
             # For num_classes update, Model and Metric are instantiated separately.
             model_config = self.config[self.subcommand].pop("model")
-            metric_config = self.config[self.subcommand].pop("metric")
+            metric_config = self.config[self.subcommand].get("metric")
```
jaegukhyun, line 343: Will it affect to https://github.com/openvinotoolkit/training_extensions/blob/fa70d8c357bf8f7d3b3c7a41532e4053d53430e4/src/otx/cli/cli.py#L374?

jaegukhyun (dismissed):

src/otx/cli/cli.py
harimkang: In the output configuration file, the metric section was currently being dropped; fix this.

src/otx/cli/utils/jsonargparse.py
```diff
@@ -351,14 +353,70 @@ def get_configuration(config_path: str | Path) -> dict:
...
+    ]
+    if work_dir is not None:
+        cli_args.extend(["--work_dir", str(work_dir)])
+    if data_root is not None:
+        cli_args.extend(["--data_root", str(data_root)])
+    for key, value in kwargs.items():
+        cli_args.extend([f"--{key}", str(value)])
+    otx_cli = OTXCLI(
+        args=cli_args,
+        run=False,
```
sungmanc, line 415: I'm not sure this is the best way to get some information needed to get the instantiated classes. Calling the `CLI` class at the function level even though it's not a `main` function is weird for me. Is it inevitable?

src/otx/cli/cli.py
```diff
@@ -326,41 +329,53 @@ def _set_extension_subcommands_parser(self, parser_subcommands: _ActionSubComman
...
+            model=self.model,
+            optimizer=self.optimizer,
+            scheduler=self.scheduler,
+            datamodule=self.datamodule,
+            work_dir=self.workspace.work_dir,
+            **engine_kwargs,
+        )
 
-    def instantiate_metric(self, metric_config: Namespace) -> Metric | None:
+    def instantiate_metric(self, metric_config: Namespace) -> Metric | partial | None:
```
sungmanc:
```suggestion
    def instantiate_metric(self, metric_config: Namespace) -> partial | None:
```
It seems that there is no case that returns the `Metric` class.

src/otx/cli/utils/jsonargparse.py
sungmanc: Same for the #L356

src/otx/core/utils/cache.py
```diff
@@ -47,6 +48,8 @@ def __init__(self, **kwargs) -> None:
     def update(self, **kwargs) -> None:
         """Replace cached arguments with arguments retrieved from the model."""
         for key, value in kwargs.items():
+            if value is None:
+                continue
```
sungmanc, line 52:
```suggestion
for key, value in kwargs.items():
        if value is not None and self._cached_args.get(key) != value:
            logger.info(f"Overriding {key} from {self._cached_args.get(key)} with {value}")
            self._cached_args[key] = value
```

sungmanc (commented):
Thanks for the PR, please check my comments

src/otx/cli/utils/jsonargparse.py
harimkang: The purpose of this function is to utilize the `OTXCLI` to collect arguments, and work up to instantiate to get each module. To make this work the same as the CLI without using the CLI class, we would have to implement code directly to do what the CLI does, which is redundant, and it would be a difficult task to make it work the same.

src/otx/core/utils/cache.py
harimkang: This will work fine for updating existing values, but will be ignored for newly added values.

src/otx/cli/cli.py
harimkang: Thanks! Done - https://github.com/openvinotoolkit/training_extensions/pull/2973/commits/089e3dda40f0ef1976924d0b29e3277a784ede93

src/otx/cli/cli.py
```diff
@@ -326,41 +328,53 @@ def _set_extension_subcommands_parser(self, parser_subcommands: _ActionSubComman
...
+            model=self.model,
+            optimizer=self.optimizer,
+            scheduler=self.scheduler,
+            datamodule=self.datamodule,
+            work_dir=self.workspace.work_dir,
+            **engine_kwargs,
+        )
 
-    def instantiate_metric(self, metric_config: Namespace) -> Metric | None:
+    def instantiate_metric(self, metric_config: Namespace) -> partial | None:
```
samet-akcay: Can you explain why this is `partial` now?

vinnamkim:
```suggestion
    def instantiate_metric(self, metric_config: Namespace) -> MetricCallable | None:
```

harimkang: Thanks for suggestion, Done - https://github.com/openvinotoolkit/training_extensions/pull/2973/commits/5e8b627b20182fd31f6b6929db5d32c4b7ad7fac

vinnamkim (commented):
I don't know why you need this because it will show a different behavior from my expectation if an user executes the code not exactly as you provided such as
```python
engine = Engine.from_config(
    config_path=recipe_path, data_root=data_root, work_dir="otx-workspace-api", seed=1234
)

***********************************************************
Execute some codes generating random variables,
so that change random number generator's inner states
***********************************************************

engine.train(
    deterministic=True,
    max_epochs=90,
    precision=16,
    metric=metric,
)

engine.test(metric=metric)
```

If you really need this, I think that it only happens if `deterministic=True`, I would recommend you to implement

```
class Engine:
    def train(self, deterministic=True):
          if deterministic:
              reset_random_number_generator()
              reinit_model_weights()
              reinit_data_sampler()
          train()
```

src/otx/cli/cli.py
harimkang: As I understand it, some metrics (mainly H-label classification Accuracy) need to update the `num_classes` value, which is instantiated in the CLI in partial form and passed to `LitModule` for use.

and retype partial to `MetricCallable`.

harimkang:
> I don't know why you need this because it will show a different behavior from my expectation if an user executes the code not exactly as you provided such as
> 
> 
> If you really need this, I think that it only happens if `deterministic=True`, I would recommend you to implement
> 
> ```
> class Engine:
>     def train(self, deterministic=True):
>           if deterministic:
>               reset_random_number_generator()
>               reinit_model_weights()
>               reinit_data_sampler()
>           train()
> ```

Yes you are right, we need to take care of this on the seed, deterministic side to get the same behavior for now (which probably doesn't look natural to the user).
I'll probably think about your suggestion in a future PR and take a look at it, thanks for the suggestion!

src/otx/core/utils/cache.py
sungmanc: It seems that it's same with the previous implementation for me.  However, since it's not important, I don't mind selecting what you want

sungmanc (approved):

jaegukhyun (approved):

vinnamkim (approved):

PR #2977, opened 2024-02-26
Enable sal map postprocessing

negvet:
### Summary

TODO: add tests

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algo/utils/xai_utils.py
+++ b/src/otx/algo/utils/xai_utils.py
@@ -25,19 +25,20 @@ def process_saliency_maps_in_pred_entity(
     work_dir: Path | None = None,
 ) -> list[Any] | list[OTXBatchPredEntityWithXAI | InstanceSegBatchPredEntityWithXAI]:
     """Process saliency maps in PredEntity."""
-    saliency_maps = predict_result[0].saliency_maps
-    pred_labels = predict_result[0].labels  # type: ignore[union-attr]
-    if pred_labels:
-        pred_labels = [pred.tolist() for pred in pred_labels]
+    for batch_id in range(len(predict_result)):
+        saliency_maps = predict_result[batch_id].saliency_maps
+        pred_labels = predict_result[batch_id].labels  # type: ignore[union-attr]
+        if pred_labels:
+            pred_labels = [pred.tolist() for pred in pred_labels]
 
-    processed_saliency_maps = process_saliency_maps(saliency_maps, explain_config, pred_labels)
+        processed_saliency_maps = process_saliency_maps(saliency_maps, explain_config, pred_labels)
 
-    if processed_saliency_maps and work_dir:
-        # Temporary saving random saliency map for image 0 (for tests)
-        s_map_to_save = next(iter(processed_saliency_maps[0].values()))
-        cv2.imwrite(str(work_dir / "saliency_map.tiff"), s_map_to_save)
+        if processed_saliency_maps and work_dir:
+            # Temporary saving random saliency map for image 0 (for tests)
+            s_map_to_save = next(iter(processed_saliency_maps[0].values()))
+            cv2.imwrite(str(work_dir / "saliency_map.tiff"), s_map_to_save)
 
-    predict_result[0].saliency_maps = processed_saliency_maps
+        predict_result[batch_id].saliency_maps = processed_saliency_maps
     return predict_result
 
 
@@ -46,13 +47,13 @@ def process_saliency_maps(
     explain_config: ExplainConfig,
     pred_labels: list | None,
 ) -> list[dict[Any, Any]]:
-    """Perform saliency map selection and post-processing."""
+    """Perform saliency map convertion to dict and post-processing."""
     if explain_config.target_explain_group == TargetExplainGroup.ALL:
-        processed_saliency_maps = select_saliency_maps_all(saliency_maps)
+        processed_saliency_maps = convert_maps_to_dict_all(saliency_maps)
     elif explain_config.target_explain_group == TargetExplainGroup.PREDICTIONS:
-        processed_saliency_maps = select_saliency_maps_predictions(saliency_maps, pred_labels)
+        processed_saliency_maps = convert_maps_to_dict_predictions(saliency_maps, pred_labels)
     elif explain_config.target_explain_group == TargetExplainGroup.IMAGE:
-        processed_saliency_maps = select_saliency_maps_image(saliency_maps)
+        processed_saliency_maps = convert_maps_to_dict_predictions_image(saliency_maps)
     else:
         msg = f"Target explain group {explain_config.target_explain_group} is not supported."
         raise ValueError(msg)
@@ -64,8 +65,8 @@ def process_saliency_maps(
     return processed_saliency_maps
 
 
-def select_saliency_maps_all(saliency_maps: np.array) -> list[dict[Any, np.array]]:
-    """Select salincy maps for TargetExplainGroup.ALL."""
+def convert_maps_to_dict_all(saliency_maps: np.array) -> list[dict[Any, np.array]]:
+    """Convert salincy maps to dict for TargetExplainGroup.ALL."""
     if saliency_maps[0].ndim != 3:
         raise ValueError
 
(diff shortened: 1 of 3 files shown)


src/otx/engine/engine.py
```diff
@@ -393,6 +394,14 @@ def predict(
             return_predictions=return_predictions,
         )
 
+        if explain:
+            from otx.algo.utils.xai_utils import process_saliency_maps_in_pred_entity
```
GalyaZalesskaya: Did you move this import so we wouldn't have to import cv2 from `otx.algo.utils.xai_utils` each time? I couldn't find if cv2 is an optional dependency or not

src/otx/algo/utils/xai_utils.py
```diff
@@ -11,44 +11,96 @@
...
-) -> list:
-    """Postptocess saliency maps."""
-    # Implement <- TODO(negvet)
-    return saliency_maps
+    explain_config: ExplainConfig,
+    pred_labels: list | None,
+) -> list[dict[Any, Any]]:
+    """Perform saliency map selection and post-processing."""
+    if explain_config.target_explain_group == TargetExplainGroup.ALL:
+        processed_saliency_maps = select_saliency_maps_all(saliency_maps)
```
GalyaZalesskaya: I suggest combining functions `select_saliency_maps_all` and ` select_saliency_maps_predictions` into ` select_saliency_maps` where `pred_labels` will be passed in case of TargetExplainGroup.PREDICTIONS.

Since both these functions have the same checks and logic It makes sense

src/otx/algo/utils/xai_utils.py
```diff
@@ -11,44 +11,96 @@
...
+    elif explain_config.target_explain_group == TargetExplainGroup.PREDICTIONS:
+        processed_saliency_maps = select_saliency_maps_predictions(saliency_maps, pred_labels)
+    elif explain_config.target_explain_group == TargetExplainGroup.IMAGE:
+        processed_saliency_maps = select_saliency_maps_image(saliency_maps)
```
GalyaZalesskaya: It's not really a `select`  but rather a `convert_saliency_maps_image`. But this just a comment, can stay as it is

src/otx/algo/utils/xai_utils.py
```diff
@@ -11,44 +11,96 @@
...
+    else:
+        msg = f"Target explain group {explain_config.target_explain_group} is not supported."
```
GalyaZalesskaya, line 58: We wont support `TargetExplainGroup.CUSTOM`, right?

src/otx/algo/utils/xai_utils.py
```diff
@@ -11,44 +11,96 @@
...
-    if raw_saliency_maps and work_dir:
-        # Temporary saving saliency map for image 0, class 0 (for tests)
-        cv2.imwrite(str(work_dir / "saliency_map.tiff"), raw_saliency_maps[0][0])
+def process_saliency_maps_in_pred_entity(
+    predict_result: list[Any] | list[OTXBatchPredEntityWithXAI | InstanceSegBatchPredEntityWithXAI],
+    explain_config: ExplainConfig,
+    work_dir: Path | None = None,
+) -> list[Any] | list[OTXBatchPredEntityWithXAI | InstanceSegBatchPredEntityWithXAI]:
+    """Process saliency maps in PredEntity."""
+    saliency_maps = predict_result[0].saliency_maps
```
GalyaZalesskaya: This TODO is still relevant right?

>     # Optimize for memory <- TODO(negvet)

sovrasov: `predict_result` is supposed to be a list of per-batch saliency maps. Does this line mean we process only the first batch?

src/otx/engine/engine.py
negvet: I have not moved this import, this is a new one. May be you mean the one that is below?

src/otx/algo/utils/xai_utils.py
negvet: It is definitely possible to combine all of them together with a couple of if-else branches. But it seems that to keep them separated make sense in terms of decoupling and simplicity. What do you think?

src/otx/algo/utils/xai_utils.py
negvet: I agree than naming is not perfect, I updated

src/otx/algo/utils/xai_utils.py
negvet: For now - no. Probably in the future.

src/otx/algo/utils/xai_utils.py
negvet: Well, probably no, we are no longer aggregate explain results anywhere from inference to inference. Let me know if I miss smth. I will also add dumping maps in the future PR.

negvet: Right, it was not supported. I fixed it.

src/otx/algo/utils/xai_utils.py
```diff
@@ -11,44 +11,100 @@
...
-    if raw_saliency_maps and work_dir:
-        # Temporary saving saliency map for image 0, class 0 (for tests)
-        cv2.imwrite(str(work_dir / "saliency_map.tiff"), raw_saliency_maps[0][0])
+def process_saliency_maps_in_pred_entity(
+    predict_result: list[Any] | list[OTXBatchPredEntityWithXAI | InstanceSegBatchPredEntityWithXAI],
+    explain_config: ExplainConfig,
+    work_dir: Path | None = None,
+) -> list[Any] | list[OTXBatchPredEntityWithXAI | InstanceSegBatchPredEntityWithXAI]:
+    """Process saliency maps in PredEntity."""
+    for batch_id in range(len(predict_result)):
```
sovrasov: `enumerate(predict_result)`?
With enumerate we can also clearly separate mutable/immutable access

src/otx/algo/utils/xai_utils.py
GalyaZalesskaya: I don't mind to go any of these ways.

src/otx/engine/engine.py
GalyaZalesskaya: I mean can we move this import on the top?

tests/unit/algo/hooks/test_saliency_map_processing.py
```diff
@@ -0,0 +1,149 @@
...
+SALIENCY_MAPS = [np.ones((NUM_CLASSES, RAW_SIZE, RAW_SIZE), dtype=np.uint8) for _ in range(BATCH_SIZE)]
+SALIENCY_MAPS_IMAGE = [np.ones((RAW_SIZE, RAW_SIZE), dtype=np.uint8) for _ in range(BATCH_SIZE)]
+
+
+@pytest.mark.parametrize("postprocess", [False, True])
+def test_process_all(postprocess) -> None:
+    explain_config = ExplainConfig(target_explain_group=TargetExplainGroup.ALL, postprocess=postprocess)
+
+    with pytest.raises(ValueError, match="Shape mismatch."):
+        processed_saliency_maps = process_saliency_maps(SALIENCY_MAPS_IMAGE, explain_config, PRED_LABELS)
```
vinnamkim, line 27:
```suggestion
@pytest.fixture
def fxt_saliency_maps():
    return [np.ones((NUM_CLASSES, RAW_SIZE, RAW_SIZE), dtype=np.uint8) for _ in range(BATCH_SIZE)]

@pytest.fixture
def fxt_saliency_maps_image():
    return [np.ones((RAW_SIZE, RAW_SIZE), dtype=np.uint8) for _ in range(BATCH_SIZE)]

@pytest.mark.parametrize("postprocess", [False, True])
def test_process_all(postprocess, fxt_saliency_maps) -> None:
    explain_config = ExplainConfig(target_explain_group=TargetExplainGroup.ALL, postprocess=postprocess)

    with pytest.raises(ValueError, match="Shape mismatch."):
        processed_saliency_maps = process_saliency_maps(fxt_saliency_maps, explain_config, PRED_LABELS)
```

Please use pytest fixture to follow the common practice and avoid to use global variables which are hard to debug if they are silently changed somewhere.

src/otx/algo/utils/xai_utils.py
negvet: But we do not need both index and a value. PredEntity is a mutable object so we might pick just a value. Take a look.

src/otx/engine/engine.py
negvet: Not sure I see which top you mean, I moved it to the top of the method. This is what you proposed?

sovrasov (approved):

src/otx/engine/engine.py
GalyaZalesskaya: I'm sorry if I wasn't clear, I meant why can't we move this import on the top of the file with other imports?

GalyaZalesskaya (approved):

harimkang (approved):

vinnamkim (approved):

src/otx/engine/engine.py
negvet: Probably we can, but what I see from the file convention, all local utility imports are imported locally in corresponding methods. I just followed this convention, probably wrongly.

PR #2993, opened 2024-02-27
Implement sampler API & add BalancedSampler

harimkang:
### Summary

- Configuration to make the Sampler mutable.
- Add `BalancedSampler`, `RepeatSampler`

BalancedSampler in OTX 1.6 : https://github.com/openvinotoolkit/training_extensions/blob/develop/src/otx/algorithms/common/adapters/torch/dataloaders/samplers/balanced_sampler.py

```
otx train \
--config src/otx/recipe/classification/multi_class_cls/otx_efficientnet_b0.yaml \
--data_root /home/harimkan/workspace/repo/datasets/otx_v2_dataset/multiclass_classification/multiclass_CUB_medium \
--seed 1234 \
--deterministic True \
--data.config.train_subset.sampler.class_path otx.algo.samplers.balanced_sampler.BalancedSampler \
--work_dir otx-cls-balanced-sampler
```

Here is the result of running a simple Regression Dataset.
*seed: 1234, deterministic: True (The example below is a random sampler generated with np.Generator, so it may look different than the current state with torch.Generator.)
image
Just for reference, the current Regression Test dataset is not a typical dataset and does not have unbalanced classes. (We need to experiment more.)
So the current default is the same RandomSampler as before.

### Checklist

### License

```python
```


src/otx/algo/samplers/balanced_sampler.py
```diff
@@ -0,0 +1,160 @@
+"""Balanced sampler for imbalanced data."""
```
sungmanc:
```suggestion
# Copyright (C) 2024 Intel Corporation
# SPDX-License-Identifier: Apache-2.0
"""Balanced sampler for imbalanced data."""
```

src/otx/algo/samplers/balanced_sampler.py
```diff
@@ -0,0 +1,160 @@
...
+
+from .repeat_sampler import RepeatSampler
+
+if TYPE_CHECKING:
+    from datumaro import Dataset as DmDataset
+
+    from otx.core.data.dataset.base import OTXDataset
+
+
+def compute_class_statistics(dm_dataset: DmDataset) -> dict[str, list[int]]:
```
sungmanc: It seems that this function just returns the dict including all labels in the dataset, then how about chaning the naming ? might `get_all_classes()`?

src/otx/algo/samplers/repeat_sampler.py
```diff
@@ -0,0 +1,137 @@
...
+                       (coef * math.sqrt(num_iters-1)) +5
+        min_repeat (float) : minimum repeats
+    """
+    if data_size == 0 or batch_size == 0:
+        return 1
+    n_iters_per_epoch = math.ceil(data_size / batch_size)
+    return math.floor(max(coef * math.sqrt(n_iters_per_epoch - 1) + 5, min_repeat))
+
+
+class RepeatSampler(Sampler):
```
vinnamkim: Why do we need this although we can set the sample size of `torch.utils.data.RandomSampler` with `num_samples` argument?

src/otx/algo/samplers/balanced_sampler.py
```diff
@@ -0,0 +1,160 @@
...
+                num_samples = math.ceil(num_samples / self.num_replicas)
+            self.total_size = num_samples * self.num_replicas
+
+        return num_samples
+
+    def __iter__(self):
+        """Iter."""
+        indices = []
+        random_generator = np.random.default_rng()
+        for _ in range(self.repeat):
```
vinnamkim: See this link https://pytorch.org/docs/2.1/_modules/torch/utils/data/sampler.html#RandomSampler and I hope you follow a flow as similar to the PyTorch codebase as possible, especially for the random generator part.

src/otx/algo/samplers/repeat_sampler.py
harimkang: @sungmanc I'm not exactly sure of the history of the use of that class, and I think Vinnam's comment is right on the money. From my quick look, it doesn't seem to be much different than `RandomSampler` with `num_samples`, other than supporting the code snippet below, but do you have any thoughts?
image

sungmanc: I just remembered the reason why we brought this class: we want to add the functionality that adjusts the repeat with float times. i.e. (1.2x num_data) or (0.7x num_data). However, since we don't have a room, this is the same with the `torch.RandomSampler`.  So, for the current status, I think using the `torch.RandomSampler` looks good.

harimkang: I removed `RepeatSampler` and modified `BalancedSampler` to inherit from `torch.utils.data.sampler.Sampler`. - https://github.com/openvinotoolkit/training_extensions/pull/2993/commits/08383dd4882e766fe1cf951b05224d8edab203d5

src/otx/algo/samplers/balanced_sampler.py
harimkang: I replaced `np.Generator` with `torch.Generator` and modified the syntax around generator accordingly. - https://github.com/openvinotoolkit/training_extensions/pull/2993/commits/f09c0192faf1045647991bfdf24c426b5eeb5585

src/otx/algo/samplers/balanced_sampler.py
harimkang: Thanks! Done - https://github.com/openvinotoolkit/training_extensions/pull/2993/commits/e36d25f61754cb929149595ff6c1642d92bf6618

src/otx/algo/samplers/balanced_sampler.py
harimkang: I renamed this to `get_idx_list_per_classes` - https://github.com/openvinotoolkit/training_extensions/pull/2993/commits/08383dd4882e766fe1cf951b05224d8edab203d5

src/otx/algo/samplers/balanced_sampler.py
```diff
@@ -0,0 +1,174 @@
...
+            return ann.label
+
+    stats: dict[str, list[int]] = {}
+    for i, item in enumerate(dm_dataset):
+        for ann in item.annotations:
+            label = get_label(ann)
+            if label in stats:
+                stats[label].append(i)
+            else:
+                stats[label] = [i]
```
vinnamkim:
```suggestion
    stats: dict[str, list[int]] = defaultdict(list)
    for item_idx, item in enumerate(dm_dataset):
        for ann in item.annotations:
            label = get_label(ann)
            stats[label].append(item_idx)
```

src/otx/algo/samplers/balanced_sampler.py
```diff
@@ -0,0 +1,174 @@
...
+def get_idx_list_per_classes(dm_dataset: DmDataset) -> dict[str, list[int]]:
+    """Compute class statistics."""
+    labels = dm_dataset.categories().get(AnnotationType.label, LabelCategories())
+
+    def get_label(ann: Annotation) -> str:
+        """Get the name of the label associated with the given annotation."""
+        try:
+            return labels.items[ann.label].name if ann.label is not None else None
+        except IndexError:
+            return ann.label
```
vinnamkim:
```suggestion
        return ann.label
```
Is there any mandatory reason to obtain a string key?

src/otx/algo/samplers/balanced_sampler.py
```diff
@@ -0,0 +1,174 @@
...
+            generator.manual_seed(seed)
+        else:
+            generator = self.generator
+
+        indices = []
+        for _ in range(self.repeat):
+            for _ in range(self.num_trials):
+                indice = torch.cat(
+                    [
+                        torch.tensor(self.img_indices[cls_indices], dtype=torch.int64)[
```
vinnamkim: `self.img_indices[cls_indices]`  can be a large vector. Change this to `torch.tensor` in advance, not to create everytime in the for loop.

src/otx/algo/samplers/balanced_sampler.py
```diff
@@ -0,0 +1,174 @@
...
+                            torch.multinomial(torch.ones(len(self.img_indices[cls_indices])), 1, generator=generator)
```
vinnamkim: As the same reason as https://github.com/openvinotoolkit/training_extensions/pull/2993/files#r1505322508, `torch.multinomial(torch.ones(len(self.img_indices[cls_indices]))` is inefficient. Use `torch.randint` rather than creating `torch.ones(len(self.img_indices[cls_indices])` and using `torch.multinomial()`. In addition, it is inefficient just sample only one index each time. Find another way to sample a batch of indices to make it faster such as 
```python
        if self.replacement:
            for _ in range(self.num_samples // 32):
                yield from torch.randint(high=n, size=(32,), dtype=torch.int64, generator=generator).tolist()
            yield from torch.randint(high=n, size=(self.num_samples % 32,), dtype=torch.int64, generator=generator).tolist()
```

src/otx/algo/samplers/balanced_sampler.py
```diff
@@ -0,0 +1,174 @@
...
+            seed = int(torch.empty((), dtype=torch.int64).random_().item())
+            generator = torch.Generator()
+            generator.manual_seed(seed)
+        else:
+            generator = self.generator
+
+        indices = []
+        for _ in range(self.repeat):
+            for _ in range(self.num_trials):
+                indice = torch.cat(
```
vinnamkim:
```suggestion
                index = torch.cat(
```

src/otx/algo/samplers/balanced_sampler.py
harimkang: Thanks. Done - https://github.com/openvinotoolkit/training_extensions/pull/2993/commits/6eb345a7526d42dafb6351fc7de10e41341786b8

src/otx/algo/samplers/balanced_sampler.py
harimkang: Thanks - i change this to use just ann.label - https://github.com/openvinotoolkit/training_extensions/pull/2993/commits/6eb345a7526d42dafb6351fc7de10e41341786b8

src/otx/algo/samplers/balanced_sampler.py
harimkang: I modified this so that `img_indices` has a value as a tensor when it is created.
- https://github.com/openvinotoolkit/training_extensions/pull/2993/commits/6eb345a7526d42dafb6351fc7de10e41341786b8

src/otx/algo/samplers/balanced_sampler.py
harimkang: Thanks for the nice comment, I replaced `torch.multinomial(torch.ones(len(self.img_indices[cls_indices]))` with `torch.randint(0, len(self.img_indices[cls_indices]), (1,), generator=self.generator)`

src/otx/algo/samplers/balanced_sampler.py
harimkang: Thanks - Done https://github.com/openvinotoolkit/training_extensions/pull/2993/commits/6eb345a7526d42dafb6351fc7de10e41341786b8

src/otx/core/data/module.py
```diff
@@ -147,6 +149,8 @@ def train_dataloader(self) -> DataLoader:
             "pin_memory": True,
             "collate_fn": dataset.collate_fn,
             "persistent_workers": config.num_workers > 0,
+            "sampler": sampler,
+            "shuffle": sampler is None,
```
jaegukhyun, line 153: Can we rely on data shuffling to custom sampler. I think some sampler wants shuffled dataset.
And also I cannot find shuffling in OTXSampler. Does this come from OTX1.x?

src/otx/core/utils/instantiators.py
```diff
@@ -92,3 +96,27 @@ def partial_instantiate_class(init: list | dict | None) -> list[partial] | None:
...
+        Sampler: The instantiated sampler object.
+    """
+    class_module, class_name = sampler_config.class_path.rsplit(".", 1)
+    module = __import__(class_module, fromlist=[class_name])
+    sampler_class = getattr(module, class_name)
+    sampler_kwargs = {**sampler_config.init_args, **kwargs}
+
+    init_signature = list(inspect.signature(sampler_class.__init__).parameters.keys())
+    if "samples_per_gpu" in init_signature:
+        sampler_kwargs["samples_per_gpu"] = batch_size
```
jaegukhyun: I think samples per gpu is too mm related words. Can we modify this to batch size?

src/otx/engine/utils/auto_configurator.py
```diff
@@ -207,13 +208,16 @@ def get_datamodule(self) -> OTXDataModule | None:
...
+        test_config = data_config.pop("test_subset")
         return OTXDataModule(
             task=self.config["data"]["task"],
             config=DataModuleConfig(
-                train_subset=SubsetConfig(**data_config.pop("train_subset")),
-                val_subset=SubsetConfig(**data_config.pop("val_subset")),
-                test_subset=SubsetConfig(**data_config.pop("test_subset")),
+                train_subset=SubsetConfig(sampler=SamplerConfig(**train_config.pop("sampler", {})), **train_config),
+                val_subset=SubsetConfig(sampler=SamplerConfig(**val_config.pop("sampler", {})), **val_config),
+                test_subset=SubsetConfig(sampler=SamplerConfig(**test_config.pop("sampler", {})), **test_config),
```
jaegukhyun, line 220: Would this SamplerConfig disappear for easy instantiation?

src/otx/recipe/_base_/data/mmdet_base.yaml
```diff
@@ -24,6 +24,8 @@ config:
       - prob: 0.5
         type: RandomFlip
       - type: PackDetInputs
+    sampler:
+      class_path: torch.utils.data.RandomSampler
```
jaegukhyun, line 28: I thought that base sampler of OTX1.x is BalancedSampler, if so could you explain the reason why we change default sampler type in OTX2.x?

src/otx/core/data/module.py
harimkang: https://github.com/pytorch/pytorch/blob/0dbef1618f19d69af075bc0c321f7c15c3595bfa/torch/utils/data/dataloader.py#L326
DataLoader does not normally allow shuffle and Sampler to coexist.

src/otx/core/utils/instantiators.py
harimkang: Thanks, On second look, `BalancedSampler` no longer uses `batch_size`, so I removed it.
https://github.com/openvinotoolkit/training_extensions/pull/2993/commits/d88d1b641ec1b4d603b945c875ceac8846309862

src/otx/engine/utils/auto_configurator.py
harimkang: In the CLI, the nested dataclass structure makes it difficult for objects to be partially instantiated, so I used a dataclass called `SamplerConfig` for this PR.
I think this could be removed when `DataModuleConfig` is refactored.

src/otx/recipe/_base_/data/mmdet_base.yaml
harimkang: I'm going to merge this PR first, then apply `BalancedSampler` to Regression Test to check the performance, and if all goes well after that, I'll apply it by default.

jaegukhyun (approved):

src/otx/algo/samplers/balanced_sampler.py
vinnamkim: Improve it further to sample a bulk of random indices in the next PR. Refer to
```python
        if self.replacement:
            for _ in range(self.num_samples // 32):
                yield from torch.randint(high=n, size=(32,), dtype=torch.int64, generator=generator).tolist()
            yield from torch.randint(high=n, size=(self.num_samples % 32,), dtype=torch.int64, generator=generator).tolist()
```

vinnamkim (approved):

src/otx/algo/samplers/balanced_sampler.py
harimkang: Yes I was thinking about this now, I will change this next time, thank you.

eunwoosh (approved):

PR #2994, opened 2024-02-27
Introduce the accuracy and change meta_info -> label_info

sungmanc:
### Summary

This PR introduces,
* Implement the `CustomAccuracy` series, to make same results with the OTX1.x.
* Converted all `meta_info` to `label_info`. Although `DataModule` deliver the `meta_info` to the `LitModule`, 
`meta_info` is only consumed at the `HLabelClassification`. It should be applied to the all tasks.
  * Add the `_reset_prediction_layers` to the `OVModel` and `VPM Models`
* Remove sort algorithm when `HLabelInfo` is generated to unify the order of `label_groups` and hlabel information.

### Checklist

### License

```python
```


src/otx/core/model/entity/base.py
```diff
@@ -494,7 +494,24 @@ def _read_ptq_config_from_ir(self, ov_model: Model) -> dict[str, Any]:
...
         return argparser.instantiate_classes(initial_ptq_config).as_dict()
 
+    def _reset_prediction_layer(self, num_classes: int) -> None:
+        return
+
     @property
     def model_adapter_parameters(self) -> dict:
         """Model parameters for export."""
         return {}
+
```
sungmanc, line 504: I just wonder why OVModel doesn't allow to update the num_classes, could please you explain? 
@kprokofi

src/otx/core/data/module.py
```diff
@@ -114,18 +114,18 @@ def __init__(
...
-        if self._is_meta_info_valid(meta_infos) is False:
+        if self._is_meta_info_valid(label_infos) is False:
             msg = "All data meta infos of subsets should be the same."
             raise ValueError(msg)
 
-        self.meta_info = next(iter(meta_infos))
+        self.label_info = next(iter(label_infos))
 
-    def _is_meta_info_valid(self, meta_infos: list[LabelInfo]) -> bool:
+    def _is_meta_info_valid(self, label_infos: list[LabelInfo]) -> bool:
```
harimkang, line 126: Shouldn't this function be renamed as well?

src/otx/core/metrics/accuracy.py
```diff
@@ -5,16 +5,198 @@
...
+
+            data_name = [label_group[0], "~" + label_group[0]]
+            confmat = NamedConfusionMatrix(task="binary", num_classes=2, row_names=data_name, col_names=data_name).to(
+                self.device,
+            )
+            conf_matrices.append(confmat(valid_preds, valid_targets))
+        return conf_matrices
+
+
+class CustomHlabelAccuracy(CustomAccuracy):
```
harimkang: Shouldn't we explain what makes `CustomHlabelAccuracy` different from traditional `HlabelAccuracy`? It seems hard to understand as a user, maybe we should add some more content to the docstring.

src/otx/core/metrics/accuracy.py
```diff
@@ -5,16 +5,198 @@
...
+        self.conf_matrix = ConfusionMatrix(task=task, num_classes=num_classes)
+        self.col_names = col_names
+        self.row_names = row_names
+
+    def __call__(self, *args: object, **kwargs: object):
+        """Call function of the Named Confusion Matrix."""
+        return self.conf_matrix(*args, **kwargs)
+
+
+class CustomAccuracy(Metric):
```
harimkang: Similarly for the other matrices, if OTX is custom, we need to explain what part of it is custom.

src/otx/core/model/module/classification.py
```diff
@@ -56,12 +56,22 @@ def __init__(
...
+
     def _log_metrics(self, meter: Metric, key: str) -> None:
         results = meter.compute()
         if results is None:
             msg = f"{meter} has no data to compute metric or there is an error computing metric"
             raise RuntimeError(msg)
 
+        # Custom Accuracy returns the dictionary, and accuracy value is in the `accuracy` key.
+        if isinstance(results, dict):
+            results = torch.tensor(results["accuracy"])
```
harimkang, line 73: I wonder if the result always has 'accuracy', even when using other matrices.

src/otx/core/metrics/fmeasure.py
```diff
@@ -652,7 +654,7 @@ def __init__(
         self.cross_class_nms = cross_class_nms
         self.preds: list[list[tuple]] = []
         self.targets: list[list[tuple]] = []
-        self.num_classes: int | None = None
+        self.num_classes = num_classes
```
harimkang, line 657: @eugene123tw I think this is probably the part that relates to the issue you encountered. Please check it out.

harimkang (commented):
Thanks for doing this! i left some comments and questions.

src/otx/core/metrics/accuracy.py
```diff
@@ -5,16 +5,198 @@
...
 
+if TYPE_CHECKING:
+    from torch import Tensor
 
-class HLabelAccuracy(Metric):
-    """Custom accuracy metric for h-label classification.
+    from otx.core.data.dataset.base import LabelInfo
+
+
+class NamedConfusionMatrix(nn.Module):
```
vinnamkim: Why don't you make it derived from `ConfusionMatrix`?

src/otx/core/data/dataset/segmentation.py
```diff
@@ -64,15 +64,15 @@ def __init__(
             image_color_channel,
             stack_images,
         )
-        self.meta_info = SegMetaInfo(
-            label_names=self.meta_info.label_names,
-            label_groups=self.meta_info.label_groups,
+        self.label_info = SegMetaInfo(
```
vinnamkim: Should this be renamed into `SegMetaInfo -> SegLabelInfo`?

src/otx/core/model/module/classification.py
sungmanc: For the classification, currently yes. All custom matrices returns the accuracy key in the results. For the future implementation, uses need to returns the same way like this.

src/otx/core/metrics/accuracy.py
sungmanc: I tried that way, however, I got the below error. I'm not sure why this happen, that's the reason why I selected this way.

```python3
class NamedConfusionMatrix(ConfusionMatrix):
    """Named Confusion Matrix to add row, col label names."""

    def __init__(
        self,
        task: str,
        num_classes: int,
        col_names: list[str],
        row_names: list[str],
    ):
        super().__init__(task=task, num_classes=num_classes)
        self.col_names = col_names
        self.row_names = row_names
```

ValueError: Unexpected keyword arguments: `col_names`, `row_names`

src/otx/core/metrics/accuracy.py
sungmanc: Actually, there is no traditional HLabelAccuracy. I added docstring what it really do. https://github.com/openvinotoolkit/training_extensions/pull/2994/commits/680ccc68aee3ef0d9d051a54517d348a867a2849

src/otx/core/metrics/accuracy.py
sungmanc: Thanks for the pointing out, https://github.com/openvinotoolkit/training_extensions/pull/2994/commits/680ccc68aee3ef0d9d051a54517d348a867a2849

src/otx/core/data/module.py
sungmanc: Done, https://github.com/openvinotoolkit/training_extensions/pull/2994/commits/1e9da826b42e4f1bb6ad08cfdd493c076fe02d2e

src/otx/core/data/dataset/segmentation.py
sungmanc: Done, https://github.com/openvinotoolkit/training_extensions/pull/2994/commits/1e9da826b42e4f1bb6ad08cfdd493c076fe02d2e. I also changed `HLabelMetaInfo --> HLabelInfo` and previous `HLabelInfo` --> `HLabelData`

src/otx/core/metrics/accuracy.py
vinnamkim:
```python
from typing import Any, Literal, Optional
from torchmetrics import ConfusionMatrix


class NamedConfusionMatrix(ConfusionMatrix):
    """Named Confusion Matrix to add row, col label names."""

    def __new__(
        cls,
        col_names: list[str],
        row_names: list[str],
        task: Literal["binary", "multiclass", "multilabel"],
        threshold: float = 0.5,
        num_classes: Optional[int] = None,
        num_labels: Optional[int] = None,
        normalize: Optional[Literal["true", "pred", "all", "none"]] = None,
        ignore_index: Optional[int] = None,
        validate_args: bool = True,
        **kwargs: Any,
    ) -> NamedConfusionMatrix:
        confusion_metric = super().__new__(
            cls,
            task=task,
            threshold=threshold,
            num_classes=num_classes,
            num_labels=num_labels,
            normalize=normalize,
            ignore_index=ignore_index,
            validate_args=validate_args,
            **kwargs,
        )

        confusion_metric.col_names = col_names
        confusion_metric.row_names = row_names
        return confusion_metric

    @property
    def col_names(self) -> list[str]:
        return self.col_names

    @property
    def row_names(self) -> list[str]:
        return self.row_names


matrix = NamedConfusionMatrix(["a", "b"], ["a", "b"], task="multiclass", num_classes=2)
print(matrix.col_names)
```

src/otx/core/metrics/accuracy.py
vinnamkim: Can we rename this to `HLabelAccuracy`

src/otx/core/metrics/accuracy.py
vinnamkim: Is there a possibility of this being used somewhere other than hlabel?

src/otx/core/metrics/accuracy.py
sungmanc: Done, https://github.com/openvinotoolkit/training_extensions/pull/2994/commits/1b3ff689e2b28f59cc011324a952d5b405d88424

src/otx/core/metrics/accuracy.py
sungmanc: `CustomAccuracy` is used for `CustomMulticlassAccuracy`, `CustomMultilabelAccuracy`, and `CustomHlabelAccuracy`.

src/otx/core/metrics/accuracy.py
sungmanc: We always open to change the name, however, for this case, I think `CustomHLabelAccuracy` is better since all other metrics including this are inherited from the `CustomAccuracy` and this comes from the different behavior that averages through the label_groups.

src/otx/core/metrics/accuracy.py
vinnamkim: Can you explain why those two classes (`CustomMulticlassAccuracy`, `CustomMultilabelAccuracy`) are needed? From your docstrings I can see that all of them are required to deal with `label_group`. However, I think that `label_group` is meaningless for both cases except h-label. If we consider this, I think that we can rename those to

1. `MulticlassAccuracyForLabelGroup` or `_MulticlassAccuracyForLabelGroup` (private only for `HlabelAccuracy`)
2. `MultilabelAccuracyForLabelGroup` or `_MultilabelAccuracyForLabelGroup` (private only for `HlabelAccuracy`)
3. `HlabelAccuracy`

src/otx/core/data/dataset/segmentation.py
vinnamkim: Why do we need `HLabelData` in addition to `HlabelInfo`? If not necessary, can we merge them in the next PR?

sungmanc: Sure, I'll merge at the next PR

src/otx/core/metrics/accuracy.py
sungmanc: You're right, I'll change the name. For the Geti and regression tests, we need to use `MulticlassAccuracyForLabelGroup`.

sungmanc: Done, https://github.com/openvinotoolkit/training_extensions/pull/2994/commits/b7ad0596c8f5e215ac3fb6435309215ba20d484d

vinnamkim (approved):

harimkang (approved):

jaegukhyun (approved):

PR #2997, opened 2024-02-27
Forward_explain & Export with XAI: Detection & InstSeg

GalyaZalesskaya:
### Summary

- Detection
- - Forward_explain 
- - Export with XAI
- - Tests
- - Validation on 
- - - PyToch task
- - - OV task 

- Instance Segmentation 
- - Forward_explain 
- - Export with XAI 
- - Tests
- - Validation on 
- - - PyToch task
- - - OV task 

TODO

- [x] Tests for Inst Segm
- [x] Fix pre-commit
- [x] Fix comments
- [x] Check the validity of saliency maps for `otx test` on XML with features
- [x] Check saving saliency maps for `otx explain` for PyTorch task

### Checklist

### License

```python
```


src/otx/core/model/entity/detection.py
```diff
@@ -84,48 +93,102 @@ def _export_parameters(self) -> dict[str, Any]:
...
+    ) -> T_OTXBatchPredEntity | OTXBatchLossEntity:
+        """Model forward function."""
+        self.model.explain_fn = self.get_explain_fn()
+
+        # If customize_inputs is overridden
+        outputs = (
+            self._forward_explain_detection(self.model, **self._customize_inputs(inputs))
+            if self._customize_inputs != ExplainableOTXDetModel._customize_inputs
+            else self.model(inputs)
+        )
```
negvet, line 125: I wonder if there is the case when `self._customize_inputs == ExplainableOTXDetModel._customize_inputs`? In this case we need to update line 114 as well, right?

I know, the same question applied for classification.

negvet: If there is not such a case, then we might eliminate this if-else statement.

src/otx/core/model/entity/detection.py
```diff
@@ -84,48 +93,102 @@ def _export_parameters(self) -> dict[str, Any]:
...
+        for param in self.parameters():
+            param.requires_grad = False
+
+        backbone_feat = self.extract_feat(inputs)
+        bbox_head_feat = self.bbox_head.forward(backbone_feat)
 
+        # Return the first output form detection head: classification scores
+        saliency_map = self.explain_fn(bbox_head_feat[0])
+
+        if mode == "loss":
```
negvet: We do not need `mode == "loss"` here I believe

src/otx/core/exporter/mmdeploy.py
```diff
@@ -63,6 +63,7 @@ def __init__(
         swap_rgb: bool = False,
         metadata: dict[tuple[str, str], str] | None = None,
         max_num_detections: int = 0,
+        additional_output_names: tuple = (),
```
sovrasov: tuple[str...]?

src/otx/core/model/entity/base.py
```diff
@@ -161,10 +161,7 @@ def _reset_model_forward(self) -> None:
     def _restore_model_forward(self) -> None:
         pass
 
-    def forward_tiles(
-        self,
-        inputs: T_OTXTileBatchDataEntity,
-    ) -> T_OTXBatchPredEntity | T_OTXBatchPredEntityWithXAI | OTXBatchLossEntity:
+    def forward_tiles(self, inputs: T_OTXTileBatchDataEntity) -> T_OTXBatchPredEntity | OTXBatchLossEntity:
```
sovrasov: Don't we expect XAI for tiling?

src/otx/core/model/entity/detection.py
```diff
@@ -84,48 +93,102 @@ def _export_parameters(self) -> dict[str, Any]:
                 ("model_info", "iou_threshold"): str(0.5),
             },
         )
+        parameters["additional_output_names"] = ("saliency_map") if self.explain_mode else ()
```
sovrasov: OTXModelExporter already contains `output_names` parameter in constructor. Can we use it (like in classification)?

sovrasov: Also, I'd rather move this parameter to `ExplainableOTXDetModel`

src/otx/core/model/entity/instance_segmentation.py
```diff
@@ -127,22 +134,101 @@ def _export_parameters(self) -> dict[str, Any]:
...
+
+        func_type = types.MethodType
+        self.model.forward = func_type(self.original_model_forward, self.model)
+        self.original_model_forward = None
+
+    @property
+    def _export_parameters(self) -> dict[str, Any]:
+        """Defines parameters required to export a particular model implementation."""
+        parameters = super()._export_parameters
+        parameters["output_names"] = ["saliency_map", "feature_vector"] if self.explain_mode else None
```
sovrasov: All the outputs should be listed here, and probably this doesn't work for mmdeploy when converting to onnx

src/otx/core/model/entity/base.py
GalyaZalesskaya: Will add in the next PR

GalyaZalesskaya:
The PR is ready for review, there are only pre-commit issues left to fix

src/otx/core/exporter/mmdeploy.py
```diff
@@ -63,14 +63,19 @@ def __init__(
...
-        super().__init__(input_size, mean, std, resize_mode, pad_value, swap_rgb, metadata)
+        super().__init__(input_size, mean, std, resize_mode, pad_value, swap_rgb, metadata, output_names)
         self._model_builder = model_builder
         model_cfg = convert_conf_to_mmconfig_dict(model_cfg, "list")
         self._model_cfg = MMConfig({"model": model_cfg, "test_pipeline": list(map(to_tuple, test_pipeline))})
         self._deploy_cfg = deploy_cfg if isinstance(deploy_cfg, MMConfig) else load_mmconfig_from_pkg(deploy_cfg)
 
         patch_input_shape(self._deploy_cfg, input_size[3], input_size[2])
+        if output_names is not None:
+            self._deploy_cfg.ir_config.output_names.extend(output_names)
```
sovrasov, line 76: It seems like we would have duplicated names here if mmdeploy config already contains something.

src/otx/algo/hooks/recording_forward_hook.py
```diff
@@ -411,48 +390,45 @@ def create_and_register_hook(cls, num_classes: int) -> BaseRecordingForwardHook:
...
             np.array: Class-wise Saliency Maps. One saliency map per each class - [batch, class_id, H, W]
         """
-        _, height, width = pred.masks[0].data.shape
-        masks, scores, labels = (
-            pred.masks[0].data,
-            pred.scores[0].data,
-            pred.labels[0].data,
-        )
-        saliency_maps = torch.zeros((num_classes, height, width), dtype=torch.float32)
+        masks, scores, labels = (pred.masks.data, pred.scores.data, pred.labels.data)
```
sovrasov: Are these entities batch predictions, or single images ones?

src/otx/algo/hooks/recording_forward_hook.py
```diff
@@ -466,6 +442,5 @@ def average_and_normalize(
 
         saliency_maps = saliency_maps.reshape((num_classes, -1))
         saliency_maps = cls._normalize_map(saliency_maps)
-        saliency_maps = saliency_maps.reshape(num_classes, height, width)
 
-        return saliency_maps.numpy()
+        return saliency_maps.reshape(num_classes, height, width)
```
sovrasov, line 454: According to the docstring, the shape is [batch, class_id, H, W]

src/otx/algo/hooks/recording_forward_hook.py
```diff
@@ -411,48 +390,45 @@ def create_and_register_hook(cls, num_classes: int) -> BaseRecordingForwardHook:
...
         Args:
-            preds (List[InstanceSegBatchPredEntity]): Predictions of Instance Segmentation model.
+            preds (InstanceSegBatchPredEntity | dict): Predictions of Instance Segmentation model.
 
         Returns:
             list[np.array]: Class-wise Saliency Maps. One saliency map per each class - [batch, class_id, H, W]
         """
         # TODO(gzalessk): Add unit tests # noqa: TD003
-        batch_size = len(preds)
+        batch_size = len(predictions)
```
sovrasov: If we use .append() in the loop over `predictions`, we can avoid explicit list allocation and simplify the loop as a bonus.

src/otx/core/exporter/mmdeploy.py
GalyaZalesskaya: In Detection and Instance Segmentation cases `ir_config.output_names` contains [`boxes`, `labels`, `masks`] and in case of explain mode I add export_parameters["output_names"] with `saliency_map`, `feature_vector`. So we shouldn't have duplicated output names. Also, if we for some reason have duplicates, and the number of outputs for the model isn't equal to the number of output names, we'll catch an error with output number mismatch

src/otx/algo/hooks/recording_forward_hook.py
GalyaZalesskaya: It's a single image prediction, we process batch predictions above in `MaskRCNNRecordingForwardHook.func`

src/otx/algo/hooks/recording_forward_hook.py
GalyaZalesskaya: Thank you for noticing, will fix

src/otx/algo/hooks/recording_forward_hook.py
GalyaZalesskaya: Good point

src/otx/core/model/entity/classification.py
```diff
@@ -77,11 +77,11 @@ def forward_explain(
...
         # If customize_inputs is overridden
-        outputs = (
-            self._forward_explain_image_classifier(self.model, **self._customize_inputs(inputs))
-            if self._customize_inputs != ExplainableOTXClsModel._customize_inputs
-            else self.model(inputs)
-        )
+        if self._customize_inputs != ExplainableOTXClsModel._customize_inputs:
+            customized_inputs = self._customize_inputs(inputs)
+        else:
+            customized_inputs = {inputs}
```
negvet: Should a set of inputs be created here?
Double asterisks operator would work only with dict, with set (not mapping, just iterable object) you would need just one asterisk.

Do we have a case when `self._customize_inputs == ExplainableOTXClsModel._customize_inputs` to test it? I do not see one..

src/otx/core/model/entity/instance_segmentation.py
```diff
@@ -127,22 +134,101 @@ def _export_parameters(self) -> dict[str, Any]:
...
+        outputs = self._forward_explain_inst_seg(self.model, **customized_inputs)
+
+        return (
+            self._customize_outputs(outputs, inputs)
+            if self._customize_outputs != ExplainableOTXInstanceSegModel._customize_outputs
+            else outputs
+        )
+
+    @staticmethod
+    def _forward_explain_inst_seg(self, inputs: torch.Tensor, data_samples: SampleList, mode: str) -> dict:
```
negvet: Please specify type of `self`

src/otx/core/model/entity/instance_segmentation.py
```diff
@@ -127,22 +134,101 @@ def _export_parameters(self) -> dict[str, Any]:
...
+
+    def get_explain_fn(self) -> Callable:
+        """Returns explain function."""
         from otx.algo.hooks.recording_forward_hook import MaskRCNNRecordingForwardHook
 
-        self.explain_hook = MaskRCNNRecordingForwardHook.create_and_register_hook(
+        explainer = MaskRCNNRecordingForwardHook.create_and_register_hook(
             num_classes=self.num_classes,
         )
+        return explainer.func
```
negvet, line 200: It is possible to align it with detection: remove create_and_register_hook and create instance right here?

tests/integration/api/test_xai.py
```diff
@@ -0,0 +1,136 @@
...
+DETECTION_LIST = [recipe for recipe in RECIPE_LIST_ALL if "/detection" in recipe and "tile" not in recipe]
+INST_SEG_LIST = [recipe for recipe in RECIPE_LIST_ALL if "instance_segmentation" in recipe and "tile" not in recipe]
+EXPLAIN_MODELS = MC_ML_CLS + DETECTION_LIST + INST_SEG_LIST
+
+MEAN_TORCH_OV_DIFF = 150
+
+
+@pytest.mark.parametrize(
+    "recipe",
+    MULTI_CLASS_CLS,
```
negvet: `MULTI_CLASS_CLS` -> `MULTI_CLASS_CLS + DETECTION_LIST + INST_SEG_LIST`?

tests/integration/cli/test_cli.py
```diff
@@ -318,12 +321,20 @@ def test_otx_explain_e2e(
...
     reference_sal_vals = {
+        # Classification
         "multi_label_cls_efficientnet_v2_light": np.array([66, 97, 84, 33, 42, 79, 0], dtype=np.uint8),
         "h_label_cls_efficientnet_v2_light": np.array([43, 84, 61, 5, 54, 31, 57], dtype=np.uint8),
+        # Detection
+        "detection_yolox_tiny": np.array([57, 96, 77, 30, 25, 30, 43, 32, 55, 33], dtype=np.uint8),
+        "detection_ssd_mobilenetv2": np.array([62, 32, 27, 13, 6, 11, 18, 21, 21, 19], dtype=np.uint8),
+        "detection_atss_mobilenetv2": np.array([22, 62, 64, 0, 27, 60, 59, 53, 37, 45], dtype=np.uint8),
+        # Instance Segmentation
+        "instance_segmentation_maskrcnn_efficientnetb2b": np.array([0, 0, 0, 0, 0, 0, 0, 0, 0, 0], dtype=np.uint8),
```
negvet: zeros are because the model is not trained?

src/otx/core/model/entity/classification.py
GalyaZalesskaya: Me too, actually. I'm not sure in which case this happens. 
I put inputs in dict in `else` branch exactly for the purpose of getting initial `inputs` after passing that as `**{inputs}` below. The type of `inputs` shouldn't affect that

src/otx/core/model/entity/instance_segmentation.py
GalyaZalesskaya: Will do together with fixing precommit, thank you :)

src/otx/core/model/entity/instance_segmentation.py
GalyaZalesskaya: Totally missed it, thank you for noticing

tests/integration/api/test_xai.py
GalyaZalesskaya: Will have to update this asseqrt condition with proper types but true:
`assert isinstance(predict_result[0], MulticlassClsBatchPredEntity)`
`assert isinstance(predict_result_explain[0], MulticlassClsBatchPredEntityWithXAI)`

tests/integration/cli/test_cli.py
GalyaZalesskaya: Because saliency maps are combined from object masks. So, the overall map is sparse and has the resolution of the initial image ((720, 1280) for instance). That's why the upper row of the saliency map doesn't have objects and contains only zeros

negvet: What about picking some other row, rather than the top one for mrcnn? just to pull this test from dummy mode

tests/integration/api/test_xai.py
negvet: Indeed, this needs to be updated

src/otx/core/model/entity/classification.py
negvet: with `obj = {value}`, obj is a set, not a dict, `**obj` would fail

GalyaZalesskaya: Thank you, I missed it

GalyaZalesskaya:
Closed due to too many fixes and comments that make reviewing too complicated.

PR #3011, opened 2024-02-29
Support dumping saliency maps

negvet:
### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algo/utils/xai_utils.py
+++ b/src/otx/algo/utils/xai_utils.py
@@ -4,9 +4,12 @@
 
 from __future__ import annotations
 
+from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
 import cv2
+import numpy as np
+from datumaro import Image
 
 from otx.core.config.explain import ExplainConfig
 from otx.core.data.entity.base import OTXBatchPredEntityWithXAI
@@ -14,29 +17,25 @@ from otx.core.data.entity.instance_segmentation import InstanceSegBatchPredEntit
 from otx.core.types.explain import TargetExplainGroup
 
 if TYPE_CHECKING:
-    from pathlib import Path
+    from lightning.pytorch.utilities.types import EVAL_DATALOADERS
 
-    import numpy as np
+    from otx.core.data.module import OTXDataModule
 
 
 def process_saliency_maps_in_pred_entity(
     predict_result: list[Any] | list[OTXBatchPredEntityWithXAI | InstanceSegBatchPredEntityWithXAI],
     explain_config: ExplainConfig,
-    work_dir: Path | None = None,
 ) -> list[Any] | list[OTXBatchPredEntityWithXAI | InstanceSegBatchPredEntityWithXAI]:
     """Process saliency maps in PredEntity."""
     for predict_result_per_batch in predict_result:
         saliency_maps = predict_result_per_batch.saliency_maps
+        imgs_info = predict_result_per_batch.imgs_info
+        ori_img_shapes = [img_info.ori_shape for img_info in imgs_info]
         pred_labels = predict_result_per_batch.labels  # type: ignore[union-attr]
         if pred_labels:
             pred_labels = [pred.tolist() for pred in pred_labels]
 
-        processed_saliency_maps = process_saliency_maps(saliency_maps, explain_config, pred_labels)
-
-        if processed_saliency_maps and work_dir:
-            # Temporary saving random saliency map for image 0 (for tests)
-            s_map_to_save = next(iter(processed_saliency_maps[0].values()))
-            cv2.imwrite(str(work_dir / "saliency_map.tiff"), s_map_to_save)
+        processed_saliency_maps = process_saliency_maps(saliency_maps, explain_config, pred_labels, ori_img_shapes)
 
         predict_result_per_batch.saliency_maps = processed_saliency_maps
     return predict_result
@@ -46,6 +45,7 @@ def process_saliency_maps(
     saliency_maps: list,
     explain_config: ExplainConfig,
     pred_labels: list | None,
+    ori_img_shapes: list,
 ) -> list[dict[Any, Any]]:
     """Perform saliency map convertion to dict and post-processing."""
     if explain_config.target_explain_group == TargetExplainGroup.ALL:
@@ -60,7 +60,9 @@ def process_saliency_maps(
 
     if explain_config.postprocess:
         for i in range(len(processed_saliency_maps)):
-            processed_saliency_maps[i] = {key: postprocess(s_map) for key, s_map in processed_saliency_maps[i].items()}
+            processed_saliency_maps[i] = {
+                key: postprocess(s_map, ori_img_shapes[i]) for key, s_map in processed_saliency_maps[i].items()
+            }
 
     return processed_saliency_maps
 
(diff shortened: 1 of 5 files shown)


src/otx/algo/utils/xai_utils.py
```diff
@@ -101,10 +103,62 @@ def convert_maps_to_dict_image(saliency_maps: np.array) -> list[dict[Any, np.arr
     return [{"map_per_image": map_per_image} for map_per_image in saliency_maps]
 
 
-def postprocess(saliency_map: np.ndarray, output_size: tuple | list = (224, 224)) -> np.ndarray:
+def postprocess(saliency_map: np.ndarray, output_size: tuple | None) -> np.ndarray:
```
sovrasov: tuple[int...]?

sovrasov: Or rather tupe[int,int] since there's an explicit unpacking into h,w

src/otx/algo/utils/xai_utils.py
```diff
@@ -101,10 +103,62 @@ def convert_maps_to_dict_image(saliency_maps: np.array) -> list[dict[Any, np.arr
...
     """Postprocess single saliency map."""
     if saliency_map.ndim != 2:
         raise ValueError
```
GalyaZalesskaya:
```suggestion
        msg = "Shape mismatch."
        raise ValueError(msg)
```

src/otx/algo/utils/xai_utils.py
```diff
@@ -101,10 +103,62 @@ def convert_maps_to_dict_image(saliency_maps: np.array) -> list[dict[Any, np.arr
...
 
-    saliency_map = cv2.resize(saliency_map, output_size)
+    if output_size:
+        h, w = output_size
+        saliency_map = cv2.resize(saliency_map, (w, h))
     return cv2.applyColorMap(saliency_map, cv2.COLORMAP_JET)
+
+
+def dump_saliency_maps(
+    predict_result: list[Any] | list[OTXBatchPredEntityWithXAI | InstanceSegBatchPredEntityWithXAI],
```
GalyaZalesskaya: Can't we do:
list[OTXBatchPredEntityWithXAI | InstanceSegBatchPredEntityWithXAI | Any]
here?

src/otx/algo/utils/xai_utils.py
```diff
@@ -101,10 +103,62 @@ def convert_maps_to_dict_image(saliency_maps: np.array) -> list[dict[Any, np.arr
...
+
+    for predict_result_per_batch in predict_result:
+        saliency_maps = predict_result_per_batch.saliency_maps
+        imgs_info = predict_result_per_batch.imgs_info
+        for pred_index in range(len(saliency_maps)):
+            img_id = imgs_info[pred_index].img_idx
+            img_data, image_save_name = _get_image_data_name(datamodule, img_id)
+
+            for class_id, s_map in saliency_maps[pred_index].items():
+                file_name_map = Path(image_save_name + "_class_" + str(class_id) + "_saliency_map.png")
```
GalyaZalesskaya, line 137: Actually, can we use here '.jpg' format instead of '.png' since it saves some memory? Just from my experience, after saving a lot of saliency maps I had memory issues

src/otx/engine/engine.py
```diff
@@ -528,14 +528,16 @@ def explain(
         checkpoint: PathLike | None = None,
         datamodule: EVAL_DATALOADERS | OTXDataModule | None = None,
         explain_config: ExplainConfig | None = None,
+        dump: bool | None = False,
```
GalyaZalesskaya, line 531: I'd suggest calling it `output`, so if it's not None, it will be both a flag to dump saliency maps and path where to dump them.

src/otx/algo/utils/xai_utils.py
```diff
@@ -101,10 +103,62 @@ def convert_maps_to_dict_image(saliency_maps: np.array) -> list[dict[Any, np.arr
...
+                save_path_map = output_dir / file_name_map
+                cv2.imwrite(str(save_path_map), s_map)
+
+                if explain_config.postprocess:
+                    file_name_overlay = Path(image_save_name + "_class_" + str(class_id) + "_overlay.png")
```
GalyaZalesskaya, line 142: Same as above

src/otx/algo/utils/xai_utils.py
negvet: Sure we can, fixed

src/otx/algo/utils/xai_utils.py
negvet: Well, it is not that clear that ".jpg" would be better.

Usually, xai model analysis is done manually by a human. Which means that the number of inspecting images is limited (up to hundreds let's say). Which is not that dramatic in terms if memory. If the user accidently runs explain for a million of images, ".jpg" is not gonna save his memory (jpg still scales linearly).
With that being said, and assuming that ".jpg" is a lossy compression, I would propose to keep png. What do you think?

src/otx/engine/engine.py
negvet: I would propose to stick to `self.work_dir / "saliency_maps"` as the default place for storing saliency maps. What about `dump` as as saving flag, and `output` as an optional path?

src/otx/algo/utils/xai_utils.py
GalyaZalesskaya: Okay, let's keep it as it is. I actually had the memory problem while compiling presentation with different saliency map results. But it can be called a pretty rare usecase.

src/otx/engine/engine.py
GalyaZalesskaya: I think we can leave it as it is

GalyaZalesskaya (approved):

src/otx/algo/utils/xai_utils.py
```diff
@@ -101,10 +103,63 @@ def convert_maps_to_dict_image(saliency_maps: np.array) -> list[dict[Any, np.arr
...
+    """Sumps saliency maps (raw and with overlay)."""
+    output_dir = output_dir / "saliency_maps"
+    output_dir.mkdir(parents=True, exist_ok=True)
+
+    for predict_result_per_batch in predict_result:
+        saliency_maps = predict_result_per_batch.saliency_maps
+        imgs_info = predict_result_per_batch.imgs_info
+        for pred_index in range(len(saliency_maps)):
+            img_id = imgs_info[pred_index].img_idx
+            img_data, image_save_name = _get_image_data_name(datamodule, img_id)
```
vinnamkim, line 134: I hope this will be called for every batch in the `engine.predict()`. This is because 
1. No double image decoding at dataloader and this line
```python
            img_data, image_save_name = _get_image_data_name(datamodule, img_id)
```
2. Not maintaining all the saliency maps during test workload but flushing saliency map into a disk streamly.

I think you can do this by registering a custom callback to PyTorchLightning Trainer.

negvet: This proposal make sense. Currently we just replicated the OTX1.x behavior. Optimizing it further would require noticeable engineering effort. Also we have a FF this week, so I would suggest to postpone this optimization for the future.
See my comments below for more details:

> Not maintaining all the saliency maps during test workload but flushing saliency map into a disk steamily.

We probably need to introduce a new prediction entity to handle such case for API users. Currently every data point in the prediction contains the numpy object itself. In the naive approach, it is possible to flush the maps to the disk on the per-bath basis and store a reference to them. Something more complicated can be done here as well. This is not well aligned with what we have now.

> No double image decoding at dataloader.

It is possible to register a custom callback with some logic at `on_predict_batch_end`. But the issue here is that there is only transformed batch available, although I need the original images to create an overlay. Not sure if I know an easy way to handle it fast.

src/otx/engine/engine.py
harimkang, line 531: why we need `None` Type for `dump`?

src/otx/algo/utils/xai_utils.py
vinnamkim: > It is possible to register a custom callback with some logic at `on_predict_batch_end`. But the issue here is that there is only transformed batch available, although I need the original images to create an overlay. Not sure if I know an easy way to handle it fast.

Why not overlay it over a transformed image? Is the saliency map computed from the transformed image I'm thinking that it is more correct that we should overlay it over the transformed image. A simple example for this is that the saliency map is computed from the transformed image with random cropping. There is no way to overlay it correctly without knowing the cropping parameters. Even if we know them, for overlaying. we just have to do the same preprocessing to make the original image same as the previously transformed image.

Anyway, I hope it to be improved in the future.

vinnamkim (approved):

sungmanc (approved):

src/otx/engine/engine.py
negvet: This is a bug, will fix it in https://github.com/openvinotoolkit/training_extensions/pull/3014: https://github.com/openvinotoolkit/training_extensions/pull/3014/commits/22dc1dbb2ae1d03bc3b73ca8d15e37a21611da03

PR #3027, opened 2024-03-04
Support explain for OV models

GalyaZalesskaya:
### Summary

Change `engine/explain` to process OV model
Update `tests/integration/api/test_engine_api.py` to cover explain for Torch and IR models

### Checklist

- [ ] I have added unit tests to cover my changes.​
- [x] I have added integration tests to cover my changes.​
- [ ] I have added e2e tests for validation.
- [ ] I have added the description of my changes into CHANGELOG in my target branch (e.g., [CHANGELOG](https://github.com/openvinotoolkit/training_extensions/blob/develop/CHANGELOG.md) in develop).​
- [ ] I have updated the documentation in my target branch accordingly (e.g., [documentation](https://github.com/openvinotoolkit/training_extensions/tree/develop/docs) in develop).
- [ ] I have [linked related issues](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

### License

- [ ] I submit _my code changes_ under the same [Apache License](https://github.com/openvinotoolkit/training_extensions/blob/develop/LICENSE) that covers the project.
      Feel free to contact the maintainers if that's a concern.
- [ ] I have updated the license header for each file (see an example below).

```python
# Copyright (C) 2023 Intel Corporation
# SPDX-License-Identifier: Apache-2.0
```

PR #3028, opened 2024-03-04
Add checkpoint argument to optimize()

sovrasov:
### Summary

Engine.optimize() now can take model weights as a checkpoint argument

### Checklist

### License

```python
```


src/otx/engine/engine.py
```diff
@@ -516,24 +518,48 @@ def optimize(
...
                     --model <CONFIG | CLASS_PATH_OR_NAME> --data_root <DATASET_PATH, str>
                     --model.model_name=<PATH_TO_IR_XML, str>
                 ```
         """
+        if checkpoint is not None:
+            checkpoint = str(checkpoint)
+        elif self.checkpoint is not None:
+            checkpoint = str(self.checkpoint)
+        else:
+            checkpoint = None
```
harimkang:
```suggestion
        checkpoint = checkpoint if checkpoint is not None else self.checkpoint
```

sungmanc: It seems that checkpoint should be inserted with .xml or .onnx suffix. Can check point be none?

src/otx/engine/engine.py
```diff
@@ -497,6 +497,7 @@ def export(
 
     def optimize(
         self,
+        checkpoint: str | Path | None = None,
```
vinnamkim:
```suggestion
        checkpoint: PathLike | None = None,
```

vinnamkim:
I'm curious what the `Engine` is if we allow it to be stateless for every command such as this PR or https://github.com/openvinotoolkit/training_extensions/pull/3027?
Is it really helpful for the user experience? Is there any requirement to make it like this?

eunwoosh:
> I'm curious what the `Engine` is if we allow it to be stateless for every command such as this PR or #3027? Is it really helpful for the user experience? Is there any requirement to make it like this?

Actually, the start point from these variants are completely for increasing user experience.
When user tries to run `test` after training a torch model, user can execute them as below.
```
otx train --config src/otx/recipe/detection/atss_mobilenetv2.yaml ...
otx test --config src/otx/recipe/detection/atss_mobilenetv2.yaml --checkpoint ...
``` 
But before changes like this is merged, if user want to execute `test`, `optimize` or etc. using exported model,
user should execute as below.
```
otx train --config src/otx/recipe/detection/atss_mobilenetv2.yaml ...
otx export --config src/otx/recipe/detection/atss_mobilenetv2.yaml --checkpoint ...
otx test --config src/otx/recipe/detection/openvino_model.yaml --model.name openvino.xml ...
otx optimize --config src/otx/recipe/detection/openvino_model.yaml --model.name openvino.xml ...
```
This isn't intuitive in user experience perspective, I think.
But after this changes, user can execute as below.
```
otx train --config src/otx/recipe/detection/atss_mobilenetv2.yaml ...
otx export --config src/otx/recipe/detection/atss_mobilenetv2.yaml --checkpoint ...
otx test --config src/otx/recipe/detection/atss_mobilenetv2.yaml --checkpoint openvino.xml ...
otx optimize --config src/otx/recipe/detection/atss_mobilenetv2.yaml --checkpoint openvino.xml ...
```
I understand what 'stateless ' means you said, but I think it can increase UX a lot.

vinnamkim:
> I understand what 'stateless ' means you said, but I think it can increase UX a lot.

First of all, I understand that `Engine` is used for CLI. However, it is not desired to design `Engine` itself to achieve some CLI behaviors. If the CLI code flow is like

1. Take in CLI user commands
2. Parse the CLI commands to Engine
3. Running Engine function

To enhance the CLI UX, you have to improve the step 2, not the step 3.

harimkang:
I don't think this is necessarily due to CLI behavior - 
Engine.test, explain, optimize can actually use 2 checkpoints. One would be `torch`, and the other would be the `onnx/IR`.
For example, if user want to test, explain, optimize with the onnx/IR checkpoint without taking the argument checkpoint, we need to create new Engine. 
From a UX perspective, this would be uncomfortable, what do you guys think?

harimkang:
> I don't think this is necessarily due to CLI behavior - Engine.test, explain, optimize can actually use 2 checkpoints. One would be `torch`, and the other would be the `onnx/IR`. For example, if user want to test, explain, optimize with the onnx/IR checkpoint without taking the argument checkpoint, we need to create new Engine. From a UX perspective, this would be uncomfortable, what do you guys think?

eunwoosh:
I just thought CLI case and I agree Vinnam's opinion in this perspective. but as Harim said, we need to consider API user case. How do you think about it?

src/otx/engine/engine.py
sovrasov: None case is checked a couple of lines later

src/otx/engine/engine.py
sovrasov: Updated

src/otx/engine/engine.py
sovrasov: Updated

sovrasov:
`checkpoint` arg introduced here is optional, and we can still do a fallback to `self.checkpoint`, which means the stateful approach also works. The problem we're trying to solve here is raised by torch/OV models discrepancy, as Harim said. Because of some design issues at model entity level, we had a twisted CLI/API for OV models, so this and similar PRs are aiming to resolve it to some extent. 
We can discuss it further on OTX 2.0 forum I believe.

harimkang (approved):

vinnamkim (approved):

jaegukhyun (approved):

harimkang:
@sovrasov 
image
Can you check it out in the next PR?

The torch's score is lower than the OV and optimized score, can this happen? If so, I don't think this is a FAIL.

PR #3005, opened 2024-02-28
Improve inst seg perf

vinnamkim:
### Summary

- Ticket no. 128949
- Stack images as possible for all tasks
- Pad images during instance segmentation data pipeline as possible
- Add non_blocking option to instance segmentation data preprocessors

### How to test
Median train iter time for this command
```console
otx train --data_root /home/vinnamki/dataset/instance_seg/coco_car_person_medium --config src/otx/recipe/instance_segmentation/maskrcnn_efficientnetb2b.yaml 
```
is improved,
| Before | After |
| :-: | :-: |
| 0.247 | 0.236 |

### Checklist

- [ ] I have added unit tests to cover my changes.​
- [ ] I have added integration tests to cover my changes.​
- [ ] I have added e2e tests for validation.
- [ ] I have added the description of my changes into CHANGELOG in my target branch (e.g., [CHANGELOG](https://github.com/openvinotoolkit/training_extensions/blob/develop/CHANGELOG.md) in develop).​
- [ ] I have updated the documentation in my target branch accordingly (e.g., [documentation](https://github.com/openvinotoolkit/training_extensions/tree/develop/docs) in develop).
- [ ] I have [linked related issues](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

### License

- [ ] I submit _my code changes_ under the same [Apache License](https://github.com/openvinotoolkit/training_extensions/blob/develop/LICENSE) that covers the project.
      Feel free to contact the maintainers if that's a concern.
- [ ] I have updated the license header for each file (see an example below).

```python
# Copyright (C) 2023 Intel Corporation
# SPDX-License-Identifier: Apache-2.0
```

PR #3039, opened 2024-03-05
Add RTMDet Inst Tiny

eugene123tw:
### Summary

This PR introduces RTMDetInst Tiny. However, it's currently not functional, giving 0% mAP after long training on small datasets. I reckon it will require another sprint to enhance both speed and accuracy.

### Checklist

### License

```python
```


src/otx/recipe/instance_segmentation/rtmdet_inst_tiny.yaml
```diff
@@ -0,0 +1,146 @@
...
+            img_scale:
+              - 640
+              - 640
+            ratio_range:
+              - 1.0
+              - 1.0
+            max_cached_images: 10
+            random_pop: false
+            pad_val:
+              - 114
```
vinnamkim, line 90: image
The default pixel pad value of `DetDataPreprocessor` is 0. Update the model configuration to align with this value, 114. Please check other model templates as well.

harimkang:
Why does this model need to be added even though it is 0% mAP on small dataset?
If this is not an urgent task, shouldn't the issue be resolved and added together?
I say this because there are no results for that model, and I don't think it's a model for Geti compatibility.

eugene123tw:
> Why does this model need to be added even though it is 0% mAP on small dataset? If this is not an urgent task, shouldn't the issue be resolved and added together? I say this because there are no results for that model, and I don't think it's a model for Geti compatibility.

This model requires further optimization in both OpenVINO and PyTorch. Given the scope of the optimization process, it'd be easier for the reviewing process if I separate the initial introduction of RTMDet from the optimization phase, particularly since the optimization phase will be more extensive.

src/otx/recipe/instance_segmentation/rtmdet_inst_tiny.yaml
eugene123tw: @vinnamkim Padding to a square aspect ratio has already been applied before DetDataPreprocessor. Specifically, YOLOX and RTMDet use padding value of 114, and has the following aug:
```
Type: Pad
pad_to_square: true
pad_val: 114
```
image

vinnamkim: If each image in a batch has different size, `DetDataPreprocessor` will pad them with zero (default) to the maximal image size of them. I observed there is image size mismatch between samples in a batch of instance segmentation models while working on https://github.com/openvinotoolkit/training_extensions/pull/3005.

harimkang (approved):

src/otx/algo/detection/mmconfigs/yolox_l.yaml
```diff
@@ -21,6 +21,7 @@ data_preprocessor:
     - 1.0
     - 1.0
     - 1.0
+  pad_value: 114
```
harimkang, line 24: One question is, why do we need this?

eugene123tw: It's based on Vinnam's comment: https://github.com/openvinotoolkit/training_extensions/pull/3039#discussion_r1515327705

src/otx/recipe/instance_segmentation/rtmdet_inst_tiny.yaml
eugene123tw: I added pad_value to preprocessor to both yolox variants and RTMDetInst: https://github.com/openvinotoolkit/training_extensions/pull/3039/files#diff-caf5ba5b8d9c9b0ee78ccdb4dec72273f5a3b0d74913ca6d9b7c6fa9aa871faaR13

vinnamkim (approved):

sungmanc (approved):

jaegukhyun (approved):

PR #3047, opened 2024-03-06
Support feature vector for cls

negvet:
### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algo/hooks/recording_forward_hook.py
+++ b/src/otx/algo/hooks/recording_forward_hook.py
@@ -16,6 +16,24 @@ if TYPE_CHECKING:
     from torch.utils.hooks import RemovableHandle
 
 
+def feature_vector_fn(feature_map: torch.Tensor | Sequence[torch.Tensor]) -> torch.Tensor:
+    """Generate the feature vector by average pooling feature maps."""
+    if isinstance(feature_map, (list, tuple)):
+        # aggregate feature maps from Feature Pyramid Network
+        feature_vectors = [torch.nn.functional.adaptive_avg_pool2d(f, (1, 1)) for f in feature_map]
+        for i in range(len(feature_vectors)):
+            batch, num_channels, *_ = feature_vectors[i].shape
+            feature_vectors[i] = feature_vectors[i].view(batch, num_channels)
+
+        if len(feature_vectors) > 1:
+            return torch.cat(feature_vectors, 1)
+        return feature_vectors[0]
+
+    feature_vector = torch.nn.functional.adaptive_avg_pool2d(feature_map, (1, 1))
+    batch, num_channels, *_ = feature_vector.shape
+    return feature_vector.view(batch, num_channels)
+
+
 class BaseRecordingForwardHook:
     """While registered with the designated PyTorch module, this class caches feature vector during forward pass.
 
--- a/src/otx/algo/hooks/recording_forward_hook.py
+++ b/src/otx/algo/hooks/recording_forward_hook.py
@@ -20,18 +20,16 @@ def feature_vector_fn(feature_map: torch.Tensor | Sequence[torch.Tensor]) -> tor
     """Generate the feature vector by average pooling feature maps."""
     if isinstance(feature_map, (list, tuple)):
         # aggregate feature maps from Feature Pyramid Network
-        feature_vectors = [torch.nn.functional.adaptive_avg_pool2d(f, (1, 1)) for f in feature_map]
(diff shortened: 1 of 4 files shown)


src/otx/core/model/entity/classification.py
```diff
@@ -622,7 +654,7 @@ def _export_parameters(self) -> dict[str, Any]:
         export_params["via_onnx"] = False
         export_params["input_size"] = self.image_size
         export_params["onnx_export_configuration"] = None
-        export_params["output_names"] = ["logits", "saliency_map"] if self.explain_mode else None
+        export_params["output_names"] = ["logits", "feature_vector", "saliency_map"] if self.explain_mode else None
```
GalyaZalesskaya: What do you think about moving explain related parameters to src/otx/core/model/entity/classification.py?

ExplainableOTXClsModel ->
```
    @property
    def _export_parameters(self) -> dict[str, Any]:
        """Defines parameters required to export a particular model implementation."""
        export_params = super()._export_parameters
        export_params["output_names"] = ["logits", "feature_vector", "saliency_map"] if self.explain_mode else None
        return export_params
```

negvet: I think it is possible, my logic was to put it near `export_params["onnx_export_configuration"]`. But I can move it up. Probably some other parameters also can be moved to the upper level class, @sovrasov?

sovrasov: We have to add the parameters which make sense at the current level of hierarchy. In `ExplainableOTXClsModel` we already know what the outputs should be in explain mode, so we can move `export_params["output_names"]` there.

Other parameters here are implementation-specific (in out case that's mmpretrain), and I'd opt to leave them as is. Harim is working on torchvision wrapper, so we'd better keep parameters which could be not generalizable at lower levels.

sovrasov (dismissed):

harimkang (dismissed):

src/otx/algo/hooks/recording_forward_hook.py
```diff
@@ -16,6 +16,24 @@
...
 
 
+def feature_vector_fn(feature_map: torch.Tensor | Sequence[torch.Tensor]) -> torch.Tensor:
+    """Generate the feature vector by average pooling feature maps."""
+    if isinstance(feature_map, (list, tuple)):
+        # aggregate feature maps from Feature Pyramid Network
+        feature_vectors = [torch.nn.functional.adaptive_avg_pool2d(f, (1, 1)) for f in feature_map]
+        for i in range(len(feature_vectors)):
+            batch, num_channels, *_ = feature_vectors[i].shape
+            feature_vectors[i] = feature_vectors[i].view(batch, num_channels)
```
vinnamkim:
```suggestion
        feature_vectors = [
             # Spatially pooling and flatten, B x C x H x W => B x C'
             torch.nn.functional.adaptive_avg_pool2d(f, (1, 1)).flatten(start_dim=1)
             for f in feature_map
        ]
```

src/otx/algo/hooks/recording_forward_hook.py
```diff
@@ -16,6 +16,24 @@
...
+
+        if len(feature_vectors) > 1:
+            return torch.cat(feature_vectors, 1)
+        return feature_vectors[0]
+
+    feature_vector = torch.nn.functional.adaptive_avg_pool2d(feature_map, (1, 1))
+    batch, num_channels, *_ = feature_vector.shape
+    return feature_vector.view(batch, num_channels)
```
vinnamkim: Here as well like https://github.com/openvinotoolkit/training_extensions/pull/3047/files#r1515331667

src/otx/core/model/entity/classification.py
```diff
@@ -673,7 +709,7 @@ def _customize_outputs(
             # Squeeze dim 4D => 3D, (1, num_classes, H, W) => (num_classes, H, W)
             predicted_s_maps = [out.saliency_map[0] for out in outputs]
 
-            predicted_f_vectors = [out.feature_vector for out in outputs]
+            predicted_f_vectors = [out.feature_vector[0] for out in outputs]
```
vinnamkim, line 713: Would you add a comment here about the reason why we extract index = 0 from `out.feature_vector`?

negvet: Sure, added

vinnamkim (approved):

sovrasov (approved):

GalyaZalesskaya (approved):

PR #3060, opened 2024-03-08
Make the same config with OTX1.x : Edit the default factor for lr scheduler and update warmup_step logic

sungmanc:
### Summary

OTX2.x behavior is different with OTX1.X (reported from @jaegukhyun)
So, the goal is to make same behavior.

This PR introduces, 
* Change the default factor for the LRScheduler (0.5 -> 0.1)
* LR scheduler's patience will results the same results with the OTX1.x. Reduced by 1.
(OTX1.x scheduler patience works at >= patience, however, OTXv2 scheduler patience works at > patience)
* Warmup scheduler makes non-zero lr at the first iteration. 

### Checklist

### License

```python
```


jaegukhyun (dismissed):

sungmanc:
For the review convenience : https://github.com/openvinotoolkit/training_extensions/pull/3060/files?file-filters%5B%5D=.py&show-viewed-files=true

src/otx/algo/callbacks/adaptive_train_scheduling.py
```diff
@@ -62,6 +62,16 @@ def on_train_start(self, trainer: Trainer, pl_module: LightningModule) -> None:
             trainer.check_val_every_n_epoch = adaptive_check_val_every_n_epoch
             self._change_early_stopping_patience(trainer.callbacks, adaptive_check_val_every_n_epoch)
             self._change_lr_scheduler_frequency(trainer.lr_scheduler_configs, adaptive_check_val_every_n_epoch)
+        else:
```
vinnamkim: Why is this under `else` branch?

src/otx/algo/callbacks/adaptive_train_scheduling.py
```diff
@@ -62,6 +62,16 @@ def on_train_start(self, trainer: Trainer, pl_module: LightningModule) -> None:
...
+            # To get the consistent result with the OTX1.x, the patience should be changed
+            # Since OTX1.x scheduler works with >= patience, however, OTX2.x scheduler works with > patience
+            for config in trainer.lr_scheduler_configs:
+                if hasattr(config, "scheduler") and hasattr(config.scheduler, "patience"):
+                    msg = (
+                        "The LR scheduler for OTX2 operates when the patience exceeds the set value, "
+                        "hence, the configured value is reduced by 1."
+                    )
+                    config.scheduler.patience -= 1
```
vinnamkim: What is the minimum value for `config.scheduler.patience`? 0 or 1? Please clip it with the minimum.

vinnamkim: Rather than adding this code, is it clear to change the default configurations (`src/otx/recipe`)?

src/otx/algo/callbacks/adaptive_train_scheduling.py
sungmanc: When if condition is satisfied, it already considers the OTX1.x case.
https://github.com/openvinotoolkit/training_extensions/blob/e0d627d8f5eec36aee793e0a426c50a5bd7c5998/src/otx/algo/callbacks/adaptive_train_scheduling.py#L137

src/otx/algo/callbacks/adaptive_train_scheduling.py
sungmanc: I also think that way, however, in that case, logic could be different.
If I get your point, you might mean that changing the value of `config.scheduler.patience=5 --> 4` 

When we consider the `adaptive_interval=2, original_patience=6` at the OTX1.x, 
* OTX1.x behavior : patience = (6/2) = 3 with reduce at  >= patience
* OTX2.x behavior (this PR): (6/2) -1 = 2 with reduce at > patience
* OTX2.x behavior (config change, -1 at the config level): (5/2) -1 = 1 with reduce at > patience

sungmanc: If you have idea, please let me know :) I also prefer to avoid inner pathcing

src/otx/algo/callbacks/adaptive_train_scheduling.py
sungmanc: Fixed the logic as we discussed, https://github.com/openvinotoolkit/training_extensions/pull/3060/commits/827720616db9ddb9b367f574254464abe34eb468

src/otx/algo/callbacks/adaptive_train_scheduling.py
sungmanc: Fixed the logic as we discussed, https://github.com/openvinotoolkit/training_extensions/pull/3060/commits/827720616db9ddb9b367f574254464abe34eb468

jaegukhyun (dismissed):

src/otx/core/model/module/base.py
```diff
@@ -159,6 +159,10 @@ def ensure_list(item: Any) -> list:  # noqa: ANN401
                 lr_scheduler_config["interval"] = scheduler.interval
             if hasattr(scheduler, "monitor"):
                 lr_scheduler_config["monitor"] = scheduler.monitor
+            if hasattr(scheduler, "patience"):
+                # To get the consistent result with the OTX1.x, the patience should be changed
+                # Since OTX1.x scheduler works with >= patience, however, OTX2.x scheduler works with > patience
+                scheduler.patience = max(1, scheduler.patience - 1)
```
vinnamkim: I recommend you to avoid runtime patching as possible unlike OTX 1.x. Why not just updating the model templates themselves?

sungmanc: I just don't want to show the different numbers from the OTX1.x configuration to consider the OTX1.x user. 
They might wonder about the different lr patience values. Do you prefer to change the configuration?  I could change

vinnamkim: I don't want to see anymore the runtime patches like this which are hidden and untraceable if possible.

chuneuny-emily (dismissed):

src/otx/core/model/module/base.py
sungmanc: Done, https://github.com/openvinotoolkit/training_extensions/pull/3060/commits/91fccf906f940caedbe05d4b09defb7f69bbed5d

vinnamkim (approved):

harimkang (approved):

jaegukhyun (approved):

PR #3064, opened 2024-03-11
Detect num_classes information for the H-label classification

sungmanc:
### Summary

This PR introduces,
* Automatically takes the `hlabel_info` from the `Datamodule`  and changes the `num_multiclass_heads` and `num_multilabel_classes` information. So, after this PR, users don't need to manually set `--model.num_multiclass_heads` and `--model.num_multilabel_casses` which is quite harmful to UX.

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/cli/cli.py
+++ b/src/otx/cli/cli.py
@@ -421,6 +421,13 @@ class OTXCLI:
                 )
                 warn(warning_msg, stacklevel=0)
                 model_config.init_args.num_classes = num_classes
+            
+            # Hlabel classification
+            from otx.core.data.dataset.classification import HLabelInfo
+            if isinstance(self.datamodule.label_info, HLabelInfo):
+                hlabel_info = self.datamodule.label_info
+                model_config.init_args.num_multiclass_heads = hlabel_info.num_multiclass_heads
+                model_config.init_args.num_multilabel_classes = hlabel_info.num_multilabel_classes
 
         # Parses the OTXModel separately to update num_classes.
         model_parser = ArgumentParser()
--- a/src/otx/cli/cli.py
+++ b/src/otx/cli/cli.py
@@ -421,9 +421,10 @@ class OTXCLI:
                 )
                 warn(warning_msg, stacklevel=0)
                 model_config.init_args.num_classes = num_classes
-            
+
             # Hlabel classification
             from otx.core.data.dataset.classification import HLabelInfo
+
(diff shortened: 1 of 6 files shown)


harimkang (commented):
One concern is that API users will eventually have to update `num_multiclass_heads` and `num_multilabel_classes` themselves.

Shouldn't we do this somewhere other than the CLI?
For example, I think this should be updated at some point when `lit_module.label_info = datamodule.label_info`.
If the metric is the problem, I think it would be better to create a new one when label_info is updated.

harimkang:
> One concern is that API users will eventually have to update `num_multiclass_heads` and `num_multilabel_classes` themselves.
> 
> Shouldn't we do this somewhere other than the CLI? For example, I think this should be updated at some point when `lit_module.label_info = datamodule.label_info`. If the metric is the problem, I think it would be better to create a new one when label_info is updated.

sungmanc:
> One concern is that API users will eventually have to update `num_multiclass_heads` and `num_multilabel_classes` themselves.
> 
> Shouldn't we do this somewhere other than the CLI? For example, I think this should be updated at some point when `lit_module.label_info = datamodule.label_info`. If the metric is the problem, I think it would be better to create a new one when label_info is updated.

I also think the API side that an user delievers the class instant via API (i.e. Engine.train(model=EfficientNetb0HLabelClassification, metric=MixedHLabelAccuracy)). It's hard to consider when we consider the delivery of class instant since the metric needs that information (num_multiclass_heads) at the init phase. I want to do that at the later phase.

src/otx/cli/cli.py
```diff
@@ -422,6 +422,14 @@ def instantiate_model(self, model_config: Namespace) -> tuple:
                 warn(warning_msg, stacklevel=0)
                 model_config.init_args.num_classes = num_classes
 
+            # Hlabel classification
+            from otx.core.data.dataset.classification import HLabelInfo
```
kprokofi, line 426: Can we move the import higher so as not to mix with the code? Either in this function or at the beginning of the file.

harimkang: Unfortunately, the CLI doesn't assume torch or lightning is installed by default, so this is a good place to be.

jaegukhyun (approved):

sungmanc:
@harimkang , @kprokofi , Thanks for the comments, do I need to fix or add something in this PR ?

harimkang (approved):

eunwoosh (approved):

PR #3094, opened 2024-03-13
Enable training on XPU devices in OTX2.0

kprokofi:
### Checklist

### License

```python
```


src/otx/engine/engine.py
```diff
@@ -804,6 +807,14 @@ def _build_trainer(self, **kwargs) -> None:
         """Instantiate the trainer based on the model parameters."""
         if self._cache.requires_update(**kwargs) or self._trainer is None:
             self._cache.update(**kwargs)
+            # set up xpu device
+            if self._device.accelerator == DeviceType.xpu:
+                self._cache.update(strategy="xpu_single")
+                # add plugin for Automatic Mixed Precision on XPU
+                if kwargs["precision"] == 16:
+                    self._cache.update(plugins=[MixedPrecisionXPUPlugin()])
+                    self._cache.args["precision"] = None
```
harimkang: This probably won't cope with the case where `kwargs["precision"] = None` and `self._cache.args["precision"] = 16`.

src/otx/engine/engine.py
```diff
@@ -904,3 +915,19 @@ def _get_anomaly_model(
...
+            from mmcv.ops.roi_align import RoIAlign
+            from mmengine.structures import instance_data
+
+            from otx.algo.detection.utils import monkey_patched_nms, monkey_patched_roi_align
+
+            long_type_tensor = Union[torch.LongTensor, torch.xpu.LongTensor]
+            bool_type_tensor = Union[torch.BoolTensor, torch.xpu.BoolTensor]
+            instance_data.IndexType = Union[str, slice, int, list, long_type_tensor, bool_type_tensor, np.ndarray]
+            NMSop.forward = monkey_patched_nms
+            RoIAlign.forward = monkey_patched_roi_align
```
harimkang: Is there any reason why this function needs to be a function in Engine? 
It could be declared somewhere like utils and Engine could just call it when it needs it, what do you think?

src/otx/algo/detection/utils/mmcv_patched_ops.py
```diff
@@ -0,0 +1,74 @@
...
+def monkey_patched_nms(
+    ctx: NMSop,
+    bboxes: torch.Tensor,
+    scores: torch.Tensor,
+    iou_threshold: float,
+    offset: float,
+    score_threshold: float,
+    max_num: int,
+) -> torch.Tensor:
+    """Runs MMCVs NMS with torchvision.nms, or forces NMS from MMCV to run on CPU."""
```
jaegukhyun, line 31: It's just a question, can we run NMS through XPU?

src/otx/algo/detection/utils/mmcv_patched_ops.py
jaegukhyun, line 1: Can we add some unit tests for checking these patched module return same results with original ones?

jaegukhyun (commented):
Do we have plan to add integration test for xpu to Github action?

harimkang:
@kprokofi Can you please target this to the [releases/2.0.0](https://github.com/openvinotoolkit/training_extensions/tree/releases/2.0.0) branch?

src/otx/engine/engine.py
```diff
@@ -904,3 +915,19 @@ def _get_anomaly_model(
         model.optimizer = optimizer
         model.scheduler = scheduler
         return model
+
+    def _patch_packages_xpu(self) -> None:
+        """Patch packages when xpu is available."""
+        if self.task in [OTXTaskType.DETECTION, OTXTaskType.INSTANCE_SEGMENTATION] and self.device.accelerator == "xpu":
```
eunwoosh: I think that `ROTATED_DETECTION` and `DETECTION_SEMI_SL` also need to be added.

src/otx/engine/engine.py
```diff
@@ -904,3 +915,19 @@ def _get_anomaly_model(
...
+            import numpy as np
+            from mmcv.ops.nms import NMSop
+            from mmcv.ops.roi_align import RoIAlign
+            from mmengine.structures import instance_data
+
+            from otx.algo.detection.utils import monkey_patched_nms, monkey_patched_roi_align
+
+            long_type_tensor = Union[torch.LongTensor, torch.xpu.LongTensor]
+            bool_type_tensor = Union[torch.BoolTensor, torch.xpu.BoolTensor]
+            instance_data.IndexType = Union[str, slice, int, list, long_type_tensor, bool_type_tensor, np.ndarray]
```
eunwoosh: It's just a question. Why is it needed?

src/otx/engine/engine.py
```diff
@@ -804,6 +807,14 @@ def _build_trainer(self, **kwargs) -> None:
         """Instantiate the trainer based on the model parameters."""
         if self._cache.requires_update(**kwargs) or self._trainer is None:
             self._cache.update(**kwargs)
+            # set up xpu device
+            if self._device.accelerator == DeviceType.xpu:
+                self._cache.update(strategy="xpu_single")
+                # add plugin for Automatic Mixed Precision on XPU
+                if kwargs["precision"] == 16:
+                    self._cache.update(plugins=[MixedPrecisionXPUPlugin()])
```
eunwoosh, line 817: As you know, loss of yolox_tiny model become NaN after some iteration in OTX 1.X, and mixed precision is disabled in OTX 1.X for that reason. Is it fixed in OTX 2.0?

eunwoosh (commented):
Thanks for your effort :) I left some comments. Please take a look.

src/otx/engine/engine.py
kprokofi: Unfortunately, if we won't patch this -> mmengine will raise an error during bbox post processing since XPU device is not supported by MMLab. It didn't happen in OTX1.6 since we used old version of mm* libraries.

src/otx/algo/detection/utils/mmcv_patched_ops.py
kprokofi: MMCV doesn't support XPU device. So, using original nms function isn't possible. It will lead to an error

src/otx/engine/engine.py
kprokofi: Yes, we can

src/otx/algo/detection/utils/mmcv_patched_ops.py
kprokofi: Ok

kprokofi: [Done](https://github.com/openvinotoolkit/training_extensions/pull/3094/commits/7e303a1a9caedfa247c9c98d47e7956d5fee01d0)

src/otx/engine/engine.py
kprokofi: [Done](https://github.com/openvinotoolkit/training_extensions/pull/3094/commits/7e303a1a9caedfa247c9c98d47e7956d5fee01d0)

src/otx/engine/engine.py
kprokofi: [Done](https://github.com/openvinotoolkit/training_extensions/pull/3094/commits/7e303a1a9caedfa247c9c98d47e7956d5fee01d0)

src/otx/engine/engine.py
kprokofi: I haven't faced this issue during benchmarking otx2.0. Yolox_tiny used in Geti works OK with this plugin

jaegukhyun (dismissed):

src/otx/engine/engine.py
```diff
@@ -804,6 +808,14 @@ def _build_trainer(self, **kwargs) -> None:
         """Instantiate the trainer based on the model parameters."""
         if self._cache.requires_update(**kwargs) or self._trainer is None:
             self._cache.update(**kwargs)
+            # set up xpu device
+            if self._device.accelerator == DeviceType.xpu:
+                self._cache.update(strategy="xpu_single")
```
vinnamkim, line 814: Why is this needed? Is there any problem for just giving `accelerator==xpu`?

src/otx/engine/engine.py
```diff
@@ -135,6 +137,8 @@ def __init__(
                 label_info=self._datamodule.label_info if self._datamodule is not None else None,
             )
         )
+        patch_packages_xpu(self.task, self.device.accelerator)
```
vinnamkim: Please move this to `XPUAccelerator.setup_device()` not `Engine`

src/otx/utils/utils.py
```diff
@@ -114,3 +126,28 @@ def remove_matched_files(directory: Path, pattern: str, file_to_leave: Path | No
...
+        from mmcv.ops.roi_align import RoIAlign
+        from mmengine.structures import instance_data
+
+        from otx.algo.detection.utils import monkey_patched_nms, monkey_patched_roi_align
+
+        long_type_tensor = Union[torch.LongTensor, torch.xpu.LongTensor]
+        bool_type_tensor = Union[torch.BoolTensor, torch.xpu.BoolTensor]
+        instance_data.IndexType = Union[str, slice, int, list, long_type_tensor, bool_type_tensor, np.ndarray]
+        NMSop.forward = monkey_patched_nms
+        RoIAlign.forward = monkey_patched_roi_align
```
vinnamkim:
```suggestion
        global _nms_op_forward, _roi_align_forward

        _nms_op_forward = NMSop.forward
        _roi_align_forward = RoIAlign.forward
        NMSop.forward = monkey_patched_nms
        RoIAlign.forward = monkey_patched_roi_align

def revert_packages_xpu():
    NMSop.forward = _nms_op_forward
    RoIAlign.forward = _roi_align_forward
    
class XPUAccelerator
....
    def teardown():
        revert_packages_xpu()
```

src/otx/engine/engine.py
vinnamkim: After looking at `xpu_precision.py` and `xpu_single.py`, I see the situation. However, my gut tells me that `SingleDeviceStrategy` and `MixedPrecisionPlugin` must be device-independent. That means this combination `accelerator=xpu`, `precision=16`, and `strategy=single` should be working. Can we re-register them with patching XPU support rather than creating explicit ones? If it is hard to be done in this PR, please create another one.

docs/source/guide/get_started/installation.rst
```diff
@@ -65,6 +65,68 @@ virtual environment.
...
+.. code-block:: shell
+
+    pip install otx
+
+5. Install requirements for training
+excluding Pytorch and MMCV.
+
+.. code-block:: shell
+
+    otx install -v --do-not-install-torch --do-no-install-mmcv
```
eunwoosh:
```suggestion
    otx install -v --do-not-install-torch --do-not-install-mmcv
```

eunwoosh: It seems that if passing `--do-no-install-mmcv`, otx doesn't install all mmXX frameworks including mmpretrain, mmdet and etc. Could you check it? In my case, installation w/o `otx install -v --do-not-install-torch` even works well.

src/otx/engine/engine.py
eunwoosh: image
It seems that yolox_tiny still have a problem. How about disabling mixed precision training w/ XPU?

kprokofi: What dataset do you use? Is it from tests/assest? Do you propose to remove it for all models?

docs/source/guide/get_started/installation.rst
kprokofi: Yesterday when I tried to install following my instructions - everything was installed properly. Installing mmcv from mim leads to the same error with mmcv.ops as we have in OTX1.6
I will check about other mmXX packages

src/otx/engine/engine.py
kprokofi: image

kprokofi: YOLOX cat train well on all 3 regression dataset in my case

kprokofi: I will check the speed difference between FP16 and FP32 on XPU. If there is no difference - let's remove it at all

docs/source/guide/get_started/installation.rst
eunwoosh:
```
    mmcv_install_args = []
    if mmcv_requirements and not do_not_install_mmcv:
        mmcv_install_args = get_mmcv_install_args(torch_requirement, mmcv_requirements)
        install_args += ["openmim"]

    # Install requirements.
    with console.status("[bold green]Working on installation...\n") as status:
        if verbose:
            logger.setLevel(logging.INFO)
            status.stop()
        console.log(f"Installation list: [yellow]{install_args}[/yellow]")
        status_code = create_command("install").main(install_args)
        if status_code == 0:
            console.log(f"Installation Complete: {install_args}")

        # https://github.com/Madoshakalaka/pipenv-setup/issues/101
        os.environ["SETUPTOOLS_USE_DISTUTILS"] = "stdlib"

        # Install mmX requirements if the task requires mmX packages using mim.
        if mmcv_install_args and status_code == 0:
            console.log(f"Installation list: [yellow]{mmcv_install_args}[/yellow]")
            status_code = mim_installation(mmcv_install_args)
            if status_code == 0:
                console.log(f"MMLab Installation Complete: {mmcv_install_args}")
```
This is "src/otx/cli/install.py". It seems that if `do_not_install_mmcv` is True, `mmcv_install_args` will be `[]`and then `if mmcv_install_args and status_code == 0` can't be exceuted.

src/otx/utils/utils.py
kprokofi: [Done](https://github.com/openvinotoolkit/training_extensions/pull/3094/commits/8fb52df736f8c03245d64994bbadd54debab8007). I moved it to accelerator to avoid working with global variables

src/otx/engine/engine.py
kprokofi: Patching strategy seems to be overwhelming. I tried to do that, it leads to errors and make code less readable in my opinion. Registering new strategy is ordinary process for new device. In lightning they have new strategy for each device.
Also, please, be aware of this draft (for a while) PR from lightning: https://github.com/Lightning-AI/pytorch-lightning/pull/17700

When they decide to actually add XPU as new device, we can remove everything and use XPU natively. Let's keep explicit strategy registering for now. I don't see any problems with that, it is how lightning works with new devices.

docs/source/guide/get_started/installation.rst
kprokofi: Actually, you are right. Also, I noticed, that if correct version of mmcv is installed, it is not collected during parsing "mmcv_install_args ". So, let's just remove "do_not_install_mmcv". I can confirm that OTX2.0 can be installed w/o problems omitting this step

harimkang (dismissed):

kprokofi:
I updated strategy: https://github.com/openvinotoolkit/training_extensions/pull/3094/commits/98e0b69e2903fa1d4ce799c1ee5480416f52b0ee removing  torch.xpu.optimize for Semantic Segmentation since it was reported by IPEX that they have a bug there in that case and to be able to train segmentation model we should remove optimization. 
I hope it will be fixed in the next IPEX releases

kprokofi:
@harimkang , seems like a problem with installing mmdet occurs during set up env for unit tests. Could you take a look?

jaegukhyun (approved):

eunwoosh (approved):

src/otx/engine/engine.py
vinnamkim: > Registering new strategy is ordinary process for new device. In lightning they have new strategy for each device.

I visited the link you gave https://github.com/Lightning-AI/pytorch-lightning/pull/17700. However, it could not see the new implementation about the strategy. On the other hand, there are modifications for the existing strategy.
image

> When they decide to actually add XPU as new device, we can remove everything and use XPU natively. Let's keep explicit strategy registering for now.

1. I don't know whether the upstream situation is resolved near future. This is because https://github.com/Lightning-AI/pytorch-lightning/pull/17700 was opened almost a year ago and is not merged even today.
2. Even if the upstream PR is merged, I can expect that we still need some customizations for us like https://github.com/openvinotoolkit/training_extensions/blob/98e0b69e2903fa1d4ce799c1ee5480416f52b0ee/src/otx/algo/detection/utils/mmcv_patched_ops.py
It means the re-registering will exist in the future as well.

> Patching strategy seems to be overwhelming. I tried to do that, it leads to errors and make code less readable in my opinion.

I don't know why it is hard actually. I think it can be

```python
from lightning.pytorch.strategies.single_device import SingleDeviceStrategy as _SingleDeviceStrategy

class SingleDeviceStrategy(_SingleDeviceStrategy):
    ....

    def setup_optimizers(self, trainer: pl.Trainer) -> None:
        if self.root_device.type == "xpu":
            return self._xpu_setup_optimizer(trainer)
        return super().setup_optimizers(trainer)

    def _xpu_setup_optimizers(self, trainer: pl.Trainer) -> None:
        """Sets up optimizers."""
        super().setup_optimizers(trainer)
        if len(self.optimizers) != 1:  # type: ignore[has-type]
            msg = "XPU strategy doesn't support multiple optimizers"
            raise RuntimeError(msg)
        if trainer.task != "SEMANTIC_SEGMENTATION":
            model, optimizer = torch.xpu.optimize(trainer.model, optimizer=self.optimizers[0])  # type: ignore[has-type]
            self.optimizers = [optimizer]
            self.model = model
```
I can see nothing that can be error-prone.

tests/unit/algo/plugins/test_plugins.py
```diff
@@ -0,0 +1,56 @@
...
+"""Test for otx.algo.plugins.xpu_precision"""
+
+
+import pytest
+import torch
+from otx.algo.plugins.xpu_precision import MixedPrecisionXPUPlugin
+from torch.optim import Optimizer
+
+
+class TestMixedPrecisionXPUPlugin:
```
vinnamkim, line 13: Are the new tests you added in this PR needed to run on the XPU devices? If so, please add pytest marker, e.g., xpu, to 
https://github.com/openvinotoolkit/training_extensions/blob/176cdf501f409d537bb12d1055a56bb0a3376eb6/pyproject.toml#L333-L338

Then, please add the pytest markers
```suggestion
@pytest.mark.xpu
class TestMixedPrecisionXPUPlugin:
```

We don't have CI pipeline for XPU but adding the pytest markers is done before that.

harimkang:
> @harimkang , seems like a problem with installing mmdet occurs during set up env for unit tests. Could you take a look?

There was a brief issue with the CI network. I rerun the test and the install is fine.

tests/unit/algo/plugins/test_plugins.py
kprokofi: it is device agnostic. I patch torch.xpu, so it can be also executed on CUDA.

eugene123tw (approved):

src/otx/engine/engine.py
kprokofi: Let's consider it in the next PR

tests/unit/algo/plugins/test_plugins.py
kprokofi: I will mark it as resolved. Please, add more comments if you want me to add smth in the different PR

PR #3139, opened 2024-03-15
Fix wrong ov datamodule update in predict, test, optimize

harimkang:
### Summary

Fix https://jira.devtools.intel.com/browse/CVS-135468

Found that changing the OV test pipeline was causing a behavior that changed not only the transforms, but also other data-related configs, so fix around it.

### Checklist

### License

```python
```


jaegukhyun (dismissed):

src/otx/engine/engine.py
```diff
@@ -533,7 +533,7 @@ def optimize(
 
         model = self.model
         if not isinstance(model, OVModel):
-            datamodule = self._auto_configurator.get_ov_datamodule()
+            optimize_datamodule = self._auto_configurator.update_ov_test_pipeline(datamodule=optimize_datamodule)
```
vinnamkim: It is basically not allowed to use a test set for model optimization. Anyway, why do we have to create a new datamodule even if we already has `self.datamodule`?

harimkang: This is the code added by https://github.com/openvinotoolkit/training_extensions/pull/3028 in that PR. 
@sovrasov Could you please answer this question?

harimkang: I removed this line to update the `datamodule` in optimize. - https://github.com/openvinotoolkit/training_extensions/pull/3139/commits/edadbb6434ecfe51b0d17c3c901f098b09817589

vinnamkim: > Anyway, why do we have to create a new datamodule even if we already has self.datamodule?

This is not just for the optimize. Why do we need these recreations for others?

harimkang: I don't know the details, but there was a request from Vlad to update the test pipeline use the pipeline in `openvino_model.yaml` when using `OVModel`. - https://github.com/openvinotoolkit/training_extensions/pull/2897#issuecomment-1931862841

vinnamkim: I see the reason. Can you update this to update the train pipeline with the OV data pipeline such as

```python
# not update_ov_test_pipeline()
optimize_datamodule = self._auto_configurator.update_ov_train_pipeline(datamodule=optimize_datamodule)
```

harimkang: I modified the `update_ov_test_pipeline` function to `update_ov_subset_pipeline` to allow it to be updated by specifying a subset. - https://github.com/openvinotoolkit/training_extensions/pull/3139/commits/13ac06895c230a7c02312ae757d62b95b68c906b

jaegukhyun (approved):

vinnamkim (approved):

PR #3142, opened 2024-03-15
Enhancing Mask Post-Processing with Customized RTMDetInst Head

eugene123tw:
### Summary
This PR improves mask post-processing for training and OV inference by introducing Customized RTMDetInst head. Improvements include optimizing CUDA memory usage and integrating Customized ONNX NMS static tracing. Additionally, it addresses issues related to RandomResize scope and reinstates Mosaic augmentation. Test coverage for the customized RTMDetInst head is also added. Experiment results are available [here](https://intel-my.sharepoint.com/:x:/p/eugene_liu/ETPbekz1zjBLre3DT9l7MrcBZbNrufb1tumJXZCRVNu8gw?e=RNbtHg&nav=MTVfezc0MzQzN0ZCLTBCMDMtNDFBRC05REE4LUQzMDUzNDRGOEQ4MX0).

### Checklist

### License

```python
```


src/otx/core/data/dataset/tile.py
```diff
@@ -76,6 +76,10 @@ def __init__(
             threshold_drop_ann=threshold_drop_ann,
         )
         self._tile_size = tile_size
+        # TODO (Eugene): Bug found in original Datumaro tile polygon function.
+        # https://github.com/eugene123tw/training_extensions/tree/eugene/fix-tile-polygon-func
+        # It lacks polygon validation, potentially leading to GeometryCollection or MultiPolygon results,
+        # which the current function doesn't handle.
```
vinnamkim, line 82: What's this?

src/otx/core/data/dataset/instance_segmentation.py
```diff
@@ -56,6 +56,8 @@ def _get_item_impl(self, index: int) -> InstanceSegDataEntity | None:
 
         # convert xywh to xyxy format
         bboxes = np.array(gt_bboxes, dtype=np.float32)
+        if len(bboxes) == 0:
+            bboxes = np.empty((0, 4))
```
vinnamkim:
```suggestion
        bboxes = np.array(gt_bboxes, dtype=np.float32) if gt_bboxes else np.empty((0, 4))
```

src/otx/core/model/entity/base.py
```diff
@@ -410,6 +413,8 @@ def _callback(result: NamedTuple, idx: int) -> None:
                     self.model.await_any()
                 self.model.infer_async(im, user_data=idx)
             self.model.await_all()
+            print("FPS:", num_samples / (time.time() - start_time))
+            print("Time per sample:", (time.time() - start_time) / num_samples)
```
vinnamkim: Is this needed?

src/otx/core/model/module/instance_segmentation.py
```diff
@@ -169,11 +169,13 @@ def _convert_pred_entity_to_compute_metric(
             inputs.polygons,
             inputs.labels,
         ):
-            rles = (
-                [encode_rle(mask) for mask in masks.data]
-                if len(masks)
-                else polygon_to_rle(polygons, *imgs_info.ori_shape)
-            )
+            rles = []
+            if len(bboxes):
```
vinnamkim: Why is this check needed?

src/otx/recipe/instance_segmentation/openvino_model.yaml
```diff
@@ -35,4 +35,4 @@ overrides:
       image_color_channel: RGB
       data_format: coco_instances
       test_subset:
-        batch_size: 2
+        batch_size: 256
```
jaegukhyun: Isn't it too large for test case?

src/otx/algo/instance_segmentation/heads/custom_rtmdet_ins_head.py
```diff
@@ -0,0 +1,364 @@
...
+from mmdet.structures.bbox import get_box_tensor, get_box_wh, scale_boxes
+from mmengine.config import ConfigDict
+
+if TYPE_CHECKING:
+    from mmengine.structures import InstanceData
+
+
+@MODELS.register_module()
+class CustomRTMDetInsSepBNHead(RTMDetInsSepBNHead):
+    """Custom RTMDet instance segmentation head."""
```
jaegukhyun: Could you add why do we need custom head and what is difference between custom head and origin one?

harimkang:
Please modify the target with this to the releases/2.0.0 branch.

src/otx/core/model/module/instance_segmentation.py
eugene123tw: To prevent unexpected errors, inputs from val set should be checked for images that lack annotations.

src/otx/core/model/entity/base.py
eugene123tw: In OTX 1.x, we log the FPS from the OV inferencer. This log is particularly valuable for Geti scenarios.

src/otx/core/model/module/instance_segmentation.py
vinnamkim: What do you exactly mean? What is the exact requirement for the training step? Is `len(bboxes) == len(rle)`? If the malformed input is given, we should stop and raise an exception, not continue rolling the wheel by introducing if-branching for the exceptional case.

src/otx/core/model/entity/base.py
vinnamkim: If so, I recommend you to use https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.profilers.SimpleProfiler.html#lightning.pytorch.profilers.SimpleProfiler in this case as you know, otherwise, pass the FPS metric values to PyTorch Trainer's logger, not adding `print()` here. You might enhance https://github.com/openvinotoolkit/training_extensions/blob/develop/src/otx/algo/callbacks/iteration_timer.py to gather the FPS metric additionally.

vinnamkim: In addition, OTX 2 inference pipeline will not be used in Geti.

src/otx/core/model/module/instance_segmentation.py
eugene123tw: No, it has nothing to do with malformed or rolling the wheel. It's exactly as I mentioned: some images in the validation set lack annotations because we don't apply the same filtering as we do during training. If you believe we should include pre-filtering for val set, we can include it in the pre-filtering process (`src/otx/core/data/pre_filtering.py`). If not, we should anticipate that inputs.bboxes will yield bboxes = [], therefore bypassing RLE postprocessing.

src/otx/core/model/entity/base.py
eugene123tw: SimpleProfiler is overly verbose in these situations, making it less useful. This is likely why we haven't made it the default option and instead opted to add a specific line of https://github.com/openvinotoolkit/training_extensions/blob/releases/2.0.0/src/otx/cli/__init__.py#L19

```py
def main() -> None:
    """Entry point for OTX CLI.

    This function is a single entry point for all OTX CLI related operations:
    """
    start = time()
    OTXCLI()
    dt = timedelta(seconds=time() - start)
    print(f"Elapsed time: {dt}")
```

The underlying concept is quite similar.

Plus, the impact of the iteration time on printing to stdout is minimal, especially considering the large batch size specified in openvino_model.yaml.

However, if you're not a fan of it, I can remove it.

src/otx/algo/instance_segmentation/heads/custom_rtmdet_ins_head.py
eugene123tw: Done. Thanks!

src/otx/recipe/instance_segmentation/openvino_model.yaml
eugene123tw: I reduced it to 64 the same as in detection.

src/otx/core/data/dataset/tile.py
eugene123tw: I've encountered an issue in `_tile_polygon` (https://github.com/openvinotoolkit/datumaro/blob/develop/src/datumaro/plugins/tiling/tile.py#L64) while running some tiling experiments. My plan is to address this problem first by creating a PR in OTX 2.x to fix it, and then report back to the datumaro repo.

The issue is from the line `inter = polygon.intersection(roi_box)`, which doesn't consistently return a sg.Polygon object; instead, it sometimes returns a sg.GeometryCollection or sg.MultiPolygon object. Additionally, the function doesn't validate the polygon, and from the experiment, it could a self-intersecting polygon, which would result in an error when computing the Shapely intersection.

vinnamkim: I jumped into this link https://github.com/eugene123tw/training_extensions/tree/eugene/fix-tile-polygon-func and found there was a patch. I was asking the reason why there is just a comment in this PR and not having the actual patch that exists in the link.

src/otx/core/model/entity/base.py
vinnamkim: I'm cautioning against the additive manner in which new implementation is being added without consideration for existing ones. Functionally, batch iteration time and sample throughput are quite similar. Therefore, instead of increasing code entropy, utilize existing functionality or extend it if possible. The current code structure ensures that throughput is logged only for OVModel, excluding OTXModel. Why does OTXModel lack this throughput reporting, and why is this code inserted only in OVModel? If you were to leave the maintenance in the future or a new person unfamiliar with the history were to join, it would be a struggling bar to understand this.

src/otx/core/model/module/instance_segmentation.py
vinnamkim: I read your comment twice but I cannot still understand it. Is there any case that `bboxes = []` but `len(masks) > 0` or `len(polygons)>0`?. To me, `bboxes` and `masks or polygons` are independent entities each other. However, you injected a constraint that `bboxes == [] => rle = []` From this statement and my conjecture, I asked you whether there is a strict constraint that `len(bboxes) == len(masks)` or `len(bboxes) == len(polygons)`.

src/otx/core/data/dataset/tile.py
eugene123tw: As it's not necessarily related to this PR, I intend to create another PR for it.

src/otx/core/model/module/instance_segmentation.py
eugene123tw: The requirement of `len(bboxes) == len(masks)` or `len(bboxes) == len(polygons)` is necessary. And, I understand how it might lead to confusion. So I've decided to remove this check and manage it within the polygon to RLE conversion function.

src/otx/core/model/entity/base.py
eugene123tw: That's a fair point. I'm removing it and exploring a more holistic approach. The rationale behind not adding it to OTXModel is that Geti exclusively runs inference/predictions on OV models, making it a valuable metric. However, if Geti won't utilise OTX 2.x for inference or prediction then it's not serving much of a purpose.

vinnamkim (approved):

harimkang (approved):

Issue #3154, opened 2024-03-19
No evaluation metric outputs for Anomaly exported & optimized models


goodsong81:
Train -> Test gives following output:
![image](https://github.com/openvinotoolkit/training_extensions/assets/4629606/5193ff23-6659-486c-8c7f-592b4ffafbb6)

But, Export or Optimize -> Test gives no evaluation metrics:
![image](https://github.com/openvinotoolkit/training_extensions/assets/4629606/3ed74957-fbff-41a5-84c6-0bf66a8f5d7b)
And even seems like testing wrong data subset; 237 images, not 8.

**How to reproduce**
On the latest develop (v2.0)
(Any datasets would reproduce the above results)
```console
otx train --config src/otx/recipe/anomaly_classification/stfpm.yaml --data_root /mnt/sdc/otx_v2_dataset/anomaly/mvtec/wood_medium --work_dir /tmp/pytest-of-songkich/pytest-7/otx-benchmark0/20240319-105856/anomaly_classification/stfpm/mvtec_wood_medium/0 --model.num_classes 2 --data.config.data_format mvtec --engine.device gpu --seed 0 --deterministic False
otx test --work_dir /tmp/pytest-of-songkich/pytest-7/otx-benchmark0/20240319-105856/anomaly_classification/stfpm/mvtec_wood_medium/0
otx export --work_dir /tmp/pytest-of-songkich/pytest-7/otx-benchmark0/20240319-105856/anomaly_classification/stfpm/mvtec_wood_medium/0
otx test --config src/otx/recipe/anomaly_classification/openvino_model.yaml --checkpoint /tmp/pytest-of-songkich/pytest-7/otx-benchmark0/20240319-105856/anomaly_classification/stfpm/mvtec_wood_medium/0/.latest/export/exported_model.xml --work_dir /tmp/pytest-of-songkich/pytest-7/otx-benchmark0/20240319-105856/anomaly_classification/stfpm/mvtec_wood_medium/0
```

harimkang:
@ashwinvaidya17 Can you take a look at this? As I understand it, the metrics on the `AnomalyModule` side are internal and not configured separately in OTX.

goodsong81:
@ashwinvaidya17 Could you check the right metrics for each task?
Currently, Train -> Test outputs
* Anomaly classification
  - image_F1Score/AUROC
* Anomaly detection
  - image_F1Score/AUROC
  - pixel_F1Score/AUROC
[...]

PR #3163, opened 2024-03-19
🐞 Add metrics to openvino model

ashwinvaidya17:
### Summary

- Addresses https://github.com/openvinotoolkit/training_extensions/issues/3154
- I am a bit unclear on metric computation for OpenVINO models. Currently I am using Anomalib's metrics. It in turn relies on TorchMetrics.
- The DataLoader issue is probably unrelated. Also, I don't expect the results to be correct as the CLI does not run validation to select meaningful thresholds.
image

### Checklist

### License

```python
```


jaegukhyun:
Could you add unit tests for this update?.
And I think it is natural to use torch metrics for openvino model, so we can assure that torch model and openvino model show same performance if they return same output on same input.
What is your main concern about using torch metrics for openvino model?

ashwinvaidya17:
> Could you add unit tests for this update?. And I think it is natural to use torch metrics for openvino model, so we can assure that torch model and openvino model show same performance if they return same output on same input. What is your main concern about using torch metrics for openvino model?

Alright! I'll add some tests.
I was just unsure about the approach for openvino model. I thought since we don't use torch tensors in openvino TorchMetrics was not the right approach. But thanks for the clarification.

harimkang:
Please modify the target with this to the releases/2.0.0 branch.

jaegukhyun (dismissed):

harimkang:
I reapplied pre-commit in my environment. - https://github.com/openvinotoolkit/training_extensions/pull/3163/commits/f30e55176d334e5e03ce358805ce84a1bb594628

harimkang (approved):

jaegukhyun (approved):

PR #3180, opened 2024-03-21
Refactor label info

vinnamkim:
### Summary
This ticket (136313) about refactoring LabelInfo in OTX 2.0. The scope is

1. Move LabelInfo from `otx/core/data/dataset.py` to the dedicated source file at `otx/core/types/label.py`
2. Prevent label_info from being over-written for OVModel. It will be serialized into a JSON string which is stored in OpenVINO IR's rt_info fields, and deserialized when OVModel is initialized.
3. Remove runtime patching for HLabel head and the HLabelInfo object can be propagated at the model initialization

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/core/types/label.py
+++ b/src/otx/core/types/label.py
@@ -13,7 +13,7 @@ from typing import TYPE_CHECKING, Any
 if TYPE_CHECKING:
     from datumaro import Label, LabelCategories
 
-__all__ = ["LabelInfo", "HLabelInfo"]
+__all__ = ["LabelInfo", "HLabelInfo", "SegLabelInfo", "NullLabelInfo"]
 
 
 @dataclass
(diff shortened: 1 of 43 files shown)


src/otx/core/types/label.py
samet-akcay, line 1: What is the motivation of moving label to types? 

As a user, I would expect to import it somewhere from data. 
either, `from otx.data import label` or `from otx.core.data import label`. When using the API, it might be hard to remember that `label` is in `types`

vinnamkim: To me, it is more neutral position between data and model. This is because the data can produce `label_info` but the model takes `label_info` at its initialization. In addition, this is a kind of concept existing in OTX and `otx.core.types` is reserved for it.

I think you are worrying about accessibility for them. https://github.com/openvinotoolkit/training_extensions/pull/3180/commits/ba7e9b18968896f9ba66dfb733183b8df1e84439 improved it.

harimkang (approved):

sungmanc (approved):

PR #3155, opened 2024-03-19
Promote OTXModel to PyTorchLightningModule and deprecate OTXLitModule

vinnamkim:
### Summary

 - Ticket no. 134585

### Checklist

- [ ] I have added unit tests to cover my changes.​
- [ ] I have added integration tests to cover my changes.​
- [ ] I have added e2e tests for validation.
- [ ] I have added the description of my changes into CHANGELOG in my target branch (e.g., [CHANGELOG](https://github.com/openvinotoolkit/training_extensions/blob/develop/CHANGELOG.md) in develop).​
- [ ] I have updated the documentation in my target branch accordingly (e.g., [documentation](https://github.com/openvinotoolkit/training_extensions/tree/develop/docs) in develop).
- [ ] I have [linked related issues](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

### License

- [ ] I submit _my code changes_ under the same [Apache License](https://github.com/openvinotoolkit/training_extensions/blob/develop/LICENSE) that covers the project.
      Feel free to contact the maintainers if that's a concern.
- [ ] I have updated the license header for each file (see an example below).

```python
# Copyright (C) 2024 Intel Corporation
# SPDX-License-Identifier: Apache-2.0
```

PR #3187, opened 2024-03-22
Fix F1 metric

vinnamkim:
### Summary

- This PR is suggested by this comment https://github.com/openvinotoolkit/training_extensions/pull/3155#issuecomment-2011271864
- ~~It can be reviewed after merging https://github.com/openvinotoolkit/training_extensions/pull/3180~~
- Rework to correct best_confidence_threshold acquisition from F1 measure and different F1 measure evaluation for validation and test.
- However, I still do not know it is correctly working. This is because there are two problems on 1) parsing config file and 2) uncertainty on F1 measure implementation itself.
- Fix OTXCLI to return error code if an exception is raised

### Checklist

### License

```python
```


src/otx/core/metrics/fmeasure.py
```diff
@@ -680,8 +689,13 @@ def update(self, preds: list[dict[str, Tensor]], target: list[dict[str, Tensor]]
...
             )
 
-    def compute(self) -> dict:
-        """Compute f1 score metric."""
+    def compute(self, best_confidence_threshold: float | None = None) -> dict:
+        """Compute f1 score metric.
+
+        best_confidence_threshold (float | None): Pre-defined best confidence threshold.
+            If this value is None, then FMeasure will find best confidence threshold and store it as member variable.
+            Defaults to None.
```
harimkang:
```suggestion
        Args:
            best_confidence_threshold (float | None): Pre-defined best confidence threshold.
                If this value is None, then FMeasure will find best confidence threshold and store it as member variable.
                Defaults to None.
```

src/otx/core/model/base.py
```diff
@@ -290,8 +291,11 @@ def _convert_pred_entity_to_compute_metric(
         """Convert given inputs to a Python dictionary for the metric computation."""
         raise NotImplementedError
 
-    def _log_metrics(self, meter: Metric, key: str) -> None:
-        results: dict[str, Tensor] = meter.compute()
+    def _log_metrics(self, meter: Metric, key: Literal["val", "test"], **compute_kwargs) -> None:
+        sig = inspect.signature(meter.compute)
+        filtered_kwargs = {key: value for key, value in compute_kwargs.items() if key in sig.parameters}
```
harimkang, line 296: How about outputting information about which keys are being filtered?

src/otx/engine/engine.py
```diff
@@ -55,6 +55,24 @@
 }
 
 
+@contextmanager
+def override_metric_callable(model: OTXModel, new_metric_callable: MetricCallable | None) -> Iterator[OTXModel]:
```
harimkang, line 59: Are there any cases where this function is used outside of `Engine`? If so, I think this function would fit in the other place (as `core/utils`..?).

If it's just for Engine, you can ignore this comment.

src/otx/engine/engine.py
```diff
@@ -263,19 +280,28 @@ def train(
...
+                **fit_kwargs,
+            )
         self.checkpoint = self.trainer.checkpoint_callback.best_model_path
+
+        if not isinstance(self.checkpoint, (Path, str)):
+            msg = "self.checkpoint should be Path or str at this time."
+            raise TypeError(msg)
+
+        best_checkpoint_symlink = Path(self.work_dir) / "best_checkpoint.ckpt"
+        best_checkpoint_symlink.symlink_to(self.checkpoint)
```
harimkang, line 296: In this case, we have an existing ckpt output in `work_dir` called `checkpoints/epoch~.ckpt`,
and `best_checkpoint.ckpt` in our `work_dir`?
If it's redundant, I think we can get rid of one, but what do you think?

tests/regression/test_regression.py
```diff
@@ -84,19 +90,40 @@ def _test_regression(
...
+
+                run_main(command_cfg=train_cfg, open_subprocess=True)
+                checkpoint = test_case.output_dir / ".latest" / "train" / "best_checkpoint.ckpt"
+                assert checkpoint.exists()
+
+                test_cfg = command_cfg.copy()
+                test_cfg[1] = "test"
+                test_cfg += ["--checkpoint", str(checkpoint)]
+
+                # TODO(harimkang): This command cannot create `metrics.csv`` file under test output directory
```
harimkang, line 102: Let me check. Thanks!

src/otx/core/metrics/fmeasure.py
```diff
@@ -643,22 +643,31 @@ class FMeasure(Metric):
...
-        self.targets: list[list[tuple]] = []
-        self.num_classes: int = num_classes
+        self.label_info: LabelInfo = label_info
 
         self._f_measure_per_confidence: dict | None = None
         self._f_measure_per_nms: dict | None = None
-        self._best_confidence_threshold: float | None = best_confidence_threshold
+        self._best_confidence_threshold: float | None = None
         self._best_nms_threshold: float | None = None
+        self._f_measure = 0.0
```
jaegukhyun, line 659: I think we can't set pre-defined confidence threshold during training phase. In Geti, user can specify confidence threshold they want, and it is expected to validate using that confidence threshold during training.

harimkang (commented):
LGTM, 
@jaegukhyun We need an update regarding the confidence threshold in OTX 2.1 for Geti integration. Please take a look at this PR in this regard.

sungmanc (dismissed):

src/otx/core/metrics/fmeasure.py
vinnamkim: I think that there was not clear interface about what you mentioned too before this PR. If you think it is necessary, add it to the constructor. But, I don't think that it is necessary.

src/otx/engine/engine.py
vinnamkim: It is dedicated to the engine

src/otx/engine/engine.py
vinnamkim: This is a symlink so that there is no additional storage cost. To me, it seems better to have a fixed file path for the best checkpoint in the `work_dir`.

vinnamkim: Added in 81743a0

harimkang (approved):

jaegukhyun (approved):

sungmanc (approved):

src/otx/core/model/base.py
vinnamkim: Added in 81743a0

PR #3193, opened 2024-03-25
Update visual prompting refactoring to develop

sungchul1:
### Summary

This PR includes:
- [x] Updates from `release/2.0` branch
- [x] Update tests
- [x] Fix ruff errors (ticket no. 135852)

### Checklist

### License

```python
```


harimkang (approved):

src/otx/algo/visual_prompting/zero_shot_segment_anything.py
```diff
@@ -729,15 +753,15 @@ def _customize_outputs(  # type: ignore[override]
...
+        masks: list[Mask] = []
         prompts: list[Points] = []
         scores: list[Tensor] = []
         labels: list[LongTensor] = []
         for predicted_masks, used_points in outputs:
             for label, predicted_mask in predicted_masks.items():
                 if len(predicted_mask) == 0:
                     continue
-                masks.append(tv_tensors.Mask(torch.stack(predicted_mask, dim=0), dtype=torch.float32))
+                masks.append(Mask(torch.stack(predicted_mask, dim=0), dtype=torch.float32))
```
vinnamkim, line 764: Is this binary mask, or not? If it is a binary mask, is there any reason that it should have fp32 precision?

sungchul1: It is wrapped with similarity scores, so it is needed to specify data type.
But it seems legacy because it was to use soft prediction for iou calculation in 1.x.
https://github.com/openvinotoolkit/training_extensions/pull/3193/files/a8f1083eeefa9b340a4890959ef3efe4c20e9d97#diff-ca889a1c6b848bafab6f551e09cceb26aa296935debfaba99a7ff18fb51b5b9bR391

vinnamkim: I cannot understand what you meant exactly. Do you mean that "It is a soft mask (non-binary) for now, but we can change it to the hard mask (binary)"?

From the link you gave, it seems that an scalar value is multiplied with the hard mask
https://github.com/openvinotoolkit/training_extensions/blob/a8f1083eeefa9b340a4890959ef3efe4c20e9d97/src/otx/algo/visual_prompting/zero_shot_segment_anything.py#L391
It seems that we don't need a full precision for each pixel to represent that mask.

sungchul1: Sorry for confusing due to link's order.
In 1.x, soft prediction to get blurred hard prediction was used in https://github.com/openvinotoolkit/training_extensions/blob/0e1de1dbb6635d4cf42a2f99d1b529661e34a897/src/otx/algorithms/visual_prompting/adapters/pytorch_lightning/callbacks/inference.py#L121-L123.
But 2.x doesn't use this function so it seems that we don't need that part anymore.
I think it doesn't affect to performance, but it needs performance check before and after.
I can't afford to do it so I'll revisit this check.

vinnamkim (approved):

Issue #3166, opened 2024-03-20
OTX 2.0 fails to evaluate the exported models trained with tiling


goodsong81:
**Describe the bug**

```colsole
Traceback (most recent call last):
  File "/home/songkich/miniconda3/envs/otx-v2/bin/otx", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/mnt/sdb/workarea/otx/src/otx/cli/__init__.py", line 17, in main
    OTXCLI()
  File "/mnt/sdb/workarea/otx/src/otx/cli/cli.py", line 56, in __init__
    self.run()
  File "/mnt/sdb/workarea/otx/src/otx/cli/cli.py", line 570, in run
    self.instantiate_classes()
  File "/mnt/sdb/workarea/otx/src/otx/cli/cli.py", line 354, in instantiate_classes
    self.model, self.optimizer, self.scheduler = self.instantiate_model(model_config=model_config)
                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/sdb/workarea/otx/src/otx/cli/cli.py", line 443, in instantiate_model
    raise AttributeError(msg)
AttributeError: The model does not have a tile_config attribute. Please check if the model supports tiling.
```

**Steps to Reproduce**

```colsole
otx train --config src/otx/recipe/instance_segmentation/maskrcnn_efficientnetb2b_tile.yaml --data_root tests/assets/car_tree_bug/ --work_dir /tmp/tile --metric otx.core.metrics.fmeasure.FMeasure --callback_monitor val/f1-score --scheduler.monitor val/f1-score --seed 0 --deterministic False --max_epochs 1
otx export --work_dir /tmp/tile/
otx test --work_dir /tmp/tile/ --config src/otx/recipe/instance_segmentation/openvino_model.yaml --checkpoint /tmp/tile/.latest/export/exported_model.xml
```

harimkang:
@eugene123tw Probably `tile_config.enable_tile` is on if we are using `openvino_model.yaml`.
We want `enable_tile` to be off in this case, right? 

Anyway,
In this case, we can use `otx test --work_dir /tmp/tile/ --checkpoint /tmp/tile/.latest/export/exported_model.xml`

But I think the example above should work fine. Can you take a look at the above case to make it work?

eugene123tw:
[...]

PR #3204, opened 2024-03-25
Fix tile CLI

eugene123tw:
### Summary

https://github.com/openvinotoolkit/training_extensions/issues/3166

### Checklist

### License

```python
```


harimkang:
@eugene123tw  I think this should be added to the 2.0.0 branch as well, right?

src/otx/engine/utils/auto_configurator.py
```diff
@@ -347,6 +347,7 @@ def update_ov_subset_pipeline(self, datamodule: OTXDataModule, subset: str = "te
         data_configuration = datamodule.config
         ov_test_config = self._load_default_config(model_name="openvino_model")["data"]["config"][f"{subset}_subset"]
         subset_config = getattr(data_configuration, f"{subset}_subset")
+        subset_config.batch_size = ov_test_config["batch_size"]
```
harimkang, line 367: Does `batch_size` also need to be updated here? and if so, could you include the batch size update in the warning message at the bottom of that code?

src/otx/engine/engine.py
```diff
@@ -340,7 +340,7 @@ def test(
         datamodule = datamodule if datamodule is not None else self.datamodule
 
         is_ir_ckpt = Path(str(checkpoint)).suffix in [".xml", ".onnx"]
-        if is_ir_ckpt and not isinstance(model, OVModel):
+        if is_ir_ckpt:
```
harimkang: The only case where `OVModel` is already used is when using `openvino_model.yaml` as config file, as far as I know. In this case, I don't think we need to update the `datamodule` pipeline and `model`. Is there any reason why this syntax should disappear?

eugene123tw: @harimkang  `datamodule` is set with `OTXTileTestDataset` following `otx test --work_dir /tmp/tile/ --config src/otx/recipe/instance_segmentation/openvino_model.yaml --checkpoint /tmp/tile/.latest/export/exported_model.xml`

As Model API has tiling API, there's no need to use OTXTileTestDataset, thus, we should reset the datamodule if model IR is specified. Tbh, I don't quite understand why there's this condition, and I think it should be changed to `is_ir_ckpt and isinstance(model, OVModel)`

harimkang: This is for the natural flow of the engine. After engine.train and export, in order to use the testing ir model, we need to create a new engine by creating an OVModel. This statement eliminates the need to create a new engine by allowing OVModel to be built and used for each task when the checkpoint is xml.

eugene123tw: Thanks for the explanation, I made a new update.

jaegukhyun (dismissed):

src/otx/engine/utils/auto_configurator.py
eugene123tw: Done!

harimkang (changes requested):
Please reflect the changes in this PR. https://github.com/openvinotoolkit/training_extensions/pull/3229

harimkang (approved):

jaegukhyun (approved):

PR #3212, opened 2024-03-27
Add unit test for export for develop branch

eunwoosh:
### Summary
It's same as #3211 but for develop branch.
This PR adds unit test for export.
Additional unit test covering left part is included in another PR.

### Checklist

### License

```python
```


tests/test_helpers.py
```diff
@@ -40,3 +46,122 @@ def generate_random_bboxes(
     areas = (x_max - x_min) * (y_max - y_min)
     bboxes = np.column_stack((x_min, y_min, x_max, y_max))
     return bboxes[areas > 0]
+
+
+@contextlib.contextmanager
+def generate_random_image_folder(width: int = 480, height: int = 360, number_of_images: int = 10) -> Iterator[str]:
```
vinnamkim: Why not making these functions (`generate_*`) as pytest fixtures?

eunwoosh: This part is copied from OTX 1.X which is also copied from Geti side. Do you think it's better to change it as fixture?

vinnamkim: Sure, it's common practice. In addition, you can control the fixture's lifecycle. For example, in this case, you can generate a random image folder once in a session and can reuse it for many test functions. https://stackoverflow.com/a/45833359

tests/unit/core/exporter/exportable_code/conftest.py
```diff
@@ -0,0 +1,17 @@
...
+
+from otx.core.exporter.exportable_code import demo
+
+# exportable_code is standalone package and files in exportable_code import 'demo_package' which is same as
+# 'otx.core.exporter.exportable_code.demo.demo_package'. To avoid import error while running test, need to
+# import 'otx.core.exporter.exportable_code.demo.demo_package' and register it as 'demo_package'.
+demo_package_file = Path(demo.__file__).parent / "demo_package" / "__init__.py"
+spec = importlib.util.spec_from_file_location("demo_package", demo_package_file)
+module = importlib.util.module_from_spec(spec)
+sys.modules["demo_package"] = module
```
vinnamkim: Would you attach an error log if we don't have this modification as comments here as well?

tests/test_helpers.py
eunwoosh: ok I'll change it.

tests/unit/core/exporter/exportable_code/conftest.py
eunwoosh: do you mean attaching error log at comment?

vinnamkim: In addition, this should be a fixture to make it restrictive for the tests under `tests/unit/core/exporter/exportable_code`

```python
@pytest.fixture(scope="module", autouse=True)
def fxt_fix_exportable_code_import_error():
    # exportable_code is standalone package and files in exportable_code import 'demo_package' which is same as
    # 'otx.core.exporter.exportable_code.demo.demo_package'. To avoid import error while running test, need to
    # import 'otx.core.exporter.exportable_code.demo.demo_package' and register it as 'demo_package'.
    demo_package_file = Path(demo.__file__).parent / "demo_package" / "__init__.py"
    spec = importlib.util.spec_from_file_location("demo_package", demo_package_file)
    module = importlib.util.module_from_spec(spec)

    tmp_mod_demo_package = sys.modules.get("demo_package", None)
    sys.modules["demo_package"] = module

    yield

    # Revert sys.modules["demo_package"] modification
    if tmp_mod_demo_package:
        sys.modules["demo_package"] = tmp_mod_demo_package
    else:
        sys.modules.pop("demo_package")
```

eunwoosh: Thanks for detailed explanation. I also agree with your opinion but it seems that fixture is executed after import, which makes error.
image

vinnamkim: > do you mean attaching error log at comment?

Yes

vinnamkim: > Thanks for detailed explanation. I also agree with your opinion but it seems that fixture is executed after import, which makes error. image

You can make move this import from the source top level to the function level to be done at runtime
`from otx.core.exporter.exportable_code.demo.demo_package.executors.asynchronous import AsyncExecutor`

tests/unit/core/exporter/exportable_code/demo/demo_package/executors/test_asynchronous.py
```diff
@@ -0,0 +1,79 @@
...
+        executor = AsyncExecutor(mock_model, mock_visualizer)
+        executor.run(MagicMock())
+
+        mock_render_result.assert_called()
+        for i in range(1, 4):
+            assert mock_render_result.call_args_list[i - 1].args == (i,)
+        mock_visualizer.show.assert_called()
+        for i in range(1, 4):
+            assert mock_visualizer.show.call_args_list[i - 1].args == (i,)
+        mock_dump_frames.assert_called()
```
vinnamkim, line 81: You are testing `AsyncExecutor` and it consists of three components: 1) `ModelWrapper`, 2) `BaseVisualizer`, and 3) `AsyncPipeline`. In this case, we basically mock those three components to test the target object. But, you mocked the two object, but rather than mocking `AsyncPipeline`, trying to mock the member function of the target test object (`AsyncExecutor.render_result()`). It is not common practice. The common practice will be
```suggestion
    def test_run(self, mocker, mock_model, mock_visualizer, mock_streamer, mock_dump_frames):
        mock_async_pipeline = mocker.patch("otx.core.exporter.exportable_code.demo.demo_package.executors.asynchronous.AsyncPipeline").return_value
        executor = AsyncExecutor(mock_model, mock_visualizer)
        executor.run(MagicMock())

        ...
```

tests/unit/core/exporter/exportable_code/conftest.py
eunwoosh: done https://github.com/openvinotoolkit/training_extensions/pull/3212/commits/cfc87b25f3ad44dfd07350cc8c7727c79a6eb298

tests/test_helpers.py
eunwoosh: done https://github.com/openvinotoolkit/training_extensions/pull/3212/commits/e7325279302ee9b77272373c33be053a29148855

tests/unit/core/exporter/exportable_code/demo/demo_package/executors/test_asynchronous.py
eunwoosh: I also mock `AsyncPipeline` and `render_result` as you said. I think purpose of unit test is to check target function works as expected. So, checking how `render_result` works is out of scope of those test function. I also implement unit test for `render_result` which can cover checking `render_result`, I think.

harimkang (approved):

tests/unit/core/exporter/exportable_code/demo/demo_package/executors/test_asynchronous.py
vinnamkim: I am still feeling the idea of ​​mocking an object's "public" function in order to test itself very strange. It deviates from the standards of software development. Thus, it makes other developers hard to understand it.

We are not just copy pasting something from OTX 1.x to OTX 2. We are doing this to achieve real improvement.

This time I'll approved this PR, However, please think about it next time.

vinnamkim (approved):

PR #3035, opened 2024-03-05
Fix multiple lr scheduler (warmup scheduler) & Add adaptive_patience to the lr scheduler

sungmanc:
### Summary

This PR introduces,
* Resolving the multi LR scheduler issue by overriding the `optimizer_step()` to enable the warmup scheduling. So, warmup logic will not be enabled by `LinearWarmupScheduler` anymore.
* Add adaptive_patience to the LRScheduler
* Add unit-tests for warmup logic

TODOs,
* Need to resolve the hard-coded index for the optimizer, and scheduler (i.e. optimizers[0]).
* For future work, we might use the callback ? to enable the warm-up scheduler, need to check the feasibility. If this way works, we could remove the `warmup_steps` and `warmup_by_epochs` at the base `OTXLitModule`.
NOTE,
* I remained the lr_scheduler_configs as list type since I got an error when I changed the type to the dictionary. I think it could be handled at the next phase since it is not important for this PR.

### Checklist

- [ ] I have added unit tests to cover my changes.​
- [ ] I have added integration tests to cover my changes.​
- [ ] I have added e2e tests for validation.
- [ ] I have added the description of my changes into CHANGELOG in my target branch (e.g., [CHANGELOG](https://github.com/openvinotoolkit/training_extensions/blob/develop/CHANGELOG.md) in develop).​
- [ ] I have updated the documentation in my target branch accordingly (e.g., [documentation](https://github.com/openvinotoolkit/training_extensions/tree/develop/docs) in develop).
- [ ] I have [linked related issues](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

### License

- [ ] I submit _my code changes_ under the same [Apache License](https://github.com/openvinotoolkit/training_extensions/blob/develop/LICENSE) that covers the project.
      Feel free to contact the maintainers if that's a concern.
- [ ] I have updated the license header for each file (see an example below).

```python
# Copyright (C) 2023 Intel Corporation
# SPDX-License-Identifier: Apache-2.0
```

PR #3216, opened 2024-03-27
Refactor optimizer and lr scheduler part

vinnamkim:
### Summary
 - Ticket no. 135933
 - This PR aims to remove the custom code in `src/otx/cli/cli.py` which is hardly coupled with `OTXModel`.
 - To do this, change type `optimizer` and `scheduler` typings from

    ```python
    optimizer: list[OptimizerCallable] | OptimizerCallable = DefaultOptimizerCallable,
    scheduler: list[LRSchedulerCallable] | LRSchedulerCallable = DefaultSchedulerCallable,
    ```

    to

    ```python
    optimizer: OptimizerCallable = DefaultOptimizerCallable,
    scheduler: LRSchedulerCallable | LRSchedulerListCallable = DefaultSchedulerCallable,
    ```

    This is to remove the `list[OptimizerCallable]` and `list[LRSchedulerCallable]` typings. The reason for this is that I faced this issue, https://github.com/omni-us/jsonargparse/issues/478, after removing `add_list_type_arguments()` from `src/otx/cli/cli.py`.

 - According to this type change, it is required to implement `LinearWarmupSchedulerCallable` newly to incorporate the linear warmup scheduler with the other main scheduler.
 - `LinearWarmupSchedulerCallable` and `OTXModel.lr_scheduler_step()` will solve this problem, https://github.com/openvinotoolkit/training_extensions/pull/3035#discussion_r1515429701

### Checklist

### License

```python
```


src/otx/algo/classification/mmconfigs/multiclass_classification/dino_v2.yaml
```diff
@@ -0,0 +1,6 @@
+backbone:
+  name: dinov2_vits14_reg
+  frozen: false
+head:
+  in_channels: 384
+  num_classes: 1000
```
harimkang, line 6: Wouldn't it be better to have this as an argument to the model? DINO doesn't actually use mm, so it seems a bit odd to have configs moved to mmconfigs.

src/otx/algo/anomaly/openvino_model.py
```diff
@@ -12,17 +13,17 @@
 
 from typing import TYPE_CHECKING, Any
 
-from lightning.pytorch import LightningModule
-
+from otx.core.metrics import NullMetricCallable
+from otx.core.metrics.types import MetricCallable
```
harimkang:
```suggestion
from otx.core.metrics.types import MetricCallable, NullMetricCallable
```

src/otx/engine/utils/auto_configurator.py
```diff
@@ -256,41 +254,60 @@ def get_model(self, model_name: str | None = None, label_info: LabelInfo | None
...
+            model_config["init_args"]["hlabel_info"] = label_info
+        elif label_info is None and self.task == OTXTaskType.H_LABEL_CLS:
+            msg = "You should explicitly give label_info for `OTXTaskType.H_LABEL_CLS` task."
+            raise ValueError(msg)
 
-        logger.warning(f"Set Default Model: {self.config['model']}")
-        return instantiate_class(args=(), init=self.config["model"])
+        return model_parser.instantiate_classes(Namespace(model=model_config)).get("model")
 
     def get_optimizer(self) -> list[OptimizerCallable] | None:
```
harimkang, line 278: This seems to be deprecated in `Engine`, do we still need this function?
The same goes for `get_scheduler`.

src/otx/algo/classification/mmconfigs/multiclass_classification/dino_v2.yaml
vinnamkim: It is currently implemented as MM-style and it must be revisited in the future. The reason for this change is because I encountered annoying test errors where `jsonargparse` kept failing to parse something in `DictConfig`. What you mentioned will be amortized in the future revisit for this model to disassemble the MM-style.

src/otx/engine/utils/auto_configurator.py
vinnamkim: If you really think it as what you said, you can remove this. However, I'll keep this because it is a big sized PR, so that I'd like to avoid any potential test error by removing it.

src/otx/core/metrics/fmeasure.py
```diff
@@ -644,6 +644,7 @@ class FMeasure(Metric):
     def __init__(
         self,
         label_info: LabelInfo,
+        *,
```
jaegukhyun, line 647: Why is this needed?

jaegukhyun (approved):

src/otx/engine/utils/auto_configurator.py
harimkang: Okay, I see, this is something I'll look into and remove later.

harimkang (approved):

src/otx/core/metrics/fmeasure.py
vinnamkim: I'd like to make sure that only `label_info` is an unique positional argument: https://realpython.com/python-asterisk-and-slash-special-parameters/#in-short-the-python-asterisk-and-slash-control-how-to-pass-values-to-functions

It does not affect to the runtime behavior.

PR #3217, opened 2024-03-27
Enable train with fixed confidence threshold f1 score metric

jaegukhyun:
### Checklist

### License

```python
```

### DIFF
--- a/src/otx/core/model/detection.py
+++ b/src/otx/core/model/detection.py
@@ -52,6 +52,7 @@ class OTXDetectionModel(
         scheduler: list[LRSchedulerCallable] | LRSchedulerCallable = DefaultSchedulerCallable,
         metric: MetricCallable = MeanAPCallable,
         torch_compile: bool = False,
+        confidence_threshold: float | None = None,
     ) -> None:
         super().__init__(
             num_classes=num_classes,
@@ -61,6 +62,7 @@ class OTXDetectionModel(
             torch_compile=torch_compile,
         )
         self.tile_config = TileConfig()
+        self.confidence_threshold: float | None = confidence_threshold
 
     def forward_tiles(self, inputs: TileBatchDetDataEntity) -> DetBatchPredEntity | DetBatchPredEntityWithXAI:
         """Unpack detection tiles.
@@ -163,24 +165,23 @@ class OTXDetectionModel(
 
     def _log_metrics(self, meter: Metric, key: Literal["val", "test"], **compute_kwargs) -> None:
         if key == "val":
-            retval = super()._log_metrics(meter, key)
-
-            # NOTE: Validation metric logging can update `best_confidence_threshold`
-            if best_confidence_threshold := getattr(meter, "best_confidence_threshold", None):
-                self.hparams["best_confidence_threshold"] = best_confidence_threshold
+            # NOTE: Validation metric logging can update `best_confidence_threshold`,
+            # if there's no pre-defined best confidence threshold
+            if self.confidence_threshold is None:
+                retval = super()._log_metrics(meter, key)
+                if best_confidence_threshold := getattr(meter, "best_confidence_threshold", None):
+                    self.hparams["best_confidence_threshold"] = best_confidence_threshold
+            else:
+                retval = super()._log_metrics(meter, key, best_confidence_threshold=self.confidence_threshold)
+                self.hparams["best_confidence_threshold"] = self.confidence_threshold
 
             return retval
 
-        if key == "test":
-            # NOTE: Test metric logging should use `best_confidence_threshold` found previously.
-            best_confidence_threshold = self.hparams.get("best_confidence_threshold", None)
-            compute_kwargs = (
-                {"best_confidence_threshold": best_confidence_threshold} if best_confidence_threshold else {}
-            )
+        # NOTE: Test metric logging should use `best_confidence_threshold` found previously.
+        best_confidence_threshold = self.hparams.get("best_confidence_threshold", None)
+        compute_kwargs = {"best_confidence_threshold": best_confidence_threshold} if best_confidence_threshold else {}
 
-            return super()._log_metrics(meter, key, **compute_kwargs)
-
-        raise ValueError(key)
+        return super()._log_metrics(meter, key, **compute_kwargs)
 
 
 class ExplainableOTXDetModel(OTXDetectionModel):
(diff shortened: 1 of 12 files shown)


vinnamkim (commented):
Why this change is needed? I doubt that there exists a real use-case to setup a confidence threshold by hands during train/val.

jaegukhyun:
> Why this change is needed? I doubt that there exists a real use-case to setup a confidence threshold by hands during train/val.

Unfortunately, Geti has parameters for this.

vinnamkim:
> > Why this change is needed? I doubt that there exists a real use-case to setup a confidence threshold by hands during train/val.
> 
> Unfortunately, Geti has parameters for this.

I think that it should be deprecated.

jaegukhyun:
> > > Why this change is needed? I doubt that there exists a real use-case to setup a confidence threshold by hands during train/val.
> > 
> > 
> > Unfortunately, Geti has parameters for this.
> 
> I think that it should be deprecated.

I'm not sure about this. Could you share your thought with other stakeholders?

vinnamkim:
> > > > Why this change is needed? I doubt that there exists a real use-case to setup a confidence threshold by hands during train/val.
> > > 
> > > 
> > > Unfortunately, Geti has parameters for this.
> > 
> > 
> > I think that it should be deprecated.
> 
> I'm not sure about this. Could you share your thought with other stakeholders?

I reported this issue to Wonju.

jaegukhyun:
> > > > > Why this change is needed? I doubt that there exists a real use-case to setup a confidence threshold by hands during train/val.
> > > > 
> > > > 
> > > > Unfortunately, Geti has parameters for this.
> > > 
> > > 
> > > I think that it should be deprecated.
> > 
> > 
> > I'm not sure about this. Could you share your thought with other stakeholders?
> 
> I reported this issue to Wonju.

Any conclusion?

vinnamkim:
> Any conclusion?

Further discussion with Geti stakeholders is needed

jaegukhyun:
> > Any conclusion?
> 
> Further discussion with Geti stakeholders is needed

Then how about adding this for now. At this moment, we need three postprocessing parameters confidence threshold, use ellipse shapes, max num detections. Now I'm considering add this as a postprocessing parameters for det, iseg model. If it is done, we can easily delete unnecessary post processing parameters

src/otx/core/model/detection.py
```diff
@@ -61,6 +62,7 @@ def __init__(
             torch_compile=torch_compile,
         )
         self.tile_config = TileConfig()
+        self.confidence_threshold: float | None = confidence_threshold
```
vinnamkim, line 65: Is this already captured by `self.hparams`? Even if not, please store this in `self.hparams` as the same place as `best_confidence_threshold`.

src/otx/core/model/detection.py
```diff
@@ -163,24 +165,23 @@ def load_state_dict(self, ckpt: dict[str, Any], *args, **kwargs) -> None:
...
-                self.hparams["best_confidence_threshold"] = best_confidence_threshold
+            # NOTE: Validation metric logging can update `best_confidence_threshold`,
+            # if there's no pre-defined best confidence threshold
+            if self.confidence_threshold is None:
+                retval = super()._log_metrics(meter, key)
+                if best_confidence_threshold := getattr(meter, "best_confidence_threshold", None):
+                    self.hparams["best_confidence_threshold"] = best_confidence_threshold
+            else:
+                retval = super()._log_metrics(meter, key, best_confidence_threshold=self.confidence_threshold)
+                self.hparams["best_confidence_threshold"] = self.confidence_threshold
```
vinnamkim, line 176: According to the naming change,
```suggestion
            # if there's no fixed confidence threshold
            if fixed_confidence_threshold := self.hparam.get("fixed_confidence_threshold", None):
                retval = super()._log_metrics(meter, key, best_confidence_threshold=fixed_confidence_threshold)
                self.hparams["best_confidence_threshold"] = fixed_confidence_threshold
            else:
                retval = super()._log_metrics(meter, key)
                self.hparams["best_confidence_threshold"] = getattr(meter, "best_confidence_threshold", None)
```

src/otx/core/model/detection.py
vinnamkim: Can we rename this to `fixed_confidence_threshold` to make it clear?
```suggestion
        self.fixed_confidence_threshold: float | None = fixed_confidence_threshold
```

vinnamkim (commented):
> Then how about adding this for now. At this moment, we need three postprocessing parameters confidence threshold, use ellipse shapes, max num detections. Now I'm considering add this as a postprocessing parameters for det, iseg model. If it is done, we can easily delete unnecessary post processing parameters

Sure. However, I left some comments. Please see

src/otx/core/model/detection.py
jaegukhyun: This logic will save best_confidence_threshold and fixed_confidence_threshold filed, and the value of both filed would be same if users specify their own threshold. Is it right?

vinnamkim: Hmm, I thought that you wanted to make them same. If not they are equal to each other, we cannot do the test with the fixed confidence threshold because your test metric logic is
https://github.com/openvinotoolkit/training_extensions/blob/01a07f6ecd16db153f5c4d6cd81717f253325ab9/src/otx/core/model/detection.py#L180-L184

Under this implementation, we have to set `best_confidence_threshold = fixed_confidence_threshold`.

Yea, personally, I prefer the two values ​​to be independent each other. If so, we need to change this part too. `fixed_confidence_threshold` should be get prioritized than `best_confidence_threshold` when it exists during the test phase.

jaegukhyun: How about fix hparam's field name as confidence threshold, if user specify fixed confidence threshold then confidence threshold would be fixed confidence threshold if not, it will be best confidence threshold from metric. And test phase will  only take confidence threshold from hparam or xml file

wonjuleee:
> > > Any conclusion?
> > 
> > 
> > Further discussion with Geti stakeholders is needed
> 
> Then how about adding this for now. At this moment, we need three postprocessing parameters confidence threshold, use ellipse shapes, max num detections. Now I'm considering add this as a postprocessing parameters for det, iseg model. If it is done, we can easily delete unnecessary post processing parameters

Hi @jaegukhyun, thank you for your efforts. After discussion with Radwan, it would be good for holding at this moment. He said, it is a wrong behavior (deterministic confidence threshold during training) and should be applicable only for testing/inference. This will be also fixed in Geti side too.

jaegukhyun:
This will be revisited after Geti's direction of this feature is determined.

Issue #3219, opened 2024-03-27
OTX2.0 Anomaly tasks show 0 accuracy on mvtec datasets


goodsong81:
**Describe the bug**
![image](https://github.com/openvinotoolkit/training_extensions/assets/4629606/f16b4e4d-bc84-47ed-9560-d10342fa457a)

**Steps to Reproduce**

On the latest releases/2.0.0 branch
```bash
otx train --config src/otx/recipe/anomaly_classification/padim.yaml --data_root /mnt/sdc/val-data/v1/anomaly/mvtec/wood_medium/ --work_dir /tmp/ano
otx test --work_dir /tmp/ano
```
Same goes for stfpm and other mvtec datasets.

ashwinvaidya17:
Yeah it is a known issue. This is due to the fact that the default dataset configured from the CLI does not contain validation set. So the thresholds are not assigned correctly. Ideally OTX should support custom dataloaders.

I've documented this in the config files
https://github.com/openvinotoolkit/training_extensions/blob/2277ec5fa01330a1279ebb07be742c29722369b2/src/otx/recipe/anomaly_classification/padim.yaml#L19

If we use this https://github.com/openvinotoolkit/training_extensions/blob/develop/src/otx/data/anomaly/anomaly.py from the API then it gives good numbers

harimkang:
> Yeah it is a known issue. This is due to the fact that the default dataset configured from the CLI does not contain validation set. So the thresholds are not assigned correctly. Ideally OTX should support custom dataloaders.
> 
> I've documented this in the config files
> 
> https://github.com/openvinotoolkit/training_extensions/blob/2277ec5fa01330a1279ebb07be742c29722369b2/src/otx/recipe/anomaly_classification/padim.yaml#L19
> 
> If we use this https://github.com/openvinotoolkit/training_extensions/blob/develop/src/otx/data/anomaly/anomaly.py from the API then it gives good numbers

@ashwinvaidya17 Is there any workaround for this?

harimkang:
https://github.com/openvinotoolkit/training_extensions/pull/3231 Merged into 2.0

PR #3231, opened 2024-03-28
Use anomaly datamodule in CLI

ashwinvaidya17:
### Summary

Addresses: https://github.com/openvinotoolkit/training_extensions/issues/3219

```bash
otx train --config src/otx/recipe/anomaly_classification/padim.yaml --data_root ~/datasets/MVTec/bottle --work_dir otx-workspace
otx test --config src/otx/recipe/anomaly_classification/padim.yaml --data_root ~/datasets/MVTec/bottle --work_dir otx-workspace 
```

### Padim
image

### Stfpm
image

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/recipe/anomaly_classification/draem.yaml
+++ b/src/otx/recipe/anomaly_classification/draem.yaml
@@ -34,7 +34,7 @@ overrides:
         transforms:
           - class_path: torchvision.transforms.v2.Resize
             init_args:
-              size: 256
+              size: [256, 256]
               antialias: True
           - class_path: torchvision.transforms.v2.ToDtype
             init_args:
@@ -50,7 +50,7 @@ overrides:
         transforms:
           - class_path: torchvision.transforms.v2.Resize
             init_args:
-              size: 256
+              size: [256, 256]
               antialias: True
           - class_path: torchvision.transforms.v2.ToDtype
             init_args:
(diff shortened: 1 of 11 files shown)


src/otx/recipe/anomaly_classification/draem.yaml
```diff
@@ -31,11 +32,10 @@ overrides:
         batch_size: 8
         num_workers: 4
         transforms:
-          - class_path: otx.core.data.transform_libs.torchvision.ResizetoLongestEdge
```
samet-akcay, line 34: what is the difference between the native torchvision transform and `otx.core.data.transform_libs.torchvision`. Ideally transforms can be imported directly from `torchvision.transforms.v2.Resize`? @harimkang any thoughts?

@ashwinvaidya17, this way of transform configuration should ideally be in anomalib too

samet-akcay (dismissed):
Thanks for addressing this @ashwinvaidya17. Let's prioritise the discussion of custom datamodule support.

src/otx/recipe/anomaly_classification/draem.yaml
harimkang: Um, looking at it, this looks like a transform for the visual_prompting model, we can use `torchvision.transforms.v2.Resize`

harimkang:
It looks like the test is still not passing, please take a look at it, and maybe we shouldn't put this commit in 2.1 and just leave it as a workaround for 2.0, because I think we should fix it the right way in 2.1. It would be great if you could work on develop when you have time.

harimkang (approved):

jaegukhyun (approved):

PR #3242, opened 2024-03-29
Refactor XAI model part - classification tasks

vinnamkim:
### Summary

- Ticket no. 135930
- The main objective of this refactoring is to break this inheritance
```text
Before:
OTXModel -> ExplainableOTXClsModel-> OTXMulticlassClsModel, OTXMultilabelClsModel, OTXHlabelClsModel

After:
OTXModel -> OTXMulticlassClsModel, OTXMultilabelClsModel, OTXHlabelClsModel
```

- This is because `ExplainableOTXClsModel` has many common functions but those are mostly used for the `MMPretrain` models. Even these common functions actually harm the freedom of implementation of models other than `MMPretrain`, such as `OTXTVModel`. It is not good for the OTX 2.1 objective: replace the old MM style models to own implemented ones. Therefore, I demoted it into a mixin class in `src/otx/core/model/utils/mmpretrain.py`.
- In addition, it also does not fit the spirit of having this interface, `OTXModel.forward_explain()`. All we needed to obtain XAI predictions are implementing just this function `OTXModel.forward_explain()` in the derived class such as
https://github.com/openvinotoolkit/training_extensions/blob/99b1de22102bda57e76635631a1ee1751e414dd8/src/otx/algo/classification/torchvision_model.py#L310-L322
- However, `OTXModel.forward_explain()` is not perfectly shaped for now because it requires helps from those two functions:
https://github.com/openvinotoolkit/training_extensions/blob/99b1de22102bda57e76635631a1ee1751e414dd8/src/otx/algo/classification/torchvision_model.py#L324-L332
This is because there are some problems in the model exportation. It will be revisited afterward.

### Checklist

### License

```python
```


src/otx/algo/classification/torchvision_model.py
```diff
@@ -161,8 +169,39 @@ def forward(
...
+    def _forward_explain(self, images: torch.Tensor) -> dict[str, torch.Tensor | list[torch.Tensor]]:
+        x = self.backbone(images)
+        backbone_feat = x
+
+        saliency_map = self.explainer.func(backbone_feat)
+
+        if len(x.shape) == 4 and not self.use_layer_norm_2d:
+            x = x.view(x.size(0), -1)
+
+        feature_vector = x
```
GalyaZalesskaya, line 186: Feature vector is calculated as 2D pooling of backbone output, as here:
```
from otx.algo.hooks.recording_forward_hook import feature_vector_fn

feature_vector = feature_vector_fn(x)
```

GalyaZalesskaya: I see that this block is just moved from the other part of the file, but apparently it was an issue with code. We don't have checks for feature vector values really (just ndims check), that's why we couldn't catch that feature vector is calculated this way.

src/otx/algo/utils/xai_utils.py
```diff
@@ -2,58 +2,83 @@
 # SPDX-License-Identifier: Apache-2.0
 """Utils used for XAI."""
 
+# TODO(someone): Typings in this file is too weak or wrong. It should be fixed.
```
GalyaZalesskaya:
```suggestion
# TODO(gzalessk): Typings in this file is too weak or wrong. It should be fixed.
```

GalyaZalesskaya: Thank you for this comment, I'll fix it

src/otx/algo/utils/xai_utils.py
```diff
@@ -2,58 +2,83 @@
...
+# For example, `pred_labels: list | None` has no object typing containered in the list.
+# On the other hand, process_saliency_maps should not produce list of dictionaries
+# (`list[dict[str, np.ndarray | torch.Tensor]]`).
+# This is because the output will be assigned to OTXBatchPredEntity.saliency_map,
+# but this has `list[np.ndarray | torch.Tensor]` typing, so that it makes a mismatch.
```
GalyaZalesskaya, line 10: Dict entity is needed not to lose the saliency map class index after selecting a subset of saliency map with selected class 
`{class_idx: class_saliency_map}`

I'll need to look into that

harimkang (dismissed):

src/otx/algo/utils/xai_utils.py
```diff
@@ -67,40 +92,43 @@ def process_saliency_maps(
...
-        msg = "Shape mismatch."
-        raise ValueError(msg)
-
     processed_saliency_maps = []
-    for maps_per_image in saliency_maps:
+    for maps_per_image in saliency_map:
+        if maps_per_image.ndim != 3:
+            msg = "Shape mismatch."
+            raise ValueError(msg)
+
```
GalyaZalesskaya, line 102: There's no need to move this check in `for` loop since all saliency maps should be the same size based on the method of its generation. But your option looks prettier.

src/otx/core/data/entity/base.py
```diff
@@ -669,25 +669,29 @@ class OTXBatchPredEntity(OTXBatchDataEntity):
...
 
     # (Optional) XAI-related outputs
-    saliency_maps: list[np.ndarray] | list[Tensor] = field(default_factory=list)
-    feature_vectors: list[np.ndarray] | list[Tensor] = field(default_factory=list)
+    # TODO(vinnamkim): These are actually plural, but their namings are not
+    # This is because ModelAPI requires the OV IR to produce `saliency_map`
+    # and `feature_vector` (singular) named outputs.
+    # It should be fixed later.
+    saliency_map: list[np.ndarray] | list[Tensor] = field(default_factory=list)
+    feature_vector: list[np.ndarray] | list[Tensor] = field(default_factory=list)
```
GalyaZalesskaya, line 686: Do you want to make updates in Model API and then rename `saliency_map -> saliency_maps` back to plural?

src/otx/core/model/utils/mmpretrain.py
```diff
@@ -48,3 +54,132 @@ def create_model(config: DictConfig, load_from: str | None = None) -> tuple[nn.M
     """
     classification_layers = get_classification_layers(config, MODELS, "model.")
     return build_mm_model(config, MODELS, load_from), classification_layers
+
+
+class ForwardExplainMixInForMMPretrain(Generic[T_OTXBatchPredEntity, T_OTXBatchDataEntity]):
```
GalyaZalesskaya: @negvet I wonder if we can come up with better naming, because it seems not really clear to me. How about `ExplainableMMPretrainModel`?

GalyaZalesskaya (commented):
Thank you for this huge update and good comments. 
@negvet can you look at this PR? I think you can provide a meaningful input here.

src/otx/algo/classification/torchvision_model.py
vinnamkim: image
This model has a wrong implementation losing spatial resolution in a saliency map. I'm about creating a ticket about this. What you mentioned also can be dealt in that ticket.

vinnamkim: Do you mind assigning it to you?

src/otx/core/model/utils/mmpretrain.py
vinnamkim: This is a mixin class that cannot be working standalone. Therefore, I think that it should include `MixIn` keyword in its name such as `ExplainableMixInMMPretrain`. I'm open to a new naming but hope `MixIn` will be in it.

src/otx/core/data/entity/base.py
vinnamkim: Yes, `feature_vector` would be too.

src/otx/algo/utils/xai_utils.py
vinnamkim: The problem is that the output goes into `OTXBatchPredEntity.saliency_map` which has `list[np.ndarray | torch.Tensor]` typing. My concern is that mistyping in runtime should not exist. If you think that `saliency_map` should have this type, `{class_idx: class_saliency_map}`. You can choose to modify `OTXBatchPredEntity`.

src/otx/algo/utils/xai_utils.py
vinnamkim: The former one produces an error not understandable easily if `saliency_maps` is an empty list.

src/otx/core/model/utils/mmpretrain.py
```diff
@@ -48,3 +54,132 @@ def create_model(config: DictConfig, load_from: str | None = None) -> tuple[nn.M
...
+    """OTX classification model which can attach a XAI hook."""
```
vinnamkim:
```suggestion
    """Mix-in class to support XAI feature for MM Pretrained models commonly.

    This is a mix-in class that cannot be used as standalone. The correct usage of it will be like
    ```python
    class MobileNetV3ForMulticlassCls(ForwardExplainMixInForMMPretrain, MMPretrainMulticlassClsModel):
    ```
    """
```

src/otx/core/model/utils/mmpretrain.py
GalyaZalesskaya: Thank you for this clarification :) How about `ExplainableMixInMMPretrainModel`?

src/otx/algo/classification/torchvision_model.py
GalyaZalesskaya: No, I don't mind, thank you
P.S. Thumbs up for the golden theme in VS code

src/otx/algo/utils/xai_utils.py
GalyaZalesskaya: Yes, this makes sense

src/otx/algo/utils/xai_utils.py
GalyaZalesskaya: Makes sense

harimkang (dismissed):

src/otx/core/model/utils/mmpretrain.py
negvet: `ExplainableMixInMMPretrainModel` sounds good to me

src/otx/algo/classification/torchvision_model.py
```diff
@@ -137,6 +139,12 @@ def __init__(
         self.softmax = nn.Softmax(dim=-1)
         self.loss = loss
 
+        self.explainer = ReciproCAMHook(
```
negvet, line 142: What will happen if the user run some unsupported model, e.g. one of `swin` models?

src/otx/core/data/entity/base.py
```diff
@@ -669,25 +669,29 @@ class OTXBatchPredEntity(OTXBatchDataEntity):
...
+        feature_vector: List of intermediate feature vectors used for model predictions.
             This field is optional and will be an empty list for non-XAI pipelines.
     """
 
     scores: list[np.ndarray] | list[Tensor]
 
     # (Optional) XAI-related outputs
-    saliency_maps: list[np.ndarray] | list[Tensor] = field(default_factory=list)
-    feature_vectors: list[np.ndarray] | list[Tensor] = field(default_factory=list)
+    # TODO(vinnamkim): These are actually plural, but their namings are not
```
negvet, line 681: `These are actually plural, but their namings are not`, this is what might lead to some minor confusion. I appreciate unification that you bring. Another good alternative was to instead update Model API to `saliency_maps` and `feature_vectors`.

GalyaZalesskaya: Yes, as I understood from Vinnam's comment, the plan is to update Model API and then revert these changes `saliency_map`->`saliency_maps`, `feature_vector` -> `feature_vectors`

vinnamkim: Yea I hope that this will be fixed in ModelAPI side first.

src/otx/algo/classification/torchvision_model.py
vinnamkim: What is the exact definition of the user you said? If the user is a kind of model developer who implements his own model, all we expect from him are implementing `OTXModel.forward_explain()` in the derived class. `OTXModel.forward_explain()` should fulfill `saliency_map` and `feature_vector` fields. If possible, it might exploit a helper component like `ReciproCAMHook` in it.

> - In addition, it also does not fit the spirit of having this interface, `OTXModel.forward_explain()`. All we needed to obtain XAI predictions are implementing just this function `OTXModel.forward_explain()` in the derived class such as
https://github.com/openvinotoolkit/training_extensions/blob/99b1de22102bda57e76635631a1ee1751e414dd8/src/otx/algo/classification/torchvision_model.py#L310-L322

negvet: I mean will some exception will raise when someone would run `swin` model with `explain_mode`? `ReciproCAM` does not support some models, such as `swin`.

GalyaZalesskaya: I think, Evgeny's point is that we have several algos in `src/otx/algo/hooks/recording_forward_hook.py` for different types of models. For transformer-based models like `swin` we should run `ViTReciproCAMHook`.

I think, I should revisit XAI in `torchvision_models` either way, so I will create if-else condition to chose proper algo for the proper architecture

GalyaZalesskaya (approved):

src/otx/algo/classification/torchvision_model.py
negvet: `ViTReciproCAMHook` most probably would not work with `swin` out of the box. `ViTReciproCAMHook` is applicable for vanila `ViT` architecture

negvet (approved):

src/otx/algo/classification/torchvision_model.py
GalyaZalesskaya: Oh, thank you for this point

harimkang (approved):

PR #3253, opened 2024-04-02
Decouple detection data pipeline from mmdet

sungchul1:
### Summary

This PR includes:
- [x] Update data pipeline recipes
    - [x] Get mmdet transforms
        - `MinIoURandomCrop`
        - `Resize`
        - `RandomFlip`
        - `PhotoMetricDistortion`
        - `RandomAffine`
        - `CachedMosaic`
        - `CachedMixUp`
        - `YOLOXHSVRandomAug`
        - `Pad`
    - Reproduced results

          | framework | Best mAP | Epoch for best mAP | Total epoch | data_time | iter_time
        -- | -- | -- | -- | -- | -- | --
        atss_mobilenetv2 | mmdet | 0.302 ± 0.021 | 15.4 ± 7.27 | 25.8 ± 5.22 | 0.044 ± 0.004 | 0.263 ± 0.088
          | torchvision | 0.302 ± 0.021 (+0.00) | 15.4 ± 7.27 | 25.8 ± 5.22 | 0.046 ± 0.003 (1.05x) | 0.284 ± 0.085 (1.08x)
        atss_resnext101 | mmdet | 0.390 ± 0.006 | 4.0 ± 1.87 | 15.4 ± 3.91 | 0.057 ± 0.075 | 0.300 ± 0.086
          | torchvision | 0.390 ± 0.006 (+0.00) | 4.0 ± 1.87 | 15.4 ± 3.91 | 0.044 ± 0.002 (0.77x) | 0.307 ± 0.088 (1.02x)
        ssd_mobilenetv2 | mmdet | 0.146 ± 0.008 | 21.8 ± 6.42 | 31.8 ± 9.86 | 0.073 ± 0.023 | 0.251 ± 0.037
          | torchvision | 0.146 ± 0.008 (+0.00) | 21.8 ± 6.42 | 31.8 ± 9.86 | 0.083 ± 0.048 (1.14x) | 0.261 ± 0.054 (1.04x)
        yolox_tiny | mmdet | 0.176 ± 0.028 | 7.8 ± 7.16 | 16.2 ± 6.57 | 0.055 ± 0.010 | 0.244 ± 0.063
          | torchvision | 0.176 ± 0.028 (+0.00) | 7.8 ± 7.16 | 16.2 ± 6.57 | 0.062 ± 0.006 (1.13x) | 0.276 ± 0.074 (1.13x)
        yolox_s | mmdet | 0.220 ± 0.041 | 24.6 ± 7.40 | 36.6 ± 8.53 | 0.057 ± 0.007 | 0.343 ± 0.061
          | torchvision | 0.220 ± 0.041 (+0.00) | 24.6 ± 7.40 | 36.6 ± 8.53 | 0.053 ± 0.002 (0.93x) | 0.340 ± 0.075 (0.99x)
        yolox_l | mmdet | 0.322 ± 0.036 | 13.0 ± 8.12 | 22.6 ± 7.92 | 0.085 ± 0.113 | 0.373 ± 0.146
          | torchvision | 0.322 ± 0.036 (+0.00) | 13.0 ± 8.12 | 22.6 ± 7.92 | 0.087 ± 0.120 (1.02x) | 0.372 ± 0.143 (1.00x)
        yolox_x | mmdet | 0.292 ± 0.047 | 20.0 ± 10.49 | 27.8 ± 7.92 | 0.055 ± 0.005 | 0.267 ± 0.071
          | torchvision | 0.292 ± 0.047 (+0.00) | 20.0 ± 10.49 | 27.8 ± 7.92 | 0.054 ± 0.003 (0.98x) | 0.262 ± 0.072 (0.98x)

    - Refactor the transforms with torchvision transform policy -> proceeded in the next PR if needed
- [x] misc
    - [x] Temporarily change `scale_factor` from (h, w) to (w, h) to fit https://github.com/open-mmlab/mmcv/blob/v2.1.0/mmcv/transforms/processing.py#L193 -> to be reverted in the next PR
       

### Checklist

### License

```python
```


src/otx/core/data/entity/base.py
```diff
@@ -247,7 +247,7 @@ def _resize_image_info(image_info: ImageInfo, size: list[int], **kwargs) -> Imag
 
     ori_h, ori_w = image_info.ori_shape
     new_h, new_w = image_info.img_shape
-    image_info.scale_factor = (new_h / ori_h, new_w / ori_w)
+    image_info.scale_factor = (new_w / ori_w, new_h / ori_h)
```
sungchul1: @vinnamkim @harimkang @jaegukhyun 
I tried to change `scale_factor` from `(h, w)` to `(w, h)` to match https://github.com/open-mmlab/mmcv/blob/v2.1.0/mmcv/transforms/processing.py#L193.
With `(h, w)`, predicted bounding boxes in `mmdet.base_dense_head` are wrongly post-processed.
Is there any conflict affecting other tasks?

vinnamkim:
Which requirement does this PR target? I don't see that this work is needed for us. AFAIK, we are implementing our own model implementations away from `mm*`. Why don't we let the existing recipe as is and introduce a new data pipeline to a new recipe equipped with our own model implementations?

sungchul1:
> Which requirement does this PR target? I don't see that this work is needed for us. AFAIK, we are implementing our own model implementations away from `mm*`. Why don't we let the existing recipe as is and introduce a new data pipeline to a new recipe equipped with our own model implementations?

This is a part of the task about decoupling from mmx.
It is required to update not only models but also data pipelines because current data pipelines are strongly coupled with mmx.
Newly created recipes are just for comparison and testing, so they will be merged with original recipes after passing the performance check.

vinnamkim:
> This is a part of the task about decoupling from mmx. It is required to update not only models but also data pipelines because current data pipelines are strongly coupled with mmx. Newly created recipes are just for comparison and testing, so they will be merged with original recipes after passing the performance check.

No, the data pipeline is not coupled. Rather, it was designed to be decoupled at the beginning. The fact that we can borrow the MM transform pipeline and build an OTX data transform pipeline shows that it is decoupled.

What you have to compare is that 
1. The old recipe equipped with `mm*` models and pipeline 
2. The new recipe equipped with own implementation models and torchvision pipeline

, not the old recipe equipped with `mm*` models but torchvision pipeline

sungchul1:
> > This is a part of the task about decoupling from mmx. It is required to update not only models but also data pipelines because current data pipelines are strongly coupled with mmx. Newly created recipes are just for comparison and testing, so they will be merged with original recipes after passing the performance check.
> 
> No, the data pipeline is not coupled. Rather, it was designed to be decoupled at the beginning. The fact that we can borrow the MM transform pipeline and build an OTX data transform pipeline shows that it is decoupled.
> 
> What you have to compare is that
> 
> 1. The old recipe equipped with `mm*` models and pipeline
> 2. The new recipe equipped with own implementation models and torchvision pipeline
> 
> , not the old recipe equipped with `mm*` models but torchvision pipeline

Yes, you're right. So @jaegukhyun and I are trying the second one.
@jaegukhyun will set models and I'm setting torchvision pipeline, because mmdetection's data pipeline & transforms are currently used on detection task.
It can be a huge PR if both are combined, so we decided to push PRs separately.

src/otx/core/data/dataset/detection.py
```diff
@@ -49,11 +49,20 @@ def _get_item_impl(self, index: int) -> DetDataEntity | None:
                 bboxes,
                 format=tv_tensors.BoundingBoxFormat.XYXY,
                 canvas_size=img_shape,
-            ),
+            )
+            if self.dm_subset.name == "train"
+            else torch.as_tensor(bboxes),  # to avoid resizing bounding boxes when validation and testing
```
vinnamkim: You are creating a if-else branch in the common module. However, it is extremely dedicated to just one local case where a data pipeline produces resized bounding box, but a model does not allow it. This is basically not desired and turning back the clock to get OTX 2 back into OTX 1.

src/otx/core/data/entity/base.py
vinnamkim: See https://github.com/openvinotoolkit/training_extensions/blob/977f99275066d1a1b895cea321053619b23305fe/src/otx/core/data/entity/base.py#L62-L67

Its definition is `(h,w)` and aligned with the other definitions, e.g., `img_shape`. What you mentioned is `mm*`'s problem. It should be handled in `src/otx/core/data/transform_libs/mm*.py`

vinnamkim:
~~> Yes, you're right. So @jaegukhyun and I are trying the second one. @jaegukhyun will set models and I'm setting torchvision pipeline, because mmdetection's data pipeline & transforms are currently used on detection task.~~

~~I'm still not able to find a good reason for removing the existing recipes from the codebase.~~

~~> Why don't we let the existing recipe as is and introduce a new data pipeline to a new recipe equipped with our own model implementations?~~

~~I believe that you can still move the existing recipes to somewhere (`src/otx/recipe/__deprecated__/**`) and make new ones.~~

Nevermind it. I found that the new recipe is `src/otx/recipe/detection/atss_mobilenetv2_tv.yaml`, not replacing the existing one.

> It can be a huge PR if both are combined, so we decided to push PRs separately.

I don't know whether it is working well. How about creating an individual PR for each model? Even in this PR, I can see that the model implementation and data pipeline is hardly coupled each other. Please see https://github.com/openvinotoolkit/training_extensions/pull/3253/files#r1547485311
I guess that you are changing the common part, not to touch the detection model itself. It is not allowed. I recommend you to open a PR that includes both data pipeline and model implementation for each model.

src/otx/core/data/dataset/detection.py
sungchul1: I agree with your comment, but I temporarily set this because this is needed if using `torchvision.transforms.v2.Resize` because it forces to transform tensors if there is `BoundingBoxes` tv_tensor.
I'm seeking the appropriate location for this patching and I'll update it.

src/otx/core/data/entity/base.py
sungchul1: Yes, the definition is right and I think this way is totally right.
But just for now, matching `scale_factor` with mmx pipelines seems more important because it can affect other models implemented from mmx.
And if `scale_factor` with `h, w` is used, there are many things to be changed in model implementations.
So more discussion is required with other assigned people.

wonjuleee: How about leaving this now and update to `(h, w)` again when merging model part?

sungchul1:
@harimkang Is it needed to add torchvision recipes into integration tests?

harimkang:
> @harimkang Is it needed to add torchvision recipes into integration tests?

sorry, i didn't get it.
If it means replacing the entire pipeline of detection models, wouldn't it be better to do it in a different PR per each model?

src/otx/core/data/dataset/detection.py
sungchul1: We don't use torchvision operations in this PR, so we can handle whether transforming bboxes or not in the data pipeline, not common modules.
And reverting transformed bboxes in metric side has two hurdles: it is needed to add additional things into `_convert_pred_entity_to_compute_metric` in `src/otx/core/model/detection.py`, and it might be complicated to use model's postprocessing functions.
So, I added a flag about this function at https://github.com/openvinotoolkit/training_extensions/pull/3253/commits/c9bcb3673721eaa63cbbc4ed09d45811f8cc2fc8 because it is the simplest way of this issue.
If you think it should be done on metric side, not data pipeline side, I'll consider to update it when merging model and data sides.

harimkang (dismissed):
LGTM, but is there any result data on whether this actually behaves the same in terms of speed (and other) as a existing mm's pipeline?

sungchul1:
> LGTM, but is there any result data on whether this actually behaves the same in terms of speed (and other) as a existing mm's pipeline?

Updated the reproduced results.

jaegukhyun (dismissed):

sungchul1:
@wonjuleee @vinnamkim Could you check unresolved conversations?

src/otx/core/data/entity/base.py
```diff
@@ -247,7 +247,7 @@ def _resize_image_info(image_info: ImageInfo, size: list[int], **kwargs) -> Imag
 
     ori_h, ori_w = image_info.ori_shape
     new_h, new_w = image_info.img_shape
-    image_info.scale_factor = (new_h / ori_h, new_w / ori_w)
+    image_info.scale_factor = (new_w / ori_w, new_h / ori_h)  # TODO (sungchul): revert to (h, w)
```
vinnamkim: Create a JIRA ticket and attach the ticket number here

sungchul1: I'll.
And I found this part, when using mmdet data pipeline, `scale_factor` which is obtained through `mmdet.Resize` is not converted from `(w, h)` to `(h, w)` when assigning it to `image_info.scale_factor` at L201 in the below file.
https://github.com/openvinotoolkit/training_extensions/blob/bd58fa3bde6d50f486a73c20ae2163c61965dc75/src/otx/core/data/transform_libs/mmdet.py#L184-L204
It seems a hidden bug which doesn't use `img_info.scale_factor`'s format.

sungchul1: > Create a JIRA ticket and attach the ticket number here

Done at https://github.com/openvinotoolkit/training_extensions/pull/3253/commits/49ad23e76c2aeb3be3121fea47fcd68c10125d6d.

harimkang (approved):

vinnamkim (approved):

PR #3264, opened 2024-04-03
Fix Dice score. Add mIoU metric for semantic segmentation

kprokofi:
### Summary
1) Ignore index was handled wrongly. Loss function also used 255 ignore index + 255 index was inserted into mask after transformations. It leads to an error during training. This PR patch Dice score to prevent such issues.

2) mIoU is more representative metric, it helps to compare models better. It is also more common choice for articles. So, adding this is very helpful for development. 

### Checklist

### License

```python
```


src/otx/core/metrics/dice.py
```diff
@@ -2,16 +2,33 @@
...
 from otx.core.types.label import LabelInfo
 
 
-def _dice_callable(label_info: LabelInfo) -> MetricCollection:
+def _segm_callable(label_info: LabelInfo) -> MetricCollection:
     return MetricCollection(
-        {"Dice": Dice(num_classes=label_info.num_classes + 1, ignore_index=label_info.num_classes)},
+        {
+            "Dice": OTXDice(num_classes=label_info.num_classes, ignore_index=-1, average="macro"),
+            "mIoU": JaccardIndex(task="multiclass", num_classes=label_info.num_classes, ignore_index=255),
```
vinnamkim: Why should they have different values for `ignore_index`?

vinnamkim: In addition, does we already has a dedicated `label_info` metadata entity, `SegLabelInfo`? By following your explanations,

> Ignore index was handled wrongly. Loss function also used 255 ignore index ...

I think that `ignore_index` in the metric should be configurable and `label_info` can be a source of that configuration. This time you can add `ignore_index` field to `SegLabelInfo` and set its default value as 255.

kprokofi: It is different, because Dice metric from torchvision doesn't support 255 common ignore index and raises an error regarding this. Moreover, it doesn't support even -1, it just ignores it w/o errors. 
Implementation for this class is not so good. Copy and paste class constructor just to delete assertion seemed to me not a good idea, so I use this workaround

kprokofi: Ok, I will try to add possibility to change ignore index, however, we should also change loss function and augmentations

vinnamkim: As you mentioned, it is configurable from the model (the loss function) and data pipeline (the augmentations) configurations. My hope is that it should be configurable at the metric side as same as the others, even if we normaly use a fixed value for `ignore_index=255` in fact.

jaegukhyun (dismissed):

src/otx/recipe/_base_/data/mmseg_base.yaml
```diff
@@ -35,6 +36,7 @@ config:
         size:
           - 512
           - 512
+        pad_val: { img: 0, seg: 255 }
```
vinnamkim:
```suggestion
        pad_val:
          img: 0
          seg: 255
```

src/otx/core/types/label.py
```diff
@@ -271,7 +271,7 @@ def from_json(cls, serialized: str) -> HLabelInfo:
 class SegLabelInfo(LabelInfo):
     """Meta information of Semantic Segmentation."""
 
-    def __init__(self, label_names: list[str], label_groups: list[list[str]]) -> None:
+    def __init__(self, label_names: list[str], label_groups: list[list[str]], ignore_index: int | None = 255) -> None:
```
vinnamkim:
```suggestion
    def __init__(self, label_names: list[str], label_groups: list[list[str]], ignore_index: int = 255) -> None:
```

vinnamkim (approved):

sovrasov (approved):

PR #3265, opened 2024-04-04
Fix reference info path for zero-shot learning

sungchul1:
### Summary

Fixed the issue about stacking reference info files in the specific path.
- Updated the path to `work_dir`, the files will be saved in tmp dir during test.

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algo/visual_prompting/zero_shot_segment_anything.py
+++ b/src/otx/algo/visual_prompting/zero_shot_segment_anything.py
@@ -6,7 +6,7 @@
 from __future__ import annotations
 
 import logging as log
-import os
+import pickle
 from collections import defaultdict
 from copy import deepcopy
 from itertools import product
@@ -263,9 +263,9 @@ class ZeroShotSegmentAnything(SegmentAnything):
                 for input_prompt in input_prompts:
                     if isinstance(input_prompt, Mask):
                         # directly use annotation information as a mask
-                        ref_mask[
-                            input_prompt == 1
-                        ] += 1  # TODO (sungchul): check if the mask is bool or int # noqa: TD003
+                        ref_mask[input_prompt == 1] += (
+                            1  # TODO (sungchul): check if the mask is bool or int # noqa: TD003
+                        )
                     else:
                         if isinstance(input_prompt, BoundingBoxes):
                             point_coords = input_prompt.reshape(-1, 2, 2)
@@ -604,7 +604,8 @@ class OTXZeroShotSegmentAnything(OTXVisualPromptingModel):
         self,
         backbone: Literal["tiny_vit", "vit_b"],
         num_classes: int = 0,
-        root_reference_info: Path | str = "vpm_zsl_reference_infos",
+        reference_info_dir: Path | str = "reference_infos",
+        infer_reference_info_root: Path | str = "../.latest/train",
         save_outputs: bool = True,
         pixel_mean: list[float] | None = [123.675, 116.28, 103.53],  # noqa: B006
         pixel_std: list[float] | None = [58.395, 57.12, 57.375],  # noqa: B006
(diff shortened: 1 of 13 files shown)


jaegukhyun:
Should we apply this PR to 2.0 releases? It would be better to apply this PR to develop if this is not an urgent issue,

sungchul1:
> Should we apply this PR to 2.0 releases? It would be better to apply this PR to develop if this is not an urgent issue,

@yunchu what do you think? I'm preparing it for develop as well, but if this PR is not necessary for 2.0, I'll update it to develop only.

src/otx/algo/visual_prompting/zero_shot_segment_anything.py
```diff
@@ -604,7 +604,8 @@ def __init__(
         self,
         backbone: Literal["tiny_vit", "vit_b"],
         num_classes: int = 0,
-        root_reference_info: Path | str = "vpm_zsl_reference_infos",
+        reference_info_dir: Path | str = "reference_infos",
+        infer_reference_info_root: Path | str = "../.latest/train",
```
harimkang, line 608: `.latest` is the output folder, which is only available in the CLI. This default value seems to need to be modified for API usage.
Also, since `.latest` is a soft-linked folder, it seems appropriate to browse to the actual folder.

src/otx/core/model/entity/visual_prompting.py
```diff
@@ -357,7 +356,8 @@ def __init__(
         max_num_requests: int | None = None,
         use_throughput_mode: bool = True,
         model_api_configuration: dict[str, Any] | None = None,
-        root_reference_info: str = "vpm_zsl_reference_infos",
+        reference_info_dir: Path | str = "reference_infos",
+        infer_reference_info_root: Path | str = "../.latest/train",
```
harimkang, line 360: Same here

harimkang (commented):
In my opinion, accessing `.latest` in code is not a good idea: not only is it only allowed in the CLI, but it's also unstable because `.latest` is actually a symlinked address. I think we need a way to resolve to the actual dir path.

src/otx/algo/visual_prompting/zero_shot_segment_anything.py
sungchul1: How can I access `work_dir` for training during test?
I want to save the files in `work_dir` for training and use them during test.
As far as I know, it seems there are only hard-coded path like that or user input like below
https://github.com/openvinotoolkit/training_extensions/blob/9b1c5c781efd70ad260fc1be4350b144bab66b92/tests/integration/api/test_auto_configuration.py#L47
https://github.com/openvinotoolkit/training_extensions/blob/9b1c5c781efd70ad260fc1be4350b144bab66b92/tests/integration/cli/test_cli.py#L107-L108

src/otx/algo/visual_prompting/zero_shot_segment_anything.py
```diff
@@ -841,29 +843,52 @@ def initialize_reference_info(self) -> None:
...
-            return None
-        if len(stamps := sorted(os.listdir(root), reverse=True)) > 0:
-            return stamps[0]
-        return None
+    def save_reference_info(self, default_root_dir: Path | str) -> None:
+        """Save reference info."""
+        reference_info = {
+            "reference_feats": self.reference_feats,
+            "used_indices": self.used_indices,
+        }
```
vinnamkim, line 851: I'm curious why these tensors cannot be not included in the model checkpoint.

sungchul1: We changed direction to return those tensors, not including them in the model.
But for OTX CLI standalone usecase, they are required for inference, so I and @wonjuleee decided to save them at the specific location and load them.

vinnamkim: If you are worrying about the big size of checkpoint file, I think you can still amortize this to the model checkpoint. You can override `load_state_dict()` and `state_dict()` with dropping the model weight part from them. This will make it more streamlinedwith the checkpoint callback.

sungchul1: Thanks for the recommendation. I'll take a look for a better direction.

vinnamkim (approved):
Approved because it is targeted to `releases/2.0.0`. Please use a model state dictionary based logic if you try to fix this on the develop.

jaegukhyun (approved):

wonjuleee:
> Approved because it is targeted to `releases/2.0.0`. Please use a model state dictionary based logic if you try to fix this on the develop.

Hi @vinnamkim, thank you for the suggestion. We have kept the reference feature within model state dictionary originally, but it made us difficult to put in reference feature dynamically. This was indeed a problem for OV IR too. So we turn to separate the reference feature from model state dictionary. This looks like more applicable for API usage from Geti VPS. What do you think?

sungchul1:
@harimkang I'll close your conversations because we discussed it in chat and thought there was no way to use configs from engine or lit module.

PR #3271, opened 2024-04-04
Decouple mmpretrain: Mobilenet-V3-large

harimkang:
### Summary

- [x] Replace mmpretrain to torchvision transforms & pipeline
- [x] Bring Classifier, Head, Neck, Loss from mmpretrain
- [x] Add Recipes
- [x] Multi-Class classification
- [x] Multi-Label classification
- [x] H-Label Classification
- [x] Revisit Ignored Label
- [x] Modify Mobilenet-v3-large Backbone & Model
- [ ] XAI support -> XAI with IR Model issue: report to Vlad & Galya https://jira.devtools.intel.com/browse/CVS-138604
- [ ] Decouple mmengine BaseModule -> Will be next PR
- [x] Unit-Test

<head>

</head>

<div style='direction:ltr'>

task | dataset | model | version | test/accuracy | train/iter_time | duration |   |  
-- | -- | -- | -- | -- | -- | -- | -- | --
MULTI_CLASS_CLS | Small_0 | mobilenet_v3_large | Old | 0.9661 | - | 22 | mmpretrain |  
  |   |   | New | 1.0 | - | 25 | Decouple   mmpretrain   + torchvision   transform |  
  | Small_1 |   | Old | 0.9661 | - | 22 |   |  
  |   |   | New | 0.9661 | - | 25 |   |  
  | Small_2 |   | Old | 0.9153 | - | 22 |   |  
  |   |   | New | 0.9831 | - | 25 |   |  
  | Medium |   | Old | 0.7873 | 0.0421 | 72 | 23   epochs |  
  |   |   | New | 0.7662 | 0.0830 | 126 | 25   epochs |  
  |   |   | New 2 | 0.7623 | 0.0644 | 108 | 27   epochs | Resize   (NEAREST)
  | Large |   | Old | 0.8334 | 0.0735 | 414 | 21   epochs |  
  |   |   | New | 0.8235 | 0.1350 | 504 | 13   epochs |  
  |   |   | New 2 | 0.8322 | 0.1143 | 480 | 16   epochs | Resize   (NEAREST)

</div>

</body>

</html>

I've noticed that torchvision transforms is slower than the mmcv pipeline, so I'm looking into it.
The first discovery is that using the existing interpolation, BILINEAR, on resize forces the tensor to be cast to float32 (NEW). Using NEAREST(0) as the interpolation is slightly faster. (NEW2)
https://github.com/pytorch/vision/blob/5181a854d8b127cf465cd22a67c1b5aaf6ccae05/torchvision/transforms/v2/functional/_geometry.py#L234-L274
I got some improvement, but it's still slow in terms of `iter_time`.

### Checklist

### License

```python
```


vinnamkim (commented):
Why is this implementation needed although we already has
https://github.com/openvinotoolkit/training_extensions/blob/develop/src/otx/algo/classification/torchvision_model.py
?

harimkang:
> Why is this implementation needed although we already has https://github.com/openvinotoolkit/training_extensions/blob/develop/src/otx/algo/classification/torchvision_model.py ?

@vinnamkim 
I proceed with the requirement below.
https://jira.devtools.intel.com/browse/CVS-137602

I'm not sure if I'm clear on this, but as I understand it, the goal is to make sure that the model using in Geti doesn't use mm.
I copy some code from mm to ensure compatibility with the model was already using in Geti.

vinnamkim:
> > Why is this implementation needed although we already has https://github.com/openvinotoolkit/training_extensions/blob/develop/src/otx/algo/classification/torchvision_model.py ?
> 
> @vinnamkim I proceed with the requirement below. https://jira.devtools.intel.com/browse/CVS-137602
> 
> I'm not sure if I'm clear on this, but as I understand it, the goal is to make sure that the model using in Geti doesn't use mm. I copy some code from mm to ensure compatibility with the model was already using in Geti.
> 
> @wonjuleee Am I understanding this correctly?

From my understanding, it is accepted that Geti keeps a OTX version field. We can maintain at least two OTX images for Geti training: `OTX 1.x` and `OTX 2.x`. It means that we don't worry about a compatibility issue between `OTX 1.x` and `OTX 2.x`. If we don't need to consider it, there is no need to stick to the MM implementations.

harimkang:
> From my understanding, it is accepted that Geti keeps a OTX version field. We can maintain at least two OTX images for Geti training: `OTX 1.x` and `OTX 2.x`. It means that we don't worry about a compatibility issue between `OTX 1.x` and `OTX 2.x`. If we don't need to consider it, there is no need to stick to the MM implementations.

@vinnamkim I will be happy if it goes as well as your suggestion. I also think this work becomes unnecessary once we no longer need to consider compatibility.
However, As i know that similar work is currently in progress for other tasks.. @wonjuleee What do you think?

tests/integration/api/test_xai.py
```diff
@@ -95,6 +95,9 @@ def test_predict_with_explain(
     if "dino" in model_name or "rtmdet_inst_tiny" in model_name:
         pytest.skip("DINO and Rtmdet_tiny are not supported.")
 
+    if "mobilenet_v3_large" in model_name:
+        pytest.skip("There's issue with mobilenet_v3_large model. Skip for now.")
```
harimkang, line 99: Hi, @sovrasov @negvet @GalyaZalesskaya  When I run the code below on this PR, the outputs of exported_model with XAI is weird (feature_vector & silency_map shape), Is there something I'm missing?
```python
from pathlib import Path

import openvino.runtime as ov
from otx.engine import Engine

recipe = "src/otx/recipe/classification/multi_class_cls/mobilenet_v3_large.yaml"
data_root = "tests/assets/classification_dataset"
work_dir = "otx-workspace-xai-test"

engine = Engine.from_config(
    config_path=recipe,
    data_root=data_root,
    work_dir=work_dir,
)


predict_result_explain_torch = engine.predict(explain=True)
ckpt_path = Path(work_dir) / "checkpoint.ckpt"
engine.trainer.save_checkpoint(ckpt_path)
exported_model_path = engine.export(checkpoint=ckpt_path, explain=True)


model = ov.Core().read_model(exported_model_path)
feature_vector_output = None
saliency_map_output = None
for output in model.outputs:
    if "feature_vector" in output.get_names():
        feature_vector_output = output
    if "saliency_map" in output.get_names():
        saliency_map_output = output
```

image
in ov model api side,
image

src/otx/algo/classification/backbones/__init__.py
```diff
@@ -5,6 +5,6 @@
 
 from .otx_efficientnet import OTXEfficientNet
 from .otx_efficientnet_v2 import OTXEfficientNetV2
-from .otx_mobilenet_v3 import OTXMobileNetV3
+from .mobilenet_v3 import OTXMobileNetV3
```
jaegukhyun, line 8: If we don't need to register models to registry, do we need this?

src/otx/algo/classification/backbones/mobilenet_v3.py
```diff
@@ -308,11 +331,10 @@ def _initialize_weights(self):
                 m.bias.data.zero_()
 
 
-@MODELS.register_module()
 class OTXMobileNetV3(MobileNetV3):
     """MobileNetV3 model for OTX."""
 
-    cfgs = dict(
+    mv_cfgs = dict(
```
jaegukhyun: What is mv mean? Could we make this more clear?

src/otx/algo/classification/heads/hlabel_cls_head.py
```diff
@@ -0,0 +1,356 @@
...
+
+"""Module for defining h-label linear classification head."""
+
+from __future__ import annotations
+
+import inspect
+from typing import Callable
+
+import torch
+from mmengine.model import BaseModule
```
jaegukhyun, line 12: Will this be replaced after?

src/otx/algo/classification/losses/asymmetric_angular_loss_with_ignore.py
```diff
@@ -6,23 +6,77 @@
 from __future__ import annotations
 
 import torch
-from mmpretrain.models.losses.utils import weight_reduce_loss
+
+# TODO(harimkang): Remove the following import after decoupling mmpretrain
 from mmpretrain.registry import MODELS
```
jaegukhyun, line 11: Just for a question, where is this registry used?

src/otx/algo/utils/mmengine_utils.py
```diff
@@ -0,0 +1,216 @@
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+"""Load Checkpoint functions."""
+# TODO(someone): Revisit mypy errors after deprecation of mmlab
+# mypy: ignore-errors
```
jaegukhyun, line 6: What kind of errors cannot be handled in this PR?

src/otx/algo/classification/backbones/__init__.py
harimkang: This is simply for use directly in `otx.algo.classification.backbones`. Are there any concerns?

src/otx/algo/classification/heads/hlabel_cls_head.py
harimkang: Yes I will be working on that after get all the other backbone stuff and mmpretrain out of the way.

src/otx/algo/classification/losses/asymmetric_angular_loss_with_ignore.py
harimkang: Currently, this PR has only been replaced for the mobilenet-v3-large usage model, so other multi-label classification models still need to be added to the registry.

src/otx/algo/utils/mmengine_utils.py
harimkang: This is copy code of some functions provided by mmengine, so it's not task-specific, and we need to clean it up later, and I think mypy will be worked on at that time.

src/otx/algo/classification/backbones/mobilenet_v3.py
harimkang: Thanks, renamed - https://github.com/openvinotoolkit/training_extensions/pull/3271/commits/ead07043aee7a5f9536075814bd86d016b52cc4a

tests/integration/api/test_xai.py
GalyaZalesskaya: Looks like output names for XAI are not aligned correctly with outputs.
Based on shapes names[965] should be `saliency map` and names[885] should be the `feature vector`. 
Probably it happened after latest export/XAI refactoring. I'll create a ticket for this issue and investigate it.

Thank you for noticing and reporting this behavior!

src/otx/algo/classification/backbones/__init__.py
jaegukhyun: I just worried about init speed. But this can provide convenience.

jaegukhyun (dismissed):

src/otx/algo/classification/backbones/mobilenet_v3.py
```diff
@@ -135,10 +134,36 @@ def conv_1x1_bn(inp, oup, loss="softmax"):
     )
 
 
+def make_divisible(value: int | float, divisor: int, min_value: int | None = None, min_ratio: float = 0.9):
```
sungchul1, line 137: Is this function used in other models? If so, it seems better to move it to `utils`.

src/otx/algo/classification/backbones/mobilenet_v3.py
```diff
@@ -308,11 +331,10 @@ def _initialize_weights(self):
                 m.bias.data.zero_()
 
 
-@MODELS.register_module()
 class OTXMobileNetV3(MobileNetV3):
     """MobileNetV3 model for OTX."""
 
-    cfgs = dict(
+    backbone_configs = dict(
```
sungchul1, line 337: Is there any problem to use the previous name?

sungchul1 (commented):
Thanks for your hard work. I left some minor comments, please comment them :)

src/otx/algo/classification/backbones/mobilenet_v3.py
harimkang: There's nothing particularly wrong with this, and i just changed it to use a clearer name. Jaeguk also wanted to use a clearer name.

src/otx/algo/classification/backbones/mobilenet_v3.py
harimkang: I'll check and move this to other backbones and models in the next PR, but for now, I see that it's only used in MV3, so I put it there.

sungchul1 (approved):

jaegukhyun (approved):

PR #3281, opened 2024-04-05
MMDet MaskRCNN ResNet50/SwinTransformer Decouple

eugene123tw:
### Summary

* Moved MMDet MaskRCNN
* Moved MMDet ResNet50 
* Moved MMDet SwinTransformer
* Moved MMDeploy MaskRCNN components

### Checklist

### License

```python
```


src/otx/core/model/instance_segmentation.py
```diff
@@ -403,6 +403,7 @@ def _customize_inputs(self, entity: InstanceSegBatchDataEntity) -> dict[str, Any
         ):
             # NOTE: ground-truth masks are resized in training, but not in inference
             height, width = img_info.img_shape if self.training else img_info.ori_shape
+            mmdet_masks: BitmapMasks | PolygonMasks
```
harimkang, line 420: Why do we need this separately?

harimkang (commented):
At first glance, it looks like you're importing and replacing modules.
@jaegukhyun Can you take a look to make sure there's no duplication with detection-related tasks?

Will the `MaskRCNN` in `src/otx/algo/instance_segmentation/maskrcnn.py`, which is used as the main, be replaced with `OTXInstanceSegModel` instead of `MMDetInstanceSegCompatibleModel` in another PR?

src/otx/core/model/instance_segmentation.py
eugene123tw: I need to add type hints for mypy to recognize.

jaegukhyun:
I have one question during review.
Will MaskRCNN family be exported using NativeExporter or MMDeployExporter?

src/otx/algo/detection/losses/utils.py
jaegukhyun, line 1: Sorry for late. This file is in otx/algo/det/losses/weighted_loss.py

src/otx/algo/instance_segmentation/mmdet/models/detectors/base.py
jaegukhyun, line 1: I think this class would give complexity to algo folder. In my case I make single stage detector to support base detector's methods. So we can try 
1. TwoStageDetector inherits SingleStageDetector
2. TwoStageDetector have methods of BaseDetector as SingleStageDetector do
3. Make BaseDetector, SingleStageDetector and TwoStageDetector will inherit BaseDetector.
I want to recommend 2nd approach, what do you think?

src/otx/algo/instance_segmentation/mmdet/models/layers/bbox_nms.py
jaegukhyun, line 1: We have nms from mmdet in otx/algo/det/ops/nms.py. Could we use it here?

eugene123tw:
> At first glance, it looks like you're importing and replacing modules. @jaegukhyun Can you take a look to make sure there's no duplication with detection-related tasks?
> 
> Will the `MaskRCNN` in `src/otx/algo/instance_segmentation/maskrcnn.py`, which is used as the main, be replaced with `OTXInstanceSegModel` instead of `MMDetInstanceSegCompatibleModel` in another PR?


Hi @harimkang, maybe I need to clarify some points. I've transferred all the essential MaskRCNN code and backbone code from MMDet to OTX 2.x, reflecting a similar approach with MMDet detection models or the mmcv conv/norm module you did. Also, reuse as many components as Jaeguk moved. 

I'm not sure of the rationale behind replacing `MMDetInstanceSegCompatibleModel` with `OTXInstanceSegModel`. As far as I understand, we still require `MMXCompatibleModel` classes to execute mmlab models on OTX 2.x without installing MMDet. I believe this PR aims to enable Geti MMDet models in OTX without having to install mmdet or mmcv.

src/otx/algo/instance_segmentation/mmdet/models/utils/misc.py
jaegukhyun, line 1: We have utils in otx/algo/det/utils/utils. It is totally fine that remove duplicated things after mergeing this PR.
I just want you to notice

src/otx/algo/instance_segmentation/mmdet/models/utils/typing_utils.py
jaegukhyun, line 1: In my case I just interpret all data structure from mmdet. So detection algo only depend on mmengine and it will be replaced by otx method in near future. What do you think about that kind of approach?

jaegukhyun (commented):
I think separate mmdet folder has pros and cons. My opinion is we don't need to specify that all of algorithms come from mmdet. @harimkang What do you think?

eugene123tw:
> I have one question during review. Will MaskRCNN family be exported using NativeExporter or MMDeployExporter?

@jaegukhyun Currently, it relies on MMDeployExporter. However, if we intend to remove mmdeploy dependency, this action should be addressed in a separate PR due to its large scope.

harimkang:
> Hi @harimkang, maybe I need to clarify some points. I've transferred all the essential MaskRCNN code and backbone code from MMDet to OTX 2.x, reflecting a similar approach with MMDet detection models or the mmcv conv/norm module you did. Also, reuse as many components as Jaeguk moved.
> 
> I'm not sure of the rationale behind replacing `MMDetInstanceSegCompatibleModel` with `OTXInstanceSegModel`. As far as I understand, we still require `MMXCompatibleModel` classes to execute mmlab models on OTX 2.x without installing MMDet. I believe this PR aims to enable Geti MMDet models in OTX without having to install mmdet or mmcv.

Yes, I think we're on the same page. But doesn't `MMDetInstanceSegCompatibleModel` also have a dependency on mm after all, and if we want to use it without installing mm, shouldn't eventually make sure that existing models don't use the `MMDetInstanceSegCompatibleModel` class in `MaskRCNN`?

Or are you saying you're going to leave the old model alone and add a new recipe (without mm)?

src/otx/algo/instance_segmentation/mmdet/models/layers/bbox_nms.py
eugene123tw: @jaegukhyun `otx/algo/det/ops/nms.py::multiclass_nms` looks different. It looks like it's designed for OV/ONNX model export, not for PyTorch training.

eugene123tw:
> > Hi @harimkang, maybe I need to clarify some points. I've transferred all the essential MaskRCNN code and backbone code from MMDet to OTX 2.x, reflecting a similar approach with MMDet detection models or the mmcv conv/norm module you did. Also, reuse as many components as Jaeguk moved.
> > I'm not sure of the rationale behind replacing `MMDetInstanceSegCompatibleModel` with `OTXInstanceSegModel`. As far as I understand, we still require `MMXCompatibleModel` classes to execute mmlab models on OTX 2.x without installing MMDet. I believe this PR aims to enable Geti MMDet models in OTX without having to install mmdet or mmcv.
> 
> Yes, I think we're on the same page. But doesn't `MMDetInstanceSegCompatibleModel` also have a dependency on mm after all, and if we want to use it without installing mm, shouldn't eventually make sure that existing models don't use the `MMDetInstanceSegCompatibleModel` class in `MaskRCNN`?
> 
> Or are you saying you're going to leave the old model alone and add a new recipe (without mm)?

@harimkang  `MMDetInstanceSegCompatibleModel` is a bridge for training MMDet models. With the decoupling of MMDet MaskRCNN in this PR, we also need to untangle data structures (such as BitmapMasks and PolygonMasks) and the instance segmentation data pipeline. However, this will require a separate PR to avoid overwhelming the current one.

Of course, there are still several mm-related components, including mmcv opset, mmdet data structures, mmdeploy, and mmengine, that still remain. It's possible to migrate all these components to OTX 2.x, it's also time-consuming. Given the scope, this migration cannot be achieved in a single PR.

Looking ahead, for torchvision.MaskRCNN, a new recipe will be necessary. And, to maintain backward compatibility, we can continue utilizing MMDetInstanceSegCompatibleModel until we find a solution for full compatibility between torchvision.MaskRCNN and MMDet MaskRCNN.

harimkang:
> Of course, there are still several mm-related components, including mmcv opset, mmdet data structures, mmdeploy, and mmengine, that still remain. It's possible to migrate all these components to OTX 2.x, it's also time-consuming. Given the scope, this migration cannot be achieved in a single PR.
> 
> Looking ahead, for torchvision.MaskRCNN, a new recipe will be necessary. And, to maintain backward compatibility, we can continue utilizing MMDetInstanceSegCompatibleModel until we find a solution for full compatibility between torchvision.MaskRCNN and MMDet MaskRCNN.

Yes I understand, thanks for the clarification! :)

src/otx/algo/instance_segmentation/mmdet/models/utils/typing_utils.py
eugene123tw: I agree. I will remove MMDet special typing.

src/otx/algo/instance_segmentation/mmdet/models/detectors/base.py
eugene123tw: @jaegukhyun we can minimize duplication by having `SingleStageDetector` inherit from `BaseDetector` as well. If that seems overly burdensome, I'm open to moving methods from `BaseDetector` to `TwoStageDetector` and having `TwoStageDetector` inherit from `BaseModel`. Specifically, I can adjust `class TwoStageDetector(BaseDetector)` to `class TwoStageDetector(BaseModel)`.

src/otx/algo/detection/heads/custom_anchor_generator.py
```diff
@@ -9,12 +9,13 @@
...
 import numpy as np
 import torch
-from mmdet.registry import TASK_UTILS
+from mmengine.registry import TASK_UTILS
 from torch.nn.modules.utils import _pair
 
 
 # This class and its supporting functions below lightly adapted from the mmdet AnchorGenerator available at:
 # https://github.com/open-mmlab/mmdetection/blob/cfd5d3a985b0249de009b67d04f37263e11cdf3d/mmdet/models/task_modules/prior_generators/anchor_generator.py
+@TASK_UTILS.register_module()
```
wonjuleee, line 18: Why should we need to register this within mmengine? As far as I understand, it is avoidable for installing even mmengine.

eugene123tw: @wonjuleee I want to highlight it will be used to build another anchor generator for RTMDetInst in an upcoming PR. Branching could be an option for handling this, that said, it's worth noting that numerous components currently depend on the mmengine registry. Removing this dependency is not within the scope of this PR. However, it's something we should address comprehensively at a later stage, although not suitable within this PR.

src/otx/algo/instance_segmentation/mmdet/models/detectors/base.py
jaegukhyun: Okay, Let's decide this after merging this PR

jaegukhyun (dismissed):

harimkang (approved):

jaegukhyun (approved):

PR #3284, opened 2024-04-07
Extend exportable code

kprokofi:
### Checklist

### License

```python
```


jaegukhyun (dismissed):
Just for the check, Will otx export {exported model} replace otx demo in OTX1.x?

kprokofi:
> Just for the check, Will otx export {exported model} replace otx demo in OTX1.x?

Yes. `otx deploy` and `otx demo` are replaced with `otx export`. Demo is now stand alone code, that used without OTX package.

kprokofi:
@eunwoosh , could you take a look at HPO issue? Did you face this problem?
image

eunwoosh:
> @eunwoosh , could you take a look at HPO issue? Did you face this problem? image

Could you check again? HPO test is set as 'xfailed', because it doesn't work in develop branch due to refactoring. It is being fixed currently.

src/otx/core/exporter/base.py
```diff
@@ -150,25 +163,45 @@ def to_exportable_code(
...
             precision (OTXExportPrecisionType, optional): precision of the exported model's weights
+            path_to_already_exported_model (PathLike, optional): path to the already
+                exported model to add it in exportable code
 
         Returns:
             Path: path to the exported model.
         """
-        work_dir = Path(demo.__file__).parent
+        work_demo_dir = Path(demo.__file__).parent
+        training_extensions_dir = Path(otx.__file__).parents[2]
```
eunwoosh: I think it doesn't work if otx is installed as non-editable mode. Could you copy LICENSE to demo directory and use it? You can refer #3255

vinnamkim (commented):
Is this PR about the ticket no. 135467? I still don't know why don't we just drop `OTXExportFormatType.EXPORTABLE_CODE` and introduce a boolean input argument such as `package_model_into_exportable_code: bool`. If we add this, the number of argument increases, but this PR too. In addition, the usage introduced by this PR seems more complicated to me.

Is there any special reason that we must keep `OTXExportFormatType.EXPORTABLE_CODE`?

kprokofi:
> Is this PR about the ticket no. 135467? I still don't know why don't we just drop `OTXExportFormatType.EXPORTABLE_CODE` and introduce a boolean input argument such as `package_model_into_exportable_code: bool`. If we add this, the number of argument increases, but this PR too. In addition, the usage introduced by this PR seems more complicated to me.
> 
> Is there any special reason that we must keep `OTXExportFormatType.EXPORTABLE_CODE`?

I don't see the difference in complexity.
If we remove OTXExportFormatType.EXPORTABLE_CODE, everything will remain same except that we need to pass additional "--save-as-zip" or whatever. It gets even more complicated, since we need to pass additional parameters to methods of the model class

Now, user just need to: otx export ... --export_type EXPORTABLE_CODE
If we remove it: otx export ... --save-as-zip

Is it make sense to delete additional export type?

vinnamkim:
> I don't see the difference in complexity. If we remove OTXExportFormatType.EXPORTABLE_CODE, everything will remain same except that we need to pass additional "--save-as-zip" or whatever. It gets even more complicated, since we need to pass additional parameters to methods of the model class
> 
> Now, user just need to: otx export ... --export_type EXPORTABLE_CODE If we remove it: otx export ... --save-as-zip
> 
> Is it make sense to delete additional export type?

Ticket no. 135467 is about packaging an optimized model (INT8), not exported model (FP16 or FP32) into an exportable code format (zip file). Following this PR, I imagine that the CLI workflow will be 
```console
otx optimize <some-path-1>
otx export --import <some-path-1> --export <some-path-2> -- export_type EXPORTABLE_CODE
```
I abbreviate a Python API case but it looks similar.

kprokofi:
> > I don't see the difference in complexity. If we remove OTXExportFormatType.EXPORTABLE_CODE, everything will remain same except that we need to pass additional "--save-as-zip" or whatever. It gets even more complicated, since we need to pass additional parameters to methods of the model class
> > Now, user just need to: otx export ... --export_type EXPORTABLE_CODE If we remove it: otx export ... --save-as-zip
> > Is it make sense to delete additional export type?
> 
> Ticket no. 135467 is about packaging an optimized model (INT8), not exported model (FP16 or FP32) into an exportable code format (zip file). Following this PR, I imagine that the CLI workflow will be
> 
> ```
> otx optimize <some-path-1>
> otx export --import <some-path-1> --export <some-path-2> -- export_type EXPORTABLE_CODE
> ```
> 
> I abbreviate a Python API case but it looks similar.

Optimized or exported models - doesn't matter. It is all about OpenVINO models

```
otx optimize ...
otx export ... --checkpoint path/to/.xml
```

If you pass .xml to export, you don't even need to change export_type, it will automatically proceed with exportable code

vinnamkim:
> Optimized or exported models - doesn't matter. It is all about OpenVINO models
> 
> otx optimize ... otx export --checkpoint path/to/.xml
> 
> If you pass .xml to export, you don't even need to change export_type

I'm asking why don't we abbreviate it to one step such as
```console
otx optimize --package-to-exportable-code
```
, and Python API too.

kprokofi:
> > Optimized or exported models - doesn't matter. It is all about OpenVINO models
> > otx optimize ... otx export --checkpoint path/to/.xml
> > If you pass .xml to export, you don't even need to change export_type
> 
> I'm asking why don't we abbreviate it to one step such as
> 
> ```
> otx optimize --package-to-exportable-code
> ```
> 
> , and Python API too.



> > Optimized or exported models - doesn't matter. It is all about OpenVINO models
> > otx optimize ... otx export --checkpoint path/to/.xml
> > If you pass .xml to export, you don't even need to change export_type
> 
> I'm asking why don't we abbreviate it to one step such as
> 
> ```
> otx optimize --package-to-exportable-code
> ```
> 
> , and Python API too.

Ok, I understand what you mean. I think we can add this functionality. I can add additional argument to `engine.optimize` and run `engine.export` after optimization if user wants to export models with demo

src/otx/core/exporter/base.py
kprokofi: Ok, let's add LICENSE to demo folder

kprokofi:
I added possibility to export demo during optimization + removed additional parameter in export with path to xml model. I found a better way

vinnamkim (dismissed):

sovrasov (approved):

eunwoosh (approved):

PR #3287, opened 2024-04-08
Add NMS iou threshold configurable parameter

jaegukhyun:
### Checklist

### License

```python
```


src/otx/algorithms/common/configs/training_base.py
```diff
@@ -231,6 +231,15 @@ class BasePostprocessing(ParameterGroup):
             affects_outcome_of=ModelLifecycle.INFERENCE,
         )
 
+        nms_iou_threshold = configurable_float(
+            default_value=0.0,
```
sovrasov: Why do we have 0 here? Should it be ~0.5?

src/otx/algorithms/detection/configs/base/data/tiling/efficientnet_iseg_tile_pipeline.py
```diff
@@ -8,7 +8,7 @@
 img_size = (512, 512)
 
 tile_cfg = dict(
-    tile_size=400, min_area_ratio=0.9, overlap_ratio=0.2, iou_threshold=0.45, max_per_img=1500, filter_empty_gt=True
+    tile_size=400, min_area_ratio=0.9, overlap_ratio=0.2, iou_threshold=0.7, max_per_img=1500, filter_empty_gt=True
```
sovrasov: How do we know that 0.7 is better?

sovrasov:
Are detectors so sensitive to IoU in NMS? We shouldn't apply any default value tricks, since it all is visible in the UI.
My initial thought was to bind 0.5 default value for all detectors and tilers, conduct a couple experiments, and call it a day. Unfortunately, that's a sacrifice we're making because of the imperfect Geti configuration system, which doesn't support per model tuning for default params.

If that's not possible (the detectors are too sensitive, we could think of changing the tiling threshold only)

src/otx/algorithms/common/configs/training_base.py
jaegukhyun: Every detectors have their own nms iou threshold in their model.py. It is different across the architecture, So here default value 0.0, which means we're going to use the default value of object detector. 
In addition, this methods can prevent confusion from OTX side. Currently OTX do not know whether hyperparameters come from defaults or users command. So this way can easily determine whether the nms iou threshold come from users or defaults confiuration.

src/otx/algorithms/detection/configs/base/data/tiling/efficientnet_iseg_tile_pipeline.py
jaegukhyun: This is for sync with models' parameter. I thought our plan is 
1. Sync model and tiler's nms iou threshold
2. Conduct exp for verifying it
@sovrasov @eugene123tw Feel free to correct me.

src/otx/algorithms/common/configs/training_base.py
sovrasov: The problem that in Geti UI, users would see 0.0 as IoU threshold, which looks like incorrect value.

src/otx/algorithms/detection/configs/base/data/tiling/efficientnet_iseg_tile_pipeline.py
sovrasov: That's right, the only problem is the default value.

eugene123tw:
> Are detectors so sensitive to IoU in NMS? We shouldn't apply any default value tricks, since it all is visible in the UI. My initial thought was to bind 0.5 default value for all detectors and tilers, conduct a couple experiments, and call it a day. Unfortunately, that's a sacrifice we're making because of the imperfect Geti configuration system, which doesn't support per model tuning for default params.
> 
> If that's not possible (the detectors are too sensitive, we could think of changing the tiling threshold only)

@jaegukhyun Based on the experiment: https://intel-my.sharepoint.com/:x:/p/eugene_liu/EWr2Xk17brBDgqZO62qlgSIBNNugJIalJZTWQJYqmaNiVg?e=SgQqpe

Let's change the default NMS IoU threshold to 0.5, what do you think?

src/otx/algorithms/detection/adapters/mmdet/configurer.py
```diff
@@ -61,6 +62,30 @@ def configure_max_num_detections(self, cfg, max_num_detections):
                 train_cfg.rpn_proposal.nms_pre = max_num_detections * 20
                 train_cfg.rpn_proposal.max_per_img = max_num_detections * 10
 
+    def configure_nms_iou_threshold(self, cfg, nms_iou_threshold):
+        """Configure nms iou threshold to user specified value if the object detector uses nms."""
+        if nms_iou_threshold > 0:
```
eugene123tw: Do we need this condition if it's always > 0?

src/otx/algorithms/detection/configs/detection/configuration.yaml
```diff
@@ -294,6 +294,26 @@ postprocessing:
...
+    min_value: 0
+    type: FLOAT
+    ui_rules:
+      action: DISABLE_EDITING
+      operator: AND
+      rules: []
+      type: UI_RULES
+    value: 0.01
+    visible_in_ui: true
+    warning: null
```
eugene123tw: Where do we add the warning for retraining?

eugene123tw (commented):
Only some minor comments

src/otx/algorithms/detection/configs/detection/configuration.yaml
jaegukhyun: I updated that info in description

jaegukhyun: I moved it to warning

src/otx/algorithms/detection/adapters/mmdet/configurer.py
jaegukhyun: Removed!

eugene123tw (dismissed):
Thanks Jaeguk!

sovrasov (dismissed):

sovrasov (approved):

eugene123tw (approved):

PR #3297, opened 2024-04-10
XAI tor tiling: detection, instance segmentation

GalyaZalesskaya:
### Summary

Support merging saliency maps for detection and instance segmentation.

### Checklist

### License

```python
```


src/otx/core/model/detection.py
```diff
@@ -74,26 +75,36 @@ def forward_tiles(self, inputs: TileBatchDetDataEntity) -> DetBatchPredEntity:
...
+            "labels": [pred_entity.labels for pred_entity in pred_entities],
+        }
+        if self.explain_mode:
+            det_batch_pred_entitity_params.update(
+                {
+                    "saliency_map": [pred_entity.saliency_map for pred_entity in pred_entities],
+                    "feature_vector": [pred_entity.feature_vector for pred_entity in pred_entities],
+                },
+            )
+        return DetBatchPredEntity(**det_batch_pred_entitity_params)  # type: ignore[arg-type]
```
GalyaZalesskaya: @sovrasov  What do you think about such way of initializing `DetBatchPredEntity` as it requires a different set of parameters with or without XAI.

I ask because I have an intention to refactor it similarly in `__customize_outputs__`:
https://github.com/openvinotoolkit/training_extensions/blob/37165e0a17d55598c49e15855e75f7768b14fe07/src/otx/core/model/detection.py#L466-L484

GalyaZalesskaya:
@eugene123tw @negvet Can I ask you to look at these changes?

src/otx/core/model/detection.py
sovrasov: I see the problem, but this approach limits capabilities of type checking.
Can we always construct `DetBatchPredEntity` without XAI, and then add XAI if needed?
```python
prediction = DetBatchPredEntity( 
         batch_size=len(predictions), 
         images=inputs.images, 
         imgs_info=inputs.imgs_info, 
         scores=scores, 
         bboxes=bboxes, 
         labels=labels)

if self.explain_mode:
        prediction.feature_vector = ...

return prediction
```

src/otx/core/utils/tile_merge.py
```diff
@@ -120,62 +139,153 @@ def merge(
...
+        map_h, map_w = saliency_maps[0].shape[1:]
+
+        image_h, image_w = shape
+        ratio = map_h / min(image_h, self.tile_size), map_w / min(image_w, self.tile_size)
+
+        image_map_h = int(image_h * ratio[0])
+        image_map_w = int(image_w * ratio[1])
+        merged_map = np.zeros((num_classes, image_map_h, image_map_w))
+
+        for i, saliency_map in enumerate(saliency_maps):
```
sovrasov, line 249: Is that code somehow equivalent to the one we have in MAPI?

src/otx/core/model/detection.py
GalyaZalesskaya: That's a great idea, I'll update the PR

src/otx/core/utils/tile_merge.py
GalyaZalesskaya: Yes, with a few updates because of the difference in tiling implementation between PyTorch  and OV tasks:
1. In OV in MAPI the first tile is an entire image resized to tile size. While in PyTorch we don't have this first tile with the whole image
2. Different format of tile coordinates: x1,y1,x2,y2 in MAPI and x,y,w,h for PyTorch

I wrote that it's basically the same function in function description.
I could optimize the code but decided to keep the visual similarity of implementations in OTX and MAPI

jaegukhyun (dismissed):

src/otx/algo/hooks/recording_forward_hook.py
```diff
@@ -434,7 +434,10 @@ def average_and_normalize(
         Returns:
             np.array: Class-wise Saliency Maps. One saliency map per each class - [class_id, H, W]
         """
-        masks, scores, labels = (pred.masks, pred.scores, pred.labels)
+        if isinstance(pred, dict):
```
vinnamkim: `pred: InstanceData` at L425. However, why is the dictionary `pred` given here?

src/otx/core/model/detection.py
```diff
@@ -74,26 +74,33 @@ def forward_tiles(self, inputs: TileBatchDetDataEntity) -> DetBatchPredEntity:
         tile_attrs: list[list[dict[str, int | str]]] = []
         merger = DetectionTileMerge(
             inputs.imgs_info,
+            self.tile_config.tile_size[0],
```
vinnamkim: `self.tile_config.tile_size` seems a tuple of two integers but I guess that you assume that they are square (`tile_size[0] == tile_size[1]`). How do you guarantee it?

src/otx/core/utils/tile_merge.py
```diff
@@ -31,20 +33,30 @@ class TileMerge(Generic[T_OTXDataEntity, T_OTXBatchPredEntity]):
     def __init__(
         self,
         img_infos: list[ImageInfo],
+        tile_size: int,
+        num_classes: int,
         iou_threshold: float = 0.45,
         max_num_instances: int = 500,
```
vinnamkim:
```suggestion
        num_classes: int,
        tile_config: TileConfig,
```
I'd rather take `tile_config` itself, if many arguments comes from it.

src/otx/algo/hooks/recording_forward_hook.py
GalyaZalesskaya: Thank you, I updated description of the function

src/otx/core/model/detection.py
GalyaZalesskaya: Thank you for pointing that out, I updated the algo to cover the use case when the tile is not square.

src/otx/core/utils/tile_merge.py
GalyaZalesskaya: Thank you, updated

src/otx/core/utils/tile_merge.py
sovrasov: Ok, I'll ask Eugene. It looks like we should adjust the implementation in MAPI

GalyaZalesskaya: Yeah, we talked about it, here is his comment:

`Previously in inference/testing we return the original image as the first tile, this decision is based on current design limitation. That said, this is for PyTorch only. In OpenVINO tiler, it still returns the original image as the first tile. `

Yes, you're right, we should adjust the implementation in MAPI

jaegukhyun (dismissed):

src/otx/core/model/detection.py
```diff
@@ -190,9 +195,17 @@ def __init__(
...
+        inputs: DetBatchDataEntity | TileBatchDetDataEntity,
     ) -> DetBatchPredEntity:
         """Model forward function."""
+        from otx.algo.explain.explain_algo import get_feature_vector
+
+        if isinstance(inputs, OTXTileBatchDataEntity):
+            return self.forward_tiles(inputs)
+
+        self.model.feature_vector_fn = get_feature_vector
+        self.model.explain_fn = self.get_explain_fn()
```
negvet, line 207: Just a comment, not to be done in this PR.

`get_feature_vector` is a function itself
`get_explain_fn()` returns the function
But the naming convention is the same, which is confusing.

What about rename `get_feature_vector` -> `feature_vector_fn`?

GalyaZalesskaya: Will do in the future PR, thanks for pointing out

negvet (approved):

sovrasov (approved):

PR #3344, opened 2024-04-17
Revisit LabelInfo to make it immutable on OTXModel side as possible

vinnamkim:
### Summary
- Ticket no. 138709
- Replace `num_classes` common `OTXModel` argument with `label_info: LabelInfoTypes` (`LabelInfoTypes := int | list[str] | LabelInfo`).
- Merge `HLabel` task's `hlabel_info` into `label_info` as well.
- Do not runtime assignments of `OTXModel.label_info` at `Engine` and `CLI` as possible to make them immutable during the train pipeline. It should be declarative at the very first stage (`OTXModel.__init__(label_info=label_info, ...)`).

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algo/action_classification/movinet.py
+++ b/src/otx/algo/action_classification/movinet.py
@@ -13,6 +13,7 @@ from otx.core.metrics.accuracy import MultiClassClsMetricCallable
 from otx.core.model.action_classification import MMActionCompatibleModel
 from otx.core.model.base import DefaultOptimizerCallable, DefaultSchedulerCallable
 from otx.core.schedulers import LRSchedulerListCallable
+from otx.core.types.label import LabelInfoTypes
 
 if TYPE_CHECKING:
     from lightning.pytorch.cli import LRSchedulerCallable, OptimizerCallable
@@ -25,7 +26,7 @@ class MoViNet(MMActionCompatibleModel):
 
     def __init__(
         self,
-        num_classes: int,
+        label_info: LabelInfoTypes,
         optimizer: OptimizerCallable = DefaultOptimizerCallable,
         scheduler: LRSchedulerCallable | LRSchedulerListCallable = DefaultSchedulerCallable,
         metric: MetricCallable = MultiClassClsMetricCallable,
@@ -33,7 +34,7 @@ class MoViNet(MMActionCompatibleModel):
     ) -> None:
         config = read_mmconfig("movinet")
         super().__init__(
-            num_classes=num_classes,
+            label_info=label_info,
             config=config,
             optimizer=optimizer,
             scheduler=scheduler,
--- a/src/otx/core/model/base.py
+++ b/src/otx/core/model/base.py
@@ -13,6 +13,7 @@ import json
 import logging
 import warnings
 from abc import abstractmethod
+from collections.abc import Sequence
 from typing import TYPE_CHECKING, Any, Callable, Generic, Literal, NamedTuple
 
 import numpy as np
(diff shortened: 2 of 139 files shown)


src/otx/algo/action_classification/movinet.py
```diff
@@ -25,15 +26,15 @@ class MoViNet(MMActionCompatibleModel):
 
     def __init__(
         self,
-        num_classes: int,
+        label_info: LabelInfoTypes,
```
harimkang, line 29: I'm not sure if this is common, I think `num_classes` is easy to understand and common from a user perspective, but `label_info`, just by looking at the words, I have no idea what to put in it.
@samet-akcay What do you think?

samet-akcay: If `label_info` is used to access to some other attributes, it's fine to use it. If it is only passed to access `num_classes`, then it does not make sense to replace `num_classes` with `label_info`. 

`num_classes` is more explicit and clearly shows what it is. As a user, I would not know what `label_info` contains, and need to go to the class definition to look at the details. 

Another comment, maybe not related to this PR,  but `LabelInfoTypes` looks too verbose. Ideally it could be just `LabelInfo`

Hope that helps. Thanks

vinnamkim: Our expectation for users is to construct the `label_info` object correctly and provide it during model instantiation. This is crucial because we intend for `OTXModel` not to keep anonymity on the model prediction. It's not recommended to simply pass an integer, for example, `OTXModel(num_classes=10)`. This is because of the anonymity I mentioned before. There is no information that the predicted class number is what exactly it is in realty.

However, it should be considered that creating `label_info` from scratch might be cumbersome for users. This object essentially encapsulates dataset characteristics and should originate from the dataset itself. Therefore, the recommended user workflow involves: `data_root -> OTXDataModule -> OTXDataModule.label_info -> OTXModel(label_info)`. I believe there's potential to design a more user-friendly interface to streamline this process further, possibly abbreviating it further to `data_root -> OTXModel(label_info)`.

harimkang: Yeah, I don't disagree with that, and we can think about it more in the future to make it more user-friendly.
Thank you for answers!

harimkang (approved):

src/otx/core/types/label.py
```diff
@@ -41,9 +48,12 @@ def from_num_classes(cls, num_classes: int) -> LabelInfo:
                 label_groups=[["label_0", ...]]
             )
         """
+        if num_classes < 0:
```
eunwoosh, line 51:
```suggestion
        if num_classes <= 0:
```
Is it intended to set `label_names` as [] if num_classes is 0 intead of returning `NullLabelInfo()`?

src/otx/core/model/base.py
```diff
@@ -123,7 +124,7 @@ def __init__(
 
         # this line allows to access init params with 'self.hparams' attribute
         # also ensures init params will be stored in ckpt
-        self.save_hyperparameters(logger=False, ignore=["model", "optimizer", "scheduler", "metric"])
+        self.save_hyperparameters(logger=False, ignore=["optimizer", "scheduler", "metric"])
```
eunwoosh, line 127: It's just a question. Why is 'model' removed?

vinnamkim: It is a kind of imprint remaining from `OTXLitModule`. `OTXLitModule.__init__()` had `model` argument

src/otx/core/types/label.py
vinnamkim: Because this PR is huge and effective everywhere, I don't want to take charge of any possibility to break the test success if it is not a severe issue. Let's revisit it after this PR.

eunwoosh (approved):

src/otx/algo/action_classification/movinet.py
samet-akcay: > However, it should be considered that creating label_info from scratch might be cumbersome for users.

That's my main concern. As a user, I would find it annoying to create another object in addition to datamodule or model. Datamodule should handle this under the hood, and the user should not care about creating another object.

PR #3366, opened 2024-04-19
Enable SSD export through NativeExporter

jaegukhyun:
### Checklist

### License

```python
```


src/otx/algo/detection/heads/base_head.py
```diff
@@ -468,3 +469,210 @@ def _bbox_post_process(
             results = results[: cfg.max_per_img]
 
         return results
+
+    def export(
+        self,
+        x: tuple[Tensor],
+        batch_data_samples: list[InstanceData],
+        rescale: bool = False,
+    ) -> list[InstanceData]:
```
eugene123tw, line 478: Does this meab that for each algo, we must now implement multiple `def export` to use NaiveExporter? Could we instead integrate the export functionality within `def predict` and `def predict_by_feat` via branching with `torch.onnx.is_in_onnx_export()`? Something like
```py
def forward(...):
  if torch.onnx.is_in_onnx_export():
     return predict_by_feat_onnx_export(...)
  return predict_by_feat(...)

def predict_by_feat_onnx_export(...):
   """Predict by feat from MMDeploy"""

def predict_by_feat(...):
   """Predict by feat from MMDet"""
```

src/otx/algo/detection/ssd.py
```diff
@@ -572,21 +601,42 @@ def _exporter(self) -> OTXModelExporter:
...
+                    "image": {0: "batch", 2: "height", 3: "width"},
+                    "boxes": {0: "batch", 1: "num_dets"},
+                    "labels": {0: "batch", 1: "num_dets"},
+                },
+                "autograd_inlining": False,
+            },
+            output_names=["feature_vector", "saliency_map"] if self.explain_mode else None,
+        )
+
+    def forward_for_tracing(self, inputs: Tensor) -> list[InstanceData]:
```
eugene123tw, line 607: I didn't see any usage of `forward_for_tracing` in the PR. Also, I couldn't locate it within OTXNativeModelExporter either. Could you provide more details?

src/otx/algo/detection/ssd.py
```diff
@@ -572,21 +601,42 @@ def _exporter(self) -> OTXModelExporter:
...
+            std=std,
+            resize_mode="standard",
+            pad_value=0,
+            swap_rgb=False,
+            via_onnx=True,  # Currently SSD should be exported through ONNX
+            onnx_export_configuration={
+                "input_names": ["image"],
+                "output_names": ["boxes", "labels"],
+                "dynamic_axes": {
+                    "image": {0: "batch", 2: "height", 3: "width"},
```
sovrasov, line 598: Why do we need dynamic image size here? In MAPI we automatically insert resize graph node, rather than making model recappable.

src/otx/algo/detection/ops/nms.py
```diff
@@ -219,3 +221,220 @@ def batched_nms(
...
+    output_index: bool = False,
+) -> tuple[torch.Tensor, torch.Tensor] | tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+    """Create a dummy onnx::NonMaxSuppression op while exporting to ONNX.
+
+    This function helps exporting to onnx with batch and multiclass NMS op. It
+    only supports class-agnostic detection results. That is, the scores is of
+    shape (N, num_bboxes, num_classes) and the boxes is of shape (N, num_boxes,
+    4).
+    """
+    if version.parse(torch.__version__) < version.parse("1.13.0"):
```
sungchul1: We don't need this condition because OTX 2.x supports torch >= 2.0.

src/otx/algo/detection/ops/nms.py
```diff
@@ -219,3 +221,220 @@ def batched_nms(
...
+                boxes per class of nms.
+            iou_threshold (float): IOU threshold of nms.
+            score_threshold (float): score threshold of nms.
+
+        Returns:
+            Tensor: Selected indices of boxes. 2-D tensor of shape
+            (num_selected_indices, 3) with each row of
+            [batch_index, class_index, box_index].
+        """
+        from mmcv.ops import nms
```
sungchul1: Isn't getting nms in this PR's scope?

src/otx/algo/detection/heads/base_head.py
jaegukhyun: I admit that implementing every export function requires lots of engineering cost. I think we should re-visit this after migrating all other detectors. Currently I add these export function to compare predict procedure more explicitly. And this approach can protect predict function so that we can assure backward compatibility between PRs.
In addition if we use torch.onnx.is_in_onnx_export() branch, then direct export through ov, not using onnx, cannot enter that branch.

src/otx/algo/detection/ssd.py
jaegukhyun: I don't search for the details of forward_for_tracing, but I believe this is the forward path definition when the models are exported

src/otx/algo/detection/ssd.py
jaegukhyun: Anchor generator generates default anchor from feature map size, and the feature map size come from image size.
If we don't specify image size as a dynamic, then ONNX convert anchor generator to generate anchors from constant values. As a result, the graph between anchor generator and object detector are broken. So we need to specify this for ONNX conversion.

src/otx/algo/detection/ops/nms.py
jaegukhyun: Nice catch! Removed by [3f62054](https://github.com/openvinotoolkit/training_extensions/pull/3366/commits/3f62054242114aea15660194ea6790d5f68aa851)

src/otx/algo/detection/ops/nms.py
jaegukhyun: Yes it is. Changed by [3f62054](https://github.com/openvinotoolkit/training_extensions/pull/3366/commits/3f62054242114aea15660194ea6790d5f68aa851)

sungchul1 (dismissed):

src/otx/algo/detection/ssd.py
eugene123tw: Thank you, Jaeguk. I've noticed a monkey patch beneath OTXModel.export.
```
self.forward = self.forward_for_tracing  # type: ignore[method-assign, assignment]
return self._exporter.export(
    self,
    output_dir,
    base_name,
    export_format,
    precision,
)
```
Since there's no documentation available, I found myself a bit confused regarding the purpose of this method. 
Besides, I'm concerned about the potential risks associated with such monkey patching, especially without any warnings or documentation. While we may not have a better solution, it would be good to mark this with a TODO and prioritize improving the documentation before resorting to monkey-patching export functions, if possible of course.

eugene123tw (approved):

eunwoosh (approved):

src/otx/algo/detection/ssd.py
jaegukhyun: @vinnamkim Do you have any opinion for this?

PR #3376, opened 2024-04-21
Decouple semantic segmentation models

kprokofi:
### Summary

Eliminate mmseg dependency for models, decouple all models, introduce updated model structure, added support for torchvision data pipeline

Unit tests will be added with next PRs.

### Checklist

### License

```python
```


kprokofi:
Strange error on CI with CUDA
image

Locally, all unit tests are passed with tox
image

src/otx/algo/segmentation/heads/custom_fcn_head.py
jaegukhyun, line 1: I think now we can change name to fcn_head.py. What do you think?

src/otx/algo/segmentation/heads/custom_ham_head.py
jaegukhyun, line 1: Also this file could change name

src/otx/algo/segmentation/losses/cross_entropy_loss_with_ignore.py
jaegukhyun, line 1: Just for a note, many of modules in this file are duplicated with detection I think. Let's merge this after decoupling all tasks.

jaegukhyun (commented):
Overall LGTM.
Could you share some regression tests between decoupling?
In some case torchvision data pipeline slower than mm's pipeline since mm use opencv and it is faster than torch tensor ops.
So we need regression tests results for accuracy and e2e time.

jaegukhyun:
And it seems unit tests run on cpu machine. So I think you need to fix unit tests failure on cpu machine.

src/otx/algo/modules/conv_module.py
```diff
@@ -355,3 +355,102 @@ def create_from_conv_bn(
         self.turn_on_efficient_conv_bn_eval(efficient_conv_bn_eval)
 
         return self
+
+
+class DepthwiseSeparableConvModule(nn.Module):
```
sungchul1: It is already implemented in develop. 
https://github.com/openvinotoolkit/training_extensions/blob/3d6bbdf673c1d1072dc9813c48cdcc3a241c976d/src/otx/algo/modules/depthwise_separable_conv_module.py#L14
Please remove it or move the previously implemented one to here.

src/otx/algo/segmentation/base_model.py
```diff
@@ -0,0 +1,124 @@
...
+import torch.nn.functional as f
+from torch import Tensor, nn
+
+from otx.algo.segmentation.losses import create_criterion
+
+if TYPE_CHECKING:
+    from otx.core.data.entity.base import ImageInfo
+
+
+class BaseSegmNNModel(nn.Module):
```
sungchul1: What is `NN`? Is it from `nn.Module`? If so, why isn't it in `BaseSegmHead`?

src/otx/algo/segmentation/heads/custom_fcn_head.py
```diff
@@ -5,98 +5,130 @@
...
+        return self.cls_seg(output)
 
 
-@MODELS.register_module()
-class CustomFCNHead(ClassIncrementalMixin, FCNHead):
+class CustomFCNHead(FCNHead):
     """Custom FCNHead implementation.
 
-    Custom FCNHead supports ignored label for class incremental learning cases.
+    Custom FCNHead supports iterative agregator and ignored label for class
```
sungchul1:
```suggestion
    Custom FCNHead supports iterative aggregator and ignored label for class
```

src/otx/algo/segmentation/heads/custom_fcn_head.py
```diff
@@ -5,98 +5,130 @@
...
-        return valid_label_mask
+    def forward(self, inputs: Tensor) -> Tensor:
+        """Forward function."""
+        output = self._forward_feature(inputs)
+        return self.cls_seg(output)
 
 
-@MODELS.register_module()
-class CustomFCNHead(ClassIncrementalMixin, FCNHead):
+class CustomFCNHead(FCNHead):
```
sungchul1: Now, it seems to need to change this name to reveal more of its characteristics.

src/otx/algo/segmentation/litehrnet.py
```diff
@@ -5,66 +5,530 @@
...
-        super().__init__(
-            label_info=label_info,
-            config=config,
-            optimizer=optimizer,
-            scheduler=scheduler,
-            metric=metric,
-            torch_compile=torch_compile,
-        )
+    @property
+    def ignore_scope(self) -> dict[str, str | dict[str, list[str]]]:
```
sungchul1, line 73: Isn't there any way not to use these export parameters in the source code?
(I don't know the history of why it came here. If it was already agreed with others, feel free to ignore it.)

src/otx/algo/segmentation/modules/iterator.py
sungchul1, line 1: `aggregators.py` looks better than `iterator.py`.

src/otx/recipe/_base_/data/mmseg_base.yaml
```diff
@@ -11,68 +11,70 @@ config:
...
-      - type: PhotoMetricDistortion
-      - type: Pad
-        size:
-          - 512
-          - 512
-        pad_val:
-          img: 0
-          seg: 255
-      - type: PackSegInputs
+      - class_path: torchvision.transforms.v2.RandomResizedCrop
```
sungchul1, line 16: It is a known issue that using torchvision operations shows slow performance.
What about replacing them with mmdet transforms implemented in torchvision pipeline like detection?

src/otx/algo/segmentation/segnext.py
```diff
@@ -4,64 +4,124 @@
...
 
 if TYPE_CHECKING:
-    from lightning.pytorch.cli import LRSchedulerCallable, OptimizerCallable
+    from torch import nn
+
+
+class SegNextB(BaseSegmNNModel):
+    """SegNextB Model."""
+
+    default_backbone_configuration: ClassVar[dict[str, Any]] = {
```
sungchul1, line 23: Is it required? They are already in the recipes.

kprokofi: It is more for API usage. If we don't add these defaults - user should pass this entirely to initialize the class. With this addition, we can just create model passing only base_model_name. Moreover, we can pass only several parameters via dict (for backbone or head) and it will override defaults. I thought it can be useful.

src/otx/recipe/_base_/data/mmseg_base.yaml
kprokofi: Do we have it implemented in OTX? I will take a look

harimkang: https://github.com/openvinotoolkit/training_extensions/blob/3d6bbdf673c1d1072dc9813c48cdcc3a241c976d/src/otx/core/data/transform_libs/torchvision.py#L549C7-L549C24
However, this approach still has the issue of being slow. See also https://github.com/openvinotoolkit/training_extensions/issues/3379

src/otx/algo/segmentation/litehrnet.py
kprokofi: Not sure what you mean, but ignore scope in such representation is required for LiteHRNet unfortunately, we can also add entire backbone to ignored scope, but it doesn't make sense to compress this model. It is from NNCF team

src/otx/algo/segmentation/heads/custom_fcn_head.py
kprokofi: OK

src/otx/algo/segmentation/base_model.py
kprokofi: I will remove NN

src/otx/recipe/_base_/data/mmseg_base.yaml
sungchul1: Yeah, conversion parts in the pipeline will be fixed or the whole pipeline can be updated by using other transform library.
But for now, please refer previously implemented ones.

src/otx/algo/segmentation/segnext.py
sungchul1: I agree with your intention. I have considered how to use it more cleanly, but I have no idea :)

src/otx/algo/segmentation/modules/iterator.py
kprokofi: done https://github.com/openvinotoolkit/training_extensions/pull/3376/commits/04c51228f2b7e7f32812e63920e6ee011a199fad

src/otx/algo/segmentation/heads/custom_fcn_head.py
kprokofi: done https://github.com/openvinotoolkit/training_extensions/pull/3376/commits/04c51228f2b7e7f32812e63920e6ee011a199fad

src/otx/algo/segmentation/heads/custom_fcn_head.py
kprokofi: removed https://github.com/openvinotoolkit/training_extensions/pull/3376/commits/04c51228f2b7e7f32812e63920e6ee011a199fad

src/otx/algo/segmentation/base_model.py
kprokofi: done https://github.com/openvinotoolkit/training_extensions/pull/3376/commits/04c51228f2b7e7f32812e63920e6ee011a199fad

src/otx/algo/modules/conv_module.py
kprokofi: deleted https://github.com/openvinotoolkit/training_extensions/pull/3376/commits/04c51228f2b7e7f32812e63920e6ee011a199fad

src/otx/algo/segmentation/backbones/mscan.py
```diff
@@ -0,0 +1,473 @@
...
+
+        Args:
+            drop_prob (float): Probability of the path to be zeroed. Default: 0.1
+        """
+        super().__init__()
+        self.drop_prob = drop_prob
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        """Forward function."""
+        return drop_path(x, self.drop_prob, self.training)
```
eugene123tw, line 53: Should we keep DropPath? It appears redundant alongside timm models (an existing dependency). Can we substitute it with timm.models.layers.DropPath for consistency?

kprokofi: We should revisit everything what can be deleted / omitted. But maybe it is better to do after FF and basic functionality. Since we have a tight schedule...

src/otx/recipe/_base_/data/mmseg_base.yaml
kprokofi: We don't support masks here unfortunately

kprokofi: I added support for masks resizing, but I don't see any difference in speed (train/data_time and iter_time is the same during training), but I see drop in accuracy and worse convergence. Let's decide on our augmentations backend in next PRs. For now, I would like to leave torchvisionv2 for segmentation.

kprokofi:
Tests are passed. I propose to merge this PR, since today is FF. 
Please, feel free to leave comments to address in the following PRs before code freeze.

I compared algorithms on KITTI dataset.

iter_time is  the same, data_train_time is worse on 0.002 sec.,  due to difference in resizing and augmentations, we have changed accuracy values about 2-4 % DICE score on this dataset. We need to align augmentations with mm*

harimkang (approved):

src/otx/recipe/_base_/data/mmseg_base.yaml
sungchul1: It is fine if there is little difference between mmdet and torcivision on semantic segmentation.
Let's discuss the future plan later as you said.

src/otx/algo/segmentation/litehrnet.py
sungchul1: I know why the `ignore_scope` is needed for export, but I'm not sure if they must be in here.
I think it would be better to separate it from the model and save it separately in a location related to the recipe (`openvino_model.yaml`?) like 1.x.
But I think my suggestion is not the scope for this PR, please consider it later.

sungchul1 (approved):

PR #3334, opened 2024-04-16
Add SchedulerCallableSupportHPO

vinnamkim:
### Summary

- Ticket no. 137956
- As same as `OptimizerCallableSupportHPO` added in #3258, the scheduler callable should also be picklable and allow access on its parameters.

### How to test
Added unit tests for this change as well.

### Checklist

- [ ] I have added unit tests to cover my changes.​
- [ ] I have added integration tests to cover my changes.​
- [ ] I have added e2e tests for validation.
- [ ] I have added the description of my changes into CHANGELOG in my target branch (e.g., [CHANGELOG](https://github.com/openvinotoolkit/training_extensions/blob/develop/CHANGELOG.md) in develop).​
- [ ] I have updated the documentation in my target branch accordingly (e.g., [documentation](https://github.com/openvinotoolkit/training_extensions/tree/develop/docs) in develop).
- [ ] I have [linked related issues](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

### License

- [ ] I submit _my code changes_ under the same [Apache License](https://github.com/openvinotoolkit/training_extensions/blob/develop/LICENSE) that covers the project.
      Feel free to contact the maintainers if that's a concern.
- [ ] I have updated the license header for each file (see an example below).

```python
# Copyright (C) 2024 Intel Corporation
# SPDX-License-Identifier: Apache-2.0
```

PR #3378, opened 2024-04-22
Add adaptive batch size feature

eunwoosh:
### Summary
This PR adds adaptive batch size feature supporting both GPU and XPU.
As same as OTX 1.X, this feature supports "Safe" mode and "Full" mode.
Safe mode checks current batch size fits to memory size and decrease it if not.
Full mode finds a batch size which uses almost memory size.

image

All integration tests on XPU are passed.

### How to test
  otx train --task ...   --data_root ...   --adaptive_bs Safe
  otx train --task ...   --data_root ...   --adaptive_bs Full

### Checklist

### License

```python
```


src/otx/core/model/base.py
```diff
@@ -38,12 +38,12 @@
 from otx.core.data.entity.tile import OTXTileBatchDataEntity
 from otx.core.exporter.native import OTXNativeModelExporter
 from otx.core.metrics import MetricInput, NullMetricCallable
-from otx.core.optimizer.callable import OptimizerCallableSupportHPO
+from otx.core.optimizer.callable import OptimizerCallableSupportPickle
```
vinnamkim: Why do you change this naming?

eunwoosh: It's because Adaptive batch size feature also uses this class because it use almost same method as HPO when trying batch size.

vinnamkim: See https://github.com/openvinotoolkit/training_extensions/pull/3334#discussion_r1568115629. It is not just for pickling.

eunwoosh: Then how about OptimizerCallableSupportHPTuning? I think it can include HPO and adaptive batch size.

vinnamkim: I don't know why the renaming is necessary

eunwoosh: so you think OptimizerCallableSupportHPO also can include meaning of adaptive batch size, right?

vinnamkim: Why not? Is batch size not a hyperparameter?

eunwoosh: ok I'll revert it

eunwoosh: done

src/otx/engine/engine.py
```diff
@@ -240,6 +245,9 @@ def train(
         """
         checkpoint = checkpoint if checkpoint is not None else self.checkpoint
 
+        if adaptive_bs != "None":
+            adapt_batch_size(engine=self, **locals(), not_increase=(adaptive_bs != "Full"))
```
harimkang, line 249: If `adaptive_bs` and `run_hpo` are running at the same time, will this cause any problems or do we need to prevent this?

harimkang (dismissed):
Thanks! i left minor comment

src/otx/engine/engine.py
eunwoosh: HPO is turned off automatically during adaptive batch size and adaptive batch size is turned of during HPO.

src/otx/engine/adaptive_bs/__init__.py
```diff
@@ -0,0 +1,9 @@
+"""API for adaptive batch size."""
+
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
```
sungchul1: AFAIK, the order is license - docstring for 2.x, isn't it?

src/otx/engine/adaptive_bs/adaptive_bs_api.py
```diff
@@ -0,0 +1,174 @@
...
+    else:
+        logger.info("Adapting batch size is done. Batch size isn't changed.")
+
+
+def _adjust_train_args(train_args: dict[str, Any]) -> dict[str, Any]:
+    train_args.update(train_args.pop("kwargs", {}))
+    train_args.pop("self", None)
+    train_args.pop("run_hpo", None)
+    train_args.pop("adaptive_bs")
+    return train_args
```
sungchul1, line 96: Because of `**locals()` in `engine.train()`, this function seems required.
Can we use only necessary arguments instead of `**locals()`?

src/otx/engine/adaptive_bs/__init__.py
eunwoosh: ah, I'll align it thanks

eunwoosh: done https://github.com/openvinotoolkit/training_extensions/pull/3378/commits/11c3e96673a65016dca411f09ab416a73eebdce6

src/otx/engine/adaptive_bs/adaptive_bs_api.py
eunwoosh: it can, but IMHO, `train_args` is more readable because developer doesn't have to consider which arguments are only for `adapt_batch_size`. And also everytime argument of engine.train is changed, argument of `adapt_batch_size` also needs to be changed if it's changed. For these reasons, I think it's better to leave it as it is. How do you think about?

sungchul1: I agree with your thought that users don't have to handle those args that can be dynamically changed.
But I'm just wondering if there is another way to except those args more cleanly.
Leave it for now but please consider it.

sungchul1 (approved):

harimkang (approved):

Issue #3379, opened 2024-04-22
Slowness issues with np array based transform (otx.core.data.transform_libs.torchvision.Resize)


harimkang:
**Describe the bug**

Despite importing MM's implementation into transform for resolving slowness issue (https://github.com/openvinotoolkit/training_extensions/pull/3253), 
the inefficient process of `np array -> tv_image -> np array` results in the following differences.

<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=en-US style='font-family:Calibri;font-size:11.0pt'>

<div style='direction:ltr'>

MULTI_CLASS_CLS | val/iter_time
-- | --
mm.Resize (np.array) | 0.174
Torchvision.v2.Resize (tv.Image) | 0.240
otx.Resize   (tv_image -> np.array) | 0.231
otx.Resize   without converting to tv.Image (np.array) | 0.173

</div>

</body>

</html>

torchvision transfrom always use tv_image: https://github.com/openvinotoolkit/training_extensions/blob/03884711f2aa1ac02512ad6f6beb0a721366a4c8/src/otx/core/data/dataset/base.py#L96

But, New `otx.Resize` based np.array: https://github.com/openvinotoolkit/training_extensions/blob/03884711f2aa1ac02512ad6f6beb0a721366a4c8/src/otx/core/data/transform_libs/torchvision.py#L494
https://github.com/openvinotoolkit/training_extensions/blob/03884711f2aa1ac02512ad6f6beb0a721366a4c8/src/otx/core/data/transform_libs/torchvision.py#L504

In my personal opinion, the best thing to do would be to improve torchvision.v2.Resize based on tv Image.
But I'm not sure how to do that right now

**Steps to Reproduce**

1.
2.
3.
4.

**Environment:**

- OS: 
- Framework version: 
- Python version:
- OpenVINO version:
- CUDA/cuDNN version:
- GPU model and memory:

sungchul1:
Currently, `otx.Resize` supports only numpy image because it uses `cv2.resize`.
[...]

PR #3380, opened 2024-04-22
Decouple mmpretrain: Efficientnet-B0

harimkang:
### Summary

Enable Efficientnet-B0 with torchvision pipeline

### Checklist

### License

```python
```


src/otx/core/data/transform_libs/torchvision.py
```diff
@@ -160,6 +160,19 @@ def _get_preprocess_shape(self, oldh: int, oldw: int, long_side_length: int) ->
         return (newh, neww)
 
 
+class ResizetoShortEdge(ResizetoLongestEdge):
```
sungchul1: It's parent class `ResizetoLongestEdge` which was previously implemented for visual prompting is dependent on torchvision ops, but torchvision ops showed slow performance.
I think `ResizetoLongestEdge` should be deprecated and using `Resize` implemented with mmdet components looks better. (will be updated)
And this `ResizetoShortEdge` should be also updated along with `Resize`.
`Resize` already supports `keep_ratio` and this option works the same with `ResizetoLongestEdge`.
What do you think?

src/otx/algo/classification/classifier/base_classifier.py
```diff
@@ -220,10 +220,10 @@ def _forward_explain(self, images: torch.Tensor) -> dict[str, torch.Tensor | lis
 
         return {
             "logits": logits,
-            "preds": preds,
-            "scores": scores,
-            "saliency_map": saliency_map,
             "feature_vector": feature_vector,
+            "saliency_map": saliency_map,
```
sungchul1, line 224: This change seems meaningless.

harimkang: The model API is affected by the order of the corresponding dict output keys.
-> This resolve https://jira.devtools.intel.com/browse/CVS-138604

jaegukhyun (dismissed):

src/otx/core/data/transform_libs/torchvision.py
harimkang: Yes, I agree.
However, `Resize` imported into OTX still has the issue of being slow on large data. - https://github.com/openvinotoolkit/training_extensions/issues/3379

This looks like it needs to be optimized and unnecessary classes removed.
Let me check this with `Resize (keep_ratio=True)` and compare the two scores.

harimkang: Thanks - Done https://github.com/openvinotoolkit/training_extensions/pull/3380/commits/13a3491503e9984ad843c1ce08649601c64da06c

sungchul1: When resizing an image with (3, 5) to (10, 10), 
`Resize (keep_ratio=True)` and `ResizetoLongestEdge` resize it to (6, 10) based on the longest edge.
`ResizetoShortEdge` resizes it to (10, 16) based on the shortes edge.
It is difference, please refer it.

goodsong81: Leaving a comment while just passing by :p 
It's very unusual where inheritance is used in this way; opposite behavior! Maybe due to code reuse, right?
It might be better to have common ResizeToEdge and have 2 children or option like `longest=True`.
Just a nitpick. Feel free to ignore.

harimkang: For now, let's apply the validation pipeline to Resize only (same with test pipeline), and I'll revisit this soon.

jaegukhyun (approved):

sungchul1 (approved):

sungchul1 (approved):

PR #3407, opened 2024-04-26
Decouple YOLOX : export

sungchul1:
### Summary

This PR includes:
- [x] Change `MMdeployExporter` to `OTXNativeModelExporter`
- [x] Add `dynamic_topk` for #preds < k

### Checklist

### License

```python
```


src/otx/algo/detection/heads/base_head.py
```diff
@@ -666,9 +666,9 @@ def _export_by_feat_single(
         return multiclass_nms(
             bboxes,
             scores,
-            cfg.max_per_img,
+            max_output_boxes_per_class=200,
```
jaegukhyun: Could you explain why this value should be static?

sungchul1: `cfg.max_per_img`
https://github.com/openvinotoolkit/training_extensions/blob/ff2cf14131184f2518146876ace2c990e0e81971/src/otx/algo/detection/mmconfigs/ssd_mobilenetv2.yaml#L22

`max_output_boxes_per_class`
https://github.com/openvinotoolkit/training_extensions/blob/ff2cf14131184f2518146876ace2c990e0e81971/src/otx/algo/detection/mmdeploy/base_detection.py#L44

Both seem different, doesn't it?
The former is from `mmconfigs` and the latter is from `deploy_cfg`.
https://github.com/open-mmlab/mmdeploy/blob/v1.3.1/mmdeploy/codebase/mmdet/models/dense_heads/base_dense_head.py#L191-L206

jaegukhyun: In this case we should follow value of model config. Since the trained model use value from model config.

sungchul1: I just follwed mm's default setting and I thought it was typo because yolox has different values between training and exporting.
https://github.com/openvinotoolkit/training_extensions/blob/ff2cf14131184f2518146876ace2c990e0e81971/src/otx/algo/detection/mmconfigs/yolox_tiny.yaml#L11
I'll revert it for SSD and ATSS, but I think yolox's one should remain for now.

sungchul1: https://github.com/openvinotoolkit/training_extensions/pull/3407/commits/2bd356dbbfb8a706ce524745912d901197b3b9a0

jaegukhyun (approved):

harimkang (approved):

PR #3408, opened 2024-04-26
Refine classification OV export

sovrasov:
### Summary
This PR removes unnecessary unnamed outputs in classification models:
image

Also,
-  output_names argument behavior was fixed for OV export in the Native exporter.
- recover LiteHRNet export parameters missing after the mmseg decoupling

### Checklist

### License

```python
```


harimkang (dismissed):

sovrasov:
Merge after https://github.com/openvinotoolkit/training_extensions/pull/3410

jaegukhyun (dismissed):

src/otx/core/exporter/base.py
```diff
@@ -280,30 +280,17 @@ def _postprocess_openvino_model(self, exported_model: openvino.Model) -> openvin
...
-
-                if not found:
-                    msg = (
-                        "Given output name to export is not in the traced_outputs, "
-                        f"{output_name} not in {traced_outputs}"
-                    )
-                    raise RuntimeError(msg)
+            if len(exported_model.outputs) >= len(self.output_names):
+                for i, name in enumerate(self.output_names):
+                    exported_model.outputs[i].tensor.set_names({name})
```
vinnamkim, line 312: How do you guarantee that `exported_model.outputs[:len(self.output_names)]` always indicate `self.output_names` and their orders is same?

src/otx/algo/classification/efficientnet.py
```diff
@@ -131,7 +131,7 @@ def _exporter(self) -> OTXModelExporter:
             resize_mode="standard",
             pad_value=0,
             swap_rgb=False,
-            via_onnx=True,  # NOTE: This should be done via onnx
+            via_onnx=False,
```
wonjuleee, line 134: Could you give more explanations why direct export from openvino has been fine with non-tracing exception?

sovrasov: I haven't faced any issues with these particular models, and I don't know why this flag was enabled here. If you have any problems with OVC on basic models, it should be reported and fixed in the next OV release.

src/otx/core/exporter/base.py
sovrasov: Unfortunately, OVC API docs do not describe any properties of the output model, so formally nothing (including naming is guaranteed). I believe that happened because of multiple backends supported. From the past experience, output order is one of a few details that is preserved. Also, I additionally clarified that with OVC owner, he confirmed that the output order is guaranteed.

vinnamkim: The inconsistency between them can also be happened by human errors. For example, the model implementation produces the output sinks in this order: `output1`, `output2`, and `output3`, but given `output_names` in the exporter can be `["output2", "output1", "output3"]`. In this case, we do not have any exception raised, then the real problem is hidden and could not be found easily.

harimkang: I have encountered an error related to this - https://jira.devtools.intel.com/browse/CVS-138604.
I found that it is simply a difference in the order of the output.
We need to clean this up. As an external user, if it doesn't work just because the order of this is different, it's hard to find the problem with this. It seems like we need to write more documentation for now.

sovrasov: Unfortunately, we don't have a fix for that on torch side as well. If we export a model via ONNX or to ONNX, incorrect order will be propagated to OV model, and the cleaning process we had before, can not catch that as well, because it has no cue about the correct output order. Here, I'm just aligning the behavior to avoid incorrect results when `via_onnx==False`.

From the top of my head, I can only propose to preprocess the torch model itself to avoid this, but we need an extra inference and a wrapper to reorder the outputs if required. Also, in the docs we can mention that the order is important.

vinnamkim: > Unfortunately, we don't have a fix for that on torch side as well. If we export a model via ONNX or to ONNX, incorrect order will be propagated to OV model, and the cleaning process we had before, can not catch that as well, because it has no cue about the correct output order. Here, I'm just aligning the behavior to avoid incorrect results when `via_onnx==False`.
> 
> From the top of my head, I can only propose to preprocess the torch model itself to avoid this, but we need an extra inference and a wrapper to reorder the outputs if required. Also, in the docs we can mention that the order is important.

Sorry, I don't understand what you mean. To me, your current implementation is hiding a restriction that the model output sink order and names should be same with `output_names`. In addition, even if the wrong `output_names` is given, it does not raise any exception and make any malfunction afterward hard to trace. However, the previous implementation at least tries to raise an error if not acceptable input is given.

What's the reason that this implementation is more acceptable than before?

sovrasov: By design output naming feature doesn't suppose the raw model outputs have names. The exporter tries to make no difference between ONNX and OV, since we should support both. ONNX doesn't support propagating output dict names, while OV does this (not perfectly though). The previous implementation was just checking if the raw model outputs correspond to the parameter passed to the exporter, and clean up the mess introduced by OVC where possible. Now the implementation just assigns the names. I assume we can do that in OV since we have to provide correct output names for ONNX model anyway.
At the same time I see your point, I'll add an extra check to OV implementation to produce warnings if the output names are inconsistent vs ones in the output dict.

harimkang (approved):

jaegukhyun (approved):

PR #3413, opened 2024-04-29
Change location of stack_batch

jaegukhyun:
### Summary
Stack batch is function for stack list of Tensors whose sizes are different.
This function is useful when the batch of Tensors has different shapes with each other.
Normal torch.stack will raise shape error, but stack batch pad to match shapes of batch to largest tensor of batch.
Stack batch requires pad_size_divisor and pad_val, but previous implementation can't configure that values.
This PR changes the location of stack batch so that stack batch can get those value from model directly. 

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algo/detection/yolox.py
+++ b/src/otx/algo/detection/yolox.py
@@ -17,11 +17,11 @@ from otx.algo.detection.heads.yolox_head import YOLOXHead
 from otx.algo.detection.necks.yolox_pafpn import YOLOXPAFPN
 from otx.algo.detection.ssd import SingleStageDetector
 from otx.algo.utils.mmconfig import read_mmconfig
-from otx.algo.utils.mmengine_utils import stack_batch
 from otx.algo.utils.support_otx_v1 import OTXv1Helper
 from otx.core.config.data import TileConfig
 from otx.core.data.entity.base import OTXBatchLossEntity
 from otx.core.data.entity.detection import DetBatchDataEntity, DetBatchPredEntity
+from otx.core.data.entity.utils import stack_batch
 from otx.core.exporter.base import OTXModelExporter
 from otx.core.exporter.mmdeploy import MMdeployExporter
 from otx.core.metrics.mean_ap import MeanAPCallable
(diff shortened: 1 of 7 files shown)


harimkang (approved):

src/otx/algo/detection/yolox.py
```diff
@@ -115,6 +116,8 @@ def _create_model(self) -> nn.Module:
         return detector
 
     def _customize_inputs(self, entity: DetBatchDataEntity) -> dict[str, Any]:
+        if isinstance(entity.images, list):
+            entity.images = stack_batch(entity.images, pad_size_divisor=32, pad_value=114)
```
wonjuleee, line 120: Shouldn't we have `pad_size_divisor` and `pad_value` in a recipe for the configuration?

src/otx/core/data/entity/base.py
```diff
@@ -607,9 +608,19 @@ def collate_fn(
...
+                "can have different image shapes among samples in the batch. However, if it is not your intention, "
+                "consider setting stack_images as False in the config."
+            )
+            warnings.warn(msg, stacklevel=1)
+            stack_images = False
+
         return OTXBatchDataEntity(
             batch_size=batch_size,
-            images=tv_tensors.wrap(cls.stack_batch(images), like=like) if stack_images else images,
+            images=tv_tensors.wrap(torch.stack(images), like=like) if stack_images else images,
```
wonjuleee, line 623: Sorry, I was failed to understand why this is better. When it comes to me, it was possible to configure the variables through `cls.stack_batch(images, pad_size_divisor=..., pad_value=...)`. Please correct me if I was wrong.

jaegukhyun: Yes it could be, but it requires change to data entity of entire tasks. I think it is too risky w.r.t release schedule

src/otx/algo/detection/yolox.py
jaegukhyun: That value in recipe is for det data preprocessor which is implemented by mmdet. We are now decoupling it and data preprocessor is divided into model and data pipeline. So we now don't take those values from recipe.

kprokofi (approved):

PR #3415, opened 2024-04-29
Update to use `BaseModule` on detection

sungchul1:
### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algo/detection/heads/base_head.py
+++ b/src/otx/algo/detection/heads/base_head.py
@@ -62,12 +62,7 @@ class BaseDenseHead(BaseModule):
     """
 
     def __init__(self, init_cfg: ConfigDict | list[ConfigDict] | dict | list[dict] | None = None) -> None:
-        super().__init__()
-
-        self._is_init = False
-
-        self.init_cfg = copy.deepcopy(init_cfg)
-
+        super().__init__(init_cfg=init_cfg)
         # `_raw_positive_infos` will be used in `get_positive_infos`, which
         # can get positive information.
         self._raw_positive_infos: dict = {}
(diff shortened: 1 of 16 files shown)


jaegukhyun:
Could you check init_weights functions are fully migrated? I think some modules in otx/algo/detection miss their original init_weights function.

sungchul1:
> Could you check init_weights functions are fully migrated? I think some modules in otx/algo/detection miss their original init_weights function.

Done at https://github.com/openvinotoolkit/training_extensions/pull/3415/commits/c9a8562ecc1c11e36f56b6aa37d624793de36806.

src/otx/algo/detection/heads/base_head.py
jaegukhyun, line 1: BaseDenseHead has its own init weights function(https://github.com/open-mmlab/mmdetection/blob/ecac3a77becc63f23d9f6980b2a36f86acd00a8a/mmdet/models/dense_heads/base_dense_head.py#L66). I removed this kind of function since we haven't cared init weights. This PR is for initializing modules exact same with mm's implementation. So we need to bring all of init_weights function to OTX implementation. Could you check all other modules?

sungchul1: I see, but as you already know this overrided function seems for both module `DeformConv2dPack` and `ModulatedDeformConv2dPack`.
I think there is no model to use those modules in otx and this part seems to be ignored. Is it necessary?

jaegukhyun: I guess not, If all other modules don't need to bring to OTX, then it is totally fine. Could you check the all other modules?

sungchul1: At least, ssd, atss, and yolox don't forward that part.

harimkang (approved):

jaegukhyun (approved):

PR #3504, opened 2024-05-15
Add torchvision.MaskRCNN

eugene123tw:
### Checklist

### License

```python
```


harimkang:
Is this PR a releases/2.0.0 target? Or is it a develop target?

sovrasov:
I think it's for develop

src/otx/recipe/instance_segmentation/maskrcnn_efficientnetb2b.yaml
```diff
@@ -51,7 +51,7 @@ overrides:
                 - 1024
           - class_path: otx.core.data.transform_libs.torchvision.Pad
             init_args:
-              size_divisor: 32
+              pad_to_square: true
```
eunwoosh, line 54: Don't below `side_divisor` have to be changed to `pad_to_square`? And could you tell me why all `side_divisor = 32` is changed to `pad_to_squre`?

src/otx/algo/instance_segmentation/maskrcnn_tv.py
```diff
@@ -0,0 +1,265 @@
...
+            (backbone.out_channels, 7, 7),
+            [256, 256, 256, 256],
+            [1024],
+            norm_layer=nn.BatchNorm2d,
+        )
+        mask_head = MaskRCNNHeads(backbone.out_channels, [256, 256, 256, 256], 1, norm_layer=nn.BatchNorm2d)
+
+        model = TVMaskRCNN(
+            backbone,
+            num_classes=91,
```
eunwoosh, line 90: Is it ok to fix it?

eunwoosh (commented):

src/otx/algo/instance_segmentation/maskrcnn_tv.py
eugene123tw: To load a pre-trained COCO model, we need to create a model with the same number of output classes as the original COCO model was trained on.

src/otx/recipe/instance_segmentation/maskrcnn_efficientnetb2b.yaml
eugene123tw: Updated. And the purpose is to ensure that the resizing op aligns with the OV fit-to-window resize pre-processing step.

eunwoosh (approved):

harimkang (approved):

src/otx/algo/instance_segmentation/maskrcnn_tv.py
```diff
@@ -0,0 +1,265 @@
...
+    ) -> None:
+        super().__init__(
+            label_info=label_info,
+            optimizer=optimizer,
+            scheduler=scheduler,
+            metric=metric,
+            torch_compile=torch_compile,
+            tile_config=tile_config,
+        )
+        self.image_size = (1, 3, 1024, 1024)
```
kprokofi, line 58: We need to think how we can avoid hardcoded numbers in the models. Image_size is something that user may want to change. Same for mean/std

src/otx/algo/instance_segmentation/maskrcnn_tv.py
kprokofi: Maybe we can use strict=False? This number of classes a bit confusing.

eugene123tw: @kprokofi Setting `strict=False` doesn't resolve the issue.

```
model = TVMaskRCNN(
            backbone,
            num_classes=num_classes,
            rpn_anchor_generator=rpn_anchor_generator,
            rpn_head=rpn_head,
            box_head=box_head,
            mask_head=mask_head,
        )

model.load_state_dict(weights.get_state_dict(progress=True, check_hash=True))

RuntimeError: Error(s) in loading state_dict for TVMaskRCNN:
	size mismatch for roi_heads.box_predictor.cls_score.weight: copying a param with shape torch.Size([91, 1024]) from checkpoint, the shape in current model is torch.Size([2, 1024]).
	size mismatch for roi_heads.box_predictor.cls_score.bias: copying a param with shape torch.Size([91]) from checkpoint, the shape in current model is torch.Size([2]).
	size mismatch for roi_heads.box_predictor.bbox_pred.weight: copying a param with shape torch.Size([364, 1024]) from checkpoint, the shape in current model is torch.Size([8, 1024]).
	size mismatch for roi_heads.box_predictor.bbox_pred.bias: copying a param with shape torch.Size([364]) from checkpoint, the shape in current model is torch.Size([8]).
	size mismatch for roi_heads.mask_predictor.mask_fcn_logits.weight: copying a param with shape torch.Size([91, 256, 1, 1]) from checkpoint, the shape in current model is torch.Size([2, 256, 1, 1]).
	size mismatch for roi_heads.mask_predictor.mask_fcn_logits.bias: copying a param with shape torch.Size([91]) from checkpoint, the shape in current model is torch.Size([2]).
```

src/otx/algo/instance_segmentation/maskrcnn_tv.py
eugene123tw: I agree, it's only used for export, and currently, there's no other way to get attributes from recipe configs. We could find a way to access recipe parameters, but I don't know how to do it. This should be addressed in a different PR.

kprokofi (approved):

src/otx/algo/instance_segmentation/maskrcnn_tv.py
kprokofi: I will put this task to our backlog

src/otx/algo/instance_segmentation/maskrcnn_tv.py
kprokofi: Ok, I see, let's keep it then

PR #3515, opened 2024-05-17
Add early stopping with warmup. Remove mandatory background label in semantic segmentation task

kprokofi:
### Summary
Motivation:
1. Early stopping from lightning doesn't have a warmup parameter and it leads to underfitting when training on small Geti target datasets. I added two thresholds to have invariant for small datasets that can fit in one batch and for larger data

2. Inserting a background label actually can lead to class mismatches for the Segmentation head, sometimes datasets don't have background classes at all. I tried to divide the Geti case with polygons and ordinary training from file masks. When dm_dataset represents annotations with Polygons -> we insert a background label and create a head with num_classes + 1. For other cases, dataset_meta.json should explicitly reflect all classes for training. So, the number of indices in masks == num categories in dataset_meta.json.
Before these changes, if we add a background class to dataset_meta.json, the training will fail, because OTX deletes the background class during prefiltering. Now, the background class must be presented in a metadata file when needed.

### Checklist

### License

```python
```


src/otx/core/data/pre_filtering.py
```diff
@@ -72,11 +72,12 @@ def remove_unused_labels(dataset: DmDataset, data_format: str) -> DmDataset:
     used_labels: list[int] = list({ann.label for item in dataset for ann in item.annotations})
     if data_format == "ava":
         used_labels = [0, *used_labels]
-    elif data_format == "common_semantic_segmentation_with_subset_dirs":
-        if 0 in used_labels:
-            used_labels = [label - 1 for label in used_labels[1:]]
-        else:
-            used_labels = [label - 1 for label in used_labels]
+    if data_format == "common_semantic_segmentation_with_subset_dirs" and len(original_categories) < len(used_labels):
```
eunwoosh, line 75:
```suggestion
    if data_format == "common_semantic_segmentation_with_subset_dirs" and len(original_categories) != len(used_labels):
```
Is it ok only to check `len(original_categories) < len(used_labels)`?

src/otx/algo/callbacks/adaptive_early_stopping.py
```diff
@@ -0,0 +1,85 @@
...
+        # two thresholds to have invariant to extra small datasets and larger datasets
+        self.warmup_iters = warmup_iters
+        self.warmup_epochs = warmup_epochs
+
+    def _should_skip_check(self, trainer: pl.Trainer) -> bool:
+        from lightning.pytorch.trainer.states import TrainerFn
+
+        current_iter = trainer.current_epoch * trainer.num_training_batches
+        warmup_threshold = max(self.warmup_epochs * trainer.num_training_batches, self.warmup_iters)
+        return trainer.state.fn != TrainerFn.FITTING or trainer.sanity_checking or current_iter < warmup_threshold
```
eunwoosh:
```suggestion
        return super()._should_skip_check(trainer) or current_iter < warmup_threshold
```
How about using parent class's function? It can reflect changes if code in lightning side is updated later.

src/otx/core/data/dataset/segmentation.py
```diff
@@ -171,6 +171,13 @@ def __init__(
             stack_images,
             to_tv_image,
         )
+        is_polygon = isinstance(dm_subset[0].annotations[-1], (Polygon, Ellipse))
+        if is_polygon:
+            # insert background class at index 0
+            self.label_info.label_names.insert(0, "background")
+            # in semantic segmentation we have only one label_group
+            self.label_info.label_groups[-1].insert(0, "background")
```
vinnamkim: You are judging a dataset-level characteristic `self.label_info` by probing just one sample `dm_subset[0]` in the dataset. This logic is really uncomfortable to see and I think that we should find another way.

src/otx/algo/callbacks/adaptive_early_stopping.py
```diff
@@ -0,0 +1,85 @@
...
+            log_rank_zero_only=log_rank_zero_only,
+        )
+        # two thresholds to have invariant to extra small datasets and larger datasets
+        self.warmup_iters = warmup_iters
+        self.warmup_epochs = warmup_epochs
+
+    def _should_skip_check(self, trainer: pl.Trainer) -> bool:
+        from lightning.pytorch.trainer.states import TrainerFn
+
+        current_iter = trainer.current_epoch * trainer.num_training_batches
```
vinnamkim:
```suggestion
        current_iter = trainer.global_step
```

src/otx/core/data/dataset/segmentation.py
vinnamkim: I think that there has been such a problem on labeling in semantic segmentation task and we should fix the problem in a correct way. Before we do this, please answer me this first.

What is the model side restriction on the number of labels? Is it possible that the model head can handle `num_classes=1` (binary classification) case? If not, is it true that we should keep `num_classes > 1` due to this restriction?

vinnamkim (changes requested):

src/otx/core/data/pre_filtering.py
kprokofi: If we have original_categories > used_labels case, the code below will filter unused categories

src/otx/core/data/dataset/segmentation.py
kprokofi: In the current code model can't handle num_classes =1 . We should change head construction to support it.
I judge about dataset only by 1 example, because we don't have cases when we have mixed annotations. If we have polygons then we have Geti case, where we don't have background label. 
Let me think how to check it in a more appropriate way

kprokofi: I added check as property relying on subset.get_annotated_type(). I think now it is better

src/otx/core/types/label.py
```diff
@@ -283,16 +282,6 @@ class SegLabelInfo(LabelInfo):
 
     ignore_index: int = 255
 
-    def __post_init__(self):
```
vinnamkim, line 286: What's the reason that you removed this `__post_init__()`? If you remove this, it means that `SegLabelInfo` allows `label_names` even if there is no `Background` label in itself. Is it mandatory to have `background` label?

src/otx/core/data/dataset/segmentation.py
```diff
@@ -185,6 +183,17 @@ def __init__(
...
         self.ignore_index = ignore_index
 
+    @property
+    def has_polygons(self) -> bool:
+        """Check if the dataset has polygons in annotations."""
+        has_polygons = False
+        for subset in self.dm_subset.subsets().values():
+            annot_types = set(subset.get_annotated_type())
+            if annot_types & {"polygon", "ellipse"}:
+                has_polygons = True
```
eunwoosh:
```suggestion
                return True
```
If one of subset has polygons, we can skip later part, I think.

tests/unit/core/types/test_label.py
```diff
@@ -12,10 +12,10 @@ def test_as_json(fxt_label_info):
 
 def test_seg_label_info():
     # Automatically insert background label at zero index
-    assert SegLabelInfo(["car", "bug", "tree"], []) == SegLabelInfo(["Background", "car", "bug", "tree"], [])
+    assert SegLabelInfo(["car", "bug", "tree"], []) == SegLabelInfo(["car", "bug", "tree"], [])
```
eunwoosh: It seems useless assert. How about removing it?

src/otx/core/types/label.py
kprokofi: That is one of point in my PR. Remove inserting mandatory background label if we have masks as annotations. Because not all datasets require that, we can have all objects being annotated. Also, instead of background we can have like "sky", "sea", what ever. Mask annotations are consist of all labels and aren't required to insert background label. I didn't like that so much time in 1.6, but we kept it for Geti and now in 2.0 we can divide Geti case from OTX use case.

Going further, sometimes we train model for example with 3 classes in head, when we actually have 2 classes. It leads even to slightly performance change.

vinnamkim: I see. But, based on this comment, https://github.com/openvinotoolkit/training_extensions/pull/3515#discussion_r1604774122, I think that the real constraint from the model side exists: the number of labels should be more than one. This should be checked in `__post_init__()`

src/otx/core/data/dataset/segmentation.py
```diff
@@ -185,6 +183,17 @@ def __init__(
         )
         self.ignore_index = ignore_index
 
+    @property
+    def has_polygons(self) -> bool:
+        """Check if the dataset has polygons in annotations."""
+        has_polygons = False
+        for subset in self.dm_subset.subsets().values():
+            annot_types = set(subset.get_annotated_type())
```
vinnamkim, line 190:
```suggestion
        annot_types = set(self.dm_subset.parent.get_annotated_type())
```
Isn't it working?

kprokofi: No, self.dm_subset has no attribute parent

src/otx/core/types/label.py
kprokofi: In case of mask annotations it will be tracked by https://github.com/openvinotoolkit/training_extensions/pull/3515/files/043abd59e713391c61686b5041296e46f67ff57f#diff-4f6bc95c31fd707daea6b53a000198947e0a7b9d100c171d8a133b6615a2dc85R75
In case of polygons -> background class will be inserted

We don't need to add __post_init__() for num_classes check

eunwoosh (dismissed):

src/otx/core/types/label.py
vinnamkim: The current implementation does not fail if the dataset has only one label. However, the model implementation wants to fail at this case. What I requested is that this model side restriction is realized here.

kprokofi: Ok, I added check and raising an error when num of labels less or equal than 1

vinnamkim (approved):

harimkang (approved):
LGTM

PR #3518, opened 2024-05-20
Fix a bug that dino_v2 model can't be run w/ HPO

eunwoosh:
### Summary
This PR fixes [CSV-141649](https://[jira.devtools.intel.com/browse/CVS-141649](https://jira.devtools.intel.com/browse/CVS-141649))

Currently, dino_v2 models can't be pickled, which makes error during HPO.
To fix it, `__reduce__` function is added into dino class.
Additionally, to cover HPO availability on all models, extend HPO integration model list to all models.
To reduce integration time, decrease HPO time ratio to 1.

### How to test
otx train --config src/otx/recipe/semantic_segmentation/dino_v2.yaml --data_root tests/assets/common_semantic_segmentation_dataset/supervised/ --run_hpo True 

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algo/classification/dino_v2.py
+++ b/src/otx/algo/classification/dino_v2.py
@@ -23,7 +23,7 @@ from otx.core.model.base import DefaultOptimizerCallable, DefaultSchedulerCallab
 from otx.core.model.classification import OTXMulticlassClsModel
 from otx.core.schedulers import LRSchedulerListCallable
 from otx.core.types.label import LabelInfoTypes
-from otx.utils.utils import get_cls_init_args
+from otx.utils.utils import get_class_initial_arguments
 
 if TYPE_CHECKING:
     from lightning.pytorch.cli import LRSchedulerCallable, OptimizerCallable
@@ -47,7 +47,7 @@ class DINOv2(nn.Module):
         num_classes: int,
     ):
         super().__init__()
-        self._init_args = get_cls_init_args()
+        self._init_args = get_class_initial_arguments()
         self.backbone = torch.hub.load(
             repo_or_dir="facebookresearch/dinov2",
             model=backbone,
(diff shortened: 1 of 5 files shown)


src/otx/algo/classification/dino_v2.py
```diff
@@ -75,6 +77,9 @@ def forward(self, imgs: torch.Tensor, labels: torch.Tensor | None = None, **kwar
             return self.loss(logits, labels)
         return self.softmax(logits)
 
+    def __reduce__(self):
```
harimkang, line 80: I was wondering if this should always be considered when an external user wants to add a new model to the OTX?

src/otx/utils/utils.py
```diff
@@ -176,3 +176,13 @@ def can_pass_tile_config(model_cls: type[OTXModel]) -> bool:
     """
     tile_config_param = inspect.signature(model_cls).parameters.get("tile_config")
     return tile_config_param is not None
+
+
+def get_cls_init_args() -> tuple:
```
harimkang: How about `get_class_initial_arguments`

src/otx/algo/classification/dino_v2.py
eunwoosh: It depends on class implementation. The reason why dino_v2 model is problematic is that it tries to get model using `torch.hub.load` whose class can't be found w/o any reference. If user want to implement a model in the similar way, then it may be necessary.

src/otx/utils/utils.py
eunwoosh: done https://github.com/openvinotoolkit/training_extensions/pull/3518/commits/61b8c550f076385d42338c32daf8522d82f7ddf8

harimkang (approved):

jaegukhyun (approved):

PR #3532, opened 2024-05-22
Add a feature to adapt max value of HPO batch size search space

eunwoosh:
### Summary
This PR adds new HPO feature.
New feature can adapt max value of HPO batch size search space automatically.
It provides user an easy way to avoid OOM due to setting wrong batch size search space.
It exploits auto-batch size feature, so it has two ways to use.

- Safe : Just check max value of batch size search space is ok and decrease it if now.
- Full : Increase max value of batch size as possible as memory can handle.

### How to test
```
  otx train --config ... --run_hpo true --hpo_config.adapt_bs_search_space_max_val Safe
  otx train --config ... --run_hpo true --hpo_config.adapt_bs_search_space_max_val Full
```

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/core/config/hpo.py
+++ b/src/otx/core/config/hpo.py
@@ -38,3 +38,4 @@ class HpoConfig:
     asynchronous_bracket: bool = True
     asynchronous_sha: bool = num_workers > 1
     metric_name: str | None = None
+    adapt_bs_search_space_max_val: Literal["None", "Safe", "Full"] = "None"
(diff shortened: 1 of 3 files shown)


jaegukhyun (approved):

src/otx/core/config/hpo.py
```diff
@@ -38,3 +38,4 @@ class HpoConfig:
     asynchronous_bracket: bool = True
     asynchronous_sha: bool = num_workers > 1
     metric_name: str | None = None
+    adapt_bs_search_space_max_val: Literal["None", "Safe", "Full"] = "None"
```
harimkang, line 41: Isn't mixing uppercase and lowercase letters unnatural from a user usability perspective?

harimkang (commented):
Thanks, i left minor comment

src/otx/core/config/hpo.py
eunwoosh: This feature uses 'auto batch size' feature, so I think it's better to align its naming to avoid confusion.

harimkang (approved):

PR #3538, opened 2024-05-24
Split API tests from integration tests

yunchu:
### Checklist

### License

```python
```

### DIFF
--- a/.github/workflows/pre_merge.yaml
+++ b/.github/workflows/pre_merge.yaml
@@ -92,21 +92,21 @@ jobs:
       matrix:
         include:
           - task: "action"
-            filter: "-k 'not api'"
+            filter: "-m 'not api'"
           - task: "classification"
-            filter: "-k 'not api'"
+            filter: "-m 'not api'"
           - task: "detection"
-            filter: "-k 'not api'"
+            filter: "-m 'not api'"
           - task: "instance_segmentation"
-            filter: "-k 'not api'"
+            filter: "-m 'not api'"
           - task: "semantic_segmentation"
-            filter: "-k 'not api'"
+            filter: "-m 'not api'"
           - task: "visual_prompting"
-            filter: "-k 'not api'"
+            filter: "-m 'not api'"
           - task: "anomaly"
-            filter: "-k 'not api'"
+            filter: "-m 'not api'"
           - task: "all"
-            filter: "-k 'not cli'"
+            filter: "-m api"
     name: Integration-Test-${{ matrix.task }}-py310
     # This is what will cancel the job concurrency
     concurrency:
--- a/.github/workflows/pre_merge.yaml
+++ b/.github/workflows/pre_merge.yaml
@@ -92,21 +92,12 @@ jobs:
       matrix:
         include:
           - task: "action"
-            filter: "-m 'not api'"
           - task: "classification"
-            filter: "-m 'not api'"
           - task: "detection"
-            filter: "-m 'not api'"
           - task: "instance_segmentation"
-            filter: "-m 'not api'"
           - task: "semantic_segmentation"
-            filter: "-m 'not api'"
           - task: "visual_prompting"
-            filter: "-m 'not api'"
           - task: "anomaly"
-            filter: "-m 'not api'"
-          # - task: "all"
-          #   filter: "-m api"
     name: Integration-Test-${{ matrix.task }}-py310
     # This is what will cancel the job concurrency
     concurrency:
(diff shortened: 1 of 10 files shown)


.github/workflows/pre_merge.yaml
```diff
@@ -81,19 +81,32 @@ jobs:
...
+    container:
+      image: 219678651685.dkr.ecr.eu-west-1.amazonaws.com/ote-ci:11.7.1.2-devel-ubuntu20.04
+      options: "--runtime=nvidia --env-file=/home/runner/.nvidia.env --ipc=host"
     needs: Unit-Test
     strategy:
       fail-fast: false
       matrix:
         include:
           - task: "action"
+            filter: "-k 'not api'"
```
vinnamkim: Rather than using `-k`, add a custom pytest.marker and use it: https://docs.pytest.org/en/latest/example/markers.html
Refer to this
https://github.com/openvinotoolkit/training_extensions/blob/ce2d99f5551b6f7ede3a6968a6e279c8f8ba185d/pyproject.toml#L366-L370

tests/integration/api/test_auto_configuration.py
```diff
@@ -12,7 +12,7 @@
 
 
 @pytest.mark.parametrize("task", pytest.TASK_LIST)
-def test_auto_configuration(
+def test_api_auto_configuration(
```
vinnamkim:
```suggestion
@pytest.mark.api
def test_auto_configuration(
```

vinnamkim (dismissed):

tests/unit/algo/anomaly/test_openvino_model.py
```diff
@@ -0,0 +1,50 @@
...
+
+        # NOTE: if we do not inject those into `model`,
+        # we will face errors during `model.export()` such as:
+        # AttributeError: 'Stfpm' object has no attribute 'normalization_metrics'
+        model.normalization_metrics = MinMax()
+        model.normalization_metrics.min = torch.tensor(-1.0)
+        model.normalization_metrics.max = torch.tensor(1.0)
+        model.image_threshold = ManualThreshold(0.0)
+        model.pixel_threshold = ManualThreshold(0.0)
+        model.task = OTXTaskType.ANOMALY_CLASSIFICATION
```
vinnamkim, line 40:
```python
model = Padim()
model.export(
            output_dir=tmpdir,
            base_name="exported_model",
            export_format=OTXExportFormatType.OPENVINO,
        )
```

Hi @ashwinvaidya17,
It seems impossible to export OTX-integrated anomalib models in this way. Can we enhance this?

vinnamkim: If not, is there anyway to enhance just this test code neater without injecting those variables explicitly?

ashwinvaidya17: These variables are set from the callbacks when the `setup` method is called in the callback.
https://github.com/openvinotoolkit/training_extensions/blob/80acb86508c12d5602eb93399a91a476b3feffd3/src/otx/core/model/anomaly.py#L230

One option would be to assign default values here https://github.com/openvinotoolkit/training_extensions/blob/80acb86508c12d5602eb93399a91a476b3feffd3/src/otx/core/model/anomaly.py#L149 so that export works when the model is created. But I don't know how it will behave with the callbacks. And, since the default values are incorrect, directly calling export on a new model will lead to an incorrect OV model.

vinnamkim: I see. Then, it would be good to raise an appropriate error for better UX in this case. For example,
```python
# If an user is trying to do
model = Padim()
model.export(
            output_dir=tmpdir,
            base_name="exported_model",
            export_format=OTXExportFormatType.OPENVINO,
        )

# Raise this kind of error
> RuntimeError(
    "Model export failed: The model must be trained before exporting. "
    "Please train the model first to ensure all required parameters are set for correct export behavior."
)
```

ashwinvaidya17: Yeah that would be better

yunchu:
@vinnamkim could you create separated PR for the anomaly related changes?

vinnamkim:
> @vinnamkim could you create separated PR for the anomaly related changes?

https://github.com/openvinotoolkit/training_extensions/pull/3557

yunchu:
close this PR because api test splitting required too many number of ci-runner resources.
instead, PR #3553 could be used to enable using new CI infra.

PR #3551, opened 2024-05-28
Sync rgb order between torch and ov inference of action classification task

jaegukhyun:
### Summary
This PR is for synchronize RGB order of ov inference and torch inference in action classification task

Currently action classification requires video loading for inference. For torch model inference OTX use video loading pipeline from mmaction. And for ov model inference OTX use datumaro video loading pipeline. 
MMAction loads video as BGR format and Datumaro loads video as RGB format. 
I'm not sure the pretrained weight is trained by BGR or RGB. 
Therefore, we need additional PR to sync between pretrained weights and torch, ov inference

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/core/data/transform_libs/torchvision.py
+++ b/src/otx/core/data/transform_libs/torchvision.py
@@ -236,7 +236,10 @@ class DecodeVideo(tvt_v2.Transform):
         start_index = 0
         frame_inds = np.concatenate(frame_inds) + start_index
 
-        outputs = torch.stack([torch.tensor(inpt[idx].data) for idx in frame_inds], dim=0)
+        outputs = torch.stack(
+            [torch.tensor(cv2.cvtColor(inpt[idx].data, cv2.COLOR_RGB2BGR)) for idx in frame_inds],
+            dim=0,
+        )
         outputs = outputs.permute(0, 3, 1, 2)
         outputs = tv_tensors.Video(outputs)
         inpt.close()
--- a/src/otx/recipe/_base_/data/mmaction_base.yaml
+++ b/src/otx/recipe/_base_/data/mmaction_base.yaml
@@ -78,7 +78,7 @@ config:
     transform_lib_type: MMACTION
     to_tv_image: False
     batch_size: 8
-    num_workers: 2
+    num_workers: 0
     transforms:
       - type: LoadVideoForClassification
       - type: DecordInit
--- a/src/otx/recipe/_base_/data/mmaction_base.yaml
+++ b/src/otx/recipe/_base_/data/mmaction_base.yaml
@@ -78,7 +78,7 @@ config:
     transform_lib_type: MMACTION
     to_tv_image: False
     batch_size: 8
-    num_workers: 0
+    num_workers: 2
     transforms:
       - type: LoadVideoForClassification
       - type: DecordInit
--- a/src/otx/recipe/_base_/data/mmaction_base.yaml
+++ b/src/otx/recipe/_base_/data/mmaction_base.yaml
@@ -5,7 +5,7 @@ config:
   mem_cache_img_max_size:
     - 500
     - 500
-  image_color_channel: RGB
+  image_color_channel: BGR
   stack_images: False
   unannotated_items_ratio: 0.0
   train_subset:
--- a/src/otx/core/data/dataset/action_classification.py
+++ b/src/otx/core/data/dataset/action_classification.py
@@ -6,7 +6,7 @@
 from __future__ import annotations
 
 from functools import partial
-from typing import Callable
+from typing import TYPE_CHECKING, Callable
 
 import torch
 from datumaro import Label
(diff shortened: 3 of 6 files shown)


harimkang (dismissed):

src/otx/core/data/transform_libs/torchvision.py
```diff
@@ -236,7 +236,10 @@ def _transform(self, inpt: Video, params: dict) -> tv_tensors.Video:
         start_index = 0
         frame_inds = np.concatenate(frame_inds) + start_index
 
-        outputs = torch.stack([torch.tensor(inpt[idx].data) for idx in frame_inds], dim=0)
+        outputs = torch.stack(
+            [torch.tensor(cv2.cvtColor(inpt[idx].data, cv2.COLOR_RGB2BGR)) for idx in frame_inds],
```
eunwoosh, line 240: It seems that this code always converts color format. Could you change a code to convert only if it's `image_color_channel` is RGB? You can check this by adding `_check_inputs` function to `DecodeVideo` class.

eunwoosh (commented):
Thanks for your work. I left a comment. Please take a look.

sungchul1:
@jaegukhyun I updated `auto_configurator` to align ov model's `image_color_channel` in releases/2.0.0.
https://github.com/openvinotoolkit/training_extensions/pull/3534/files#diff-5fef67b60dac607755ebe35a63b56c1c773c540814354b4c7e3c1a1447c09581R387
Could you check if this update can be applied to this PR?

jaegukhyun:
> @jaegukhyun I updated `auto_configurator` to align ov model's `image_color_channel` in releases/2.0.0. https://github.com/openvinotoolkit/training_extensions/pull/3534/files#diff-5fef67b60dac607755ebe35a63b56c1c773c540814354b4c7e3c1a1447c09581R387 Could you check if this update can be applied to this PR?

@sungchul2 
It seems this PR should check the image_color_channel of openvino_model.py in recipe folder. Is it right?

src/otx/core/data/transform_libs/torchvision.py
jaegukhyun: image_color_channel is not counted in action task since it does not consider video case. So firstly we need to sync default channels of torch and ov inference. And then secondly we need to implement the way to apply image_color_channel for the video case.
This PR aims the first step, and we need additional PR for the second step

sungchul1:
> > @jaegukhyun I updated `auto_configurator` to align ov model's `image_color_channel` in releases/2.0.0. https://github.com/openvinotoolkit/training_extensions/pull/3534/files#diff-5fef67b60dac607755ebe35a63b56c1c773c540814354b4c7e3c1a1447c09581R387 Could you check if this update can be applied to this PR?
> 
> @sungchul2 It seems this PR should check the image_color_channel of openvino_model.py in recipe folder. Is it right?

Right. Without this update, your below change is meaningless because otx in develop doesn't reflect `image_color_channel` in `openvino_model.yaml`.
https://github.com/openvinotoolkit/training_extensions/pull/3551/files#diff-662e7ef53bb9f36e726723c96cc1f6033ace1d29f557c236cadd777f1a36d7ebR21

jaegukhyun:
> > > @jaegukhyun I updated `auto_configurator` to align ov model's `image_color_channel` in releases/2.0.0. https://github.com/openvinotoolkit/training_extensions/pull/3534/files#diff-5fef67b60dac607755ebe35a63b56c1c773c540814354b4c7e3c1a1447c09581R387 Could you check if this update can be applied to this PR?
> > 
> > 
> > @sungchul2 It seems this PR should check the image_color_channel of openvino_model.py in recipe folder. Is it right?
> 
> Right. Without this update, your below change is meaningless because otx in develop doesn't reflect `image_color_channel` in `openvino_model.yaml`. https://github.com/openvinotoolkit/training_extensions/pull/3551/files#diff-662e7ef53bb9f36e726723c96cc1f6033ace1d29f557c236cadd777f1a36d7ebR21

Thanks for the notice. Currently action task do not use image color channel both on torch and ov inference, but torch and ov inference use different format. This PR is trying to sync the default color format of torch and ov inference. After this PR we need to link image_color_channel for torch and ov inference of action classification. At that PR we need to consider your point

src/otx/core/data/transform_libs/torchvision.py
jaegukhyun: @eunwoosh  I added error msg when users change ImageColorChannel value to RGB. Currently action classification task should use BGR format.

eunwoosh (dismissed):

src/otx/core/data/transform_libs/torchvision.py
sungchul1: I don't understand why this part is needed if `ImageColorChannel=BGR` is forced.
Because `cv2.COLOR_RGB2BGR` and `cv2.COLOR_BGR2RGB` are actually the same operation.
If BGR image is given, this part will convert the image to RGB.
https://docs.opencv.org/4.x/d8/d01/group__imgproc__color__conversions.html
image

jaegukhyun: Video from datumaro always RGB format. However Video from mmaction always BGR format. So this PR aims to change the color format of video from datumaro to BGR format

eunwoosh: Yes, as Jaeguk said, current action classification task doesn't support ImageColorChannel. So user should use BGR format regardless of `ImageColorChannel` value. To avoid confusion, this code notifies user that it's not possible to set `ImageColorChannel`.

sungchul1: Thanks for clarification.

src/otx/recipe/_base_/data/mmaction_base.yaml
```diff
@@ -78,7 +78,7 @@ config:
     transform_lib_type: MMACTION
     to_tv_image: False
     batch_size: 8
-    num_workers: 2
+    num_workers: 0
```
eunwoosh: Could you tell me why it's set to 0?

jaegukhyun: Thanks for the notice, I reverted it

eunwoosh (approved):

sungchul1 (approved):

PR #3558, opened 2024-05-29
Fix Anomaly OV export flag

ashwinvaidya17:
### Summary

 Fixes: CVS-140888

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/core/model/anomaly.py
+++ b/src/otx/core/model/anomaly.py
@@ -77,7 +77,7 @@ class _AnomalyModelExporter(OTXModelExporter):
             input_size=(1, 3, *image_shape),
             mean=mean_values,
             std=scale_values,
-            swap_rgb=True,  # BGR -> RGB
+            swap_rgb=False,
         )
 
     @property


src/otx/core/model/anomaly.py
```diff
@@ -77,7 +77,7 @@ def __init__(
             input_size=(1, 3, *image_shape),
             mean=mean_values,
             std=scale_values,
-            swap_rgb=True,  # BGR -> RGB
+            swap_rgb=False,
```
ashwinvaidya17, line 80: Tbh I don't really understand this flag. While the computed metrics are correct, the models do require images in RGB format. And according to this line setting it to True converts it to RGB https://github.com/openvinotoolkit/model_api/blob/a058258bbfebedb06843450b71730c9df56defc6/model_api/python/model_api/adapters/utils.py#L530. I wonder if it is because OTX provides RGB image and when this is set to True, the actual image the model reads is BGR

sovrasov: That flag is required when we feed images read by cv2.imread directly to the wrapper. That is the intended case for ModelAPI, but in OTX we have RGB convention for both wrappers and models, which means wrappers consume RGB and don't do channel swapping internally. cvtColor actually doesn't know the actual color order, and cv2.COLOR_BGR2RGB flag just means the channels swap.

sovrasov (approved):

harimkang (approved):

PR #3561, opened 2024-05-30
Fix Optimize in Anomaly Task

ashwinvaidya17:
### Summary

- Use `val_dataloader` for NNCF optimization
- Fixes: CVS-141031

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algo/anomaly/openvino_model.py
+++ b/src/otx/algo/anomaly/openvino_model.py
@@ -14,17 +14,21 @@ from __future__ import annotations
 from typing import TYPE_CHECKING, Any, Sequence
 
 import numpy as np
+import openvino
 import torch
 from anomalib.metrics import create_metric_collection
 from lightning import Callback, Trainer
 from torchvision.transforms.functional import resize
 
 from otx.core.data.entity.anomaly import AnomalyClassificationDataBatch
+from otx.core.data.module import OTXDataModule
 from otx.core.metrics.types import MetricCallable, NullMetricCallable
 from otx.core.model.anomaly import AnomalyModelInputs
 from otx.core.model.base import OVModel
 
 if TYPE_CHECKING:
+    from pathlib import Path
+
     from anomalib.metrics import AnomalibMetricCollection
     from model_api.models import Model
     from model_api.models.anomaly import AnomalyResult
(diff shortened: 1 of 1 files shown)


src/otx/algo/anomaly/openvino_model.py
```diff
@@ -133,6 +137,55 @@ def _create_model(self) -> Model:
...
+            nodes = model.get_ops()
+            return any(op.get_type_name() == "FakeQuantize" for op in nodes)
+
+        ov_model = openvino.Core().read_model(self.model_name)
+
+        if check_if_quantized(ov_model):
+            msg = "Model is already optimized by PTQ"
+            raise RuntimeError(msg)
+
+        train_dataset = data_module.val_dataloader()
```
ashwinvaidya17: Since this is the only line that is different is there a better way to update this method without duplicating the code?

harimkang: Is there any other reason why Anomaly specifically uses `val_dataloader` in anomaly?
@sovrasov What do you think about this? If this is what we need, shouldn't we be able to specify a subset in `OVModel.optimize`?

harimkang:
```suggestion
        val_dataset = data_module.val_dataloader()
```

src/otx/algo/anomaly/openvino_model.py
```diff
@@ -133,6 +137,55 @@ def _create_model(self) -> Model:
...
+
+        ptq_config_from_ir = self._read_ptq_config_from_ir(ov_model)
+        if ptq_config is not None:
+            ptq_config_from_ir.update(ptq_config)
+            ptq_config = ptq_config_from_ir
+        else:
+            ptq_config = ptq_config_from_ir
+
+        quantization_dataset = nncf.Dataset(train_dataset, self.transform_fn)  # type: ignore[attr-defined]
```
harimkang:
```suggestion
        quantization_dataset = nncf.Dataset(val_dataset, self.transform_fn)  # type: ignore[attr-defined]
```

harimkang (dismissed):

chuneuny-emily:
@ashwinvaidya17 - Are we using test set for optimization? Did you check optimization works fine with this fix?

src/otx/algo/anomaly/openvino_model.py
ashwinvaidya17: I am not the expert here. I just referred to Adrian's code here https://github.com/openvinotoolkit/anomalib/blob/e5b91d6a32c469ec0b9c4c161272b68114ba6fa3/src/anomalib/models/components/base/export_mixin.py#L245
The only difference is that the val dataset also contains the anomaly class.

ashwinvaidya17:
> @ashwinvaidya17 - Are we using test set for optimization? Did you check optimization works fine with this fix?

I tried optimization on one category (`hazelnut_large`) it seems to be working fine. I think it does use the test set for optimization as the val split is set to test set. Ideally it would be better if we could split the test set into val set and test set.

harimkang (approved):

chuneuny-emily (approved):

PR #3569, opened 2024-06-02
Bug fix in Semantic Segmentation + enable DINOV2 export in ONNX

kprokofi:
### Summary
1) Only without registers DINOV2 can be exported to ONNX. The accuracy difference is insignificant
2) Bug fix with ignore index in labels pre-filtering. Bug fix with overriding default values in segmentation models

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algo/segmentation/dino_v2_seg.py
+++ b/src/otx/algo/segmentation/dino_v2_seg.py
@@ -46,9 +46,8 @@ class OTXDinoV2Seg(TorchVisionCompatibleModel):
 
     def _create_model(self) -> nn.Module:
         # merge configurations with defaults overriding them
-        backbone_configuration = self.backbone_configuration | DinoV2Seg.default_backbone_configuration
-        decode_head_configuration = self.decode_head_configuration | DinoV2Seg.default_decode_head_configuration
-        # initialize backbones
+        backbone_configuration = DinoV2Seg.default_backbone_configuration | self.backbone_configuration
+        decode_head_configuration = DinoV2Seg.default_decode_head_configuration | self.decode_head_configuration
         backbone = DinoVisionTransformer(**backbone_configuration)
         decode_head = FCNHead(num_classes=self.num_classes, **decode_head_configuration)
         return DinoV2Seg(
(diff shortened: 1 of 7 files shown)


src/otx/core/data/pre_filtering.py
```diff
@@ -16,7 +16,12 @@
     from datumaro.components.dataset_base import DatasetItem
 
 
-def pre_filtering(dataset: DmDataset, data_format: str, unannotated_items_ratio: float) -> DmDataset:
+def pre_filtering(
+    dataset: DmDataset,
+    data_format: str,
+    unannotated_items_ratio: float,
+    ignore_index: int | None = None,
```
eunwoosh, line 23: Could you add docstring of this argument?

src/otx/algo/segmentation/dino_v2_seg.py
```diff
@@ -46,9 +46,8 @@ class OTXDinoV2Seg(TorchVisionCompatibleModel):
 
     def _create_model(self) -> nn.Module:
         # merge configurations with defaults overriding them
-        backbone_configuration = self.backbone_configuration | DinoV2Seg.default_backbone_configuration
-        decode_head_configuration = self.decode_head_configuration | DinoV2Seg.default_decode_head_configuration
-        # initialize backbones
+        backbone_configuration = DinoV2Seg.default_backbone_configuration | self.backbone_configuration
+        decode_head_configuration = DinoV2Seg.default_decode_head_configuration | self.decode_head_configuration
```
harimkang, line 50: Why is `self.**`_configuration separate, and is this something that can actually be used?

eunwoosh: a_dict | b_dict is same as a_dict.update(b_dict). So, self.**_configuration has a priority than default configuration.

eunwoosh (approved):

harimkang (approved):

src/otx/algo/segmentation/dino_v2_seg.py
kprokofi: I added this default configuration for simple API usage. Let's imagine that we don't have it. Then user should create model and pass all parameters to initialize backbone and head. It is obvious overhead. Now, he can provide only few parameters to override default values.
self.** is what comes from config or what is passed to model class.

PR #3609, opened 2024-06-12
Fix setuptools dependency

harimkang:
### Checklist

### License

```python
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ dependencies = [
     "importlib_resources==6.4.0",
     "docstring_parser==0.16", # CLI help-formatter
     "rich_argparse==1.4.0", # CLI help-formatter
-    "setuptools==69.5.1",
+    "setuptools<70",
 ]
 
 [project.optional-dependencies]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,6 +36,7 @@ dependencies = [
     "importlib_resources==6.4.0",
     "docstring_parser==0.16", # CLI help-formatter
     "rich_argparse==1.4.0", # CLI help-formatter
+    # TODO(ashwinvaidya17): https://github.com/openvinotoolkit/anomalib/issues/2126
     "setuptools<70",
 ]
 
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -36,6 +36,7 @@ dependencies = [
     "importlib_resources==6.4.0",
     "docstring_parser==0.16", # CLI help-formatter
     "rich_argparse==1.4.0", # CLI help-formatter
+    "setuptools==69.5.1",
```
yunchu: how about limiting like this?
```suggestion
    "setuptools<70",
```

harimkang: Thanks, Done -  https://github.com/openvinotoolkit/training_extensions/pull/3609/commits/f924ead7e36a3ee0f555343abb6ec0c2aea961e9

pyproject.toml
```diff
@@ -36,6 +36,7 @@ dependencies = [
     "importlib_resources==6.4.0",
     "docstring_parser==0.16", # CLI help-formatter
     "rich_argparse==1.4.0", # CLI help-formatter
+    "setuptools<70",
```
eunwoosh, line 40: I think it's better to make an issue and add a comment to refer it for maintenance. How do you think about it?

yunchu: before filing an issue, need to find root cause of it. it's probably not related with otx itself so we need to clarify the actual target of the issue.

eunwoosh: There is relevant issue https://github.com/pypa/setuptools/issues/4376.
It seems that `setuptools` has a code `from pkg_resources.extern import packaging` in __init__.py before v70 but not include it in `__all__`.
So, it means that using `pkg_resources .packaging` isn't guaranteed and it's removed since v70.
Maybe it needs to be fixed by torch side, but I can't find an issue in torch repo.

harimkang: This seems to be an issue on the anomalib side. I'll open an issue.

harimkang: Add comments - https://github.com/openvinotoolkit/training_extensions/pull/3609/commits/f8462c6ce9db4d518d6069d97683992d54cf84bf

eunwoosh (approved):

yunchu (approved):

chuneuny-emily (approved):

PR #3615, opened 2024-06-13
Various fixes for XAI in 2.1

GalyaZalesskaya:
### Summary

Issue: [CVS-142112](https://jira.devtools.intel.com/browse/CVS-142112)

Found and fixed various issues:
- Dump saliency maps using label names, if possible
- The score_treshold for classification, detection and inst_seg is too low. This means that choosing saliency maps for only predicted classes works incorrectly. Added additional score_thr=0.3 to choose saliency map with high confidence
- Fix softmax for detection saliency map generation algorithm
- Fix saliency maps for YOLOX. Added cropping saliency maps if it was padded.  
    YOLOX resizes the initial image, keeping the aspect ratio and padding the image to square. The generated saliency map is also padded so after overlaying it on the initial image, it looks distorted. Added cropping to fix it.
- Fix h-label cls saliency map dumping for predicted classes. Converted predicted labels from h-label format to index in labels_name list based on hierarchical label info, e.g. [0,0,1] -> 5 [CVS-144040](https://jira.devtools.intel.com/browse/CVS-144040)

image
 
TODO:
- Fix XAI for MobileNet V3 
- Fix XAI for multilabel cls 
- Apply these changes to Model API to update OV task

### Checklist

### License

```python
```


src/otx/algo/utils/xai_utils.py
```diff
@@ -47,31 +49,57 @@
...
-
-        processed_saliency_maps = process_saliency_maps(saliency_map, explain_config, pred_labels, ori_img_shapes)
+        paddings = [img_info.padding for img_info in imgs_info]
+        image_shape = imgs_info[0].img_shape
+
+        pred_labels = []
+        for labels, scores in zip(predict_result_per_batch.labels, predict_result_per_batch.scores):
+            if isinstance(label_info, HLabelInfo):
+                pred_labels.append(_convert_labels_from_hcls_format(labels, scores, label_info))
+            elif labels.shape == scores.shape:
```
goodsong81, line 76: What do you mean by this condition?

src/otx/algo/utils/xai_utils.py
```diff
@@ -47,31 +49,57 @@
...
+                # Filter out predictions with scores less than 0.3
+                pred_labels.append(labels[scores > 0.3].tolist())
```
goodsong81: 0.3 looks somewhat arbitrary to me, but it's OK if working well.
(How did you test this part of fix?)
I'd recommend to create constant in the module, class or recipe with default 0.3 to be referenced here.

src/otx/algo/utils/xai_utils.py
```diff
@@ -195,3 +230,51 @@ def _get_overlay(img: np.ndarray, s_map: np.ndarray, weight: float = 0.3) -> np.
...
+    # Padding: number of pixels to pad all borders (left, top, right, bottom)
+    height, width = image_shape
+
+    for i, image_map in enumerate(batch_saliency_map):
+        left, top, right, bottom = paddings[i]
+        left_ratio, top_ratio, right_ratio, bottom_ratio = left / width, top / height, right / width, bottom / height
+        for class_idx, class_map in image_map.items():
+            map_h, map_w = class_map.shape
+            d_top, d_bottom = int(top_ratio * map_h), int(bottom_ratio * map_h)
+            d_left, d_right = int(left_ratio * map_w), int(right_ratio * map_w)
```
goodsong81, line 254: Why don't you use `round()`? Or, we can even adopt bi-linear interpolation, like ROIAlign in MaskRCNN, to be more precise. (in the future)

src/otx/algo/utils/xai_utils.py
```diff
@@ -195,3 +230,51 @@ def _get_overlay(img: np.ndarray, s_map: np.ndarray, weight: float = 0.3) -> np.
...
+            label_str = label_info.all_groups[i][j]
+            pred_labels.append(label_info.label_to_idx[label_str])
+    if label_info.num_multilabel_classes:
+        for i in range(label_info.num_multilabel_classes):
+            j = label_info.num_multiclass_heads + i
+            if labels[j] and scores[j] > 0.3:
+                label_str = label_info.all_groups[j][0]
+                pred_labels.append(label_info.label_to_idx[label_str])
+
+    return pred_labels
```
goodsong81: Same here. Needs some constant for 0.3 in order for common reference.

src/otx/engine/engine.py
```diff
@@ -479,7 +479,12 @@ def predict(
             if explain_config is None:
                 explain_config = ExplainConfig()
 
-            predict_result = process_saliency_maps_in_pred_entity(predict_result, explain_config)
+            for transform in datamodule.config.test_subset.transforms:
+                tranf_name = transform["class_path"].split(".")[-1]
+                if tranf_name == "Resize" and transform["init_args"].get("keep_ratio", False):
+                    explain_config.crop_padded_map = True
```
goodsong81: Does the "keep_ratio=True" mean that we need to crop & pad the saliency map? Is it general for all tasks?

src/otx/engine/engine.py
```diff
@@ -724,7 +729,12 @@ def explain(
         if explain_config is None:
             explain_config = ExplainConfig()
 
-        predict_result = process_saliency_maps_in_pred_entity(predict_result, explain_config)
+        for transform in datamodule.config.test_subset.transforms:
+            tranf_name = transform["class_path"].split(".")[-1]
+            if tranf_name == "Resize" and transform["init_args"].get("keep_ratio", False):
+                explain_config.crop_padded_map = True
```
goodsong81: Duplicated logics. Can we refactor to a common place?

goodsong81 (commented):
Thank you for the fix. Please review my comments.

tests/integration/api/test_xai.py
```diff
@@ -177,6 +177,11 @@ def test_predict_with_explain(
         # Tickets: 142087, 141639
         return
 
+    if "yolox" in recipe:
+        # The cropping of the padded saliency maps is not implemented for OV (Model API) yet,
+        # so the saliency maps for PyTorch and OV are different.
+        return
```
harimkang: Is there a reason why the skip associated with this comment only happens in `yolox`?
And is there a ticket for this?

src/otx/algo/utils/xai_utils.py
GalyaZalesskaya: It's the most common case for filtering out predictions which is applied to most of the models. It's a simple check that we actually can filter out labels based on scores. This way the same size is needed condition.

If it's not true, like for example it happens for tv_* models that return just 0-size tensor scalar with int for predicted label (`tensor(3)` instead of `tensor([3])`) we procced without filtering.

src/otx/algo/utils/xai_utils.py
GalyaZalesskaya: I followed the same approach we have in merging saliency maps for tiling where I also had to crop maps:

https://github.com/openvinotoolkit/training_extensions/blob/785a560dbff160d2a25e64ca114bd48fb943a632/src/otx/core/utils/tile_merge.py#L258-L259

src/otx/engine/engine.py
GalyaZalesskaya: Yes, but based on my investigation it's true only for YOLOX-based models

src/otx/algo/utils/xai_utils.py
goodsong81: The coordinates will be shifted to left a bit due to this "floor" operation.
But let's keep it as is for consistency or backward compatibility.

harimkang (approved):

goodsong81 (approved):

src/otx/algo/utils/xai_utils.py
GalyaZalesskaya: It's a small thr just to filter out low-confidence predictions like 0.05-0.1, so it can be any number from `0.2` to `0.5`. I chose middle value and manually chose it works for test images.

PR #3634, opened 2024-06-18
Fix detection prediction outputs

sungchul1:
### Summary

Ticket no. : 144396

Torch models have abnormal outputs, which seem unfiltered (especially ATSS and SSD).
It can yield wrong performance measurements by adding very many false positives.

box color : blue - GT | red - predictions
image

But, OV model have normal outputs because outputs are filtered before wrapped by `DetectionResult`

figure order : rtmdet_tiny OV Model | rtmdet_tiny Torch Model | atss_mobilenetv2 Torch Model
image

To fix this issue, this PR includes:
- [x] Use `confidence_threshold` to filter outputs 
    - model api uses it (default : 0.5) to filter outputs ([link](https://github.com/openvinotoolkit/model_api/blob/a9c116de74da4e0f7b72d869af1480b525bd6c6b/model_api/python/model_api/models/ssd.py#L78)). To match the results, use `confidence_threshold` from checkpoint before wrapping outputs with `DetBatchPredEntity`.
    figure order : rtmdet_tiny OV Model | rtmdet_tiny Torch Model
    image

### Checklist

### License

```python
```


eunwoosh (dismissed):

src/otx/algo/detection/atss.py
```diff
@@ -116,15 +116,17 @@ def _customize_outputs(
         for img_info, prediction in zip(inputs.imgs_info, predictions):
             if not isinstance(prediction, InstanceData):
                 raise TypeError(prediction)
-            scores.append(prediction.scores)  # type: ignore[attr-defined]
+
+            filtered_idx = torch.where(prediction.scores > self.hparams.get("best_confidence_threshold", 0.5))  # type: ignore[attr-defined]
```
harimkang: If there is no `best_confidence_threshold` here, what do you think about `warning` or `logging` that it is used as 0.5?

chuneuny-emily (dismissed):

src/otx/algo/detection/atss.py
sungchul1: Done https://github.com/openvinotoolkit/training_extensions/pull/3634/commits/5ede7974abc3b840c06054c54776424fb7c79240

harimkang (approved):

eunwoosh (approved):

PR #3638, opened 2024-06-18
Refactor Visual Prompting OV model to use MAPI

sovrasov:
### Summary

Inference functionality was moved to MAPI

### Checklist

### License

```python
```


sovrasov:
e2e: https://github.com/openvinotoolkit/training_extensions/actions/runs/9649616346

src/otx/core/model/visual_prompting.py
```diff
@@ -577,9 +582,14 @@ def _customize_outputs(
...
-            masks.append(torch.as_tensor(output["hard_prediction"], device=self.device))
-            scores.append(torch.as_tensor(output["scores"], device=self.device))
+        for image_output in outputs:
+            masks.extend(
+                [
+                    torch.as_tensor(hard_prediction, device=self.device)
+                    for hard_prediction in image_output.hard_predictions
+                ],
+            )
+            scores.extend([torch.as_tensor(score, device=self.device) for score in image_output.scores])
```
sungchul1, line 592: It seems that `len(masks)` and `len(scores)` aren't matched with `len(outputs)`, doesn't it?

src/otx/core/model/visual_prompting.py
```diff
@@ -780,162 +813,81 @@ def __init__(
...
         reset_feat: bool = False,
-        default_threshold_reference: float = 0.3,
-        is_cascade: bool = False,
     ) -> tuple[dict[str, np.ndarray], list[np.ndarray]]:
         """`Learn` for reference features."""
-        if reset_feat or self.reference_feats is None:
-            self.initialize_reference_info()
+        images, processed_prompts = self._customize_inputs(inputs)
+        reference_masks: list[np.ndarray] = []
+        reference_features = VisualPromptingFeatures(np.array(0), np.array(0))
```
sungchul1, line 871: Is it needed to initialize `reference_features`?

src/otx/core/model/visual_prompting.py
```diff
@@ -780,162 +813,81 @@ def __init__(
...
 
-        images, metas, processed_prompts = self._customize_inputs(inputs)
-        largest_label: int = max(sum([[int(p) for p in prompt] for prompt in processed_prompts], []))
-        self.expand_reference_info(largest_label)
+        for image, prompts in zip(images, processed_prompts):
+            reference_features, masks = self.model.learn(image, reset_features=reset_feat, **prompts)
```
sungchul1, line 874: Reference features per label are overridden by the latest input, right?

src/otx/core/model/visual_prompting.py
```diff
@@ -966,47 +918,38 @@ def forward(  # type: ignore[override]
...
+                bboxes: list[Prompt] = []
                 for prompt, label in zip(prompts, labels):  # type: ignore[arg-type]
                     if isinstance(prompt, tv_tensors.BoundingBoxes):
-                        bboxes.append(prompt.cpu().numpy())
-                        _labels["bboxes"].append(label.cpu().numpy())
+                        bboxes.append(Prompt(prompt.cpu().numpy(), label.cpu().numpy()))
                     elif isinstance(prompt, Points):
-                        points.append(prompt.cpu().numpy())
-                        _labels["points"].append(label.cpu().numpy())
+                        points.append(Prompt(prompt.cpu().numpy(), label.cpu().numpy()))
```
sungchul1, line 947: Could you leave TODO comment about polygon support?

src/otx/core/model/visual_prompting.py
sovrasov: Yes, it's a good practice to store a safe initializer value in a variable which is going to be modified depending on some condition.

src/otx/core/model/visual_prompting.py
sovrasov: Right, learn returns the latest state of features, now MAPI class manages features extension

src/otx/core/model/visual_prompting.py
sovrasov: You're right. Each `image_output` contains per prompt outputs. Previously, outputs from all images and all prompt were flattened into `outputs` list, while now we have 2 levels structure, and need to extend the list to imitate the previous flattened result.

src/otx/core/model/visual_prompting.py
sovrasov: Done

src/otx/core/exporter/visual_prompting.py
```diff
@@ -93,12 +95,17 @@ def to_openvino(
...
                 base_model_name,
                 OTXPrecisionType.FP32,
-                False,
+                embed_metadata=False,
             )
             exported_model = openvino.convert_model(tmp_dir / (base_model_name + ".onnx"))
 
         exported_model = self._postprocess_openvino_model(exported_model)
 
+        if self.metadata is not None:
```
eunwoosh, line 104: It seems that this part can be executed in `self._postprocess_openvino_model` if you update `src/otx/core/types/export.py`. How about doing that to reduce duplicated code?

src/otx/core/exporter/visual_prompting.py
```diff
@@ -136,7 +144,12 @@ def to_onnx(
         )
 
         onnx_model = onnx.load(save_path)
-        onnx_model = self._postprocess_onnx_model(onnx_model, embed_metadata, precision)
+        onnx_model = self._postprocess_onnx_model(onnx_model, False, precision)
+
+        if self.metadata is not None and embed_metadata:
```
eunwoosh, line 149: Same as above.

src/otx/recipe/visual_prompting/openvino_model.yaml
```diff
@@ -5,7 +5,7 @@ model:
     model_name: segment_anything
     model_type: Visual_Prompting
     async_inference: False
-    use_throughput_mode: True
+    use_throughput_mode: False
```
eunwoosh, line 8: It's just a question. Why is this value set to `False`?

eunwoosh (commented):

src/otx/core/model/visual_prompting.py
sungchul1: Sorry for that because it seems implementation bug.
Currently, batch size for test is 1, so it doesn't matter to use flattened outputs.
However, for multi batch testing, entity has batch-wise components (masks and scores), not flattened one.
I'll fix it quickly, please reflect it.

sungchul1: https://github.com/openvinotoolkit/training_extensions/pull/3695

sovrasov: Let's fix after this PR is merged

src/otx/recipe/visual_prompting/openvino_model.yaml
sovrasov: Throughput mode only makes sense when async inference is executed, which is not the case for VPT now

src/otx/core/exporter/visual_prompting.py
sovrasov: No, it can not, unless we introduce more global changes to exporter architecture. 
We have two models here, and therefore two difference `model_type` parameters should be written.
Generalizing that, we should extend export to the case of multiple IRs per model, each of those could require an individual set of export parameters. Also, this is the only why we have a standalone exporter for visual prompting.

sovrasov: For now, I'd like to keep the core exporter clean, while making workarounds as explicit as possible, at the cost of some duplication though.
Here we can also temporally mock `self.metadata`, but as I said let's keep workarounds explicit.

eunwoosh (commented):
LGTM. Could you update integration test?

sovrasov:
tox -vv -e integration-test-visual_prompting
image

eunwoosh (approved):

sungchul1 (approved):
LGTM, but I think benchmark test results are also needed to check if the performance is reproduced.

kprokofi (approved):

sovrasov:
otx test with OV ZSL model on coco_car_person_medium data
```
# this PR

│      test/data_time       │   0.0006355935474857688   │
│         test/dice         │    0.6657536029815674     │
│       test/f1-score       │    0.6657536029815674     │
│         test/iou          │    0.4989734888076782     │
│      test/iter_time       │    0.7157989144325256     |
Elapsed time: 0:01:50.238978

# develop

│      test/data_time       │   0.0006135150324553251   │
│         test/dice         │    0.6657536029815674     │
│       test/f1-score       │    0.6657536029815674     │
│         test/iou          │    0.4989734888076782     │
│      test/iter_time       │    0.7240481376647949

Elapsed time: 0:01:51.517192
```

PR #3647, opened 2024-06-20
Add Semi-SL for cls Benchmark Test

harimkang:
### Summary

- Add TVModels into Singe-label Benchmark Test
- Add Semi-SL Benchmark Test

### Checklist

### License

```python
```

### DIFF
--- a/tests/perf/test_classification.py
+++ b/tests/perf/test_classification.py
@@ -22,6 +22,9 @@ class TestPerfSingleLabelClassification(PerfTestBase):
         Benchmark.Model(task="classification/multi_class_cls", name="mobilenet_v3_large", category="accuracy"),
         Benchmark.Model(task="classification/multi_class_cls", name="deit_tiny", category="other"),
         Benchmark.Model(task="classification/multi_class_cls", name="dino_v2", category="other"),
+        Benchmark.Model(task="classification/multi_class_cls", name="tv_efficientnet_b3", category="other"),
+        Benchmark.Model(task="classification/multi_class_cls", name="tv_efficientnet_v2_l", category="other"),
+        Benchmark.Model(task="classification/multi_class_cls", name="tv_mobilenet_v3_small", category="other"),
     ]
 
     DATASET_TEST_CASES = [
@@ -267,6 +270,11 @@ class TestPerfSemiSLMultiClass(PerfTestBase):
         Benchmark.Model(task="classification/multi_class_cls", name="efficientnet_b0_semisl", category="speed"),
         Benchmark.Model(task="classification/multi_class_cls", name="mobilenet_v3_large_semisl", category="speed"),
         Benchmark.Model(task="classification/multi_class_cls", name="efficientnet_v2_semisl", category="accuracy"),
+        Benchmark.Model(task="classification/multi_class_cls", name="deit_tiny_semisl", category="other"),
+        Benchmark.Model(task="classification/multi_class_cls", name="dino_v2_semisl", category="other"),
+        Benchmark.Model(task="classification/multi_class_cls", name="tv_efficientnet_b3_semisl", category="other"),
+        Benchmark.Model(task="classification/multi_class_cls", name="tv_efficientnet_v2_l_semisl", category="other"),
+        Benchmark.Model(task="classification/multi_class_cls", name="tv_mobilenet_v3_small_semisl", category="other"),
     ]
 
     DATASET_TEST_CASES = (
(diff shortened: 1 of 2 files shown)


chuneuny-emily (approved):

sungchul1 (approved):

tests/perf/test_classification.py
```diff
@@ -258,3 +261,121 @@ def test_perf(
...
+        Benchmark.Model(task="classification/multi_class_cls", name="tv_efficientnet_v2_l_semisl", category="other"),
+        Benchmark.Model(task="classification/multi_class_cls", name="tv_mobilenet_v3_small_semisl", category="other"),
+    ]
+
+    DATASET_TEST_CASES = (
+        [
+            Benchmark.Dataset(
+                name=f"cifar10@{num_label}_{idx}",
+                path=Path(f"multiclass_classification/semi-sl/cifar10@{num_label}_{idx}/supervised"),
+                group="cifar10",
```
eunwoosh, line 285: Group should be one of "small", "medium", "large". Could you update it?

tests/perf/test_classification.py
```diff
@@ -258,3 +261,121 @@ def test_perf(
...
+                },
+            )
+            for idx in (0, 1, 2)
+            for num_label in (4, 10, 25)
+        ]
+        + [
+            Benchmark.Dataset(
+                name=f"svhn@{num_label}_{idx}",
+                path=Path(f"multiclass_classification/semi-sl/svhn@{num_label}_{idx}/supervised"),
+                group="svhn",
```
eunwoosh, line 304: Same as above.

tests/perf/test_classification.py
```diff
@@ -258,3 +261,121 @@ def test_perf(
...
+                },
+            )
+            for idx in (0, 1, 2)
+            for num_label in (4, 10, 25)
+        ]
+        + [
+            Benchmark.Dataset(
+                name=f"fmnist@{num_label}_{idx}",
+                path=Path(f"multiclass_classification/semi-sl/fmnist@{num_label}_{idx}/supervised"),
+                group="fmnist",
```
eunwoosh, line 323: Same as above.

tests/perf/test_classification.py
```diff
@@ -258,3 +261,121 @@ def test_perf(
...
+                extra_overrides={
+                    "train": {
+                        "data.config.train_subset.subset_name": "train_data",
+                        "data.config.val_subset.subset_name": "val_data",
+                        "data.config.test_subset.subset_name": "val_data",
+                        "deterministic": "True",
+                    },
+                },
+            )
+            for idx in (0, 1, 2)
```
eunwoosh, line 297: If current configuration is used, 9 cases of every each dataset is tested with every models, which can make performence test quite longer. I think it's better to divide dataset in to 3 types("small", "medium", "large") and each type is corresponding to single case. How do you think about it?

tests/perf/test_classification.py
```diff
@@ -258,3 +261,121 @@ def test_perf(
             criteria=self.BENCHMARK_CRITERIA,
             resume_from=fxt_resume_from,
         )
+
+
+class TestPerfSemiSLMultiClass(PerfTestBase):
+    """Benchmark single-label classification for Semi-SL task."""
+
+    MODEL_TEST_CASES = [  # noqa: RUF012
+        Benchmark.Model(task="classification/multi_class_cls", name="efficientnet_b0_semisl", category="speed"),
```
eunwoosh, line 270: It seems that two models are set to `speed` and no model for `balance`. Is this your intention?

eunwoosh (commented):
Thanks for your work. I left some comments. Could you check them?

tests/perf/test_classification.py
harimkang: I discussed this with Emily and we decided to go this way.

tests/perf/test_classification.py
harimkang: As far as I know, this is just a str type to collect the results in Summary, so I think it's okay to leave it as it is, but what do you think? Will this cause errors elsewhere?

The dataset used here doesn't have a clear distinction between small, medium, and large.

tests/perf/test_classification.py
harimkang: I'll update this later when see the results.

tests/perf/test_classification.py
eunwoosh: Currently, only "small", "medium", "large" and "all" can be passed to `data_group` arguments. Please check "tests/conftest.py". So, only if "all" is passed to the argument, these datasets won't be skipped. If you think it's ok, then I have objection to leave it as it is.

PR #3652, opened 2024-06-20
Update benchmark tool for otx2.1

chuneuny-emily:
### Summary
This PR makes the following improvements:

1. **Deterministic setting optimization**: Eliminated redundant deterministic configurations, using user settings when available or defaulting to task-specific values.

2. **Add newly added models in OTX2.1**: Added rtmdet for detection and instance segmentation, and maskrcnn_r50_tv for instance segmentation.

### Checklist

### License

```python
```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -94,7 +94,7 @@ def pytest_addoption(parser: pytest.Parser):
     )
     parser.addoption(
         "--deterministic",
-        action="store",
+        choices=["true", "false"],
         default=None,
         help="Turn on deterministic training (true/false).",
     )
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -94,9 +94,9 @@ def pytest_addoption(parser: pytest.Parser):
     )
     parser.addoption(
         "--deterministic",
-        action="store_true",
-        default=False,
-        help="Turn on deterministic training.",
+        action="store",
+        default=None,
+        help="Turn on deterministic training (true/false).",
     )
     parser.addoption(
         "--user-name",
--- a/tests/perf/test_action.py
+++ b/tests/perf/test_action.py
@@ -20,10 +20,7 @@ log = logging.getLogger(__name__)
 def fxt_deterministic(request: pytest.FixtureRequest) -> bool:
     """Override the deterministic setting for action classification task."""
     deterministic = request.config.getoption("--deterministic")
-    if deterministic is None:
-        deterministic = True
-    else:
-        deterministic = deterministic.lower() == "true"
+    deterministic = True if deterministic is None else deterministic.lower() == "true"
     msg = f"deterministic={deterministic}"
     log.info(msg)
     return deterministic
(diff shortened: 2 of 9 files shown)


tests/perf/test_action.py
```diff
@@ -5,13 +5,26 @@
...
+log = logging.getLogger(__name__)
+
+
+@pytest.fixture(scope="session")
+def fxt_deterministic(request: pytest.FixtureRequest) -> bool:
+    """Override the deterministic setting for action classification task."""
+    deterministic = request.config.getoption("--deterministic")
+    deterministic = True if deterministic is None else deterministic.lower() == "true"
+    msg = f"deterministic={deterministic}"
+    log.info(msg)
```
eunwoosh:
```suggestion
    log.info(f"{deterministic=}")
```
f"{deterministic=}" is same as "deterministic=${deterministic}"

tests/perf/test_anomaly.py
```diff
@@ -5,32 +5,40 @@
...
 
 import pytest
 
 from .benchmark import Benchmark
 from .conftest import PerfTestBase
 
+log = logging.getLogger(__name__)
+
+
+@pytest.fixture(scope="session")
```
eunwoosh: It seems that default value of the deterministic is consider as `False`. Then, how about using this code at `tests/perf/conftest.py::fxt_deterministic` and using that fixture? If then, we don't have to override it in every tasks.

tests/conftest.py
```diff
@@ -94,9 +94,9 @@ def pytest_addoption(parser: pytest.Parser):
     )
     parser.addoption(
         "--deterministic",
-        action="store_true",
-        default=False,
-        help="Turn on deterministic training.",
+        action="store",
```
eunwoosh:
```suggestion
        choice=["true", "false", None],
```
If value should be one of "true", "false", None, then how about using choice?

eunwoosh (commented):

sungchul1 (dismissed):

tests/conftest.py
chuneuny-emily: user can set it to true or false. default value is None. If user does not set deterministic, it will be used the task specific fixed value.

tests/perf/test_anomaly.py
chuneuny-emily: No. default deterministic value is None. If user does not set deterministic, it will be used the task specific fixed value.

chuneuny-emily: Each task has the default deterministic value differently. e.g. true for action, classification and detection. false for anomaly, ins seg, sseg and vp.

tests/conftest.py
eunwoosh: yes right. so what user can set to `deterministic` is one of three values, right? I think it's better to prevent user from setting other value.

tests/perf/test_action.py
chuneuny-emily: ok. can I change it later? I'll keep in mind this compact code.

eunwoosh (approved):

sungchul1 (approved):

PR #3710, opened 2024-07-04
Add HuggingFace Model Wrapper for Multi-Class Classification

harimkang:
### Summary

- Add HuggingFaceModelForMulticlassCls
- Training, Testing, Exporting works

image
*The results of running the existing models are from workstation, and for the hugging face model, the results are from a different machine, so it is difficult to compare the time directly. Just a note.

### Checked models
- "nateraw/vit-base-beans", # Available
- "google/vit-base-patch16-224", # Available
- "google/mobilenet_v1_1.0_224", # Available
- "google/efficientnet-b0", # Available
- "microsoft/beit-base-patch16-224-pt22k-ft22k", # Available
- "microsoft/swin-tiny-patch4-window7-224", # Available
- "microsoft/resnet-18", # Available **(but show NAN loss)**
- "facebook/convnextv2-tiny-1k-224", # Available
- "facebook/dinov2-small-imagenet1k-1-layer", # Available
- "facebook/deit-tiny-patch16-224", # Available

I haven't checked them all (too many models), but most of the models in the list below are available.
https://huggingface.co/models?pipeline_tag=image-classification

image

### How to test

Install transformers
`pip install transformers`

API Examples
```python
from otx.algo.classification.huggingface_model import HuggingFaceModelForMulticlassCls
from otx.engine import Engine

data_root = "/datasets/otx_v2_dataset/multiclass_classification/multiclass_food101_large"

otx_model = HuggingFaceModelForMulticlassCls(
    model_name_or_path="facebook/deit-tiny-patch16-224",
    label_info=20,
)

engine = Engine(
    data_root=data_root,
    model=otx_model,
    work_dir="otx-workspace-hf",
)

engine.train(max_epochs=2)
engine.test()
engine.export()
engine.test(checkpoint="otx-workspace-hf/exported_model.xml")
```

CLI Examples
```shell
(venv) ~/otx$ otx train \
    --model otx.algo.classification.huggingface_model.HuggingFaceModelForMulticlassCls \
    --model.model_name_or_path facebook/deit-tiny-patch16-224 \
    --data_root /datasets/otx_v2_dataset/multiclass_classification/multiclass_food101_large \
    --work_dir otx-workspace-hf-cli --max_epochs 3
```

### Checklist

### License

```python
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -103,6 +103,10 @@ base = [
     "anomalib[core]==1.1.0",
 ]
 
+transformers = [
+    "transformers==4.42.3",
+]
+
 mmlab = [
     "mmdet==3.2.0",
     "mmpretrain==1.1.1",
--- a/src/otx/algo/classification/huggingface_model.py
+++ b/src/otx/algo/classification/huggingface_model.py
@@ -1,0 +1,130 @@
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+"""Hugging-Face pretrained model for the OTX classification."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+import torch
+from torch import Tensor, nn
+from transformers import AutoModelForImageClassification
+
+from otx.core.data.entity.base import OTXBatchLossEntity
+from otx.core.data.entity.classification import (
+    MulticlassClsBatchDataEntity,
+    MulticlassClsBatchPredEntity,
+)
+from otx.core.exporter.base import OTXModelExporter
+from otx.core.exporter.native import OTXNativeModelExporter
+from otx.core.metrics.accuracy import MultiClassClsMetricCallable
+from otx.core.model.base import DefaultOptimizerCallable, DefaultSchedulerCallable
+from otx.core.model.classification import OTXMulticlassClsModel
+from otx.core.schedulers import LRSchedulerListCallable
+from otx.core.types.label import LabelInfoTypes
+
+if TYPE_CHECKING:
+    from lightning.pytorch.cli import LRSchedulerCallable, OptimizerCallable
+    from transformers.modeling_outputs import ImageClassifierOutput
+
+    from otx.core.metrics import MetricCallable
+
+
+class HuggingFaceModelForMulticlassCls(OTXMulticlassClsModel):
+    """HuggingFaceModelForMulticlassCls is a class that represents a Hugging Face model for multiclass classification.
+
+    Args:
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -103,6 +103,10 @@ base = [
     "anomalib[core]==1.1.0",
 ]
 
+transformers = [
```
eunwoosh, line 106:
```suggestion
huggingface = [
```
It's for enabling huggingface, so how about to change like this?

eunwoosh (commented):
Thanks for your work :) I left minor comment, Please take a look.

pyproject.toml
harimkang: This is ambiguous, but the library is actually a `transformers` integration. It's an integration of the transformers api, which supports the hugging face model, so I think transformers makes more sense for me. What do you think?

eunwoosh: I see. my thought is that the goal of integrating a `transformers` is for using huggingface model and name `transformers` can mean various thing. So, I think when user wants to use huggingface model, using `pip install otx[huggingface]` is more intuitive. This is just my thought. If you think different please resolve this :)

eunwoosh (approved):

src/otx/algo/classification/huggingface_model.py
sungchul1, line 1: Of course mmx will be removed, but what do you think about that gathering 3rd party frameworks' (huggingface and torchvision) models into `otx/core/model/classification.py`?
Like
https://github.com/openvinotoolkit/training_extensions/blob/e67a5fcd17fb350191771ed79dd8d4ed246c0b10/src/otx/core/model/classification.py#L90-L233

sungchul1: Actually, there are usually model-related files in task-specific directory:
image
I think supporting various frameworks is optional, isn't it?
If no, feel free to ignore it :D

harimkang: I'm not sure, I think `core` should be a parent class that can be used as a base, and `algo` will be the actual model implementation. (As a user, it seems appropriate to use the `algo`'s model class to build the model.) Also, `transformers` should be installed as optional when I discussed with Kirill, it is not installed in the current test env (not `base`). So I still think this is better off in an `algo`. @kprokofi what do you think?

pyproject.toml
harimkang: Yeah we'll have to think about this a bit more, thanks for the comment!

sungchul1 (approved):

src/otx/algo/classification/huggingface_model.py
kprokofi: Yeah, I agree with Harim, we should use core/model as the base. I also thinking about removing MMmodels from core.

PR #3723, opened 2024-07-11
Fix Combined Dataloader & unlabeled warmup loss to work correctly in Semi-SL case

harimkang:
### Summary

Fixes some edge cases showing unintended behavior.
1. If the number of iters in the unlabeled dataloader is less than the number of iters in the labeled dataloader, find that combined dataloader are acting on the basis of the unlabeled dataloader and fix this.
2. Found that unlabeled warmup loss in Semi-SL was not working properly and had a shorter warmup time than expected, so fix this.

<head>

</head>

<div style='direction:ltr'>

Model | Method | Test/Accuracy | E2E Time | Epochs
-- | -- | -- | -- | --
EfficientNet-B0 | SL     (Pseudo-Labeled dataset) | 0.5863 | 0:00:27 | 24
  | Semi-SL     (batch size: 16/48) | 0.4675 | 0:08:55 | 60
  | Semi-SL #Hot-Fix     (batch size: 16/48) | 0.6058 | 0:17:24 | 115

</div>

</body>

</html>

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algo/callbacks/unlabeled_loss_warmup.py
+++ b/src/otx/algo/callbacks/unlabeled_loss_warmup.py
@@ -36,7 +36,12 @@ class UnlabeledLossWarmUpCallback(Callback):
         """Calculate the unlabeled warm-up loss coefficient before training iteration."""
         if self.unlabeled_coef < 1.0:
             if self.total_steps == 0:
-                self.total_steps = int(trainer.max_epochs * len(trainer.train_dataloader) * self.warmup_steps_ratio)
+                dataloader = (
+                    trainer.train_dataloader["labeled"]
+                    if isinstance(trainer.train_dataloader, dict)
+                    else trainer.train_dataloader
+                )
+                self.total_steps = int(trainer.max_epochs * len(dataloader) * self.warmup_steps_ratio)
             self.unlabeled_coef = 0.5 * (
                 1 - math.cos(min(math.pi, (2 * math.pi * self.current_step) / self.total_steps))
             )
--- a/src/otx/core/data/module.py
+++ b/src/otx/core/data/module.py
@@ -246,9 +246,7 @@ class OTXDataModule(LightningDataModule):
 
     def _is_meta_info_valid(self, label_infos: list[LabelInfo]) -> bool:
         """Check whether there are mismatches in the metainfo for the all subsets."""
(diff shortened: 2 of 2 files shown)


sungchul1 (approved):
LGTM, but could you explain if it is okay because Semi-SL takes a much longer training time than SL (about 39x based on the table) but has less performance improvement?

harimkang:
> LGTM, but could you explain if it is okay because Semi-SL takes a much longer training time than SL (about 39x based on the table) but has less performance improvement?

For the table I used here, the experimental dataset is a bit unusual and relatively large, so I don't expect the change in PR to make much of a difference. Other than the change in the coefficient of the unlabeled loss, the schedule setting itself doesn't make a difference, so I expect it to behave more epochs if it can get more gain, but not less. I also expect this to show similar behavior to the minimum SL.

eunwoosh (approved):

sungchul1:
> > LGTM, but could you explain if it is okay because Semi-SL takes a much longer training time than SL (about 39x based on the table) but has less performance improvement?
> 
> For the table I used here, the experimental dataset is a bit unusual and relatively large, so I don't expect the change in PR to make much of a difference. Other than the change in the coefficient of the unlabeled loss, the schedule setting itself doesn't make a difference, so I expect it to behave more epochs if it can get more gain, but not less. I also expect this to show similar behavior to the minimum SL.

Could you explain why you used unusual and relatively large dataset? To check the feasibility, I think common dataset should be used to accurately check improvement. And I'd like to suggest the experiments with two versions of datasets: #labeled > #unlabeled and #labeled < #unlabeled to see robust experimental results. What do you think?

harimkang:
> Could you explain why you used unusual and relatively large dataset? To check the feasibility, I think common dataset should be used to accurately check improvement. And I'd like to suggest the experiments with two versions of datasets: #labeled > #unlabeled and #labeled < #unlabeled to see robust experimental results. What do you think?

This PR is a fix for the edge case, and there is little change in the typical Semi-SL case. This is an issue I discovered while experimenting with edge cases. Also, the results shared in the existing Semi-SL are mostly #labeled < #unlabeled, while the tables currently shared in this PR are #labeled > #unlabeled.
Anyway, I will confirm this with a benchmark test in the future. Thanks for the comment.

PR #3727, opened 2024-07-12
Disable mixed preicision to NMS on CPU

eunwoosh:
### Summary
when using mixed precision on CPU, lightning uses bfp16 because torch amp doesn't support fp16 on CPU.
But torch nms kernel doesn't support bfp16, which makes error when running detection models on CPU.
This PR fixes it.

**Integration test result**
image
detection intg test : pytest tests/integration -ra --showlocals --task detection --open-subprocess
image
ins seg intg test : pytest tests/integration -ra --showlocals --task instance_segmentation --open-subprocess

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algo/common/utils/nms.py
+++ b/src/otx/algo/common/utils/nms.py
@@ -39,7 +39,11 @@ class NMSop(torch.autograd.Function):
             valid_mask = scores > score_threshold
             bboxes, scores = bboxes[valid_mask], scores[valid_mask]
             valid_inds = torch.nonzero(valid_mask, as_tuple=False).squeeze(dim=1)
-        inds = torch_nms(bboxes, scores, iou_threshold)
+        if bboxes.get_device() == -1 and bboxes.dtype == torch.bfloat16:  # torch nms kernel doesn't support bfloat16
+            with torch.autocast(device_type="cpu", dtype=torch.bfloat16, enabled=False):
+                inds = torch_nms(bboxes.float(), scores.float(), iou_threshold)
+        else:
+            inds = torch_nms(bboxes, scores, iou_threshold)
 
         if max_num > 0:
             inds = inds[:max_num]
--- a/tests/unit/algo/common/utils/test_nms.py
+++ b/tests/unit/algo/common/utils/test_nms.py
@@ -133,7 +133,7 @@ class TestNMSop:
         assert (mock_torch_nms.call_args[0][1] == mock_scores[0]).all()
         assert list(nms) == [0]
 
-    def test_forward_cpu_bfx(
+    def test_forward_cpu_bf16(
         self,
         mock_torch_nms: MagicMock,
         mock_torch_autocast: MagicMock,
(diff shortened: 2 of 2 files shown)


src/otx/algo/common/utils/nms.py
```diff
@@ -39,7 +39,11 @@ def forward(
             valid_mask = scores > score_threshold
             bboxes, scores = bboxes[valid_mask], scores[valid_mask]
             valid_inds = torch.nonzero(valid_mask, as_tuple=False).squeeze(dim=1)
-        inds = torch_nms(bboxes, scores, iou_threshold)
+        if bboxes.get_device() == -1 and bboxes.dtype == torch.bfloat16:  # torch nms kernel doesn't support bfloat16
+            with torch.autocast(device_type="cpu", dtype=torch.bfloat16, enabled=False):
+                inds = torch_nms(bboxes.float(), scores.float(), iou_threshold)
```
harimkang, line 44: Is there any way we can throw a message in this case?

harimkang: Looking at the current code, I think the `NMSop` class could hold certain flags (flag for needs of `autocast`) and print the correct message when the first flag changes. What do you think, is this possible?

eunwoosh: > Is there any way we can throw a message in this case?

yes It's not hard, but I'm not sure it is needed. because mixed precision automatically changed precision in available operation by torch originally and nms is not supported operation for CPU. So I think whether autocast is disabled implicitly or explicitly, they are same.

> Looking at the current code, I think the `NMSop` class could hold certain flags (flag for needs of `autocast`) and print the correct message when the first flag changes. What do you think, is this possible?

I'm not sure I understand well what you said, but if you mean need for keeping a flag to decide to autocast or not, I think this thing should be handled by torch side. If I misunderstood, please correct me :)

src/otx/algo/common/utils/nms.py
```diff
@@ -39,7 +39,11 @@ def forward(
             valid_mask = scores > score_threshold
             bboxes, scores = bboxes[valid_mask], scores[valid_mask]
             valid_inds = torch.nonzero(valid_mask, as_tuple=False).squeeze(dim=1)
-        inds = torch_nms(bboxes, scores, iou_threshold)
+        if bboxes.get_device() == -1 and bboxes.dtype == torch.bfloat16:  # torch nms kernel doesn't support bfloat16
+            with torch.autocast(device_type="cpu", dtype=torch.bfloat16, enabled=False):
```
sungchul1, line 43: Just question, is `torch.autocast` needed because `bboxes` and `scores` were created in autocast-enabled region?

eunwoosh: I added it to be sure that automatic cast is disabled in nms operations. I refered [this](https://pytorch.org/docs/stable/amp.html#autocasting) when I implemented.

harimkang (dismissed):
Can we add a unit-test to make sure this works?

harimkang (approved):

sungchul1 (approved):

PR #3646, opened 2024-06-19
Upgrade NNCF and OV versions

sovrasov:
### Summary
OV and NNCF versions are promoted

PR #3736, opened 2024-07-15
Enable direct export of ViTs

sovrasov:
### Summary

Redundant ONNX conversion step is eliminated. This also avoids extra FS usage


sovrasov:
Need to merge https://github.com/openvinotoolkit/training_extensions/pull/3646 first [done]

sovrasov:
@harimkang please have a look at 5a50259. We have a couple of unexpected outputs on export introduced in https://github.com/openvinotoolkit/training_extensions/commit/e402a84f99187302ce8b72576c3fd668cff3dd4e

src/otx/algo/classification/vit.py
```diff
@@ -140,14 +139,18 @@ def _forward_explain_image_classifier(
...
-        return {
+        outputs = {
             "logits": logits,
             "feature_vector": feature_vector,
             "saliency_map": saliency_map,
-            "scores": scores,
-            "labels": labels,
         }
 
+        if not torch._C._is_tracing():  # noqa: SLF001
```
eunwoosh:
```suggestion
        if not torch.jit.is_tracing():  # noqa: SLF001
```
Is it ok to use as above? If it's, how about avoiding to use internal funciton?

sovrasov: Updated

eunwoosh (approved):

kprokofi (approved):

PR #3746, opened 2024-07-16
Add '--test-only' flag in perf test

eunwoosh:
### Summary
This PR adds `--test-only` flags to perf test. When it's set, perf test copies all necessary files from resume directory and executes only `otx test` with specified operation. If files can't be found from resume directory, necessary operations will be executed automatically.
Additionally, bug that error is raised when setting device to `cpu` is fixed.

### Checklist

### License

```python
```

### DIFF
--- a/tests/perf/benchmark.py
+++ b/tests/perf/benchmark.py
@@ -51,9 +51,12 @@ class Benchmark:
         deterministic (bool): Whether to turn on deterministic training mode. Defaults to False.
         accelerator (str): Accelerator device on which to run benchmark. Defaults to gpu.
         reference_results (pd.DataFrame): Reference benchmark results for performance checking.
-        resume(Path | None, optional):
+        resume (Path | None, optional):
             Previous performance directory to load. If training was already done in previous performance test,
             training is skipped and refer previous result.
+        resume_from (str):
+            Which otx command to start from when resume argument is given.
+            Commands executed before `resume_from` are skipped and necessary OV IRs are copied from resume directory.
     """
 
     @dataclass
--- a/tests/perf/benchmark.py
+++ b/tests/perf/benchmark.py
@@ -314,7 +314,9 @@ class Benchmark:
         prev_otx_cmd = None
         for otx_cmd in ["train", "export", "optimize"]:
             prev_symlink = prev_work_dir / ".latest" / otx_cmd
-            if not prev_symlink.exists():
+            try:  # check symlink exists
+                prev_symlink.resolve(strict=True)
+            except FileNotFoundError:
                 break
             prev_cmd_dir_name = prev_symlink.resolve().name
             prev_cmd_dir = prev_work_dir / prev_cmd_dir_name
--- a/tests/perf/benchmark.py
+++ b/tests/perf/benchmark.py
@@ -51,12 +51,12 @@ class Benchmark:
         deterministic (bool): Whether to turn on deterministic training mode. Defaults to False.
         accelerator (str): Accelerator device on which to run benchmark. Defaults to gpu.
         reference_results (pd.DataFrame): Reference benchmark results for performance checking.
-        resume (Path | None, optional):
+        resume_from (Path | None, optional):
             Previous performance directory to load. If training was already done in previous performance test,
             training is skipped and refer previous result.
-        resume_from (str):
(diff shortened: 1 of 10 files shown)


goodsong81 (commented):
Isn't it a little bit misleading not executing `otx export` when specifying `--resume-from export`? It sounds like rather `--resume-after` to me.
(Not saying that it should be renamed to "after", thouth)

eunwoosh:
> Isn't it a little bit misleading not executing `otx export` when specifying `--resume-from export`? It sounds like rather `--resume-after` to me. (Not saying that it should be renamed to "after", thouth)

you're right, but it's hard to find proper name. So far, `--resume-from` is best candidate, I think. I'll think more about that.

kprokofi (commented):
I don't understand, why should we add second option with resume_from. Why only one resume is not enough to continue training? Let's imagine we have some workspace with already completed train stage for small datasets. Can we just write: pytest test_*.py --work-dir /path/to/it --eval-upto optimize --resume. After that I would expect that export -> optimize on small datasets will be executed

It looks complicated for me with two resumes

eunwoosh:
> I don't understand, why should we add second option with resume_from. Why only one resume is not enough to continue training? Let's imagine we have some workspace with already completed train stage for small datasets. Can we just write: pytest test_*.py --work-dir /path/to/it --eval-upto optimize --resume. After that I would expect that export -> optimize on small datasets will be executed
> 
> It looks complicated for me with two resumes

It's for saving more time when running huge number of models. Current perf test resume feature just skips training part. So, if user uses `--resume-from`, then perf test starts from `otx test` after training. Of course, skipping training part can save a lot of time, but sometimes skipping additional parts is preferred. For examples, recently I tried to benchmark score using all OTX models' optimized OV IR. With current resume feature, I need to run `otx test` 15 times using trained model, OV IR and quantized OV IR for every model (repeat 5 x 3 `otx test`). In this case, skipping `otx test` w/ trained model and OV IR can save a lot of time. And it can be beneficial when using environment w/o any acclerator where `otx test` w/ trained model spends lots of time.

goodsong81:
> > I don't understand, why should we add second option with resume_from. Why only one resume is not enough to continue training? Let's imagine we have some workspace with already completed train stage for small datasets. Can we just write: pytest test_*.py --work-dir /path/to/it --eval-upto optimize --resume. After that I would expect that export -> optimize on small datasets will be executed
> > It looks complicated for me with two resumes
> 
> It's for saving more time when running huge number of models. Current perf test resume feature just skips training part. So, if user uses `--resume-from`, then perf test starts from `otx test` after training. Of course, skipping training part can save a lot of time, but sometimes skipping additional parts is preferred. For examples, recently I tried to benchmark score using all OTX models' optimized OV IR. With current resume feature, I need to run `otx test` 15 times using trained model, OV IR and quantized OV IR for every model (repeat 5 x 3 `otx test`). In this case, skipping `otx test` w/ trained model and OV IR can save a lot of time. And it can be beneficial when using environment w/o any acclerator where `otx test` w/ trained model spends lots of time.

I think @kprokofi's comment makes sense. Isn't it some kind of test-only feature that you really need?
Why don't you add some flag like "--test train|export|optimize|all"?

eunwoosh:
I understand what you said. I'll revert "--resume" to "--resume-from" and change "--resume-from" to "--test-only".

tests/perf/benchmark.py
```diff
@@ -128,23 +136,26 @@ def __init__(
         self.deterministic = deterministic
         self.accelerator = accelerator
         self.reference_results = reference_results
+        self.resume_from = resume_from
+        if (test_only == "export" and eval_upto == "train") or (
```
sungchul1, line 140: What do you think about using enum instead of comparing strings?

eunwoosh: It's a fine idea but there is some ambiguous point there. `eval_upto` is expected to get `train`, `export` or `optimize` but `resume_from` can also get `all` including them. So, even if there is enum class, using it for both of them looks a little bit weird. Also comparing between values including `all` is also other ambiguous point. For these reasons, I think it's better to leave it as it is for now.

goodsong81 (approved):
LGTM

sungchul1 (approved):

PR #3759, opened 2024-07-23
Enable to use input_size at transforms in recipe

eunwoosh:
### Summary
This PR enables to use input_size at transforms in recipe.
It's for unified input size. So, if value of all init_args in recipe uses input_size, all necessary values are automatically updated by changing only input_size.
For example, let's assume that we have recipe as below.
```
  data:
    train_subset:
      transforms:
        - class_path: otx.core.data.transform_libs.torchvision.RandomResize
          init_args:
            scale
            - 1000
            - 1000
        - class_path: otx.core.data.transform_libs.torchvision.RandomCrop
          init_args:
            crop_size
            - 500
            - 500
```
It can be changed as below using input_size.
```
  data:
    input_size:
    - 1000
    - 1000
    train_subset:
      transforms:
        - class_path: otx.core.data.transform_libs.torchvision.RandomResize
          init_args:
            scale: $(inputs_size)
        - class_path: otx.core.data.transform_libs.torchvision.RandomCrop
          init_args:
            crop_size: $(input_size) * 0.5
```
As you can see above, `$(input_size)` should be used to notify to use input_size.
Only simple multiplication or division evaluation is supported. So, addition or subtraction can't be used.
Type of input_size can be `int` or `tuple[int, int]`. If `int` value is given, it's considered as `tuple[int, int]` with same value.
whether passing `int` or `tuple[int, int]` to transform class is decided based on type hint. If there is no type hint, `tuple[int, int]` is passed.

input_size can be overrided by subset config. For example, if we want to use (1000, 1000) input size at train and (800, 800) at validation, recipe below can be used.
```
  data:
    input_size:
    - 1000
    - 1000
    train_subset:
      transforms:
        - class_path: otx.core.data.transform_libs.torchvision.RandomResize
          init_args:
            scale: $(inputs_size)
    val_subset:
      input_size:
      - 800
      - 800
      transforms:
        - class_path: otx.core.data.transform_libs.torchvision.RandomResize
          init_args:
            scale: $(inputs_size)
```
Of course, setting input_size only in subset confg rather than data config is available.
But input_size can't use `$(input_size)`. It's available only in `transforms`.

### Checklist

### License

```python
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ All notable changes to this project will be documented in this file.
 
 - Enable to override data configurations
   (<https://github.com/openvinotoolkit/training_extensions/pull/3748>)
-- Enable to use input_size at transforms in reipce
+- Enable to use input_size at transforms in recipe
   (<https://github.com/openvinotoolkit/training_extensions/pull/3759>)
 
 ### Bug fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ All notable changes to this project will be documented in this file.
 
 - Enable to override data configurations
   (<https://github.com/openvinotoolkit/training_extensions/pull/3748>)
+- Enable to use input_size at transforms in reipce
+  (<https://github.com/openvinotoolkit/training_extensions/pull/3759>)
 
 ### Bug fixes
 
--- a/src/otx/core/data/transform_libs/torchvision.py
+++ b/src/otx/core/data/transform_libs/torchvision.py
@@ -3125,8 +3125,7 @@ class TorchVisionTransformLib:
         input_size = getattr(config, "input_size", None)
         transforms = []
         for cfg_transform in config.transforms:
-            if input_size is not None:
-                cls._configure_input_size(cfg_transform, input_size)
+            cls._configure_input_size(cfg_transform, input_size)
             transform = cls._dispatch_transform(cfg_transform)
             transforms.append(transform)
 
@@ -3144,9 +3143,10 @@ class TorchVisionTransformLib:
         The function decides to pass tuple type or int type based on the type hint of the argument.
         float point values are rounded to int.
         """
-        if input_size is None:
-            return
-        _input_size: tuple[int, int] = (input_size, input_size) if isinstance(input_size, int) else tuple(input_size)  # type: ignore[assignment]
+        if input_size is not None:
+            _input_size: tuple[int, int] = (
+                (input_size, input_size) if isinstance(input_size, int) else tuple(input_size)
+            )  # type: ignore[assignment]
 
         def check_type(value: Any, expected_type: Any) -> bool:  # noqa: ANN401
             try:
@@ -3159,6 +3159,13 @@ class TorchVisionTransformLib:
         for key, val in cfg_transform["init_args"].items():
             if not (isinstance(val, str) and "$(input_size)" in val):
                 continue
+            if input_size is None:
+                msg = (
+                    f"{cfg_transform['class_path'].split('.')[-1]} initial argument has `$(input_size)`, "
+                    "but input_size is set to None."
+                )
+                raise RuntimeError(msg)
+
             if model_cls is None:
                 model_cls = import_object_from_module(cfg_transform["class_path"])
 
--- a/src/otx/core/data/transform_libs/torchvision.py
+++ b/src/otx/core/data/transform_libs/torchvision.py
@@ -70,7 +70,7 @@ from otx.core.data.transform_libs.utils import (
     translate_masks,
     translate_polygons,
 )
-from otx.core.utils.utils import get_obj_from_str
+from otx.core.utils.utils import import_object_from_module
 
 if TYPE_CHECKING:
     from otx.core.config.data import SubsetConfig
@@ -3125,14 +3125,14 @@ class TorchVisionTransformLib:
         transforms = []
         for cfg_transform in config.transforms:
             if (input_size := getattr(config, "input_size", None)) is not None:
-                cls._eval_input_size(cfg_transform, input_size)
+                cls._configure_input_size(cfg_transform, input_size)
             transform = cls._dispatch_transform(cfg_transform)
             transforms.append(transform)
 
         return Compose(transforms)
 
     @classmethod
-    def _eval_input_size(cls, cfg_transform: dict[str, Any], input_size: int | tuple[int, int] | None) -> None:
+    def _configure_input_size(cls, cfg_transform: dict[str, Any], input_size: int | tuple[int, int] | None) -> None:
         """Evaluate the input_size and replace the placeholder in the init_args.
 
         Input size should be specified as $(input_size). (e.g. $(input_size) * 0.5)
@@ -3159,24 +3159,24 @@ class TorchVisionTransformLib:
             if not (isinstance(val, str) and "$(input_size)" in val):
                 continue
             if model_cls is None:
-                model_cls = get_obj_from_str(cfg_transform["class_path"])
+                model_cls = import_object_from_module(cfg_transform["class_path"])
 
             available_types = typing.get_type_hints(model_cls.__init__).get(key)
             if available_types is None or check_type(_input_size, available_types):  # pass tuple[int, int]
-                cfg_transform["init_args"][key] = cls._safe_eval(
+                cfg_transform["init_args"][key] = cls._eval_input_size_str(
                     val.replace("$(input_size)", f"({','.join(str(val) for val in _input_size)})"),
                 )
             elif check_type(_input_size[0], available_types):  # pass int
-                cfg_transform["init_args"][key] = cls._safe_eval(val.replace("$(input_size)", str(_input_size[0])))
+                cfg_transform["init_args"][key] = cls._eval_input_size_str(val.replace("$(input_size)", str(_input_size[0])))
             else:
                 msg = f"{key} argument should be able to get int or tuple[int, int], but it can get {available_types}"
                 raise RuntimeError(msg)
 
     @classmethod
-    def _safe_eval(cls, str_to_eval: str) -> tuple[int, ...] | int:
-        """Safe eval function for _eval_input_size.
+    def _eval_input_size_str(cls, str_to_eval: str) -> tuple[int, ...] | int:
+        """Safe eval function for _configure_input_size.
 
-        The function is implemented for `_eval_input_size`, so implementation is aligned to it as below
+        The function is implemented for `_configure_input_size`, so implementation is aligned to it as below
         - Only multiplication or division evaluation are supported.
         - Only constant and tuple can be operand.
         - tuple is changed to numpy array before evaluation.
--- a/src/otx/core/data/transform_libs/torchvision.py
+++ b/src/otx/core/data/transform_libs/torchvision.py
@@ -3125,7 +3125,8 @@ class TorchVisionTransformLib:
         input_size = getattr(config, "input_size", None)
         transforms = []
         for cfg_transform in config.transforms:
-            cls._configure_input_size(cfg_transform, input_size)
+            if isinstance(cfg_transform, (dict, DictConfig)):
+                cls._configure_input_size(cfg_transform, input_size)
             transform = cls._dispatch_transform(cfg_transform)
             transforms.append(transform)
 
@@ -3156,7 +3157,7 @@ class TorchVisionTransformLib:
             return True
 
         model_cls = None
-        for key, val in cfg_transform["init_args"].items():
+        for key, val in cfg_transform.get("init_args", {}).items():
             if not (isinstance(val, str) and "$(input_size)" in val):
                 continue
             if input_size is None:
--- a/src/otx/core/data/transform_libs/torchvision.py
+++ b/src/otx/core/data/transform_libs/torchvision.py
@@ -3145,8 +3145,8 @@ class TorchVisionTransformLib:
         """
         if input_size is not None:
             _input_size: tuple[int, int] = (
-                (input_size, input_size) if isinstance(input_size, int) else tuple(input_size)
-            )  # type: ignore[assignment]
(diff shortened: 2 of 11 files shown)


src/otx/core/utils/utils.py
```diff
@@ -84,3 +85,10 @@ def get_idx_list_per_classes(dm_dataset: DmDataset, use_string_label: bool = Fal
     for k in stats:
         stats[k] = list(dict.fromkeys(stats[k]))
     return stats
+
+
+def get_obj_from_str(obj_path: str) -> Any:  # noqa: ANN401
+    """Get object from import format string."""
+    module_name, obj_name = obj_path.rsplit(".", 1)
+    module = importlib.import_module(module_name)
+    return getattr(module, obj_name)
```
goodsong81: Function name seems not telling what it does; `import_object_from_module()`?

src/otx/core/data/transform_libs/torchvision.py
```diff
@@ -2364,7 +2369,7 @@ class RandomCrop(tvt_v2.Transform, NumpytoTVTensorMixin):
 
     def __init__(
         self,
-        crop_size: tuple,  # (H, W)
+        crop_size: tuple[int, int],  # (H, W)
```
goodsong81, line 2372: Is this always true in OTX? No (W, H)? OK if it is so but we'd better modify test cases for non-square input size to check for the future changes.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ All notable changes to this project will be documented in this file.
 
 - Enable to override data configurations
   (<https://github.com/openvinotoolkit/training_extensions/pull/3748>)
+- Enable to use input_size at transforms in reipce
```
sungchul1:
```suggestion
- Enable to use input_size at transforms in recipe
```

tests/unit/core/data/test_module.py
```diff
@@ -115,13 +119,17 @@ def test_init(
             train_subset=fxt_config.train_subset,
             val_subset=fxt_config.val_subset,
             test_subset=fxt_config.test_subset,
+            input_size=input_size,
```
harimkang: Maybe we need an additional test to see if the overriding works depending on the position of `input_size`?
Also, could there be a test case to see if it works if `input_size` is None and each transform has a corresponding value?

It would be nice to have a test case to verify that it works in the two cases below.
1. `input_size` is None & `each_subset.input_size` is different.
2. input_size` is None & `each_subset.input_size` is None & `each_transform.scale` is not None.

src/otx/core/data/transform_libs/torchvision.py
```diff
@@ -3119,11 +3124,100 @@ def generate(cls, config: SubsetConfig) -> Compose:
 
         transforms = []
         for cfg_transform in config.transforms:
+            if (input_size := getattr(config, "input_size", None)) is not None:
```
sungchul1: Do we need to get `input_size` every iteration?

src/otx/core/data/transform_libs/torchvision.py
```diff
@@ -3119,11 +3124,100 @@ def generate(cls, config: SubsetConfig) -> Compose:
...
+    @classmethod
+    def _safe_eval(cls, str_to_eval: str) -> tuple[int, ...] | int:
+        """Safe eval function for _eval_input_size.
+
+        The function is implemented for `_eval_input_size`, so implementation is aligned to it as below
+        - Only multiplication or division evaluation are supported.
+        - Only constant and tuple can be operand.
+        - tuple is changed to numpy array before evaluation.
+        - result value is rounded to int.
+        """
```
goodsong81: To me, it looks like doing actually `_eval_input_size_str()`.

src/otx/core/data/transform_libs/torchvision.py
```diff
@@ -3119,11 +3124,100 @@ def generate(cls, config: SubsetConfig) -> Compose:
...
+        Input size should be specified as $(input_size). (e.g. $(input_size) * 0.5)
+        Only simple multiplication or division evaluation is supported. For example,
+        $(input_size) * -0.5    => supported
+        $(input_size) * 2.1 / 3 => supported
+        $(input_size) + 1       => not supported
+        The function decides to pass tuple type or int type based on the type hint of the argument.
+        float point values are rounded to int.
+        """
+        if input_size is None:
+            return
```
sungchul1: It seems this condition is not passed because checking whether `input_size` is None is already in L3127.
I think we can use it and remove L3127.

goodsong81: And this function is doing `_(re)configure_input_size()` by filling the template if needed.

goodsong81 (dismissed):
Quite nice feature for recipe users or developers.
Generally looks good, but it might be controversial if it's "too much".
Left minor comments from a code reader's viewpoint. Feel free to ignore.

src/otx/core/data/transform_libs/torchvision.py
eunwoosh: yes, I think it's better to change function name. Thanks :)

src/otx/core/data/transform_libs/torchvision.py
```diff
@@ -3119,11 +3124,100 @@ def generate(cls, config: SubsetConfig) -> Compose:
...
+        for key, val in cfg_transform["init_args"].items():
+            if not (isinstance(val, str) and "$(input_size)" in val):
+                continue
+            if model_cls is None:
+                model_cls = get_obj_from_str(cfg_transform["class_path"])
+
+            available_types = typing.get_type_hints(model_cls.__init__).get(key)
+            if available_types is None or check_type(_input_size, available_types):  # pass tuple[int, int]
+                cfg_transform["init_args"][key] = cls._safe_eval(
+                    val.replace("$(input_size)", f"({','.join(str(val) for val in _input_size)})"),
```
sungchul1: If the purpose is to get `(1, 1) * 2` from `val="$(input_size) * 2"` and `input_size=(1, 1)`, what about using `{input_size}` instead for below?
```suggestion
                    val.format(input_size=_input_size),
```

tests/unit/core/data/test_module.py
eunwoosh: Nice comment. I'll add it for these cases

src/otx/core/utils/utils.py
eunwoosh: I see. I'll make it more readable

src/otx/core/data/transform_libs/torchvision.py
```diff
@@ -3119,11 +3124,100 @@ def generate(cls, config: SubsetConfig) -> Compose:
...
+                )
+            elif check_type(_input_size[0], available_types):  # pass int
```
sungchul1, line 3178:
```suggestion
            elif all(check_type(_size, available_types) for _size in _input_size):  # pass int
```

src/otx/core/data/transform_libs/torchvision.py
```diff
@@ -3119,11 +3124,100 @@ def generate(cls, config: SubsetConfig) -> Compose:
...
+        Only simple multiplication or division evaluation is supported. For example,
+        $(input_size) * -0.5    => supported
+        $(input_size) * 2.1 / 3 => supported
+        $(input_size) + 1       => not supported
+        The function decides to pass tuple type or int type based on the type hint of the argument.
+        float point values are rounded to int.
+        """
+        if input_size is None:
+            return
+        _input_size: tuple[int, int] = (input_size, input_size) if isinstance(input_size, int) else tuple(input_size)  # type: ignore[assignment]
```
sungchul1: Do we need to use this conversion from int to tuple?
I'm not sure if it is necessary because L3170 uses a single integer.

sungchul1 (commented):
Thanks for your hard work. :)
I left some comments, please take a look.
And I think this PR has to include updated recipes to check unit and integration tests that are affected by this change.
What do you think?

src/otx/core/data/transform_libs/torchvision.py
eunwoosh: I just updated type hint for the input size feature, so I don't know well this. But when I checked the code, crop_size should be (H, W) because this class refers `ImageInfo` class and it has a tuple w/ (H, W) order. I'll update the unit test to use non-square value.

src/otx/core/data/transform_libs/torchvision.py
eunwoosh: Thanks for pointing. I'll update it.

src/otx/core/data/transform_libs/torchvision.py
eunwoosh: I think leaving the safeguard is better choice. How about leaving it as it is?

src/otx/core/data/transform_libs/torchvision.py
eunwoosh: oh nice point. thanks

src/otx/core/data/transform_libs/torchvision.py
eunwoosh: should it check both `input_size[0]` and `input_size[1]`? I think it's ok to check just `input_size[0]` because `input_size[0]` is given as the argument. How do you think about it?

src/otx/core/data/transform_libs/torchvision.py
eunwoosh: If we don't change it to tuple, then we should check what type `input_size` is currently because `input_size` can be either int or tuple. So, we need to add a few code to check current type of `input_size` after `if available_types is None or check_type(_input_size, available_types)`. For this reason, I think fixing input_size type to tuple makes code more readable.

src/otx/core/data/transform_libs/torchvision.py
eunwoosh: done https://github.com/openvinotoolkit/training_extensions/pull/3759/commits/589c4c57264a27a5da2cfd9bb9aa3b2788c53bef

src/otx/core/data/transform_libs/torchvision.py
eunwoosh: done https://github.com/openvinotoolkit/training_extensions/pull/3759/commits/589c4c57264a27a5da2cfd9bb9aa3b2788c53bef

src/otx/core/utils/utils.py
eunwoosh: done https://github.com/openvinotoolkit/training_extensions/pull/3759/commits/589c4c57264a27a5da2cfd9bb9aa3b2788c53bef

eunwoosh: done https://github.com/openvinotoolkit/training_extensions/pull/3759/commits/589c4c57264a27a5da2cfd9bb9aa3b2788c53bef

src/otx/core/data/transform_libs/torchvision.py
eunwoosh: done https://github.com/openvinotoolkit/training_extensions/pull/3759/commits/2385183a05f5e4bc871e2c461287abcbf344008c

CHANGELOG.md
eunwoosh: done https://github.com/openvinotoolkit/training_extensions/pull/3759/commits/080932be38ac791a29b6f3de922fa094bf58a7f3

src/otx/core/data/transform_libs/torchvision.py
eunwoosh: done https://github.com/openvinotoolkit/training_extensions/pull/3759/commits/080932be38ac791a29b6f3de922fa094bf58a7f3

src/otx/core/data/transform_libs/torchvision.py
eunwoosh: done https://github.com/openvinotoolkit/training_extensions/pull/3759/commits/080932be38ac791a29b6f3de922fa094bf58a7f3

src/otx/core/data/transform_libs/torchvision.py
sungchul1: If so, let's keep this.

src/otx/core/data/transform_libs/torchvision.py
sungchul1: I don't think about it.
Please check the below example.
```python
model_cls = get_obj_from_str("otx.core.data.transform_libs.torchvision.Resize") # <class 'otx.core.data.transform_libs.torchvision.Resize'>
key = "scale"
val = "$(input_size)"
available_types = typing.get_type_hints(model_cls.__init__).get(key) # int | tuple[int, int] | None

# for tuple
_input_size = (1, 1)
check_type(_input_size, available_types) # True with (1, 1)

# for int
_input_size = 1
check_type(_input_size, available_types) # True with 1

# for str
_input_size = "1"
check_type(_input_size, available_types) # False with TypeCheckError set at L3153

# both tuple and int can use the same argument for `cls._safe_eval`
cls._safe_eval(val.replace("$(input_size)", str(_input_size))
```
I think we don't need an additional condition you said because `check_type` can support both types.
If I'm wrong, please correct me :)

sungchul1: I thought it again, you seem to consider the cases about only single tuple argument or int argument, is it correct?
If so, I think your choice is more safe, please ignore it and let's keep this.

tests/unit/core/data/test_module.py
eunwoosh: done https://github.com/openvinotoolkit/training_extensions/pull/3759/commits/8b3dc51d8e3b64d076e47c6bae6ea4d366b8a022
I added more unit test and change code to make proper error message if input_size is set to None and transform has `$(input_size)`.

src/otx/core/data/transform_libs/torchvision.py
eunwoosh: yes right. I considered that case. Anyway thanks for thinking how to refine my code :)

harimkang (approved):

kprokofi (approved):

PR #3760, opened 2024-07-23
[RFC] Engine Refactor Proposal | Alternative 2

ashwinvaidya17:
# Motivation

Refer to https://github.com/openvinotoolkit/training_extensions/pull/3752 for the motivation

This PR proposes an alternative design. It also includes one solution to register heterogeneous models to the CLI (there might be better approaches). 

```python
def test_engine_anomaly():
    model = Padim()
    engine = Engine(model)
    datamodule = MVTec(root="./datasets/MVTec")
    engine.train(datamodule)
    assert isinstance(engine, AnomalyEngine)


def test_engine_classification():
    datamodule = OTXDataModule(
        task=OTXTaskType.MULTI_CLASS_CLS,
        ...
    )
    model = MobileNetV3ForMulticlassCls(
        label_info=datamodule.label_info,
    )
    engine = Engine(model)
    engine.train(datamodule, max_epochs=1, deterministic=False, val_check_interval=0.5)
    assert isinstance(engine, LightningEngine)


def test_engine_ultralytics():
    model = YOLO("yolov8n.pt")
    engine = Engine(model)
    dataset = ClassificationDataset(root="./datasets/hazelnut_toy", args=DEFAULT_CFG)
    dataloader = DataLoader(dataset, batch_size=32, shuffle=True)
    engine.train(datamodule=dataloader)
    assert isinstance(engine, UltralyticsEngine)
```


src/otx/engine/engine_poc.py
```diff
@@ -0,0 +1,42 @@
...
+from .base import BaseEngine
+
+logger = logging.getLogger(__name__)
+
+
+class AutoConfigurator:
+    """Mock autoconfigurator for the engine."""
+
+    def __init__(self, model: nn.Module | None = None, data_root: Path | None = None, task: str | None = None):
+        self._engine = self._configure_engine(model)  # ideally we want to pass the data_root and task as well
```
harimkang, line 17: Currently, the role of auto-configuration is to check for task, data, and model inputs, regardless of the engine, and provide default settings for anything the user hasn't entered. Is there any reason to configure the engine internally? If it's just for the backend, it would be nice to have a different way to configure the default for each backend rather than configuring the engine directly. What do you think?

src/otx/engine/engine_poc.py
```diff
@@ -0,0 +1,42 @@
...
+    """Automatically selects the engine based on the model passed to the engine."""
+
+    def __new__(
+        cls,
+        model: nn.Module,
+        data_root: Path | None = None,
+        **kwargs,
+    ) -> BaseEngine:
+        """This takes in all the parameters that are currently passed to the OTX Engine's `__init__` method."""
+        autoconfigurator = AutoConfigurator(model, data_root=data_root, **kwargs)
```
harimkang, line 41: Engine -> AutoConfigurator -> Engine : I think their relationship with each other is strange.

src/otx/engine/base.py
```diff
@@ -0,0 +1,50 @@
+from abc import ABC, abstractmethod
+from typing import Any
+
+from torch import nn
+
+METRICS = dict[str, float]
+ANNOTATIONS = Any
+
+
+class BaseEngine(ABC):
```
harimkang, line 10: As we talked about, the arguments in `BaseEngine` will be the same as those in the current otx Engine, only the Type will be made more general as needed, right?

src/otx/engine/base.py
```diff
@@ -0,0 +1,50 @@
...
+    BASE_MODEL: nn.Module  # Use this to register models to the CLI
+
+    @classmethod
+    @abstractmethod
+    def is_valid_model(cls, model: nn.Module) -> bool:
+        pass
+
+    @abstractmethod
+    def train(self, model: nn.Module, **kwargs) -> METRICS:
```
harimkang, line 19: There's still a model here, which I think might be confusing for people looking at this PR.

PR #3762, opened 2024-07-24
Add otx micro benchmark

sovrasov:
### Summary

`otx benchmark` command aims to provide micro benchmarking capabilities for OTX users.
Given a model (OV or torch) `otx benchmark` could quickly estimate realistic deployment performance and explore the effect of varying batch size. The evaluation works on synthetic data, which is provided by `OTXModel.get_dummy_input(batch_size)` method. Also, in case of torch models theoretical complexity and amount of parameters are available.
image

Issues:
- `get_dummy_input()` should be implemented for each model type. Current limitation is hard-coded per model input resolutions. The same problem is valid for export.
- torch 2.1 messes up with Tensor types in dispatch mode, a workaround was added. There is a chance torch 2.3 resolves the issue.
- Models coverage.
- Unclear console output model in Engine.

### Checklist

### License

```python
```


src/otx/algo/classification/heads/vision_transformer_head.py
```diff
@@ -141,4 +141,4 @@ def predict(
 
     def _get_predictions(self, cls_score: torch.Tensor) -> torch.Tensor:
         """Get the score from the classification score."""
-        return functional.softmax(cls_score, dim=1)
+        return functional.softmax(cls_score, dim=-1)
```
eunwoosh, line 144: It's just a question. Why did you change it?

src/otx/core/model/base.py
```diff
@@ -781,6 +782,9 @@ def tile_config(self, tile_config: TileConfig) -> None:
 
         self._tile_config = tile_config
 
+    def get_dummy_input(self, batch_size: int = 1) -> OTXBatchDataEntity[Any]:
+        raise RuntimeError("get_dummy_input() is not implemented")
```
eunwoosh:
```suggestion
        raise NotImplementedError
```

src/otx/core/model/classification.py
```diff
@@ -86,6 +87,13 @@ def _convert_pred_entity_to_compute_metric(
             "target": target,
         }
 
+    def get_dummy_input(self, batch_size: int = 1) -> MulticlassClsBatchDataEntity:
+        labels = [torch.LongTensor([0])] * batch_size
+        images = [torch.rand(*self.image_size[1:]) for _ in range(batch_size)]
+        labels = [torch.LongTensor([0])] * batch_size
```
eunwoosh, line 93:
```suggestion
```

src/otx/engine/engine.py
```diff
@@ -778,6 +779,80 @@ def explain(
...
+        elif isinstance(self.model, OVModel):
+            msg = "To run benchmark on OV model, checkpoint must be specified."
+            raise RuntimeError(msg)
+
+        self.model.eval()
+
+        def dummy_infer(model: OTXModel, batch_size: int = 1, extra_stats: bool = False) -> dict[str, Any]:
+            input_batch = model.get_dummy_input(batch_size)
+            stats = {}
+            start = time.time()
```
```suggestion
            start = time.perf_counter()
```

src/otx/engine/engine.py
```diff
@@ -778,6 +779,80 @@ def explain(
...
+                model_fwd = lambda: model.forward(input_batch)
+                depth = 3 if extended_stats else 0
+                fwd_flops = measure_flops(model.model, model_fwd, print_stats_depth=depth)
+                stats["flops"] = convert_num_with_suffix(fwd_flops, get_suffix_str(fwd_flops*10**4))
+
+                params_num = sum(p.numel() for p in self.model.parameters() if p.requires_grad)
+                stats["params"] = convert_num_with_suffix(params_num, get_suffix_str(params_num*100))
+            else:
+                model.forward(input_batch)
+            end = time.time()
```
```suggestion
            end = time.perf_counter()
```

src/otx/engine/engine.py
```diff
@@ -778,6 +779,80 @@ def explain(
         model.explain_mode = False
         return predict_result
 
+    def benchmark(
```
eunwoosh, line 786: I think it becomes more useful if there is a way to save a result as file. How do you think about it?

src/otx/core/model/classification.py
```diff
@@ -711,6 +730,13 @@ def _convert_pred_entity_to_compute_metric(
             "target": target,
         }
 
+    def get_dummy_input(self, batch_size: int = 1) -> MulticlassClsBatchDataEntity:
+        """Returns a dummy input for classification OV model"""
+        # Resize is embedded to the OV model, which means we don't need to know the actual size
+        images = [torch.rand(3, 224, 224) for _ in range(batch_size)]
```
eunwoosh: Don't we need to consider a dataset having multiple input size?

src/otx/engine/engine.py
```diff
@@ -778,6 +779,80 @@ def explain(
...
+        def dummy_infer(model: OTXModel, batch_size: int = 1, extra_stats: bool = False) -> dict[str, Any]:
+            input_batch = model.get_dummy_input(batch_size)
+            stats = {}
+            start = time.time()
+            if extra_stats:
+                from torch.utils.flop_counter import get_suffix_str, convert_num_with_suffix
+
+                model_fwd = lambda: model.forward(input_batch)
+                depth = 3 if extended_stats else 0
+                fwd_flops = measure_flops(model.model, model_fwd, print_stats_depth=depth)
```
eunwoosh: Is it ok to measure flops while mesureing speed? Doesn't it spend more time than just executing forward?

src/otx/engine/engine.py
```diff
@@ -778,6 +779,80 @@ def explain(
...
+        for _ in range(warmup_iters):
+            dummy_infer(self.model, batch_size)
+
+        total_time = 0
+        for _ in range(n_iters):
+            inference_stats = dummy_infer(self.model, batch_size)
+            total_time += inference_stats["elapsed"]
+        total_time /= (n_iters * batch_size)
+
+        final_stats = {"latency": f"{total_time:.3f} s", "troughput": f"{(1 / total_time):.3f} FPS"}
```
eunwoosh: As far as I know, latency maens how long do I need to wait for getting result from the model. But current latency is measured as `total_time / (n_iters * batch_size)`. I think actual latency can be different because model returns a result after processing whole batch. How do you think about it?

eunwoosh (commented):

src/otx/algo/classification/heads/vision_transformer_head.py
sovrasov: Flops counter at some point could cut the batch dimension, so we have to use relative indexing. Let's see if we face similar issues with other models

src/otx/core/model/classification.py
sovrasov: I wouldn't say this is true in production, we're a little bit biased by web-scraped data. Usually, the image source is a camera with a stable resolution. 
Changing the size from the default one makes sense: then resize would be really executed. The resize complexity is proportional to output HW, therefore the input image shape is not that important (it shouldn't be too small though to prevent extensive cache usage). I'll change the size from default to let resize be executed

src/otx/engine/engine.py
sovrasov: I the benchmark scenario I estimate flops with a standalone call, but overall flops counter is a bunch of lightweight hooks on top of torch operations. I'll think about refactoring.

src/otx/core/model/base.py
sovrasov: Fixed

src/otx/engine/engine.py
sovrasov: Updated

src/otx/engine/engine.py
sovrasov: Updated

src/otx/core/model/classification.py
sovrasov: Updated

kprokofi (commented):
Thank you for this valuable feature.

I have few comments/proposals. I checked this functionality with segmentation models.

1. Currently, it is required to provide data_root for all OTX CLI commands. Since this functionality doesn't require any datasets, can we add some default datamodule (if there is no other provided) to run otx benchmark without --data_root option?

2. It will be a good option if we integrate otx.export (OV) to otx.benchmark. For example, I want to test torch models and OV models. For now, I should train them first (or remove checkpoint loading logic in OTX and use default weights), then export and after that use otx benchmark. It ends up to 2-3 commands and searching for checkpoint paths. Could we add like "convert_to_openvino" or "test_openvino" option, call export and call benchmark again with OV model?

sovrasov:
> Thank you for this valuable feature.
> 
> I have few comments/proposals. I checked this functionality with segmentation models.
> 
> 1. Currently, it is required to provide data_root for all OTX CLI commands. Since this functionality doesn't require any datasets, can we add some default datamodule (if there is no other provided) to run otx benchmark without --data_root option?
> 2. It will be a good option if we integrate otx.export (OV) to otx.benchmark. For example, I want to test torch models and OV models. For now, I should train them first (or remove checkpoint loading logic in OTX and use default weights), then export and after that use otx benchmark. It ends up to 2-3 commands and searching for checkpoint paths. Could we add like "convert_to_openvino" or "test_openvino" option, call export and call benchmark again with OV model?

1. That's the limitation of current OTX implementation. Without having a dataset one should perform full model auto-configuration, including setting num classes for instance. In theory, export command also doesn't require any data, but we just don't have a mechanism to skip the creation of an actual dataloader. I'll think about an API-based benchmark, so we could streamline the workflow at least there.
2. The best we can do now is to benchmark a torch model without `--checkpoint` specified (~no training for torch model is required). otx * tools are supposed to be orthogonal, so none of it implicitly uses another. This introduces a clear understanding of dependencies required for each otx * tool, but it could take more time to achieve the result. If we relax otx export requirements (omit --checkpoint), that could mitigate the need of pretraining weights, but still 2 calls are required.

sovrasov:
Don't merge docs are WIP

src/otx/core/model/visual_prompting.py
```diff
@@ -916,6 +974,9 @@ def _customize_inputs(  # type: ignore[override]
         """Customize OTX input batch data entity."""
         images: list[np.ndarray] = []
         processed_prompts: list[dict[str, Any]] = []
+        print(entity.labels)
+        print(entity.prompts)
+        print(entity.polygons)
```
harimkang: Is there a reason to output this customize_inputs every time?

src/otx/engine/engine.py
```diff
@@ -778,6 +780,97 @@ def explain(
...
+            checkpoint (PathLike | None, optional): _description_. Defaults to None.
+            batch_size (int, optional): _description_. Defaults to 1.
+            n_iters (int, optional): _description_. Defaults to 10.
+            extended_stats (bool, optional): _description_. Defaults to False.
+
+        Raises:
+            RuntimeError: _description_
+
+        Returns:
+            dict[str, str]: _description_
```
harimkang: Looks like it needs an update.

src/otx/algo/segmentation/litehrnet.py
```diff
@@ -558,6 +558,9 @@ def _optimization_config(self) -> dict[str, Any]:
     @property
     def _exporter(self) -> OTXModelExporter:
         """Creates OTXModelExporter object that can export the model."""
+        if self.image_size is None:
+            raise ValueError(self.image_size)
```
sungchul1: Could you update an error message including other tasks?

src/otx/core/model/instance_segmentation.py
```diff
@@ -361,6 +363,24 @@ def _convert_pred_entity_to_compute_metric(
             )
         return {"preds": pred_info, "target": target_info}
 
+    def get_dummy_input(self, batch_size: int = 1) -> InstanceSegBatchDataEntity:
+        """Returns a dummy input for instance segmentation model."""
+        if self.image_size is None:
+            raise ValueError(self.image_size)
+
+        # images = torch.rand(batch_size, *self.image_size[1:])
```
sungchul1:
```suggestion

```
if it is unnecessary

src/otx/engine/engine.py
```diff
@@ -778,6 +780,97 @@ def explain(
...
+        for _ in range(warmup_iters):
+            dummy_infer(self.model, batch_size)
+
+        total_time = 0.0
+        for _ in range(n_iters):
+            total_time += dummy_infer(self.model, batch_size)
+        latency = total_time / n_iters
+        fps = batch_size / latency
+
+        final_stats = {"latency": f"{latency:.3f} s", "troughput": f"{(fps):.3f} FPS"}
```
sungchul1:
```suggestion
        final_stats = {"latency": f"{latency:.3f} s", "throughput": f"{(fps):.3f} FPS"}
```

src/otx/core/model/segmentation.py
```diff
@@ -148,7 +167,7 @@ def __init__(
         self.decode_head_configuration = decode_head_configuration if decode_head_configuration is not None else {}
         export_image_configuration = export_image_configuration if export_image_configuration is not None else {}
         self.criterion_configuration = criterion_configuration
-        self.image_size = tuple(export_image_configuration.get("image_size", (1, 3, 512, 512)))
+        self.image_size = tuple([int(i) for i in export_image_configuration.get("image_size", (1, 3, 512, 512))])
```
sungchul1: Does `image_size` in `export_image_configuration` have non-integer values?

sungchul1 (commented):
Thanks for your hard work! It seems the feature is what we really need.
I left some minor comments.

src/otx/core/model/anomaly.py
```diff
@@ -221,13 +223,24 @@ def configure_optimizers(self) -> tuple[list[torch.optim.Optimizer], list[torch.
...
         return super().configure_optimizers()  # type: ignore[misc]
 
+    def validation_step(
+        self,
+        inputs: AnomalyModelInputs,
+        batch_idx: int = 0,
+    ) -> STEP_OUTPUT:
+        """Call validation step of the anomalib model."""
+        raise NotImplementedError
+
```
eunwoosh, line 233: It's just a question. Is it necessary?

src/otx/core/model/anomaly.py
```diff
@@ -221,13 +223,24 @@ def configure_optimizers(self) -> tuple[list[torch.optim.Optimizer], list[torch.
...
         inputs: AnomalyModelInputs,
     ) -> AnomalyModelOutputs:
         """Wrap forward method of the Anomalib model."""
-        _inputs: dict = self._customize_inputs(inputs)
-        outputs = self.model.model.forward(_inputs)
+        outputs = self.validation_step(inputs)
+        _PostProcessorCallback._post_process(outputs)  # noqa: SLF001
+        _PostProcessorCallback._compute_scores_and_labels(self, outputs)  # noqa: SLF001
+        _MinMaxNormalizationCallback._normalize_batch(outputs, self)  # noqa: SLF001
+
```
eunwoosh: It's also question. Could you explain this?

sovrasov: Anomaly model doesn't use this forward method, instead only train and validation steps are overloaded. That's an irregularity in OTX design which is to be fixed. cc @ashwinvaidya17

src/otx/core/model/anomaly.py
sovrasov: Final anomaly model is inherited from several classes including this one, and to workaround missing forward implementation I need to call `validation_step` from the other anomaly-related class here. That's why I need this declaration. This class is a subject of the future anomalib integration refactoring planned by Ashwin and Dick.

src/otx/core/model/anomaly.py
ashwinvaidya17: For more context. We have a post processing callback and normalization callback in anomalib that ensures that all the models share the same post processing + normalization steps without requiring the model implementers to worry about it.

- Post processing deals with converting anomaly maps to segmentation maps, segmentation maps to bounding boxes, and/or anomaly maps to anomaly score. This depends on the task type of the underlying model.  
- Normalization callback converts raw anomaly maps to maps in range [0, 1]. The normalization algorithm is customizable and is independent of the underlying models.

Dick is working on a new design that might potentially solve this issue. It's in PoC so I am not aware of the complete details. @sovrasov can you add a ToDo with mentioning me so I don't forget to revisit this? Currently, OTX only uses min-max normalization so this will work. I can refactor this once we have the new design in Anomalib.

src/otx/core/model/visual_prompting.py
sovrasov: The PR is still WIP, to be deleted

src/otx/engine/engine.py
sovrasov: Updated

src/otx/core/model/instance_segmentation.py
sovrasov: Deleted

src/otx/engine/engine.py
sovrasov: Fixed

ashwinvaidya17 (commented):
This is a nice addition. Looks good from Anomaly perspective.

src/otx/engine/engine.py
```diff
@@ -778,6 +780,100 @@ def explain(
...
+            try:
+                from torch.utils.flop_counter import convert_num_with_suffix, get_suffix_str
+
+                input_batch = self.model.get_dummy_input(1)
+                model_fwd = lambda: self.model.forward(input_batch)
+                depth = 3 if extended_stats else 0
+                fwd_flops = measure_flops(self.model.model, model_fwd, print_stats_depth=depth)
+                flops_str = convert_num_with_suffix(fwd_flops, get_suffix_str(fwd_flops * 10**3))
+                final_stats["complexity"] = flops_str + " MACs"
+            except Exception as e:
```
kprokofi, line 879: I think Flops is more common complexity indicator. Can we use it instead of MACs?

sovrasov: In that case, we'd have to multiply the output of the flops counter by 2, because MACs are actually counted according to the source code, but this doesn't make sense, since accumulation and multiplication are effectively fused in modern hardware.
See https://www.kdnuggets.com/2023/06/calculate-computational-efficiency-deep-learning-models-flops-macs.html#:~:text=As%20mentioned%20in%20the%20previous,MACs%20equals%20roughly%20two%20FLOPs for instance

kprokofi (dismissed):
Thank you for this feature, overall looks good

src/otx/algo/segmentation/litehrnet.py
sovrasov: Updated

src/otx/core/model/segmentation.py
sovrasov: I tried to introduce a more strict type control here, but explicit conversion doesn't make mypy happy, so I just omitted type annotation in `OTXSegmentationModel`. Can be reverted here

src/otx/core/model/anomaly.py
sovrasov: TODO has been added

src/otx/algo/segmentation/litehrnet.py
sungchul1: Actually, it seems like this hasn't been updated.

src/otx/core/model/segmentation.py
sungchul1: For safety, forcing type conversion seems fine. Could you update type annotation for `self.image_size` to `tuple[int, ...] | None` at L40?
https://github.com/openvinotoolkit/training_extensions/blob/9a1e80092d26162326dfeb15a051d451da956f97/src/otx/core/model/segmentation.py#L40

harimkang (dismissed):

src/otx/algo/segmentation/litehrnet.py
sovrasov: Updated this time, last time I've done that for all other models

src/otx/core/model/segmentation.py
sovrasov: Int tuple type was restored

eunwoosh (approved):

kprokofi (approved):

Issue #3635, opened 2024-06-18
Problems in multi-card distributed training


nowbug:
**The problem of distributed training blocking**

**Steps to Reproduce**

1、Minimum code block

from otx.engine import Engine

engine = Engine(model="yolox_s",  data_root="pwd")
engine.train(num_nodes=2) 

2.I tried other code to troubleshoot my environment.

import lightning as L
from lightning.pytorch.demos.boring_classes import BoringModel

ngpus = 2
model = BoringModel()
trainer = L.Trainer(max_epochs=10,
                    devices=ngpus)

trainer.fit(model)

log:
Initializing distributed: GLOBAL_RANK: 0, MEMBER: 1/2
Initializing distributed: GLOBAL_RANK: 1, MEMBER: 2/2
----------------------------------------------------------------------------------------------------
distributed_backend=nccl
All distributed processes registered. Starting with 2 processes

**Environment:**

- OS: 
- Framework version: 
- Python version: 3.10
- OpenVINO version:
- CUDA/cuDNN version: 12.2
- GPU model and memory: 24G(4090)*2

nowbug:
When I run it, it will card the owner.  Initializing distributed: GLOBAL_RANK: 0, MEMBER: 1/2

harimkang:
@eunwoosh could you take a look at this issue?

harimkang:
I found some open issues related to this and commented on them.
- https://github.com/Lightning-AI/torchmetrics/issues/2477
- https://github.com/Lightning-AI/pytorch-lightning/issues/18803

eunwoosh:
Hi  @nowbug , thanks for finding the issue. First of all I want to say that OTX 2.0 currently doesn't validate distributed training, so it can be a little bit unstable. Nevertheless, OTX is based on pytorch lightning, so I think distributed training is available in most cases. OTX have a plan to support distributed training in the near future, so it can become stable soon.
And I tested with your second code snippet, and I found a bug as @harimkang  said. So, I opened PR to fix it.
I also found that distributed training is stuck in some cases, and I suspect number of dataset is cause of the problem. I'll fix that bug after finding more.

nowbug:
@eunwoosh Thank you for your response. I'm looking forward to the upcoming versions of OTX.

PR #3778, opened 2024-07-31
Add num_devices in Engine for multi-gpu training

harimkang:
### Summary

I noticed that multi-gpu settings are not available through Engine, fix this.
https://jira.devtools.intel.com/browse/CVS-148420
This allows us to set up multi-gpu in the API and CLI.

- Add `num_devices` property and setter function in `Engine`
- Update Docs to use multi-gpu training
- Update CHANGELOG

image

### How to test

1. API
```python
engine = Engine(..., num_devices=2)
```

2. CLI
```shell
otx train ... --engine.num_devices 2
```

### Checklist

### License

```python
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ All notable changes to this project will be documented in this file.
   (<https://github.com/openvinotoolkit/training_extensions/pull/3759>)
 - Enable to use polygon and bitmap mask as prompt inputs for zero-shot learning
   (<https://github.com/openvinotoolkit/training_extensions/pull/3769>)
+- Add num_devices in Engine for multi-gpu training
+  (<https://github.com/openvinotoolkit/training_extensions/pull/3778>)
 
 ### Bug fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,8 +27,6 @@ All notable changes to this project will be documented in this file.
   (<https://github.com/openvinotoolkit/training_extensions/pull/3759>)
 - Enable to use polygon and bitmap mask as prompt inputs for zero-shot learning
   (<https://github.com/openvinotoolkit/training_extensions/pull/3769>)
-- Add num_devices in Engine for multi-gpu training
-  (<https://github.com/openvinotoolkit/training_extensions/pull/3778>)
 
 ### Bug fixes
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -27,6 +27,8 @@ All notable changes to this project will be documented in this file.
   (<https://github.com/openvinotoolkit/training_extensions/pull/3759>)
 - Enable to use polygon and bitmap mask as prompt inputs for zero-shot learning
   (<https://github.com/openvinotoolkit/training_extensions/pull/3769>)
+- Add num_devices in Engine for multi-gpu training
+  (<https://github.com/openvinotoolkit/training_extensions/pull/3778>)
```
sungchul1: I think it'd be better to be moved into `Bug fixes`.
The ticket was set as a bug and this PR seems to fix that multi-gpu settings are unavailable through Engine.
What do you think?

harimkang: Thanks! i updated CHANGELOG - https://github.com/openvinotoolkit/training_extensions/pull/3778/commits/b6ad495c275325a80aebc80278fd518103e30518

sungchul1 (approved):

eunwoosh:
I'm not sure OTX is already supporting multi GPU training. As far as I know, we haven't checked all models can be trained on multi GPU. And I also think that we should add integration test to validate distributed training if we really support it. If it's just for preparing, then I think it's ok to merge after reverting documentation.

harimkang:
> I'm not sure OTX is already supporting multi GPU training. As far as I know, we haven't checked all models can be trained on multi GPU. And I also think that we should add integration test to validate distributed training if we really support it. If it's just for preparing, then I think it's ok to merge after reverting documentation.

Yes you are right, but this is a solution to the issue that there is no way to use it, so it is a PR to fix the issue. I agree that we should do validation for all models, but we should at least make this work in OTX. (Anyway, They also all work for Classification.)

eunwoosh:
> > I'm not sure OTX is already supporting multi GPU training. As far as I know, we haven't checked all models can be trained on multi GPU. And I also think that we should add integration test to validate distributed training if we really support it. If it's just for preparing, then I think it's ok to merge after reverting documentation.
> 
> Yes you are right, but this is a solution to the issue that there is no way to use it, so it is a PR to fix the issue. I agree that we should do validation for all models, but we should at least make this work in OTX. (Anyway, They also all work for Classification.)

I understood. But honestly, I think it's hard to say that OTX supports multi GPU training. Currently, yolox models can't be trained on multi GPU (please refer #3635). I propose to notify that classification is only validated in documentation at least.

kprokofi (approved):
Agree with Eunwoo comment. We need to validate all tasks/models and handle the errors if it is not possible to train on multi GPUs

PR #3783, opened 2024-08-02
Refactoring base module (`ConvModule`)

sungchul1:
### Summary

This PR includes:

- [x] Split `ConvModule` to `Conv2dModule` and `Conv3dModule` (for x3d)
- [x] Remove `conv_cfg`
- [x] Remove `build_conv_layer`
- [x] Move `DepthwiseSeparableConvModule` into `conv_module`

### Checklist

### License

```python
```


src/otx/algo/modules/conv_module.py
```diff
@@ -95,8 +95,6 @@ class ConvModule(nn.Module):
         bias (bool | str): If specified as `auto`, it will be decided by the
             norm_cfg. Bias will be set as True if `norm_cfg` is None, otherwise
             False. Default: "auto".
-        conv_cfg (dict): Config dict for convolution layer. Default: None,
```
kprokofi, line 98: Could you update other sections of the docstring like "It is based upon three build methods: `build_conv_layer()`..."

src/otx/algo/modules/conv_module.py
```diff
@@ -197,18 +192,18 @@ def __init__(
...
         else:
-            self.norm_name = None
+            self.norm_name = None  # type: ignore[assignment]
 
         self.turn_on_efficient_conv_bn_eval(efficient_conv_bn_eval)
 
         # build activation layer
         if self.with_activation:
-            act_cfg_ = act_cfg.copy()
+            act_cfg_ = act_cfg.copy()  # type: ignore[union-attr]
```
kprokofi, line 152: Do you plan to eliminate "act_cfg" and "norm_cfg" in the this PR or different PR is expected?

src/otx/algo/modules/conv_module.py
```diff
@@ -248,26 +243,26 @@ def init_weights(self) -> None:
         # Note: For PyTorch's conv layers, they will be overwritten by our
         #    initialization implementation using default ``kaiming_init``.
         if not hasattr(self.conv, "init_weights"):
-            if self.with_activation and self.act_cfg["type"] == "LeakyReLU":
+            if self.with_activation and self.act_cfg["type"] == "LeakyReLU":  # type: ignore[index]
```
kprokofi, line 192: I have a question. Can we remove such mypy problems without #type ignore. You removed the ignoring of the entire file, but now a lot of "# type: ignore" is added there. What's the idea?

src/otx/algo/modules/conv_module.py
```diff
@@ -3,31 +3,31 @@
...
-from .conv import build_conv_layer
 from .norm import build_norm_layer
 from .padding import build_padding_layer
 
 if TYPE_CHECKING:
     from torch.nn.modules.conv import _ConvNd as ConvNd
 
 
-def efficient_conv_bn_eval_forward(bn: BatchNorm, conv: ConvNd, x: torch.Tensor) -> torch.Tensor:
+def efficient_conv_bn_eval_forward(bn: BatchNorm, conv: ConvNd, x: Tensor) -> Tensor:
```
kprokofi: Do we use it for some models?

src/otx/algo/modules/conv_module.py
```diff
@@ -248,26 +243,26 @@ def init_weights(self) -> None:
         # Note: For PyTorch's conv layers, they will be overwritten by our
         #    initialization implementation using default ``kaiming_init``.
         if not hasattr(self.conv, "init_weights"):
-            if self.with_activation and self.act_cfg["type"] == "LeakyReLU":
+            if self.with_activation and self.act_cfg["type"] == "LeakyReLU":  # type: ignore[index]
                 nonlinearity = "leaky_relu"
-                a = self.act_cfg.get("negative_slope", 0.01)
+                a = self.act_cfg.get("negative_slope", 0.01)  # type: ignore[union-attr]
```
kprokofi, line 194: Also, I see some asserts above that just removed from "black" scope with "noqa". I think we should change this behavior. Could you remove asserts and add raising an error instead?

src/otx/algo/modules/conv_module.py
```diff
@@ -248,26 +243,26 @@ def init_weights(self) -> None:
...
             else:
                 nonlinearity = "relu"
                 a = 0
             kaiming_init(self.conv, a=a, nonlinearity=nonlinearity)
```
kprokofi, line 198: Do we need custom "init_weights" method for ConvModule? I think initializing of the model weights should be done on the model side. I propose to remove redundant functionality from ConvModule scope

src/otx/algo/modules/conv_module.py
```diff
@@ -248,26 +243,26 @@ def init_weights(self) -> None:
...
+            x (Tensor): Input tensor.
             activate (bool, optional): Whether to apply activation. Defaults to True.
             norm (bool, optional): Whether to apply normalization. Defaults to True.
 
         Returns:
-            torch.Tensor: Output tensor.
+            Tensor: Output tensor.
         """
         layer_index = 0
         while layer_index < len(self.order):
```
kprokofi: Do we really change order somewhere? I think it is redundant "feature". If we don't really change the order, I propose to simplify this part

kprokofi (commented):
Thank you for your work! 
I left some questions/comments

src/otx/algo/modules/conv_module.py
eunwoosh: I think it's better to keep mypy to check this file. Although it could be a little bit verbose, but it can still keep code clearer.

src/otx/algo/instance_segmentation/layers/transformer.py
```diff
@@ -91,11 +90,11 @@ class PatchEmbed(BaseModule):
 
     We use a conv layer to implement PatchEmbed.
 
+    TODO (sungchul): it is duplicated with otx.algo.modules.transformer.PatchEmbed
```
eunwoosh, line 93: If then, how about removing one and using other one in all code?

src/otx/algo/modules/conv_module.py
sungchul1: I'll do it in the next PR.

eunwoosh (commented):
Thanks for your work. mostly LGTM. I left minor comment. Please take a look.

src/otx/algo/modules/conv_module.py
sungchul1: Oh, it seems that it is not used anywhere. @harimkang could you check if it is used?

src/otx/algo/modules/conv_module.py
sungchul1: I did it to specify lines to remove all mypy errors in the next PR.
To remove them, `norm_cfg` and `act_cfg` should be removed as well.

src/otx/algo/modules/conv_module.py
harimkang: Yes I don't think it's used, but need to check the other tasks as well.

src/otx/algo/modules/conv_module.py
sungchul1: I agree with simplifying `init_weights` structure.
But I think removing `init_weights` must be accompanied by updating `BaseModule`, and it is not a scope for this PR.
Let's talk about the plan for this.
https://github.com/openvinotoolkit/training_extensions/blob/6012bd298a6d8c53dfbbd5235684b52ed4926361/src/otx/algo/modules/base_module.py#L134-L143

src/otx/algo/instance_segmentation/layers/transformer.py
sungchul1: How about doing this when refactoring instance segmentation?

src/otx/algo/modules/conv_module.py
sungchul1: Yes, it is not an essential feature.
But `DepthwiseSeparableConvModule` uses `Torch` instead of `torch.Tensor` and in OTX `Tensor` seems to be used more than `torch.Tensor`.
So I just tried to match the format used within the file and follow this trend.

src/otx/algo/modules/conv_module.py
sungchul1: image
It seems not to be used anywhere, I'll remove it.
But, it seems a good feature by reducing memory and computation time, need more discussion about whether to use this.

eunwoosh (dismissed):

src/otx/algo/modules/conv_module.py
harimkang: this should fundamentally solve the mypy problem. As Sungchul said, if `act_cfg` is cleared, it seems like it would solve it.

src/otx/algo/modules/conv_module.py
sungchul1: Done ec9c7f1

src/otx/algo/modules/conv_module.py
sungchul1: Done at 28e7098 except for `norm_cfg` to be removed.

src/otx/algo/modules/conv_module.py
sungchul1: Done ae5e0ae

harimkang (dismissed):

src/otx/algo/modules/conv_module.py
kprokofi: Yes, I don't have any objections about using Torch for typing. My comment was about "self.order". This parameter set the order of the components (conv, norm, act). I wonder do we change the order somewhere and it is useful feature or not. Could we remove "order" functionality from the ConvModule?

src/otx/algo/modules/conv_module.py
kprokofi: Ok. Let's get back to this after removing/modifying BaseModule

src/otx/algo/modules/conv_module.py
sungchul1: Sorry for the misunderstanding.
It seems not to be used anywhere, I'll remove it and and let's check integration tests.
ec353de

kprokofi (approved):

eunwoosh (approved):

PR #3788, opened 2024-08-05
Support configurable input size

eunwoosh:
### Summary
 This PR includes
* adds configurable input size and adaptive input size to OTX.
* refactor input size in model code.

#### adds configurable input size feature to OTX
User can configure desired input size after this PR. `input_size` value in 'data' part of recipe is automatically passed to model unless model is already initialized before it's passed to the engine in API use case.
Also, user can use adaptive input size which finds appropriate input size based on dataset statistics. Also, user can force to use downscale only which doesn't change input size if bigger input size is considered as proper input size.
All interfaces for input size are added to `OTXDataModule`.
So CLI user can uses it by adding `--data.input_size 1000` for configuring input size and `--data.adaptive_input_size "auto"` for adaptive input size.
API user can use it by setting arguments of `OTXDataModule`. If model is initialized in engine, input_size will be automatically set but if not, then user should set input_size argument of the model by himself or herself.

#### refactor input size in model code
Previously, how input_size is manged in model side is different depending on each model.
Now, it's unified. All models get `input_size` argument and keep it as attribute except zvp model which uses fixed input_size.
Model also initializes own module differently based on input_size if necessary.

### How to test
* otx train --config ...  --data_root ...  --data.input_size 1024
* otx train --config ...  --data_root ...  --data.adaptive_input_size "auto | downscale"
 

### Checklist

### License

```python
```


src/otx/algo/classification/huggingface_model.py
```diff
@@ -70,13 +78,23 @@ def __init__(
...
             metric=metric,
             torch_compile=torch_compile,
+            input_size=input_size,
         )
 
     def _create_model(self) -> nn.Module:
+        model_config, _ = PretrainedConfig.get_config_dict(self.model_name)
+        kwargs = {}
+        if "image_size" in model_config:
+            kwargs["image_size"] = self.input_size[-1]
```
goodsong81: What's the meaning of -1 here?

src/otx/algo/classification/mobilenet_v3.py
```diff
@@ -61,6 +61,7 @@ class MobileNetV3ForMulticlassCls(OTXMulticlassClsModel):
         metric (MetricCallable, optional): The metric callable. Defaults to MultiClassClsMetricCallable.
         torch_compile (bool, optional): Whether to compile the model using TorchScript. Defaults to False.
         freeze_backbone (bool, optional): Whether to freeze the backbone layers during training. Defaults to False.
+        input_size (tuple[int, ...], optional): The input size of the model. Defaults to (1, 3, 224, 224)
```
goodsong81: (Not specific to this module, but in general)
Can we control the batch size and number of channels by `input_size` argument?
It's weird to see

- batch size setting while it's dynamic in most cases (-1)
- number of channels setting while the mean/std are hardcoded as 3D vectors.

src/otx/algo/detection/rtdetr.py
```diff
@@ -212,13 +242,13 @@ def _build_model(self, num_classes: int) -> nn.Module:
...
             expansion=0.5,
-            eval_spatial_size=self.image_size[2:],
+            eval_spatial_size=self.input_size[-2:],
         )
         decoder = RTDETRTransformer(
             num_classes=num_classes,
             num_decoder_layers=3,
             feat_channels=[256, 256, 256],
-            eval_spatial_size=self.image_size[2:],
+            eval_spatial_size=self.input_size[-2:],
```
goodsong81: Are the spatial sizes are always [H, W] in order?

src/otx/algo/detection/yolox.py
```diff
@@ -40,16 +71,15 @@ def _customize_inputs(
     @property
     def _exporter(self) -> OTXModelExporter:
         """Creates OTXModelExporter object that can export the model."""
-        if self.image_size is None:
-            msg = f"Image size attribute is not set for {self.__class__}"
+        if self.input_size is None:
+            msg = f"Inputhsize attribute is not set for {self.__class__}"
```
goodsong81:
```suggestion
            msg = f"Input size attribute is not set for {self.__class__}"
```

src/otx/cli/cli.py
```diff
@@ -331,11 +332,24 @@ def instantiate_classes(self, instantiate_engine: bool = True) -> None:
...
             self.config_init = self.parser.instantiate_classes(self.config)
             self.workspace = self.get_config_value(self.config_init, "workspace")
             self.datamodule = self.get_config_value(self.config_init, "data")
 
+            # pass OTXDataModule input size to the model
+            if (input_size := self.datamodule.input_size) is not None:
+                input_size = (input_size, input_size) if isinstance(input_size, int) else tuple(input_size)  # type: ignore[assignment]
+                model_config["init_args"]["input_size"] = (
+                    tuple(model_config["init_args"]["input_size"][:-2]) + input_size  # type: ignore[operator]
+                )
```
goodsong81: Definition of `input_size` is not consistent.
Sometimes it's Hxw, but BxCxHxW mostly.
Better if consolidated. But not that important, just for code readers.

src/otx/core/data/module.py
```diff
@@ -63,18 +63,15 @@ def __init__(
         auto_num_workers: bool = False,
         device: DeviceType = DeviceType.auto,
         input_size: int | tuple[int, int] | None = None,
+        adaptive_input_size: Literal["auto", "downscale", "none"] = "none",
```
goodsong81:
```suggestion
        input_size_adapter: Literal["auto", "downscale", "none"] = "none",
```
Might be consistent w/ `input_size_multiplier`

src/otx/core/data/utils/utils.py
```diff
@@ -0,0 +1,269 @@
...
+        input_size_multiplier (int | None, optional):
+            Multiplier for input size. If it's set, return the input size which can be divisible by the value.
+            Defaults to None.
+
+    Returns:
+        tuple[int, int] | None: Recommended input size based on dataset statistics.
+    """
+    min_recognizable_object_size = 32  # Minimum object size recognizable by NNs: typically 16 ~ 32
+    # meaning NxN input pixels being downscaled to 1x1 on feature map
+    min_detection_input_size = 256  # Minimum input size for object detection
```
goodsong81: Better to be configurable via params w/ default value, or at least declared as CONSTANT.

src/otx/core/model/action_classification.py
```diff
@@ -37,21 +37,23 @@ class OTXActionClsModel(OTXModel[ActionClsBatchDataEntity, ActionClsBatchPredEnt
...
         self.std = (255.0, 255.0, 255.0)
         super().__init__(
             label_info=label_info,
+            input_size=input_size,
             optimizer=optimizer,
             scheduler=scheduler,
             metric=metric,
             torch_compile=torch_compile,
         )
+        self.input_size: tuple[int, int, int, int, int, int]
```
goodsong81: Just out of curiosity, why this needed?

src/otx/engine/utils/auto_configurator.py
```diff
@@ -240,13 +245,19 @@ def get_datamodule(self) -> OTXDataModule | None:
...
+        label_info: LabelInfoTypes | None = None,
+        input_size: tuple[int, ...] | int | None = None,
+    ) -> OTXModel:
         """Retrieves the OTXModel instance based on the provided model name and meta information.
 
         Args:
             model_name (str | None): The name of the model to retrieve. If None, the default model will be used.
             label_info (LabelInfoTypes | None): The meta information about the labels.
                 If provided, the number of classes will be updated in the model's configuration.
+            input_size (tuple[int, ...] | int | None, optional): Input size of the model. Defaults to None.
```
goodsong81: Maybe we need some description somewhere.
100 -> HxW=100x100
(200,100) -> HxW=200x100
BTW, cannot set (1,3,200,100)?

goodsong81 (commented):
Nice work. A few points to mention are as follows:
* `hasatrr()` based branch logics might be quickly rusted and tangled. Hope it would be based on formal interfaces in the future.
* `input_size` term might be better to be more consistent in terms of usage

src/otx/algo/detection/rtdetr.py
eunwoosh: yes it should be (h, w) order

src/otx/core/data/utils/utils.py
eunwoosh: yes, I changed them constant.

src/otx/core/model/action_classification.py
eunwoosh: OTXModel has `tuple[int, ...] | None` as typehint and mypy raises an error for that. This can specify exact input_size type hint for each model.

src/otx/algo/classification/huggingface_model.py
eunwoosh: it needs to get single value so I just gave one of input size.

src/otx/algo/classification/mobilenet_v3.py
eunwoosh: now, input size includes only height and wdith.

src/otx/algo/detection/yolox.py
eunwoosh: done

src/otx/core/data/module.py
eunwoosh: IMHO, adaptive_input_size looks more suitable for me. how about leaving it as it is?

src/otx/engine/utils/auto_configurator.py
eunwoosh: I updated it

eunwoosh:
> Nice work. A few points to mention are as follows:
> 
> * `hasatrr()` based branch logics might be quickly rusted and tangled. Hope it would be based on formal interfaces in the future.
> * `input_size` term might be better to be more consistent in terms of usage

Thanks for your feedback :) I updated the code applying your comment. Please review it.

j99ca:
Very nice work @eunwoosh! Can the input sizes be set to non-square resolutions if the backbone used square input sizes? I imagine with a proper pooling layer towards the end layer should make this not much of a problem?

E.g. I want to use 320x256 (WxH) on MobilenetV3 which was trained with 224x224?

eunwoosh:
> Very nice work @eunwoosh! Can the input sizes be set to non-square resolutions if the backbone used square input sizes? I imagine with a proper pooling layer towards the end layer should make this not much of a problem?
> 
> E.g. I want to use 320x256 (WxH) on MobilenetV3 which was trained with 224x224?

Thanks :) yes, you can change input size to non-square size. I think your example will work.

src/otx/core/data/module.py
goodsong81: It looked like "input size adaptation method or criteria" to me so that I suggested the "adapter". Feel free to keep it or not :)

goodsong81 (dismissed):

src/otx/core/data/module.py
```diff
@@ -62,18 +62,15 @@ def __init__(
...
```
harimkang: Can we consider this?
```suggestion
        adaptive_input_size: Literal["auto", "downscale"] | None = None,
```

src/otx/core/data/module.py
```diff
@@ -62,18 +62,15 @@ def __init__(
...
+        input_size_multiplier: int = 1,
```
harimkang, line 79: When is this value used?

In my opinion, it's hard for people using the API to understand the functionality at a glance by the name of the argument. Is it possible to add a docstring for the `OTXDataModule`?

src/otx/core/model/action_classification.py
```diff
@@ -37,21 +37,23 @@ class OTXActionClsModel(OTXModel[ActionClsBatchDataEntity, ActionClsBatchPredEnt
...
         self.std = (255.0, 255.0, 255.0)
         super().__init__(
             label_info=label_info,
+            input_size=input_size,
             optimizer=optimizer,
             scheduler=scheduler,
             metric=metric,
             torch_compile=torch_compile,
         )
+        self.input_size: tuple[int, int]
```
harimkang, line 56: Why are we showing the type of this here? Doesn't the `OTXModel` already declare `input_size` as a tuple?

src/otx/core/data/module.py
```diff
@@ -95,7 +92,7 @@ def __init__(
         self.device = device
 
         self.subsets: dict[str, OTXDataset] = {}
-        self.save_hyperparameters()
+        self.save_hyperparameters(ignore=["input_size"])
```
sungchul1, line 108: Just a question, why is it ignored?
Is it an important key for training, especially if the default values are not used?
I thought it is important to save the hyper parameters that are used in training.

src/otx/cli/cli.py
```diff
@@ -23,6 +23,7 @@
 from otx.cli.utils.workspace import Workspace
 from otx.core.types.task import OTXTaskType
 from otx.core.utils.imports import get_otx_root_path
+from otx.utils.utils import get_model_cls_from_config
```
harimkang: This import internally imports an `OTXModel`, which requires `torch`. The OTX CLI currently has a case where it works without torch, so the import location doesn't seem appropriate.

src/otx/cli/cli.py
```diff
@@ -331,11 +332,23 @@ def instantiate_classes(self, instantiate_engine: bool = True) -> None:
             # For num_classes update, Model and Metric are instantiated separately.
             model_config = self.config[self.subcommand].pop("model")
 
+            # if adaptive_input_size will be executed and the model has input_size_multiplier, pass it to OTXDataModule
+            if self.config[self.subcommand].data.adaptive_input_size != "none":
+                model_cls = get_model_cls_from_config(model_config)
+                self.config[self.subcommand].data.input_size_multiplier = model_cls.input_size_multiplier
```
harimkang: Is this necessary because different models have different `input_size_multiplier`?

src/otx/cli/cli.py
```diff
@@ -331,11 +332,23 @@ def instantiate_classes(self, instantiate_engine: bool = True) -> None:
...
+
             # Instantiate the things that don't need to special handling
             self.config_init = self.parser.instantiate_classes(self.config)
             self.workspace = self.get_config_value(self.config_init, "workspace")
             self.datamodule = self.get_config_value(self.config_init, "data")
 
+            # pass OTXDataModule input size to the model
+            if (input_size := self.datamodule.input_size) is not None and "input_size" in model_config["init_args"]:
+                # TODO(eunwoosh): After configurable input size is applied to anomaly, remove input_size check
```
harimkang: I think it would be better to just let Anomaly take `input_size` as well. Are we going to have trouble with this?

src/otx/core/model/base.py
```diff
@@ -100,10 +99,12 @@ class OTXModel(LightningModule, Generic[T_OTXBatchDataEntity, T_OTXBatchPredEnti
     """
 
     _OPTIMIZED_MODEL_BASE_NAME: str = "optimized_model"
+    input_size_multiplier = 1
```
harimkang: In my opinion, it would be nice to have a docstring for this, this is a variable that doesn't make sense at first glance.

harimkang (commented):
LGTM, i left some comments

src/otx/core/model/action_classification.py
eunwoosh: `OTXModel ` uses `tuple[int, int] | None` as type hint of `input_size`, which makes mypy error. For that reason, type of `input_size` is declared explicitly here.

src/otx/cli/cli.py
eunwoosh: yes you're right. it depends on each model.

src/otx/cli/cli.py
eunwoosh: Anomaly has complex hierarchical structure, which makes it hard to update input_size feature on anomaly task. And it's being refactored So, I excluded anomaly tasks from this PR. I think it's better to update it after refactoring is done.

src/otx/core/data/module.py
eunwoosh: Some models require input_size to be multiple of specific value (e.g. yolox should use 32xN input_size). I'll update docstring for this.

src/otx/core/model/classification.py
```diff
@@ -56,12 +57,13 @@ def __init__(
...
         super().__init__(
             label_info=label_info,
+            input_size=input_size,
             optimizer=optimizer,
             scheduler=scheduler,
             metric=metric,
             torch_compile=torch_compile,
         )
-        self.image_size = (1, 3, 224, 224)
+        self.input_size: tuple[int, int]
```
sungchul1, line 66: Why is this added to every model? Isn't it possible to add it to the base model only?

src/otx/core/model/action_classification.py
harimkang: I have one more question. Task-specific `OTXModels` don't seem to have an `input_size` of `None`, nor do they have a default, so is there any reason for an `OTXModel` to have an `input_size` type of `None`?

eunwoosh: yes it's because some classes like `OVModel`  inherited from `OTXModel` need to set it to None.

src/otx/core/data/module.py
```diff
@@ -62,18 +62,15 @@ def __init__(
...
         """Constructor."""
         super().__init__()
         self.task = task
         self.data_format = data_format
         self.data_root = data_root
 
-        if input_size is not None:
-            for subset_cfg in [train_subset, val_subset, test_subset, unlabeled_subset]:
-                if subset_cfg.input_size is None:
-                    subset_cfg.input_size = input_size
```
sungchul1, line 75: I don't understand why this is removed, recipe updates, and `how torchvision transform factor applies $(input_size) is slightly changed` in your summary.
In original impl, if users want to set different `input_size` only for validation, the users can just set an additional `input_size` in the validation section.
But your update doesn't consider `input_size` as the global parameter.
I think that subset config manually set by the users in the recipe must have a higher priority.

src/otx/recipe/detection/rtdetr_101.yaml
sungchul1, line 1: Just for your implementation, rtdetr family uses a global `input_size`, not for each subset.

sungchul1 (commented):
Thanks for your huge update.
I left some comments and I'd like to suggest something more.
1. Could you add documentation for this key feature?
2. Could you update integration tests for this update?

src/otx/core/model/classification.py
eunwoosh: Could you refer the question Harim asked? I left some comments for that. Here is link https://github.com/openvinotoolkit/training_extensions/pull/3788#discussion_r1713102405 :)

src/otx/core/data/module.py
eunwoosh: First of all, the reason why it's added is because lightning compares the value if same attribute exists in both model and dataset side. It's ok for other tasks but for anomaly tasks, it makes some errors because 3 parent classes of anomaly model have input_size attribute and it makes tricky problem currently.
And I think we don't have to save this value in `OTXDatamodule` because input_size values are saved in subset config and transforms as below
image

src/otx/core/data/module.py
eunwoosh: I want to keep this policy in this PR but there is no way to distinguish between input_size user sets and existing input_size. For example, when using yolox_tiny model which sets input_size to val_subset and test_subset, user need to remove those values in the recipe to use configurable input_size because user's input_size can't be applied with previous policy. So, I have thought how to deal with it, there is no way except current change. If you have some idea of this, please tell me. I can try it instead of this :)

src/otx/recipe/detection/rtdetr_101.yaml
eunwoosh: Sorry I can't understand your intention. Could you explain more details?

src/otx/core/data/module.py
eunwoosh: done https://github.com/openvinotoolkit/training_extensions/pull/3788/commits/0adb7ea54bb57b761f75b423167b45772d08d416

src/otx/core/data/module.py
eunwoosh: done https://github.com/openvinotoolkit/training_extensions/pull/3788/commits/0adb7ea54bb57b761f75b423167b45772d08d416

src/otx/cli/cli.py
eunwoosh: done https://github.com/openvinotoolkit/training_extensions/pull/3788/commits/0adb7ea54bb57b761f75b423167b45772d08d416

src/otx/core/model/base.py
eunwoosh: done https://github.com/openvinotoolkit/training_extensions/pull/3788/commits/0adb7ea54bb57b761f75b423167b45772d08d416

harimkang (dismissed):

src/otx/core/data/module.py
sungchul1: I think it doesn't matter that users remove those values for their custom configurations.
It is their responsibility because we just support our optimal configurations.

sungchul1: Of course, I'm not sure if it is possible, but we can update `input_size` in valset and testset to `$(input_size) * 0.65` in yolox_tiny.

src/otx/recipe/detection/rtdetr_101.yaml
sungchul1: Just misunderstanding, don't mind.

src/otx/core/data/module.py
eunwoosh: thanks for sharing your idea. I updated it as you said. https://github.com/openvinotoolkit/training_extensions/pull/3788/commits/8d34d2c3c0cf5891865238881f92686df1075b8e

src/otx/core/data/utils/utils.py
```diff
@@ -0,0 +1,279 @@
...
+        return {}
+
+    stat = compute_robust_statistics(np.log(values))
+    stat = {k: float(np.exp(v)) for k, v in stat.items()}
+    # Normal scale std is easier to understand
+    stat["std"] = float(np.std(values))
+    return stat
+
+
+def compute_robust_dataset_statistics(dataset: DatasetSubset, max_samples: int = 1000) -> dict[str, Any]:
```
eunwoosh, line 78: @eugene123tw , could you check this part? :)

eugene123tw (dismissed):

eunwoosh:
> Thanks for your huge update. I left some comments and I'd like to suggest something more.
> 
> 1. Could you add documentation for this key feature?
> 2. Could you update integration tests for this update?

Thanks for comments. I added integration test and I'll make a separate PR to write documentation. Please review changes

src/otx/recipe/detection/yolox_tiny.yaml
```diff
@@ -38,23 +38,23 @@ overrides:
...
+      - 416
     train_subset:
       batch_size: 8
       transforms:
         - class_path: otx.core.data.transform_libs.torchvision.CachedMosaic
           init_args:
             random_pop: false
             max_cached_images: 20
-            img_scale: $(input_size) # (H, W)
+            img_scale: $(input_size) * 1.538 # 640x640
```
sovrasov, line 50: Why do we need this trick? Can we just declare 640 as input?

src/otx/engine/utils/auto_configurator.py
```diff
@@ -274,6 +286,11 @@ def get_model(self, model_name: str | None = None, label_info: LabelInfoTypes |
 
         model_config = deepcopy(self.config["model"])
 
+        if input_size is not None:
+            model_config["init_args"]["input_size"] = (
+                (input_size, input_size) if isinstance(input_size, int) else input_size
+            )
```
sovrasov, line 292: Are we safe to forward also mean/scale form recipe config here?
Related to https://github.com/openvinotoolkit/training_extensions/issues/3827

src/otx/recipe/detection/yolox_tiny.yaml
eunwoosh: It can, but if setting input_size in subset config, data.input_size is ignored, so user can't use configurable input size until removing input_size in subset config. So, I unified all subset config to use data.input_size.

src/otx/engine/utils/auto_configurator.py
eunwoosh: yes I think. but currently, there are no arguments for model init function. We need to refactor model code to remove hard coded mean and std. It's already discussed so, we can refactor that part later.

src/otx/core/data/transform_libs/torchvision.py
```diff
@@ -3508,22 +3508,20 @@ class GenerateTarget(tvt_v2.Transform, NumpytoTVTensorMixin):
             the specific codec for more details.
 
     Args:
-        encoder (dict | list[dict]): The codec config for keypoint encoding.
-            Both single encoder and multiple encoders (given as a list) are
-            supported
-        target_type (str, deprecated): This argument is deprecated and has no
-            effect. Defaults to ``None``
+        input_size (tuple[int, int]): Input image size in [w, h]
```
sungchul1: Could you leave TODO comments to change shape order?

eunwoosh: done https://github.com/openvinotoolkit/training_extensions/pull/3788/commits/9dccf2d0550c5bff84e61d09dd080a4017ff2c0a

harimkang (dismissed):
LGTM!

wonjuleee (dismissed):
LGTM

src/otx/core/data/transform_libs/torchvision.py
eunwoosh: I changed order from (w, h) to (h, w) https://github.com/openvinotoolkit/training_extensions/pull/3788/commits/9127e5d152ead7cef2fc98a9b5dde77002dbdce0

harimkang (dismissed):

wonjuleee (dismissed):

sungchul1 (dismissed):

harimkang (dismissed):

sovrasov (dismissed):

sooahleex (approved):
Looks good to me for h-label head!

harimkang (approved):

goodsong81 (approved):
LGTM w/ one last comment: we may want to touch the mem cache max image size to make this feature more effective in training
(Maybe in another PR? :p)

eunwoosh:
> LGTM w/ one last comment: we may want to touch the mem cache max image size to make this feature more effective in training (Maybe in another PR? :p)

yes, we can consider including that feature in the next release.

goodsong81:
> > LGTM w/ one last comment: we may want to touch the mem cache max image size to make this feature more effective in training (Maybe in another PR? :p)
> 
> yes, we can consider including that feature in the next release.

FYI, the configured input size could be set to the max memcache image size.
image
image

eunwoosh:
@goodsong81  I understand your intention. If it doesn't need huge code change, maybe it can be included in this release. Thanks for comment :)

PR #3790, opened 2024-08-05
Refactor OTXModel for Classification Task

harimkang:
### Summary

Motivation: Classification is divided into 3 main sub-tasks (`MULTI_CLASS_CLS`, `MULTI_LABEL_CLS`, `H_LABEL_CLS`) The models differ only in the `head`, but otherwise have the same structure and code. Also, the model classes are separated for each task, so there are too many classes, which can be bad in terms of usability. It also creates unnecessary duplicate code. So I consolidated this into one `OTXModel` where user can select `task`, `train_type`, and reduce duplicate code and increase usability.

- Introduce New OTXModel Class for Classification that include all task & train_type
- Introduce MetricCallablePerTask for Union sub-task metric of classification
- Replace EfficientNet-B0, V2, MobileNet-V3, ViT (all-recipes)
- Remove dino_v2.py (Replace with vit.py in https://github.com/openvinotoolkit/training_extensions/pull/3719)
- Benchmark Result
image

### How to test

1. As-Is
```python
# Multi-Class Classification
from otx.algo.classification.efficientnet import EfficientNetForMulticlassCls
multi_class_model = EfficientNetForMulticlassCls(**)

# Multi-Label Classification
from otx.algo.classification.efficientnet import EfficientNetForMultilabelCls
multi_label_model = EfficientNetForMultilabelCls(**)

# H-Label Classification
from otx.algo.classification.efficientnet import EfficientNetForHLabelCls
h_label_model = EfficientNetForHLabelCls(**)

# Semi-SL
from otx.algo.classification.efficientnet import EfficientNetForMulticlassClsSemiSL
semi_sl_model = EfficientNetForMulticlassClsSemiSL(**)
```

2. To-Be
```python
# Multi-Class Classification
from otx.algo.classification.efficientnet import EfficientNetForClassification
multi_class_model = EfficientNetForClassification(**, task="MULTI_CLASS_CLS")

# Multi-Label Classification
multi_label_model = EfficientNetForClassification(**, task="MULTI_LABEL_CLS")

# H-Label Classification
h_label_model = EfficientNetForClassification(**, task="H_LABEL_CLS")

# Semi-SL
semi_sl_model = EfficientNetForClassification(**, train_type="SEMI_SUPERVISED")
```

### Checklist

### License

```python
```


src/otx/algo/classification/efficientnet.py
```diff
@@ -25,41 +24,36 @@
...
 if TYPE_CHECKING:
     from lightning.pytorch.cli import LRSchedulerCallable, OptimizerCallable
 
-    from otx.core.metrics import MetricCallable
+    from otx.core.metrics import MetricCallablePerTask
 
 
-class EfficientNetForMulticlassCls(OTXMulticlassClsModel):
-    """EfficientNet Model for multi-class classification task."""
+class EfficientNetForClassification(OTXClassificationModel):
```
sungchul1, line 40: I think `ForClassification` can be ignored because it is already located in `algo/classification`.
What do you think?

src/otx/algo/classification/huggingface_model.py
```diff
@@ -13,25 +13,24 @@
...
 if TYPE_CHECKING:
     from lightning.pytorch.cli import LRSchedulerCallable, OptimizerCallable
     from transformers.modeling_outputs import ImageClassifierOutput
 
-    from otx.core.metrics import MetricCallable
+    from otx.core.metrics import MetricCallablePerTask
 
 
-class HuggingFaceModelForMulticlassCls(OTXMulticlassClsModel):
+class HuggingFaceModelForMulticlassCls(OTXClassificationModel):
```
sungchul1, line 33: Just a question, is the reason why `MulticlassCls` is in the class name that huggingface only supports multi class cls?

src/otx/recipe/classification/multi_label_cls/mobilenet_v3_large.yaml
```diff
@@ -1,8 +1,9 @@
 model:
-  class_path: otx.algo.classification.mobilenet_v3.MobileNetV3ForMultilabelCls
+  class_path: otx.algo.classification.mobilenet_v3.MobileNetV3ForClassification
   init_args:
     mode: large
     label_info: 1000
+    task: MULTI_LABEL_CLS
```
sungchul1, line 6: Just a question, don't we need to use `task` or `train_type` as global configurations?

sungchul1 (commented):
Thanks for your effort.
I left some comments, please take a look.
And could you update license information in all files?

src/otx/algo/classification/efficientnet.py
eunwoosh: I also agree with that.

harimkang: I agree to a certain extent. It's hard to tell the tasks apart by looking at the class name, but the import address does, right?
@kprokofi What do you think about this, it's a simple naming conundrum, but in the future we may have a model that supports multiple tasks simultaneously, what should it be called?

src/otx/recipe/classification/multi_label_cls/mobilenet_v3_large.yaml
harimkang: For now, the current refactoring is to separate the 3 sub-tasks of Classification. Actually, I think this could be added to the `OTXModel` more globally as you said, but the other tasks don't need to be given a choice because each model is currently only available for a specific task.
@kprokofi I think we can take that as a discussion topic as well.

src/otx/algo/classification/huggingface_model.py
harimkang: I'm thinking about removing classes like `OTXMulticlassClsModel` (sub-task OTXModel) in the next PR, so now i'm considering about unifying the `OTXModel` used by all classification models into an `OTXClassificationModel` for now. Any other thoughts on this?

harimkang:
> Thanks for your effort. I left some comments, please take a look. And could you update license information in all files?

Thanks for comment! Update license comments in all files - https://github.com/openvinotoolkit/training_extensions/pull/3790/commits/bd47ac81558f43d3c0ad7dd0a52a1d6a7621b1c9

src/otx/algo/classification/efficientnet.py
harimkang: First, I'm going to clear the `ForClassification`.

src/otx/recipe/classification/multi_label_cls/mobilenet_v3_large.yaml
sungchul1: Yes, we need to discuss it more.
I think `task` for detection can be divided into `detection`, `rotated detection`, and `keypoint detection`, and `task` for visual prompting can be divided into `fine tuning` and `zero-shot learning`.
Also, `train_type` can be used across all tasks like semi-supervised learning or self-supervised learning for semantic segmentation.

harimkang:
@kprokofi After discussing with @eunwoosh and @sungchul2, we've identified some shortcomings in the current version of the PR structure, 
Main shortcoming: Since only the Model is integrated and not the DataEntity, an if statement was added in OTXClassificationModel side & otx/algo models, which will make it harder to maintain in the future.

so i'm going to create a new proposal to keep the MULTI_CLASS_CLS, MULTI_LABEL_CLS, and H_LABEL_CLS task-specific model classes as is (develop), but to create a new class that will collect as much duplicated code as possible.
i'll come back with new PR and compare it to that one. and we can discuss there

harimkang:
Replace it with this: https://github.com/openvinotoolkit/training_extensions/pull/3808

PR #3808, opened 2024-08-07
Refactor classification OTXModels

harimkang:
### Summary

- Move as much duplicate code as possible to the parent class.
- Modify the SemiSL model to allow to replace the existing model with train_type.
- Remove dino_v2.py (Replace with vit.py in https://github.com/openvinotoolkit/training_extensions/pull/3719)
- Next PR: torchvision models
image

### Checklist

### License

```python
```


src/otx/algo/classification/timm_model.py
```diff
@@ -0,0 +1,299 @@
...
+    from otx.core.metrics import MetricCallable
+
+
+class TimmModelForMulticlassCls(OTXMulticlassClsModel):
+    """TimmModel for multi-class classification task."""
+
+    def __init__(
+        self,
+        label_info: LabelInfoTypes,
+        backbone: TimmModelType = "efficientnetv2_s_21k",
```
sungchul1: Just a question, is it needed to set the default backbone?
If not, I think this class seems a wrapper for timm, so it would be better to be more general.
What do you think?

harimkang: Yes, I agree, I cleared the default value for backbone. - https://github.com/openvinotoolkit/training_extensions/pull/3808/commits/86b9ecacaadb7c87847e42292b0a7371fdc351f4, https://github.com/openvinotoolkit/training_extensions/pull/3808/commits/e6306574e9548c9563ad185465cba6243305aca1

src/otx/core/model/classification.py
```diff
@@ -132,7 +306,7 @@ def _convert_pred_entity_to_compute_metric(
 
     def forward_for_tracing(self, image: Tensor) -> Tensor | dict[str, Tensor]:
         """Model forward function used for the model tracing during model exportation."""
-        return self.model.forward(image, mode="tensor")
```
eunwoosh, line 135: It's just a question. Is it ok to remove `mode="tensor"`?

src/otx/algo/classification/timm_model.py
```diff
@@ -0,0 +1,299 @@
...
+            feature_vector=outputs["feature_vector"],
+        )
+
+    def forward_for_tracing(self, image: torch.Tensor) -> torch.Tensor | dict[str, torch.Tensor]:
+        """Model forward function used for the model tracing during model exportation."""
+        if self.explain_mode:
+            return self.model(images=image, mode="explain")
+
+        return self.model(images=image, mode="tensor")
+
```
eunwoosh, line 140: It seems that these two functions are common in classification model. Then how about moving them to the parent class?

src/otx/algo/classification/huggingface_model.py
```diff
@@ -105,22 +103,6 @@ def _customize_outputs(
             labels=preds,
```
eunwoosh, line 103: Can't `_customize_outputs` be removed?

eunwoosh (commented):
Thanks for your work :) I think it'd be better if we refactor semi-sl structure later.

sungchul1 (approved):

src/otx/algo/classification/timm_model.py
harimkang: These two functions are dependent on the `ImageClassifier`, so they don't apply to all classifications. I'll see if we can refactor `ImageClassifier` in the future to clean it up if possible.

src/otx/core/model/classification.py
harimkang: `mode` is currently out of place in core because it relies on `ImageClassifier`.

src/otx/algo/classification/huggingface_model.py
harimkang: I left this as is to fix the output type as `from transformers.modeling_outputs import ImageClassifierOutput` . Therefore, the `loss_dict` call is different and we need to leave this in.

eunwoosh (approved):

src/otx/algo/classification/efficientnet.py
```diff
@@ -436,11 +262,12 @@ def _build_model(self, head_config: dict) -> nn.Module:
...
 
+        backbone = OTXEfficientNet(version=self.version, pretrained=self.pretrained)
         return ImageClassifier(
-            backbone=OTXEfficientNet(version=self.version, pretrained=True),
+            backbone=backbone,
             neck=GlobalAveragePooling(dim=2),
             head=HierarchicalLinearClsHead(
-                in_channels=1280,
+                in_channels=backbone.num_features,
                 multiclass_loss=nn.CrossEntropyLoss(),
```
kprokofi, line 271: Is it possible to move loss computation from head to classifier as we discussed?

kprokofi (commented):
Overall looks good, thank you, Harim!
However, as we discussed on our refactoring meeting, we agreed to separate the areas of responsibility for the loss and the model head. Could you make such changes for CLS task as well?
You can create separate PR though

src/otx/algo/classification/efficientnet.py
harimkang: i will do next pr! thanks!

PR #3809, opened 2024-08-07
Refactoring `ConvModule` by removing `act_cfg`

sungchul1:
### Summary

This PR includes:

- [x] Replace `act_cfg` with `activation_callable`
- [x] Remove `build_activation_layer`

### Checklist

### License

```python
```


harimkang (dismissed):
LGTM!

src/otx/algo/action_classification/backbones/x3d.py
```diff
@@ -139,13 +139,13 @@ def __init__(
...
             norm_cfg=self.norm_cfg,
-            act_cfg=None,
+            activation_callable=None,
         )
 
         if self.se_ratio is not None:
             self.se_module = SEModule(planes, self.se_ratio)
 
-        self.relu = build_activation_layer(self.act_cfg) if self.act_cfg else build_activation_layer({})
+        self.relu = self.activation_callable() if self.activation_callable else nn.ReLU(inplace=True)
```
eunwoosh, line 148: If we can set other activation layer, then how about changing name from `self.relu`?

src/otx/algo/classification/backbones/efficientnet.py
```diff
@@ -212,7 +214,7 @@ def __init__(
             )
         else:
             self.fc2 = nn.Linear(in_features=mid_channels, out_features=channels)
-        self.sigmoid = build_activation_layer({"type": out_activation})
+        self.sigmoid = out_activation_callable()
```
eunwoosh, line 217: Same as above.

src/otx/algo/detection/backbones/csp_darknet.py
```diff
@@ -148,7 +149,7 @@ def __init__(
         arch_ovewrite: list | None = None,
         spp_kernal_sizes: tuple[int, ...] = (5, 9, 13),
         norm_cfg: dict | None = None,
-        act_cfg: dict | None = None,
+        activation_callable: Callable[..., nn.Module] | None = partial(nn.LeakyReLU, negative_slope=0.1),
```
eunwoosh: It seems that Swish is used as default, but it's changed to leakyrelu now. is it ok?

eunwoosh (commented):
Thanks for your work! I left some comments. Please take a look.

src/otx/algo/action_classification/backbones/x3d.py
sungchul1: I think it depends on pretrained weights, if changing this we must add changing keys of pretrained weights as well.

src/otx/algo/detection/backbones/csp_darknet.py
sungchul1: Thanks for finding, updated at 9f0d5a7.

eunwoosh (approved):

harimkang:
I've canceled the action of this PR due to issues with the integration test at the moment, please update when this PR is merged. https://github.com/openvinotoolkit/training_extensions/pull/3815

kprokofi (commented):
Thank you for the update. I have a general question. 
Can't we pass and use already instantiated activation instead of activation class (partial)?

sungchul1:
> Thank you for the update. I have a general question. Can't we pass and use already instantiated activation instead of activation class (partial)?

I considered this before, but I haven't proceeded for now because original `ConvModule`'s purpose is an instantiation using given dict config, such as `conv_cfg`, `norm_cfg`, or `act_cfg`. 
I just updated this module maintaining this purpose, but I will try your suggestion if you or others (@harimkang, @eunwoosh, or any others 😊) think that using conv, norm, or activation instances as is is needed.
I think adding this is not hard work.

eunwoosh:
> > Thank you for the update. I have a general question. Can't we pass and use already instantiated activation instead of activation class (partial)?
> 
> I considered this before, but I haven't proceeded for now because original `ConvModule`'s purpose is an instantiation using given dict config, such as `conv_cfg`, `norm_cfg`, or `act_cfg`. I just updated this module maintaining this purpose, but I will try your suggestion if you or others (@harimkang, @eunwoosh, or any others 😊) think that using conv, norm, or activation instances as is is needed. I think adding this is not hard work.

I actually think this approach when reviewing norm refactoring but it also can be activation layer. I think current approach is ok but Kirill's opinion is also good.

sungchul1:
> Thank you for the update. I have a general question. Can't we pass and use already instantiated activation instead of activation class (partial)?

I'll try this targeting the next version as a new minor feature, conv and normalization can be also appiled.

harimkang (approved):

PR #3811, opened 2024-08-07
Refactor OTXAnomaly

ashwinvaidya17:
### Summary

Makes `OTXModel` the base class of `OTXAnomaly`.

### Checklist

### License

```python
```


sovrasov (approved):

src/otx/algo/anomaly/padim.py
```diff
@@ -132,3 +131,16 @@ def predict_step(
         if not isinstance(inputs, dict):
             inputs = self._customize_inputs(inputs)
         return AnomalibPadim.predict_step(self, inputs, batch_idx, **kwargs)  # type: ignore[misc]
+
+    def forward(
```
eunwoosh, line 135: If this function is neccesary for both padim and stfpm, how about moving it to `OTXAnomaly` class?

src/otx/core/model/anomaly.py
```diff
@@ -154,37 +155,6 @@ def configure_callbacks(self) -> list[Callback]:
             ),
         ]
 
-    def on_test_batch_end(
```
eunwoosh, line 157: It's just a question. Is it ok to remove? It seems that there is no corresponding code addition to this.

eunwoosh (commented):

src/otx/algo/anomaly/padim.py
ashwinvaidya17: The problem is that it calls `self.validation_step` which ends up calling OTXModel's validation step. We need AnomalibModule's validation step

src/otx/core/model/anomaly.py
ashwinvaidya17: Yeah. I think we can remove configure callbacks as well. I'll update the PR

eunwoosh (approved):

src/otx/core/model/anomaly.py
ashwinvaidya17: Yeah. I just tested it right now.

kprokofi (approved):

PR #3817, opened 2024-08-08
Move duplicate code in anomaly models to mixin

ashwinvaidya17:
### Summary

- Based off https://github.com/openvinotoolkit/training_extensions/pull/3811

### Checklist

### License

```python
```


src/otx/algo/anomaly/padim.py
```diff
@@ -7,23 +7,15 @@
...
 
-if TYPE_CHECKING:
-    from lightning.pytorch.utilities.types import STEP_OUTPUT
-    from torch.optim.optimizer import Optimizer
 
-    from otx.core.model.anomaly import AnomalyModelInputs
-
-
-class Padim(OTXAnomaly, OTXModel, AnomalibPadim):
+class Padim(AnomalyMixin, AnomalibPadim, OTXAnomaly):
```
kprokofi, line 16: What functionality do we inherit from AnomalibPadim? Can't we avoid inheritance from AnomalyMixin and AnomalibPadim? We could just call AnomalibPadim from anomalilib directly in OTXAnomaly methods and override any needed methods. Isn't it possible?

ashwinvaidya17: This helps deal with all the `super` calls. The problem is that AnomalibPadim itself is a lightning module with its own lightning hooks. Currently, there was duplicate code in OTX version of Stfpm and Padim which called `convert_inputs/outputs` and passed these to the base (anomalib's stfpm and padim) models. Since each model did this separately, I moved these to the mixin. OTXAnomaly needs to handle all the anomalib models. So, in the future we can add the rest of the models. Not sure if I am clear. Basically, because the mixin is before `AnomalibPadim` (or `AnomalibStfpm`), it will call `AnomalibPadim` when `super()` is called

ashwinvaidya17: Maybe this might help.
Here is one line of the training step from stfpm

https://github.com/openvinotoolkit/anomalib/blob/1a041d137d4d58cc9cdce563519c3f376357beee/src/anomalib/models/image/stfpm/lightning_model.py#L63

Here is the training step from padim

https://github.com/openvinotoolkit/anomalib/blob/1a041d137d4d58cc9cdce563519c3f376357beee/src/anomalib/models/image/padim/lightning_model.py#L77

So we can't have AnomalibPadim in OTXAnomaly

harimkang (commented):
LGTM, but integration test is fail. please revisit it.

```shell
ValueError: 'ANOMALY_CLASSIFICATION' is not a valid TaskType
```

harimkang (approved):

kprokofi:
Could you resolve merge conflicts?

kprokofi:
@chuneuny-emily , can we consider this PR for 2.2?

ashwinvaidya17:
I am closing this as the conflicts breaks this design. I'll revisit this later.

PR #3860, opened 2024-08-17
Refactoring detection task

sungchul1:
### Summary

This PR includes:
- [x] Create criterion modules for each model
- [x] Enable Factory template for building model
- [ ] Remove dict type config -> in the next PR due to the conflicts w/ iseg

### Checklist

### License

```python
```


sungchul1:
@kprokofi @harimkang We changed modules' names that are affected by factory class to `~Module` such as `SSDHeadModule` (previously `NNSSDHead`), but I found the below part that uses a module's name to internally patch a variable.
I'm not sure if similar parts are there more, but I think changing the module's name doesn't seem like an appropriate decision because we need to change all parts that are affected by our new decision.
I'd like to suggest setting factory's name such as `SSDHeadBuilder` instead.
https://github.com/openvinotoolkit/training_extensions/blob/71a819e42f77e1d1f76f59c0ea1cfc0734b58d2b/src/otx/core/model/detection.py#L468

eunwoosh:
> @kprokofi @harimkang We changed modules' names that are affected by factory class to `~Module` such as `SSDHeadModule` (previously `NNSSDHead`), but I found the below part that uses a module's name to internally patch a variable. I'm not sure if similar parts are there more, but I think changing the module's name doesn't seem like an appropriate decision because we need to change all parts that are affected by our new decision. I'd like to suggest setting factory's name such as `SSDHeadBuilder` instead.
> 
> https://github.com/openvinotoolkit/training_extensions/blob/71a819e42f77e1d1f76f59c0ea1cfc0734b58d2b/src/otx/core/model/detection.py#L468

Honestly, I think it's better to make factory function than factory class. Recently, factory class with `__new__` implementation is often used, but it looks weird that output of `SomeClass(...)` isn't instance of `SomeClass`. If class is preferred, than I think using class method like `SSDHeadBuilder.make` is better. Anyway it looks good to decide a policy of factory class for model at this point.

sungchul1:
> Honestly, I think it's better to make factory function than factory class. Recently, factory class with `__new__` implementation is often used, but it looks weird that output of `SomeClass(...)` isn't instance of `SomeClass`. If class is preferred, than I think using class method like `SSDHeadBuilder.make` is better. Anyway it looks good to decide a policy of factory class for model at this point.

Yes, I think it's a problem that `output of SomeClass(...) isn't instance of SomeClass` as @eunwoosh said.
Using factory function can be alternative, but I think it is important to maintain module's name no matter what method we choose.
If we use factory class, we can change factory's name like below:
```python
class SSDHead: ...

class SSDHeadBuilder:
    SSDHEAD_CFG: ClassVar[dict[str, Any]] = {...}
    def __new__(cls, version: str, ...):
        ...
        return SSDHead(**SSDHEAD_CFG[version], ...)
```
For factory function, we can use below:
```python
SSDHEAD_CFG: dict[str, Any] = {...}
def ssd_head_builder(version: str, ...):
    ...
    return SSDHead(**SSDHEAD_CFG[version], ...)
```

src/otx/algo/detection/backbones/backbone_factory.py
```diff
@@ -0,0 +1,67 @@
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
+"""Backbone factory for detection."""
+
+from torch import nn
+
+
+class DetectionBackboneFactory:
```
eunwoosh: I'm not sure it's right that this one class is in charge of making whole backbones for all detection models. Aren't each backbone compatible to other head? If then, I think making an ATSS backbone should be processed by ATSS part which means not ATSS model but file, class or function for ATSS. Anyway, this is my thought and if it was decided to implement like this, then I'll follow it.

src/otx/algo/detection/ssd.py
```diff
@@ -42,18 +43,30 @@
...
+    Default input size per model:
+        - ssd_mobilenetv2 : (864, 864)
+    """
+
+    mean: tuple[float, float, float] = (0.0, 0.0, 0.0)
+    std: tuple[float, float, float] = (255.0, 255.0, 255.0)
 
     def __init__(
         self,
+        model_version: str,
```
eunwoosh: Currently, `SSD` model supports only mobilenetv2. Should it get `model_version` at that case? How about leaving it as it is? If it look a little bit weird that `SSD` class always have mobilenet_v2 backbone, then we can consider changing `SSD` to `Mobilenetv2SSD`.

src/otx/algo/detection/heads/atss_head.py
```diff
@@ -259,182 +239,17 @@ def loss_by_feat(  # type: ignore[override]
...
-        return self.loss_centerness(pos_centerness, centerness_targets, avg_factor=avg_factor)
-
-    def _get_loss_bbox(
-        self,
-        pos_bbox_targets: Tensor,
-        pos_bbox_pred: Tensor,
-        centerness_targets: Tensor,
-    ) -> Tensor:
-        return self.loss_bbox(pos_bbox_pred, pos_bbox_targets, weight=centerness_targets, avg_factor=1.0)
+        return {
```
eunwoosh, line 242: It can be confusing that function name is `loss_by_feat` but it doesn't return loss. How about changing name?

src/otx/algo/detection/huggingface_model.py
```diff
@@ -72,6 +72,7 @@ def __init__(
         self.load_from = None
 
         super().__init__(
+            model_version=model_name_or_path,
```
eunwoosh: How about removing `self.model_name` from this class? It seems that class has two duplicated attirbutes.

src/otx/algo/detection/detectors/single_stage_detector.py
```diff
@@ -26,6 +26,7 @@ class SingleStageDetector(BaseModule):
     Args:
         backbone (nn.Module): Backbone module.
         bbox_head (nn.Module): Bbox head module.
+        criterion (nn.Module | None, optional): Criterion module.
```
eunwoosh:
```suggestion
        criterion (nn.Module): Criterion module.
```

src/otx/algo/common/utils/prior_generators/base_prior_generator.py
```diff
@@ -0,0 +1,37 @@
...
+from __future__ import annotations
+
+from abc import abstractmethod
+from typing import TYPE_CHECKING, Callable
+
+if TYPE_CHECKING:
+    from torch import Tensor
+
+
+class BasePriorGenerator:
```
eunwoosh:
```suggestion
class BasePriorGenerator(ABC):
```
How about making it as abstract class explictly?

src/otx/algo/common/utils/coders/base_bbox_coder.py
```diff
@@ -0,0 +1,26 @@
...
+# SPDX-License-Identifier: Apache-2.0
+#
+"""Base bounding box coder."""
+
+from abc import abstractmethod
+
+from torch import Tensor
+
+
+class BaseBBoxCoder:
```
eunwoosh:
```suggestion
class BaseBBoxCoder(ABC):
```
Same as above

eunwoosh (commented):
Thanks for your work :)
I agree overall direction of this PR, I have a one thing to discuss.
I'm not sure that we should change every module like a factory class even if that module supports single case. IMHO, it's better to leave it as it is and change it later if that module needs to support multiple cases.
How do you think about that?

src/otx/algo/common/utils/coders/base_bbox_coder.py
sungchul1: Done c3dfef0

src/otx/algo/common/utils/prior_generators/base_prior_generator.py
sungchul1: Done c3dfef0

src/otx/algo/detection/detectors/single_stage_detector.py
sungchul1: Done 4277eba

src/otx/algo/detection/huggingface_model.py
sungchul1: Done 4277eba

src/otx/algo/detection/ssd.py
sungchul1: If so, RTMDet or RTMDetInst that has single version should be changed as well.
And since model class inherits OTXModel, `model_version` must exist.

eunwoosh: Actually, I left a comment at SSD, but I want to ask general question. Of course, we can consider changing rtmdet also.
And about `OTXDetectionModel` (you mean this right?), we can change that argument as optional.

sungchul1: Honestly, I think general model's name (`SSD`, and `YOLOX`) must be used, not including specific backbone's name.
Also, `model_version` (changed to `model_name`) is also used as a required argument to distinguish sub models.
You can refer semantic segmentation part.

src/otx/algo/detection/backbones/backbone_factory.py
sungchul1: I first thought of using it only for ATSS and SSD, but it seemed better to use unified model loading structure.

src/otx/algo/detection/heads/atss_head.py
sungchul1: I think it'd better to compare it with `predict_by_feat` and `export_by_feat`, and combine redundant parts if there are.
But it can be a huge work, I'm thinking about splitting PRs or not.

src/otx/algo/detection/atss.py
```diff
@@ -33,11 +34,35 @@
     from otx.core.types.label import LabelInfoTypes
 
 
+AVAILABLE_MODEL_VERSIONS: list[str] = [
+    "atss_mobilenetv2",
+    "atss_resnext101",
```
harimkang: Do we need the word `atss_`? class already tells us that this is an ATSS.

src/otx/algo/detection/atss.py
```diff
@@ -33,11 +34,35 @@
...
+        - atss_mobilenetv2 : (800, 992)
+        - atss_resnext101 : (800, 992)
+    """
+
+    mean: tuple[float, float, float] = (0.0, 0.0, 0.0)
+    std: tuple[float, float, float] = (255.0, 255.0, 255.0)
 
     def __init__(
         self,
+        model_name: str,
```
harimkang: How about `Literal["atss_mobilenetv2", "atss_resnext101"]`?

src/otx/algo/detection/backbones/backbone_factory.py
harimkang: I had a similar impression with Eunwoo, isn't the existing way better than using multiple if statements without unification?
I don't think it's a good thing that we have to modify this backbone factory class every time we add a backbone.

To be honest, I don't know why we need this class.

src/otx/algo/detection/detectors/single_stage_detector.py
```diff
@@ -36,6 +37,7 @@ def __init__(
         self,
         backbone: nn.Module,
         bbox_head: nn.Module,
+        criterion: nn.Module,
```
harimkang, line 40: Is there any case where this is `None`?

src/otx/core/model/detection.py
```diff
@@ -461,11 +467,14 @@ def _forward_explain_detection(
...
-        from otx.algo.detection.heads import SSDHead
+        from otx.algo.detection.heads.ssd_head import SSDHeadModule
         from otx.algo.explain.explain_algo import DetClassProbabilityMap
 
         # SSD-like heads also have background class
-        background_class = hasattr(self.model, "bbox_head") and isinstance(self.model.bbox_head, SSDHead)
+        background_class = hasattr(self.model, "bbox_head") and isinstance(
+            self.model.bbox_head,
+            SSDHeadModule,
+        )  # TODO (sungchul): revert module's name?
```
harimkang, line 477: Are we sure `SSDHead` is the only one that fits the bill?

src/otx/core/model/detection.py
```diff
@@ -35,14 +35,18 @@
...
-    from otx.algo.detection.base_models import SingleStageDetector
+    from otx.algo.detection.detectors import SingleStageDetector
 
 
 class OTXDetectionModel(OTXModel[DetBatchDataEntity, DetBatchPredEntity]):
     """Base class for the detection models used in OTX."""
 
     input_size: tuple[int, int]
 
+    def __init__(self, model_name: str, *args, **kwargs) -> None:
```
harimkang, line 46: Is it right that every `OTXDetectionModel` should be given a `model_name`? I was under the impression that only models with multiple backbones or versions provide this, is that not the case? Also, I think the default should exist even if it is `None`.

src/otx/algo/detection/backbones/backbone_factory.py
eunwoosh: Honestly, it doesn't look good to add a code to make a backbone every time we need. I think it's useful if each backbone isn't depending on the head, so we can avoid duplicated code. But current implementation includes `if` statements for every head.

src/otx/algo/detection/atss.py
sungchul1: I think so, but if it is changed, others should be changed such as `yolox_tiny` -> `tiny` or `segnext_base` -> `base` like classification use case. Just like we used to use `variant` before.

src/otx/algo/detection/detectors/single_stage_detector.py
sungchul1: No, for detection.

src/otx/core/model/detection.py
sungchul1: I have no idea about xai, @GalyaZalesskaya could you give us some advice?

src/otx/core/model/detection.py
sungchul1: Yes, for those cases. But I don't think that some models have this and others don't have this is appropriate.
And I have no idea about setting the default (assume it is `atss_mobilenetv2`) because this default is not applied across all detection models.
More, I referred segmentation part below
https://github.com/openvinotoolkit/training_extensions/blob/db24a410e1a605db2915e41581238bf64a03d496/src/otx/core/model/segmentation.py#L41-L52

src/otx/algo/detection/backbones/backbone_factory.py
sungchul1: Removed this f38b613

sungchul1: Removed this f38b613

src/otx/algo/detection/atss.py
sungchul1: Done c3f5668

sungchul1:
> Thanks for your work :) I agree overall direction of this PR, I have a one thing to discuss. I'm not sure that we should change every module like a factory class even if that module supports single case. IMHO, it's better to leave it as it is and change it later if that module needs to support multiple cases. How do you think about that?

Yeah, it seems unnecessary but I think it must exist for scalability and unified structure.

src/otx/core/model/detection.py
harimkang: Okay, so we already have segmentation. Actually, this needs a little more discussion. I'm not sure if `model_name` is necessary for a common class. We'll talk more about it later.

harimkang (approved):

src/otx/core/model/detection.py
GalyaZalesskaya: Background class is needed and used only to explain SSD model among all other detection models

eunwoosh (approved):

src/otx/core/model/detection.py
eunwoosh: I agree with that.

PR #3863, opened 2024-08-19
Update Semantic Segmentation refactoring (reply comments from previous PR)

kprokofi:
### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algo/segmentation/backbones/litehrnet.py
+++ b/src/otx/algo/segmentation/backbones/litehrnet.py
@@ -1247,7 +1247,7 @@ class LiteHRNetBackbone:
 
     LITEHRNET_CFG: ClassVar[dict[str, Any]] = {
         "lite_hrnet_s": {
-            "stem": Stem(extra_stride=True),
+            "stem": {"extra_stride": True},
             "num_stages": 2,
             "stages_spec": {
                 "num_modules": [4, 4],
@@ -1261,7 +1261,6 @@ class LiteHRNetBackbone:
             "pretrained_weights": "https://storage.openvinotoolkit.org/repositories/openvino_training_extensions/models/custom_semantic_segmentation/litehrnetsv2_imagenet1k_rsc.pth",
         },
         "lite_hrnet_18": {
-            "stem": Stem(),
             "num_stages": 3,
             "stages_spec": {
                 "num_modules": [2, 4, 2],
@@ -1275,7 +1274,7 @@ class LiteHRNetBackbone:
             "pretrained_weights": "https://storage.openvinotoolkit.org/repositories/openvino_training_extensions/models/custom_semantic_segmentation/litehrnet18_imagenet1k_rsc.pth",
         },
         "lite_hrnet_x": {
-            "stem": Stem(stem_channels=60, out_channels=60, strides=(2, 1)),
+            "stem": {"stem_channels": 60, "out_channels": 60, "strides": (2, 1)},
             "num_stages": 4,
             "stages_spec": {
                 "weighting_module_version": "v1",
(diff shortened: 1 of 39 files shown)


src/otx/algo/segmentation/backbones/litehrnet.py
```diff
@@ -1435,14 +1247,7 @@ class LiteHRNetBackbone:
...
         "lite_hrnet_s": {
-            "stem": {
-                "stem_channels": 32,
-                "out_channels": 32,
-                "expand_ratio": 1,
-                "strides": [2, 2],
-                "extra_stride": True,
-                "input_norm": False,
-            },
+            "stem": Stem(extra_stride=True),
```
eunwoosh: I'm a little bit concerned that setting `Stem` as class variable. Because it's shared to all class instances, if one of instances changes value in the `Stem`, other insance which uses `Stem` should be affected. How about leaving it as it is for now and discuss about it to find better design?

src/otx/algo/segmentation/huggingface_model.py
```diff
@@ -52,7 +52,7 @@ class HuggingFaceModelForSegmentation(OTXSegmentationModel):
     Example:
         1. API
             >>> model = HuggingFaceModelForSegmentation(
-            ...     model_name_or_path="nvidia/segformer-b0-finetuned-ade-512-512",
+            ...     model_version="nvidia/segformer-b0-finetuned-ade-512-512",
```
harimkang: Actually, I think this is more of a `name` than a `version`, but what do you think?

kprokofi: Yes, but the other models and OTXSegmentationModel has "model_version" naming. I think we need to align this naming. We can change model_version to model_name everywhere though

harimkang: Hmm, but i think it's a little bit differen to have multiple versions of one architecture and to provide multiple architectures in one framework. In the case of the hugging-face example, different versions have different architectures, and those architectures can have multiple versions.

src/otx/algo/segmentation/backbones/litehrnet.py
kprokofi: I will move it to __new__ method.

src/otx/algo/segmentation/huggingface_model.py
kprokofi: I decided to go with "model_name" everywhere since "lite_hrnet_x" or "segnext_small" are more names, than versions. "model_version" seems to be confusing as parameter name in OTXSegmentationModel, but model_name looks good. Also, HuggingFace and DINOV2 look better with "model_name". 
So, I propose to keep this naming :)

src/otx/algo/segmentation/backbones/litehrnet.py
kprokofi: moved https://github.com/openvinotoolkit/training_extensions/pull/3863/commits/e7ea93c460a67587eaab5362b65686150531053a

src/otx/algo/segmentation/backbones/litehrnet.py
```diff
@@ -1500,10 +1290,11 @@ class LiteHRNetBackbone:
...
+    def __new__(cls, model_name: str) -> LiteHRNetModule:
         """Constructor for LiteHRNet backbone."""
-        if version not in cls.LITEHRNET_CFG:
-            msg = f"model type '{version}' is not supported"
+        if model_name not in cls.LITEHRNET_CFG:
+            msg = f"model type '{model_name}' is not supported"
             raise KeyError(msg)
-
-        return NNLiteHRNet(**cls.LITEHRNET_CFG[version])
+        stem_configuration = cls.LITEHRNET_CFG[model_name].pop("stem", {})
```
sungchul1: Just a suggestion, how about using a common name such as `config` or `default_config` across all tasks?

sungchul1: I left the same comment in https://github.com/openvinotoolkit/training_extensions/pull/3864#discussion_r1724377555

harimkang: I expect this can be resolved in one PR after discussion.

harimkang (approved):

GalyaZalesskaya (approved):

PR #3865, opened 2024-08-20
Instance Segmentation Model refactoring

kprokofi:
### Summary

- [x] Remove code duplication for torchvision MascRCNN
- [x] Add factory template for all models
- [x] Move loss out of the heads
- [x] Remove mmdict like approach (done partially due to overlap with detection task)
- [x] Update/add unit tests

### Checklist

### License

```python
```


sungchul1:
@kprokofi could you remove the duplicated one? https://github.com/openvinotoolkit/training_extensions/pull/3783#discussion_r1703456339
https://github.com/openvinotoolkit/training_extensions/blob/a5470fc3665228be8c3fde0bdc0728b11b26c057/src/otx/algo/instance_segmentation/layers/transformer.py#L88-L93

sungchul2:
@kprokofi @harimkang We changed modules' names that are affected by factory class to `~Module` such as `SSDHeadModule` (previously `NNSSDHead`), but I found the below part that uses a module's name to internally patch a variable.
I'm not sure if similar parts are there more, but I think changing the module's name doesn't seem like an appropriate decision because we need to change all parts that are affected by our new decision.
I'd like to suggest setting factory's name such as `SSDHeadBuilder` instead.
https://github.com/openvinotoolkit/training_extensions/blob/71a819e42f77e1d1f76f59c0ea1cfc0734b58d2b/src/otx/core/model/detection.py#L468

kprokofi:
This PR will be opened after https://github.com/openvinotoolkit/training_extensions/pull/3860 merged

src/otx/algo/instance_segmentation/maskrcnn.py
```diff
@@ -5,111 +5,56 @@
...
-        """Load the previous OTX ckpt according to OTX2.0."""
-        return OTXv1Helper.load_iseg_ckpt(state_dict, add_prefix)
-
-
-class MaskRCNNResNet50(MaskRCNN):
-    """MaskRCNN with ResNet50 backbone."""
+    AVAILABLE_MODEL_VERSIONS: ClassVar[list[str]] = [
+        "maskrcnn_resnet_50",
+        "maskrcnn_efficientnet_b2b",
+        "maskrcnn_swin_tiny",
```
eunwoosh, line 39: I think maskrcnn doesn't have to be added as prefix. If it's for using a full name to get other modules from the factory, then how about removing maskrcnn and adding it at that case? We may consider adding a function named `_get_fullname`.

src/otx/algo/detection/heads/anchor_head.py
```diff
@@ -55,31 +49,26 @@ def __init__(
...
             msg = f"num_classes={num_classes} is too small"
             raise ValueError(msg)
         self.reg_decoded_bbox = reg_decoded_bbox
 
         self.bbox_coder = bbox_coder
-        self.loss_cls = loss_cls
-        self.loss_bbox = loss_bbox
         self.train_cfg = train_cfg
         self.test_cfg = test_cfg
         if self.train_cfg:
```
eunwoosh, line 78: Is `if self.train_cfg:` necessary?

src/otx/algo/instance_segmentation/heads/rpn_head.py
```diff
@@ -26,43 +26,105 @@
...
+        self.nms_iou_threshold = nms_iou_threshold
+        self.nms_pre = nms_pre
+        self.with_nms = with_nms
+        self.min_bbox_size = min_bbox_size
+        self.max_per_img = max_per_img
+        self.pos_weight = pos_weight
+        self.reg_decoded_bbox = reg_decoded_bbox
+        self.allowed_border = allowed_border
+        self.assigner = assigner
+        self.sampler = sampler
```
eunwoosh: It seems that these two attributes are used in `AnchorHead`. If then, how about passing two attributes to `AnchorHead` and initialized them there?

src/otx/algo/instance_segmentation/heads/rpn_head.py
```diff
@@ -107,48 +169,58 @@ def forward_single(self, x: Tensor) -> tuple[Tensor, Tensor]:
...
-        proposal_cfg: dict | None = None,
-    ) -> tuple[dict, list[InstanceData]]:
+        entity: InstanceSegBatchDataEntity,  # type: ignore[override]
+    ) -> tuple:
         """Forward propagation of the head, then calculate loss and predictions from the features and data samples.
 
         Args:
             x (tuple[Tensor]): Features from FPN.
-            batch_data_samples (list[InstanceSegBatchDataEntity]): Each item contains
+            entity (InstanceSegBatchDataEntity): entity contains
```
eunwoosh:
```suggestion
            entity (InstanceSegBatchDataEntity): Entity contains
```

src/otx/algo/instance_segmentation/maskrcnn.py
```diff
@@ -5,111 +5,56 @@
...
-        )
+    effnet_std = (1.0, 1.0, 1.0)
 
     def _build_model(self, num_classes: int) -> TwoStageDetector:
-        train_cfg = {
+        if self.model_name not in self.AVAILABLE_MODEL_VERSIONS:
+            msg = f"Model version {self.model_name} is not supported."
+            raise ValueError(msg)
+
+        train_cfg = {  # TODO(Sungchul, Kirill): depricate it
```
eunwoosh: Could you remove assigner and assigner in rpn and rcnn?

src/otx/algo/instance_segmentation/losses/rpn_loss.py
```diff
@@ -0,0 +1,129 @@
...
+from torch import Tensor, nn
+
+from otx.algo.common.utils.utils import multi_apply
+
+
+class RPNCriterion(nn.Module):
+    """RPNCriterion is a loss criterion used in the Region Proposal Network (RPN) algorithm.
+
+    Args:
+        num_classes (int): The number of object classes.
```
eunwoosh: Argument docstring isn't matched with current arguments. Could you update it?

src/otx/algo/instance_segmentation/losses/rpn_loss.py
```diff
@@ -0,0 +1,129 @@
...
+        cls_reg_targets: tuple[list[Tensor], list[Tensor], list[Tensor], list[Tensor], int],
+        bbox_preds: list[Tensor],
+        cls_scores: list[Tensor],
+    ) -> dict:
+        """Calculate the loss based on the features extracted by the detection head.
+
+        TODO (kirill): it is related to RPNHead for iseg, will be deprecated
+
+        Args:
+            cls_scores (list[Tensor]): Box scores for each scale level
```
eunwoosh: Same as above

src/otx/algo/instance_segmentation/heads/rpn_head.py
```diff
@@ -184,42 +256,13 @@ def predict(
...
-        losses = super().loss_by_feat(
+        return self.predict_by_feat(
             cls_scores,
             bbox_preds,
-            batch_gt_instances,
-            batch_img_metas,
-            batch_gt_instances_ignore=batch_gt_instances_ignore,
+            batch_img_metas=batch_img_metas,
+            rescale=rescale,
+            cfg=self.test_cfg,
```
eunwoosh, line 242: It's just a question. Why wasn't `self.test_cfg` passed before?

src/otx/algo/instance_segmentation/heads/rpn_head.py
```diff
@@ -107,48 +169,58 @@ def forward_single(self, x: Tensor) -> tuple[Tensor, Tensor]:
...
-    def loss_and_predict(
+    def get_preds_and_targets(
         self,
         x: tuple[Tensor],
-        rpn_entity: InstanceSegBatchDataEntity,
-        proposal_cfg: dict | None = None,
-    ) -> tuple[dict, list[InstanceData]]:
+        entity: InstanceSegBatchDataEntity,  # type: ignore[override]
+    ) -> tuple:
         """Forward propagation of the head, then calculate loss and predictions from the features and data samples.
```
eunwoosh:
```suggestion
        """Forward propagation of the head, then calculate predictions from the features and data samples.
```

src/otx/algo/detection/heads/anchor_head.py
```diff
@@ -402,72 +391,15 @@ def get_targets(
...
+    def get_targets_for_loss(
         self,
         cls_scores: list[Tensor],
-        bbox_preds: list[Tensor],
         batch_gt_instances: list[InstanceData],
         batch_img_metas: list[dict],
         batch_gt_instances_ignore: list[InstanceData] | None = None,
-    ) -> dict:
+    ) -> tuple:
         """Calculate the loss based on the features extracted by the detection head.
```
eunwoosh: Could you update this docstring?

src/otx/algo/instance_segmentation/heads/rpn_head.py
```diff
@@ -439,23 +465,16 @@ def export_by_feat(
                 msg = "cls_score and bbox_pred should have the same size"
                 raise ValueError(msg)
             cls_score = cls_score.permute(0, 2, 3, 1)
-            if self.use_sigmoid_cls:
```
eunwoosh, line 442: It's just a question. Don't we have to consider a case not to use sigmoid now?

src/otx/algo/detection/heads/anchor_head.py
```diff
@@ -55,31 +49,26 @@ def __init__(
...
         train_cfg: dict,
-        loss_cls: nn.Module | None = None,  # TODO (kirill): deprecated
-        loss_bbox: nn.Module | None = None,  # TODO (kirill): deprecated
         test_cfg: dict | None = None,
         feat_channels: int = 256,
         reg_decoded_bbox: bool = False,
         init_cfg: dict | list[dict] | None = None,
     ) -> None:
+        if init_cfg is None:
+            init_cfg = {"type": "Normal", "layer": "Conv2d", "std": 0.01}
```
sungchul1: It seems only for RPNHead and if it is here other modules can be affected.

src/otx/algo/detection/heads/atss_head.py
```diff
@@ -182,7 +182,7 @@ def forward_single(self, x: Tensor, scale: Scale) -> tuple[Tensor, ...]:  # type
         centerness = self.atss_centerness(reg_feat)
         return cls_score, bbox_pred, centerness
 
-    def loss_by_feat(  # type: ignore[override]
+    def forward_for_loss(  # type: ignore[override]
```
sungchul1: It seems that `loss_by_feat` is changed to `forward_for_loss`.
I have 2 questions:
1. Why have `predict_by_feat` and `export_by_feat` not been changed?
2. Is `forward_for_loss` an appropriate name? Did you think any other candidate? Because `model.loss` -> `head.get_preds_and_targets` -> `head.forward_for_loss` forward flow seems less friendly than the previous one.

eunwoosh (commented):

src/otx/algo/instance_segmentation/backbones/maskrcnn_backbones.py
```diff
@@ -0,0 +1,56 @@
...
+from typing import Any, ClassVar
+
+from torch import nn
+
+from otx.algo.common.backbones import ResNet, build_model_including_pytorchcv
+from otx.algo.instance_segmentation.backbones.swin import SwinTransformer
+from otx.algo.modules.norm import build_norm_layer
+
+
+class MaskRCNNBackbone:
```
sungchul1: If these backbones are used for MaskRCNN, it can be located in `MaskRCNN` object like `ATSS` did.
What do you think?

src/otx/algo/instance_segmentation/heads/bbox_head.py
```diff
@@ -391,3 +374,306 @@ def export_by_feat(
...
+        if concat:
+            labels = torch.cat(labels, 0)
+            label_weights = torch.cat(label_weights, 0)
+            bbox_targets = torch.cat(bbox_targets, 0)
+            bbox_weights = torch.cat(bbox_weights, 0)
+            valid_label_mask = torch.cat(valid_label_mask, 0)
+        return labels, label_weights, bbox_targets, bbox_weights, valid_label_mask
+
+
+class ConvFCBBoxHead:
```
sungchul1, line 641: What do you think about using another name or builder function mentioned at https://github.com/openvinotoolkit/training_extensions/pull/3860#issuecomment-2306470778?

src/otx/algo/instance_segmentation/heads/roi_head.py
```diff
@@ -364,7 +385,6 @@ def export_bbox(
         x: tuple[Tensor],
         batch_img_metas: list[dict],
         rpn_results_list: tuple[Tensor, Tensor],
-        rcnn_test_cfg: dict,
         rescale: bool = False,
     ) -> tuple[Tensor, ...]:
         """Rewrite `predict_bbox` of `StandardRoIHead` for default backend.
```
sungchul1:
```suggestion
        """Rewrite `predict_bbox` of `RoIHead` for default backend.
```

src/otx/algo/instance_segmentation/losses/roi_loss.py
```diff
@@ -0,0 +1,128 @@
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) OpenMMLab. All rights reserved.
+#
+"""ATSS criterion."""
```
sungchul1: It needs to be updated.

sungchul1: src/otx/algo/instance_segmentation/losses/rpn_loss.py as well.

src/otx/algo/instance_segmentation/maskrcnn.py
sungchul1: I think we need to discuss about name policy for models.
About this, @harimkang also raised his hand and I left  https://github.com/openvinotoolkit/training_extensions/pull/3860#discussion_r1730693263.

src/otx/algo/instance_segmentation/maskrcnn.py
sungchul1: And I think we can remove L174-5 as well.

src/otx/algo/instance_segmentation/maskrcnn_tv.py
```diff
@@ -5,140 +5,107 @@
...
-        losses.update(proposal_losses)
+        model.load_state_dict(weights.get_state_dict(progress=True, check_hash=True))
 
-        if self.training:
-            return losses
-        scale_factors = [
-            img_meta.scale_factor if img_meta.scale_factor else (1.0, 1.0) for img_meta in entity.imgs_info
-        ]
+        # Replace RoIHeads since torchvision does not allow customized roi_heads.
+        model.roi_heads = TVRoIHeads(
```
sungchul1, line 75: `TVRoIHeads` seems to inherit `RoIHeads`, so both seem to have the same module structure. (It may not be true because I don't know well :)) 
Is there any way to patch `TVRoIHeads` to `RoIHeads` before instantiation?

sungchul1:
@kprokofi Could you check the performance before and after this PR?

src/otx/algo/detection/heads/anchor_head.py
kprokofi: No, It will be removed from all heads with next PR from me or Sungchul.

src/otx/algo/detection/heads/base_head.py
```diff
@@ -55,8 +55,12 @@ def get_positive_infos(self) -> list[InstanceData] | None:
             positive_infos.append(pos_info)
         return positive_infos
 
-    def loss(self, x: tuple[Tensor], entity: DetBatchDataEntity) -> dict:
-        """Perform forward propagation and loss calculation of the detection head.
+    def get_preds_and_targets(
```
eugene123tw: I believe this is similar to the question raised by @sungchul2. Could we combine `get_preds_and_targets` and `forward_to_loss` into a single function and rename it to `prepare_loss_inputs`? reflecting that the function is preparing or gathering the necessary components (targets, predictions, etc.) that will be used in further loss calculation. The current name, `get_preds_and_targets`, is misleading, as the function isn't just retrieving predictions and targets. Additionally, `forward_for_loss` might suggest that it's computing the loss, but it's actually just preparing the inputs for loss computation.

src/otx/algo/instance_segmentation/heads/bbox_head.py
```diff
@@ -9,103 +9,102 @@
...
-            out_dim_reg = box_dim if reg_class_agnostic else box_dim * num_classes
-            self.fc_reg = nn.Linear(in_features=in_channels, out_features=out_dim_reg)
+        in_channels = self.in_channels * self.roi_feat_area
+        # need to add background class
+        cls_channels = num_classes + 1
+        self.fc_cls = nn.Linear(in_features=in_channels, out_features=cls_channels)
+        box_dim = self.bbox_coder.encode_size
+        out_dim_reg = box_dim if reg_class_agnostic else box_dim * num_classes
+        self.fc_reg = nn.Linear(in_features=in_channels, out_features=out_dim_reg)
         self.debug_imgs = None
```
eugene123tw: Probably `self.debug_imgs` is not used and we can remove it.

src/otx/algo/instance_segmentation/heads/bbox_head.py
```diff
@@ -175,7 +167,6 @@ def predict_by_feat(
         cls_scores: tuple[Tensor],
```
eugene123tw, line 167: Are we planning to rename all `*_by_feat` methods, and change to `forward_for_*`?

src/otx/algo/instance_segmentation/heads/rtmdet_ins_head.py
```diff
@@ -530,14 +530,14 @@ def _mask_predict_by_feat_single(self, mask_feat: Tensor, kernels: Tensor, prior
                 x = torch.nn.functional.relu(x)
         return x.reshape(num_inst, h, w)
 
-    def loss_mask_by_feat(
+    def get_mask_targets(
```
eugene123tw: Could we rename to `prepare_mask_loss_inputs`?

src/otx/algo/instance_segmentation/losses/rpn_loss.py
```diff
@@ -0,0 +1,129 @@
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) OpenMMLab. All rights reserved.
+#
+"""ATSS criterion."""
```
eugene123tw:
```suggestion
"""RPN Loss Criterion."""
```

src/otx/algo/instance_segmentation/losses/rpn_loss.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def forward(
+        self,
+        cls_reg_targets: tuple[list[Tensor], list[Tensor], list[Tensor], list[Tensor], int],
+        bbox_preds: list[Tensor],
+        cls_scores: list[Tensor],
+    ) -> dict:
+        """Calculate the loss based on the features extracted by the detection head.
+
+        TODO (kirill): it is related to RPNHead for iseg, will be deprecated
```
eugene123tw: Should this be removed as well?

src/otx/algo/instance_segmentation/necks/fpn.py
```diff
@@ -162,3 +162,17 @@ def forward(self, inputs: tuple[Tensor]) -> tuple:
             for _ in range(self.num_outs - used_backbone_levels):
                 outs.append(torch.nn.functional.max_pool2d(outs[-1], 1, stride=2))
         return tuple(outs)
+
+
+class FPN:
```
eugene123tw: Could we remove this and instead use the FPN implementation in src/otx/algo/detection/necks/fpn.py to reduce duplication?

src/otx/algo/instance_segmentation/maskrcnn.py
```diff
@@ -178,512 +123,198 @@ def _build_model(self, num_classes: int) -> TwoStageDetector:
...
+            return {
+                "ignored_scope": {
+                    "types": ["Add", "Divide", "Multiply", "Sigmoid"],
+                    "validate": False,
+                },
+                "preset": "mixed",
+            }
+
+        if self.model_name == "maskrcnn_swin_t":
+            return {
```
eugene123tw: this is updated to: `return {"model_type": "transformer"}`

src/otx/algo/instance_segmentation/maskrcnn_tv.py
```diff
@@ -5,140 +5,107 @@
...
+        rpn_head = RPNHead(model_name=self.model_name, anchorgen=rpn_anchor_generator)
+        box_head = FastRCNNConvFCHead(model_name=self.model_name)
+        mask_head = MaskRCNNHeads(model_name=self.model_name)
 
-        detections, detector_losses = self.roi_heads(
-            features,
-            proposals,
-            image_list.image_sizes,
-            targets,
+        model = MaskRCNNTV(
```
eugene123tw: We should consider better naming conventions here, as we currently have `TV` used both as a prefix and a suffix in `TVMaskRCNN` and `MaskRCNNTV`, which makes it unclear what the differences are. Previously, the name `OTXTVMaskRCNN` was used (maybe?), likely to distinguish between OTX model module and the PyTorch TVMaskRCNN model module.

eugene123tw (commented):
Thanks, Kirill, for your efforts in refactoring Masks. I have a few questions, mainly regarding naming conventions. I believe some adjustments could help us better understand the purpose of the functions, methods, and classes.

src/otx/algo/instance_segmentation/maskrcnn_tv.py
eugene123tw: @sungchul2 RoIHeads comes from torchvision, while RoIHead is adapted from mmdetection, so they are inherently different packages. Aligning them in this PR would be challenging I think.

src/otx/algo/detection/heads/base_head.py
kprokofi: Let's combine them. I agree

src/otx/algo/instance_segmentation/maskrcnn.py
kprokofi: updated https://github.com/openvinotoolkit/training_extensions/pull/3865/commits/0f4409c90de69578b2e4b09e66f2b5b4c072e436

src/otx/algo/instance_segmentation/necks/fpn.py
kprokofi: Done https://github.com/openvinotoolkit/training_extensions/pull/3865/commits/0f4409c90de69578b2e4b09e66f2b5b4c072e436

src/otx/algo/instance_segmentation/losses/rpn_loss.py
kprokofi: Done https://github.com/openvinotoolkit/training_extensions/pull/3865/commits/0f4409c90de69578b2e4b09e66f2b5b4c072e436

src/otx/algo/instance_segmentation/losses/rpn_loss.py
kprokofi: Done https://github.com/openvinotoolkit/training_extensions/pull/3865/commits/0f4409c90de69578b2e4b09e66f2b5b4c072e436

src/otx/algo/instance_segmentation/heads/rtmdet_ins_head.py
kprokofi: Done https://github.com/openvinotoolkit/training_extensions/pull/3865/commits/0f4409c90de69578b2e4b09e66f2b5b4c072e436

src/otx/algo/instance_segmentation/heads/bbox_head.py
kprokofi: Done https://github.com/openvinotoolkit/training_extensions/pull/3865/commits/0f4409c90de69578b2e4b09e66f2b5b4c072e436

src/otx/algo/instance_segmentation/losses/roi_loss.py
kprokofi: Done https://github.com/openvinotoolkit/training_extensions/pull/3865/commits/0f4409c90de69578b2e4b09e66f2b5b4c072e436

src/otx/algo/detection/heads/atss_head.py
kprokofi: I will combine it with get_preds_and_targets as Eugene proposed

src/otx/algo/instance_segmentation/heads/rpn_head.py
kprokofi: I am not sure it it is required. We don't use softmax based solution for solving classification problem in our tasks. So, that's why I removed it

src/otx/algo/instance_segmentation/maskrcnn.py
kprokofi: Done https://github.com/openvinotoolkit/training_extensions/pull/3865/commits/0f4409c90de69578b2e4b09e66f2b5b4c072e436

src/otx/algo/instance_segmentation/heads/rpn_head.py
kprokofi: Done https://github.com/openvinotoolkit/training_extensions/pull/3865/commits/0f4409c90de69578b2e4b09e66f2b5b4c072e436

src/otx/algo/instance_segmentation/heads/rpn_head.py
kprokofi: I removed it for now, since they are initialized from train_cfg. I will remove this behavior in the next PR

src/otx/algo/instance_segmentation/heads/roi_head.py
kprokofi: Done https://github.com/openvinotoolkit/training_extensions/pull/3865/commits/c3bc2458fb51a22f86f59b7ffc7a09d2acf0389f

src/otx/algo/detection/heads/anchor_head.py
kprokofi: Done https://github.com/openvinotoolkit/training_extensions/pull/3865/commits/c3bc2458fb51a22f86f59b7ffc7a09d2acf0389f

src/otx/algo/instance_segmentation/heads/rpn_head.py
kprokofi: Done https://github.com/openvinotoolkit/training_extensions/pull/3865/commits/c3bc2458fb51a22f86f59b7ffc7a09d2acf0389f

src/otx/algo/instance_segmentation/losses/rpn_loss.py
kprokofi: Done https://github.com/openvinotoolkit/training_extensions/pull/3865/commits/c3bc2458fb51a22f86f59b7ffc7a09d2acf0389f

src/otx/algo/instance_segmentation/losses/rpn_loss.py
kprokofi: Done https://github.com/openvinotoolkit/training_extensions/pull/3865/commits/c3bc2458fb51a22f86f59b7ffc7a09d2acf0389f

src/otx/algo/instance_segmentation/maskrcnn.py
kprokofi: It is about the name alignment across our tasks. I can explain why I chose full name. For some tasks, we could face a problem where we share some modules and few models has small/medium/large versions. Also, for DINOV2 it is required to have full name to download model. Next, for HuggingFace wrapper we should also provide full name.

src/otx/algo/instance_segmentation/heads/rpn_head.py
kprokofi: It was later implicitly used in another method. So, basically nothing changed. Never mind, I will eliminate this behavior with the following PR.
image

src/otx/algo/instance_segmentation/maskrcnn.py
sungchul1: I think maintaining this policy seems better.

src/otx/algo/instance_segmentation/maskrcnn.py
sungchul1: Could you check if L174-5 is needed?

src/otx/algo/instance_segmentation/maskrcnn.py
eunwoosh: I understand. Actually, what I want to say is that it looks weird that we should pass "maskrcnn_resnet_50" as `model_name` argument to `MaskRCNN` class. So, I think it may be better that user passes "resnet_50" as model_name and `MaskRCNN` internally passes "maskrcnn_resnet_50" to common modules or other factory functions. Let's discuss it later as sungchul said.

src/otx/algo/instance_segmentation/maskrcnn_tv.py
sungchul1: Because both `TVRoIHeads` and `torchvision.RoIHeads` (set in `torchvision.FasterRCNN`) share `__init__` and have the same arguments, I roughly thought it could be seamlessly updated like this:
```python
from torchvision.models.detection.roi_heads import RoIHeads

RoIHeads = TVRoIHeads
model = MaskRCNNTV(...)
```
But I agree with that this agenda is not a scope of this PR as you said, so let's keep this.

sungchul1:
@kprokofi could you update all of the docstrings? It looks outdated or mismatched.
And for rtmdet_inst, need to check in some things it is written as `inst` (`rtmdet_inst`), and in some things it is written as `ins` (`rtmdet_ins_head`).

src/otx/algo/detection/heads/base_head.py
sungchul1: Additionally, I think it would be better to combine common parts across forwards (loss, predict, export) if there is.
https://github.com/openvinotoolkit/training_extensions/pull/3860#discussion_r1728972851

src/otx/algo/instance_segmentation/heads/bbox_head.py
kprokofi: I have a plan to eliminate *_by_feat,  but this PR becomes too large with such change. Let's consider it as the next task

src/otx/algo/detection/heads/base_head.py
kprokofi: Done. https://github.com/openvinotoolkit/training_extensions/pull/3865/commits/2e392af37a8f4f677dcfce5665844f6c8d96ac25
Predict and export can be combined in next PR to not extend this one more

src/otx/algo/instance_segmentation/heads/bbox_head.py
kprokofi: I changed it to BBoxHead 
https://github.com/openvinotoolkit/training_extensions/pull/3865/commits/2e392af37a8f4f677dcfce5665844f6c8d96ac25

src/otx/algo/instance_segmentation/backbones/maskrcnn_backbones.py
kprokofi: Done https://github.com/openvinotoolkit/training_extensions/pull/3865/commits/2e392af37a8f4f677dcfce5665844f6c8d96ac25

src/otx/algo/detection/heads/atss_head.py
kprokofi: Done https://github.com/openvinotoolkit/training_extensions/pull/3865/commits/2e392af37a8f4f677dcfce5665844f6c8d96ac25

src/otx/algo/detection/heads/anchor_head.py
kprokofi: Done https://github.com/openvinotoolkit/training_extensions/pull/3865/commits/2e392af37a8f4f677dcfce5665844f6c8d96ac25

kprokofi:
> @kprokofi could you update all of the docstrings? It looks outdated or mismatched. And for rtmdet_inst, need to check in some things it is written as `inst` (`rtmdet_inst`), and in some things it is written as `ins` (`rtmdet_ins_head`).

I aligned names and fixed found problems with docstrings

src/otx/algo/instance_segmentation/maskrcnn_tv.py
kprokofi: I renamed it to MaskRCNNTV and named model from torchvision (in src/otx/algo/instance_segmentation/maskrcnn_tv.py) as simply MaskRCNN

kprokofi:
> @kprokofi Could you check the performance before and after this PR?

pref tests are running

kprokofi:
Strange unit test fail, Point entity for VPT task has no attribute "canvas_size". I can't reproduce it directly with pytest, only via "tox" and running all tests...

sungchul1:
@kprokofi  I didn't reproduce unit test issue on both pytest and tox in local, need to check.

kprokofi:
> @kprokofi I didn't reproduce unit test issue on both pytest and tox in local, need to check.

I found the problem. I added accidently from src.otx... import ImageInfo, instead of otx.... import ImageInfo. That leaded to some import errors maybe and misled about the underlying problem.

src/otx/algo/detection/heads/atss_head.py
```diff
@@ -182,36 +187,24 @@ def forward_single(self, x: Tensor, scale: Scale) -> tuple[Tensor, ...]:  # type
...
-                Batch of gt_instances_ignore. It includes ``bboxes`` attribute
-                data that is ignored during training and testing.
-                Defaults to None.
+            x (tuple[Tensor]): Features from the upstream network, each is
+                a 4D-tensor.
+            entity (DetBatchDataEntity): Entity from OTX dataset.
 
         Returns:
-            dict[str, Tensor]: A dictionary of raw outputs.
+            dict: A dictionary of loss components.
```
sungchul1: The output doesn't contain any loss components, why is it changed?

src/otx/algo/detection/heads/rtmdet_head.py
```diff
@@ -157,35 +158,26 @@ def forward(self, feats: tuple[Tensor, ...]) -> tuple:
...
-                Batch of gt_instances_ignore. It includes ``bboxes`` attribute
-                data that is ignored during training and testing.
-                Defaults to None.
+            x (tuple[Tensor]): Features from the upstream network, each is
+                a 4D-tensor.
+            entity (DetBatchDataEntity): Entity from OTX dataset.
 
         Returns:
-            dict[str, Tensor]: A dictionary of raw outputs.
+            dict: A dictionary of loss components.
```
sungchul1: Same with the above

src/otx/algo/detection/heads/atss_head.py
sungchul1: In addition to this, the docstring in rtmdet_head, atss_head, and yolox_head were updated.

src/otx/algo/instance_segmentation/heads/bbox_head.py
```diff
@@ -391,3 +374,306 @@ def export_by_feat(
...
+        if concat:
+            labels = torch.cat(labels, 0)
+            label_weights = torch.cat(label_weights, 0)
+            bbox_targets = torch.cat(bbox_targets, 0)
+            bbox_weights = torch.cat(bbox_weights, 0)
+            valid_label_mask = torch.cat(valid_label_mask, 0)
+        return labels, label_weights, bbox_targets, bbox_weights, valid_label_mask
+
+
+class BBoxHead:
```
sungchul1: If this factory only returns `ConvFCBBoxHeadModule`, how about changing this name to `ConvFCBBoxHead` instead?

src/otx/algo/detection/heads/atss_head.py
kprokofi: "prepare_loss_inputs" returns components for the loss computation.

src/otx/algo/instance_segmentation/heads/bbox_head.py
kprokofi: You left previously comment about "ConvFCBBoxHead" unappropriated naming and I changed it to simple one
image

kprokofi:
Perf tests:

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/kprokofi/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/kprokofi/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#467886" vlink="#96607D">


data_group |   | all/develop | all/develop | all/refactoring | all/refactoring
-- | -- | -- | -- | -- | --
metric |   | test/f1-score | test/f1-score | test/f1-score | test/f1-score
stat |   | mean | std | mean | std
task | model |   |   |   |  
instance_segmentation | all | 0.51328727 | 0.13603715 | 0.506485189 | 0.147367714
instance_segmentation | maskrcnn_efficientnetb2b | 0.530471097 | 0.064085943 | 0.537792064 | 0.072942254
instance_segmentation | maskrcnn_r50 | 0.489827528 | 0.099895469 | 0.4898914 | 0.104490717
instance_segmentation | maskrcnn_r50_tv | 0.54778025 | 0.133141648 | 0.554633201 | 0.138719154
instance_segmentation | maskrcnn_swint | 0.589406501 | 0.100718406 | 0.547774035 | 0.137834792
instance_segmentation | rtmdet_inst_tiny | 0.408950974 | 0.194595874 | 0.402335247 | 0.212427851



</body>

</html>

I need to look at SwinT if the drop is random or not.

src/otx/algo/instance_segmentation/heads/bbox_head.py
sungchul1: Ah, that comment was the suggestion about using factory names to ~builder instead of module's name as is.
Because some parts need module's name to check some conditions such as xai.

src/otx/algo/detection/heads/atss_head.py
sungchul1: Ah, I thought that loss components are loss values.

src/otx/algo/instance_segmentation/heads/bbox_head.py
kprokofi: Renamed back https://github.com/openvinotoolkit/training_extensions/pull/3865/commits/3a68981ed9d3cbc18f1d6c9b84cf6de1b9edff5b

kprokofi: Renaming all factories with builder suffix  should be considered for all tasks. We may consider that

src/otx/algo/detection/heads/rtmdet_head.py
kprokofi: made it clearer: https://github.com/openvinotoolkit/training_extensions/pull/3865/commits/3a68981ed9d3cbc18f1d6c9b84cf6de1b9edff5b

src/otx/algo/detection/heads/atss_head.py
kprokofi: made it clearer: https://github.com/openvinotoolkit/training_extensions/commit/3a68981ed9d3cbc18f1d6c9b84cf6de1b9edff5b

kprokofi:
I tested one more time and got to conclusion, that some fluctuation is normal.
Develop (5 runs):  
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/kprokofi/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/kprokofi/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#467886" vlink="#96607D">


  | small | small
-- | -- | --
  | test/f1-score | test/f1-score
  | mean | std
model |   |  
maskrcnn_swint | 0.489417503 | 0.013157227



</body>

</html>

This branch (5 runs):
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/kprokofi/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/kprokofi/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#467886" vlink="#96607D">


  | small | small
-- | -- | --
  | test/f1-score | test/f1-score
  | mean | std
model |   |  
maskrcnn_swint | 0.507344836 | 0.079074585



</body>

</html>

I remeasured develop one more time (5 runs, averaged):
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/kprokofi/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/kprokofi/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#467886" vlink="#96607D">


  | small | small
-- | -- | --
  | test/f1-score | test/f1-score
  | mean | std
model |   |  
maskrcnn_swint | 0.500201058 | 0.032963669



</body>

</html>

So, even setting seeds doesn't guarantee the full reproducibility.  Next, even with results above, 0.589 +- 0.1 ~= 0.548 +- 0.137

I also compared the code once again and checked that snapshot from develop produces the same result on that branch. So, I hope, there is no difference in terms of performance.

I propose to merge this PR  before https://github.com/openvinotoolkit/training_extensions/pull/3917

eunwoosh (approved):

sungchul1 (approved):
Thanks for your hard work!
I have a minor question, should `RTMDetInst` inherit `ExplainableOTXInstanceSegModel` even if explainable is not supported?

kprokofi:
> Thanks for your hard work! I have a minor question, should `RTMDetInst` inherit `ExplainableOTXInstanceSegModel` even if explainable is not supported?

When I implemented RTDetr, in ExplainableOTXDetModel was some required functionality. I guess, RTMDet has the same problem. We should refactor this in the future and maybe merge Explainable*Model with base OTX*Model to avoid any discrepancy.

PR #3904, opened 2024-08-28
Fix a bug that training is stuck while detection model is trained on distrubited environment

### LINKED ISSUE #3635: Problems in multi-card distributed training

eunwoosh:
### Summary
This PR fixes #3635

### How this change solves the bug
In distributed training, torchmetric package tries to sync values across processes.
If one process has a value in bbox, label or scores and other process not, then torchmetric makes empty torch Tensor.
But if dtype of tensor between processes is different, then process is stuck during sync.
To avoid it, this PR specifies dtype properly.

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/core/data/dataset/detection.py
+++ b/src/otx/core/data/dataset/detection.py
@@ -51,7 +51,7 @@ class OTXDetectionDataset(OTXDataset[DetDataEntity]):
                 canvas_size=img_shape,
                 dtype=torch.float32,
             ),
-            labels=torch.as_tensor([ann.label for ann in bbox_anns], dtype=torch.int32),
+            labels=torch.as_tensor([ann.label for ann in bbox_anns], dtype=torch.long),
         )
 
         return self._apply_transforms(entity)
--- a/src/otx/core/data/dataset/detection.py
+++ b/src/otx/core/data/dataset/detection.py
@@ -49,8 +49,9 @@ class OTXDetectionDataset(OTXDataset[DetDataEntity]):
                 bboxes,
                 format=tv_tensors.BoundingBoxFormat.XYXY,
                 canvas_size=img_shape,
+                dtype=torch.float32,
             ),
-            labels=torch.as_tensor([ann.label for ann in bbox_anns]),
+            labels=torch.as_tensor([ann.label for ann in bbox_anns], dtype=torch.int32),
         )
 
         return self._apply_transforms(entity)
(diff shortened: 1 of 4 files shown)


src/otx/core/data/dataset/detection.py
```diff
@@ -49,8 +49,9 @@ def _get_item_impl(self, index: int) -> DetDataEntity | None:
                 bboxes,
                 format=tv_tensors.BoundingBoxFormat.XYXY,
                 canvas_size=img_shape,
+                dtype=torch.float32,
             ),
-            labels=torch.as_tensor([ann.label for ann in bbox_anns]),
+            labels=torch.as_tensor([ann.label for ann in bbox_anns], dtype=torch.int32),
```
sungchul1: Could you align it with iseg to int64 if it works?

eunwoosh: I think it should work but is it needed to set it to torch.int64? I think there is no case to use more labels than upper limit of int32.

eunwoosh: It seems that some code in detection head is using `torch.long` for label calculation and need to align it with that. maybe this is why iseg uses torch.int64. Anyway, I updated it to `torch.long`. Thanks.

sungchul1: I think so, otx seems to use double and just aligning it :)

sungchul1 (approved):

harimkang (approved):

PR #3871, opened 2024-08-21
Fix e2e code error

chuneuny-emily:
### Checklist

- [ ] I have added unit tests to cover my changes.​
- [ ] I have added integration tests to cover my changes.​
- [ ] I have ran e2e tests and there is no issues.
- [ ] I have added the description of my changes into CHANGELOG in my target branch (e.g., [CHANGELOG](https://github.com/openvinotoolkit/training_extensions/blob/develop/CHANGELOG.md) in develop).​
- [ ] I have updated the documentation in my target branch accordingly (e.g., [documentation](https://github.com/openvinotoolkit/training_extensions/tree/develop/docs) in develop).
- [ ] I have [linked related issues](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

### License

- [ ] I submit _my code changes_ under the same [Apache License](https://github.com/openvinotoolkit/training_extensions/blob/develop/LICENSE) that covers the project.
      Feel free to contact the maintainers if that's a concern.
- [ ] I have updated the license header for each file (see an example below).

```python
# Copyright (C) 2024 Intel Corporation
# SPDX-License-Identifier: Apache-2.0
```

PR #3917, opened 2024-08-30
Add YOLOv9

sungchul1:
### Summary

This PR includes:
- [x] Integrate YOLOv9 into OTX
- [x] Refactoring
- [x] Add related tests
- [x] Update others

The next tasks are:
- Integrate YOLOv7 into OTX
- Optimize performance and update recipes
- check e2e after merging back 2.2 due to https://github.com/openvinotoolkit/training_extensions/pull/3871

### Checklist

### License

```python
```


src/otx/algo/common/losses/iou_loss.py
```diff
@@ -76,6 +78,94 @@ def forward(
...
+        loss_weight (float): Weight of loss.
+        mode (str): Loss scaling mode, including "linear", "square", and "log".
+            Default: 'log'
+    """
+
+    def __init__(
+        self,
+        linear: bool = False,
+        eps: float = 1e-6,
+        reduction: str = "mean",
```
eunwoosh:
```suggestion
        reduction: Literal["none", "mean", "sum"] = "mean",
```
How about using literal as above?

src/otx/algo/common/losses/iou_loss.py
```diff
@@ -76,6 +78,94 @@ def forward(
...
+        loss_weight: float = 1.0,
+        mode: str = "log",
```
eunwoosh:
```suggestion
        mode: Literal["linear", "square", "log"] = "log",
```
It's same as above.

src/otx/algo/common/losses/iou_loss.py
```diff
@@ -104,3 +194,53 @@ def giou_loss(pred: Tensor, target: Tensor, eps: float = 1e-7) -> Tensor:
         gious = gious.to(torch.float16)
 
     return 1 - gious
+
+
+@weighted_loss
+def iou_loss(pred: Tensor, target: Tensor, linear: bool = False, mode: str = "log", eps: float = 1e-6) -> Tensor:
```
eunwoosh:
```suggestion
def iou_loss(pred: Tensor, target: Tensor, linear: bool = False, mode: Literal["linear", "square", "log"] = "log", eps: float = 1e-6) -> Tensor:
```
Same as above.

src/otx/algo/detection/yolov9.py
```diff
@@ -0,0 +1,196 @@
...
+        - yolov9_e : (640, 640) # not supported yet
+    """
+
+    input_size_multiplier = 32  # TODO (sungchul): need to check
+    mean: tuple[float, float, float] = (0.0, 0.0, 0.0)
+    std: tuple[float, float, float] = (255.0, 255.0, 255.0)
+
+    def __init__(
+        self,
+        model_name: Literal["yolov9_t", "yolov9_s", "yolov9_m", "yolov9_c", "yolov9_e"],
```
eunwoosh:
```suggestion
        model_name: Literal["yolov9_s", "yolov9_m", "yolov9_c"],
```
How about removing them for now and updating when they'll become available.

src/otx/algo/detection/backbones/gelan.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+class GELANModule(nn.Module):
+    """Generalized Efficient Layer Aggregation Network (GELAN) implementation for YOLOv9.
+
+    Args:
+        first_dim (int): The number of input channels.
+        block_entry_cfg (dict[str, Any]): The configuration for the entry block.
+        csp_channels (list[list[int]]): The configuration for the CSP blocks.
+        csp_args (dict[str, Any], optional): The configuration for the CSP blocks. Defaults to None.
```
eunwoosh:
```suggestion
        csp_args (dict[str, Any] | None, optional): The configuration for the CSP blocks. Defaults to None.
```

src/otx/algo/detection/utils/yolov7_v9_utils.py
```diff
@@ -0,0 +1,257 @@
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) OpenMMLab. All rights reserved.
```
eunwoosh: Is this code from mmlab?

src/otx/algo/detection/necks/yolo_neck.py
```diff
@@ -0,0 +1,115 @@
...
+from otx.algo.detection.layers import SPPELAN, Concat, RepNCSPELAN
+from otx.algo.detection.utils.yolov7_v9_utils import set_info_into_instance
+
+
+class YOLONeckModule(nn.Module):
+    """Neck module for YOLOv7 and v9.
+
+    Args:
+        elan_channels (list[dict[str, int]]): The ELAN channels.
+        concat_sources (list[list[str, int]]): The sources to concatenate.
```
eunwoosh:
```suggestion
        concat_sources (list[list[str | int]]): The sources to concatenate.
```

src/otx/algo/detection/layers/elan_layer.py
```diff
@@ -0,0 +1,360 @@
...
+from torch import Tensor, nn
+
+from otx.algo.detection.utils.yolov7_v9_utils import auto_pad
+from otx.algo.modules import Conv2dModule, build_activation_layer
+
+logger = logging.getLogger(__name__)
+
+
+class ELAN(nn.Module):
+    """ELAN  structure.
```
eunwoosh:
```suggestion
    """ELAN structure.
```

src/otx/algo/detection/losses/yolov9_loss.py
```diff
@@ -0,0 +1,451 @@
...
+    # Compute CIoU
+    ciou = diou - alpha * v
+    return ciou.to(dtype)
+
+
+class BCELoss(nn.Module):
+    """Binary Cross Entropy Loss.
+
+    TODO (author): Refactor the device, should be assign by config
+    TODO (author): origin v9 assign pos_weight == 1?
```
eunwoosh, line 96: It's just a question. What is it about?

src/otx/core/model/base.py
```diff
@@ -164,7 +164,7 @@ def training_step(self, batch: T_OTXBatchDataEntity, batch_idx: int) -> Tensor:
                     prog_bar=True,
                 )
 
-            total_train_loss = sum(train_loss.values())
+            total_train_loss = train_loss.get("loss", sum(train_loss.values()))
```
eunwoosh: So, it's an introduction of new way to set final loss, right? Doesn't it affect other models? because "loss" looks too general word, I think some models which have set "loss" in dict can be affected.

src/otx/algo/detection/yolov9.py
```diff
@@ -0,0 +1,196 @@
...
+                    "labels": {0: "batch", 1: "num_dets"},
+                },
+                "keep_initializers_as_inputs": False,
+                "verbose": False,
+                "autograd_inlining": False,
+            },
+            output_names=None,  # TODO (someone): support XAI
+        )
+
+    def to(self, *args, **kwargs) -> Self:
```
eunwoosh, line 138: It's just a question. What is it for?

src/otx/algo/detection/yolov9.py
```diff
@@ -0,0 +1,196 @@
...
+            tile_config=tile_config,
+        )
+
+    def _build_model(self, num_classes: int) -> SingleStageDetector:
+        backbone = GELAN(model_name=self.model_name)
+        neck = YOLONeck(model_name=self.model_name)
+        bbox_head = YOLOHead(model_name=self.model_name, num_classes=num_classes)
+
+        # patch _load_from_state_dict
+        SingleStageDetector._load_from_state_dict = _load_from_state_dict_for_yolov9  # type: ignore[method-assign] # noqa: SLF001
```
eunwoosh: Instead of changing class, how about changing instance?

eunwoosh (commented):

src/otx/algo/detection/losses/yolov9_loss.py
sungchul1: It is from the reference code, I left it to follow up periodically.

src/otx/core/model/base.py
sungchul1: I'll check it and leave here.

src/otx/algo/detection/yolov9.py
sungchul1: Some modules related to `Vec2Box` are not converted to the target device.
I'm not sure, but I'm guessing that it is because `Vec2Box` must be created after model instanstiation because auto anchor generation in the module is based on model forward's output.
Here is to manually convert the modules.

src/otx/algo/detection/yolov9.py
sungchul1: If using `detector._load_from_state_dict = _load_from_state_dict_for_yolov9` instead, the instance's methods are not reachable in `_load_from_state_dict_for_yolov9` due to `self` is not given like below
https://github.com/openvinotoolkit/training_extensions/blob/e38a7aa453326274ebf25ab5fd7017e65de985e6/src/otx/algo/utils/mmengine_utils.py#L146-L154

src/otx/algo/detection/layers/elan_layer.py
sungchul1: Done at 73b7672

src/otx/algo/detection/necks/yolo_neck.py
sungchul1: Done at 73b7672

src/otx/algo/detection/utils/yolov7_v9_utils.py
sungchul1: Thanks for finding, removed at 73b7672

src/otx/algo/detection/backbones/gelan.py
sungchul1: Done at 73b7672

src/otx/algo/detection/yolov9.py
sungchul1: Done at 73b7672

src/otx/algo/common/losses/iou_loss.py
sungchul1: Done at 73b7672

src/otx/algo/common/losses/iou_loss.py
sungchul1: Done at 73b7672

src/otx/algo/common/losses/iou_loss.py
sungchul1: Done at 73b7672

src/otx/algo/detection/yolov9.py
eunwoosh: It can be changed as below
```
detector._load_from_state_dict = _load_from_state_dict_for_yolov9.__get__(detector, SingleStageDetector)
```
Please refer https://stackoverflow.com/questions/50599045/python-replacing-a-function-within-a-class-of-a-module

sungchul1: Thanks, it works, done at 2bf9c3b

src/otx/core/model/base.py
sungchul1: I tried to find what will be affected by this change and I found visual prompting task only.
But vpm uses `loss` as total loss that is computed with specific way (`20 * focal loss + dice loss + iou loss`), not just using summation.

But I think this change looks more general because `sum(train_loss.values())` doesn't consider whether `train_loss` has `loss` key or not, assuming `loss` is used as total loss.
For example, it seems that vpm loss has been calculated incorrectly due to this part.
Since https://github.com/openvinotoolkit/training_extensions/pull/3155, vpm total loss has been computed twice.
For example, if focal loss=1, dice loss=1, and iou loss=1, total loss should be 20 * focal loss + dice loss + iou loss = 22.
But current total loss is computed once more in that part, so it becomes 22 + 1 + 1 + 1 = 45.
Of course, it can derive fast convergence with twice higher loss and gradient, but it can also derive unstable training and training fluctuation.
Due to this change, it will return to normal.

eunwoosh: so it means that the only model affected is VPM and it has miscalculated loss and it can be solved by current change, right?

src/otx/algo/detection/utils/yolov7_v9_utils.py
```diff
@@ -0,0 +1,256 @@
...
+        preds_anc = torch.concat(preds_anc, dim=1)
+        preds_box = torch.concat(preds_box, dim=1)
+
+        pred_lt_rb = preds_box * self.scaler.view(1, -1, 1)
+        lt, rb = pred_lt_rb.chunk(2, dim=-1)
+        preds_box = torch.cat([self.anchor_grid - lt, self.anchor_grid + rb], dim=-1)
+        return preds_cls, preds_anc, preds_box
+
+
+def bbox_nms(
```
kprokofi: Can't we use already introduced for other models batch_nms or torch_nms?

src/otx/algo/detection/utils/yolov7_v9_utils.py
```diff
@@ -0,0 +1,256 @@
...
+        w = torch.arange(0, width, stride) + shift
+        anchor_h, anchor_w = torch.meshgrid(h, w, indexing="ij")
+        anchor = torch.stack([anchor_w.flatten(), anchor_h.flatten()], dim=-1)
+        anchors.append(anchor)
+    all_anchors = torch.cat(anchors, dim=0)
+    all_scalers = torch.cat(scaler, dim=0)
+    return all_anchors, all_scalers
+
+
+def transform_bbox(bbox: Tensor, indicator: str = "xywh -> xyxy") -> Tensor:
```
kprokofi: Can't we use torchvision functionality to convert bboxes as we did for RTDetr, for example?

src/otx/algo/detection/utils/yolov7_v9_utils.py
```diff
@@ -0,0 +1,256 @@
...
+        x (int | Tensor): The input value.
+        div (int): The divisor value. Defaults to 1.
+
+    Returns:
+        int | Tensor: The rounded up value.
+    """
+    return x + (-x % div)
+
+
+def generate_anchors(image_size: tuple[int, int], strides: list[int]) -> tuple[Tensor, Tensor]:
```
kprokofi: We have AnchorGenerator, isn't it possible to use its functionality to generate anchors? Or maybe we could change AnchorGenerator to align with YOLO interface

src/otx/algo/detection/utils/yolov7_v9_utils.py
```diff
@@ -0,0 +1,256 @@
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
+"""Utils for YOLOv7 and v9.
```
kprokofi: Can't we merge this file with general detection utils? The naming overall looks a bit strange and I think we can combine these bbox related utils with other "utils"

src/otx/algo/modules/activation.py
```diff
@@ -57,7 +57,6 @@ def forward(self, x: Tensor) -> Tensor:
     nn.Sigmoid,
     Swish,
     nn.GELU,
-    nn.SiLU,
```
kprokofi, line 60: we use nn.SiLU for RTMDet, could you check if it is OK to delete it? I think we should keep SiLU

src/otx/core/model/base.py
kprokofi: It means that we cannot use just "loss" in our models, since only this "loss" can be optimized at the end. Can't we avoid this behavior or at least rename it like "summary_loss", "sum_loss", etc.

src/otx/algo/detection/yolov9.py
```diff
@@ -0,0 +1,192 @@
...
+if TYPE_CHECKING:
+    from lightning.pytorch.cli import LRSchedulerCallable, OptimizerCallable
+    from typing_extensions import Self
+
+    from otx.core.metrics import MetricCallable
+    from otx.core.schedulers import LRSchedulerListCallable
+    from otx.core.types.label import LabelInfoTypes
+
+
+PRETRAINED_WEIGHTS: dict[str, str] = {
```
kprokofi, line 32: We have difference here across our tasks. I personally, add this as a class attribute, but you add it as a global parameter. We need to have common decision

src/otx/algo/detection/yolov9.py
```diff
@@ -0,0 +1,192 @@
...
+    "yolov9_s": "https://github.com/WongKinYiu/YOLO/releases/download/v1.0-alpha/v9-s.pt",
+    "yolov9_m": "https://github.com/WongKinYiu/YOLO/releases/download/v1.0-alpha/v9-m.pt",
+    "yolov9_c": "https://github.com/WongKinYiu/YOLO/releases/download/v1.0-alpha/v9-c.pt",
+}
+
+
+def _load_from_state_dict_for_yolov9(
```
kprokofi: I am not sure whether it is proper place for that... 
Maybe it is better to create separate class like YOLOSingleStageDetector, inherit from SingleStageDetector and reload _load_from_state_dict or add it somehow to the SingleStageDetector.  What do you think?

src/otx/algo/detection/heads/yolo_head.py
```diff
@@ -0,0 +1,729 @@
...
+
+    def forward(self, x: Tensor) -> tuple[Tensor, Tensor, Tensor]:
+        """Forward function."""
+        anchor_x = self.anchor_conv(x)
+        class_x = self.class_conv(x)
+        anchor_x, vector_x = self.anc2vec(anchor_x)
+        return class_x, anchor_x, vector_x
+
+
+class ISingleHeadDetection(nn.Module):
```
kprokofi: The name of this class is a bit strange for me. If it is used for V7, maybe we could rename these two classes as SingleHeadDetectionV9 and SingleHeadDetectionV7...

src/otx/algo/detection/heads/yolo_head.py
```diff
@@ -0,0 +1,729 @@
...
+            x = layer(model_input)
+            outputs[-1] = x
+            if hasattr(layer, "tags"):
+                outputs[layer.tags] = x
+
+        if self.is_aux:
+            return outputs["Main"], outputs["AUX"]
+        return outputs["Main"], None
+
+    def loss(self, x: tuple[Tensor], entity: DetBatchDataEntity) -> dict:
```
kprokofi: Could you align it with my PR? Or it is better to do after those PR merged, of course

src/otx/algo/common/losses/iou_loss.py
```diff
@@ -76,6 +79,96 @@ def forward(
         )
 
 
+class IoULoss(nn.Module):
```
kprokofi, line 82: Do we use losses from this module for other tasks except for Object Detection?

src/otx/algo/detection/yolov9.py
```diff
@@ -0,0 +1,192 @@
...
+
+        return OTXNativeModelExporter(
+            task_level_export_parameters=self._export_parameters,
+            input_size=(1, 3, *self.input_size),
+            mean=self.mean,
+            std=self.std,
+            resize_mode="fit_to_window_letterbox",
+            pad_value=114,
+            swap_rgb=True,
+            via_onnx=True,
```
kprokofi, line 120: Are there any errors using direct OpenVINO conversion tool?

tests/perf/test_detection.py
```diff
@@ -40,6 +40,9 @@ class TestPerfObjectDetection(PerfTestBase):
         Benchmark.Model(task="detection", name="rtdetr_18", category="other"),
         Benchmark.Model(task="detection", name="rtdetr_50", category="other"),
         Benchmark.Model(task="detection", name="rtdetr_101", category="other"),
+        Benchmark.Model(task="detection", name="yolov9_s", category="other"),
+        Benchmark.Model(task="detection", name="yolov9_m", category="other"),
```
kprokofi, line 44: off topic: I wonder, in what moment we will have too big list of the models to feasibly test them? :)

kprokofi (commented):
Thank you Sungchul for this YOLO integration. It is great to see that we now can support SOTA YOLO models. I left few comments, please take a look.

src/otx/algo/common/losses/iou_loss.py
sungchul1: No, but I just moved it into common/iou_loss because it looked strange that there are two files (common and detection) about iou loss.

src/otx/core/model/base.py
sungchul1: I agree with your suggestion, @kprokofi, it seems the easiest way for the final loss decision.
But I have two concerns:
1. It may make the logging loss part more complicated. Referring to the below code, if `train_loss` has `loss` key, its value is logged at L159 with `train/loss` and it is a correct value. But at L168, a new loss with all values added is overwritten to `train/loss`. We can resolve this issue by changing the key name at L168 as you said (assuming it to `sum_loss`), but if there is a task that uses `loss` as a single loss, we have two keys just for this single loss, `loss` and `sum_loss`.
https://github.com/openvinotoolkit/training_extensions/blob/709d841f4a2c9e8dc73b5e416a1deb7536a6b761/src/otx/core/model/base.py#L157-L175

2. Actually, I'm not sure that `training_step` is a proper location for loss summation. I think that the whole process of loss calculation must be done in loss modules. I'd like to suggest removing L167~174, unifying the key of final loss to `loss`, and just returning the value of `loss` as the final loss. For this, every loss module should be updated and it is not within the scope of this PR.

sungchul1: > so it means that the only model affected is VPM and it has miscalculated loss and it can be solved by current change, right?

Correct.

src/otx/algo/modules/activation.py
sungchul1: nn.SiLU has `inplace` on torch 2.4.0 that is included in develop environment.
image

src/otx/algo/detection/yolov9.py
sungchul1: I agree, maybe it seems because we did refactoring parallel.
I referred vit for classification.
https://github.com/openvinotoolkit/training_extensions/blob/709d841f4a2c9e8dc73b5e416a1deb7536a6b761/src/otx/algo/classification/vit.py#L48-L63

I'll unify this pretrained weight list location for all tasks in the next PR.

src/otx/core/model/base.py
eunwoosh: I agree with the opinion that final loss should be calculated in loss module. I think current problem comes from that same way is used for logging a loss and calculating a loss. If we separate it and change final loss to be calculated in loss module, it can be better.

src/otx/algo/detection/utils/yolov7_v9_utils.py
sungchul1: Ah, it isn't used anymore, removed at bddc817.

src/otx/algo/detection/utils/yolov7_v9_utils.py
sungchul1: As I checked, this module is for yolov7. So I removed at bddc817 for now and I'll consider your suggestion for yolov7 integration.

src/otx/algo/detection/utils/yolov7_v9_utils.py
sungchul1: Thanks for your suggestion, but I think it seems to take a time for this.
Can I keep this now and check it later? I left TODO comment at 960f3d5.

src/otx/algo/detection/utils/yolov7_v9_utils.py
sungchul1: Done at d106c7c

src/otx/algo/detection/yolov9.py
sungchul1: Thanks for the suggestion, done at e9b95ac.

src/otx/algo/detection/heads/yolo_head.py
sungchul1: Done at 8406227

src/otx/algo/detection/heads/yolo_head.py
sungchul1: Sure, I'll update it after merging your iseg refactoring PR.

tests/perf/test_detection.py
sungchul1: I'm not sure and I think we need to replace old models with new advanced models if possible.

src/otx/algo/detection/yolov9.py
sungchul1: I tried with openvino 2024.3, but got the below error:
```shell
  File "venv/otx_v2/lib/python3.11/site-packages/model_api/models/ssd.py", line 121, in __init__
    self.labels_layer = find_layer_by_name(labels_layer, layers)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/otx_v2/lib/python3.11/site-packages/model_api/models/ssd.py", line 82, in find_layer_by_name
    suitable_layers = [layer_name for layer_name in layers if name in layer_name]
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/otx_v2/lib/python3.11/site-packages/model_api/models/ssd.py", line 82, in <listcomp>
    suitable_layers = [layer_name for layer_name in layers if name in layer_name]
                                                              ^^^^^^^^^^^^^^^^^^
TypeError: argument of type 'openvino._pyopenvino.Output' is not iterable
```

It looks that output name is not set and actually output name is not inserted into `openvino.convert_model`.
https://github.com/openvinotoolkit/training_extensions/blob/709d841f4a2c9e8dc73b5e416a1deb7536a6b761/src/otx/core/exporter/native.py#L85-L89

I re-tried it with `output=["boxes", "labels"]` as an argument like below,
```python
exported_model = openvino.convert_model(
    model,
    example_input=dummy_tensor,
    input=(openvino.runtime.PartialShape(self.input_size),),
    output=["boxes", "labels"],
)
```

but another error occurred:
```shell
  File "venv/otx_v2/lib/python3.11/site-packages/openvino/tools/ovc/convert_impl.py", line 494, in _convert
    ov_model = driver(argv, {"conversion_parameters": non_default_params})
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/otx_v2/lib/python3.11/site-packages/openvino/tools/ovc/convert_impl.py", line 250, in driver
    ov_model = moc_emit_ir(prepare_ir(argv), argv)
                           ^^^^^^^^^^^^^^^^
  File "venv/otx_v2/lib/python3.11/site-packages/openvino/tools/ovc/convert_impl.py", line 194, in prepare_ir
    ov_model = moc_pipeline(argv, moc_front_end)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/otx_v2/lib/python3.11/site-packages/openvino/tools/ovc/moc_frontend/pipeline.py", line 136, in moc_pipeline
    oplaces.append(out_desc["name"])
                   ~~~~~~~~^^^^^^^^
KeyError: 'name'
```

It seems that `fe_output_user_data_repack` return the list of dicts that don't have `name` key.
It seems openvino issue that should be escalated and after resolving this, we can try openvino conversion without onnx pipeline again.
https://github.com/openvinotoolkit/openvino/blob/2024.3.0/tools/ovc/openvino/tools/ovc/moc_frontend/pipeline.py#L134-L136
https://github.com/openvinotoolkit/openvino/blob/2024.3.0/tools/ovc/openvino/tools/ovc/moc_frontend/extractor.py#L308-L339

src/otx/core/model/base.py
eunwoosh: I think deciding how to deal with it is out of this PR, but I agree with Kirill's opinion. Developers have to know setting "loss" mean that that value will be used as final loss. I think it is likely to be missed. To avoid it, how about changing it to "sum_loss" or others values that is difficult to set by accident for now?

src/otx/algo/detection/yolov9.py
sungchul1: Made an issue https://github.com/openvinotoolkit/openvino/issues/26457

src/otx/core/model/base.py
sungchul1: I agree with your opinions about distinguishing the final loss with others.
I'd like to suggest `total_loss`, more general name convention than `sum_loss`.
Because some tasks use weighted summation loss, not just adding losses like this YOLOv9.

And the whole process of finding the loss must be done within the loss modules as I said.
Let's do it in the next PR.

src/otx/algo/common/losses/iou_loss.py
kprokofi: I think we could consider moving this loss to the Detection folder instead.

sungchul1: I'll revert it in the next PR about yolov7 integration.

harimkang (dismissed):
LGTM!

eunwoosh (approved):
LGTM

harimkang (approved):

src/otx/algo/detection/heads/yolo_head.py
```diff
@@ -0,0 +1,729 @@
...
+
+        padded_bboxes, padded_labels = self.pad_bbox_labels(entity.bboxes, entity.labels)
+        merged_padded_labels_bboxes = torch.cat((padded_labels, padded_bboxes), dim=-1)
+        return {
+            "main_preds": main_preds,
+            "aux_preds": aux_preds,
+            "targets": merged_padded_labels_bboxes,
+        }
+
+    def loss_by_feat(self, *args, **kwargs) -> NoReturn:
```
kprokofi, line 553: Could you please remove it in the next PR?

kprokofi (approved):

PR #3925, opened 2024-09-02
Update label info

ashwinvaidya17:
### Summary

- Fixes: CVS-151045

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algo/anomaly/padim.py
+++ b/src/otx/algo/anomaly/padim.py
@@ -14,6 +14,7 @@ from anomalib.callbacks.post_processor import _PostProcessorCallback
 from anomalib.models.image import Padim as AnomalibPadim
 
 from otx.core.model.anomaly import OTXAnomaly
+from otx.core.types.label import AnomalyLabelInfo
 from otx.core.types.task import OTXTaskType
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


src/otx/algo/anomaly/padim.py
```diff
@@ -40,6 +42,7 @@ class Padim(OTXAnomaly, AnomalibPadim):
 
     def __init__(
         self,
+        label_info: LabelInfoTypes = AnomalyLabelInfo(),
```
ashwinvaidya17, line 45: @chuneuny-emily do you think I need to pass optimizer and scheduler parameters to the model as well? While it is unused in anomaly models, I can see the reference converter
https://github.com/openvinotoolkit/training_extensions/blob/7f1c7da57365ae362b81542616a8d0ed6517a8f9/src/otx/tools/converter.py#L242

eunwoosh (approved):

src/otx/algo/anomaly/padim.py
chuneuny-emily: > @chuneuny-emily do you think I need to pass optimizer and scheduler parameters to the model as well? While it is unused in anomaly models, I can see the reference converter
> 
> https://github.com/openvinotoolkit/training_extensions/blob/7f1c7da57365ae362b81542616a8d0ed6517a8f9/src/otx/tools/converter.py#L242

If converter.py is using the "optimizer" and scheduler parameters, you may need to pass these to the model. Otherwise, an error may occur while reading the configuration file.

If there is a part of the code that explicitly sets the learning rate of the optimization function, such as config["model"]["init_args"]["optimizer"]["init_args"]["lr"], this means that the optimization function is set in the configuration file. defined and expected to be used. If there is no "optimizer" or "scheduler" section in the configuration file, the above code may cause an error such as KeyError. @wonjuleee - What do you think?

harimkang (approved):

PR #3930, opened 2024-09-04
Enhance labels handling from arrow format for Semantic Segmentation

kprokofi:
### Summary
Make use of Empty label instead of background label. Derive label names from Datumaro arrow dataset
Before:
image
After:
image

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/core/types/export.py
+++ b/src/otx/core/types/export.py
@@ -117,7 +117,7 @@ class TaskLevelExportParameters:
                     all_label_ids += lbl.replace(" ", "_") + " "
 
         if has_labels_ids:
-            for lbl in self.label_info.labels_id:
+            for lbl in self.label_info.labels_id:  # type: ignore[attr-defined]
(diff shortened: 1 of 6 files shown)


src/otx/core/types/export.py
```diff
@@ -98,18 +98,26 @@ def to_metadata(self) -> dict[tuple[str, str], str]:
             dict[tuple[str, str], str]: It will be directly delivered to
             OpenVINO IR's `rt_info` or ONNX metadata slot.
         """
+        has_labels_ids = hasattr(self.label_info, "labels_id")
```
sovrasov, line 101: I think `labels_id` should be propagated to the `LabelInfo` root class and become a first-class citizen alongside with label names. That'd allow to use both labels and IDs transparently without spreading `is None` checks withing the whole OTX code. We can't generate artificial IDs if the real ones is not accessible.

sovrasov (commented):
Good catch, but some generalization is required, since the problem affects labels handling in general

PR #3933, opened 2024-09-05
Remove datumaro attribute id from tiling, add subset names

eugene123tw:
### Summary

* Remove datumaro attribute id from tiling
* Including arrow subset names ("TRAINING", "VALIDATION")

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/core/data/dataset/tile.py
+++ b/src/otx/core/data/dataset/tile.py
@@ -53,6 +53,10 @@ if TYPE_CHECKING:
 # NOTE: Disable private-member-access (SLF001).
 # This is a workaround so we could apply the same transforms to tiles as the original dataset.
 
+# NOTE: Datumaro subset name should be standardized.
+TRAIN_SUBSET_NAMES = ("train", "TRAINING")
+VAL_SUBSET_NAMES = ("val", "VALIDATION")
+
 
 class OTXTileTransform(Tile):
     """OTX tile transform.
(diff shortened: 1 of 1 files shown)


src/otx/core/data/dataset/tile.py
```diff
@@ -188,7 +192,7 @@ def create(
         Returns:
             OTXTileDataset: Tile dataset.
         """
-        if dataset.dm_subset[0].subset == "train":
+        if dataset.dm_subset[0].subset in TRAIN_SUBSET_NAMES:
```
harimkang, line 195: Actually, this `subset_name` can be customized by the user as a parameter in the configuration or CLI if desired (`--data.train_subset.subset_name`).
It's actually correct to compare it to `data.train_subset.subset_name`.
But I can't think of a way to do that unless this class is holding all the data config..

eugene123tw: @harimkang, I think standardizing subset names would completely resolve this issue.

wonjuleee (approved):
image
It works on my side. Let's merge quickly.

src/otx/core/data/dataset/tile.py
wonjuleee: I have thought about this but I don't think we have any additional option for now. @harimkang, how about discussing this later?

harimkang (approved):

PR #3945, opened 2024-09-09
Add integration tests for keypoint detection task

sovrasov:
### Summary
- Integration tests (excluding OV inference, explain, and adaptive input size)
- Fix export

### Checklist


sovrasov:
@yunchu it looks like the tests are not really launched on CI. What should be done to fix this?

src/otx/algo/keypoint_detection/rtmpose.py
```diff
@@ -89,6 +89,9 @@ def __init__(
             torch_compile=torch_compile,
         )
 
+        self.mean = (123.675, 116.28, 103.53)
```
kprokofi: could you use class variables instead?

sovrasov: Done

eunwoosh (approved):

kprokofi (approved):

PR #3946, opened 2024-09-09
Change categories mapping logic

kprokofi:
### Checklist

### License

```python
```

### DIFF
--- a/src/otx/core/data/pre_filtering.py
+++ b/src/otx/core/data/pre_filtering.py
@@ -88,7 +88,7 @@ def remove_unused_labels(dataset: DmDataset, data_format: str, ignore_index: int
         raise ValueError(msg)
     if len(used_labels) == len(original_categories):
         return dataset
-    if data_format == "arrow" and sorted(used_labels)[-1] != len(original_categories) - 1:
+    if data_format == "arrow" and max(used_labels) != len(original_categories) - 1:
         # we assume that empty label is always the last one. If it is not explicitly added to the dataset,
         # (not in the used labels) it will be filtered out.
         mapping = {cat: cat for cat in original_categories[:-1]}
--- a/src/otx/core/data/pre_filtering.py
+++ b/src/otx/core/data/pre_filtering.py
@@ -88,7 +88,15 @@ def remove_unused_labels(dataset: DmDataset, data_format: str, ignore_index: int
         raise ValueError(msg)
     if len(used_labels) == len(original_categories):
         return dataset
+    if data_format == "arrow" and sorted(used_labels)[-1] != len(original_categories) - 1:
+        # we assume that empty label is always the last one. If it is not explicitly added to the dataset,
+        # (not in the used labels) it will be filtered out.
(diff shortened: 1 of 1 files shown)


src/otx/core/data/pre_filtering.py
```diff
@@ -88,7 +88,15 @@ def remove_unused_labels(dataset: DmDataset, data_format: str, ignore_index: int
         raise ValueError(msg)
     if len(used_labels) == len(original_categories):
         return dataset
+    if data_format == "arrow" and sorted(used_labels)[-1] != len(original_categories) - 1:
```
wonjuleee, line 91: Is it right to compare the empty label with the length of original categories?

eunwoosh:
```suggestion
    if data_format == "arrow" and max(used_labels) != len(original_categories) - 1:
```

eunwoosh (commented):
Sorry but I can't understand what it's for. Could you explain more?

kprokofi:
> Sorry but I can't understand what it's for. Could you explain more?

We now have a logic that removes unused labels from dataset. Sometimes it can happen that this pre-filtering removes unannotated objects from Geti:
image
It is undesired behavior that leads to 0 accuracy and issues. Unfortunately, Datumaro returns only IDs, not names. It is hard to distinguish empty label. But I see, that empty label is always in separated label group and it should be the last one. What I did - it is a temporary workaround.

src/otx/core/data/pre_filtering.py
```diff
@@ -88,7 +88,15 @@ def remove_unused_labels(dataset: DmDataset, data_format: str, ignore_index: int
         raise ValueError(msg)
     if len(used_labels) == len(original_categories):
         return dataset
+    if data_format == "arrow" and max(used_labels) != len(original_categories) - 1:
+        # we assume that empty label is always the last one. If it is not explicitly added to the dataset,
+        # (not in the used labels) it will be filtered out.
```
harimkang, line 93: Can we verify that this assumption is always correct? Is there any code or documentation can reference for this?

kprokofi: I will try to find it on the Datumaro side

eunwoosh:
@kprokofi , thanks for your kind explanation. Could you add TODO docstring?

src/otx/core/data/pre_filtering.py
kprokofi: Unfortunately, I haven't found the 100% proof for that. I can only see that labels are sorted in label schema by assigned ID
https://github.com/intel-innersource/frameworks.ai.interactive-ai-workflow.sonoma-creek/blob/386159bd64df784985efb2cc460d9a4b00d63435/src/sc_sdk/sc_sdk/otx/api/entities/label_schema.py#LL314
 And Empty label should have the largest ID value.

harimkang: @wonjuleee Can you confirm that we can use this workaround?

harimkang (approved):

chuneuny-emily (approved):

PR #3951, opened 2024-09-11
Move templates from OTX1.X to OTX2.X

kprokofi:
### Summary
Move templates from OTX1.6 to OTX2.2 to preserve contract between Geti and OTX and to be able to control hyperparameters from OTX2.X side

This is a workaround. In the future we will migrate to our OTX2.X recipes.

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/recipe/_base_/train.yaml
+++ b/src/otx/recipe/_base_/train.yaml
@@ -9,7 +9,7 @@ callbacks:
       check_on_train_epoch_end: false
       min_delta: 0.001
       warmup_iters: 30
-      warmup_epochs: 7
+      warmup_epochs: 3
   - class_path: lightning.pytorch.callbacks.RichProgressBar
     init_args:
       refresh_rate: 1
--- a/src/otx/recipe/_base_/train.yaml
+++ b/src/otx/recipe/_base_/train.yaml
@@ -9,7 +9,7 @@ callbacks:
       check_on_train_epoch_end: false
       min_delta: 0.001
       warmup_iters: 30
-      warmup_epochs: 3
+      warmup_epochs: 7
   - class_path: lightning.pytorch.callbacks.RichProgressBar
     init_args:
       refresh_rate: 1
--- a/src/otx/tools/templates/classification/mobilenet_v3_large_1_cls_incr/template.yaml
+++ b/src/otx/tools/templates/classification/mobilenet_v3_large_1_cls_incr/template.yaml
@@ -1,0 +1,62 @@
+# Description.
+model_template_id: Custom_Image_Classification_MobileNet-V3-large-1x
+name: MobileNet-V3-large-1x
+task_type: CLASSIFICATION
+task_family: VISION
+instantiation: "CLASS"
+summary: Class-Incremental Image Classification for MobileNet-V3-large-1x
+application: ~
+
+# Algo backend.
+framework: OTXClassification v1.2.3
+
+# Task implementations.
+entrypoints:
+  base: otx.algorithms.classification.adapters.mmcls.task.MMClassificationTask
+  openvino: otx.algorithms.classification.adapters.openvino.task.ClassificationOpenVINOTask
+  nncf: otx.algorithms.classification.adapters.mmcls.nncf.task.ClassificationNNCFTask
+
+# Capabilities.
+capabilities:
+  - compute_representations
+
+# Hyperparameters.
+hyper_parameters:
+  base_path: ../configuration.yaml
+  parameter_overrides:
+    learning_parameters:
+      batch_size:
+        default_value: 64
+        auto_hpo_state: POSSIBLE
+      learning_rate:
+        default_value: 0.0058
+        auto_hpo_state: POSSIBLE
+      learning_rate_warmup_iters:
+        default_value: 10
+      num_iters:
+        default_value: 90
+    nncf_optimization:
+      enable_quantization:
+        default_value: true
+      enable_pruning:
+        default_value: false
+      pruning_supported:
+        default_value: true
+      maximal_accuracy_degradation:
+        default_value: 1.0
+    algo_backend:
+      train_type:
+        default_value: Incremental
+
+# Training resources.
+max_nodes: 1
+training_targets:
+  - GPU
+  - CPU
+
+# Stats.
+gigaflops: 0.44
+size: 4.29
+
+# Model spec
+model_category: SPEED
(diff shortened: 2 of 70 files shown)


src/otx/recipe/_base_/train.yaml
```diff
@@ -9,7 +9,7 @@ callbacks:
       check_on_train_epoch_end: false
       min_delta: 0.001
       warmup_iters: 30
-      warmup_epochs: 3
+      warmup_epochs: 7
```
sungchul1: Why is this update needed?
And I think this change must affect other tasks' performance, could you check performance difference before and after this update?

src/otx/tools/templates/detection/detection/configuration.yaml
```diff
@@ -0,0 +1,701 @@
...
+    enum_name: InputSizePreset
+    header: Configure model input size.
+    options:
+      DEFAULT: "Default"
+      AUTO: "Auto"
+      _256x256: "256x256"
+      _384x384: "384x384"
+      _512x512: "512x512"
+      _768x768: "768x768"
+      _1024x1024: "1024x1024"
```
sungchul1, line 264: I don't know much about this, aren't other input sizes needed, like (640, 640) for yolox or (800, 992) for atss?

src/otx/recipe/_base_/train.yaml
harimkang: Yes in my opinion there shouldn't be any modifications to `/_base_/train.yaml`. If we need this, we need to update this in each model recipe.

src/otx/tools/templates/detection/detection/configuration.yaml
harimkang: This is a copy of OTX 1.6, which is set to invisible in the Geti UI, so only Default is used for Geti. If this needs to be visible in the UI in the future, it might be worth revisiting, but this is just a temporary file for now.

src/otx/tools/templates/classification/mobilenet_v3_large_1_cls_incr/template.yaml
```diff
@@ -26,7 +26,7 @@ hyper_parameters:
         default_value: 0.0058
         auto_hpo_state: POSSIBLE
       learning_rate_warmup_iters:
-        default_value: 100
+        default_value: 10
```
kprokofi: @harimkang , please note, that in OTX2.2 it is not the epochs, but iterations. So, the learning rate will be warmed up only 10 iterations. I think it is a bit useless.

src/otx/tools/templates/detection/detection/cspdarknet_yolox_l/template.yaml
```diff
@@ -28,7 +28,7 @@ hyper_parameters:
         default_value: 0.001
         auto_hpo_state: POSSIBLE
       learning_rate_warmup_iters:
-        default_value: 100
+        default_value: 3
```
kprokofi, line 31: @sungchul2 , it is not epochs. It is iterations. 3 iterations of warm-up is useless. Please, consider higher value

src/otx/recipe/_base_/train.yaml
kprokofi: Okay, I will revert it, but to be honest, 7-10 epochs are reasonable value to start early stopping, 3 epochs are not enough for many models to stabilize on small datasets

src/otx/tools/templates/classification/mobilenet_v3_large_1_cls_incr/template.yaml
harimkang: Yeah.. 
As with the other models, set it to 0 to disable it for now. https://github.com/openvinotoolkit/training_extensions/pull/3951/commits/5ca8b4341248bfd33142504c0de071f75770c546

src/otx/recipe/_base_/train.yaml
kprokofi: reverted https://github.com/openvinotoolkit/training_extensions/pull/3951/commits/d3e910e76cf62f2f63f8ea5623041c6bafc4e9f9

sungchul1: We cannot be sure until we test it experimentally because small epochs may be enough on small datasets because models only have to adapt to the distribution for a small number of samples.
But this is not in the scope of this PR that temporarily adds templates.

src/otx/tools/templates/detection/detection/cspdarknet_yolox_l/template.yaml
sungchul1: I know and I think it is strange setting as well :D
But it is just for following recipes.

kprokofi: Do we have also iterations for this parameter in OTX1.6? Or maybe it was epochs

sungchul1: I'm not sure but maybe iteration. @harimkang do you have any idea?

harimkang: But this is actually going to be converted to an OTX 2.2 recipe now, so we can use it the way we're using it now, right? (We just need to choose whether we want to use epochs or iter, and it's up to the IAI team to change this key to iter -> epoch.)

kprokofi: Yes, let's consider hyperparameters update, especially for warm-up and early stopping in the nearest future.

eugene123tw (commented):
Thanks, Kirill! I have a few general questions:

* Even though new models aren’t introduced in OTX 1.x, should we still include them in `converter.py`? For example, we didn’t have `Custom_Instance_Segmentation_RTMDet_tiny` in OTX 1.x, but it's still present in the converter. I assume this is more about ensuring Geti recognizes the format rather than converting from OTX 1.x to OTX 2.x. Is that correct?

kprokofi:
> Thanks, Kirill! I have a few general questions:
> 
> * Even though new models aren’t introduced in OTX 1.x, should we still include them in `converter.py`? For example, we didn’t have `Custom_Instance_Segmentation_RTMDet_tiny` in OTX 1.x, but it's still present in the converter. I assume this is more about ensuring Geti recognizes the format rather than converting from OTX 1.x to OTX 2.x. Is that correct?

Actually, we added new model templates and updated converter accordingly. RTMDet model will be added in Geti new version. Yunchu will update docker image and OTX2.2 will have separated templates from 1.6. So, adding new models are not a problem.
The main problem today is that all components from Geti are relied on the old contract between OTX and Geti. So, we cannot change the mapping between new recipes and Geti fast. IAI team will adapt their side for our new recipes, but it requires some time and for now we need to preserve old templates.

harimkang (approved):

sungchul1 (approved):

PR #3957, opened 2024-09-13
Add OV model for keypoint detection

sovrasov:
### Summary
- OVModel
- fix torch functionality (export & benchmark)
- minor fix in metric computation
- CLI&Unit tests

Results on coco_keypoint_single_obj/medium:
| Model   | PCK accuracy |
|---|---|
| RTMPose-tiny-single   | 0.7216 |
| RTMPose-tiny-single OV  | 0.7216  |
| RTMPose-tiny-single OV INT8  | 0.7208  |


src/otx/core/model/keypoint_detection.py
```diff
@@ -124,7 +123,7 @@ def _convert_pred_entity_to_compute_metric(
             ],
             "target": [
                 {
-                    "keypoints": kpt,
+                    "keypoints": kpt,  # / np.array(self.input_size[::-1]),
```
kprokofi: is this comment needed?

src/otx/core/model/keypoint_detection.py
```diff
@@ -154,9 +153,100 @@ def forward_for_tracing(self, image: torch.Tensor) -> torch.Tensor | tuple[torch
...
-            task_type="detection",
+            model_type="keypoint_detection",
+            task_type="keypoint_detection",
             confidence_threshold=self.hparams.get("best_confidence_threshold", None),
-            iou_threshold=0.5,
-            tile_config=self.tile_config if self.tile_config.enable_tiler else None,
         )
+
+    def get_dummy_input(self, batch_size: int = 1) -> KeypointDetBatchDataEntity:
+        """Returns a dummy input for classification model."""
```
kprokofi:
```suggestion
        """Returns a dummy input for key point detection model."""
```

kprokofi (dismissed):

src/otx/recipe/keypoint_detection/openvino_model.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+    test_subset:
+      batch_size: 64
+      num_workers: 2
+      transforms:
+        - class_path: otx.core.data.transform_libs.torchvision.GetBBoxCenterScale
+        - class_path: otx.core.data.transform_libs.torchvision.TopdownAffine
+          init_args:
+            input_size: $(input_size)
+            is_numpy_to_tvtensor: true
```
eunwoosh: Is it necessary to convert it to tv tensor?

src/otx/core/metrics/pck.py
```diff
@@ -190,19 +178,33 @@ def update(self, preds: list[dict[str, Tensor]], target: list[dict[str, Tensor]]
...
+        kpts_visible = []
+        for p in self.targets:
+            if len(p[1].shape) == 3 and p[1].shape[0] == 1:
+                kpts_visible.append(p[1].squeeze())
+            else:
+                kpts_visible.append(p[1])
+
+        kpts_visible_stacked = np.stack(kpts_visible)
+
+        normalize = np.tile(np.array([self.input_size[::-1]]), (pred_kpts.shape[0], 1))
```
eunwoosh, line 198: the reason why input size order is reversed is that norm_factor order is (w, h)? Referring docstring of the function, it should be (h, w). Which one is right?

eunwoosh (commented):

src/otx/recipe/keypoint_detection/openvino_model.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - data.test_subset.transforms
+
+  data:
+    stack_images: false
+    train_subset:
+      batch_size: 1
+      num_workers: 2
+      transforms:
+        - class_path: otx.core.data.transform_libs.torchvision.GetBBoxCenterScale
+        - class_path: otx.core.data.transform_libs.torchvision.TopdownAffine
```
sungchul2, line 30: Are these transformations not supported in model api?

src/otx/recipe/keypoint_detection/openvino_model.yaml
```diff
@@ -0,0 +1,53 @@
...
+          init_args:
+            input_size: $(input_size)
+            is_numpy_to_tvtensor: true
+
+    test_subset:
+      batch_size: 64
+      num_workers: 2
+      transforms:
+        - class_path: otx.core.data.transform_libs.torchvision.GetBBoxCenterScale
+        - class_path: otx.core.data.transform_libs.torchvision.TopdownAffine
```
sungchul1, line 48: Are these transformations necessary for ov inference? Isn't there any post-processing for predicted key points in model api?

sovrasov: MAPI has post-processing for keypoints, while OTX does not, which means OTX doesn't expect postprocessed keypoints to be predicted and handled in the metric for instance. By design, OTX OVModel is just a quick substitute to a torch one, and if torch model is not fully functional, OVModel has to provide a similar output to keep the implementation clean.

src/otx/recipe/keypoint_detection/openvino_model.yaml
sovrasov: Right, we don't need this

src/otx/core/metrics/pck.py
sovrasov: It's wh, since we use xy keypoints coordinates. The function is order-agnostic, we can use it for both xy and xy. The main point is that we need to pass aligned kps and normalization, which are in xy format for our case.

chuneuny-emily:
@sovrasov - Could you please designate the target branch as the develop branch? OTX2.2.0 has already proven that KeyPoint detection supports train and export. OV IR inference validation will be available in the upcoming OTX2.3.0.

src/otx/recipe/keypoint_detection/openvino_model.yaml
sungchul1: I agree with OVModel must be quick substitutes to torch ones.
But I think OTX must expect post-processed outputs for validation and test because what we and users really want is processed ones, not raw predictions.
And det, iseg, and vpm seem to use post-processed predictions for metric calculation or returning outputs (I don't know why semantic segmentation and keypoint detection don't adapt this pipeline).
Could you share your opinion about this @kprokofi @harimkang?

src/otx/core/metrics/pck.py
eunwoosh: Thanks for explanation. Could you update `norm_factor` docstring at `keypoint_pck_accuracy` then?

src/otx/recipe/keypoint_detection/openvino_model.yaml
sovrasov: The same problem applies to hierarchical classification: we have to disregard postprocessing coming from MAPI, because otherwise we'd have to reproduce the same postprocessing in torch. Also, we're using the same prediction entities for OV and torch models. If the outputs of OV and torch are different, while the same entities are used, that'd rise questions from user's side, let alone neither of output formats is documented.
If we think about usecases for real inference (the results are supposedly utilized outside OTX), I won't say OTX batch entities are somehow convenient. The real strong reason to prefer OTX inference over MAPI is CUDA, but we don't provide meaningful inference results in that scenario and do not even target it.

sovrasov:
@chuneuny-emily just in case, export of keypoint model in 2.2 can be launched, but it generates an incorrect model, so it can be used only as a mock.

src/otx/recipe/keypoint_detection/openvino_model.yaml
sungchul1: But for detection case, post-processing for predicted bounding boxes is already being used, isn't it?
https://github.com/openvinotoolkit/model_api/blob/master/model_api/python/model_api/models/detection_model.py#L78-L124

Current OTX's models are based on mmlab models and as far as I know those models usually have their own post-processing pipeline like below detection usecase.
https://github.com/openvinotoolkit/training_extensions/blob/b99770a95d2b5ce8c72ec85534a4d690428eb0de/src/otx/algo/detection/heads/base_head.py#L313-L374

Current OTX detection models don't do pre-processing for ground truths during validation or inference considering the above post-processing.
(the arg `transform_bbox` of `otx.core.data.transform_libs.torchvision.Resize` is set to False as default)
https://github.com/openvinotoolkit/training_extensions/blob/b99770a95d2b5ce8c72ec85534a4d690428eb0de/src/otx/recipe/_base_/data/detection.yaml#L37-L77

I agree with your opinion about using unified prediction entity for both otx and ov models, but I think the predictions must be post-processed one. If no, I think it's like losing one of MAPI's core features in OTX.

sovrasov: I see that for some of the models predictions made by torch and OV models could be close to the ready-to use one, but that's nothing more than a coincidence. OTX defines prediction format by defining an evaluation procedure, and that's it. Previously, I proposed introduction of two modes of prediction in OTX: one for metric computation, and one for generating user-friendly predictions. My proposal was declined. We can discuss this once again, but the current arrangement is that inference is done by MAPI, not OTX.
For kp task to resolve the issue you mention, we can file a ticket to Wonju, so he refactors torch & evaluation to support prediction & evaluation of postprocessed keypoints coordinates, then the OV model can be upgraded accordingly.

PR #3973, opened 2024-09-23
Fix config converter for tiling

eugene123tw:
### Summary

Fix config converter for tiling

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/tools/converter.py
+++ b/src/otx/tools/converter.py
@@ -438,6 +438,9 @@ class ConfigConverter:
         model_parser.add_subclass_arguments(OTXModel, "model", required=False, fail_untyped=False, skip={"label_info"})
         model = model_parser.instantiate_classes(Namespace(model=model_config)).get("model")
 
+        if hasattr(model, "tile_config"):
+            model.tile_config = datamodule.tile_config
+
         # Instantiate Engine
         config_work_dir = config.pop("work_dir", config["engine"].pop("work_dir", None))
         config["engine"]["work_dir"] = work_dir if work_dir is not None else config_work_dir


src/otx/tools/converter.py
```diff
@@ -438,6 +438,9 @@ def instantiate(
         model_parser.add_subclass_arguments(OTXModel, "model", required=False, fail_untyped=False, skip={"label_info"})
         model = model_parser.instantiate_classes(Namespace(model=model_config)).get("model")
 
+        if hasattr(model, "tile_config"):
+            model.tile_config = datamodule.tile_config
```
harimkang, line 442: Personally, I think this code should be inside Engine, what do you think?

harimkang (approved):

src/otx/tools/converter.py
eunwoosh: Should it be applied to OTX alone use case? If it's true, I also think that it should be implemented in Engine (or maybe CLI also).

chuneuny-emily (approved):

src/otx/tools/converter.py
eugene123tw: @harimkang We currently handle patching in the CLI via `instantiate_model`, but we missed including tiling patching during the Geti integration. A more secure approach would be to handle this patching directly in  `LightningModule`, so it's not overlooked in future updates or changes to the CLI or Geti integration. That being said, it's not the most beautiful way to patch via `self._trainer.datamodule`.

```python
class OTXInstanceSegModel(OTXModel[InstanceSegBatchDataEntity, InstanceSegBatchPredEntity]):
    def __init__(
        self,
        label_info: LabelInfoTypes,
        input_size: tuple[int, int],
        optimizer: OptimizerCallable = DefaultOptimizerCallable,
        scheduler: LRSchedulerCallable | LRSchedulerListCallable = DefaultSchedulerCallable,
        metric: MetricCallable = MaskRLEMeanAPFMeasureCallable,
        torch_compile: bool = False,
        tile_config: TileConfig = TileConfig(enable_tiler=False),
    ) -> None:
        ...
        
    def setup(self, stage):
        super().setup(stage)
        self.tile_config = self._trainer.datamodule.tile_config
```

harimkang: Yes, proceed like this for now, but I think it would be good to change it to a cleaner method later. This seems like a difficult structure to understand.

PR #3977, opened 2024-09-24
MaskDINO ResNet50

eugene123tw:
### Summary

Introduces MaskDINO ResNet50 model:
* Added new MaskDINO modules
* Integrated Detectron2 ResNet backbone.
  * Note: While I plan to use the existing ResNet50, the current detectron2.R50 will remain for now.
* Fixed specific issues in the RT-DETR decoder

### Checklist

### License

```python
```


src/otx/algo/instance_segmentation/backbones/detectron_resnet.py
```diff
@@ -0,0 +1,332 @@
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
+
+"""Detectron2 ResNet."""
```
eunwoosh: Isn't it necessary to add detectron copyright?

src/otx/algo/instance_segmentation/maskdino.py
```diff
@@ -0,0 +1,509 @@
...
+            batch_masks,
+        )
+
+
+class MaskDINOR50(ExplainableOTXInstanceSegModel):
+    """OTX MaskDINO model with ResNet50 backbone."""
+
+    load_from = "https://github.com/IDEA-Research/detrex-storage/releases/download/maskdino-v0.1.0/maskdino_r50_50ep_300q_hid2048_3sd1_instance_maskenhanced_mask46.3ap_box51.7ap.pth"
+    image_size = (1, 3, 1024, 1024)
+    tile_image_size = (1, 3, 512, 512)
```
eunwoosh: Could you remove them and align a way to set input_size to other models?

src/otx/algo/instance_segmentation/maskdino.py
```diff
@@ -0,0 +1,509 @@
...
+                def step(self, closure: Callable | None = None) -> None:
+                    all_params = itertools.chain(*[x["params"] for x in self.param_groups])
+                    torch.nn.utils.clip_grad_norm_(all_params, clip_gradients_value)
+                    super().step(closure=closure)
+
+            return FullModelGradientClippingOptimizer
+
+        return add_full_model_gradient_clipping(torch.optim.AdamW)(params, base_lr)
+
+    def _customize_inputs(self, entity: InstanceSegBatchDataEntity) -> dict[str, InstanceSegBatchDataEntity]:
```
eunwoosh: I think it looks ok to use parent class' function. How about removing it?

src/otx/algo/instance_segmentation/maskdino.py
```diff
@@ -0,0 +1,509 @@
...
+            lr_scheduler_config = {"scheduler": scheduler}
+            if hasattr(scheduler, "interval"):
+                lr_scheduler_config["interval"] = scheduler.interval
+            if hasattr(scheduler, "monitor"):
+                lr_scheduler_config["monitor"] = scheduler.monitor
+            lr_scheduler_configs.append(lr_scheduler_config)
+
+        return [optimizer], lr_scheduler_configs
+
+    def _build_optimizer(self, model: nn.Module) -> torch.optim.Optimizer:
```
eunwoosh: I think this function ignore the recipe. So, even if user set optimizer or learning rate, it can't be applied. Could you change a code to use given `OptimizerCallable`?

src/otx/algo/instance_segmentation/backbones/detectron_resnet.py
```diff
@@ -0,0 +1,332 @@
...
+        c2_msra_fill(self.conv1)
+
+    def forward(self, x: Tensor) -> Tensor:
+        """Forward pass."""
+        x = self.conv1(x)
+        x = f.relu_(x)
+        return f.max_pool2d(x, kernel_size=3, stride=2, padding=1)
+
+
+class ResNet(nn.Module):
```
eunwoosh, line 134: This is a question. Isn't it possible to user existing resnet class?

src/otx/algo/instance_segmentation/heads/pixel_decoder/maskdino_encoder.py
```diff
@@ -0,0 +1,444 @@
...
+            transformer_dropout: dropout probability in transformer
+            transformer_nheads: number of heads in transformer
+            transformer_dim_feedforward: dimension of feedforward network
+            transformer_enc_layers: number of transformer encoder layers
+            conv_dims: number of output channels for the intermediate conv layers.
+            mask_dim: number of output channels for the final conv layer.
+            norm (str): normalization for all conv layers
+            num_feature_levels: feature scales used
+            total_num_feature_levels: total feautre scales used (include the downsampled features)
+        """
```
eunwoosh: Could you move this doctsring to one above it to align it to other OTX style?

src/otx/algo/utils/mmengine_utils.py
```diff
@@ -206,6 +206,9 @@ def load_checkpoint_to_model(
     # TODO(Kirill): remove this when RTDETR weights is updloaded to openvino storage.
     state_dict = checkpoint["ema"]["module"] if "ema" in checkpoint else checkpoint.get("state_dict", checkpoint)
 
+    # TODO(Eugene): remove this when MaskDINO weights is updloaded to openvino storage.
+    state_dict = state_dict["model"] if "model" in state_dict else state_dict
```
eunwoosh, line 210: Aren't other models affected by this change?

src/otx/algo/instance_segmentation/heads/transformer_decoder/maskdino_decoder.py
```diff
@@ -0,0 +1,425 @@
...
+        total_num_feature_levels: int = 4,
+        dropout: float = 0.0,
+        nhead: int = 8,
+        dec_n_points: int = 4,
+        query_dim: int = 4,
+    ) -> None:
+        """NOTE: this interface is experimental.
+
+        Args:
+            in_channels: channels of the input features
```
eunwoosh: Could you move this docstring to the one above?

src/otx/algo/instance_segmentation/losses/maskdino_criterion.py
```diff
@@ -0,0 +1,405 @@
...
+        matcher: HungarianMatcher,
+        weight_dict: dict[str, float],
+        eos_coef: float,
+        losses: list[str],
+        num_points: int,
+        oversample_ratio: float,
+        importance_sample_ratio: float,
+        dn_losses: list[str],
+    ) -> None:
+        """Create the criterion.
```
eunwoosh: Could you move this docstring to the one above?

src/otx/algo/instance_segmentation/utils/utils.py
```diff
@@ -1,18 +1,23 @@
...
+from typing import Callable
+
 import torch
+import torch.nn.functional as f
+from torch import Tensor, nn
 
 from otx.algo.utils.mmengine_utils import InstanceData
 from otx.core.data.entity.instance_segmentation import InstanceSegBatchDataEntity
 
+TORCH_VERSION = tuple(int(x) for x in torch.__version__.split(".")[:2])
```
eunwoosh: Could you remove it? OTX uses fixed torch version, so it's unnecessary.

src/otx/algo/instance_segmentation/utils/utils.py
```diff
@@ -127,3 +132,276 @@ def empty_instances(
...
+        pos_h = torch.stack((pos_h[:, :, 0::2].sin(), pos_h[:, :, 1::2].cos()), dim=3).flatten(2)
+
+        pos = torch.cat((pos_y, pos_x, pos_w, pos_h), dim=2)
+    else:
+        msg = f"Unknown pos_tensor shape(-1):{pos_tensor.size(-1)}"
+        raise ValueError(msg)
+    return pos
+
+
+def _get_clones(module: nn.Module, nums: int) -> nn.ModuleList:
```
eunwoosh: If it's used outside this file, then how about renaming it to `get_clones`?

src/otx/algo/instance_segmentation/utils/utils.py
```diff
@@ -127,3 +132,276 @@ def empty_instances(
...
+        self.layers = nn.ModuleList(nn.Linear(n, k) for n, k in zip([input_dim, *h], [*h, output_dim], strict=True))
+
+    def forward(self, x: Tensor) -> Tensor:
+        """Forward pass."""
+        for i, layer in enumerate(self.layers):
+            x = f.relu(layer(x)) if i < self.num_layers - 1 else layer(x)
+        return x
+
+
+def inverse_sigmoid(x: Tensor, eps: float = 1e-5) -> Tensor:
```
eunwoosh: Same function exists at algo/detection/utils.py. How about moving a code to common space ans using it?

eunwoosh (commented):
Thanks for adding a new model :) I left some comments, please take a look.
And there are many classes and functions which don't have docstring for input arguments and output value. Could you update it?
And although quite a lot of functions and classes are added, unit test for MaskDINOR50 is only added. 
Could you add unit test for every each module?

src/otx/algo/instance_segmentation/backbones/detectron_resnet.py
eugene123tw: I've made some adjustments to the comment, but I'm unsure about adding Detectron2's copyright notice (i.e., 'Copyright (c) Facebook'), as we will have 2 copyrights, and we haven't done so for the other models. I think the process for including license information should be addressed separately.

src/otx/algo/instance_segmentation/backbones/detectron_resnet.py
eugene123tw: Not at the moment. The structure of detectron.ResNet differs from the version of ResNet we have in MMDet, meaning we will have issues loading weights without implementing a function to handle the layer discrepancies. We can consider implementing this later, but it's not a priority right now.

src/otx/algo/instance_segmentation/maskdino.py
eugene123tw: Updated

src/otx/algo/instance_segmentation/maskdino.py
eugene123tw: Updated

src/otx/algo/instance_segmentation/heads/transformer_decoder/maskdino_decoder.py
eugene123tw: Updated

src/otx/algo/instance_segmentation/losses/maskdino_criterion.py
eugene123tw: Updated

src/otx/algo/instance_segmentation/utils/utils.py
eugene123tw: Updated

src/otx/algo/instance_segmentation/utils/utils.py
eugene123tw: Updated

src/otx/algo/instance_segmentation/maskdino.py
eugene123tw: Updated.

src/otx/algo/instance_segmentation/utils/utils.py
eugene123tw: Updated

src/otx/algo/utils/mmengine_utils.py
eugene123tw: No, only MaskDINO stores weights under `model` key. RTDETR pretrained weights do something similar, saving weights under `module` key. I'm following the same approach as RTDETR. That said, I can override `_create_model` to load the weights. However, neither option is entirely straightforward imo.

src/otx/algo/instance_segmentation/maskdino.py
```diff
@@ -0,0 +1,518 @@
...
+            task_level_export_parameters=self._export_parameters,
+            input_size=(1, 3, *self.input_size),
+            mean=self.mean,
+            std=self.std,
+            resize_mode="standard",
+            swap_rgb=False,
+            via_onnx=True,
+            onnx_export_configuration={
+                "input_names": ["image"],
+                "output_names": ["boxes", "labels", "masks"],
```
sovrasov: `"output_names"` here is not required, because the exporter will use the names coming several lines below

src/otx/algo/instance_segmentation/backbones/detectron_resnet.py
eunwoosh: I asked it and heard that if we don't change code at all, then we need to add their copy right and if not, we need to add both intel and their copy right. Could you add it? you can refer src/otx/algo/classification/heads/linear_head.py

src/otx/algo/instance_segmentation/maskdino.py
```diff
@@ -0,0 +1,517 @@
...
+                    hyperparams["weight_decay"] = 0.0
+                if isinstance(module, norm_module_types):
+                    hyperparams["weight_decay"] = weight_decay_norm
+                if isinstance(module, torch.nn.Embedding):
+                    hyperparams["weight_decay"] = weight_decay_embed
+                params.append({"params": [value], **hyperparams})
+        return params
+
+    @staticmethod
+    def _add_grad_clipping(
```
eunwoosh: Isn't it possible to use `gradient_clip_val` of torchlightning?

src/otx/algo/instance_segmentation/maskdino.py
```diff
@@ -0,0 +1,517 @@
...
+            lr_scheduler_configs.append(lr_scheduler_config)
+
+        return [optimizer_with_grad_clip], lr_scheduler_configs
+
+    def _get_optim_params(self, model: nn.Module) -> list[dict[str, Any]]:
+        """Get optimizer parameters."""
+        _optimizer = self.optimizer_callable(self.parameters())
+
+        # Configurable from MaskDINO recipe
+        base_lr = _optimizer.defaults["lr"]
```
eunwoosh, line 376: User may use other configurations such as `eps` or etc.. Could you consider it?

src/otx/algo/instance_segmentation/heads/transformer_decoder/maskdino_decoder.py
```diff
@@ -0,0 +1,422 @@
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
+"""MaskDINODecoder module."""
+from __future__ import annotations
+
+import torch
+from torch import Tensor, nn
+
+from otx.algo.detection.utils.utils import inverse_sigmoid
```
eunwoosh: Could you move `inverse_sigmoid` to common space?

eunwoosh (commented):
Thanks for applying my comments. I added a few comments. Please check it.
And could you align all docstring to OTX style and add unit test for all added modules?

src/otx/algo/instance_segmentation/maskdino.py
eugene123tw: I agree that basic optimizer parameters (e.g., optimizer, learning rate, weight decay) should be configurable for now. However, I don't think it's necessary to expose advanced optimizer settings in this PR, as I'm not yet familiar with their impact on MaskDINO. Given the above reason, I suggest merging this PR as is and addressing optimization and fine-tuning in a later phase.

src/otx/algo/instance_segmentation/maskdino.py
eugene123tw: I think it's possible, but I plan to carry out further experiments and fine-tuning in the next phase.

src/otx/algo/instance_segmentation/heads/transformer_decoder/maskdino_decoder.py
eugene123tw: Updated

eugene123tw:
> Thanks for applying my comments. I added a few comments. Please check it. And could you align all docstring to OTX style and add unit test for all added modules?

Module unit test covered under: `tests/unit/algo/instance_segmentation/test_maskdino.py`

sovrasov (dismissed):

src/otx/algo/instance_segmentation/backbones/detectron_resnet.py
sungchul1: I think we must pursue to reuse the models implemented in otx.
If there are models redundantly used across tasks, it should be refactored as you said.
But for newly added models, adding ways to adapt them to otx models will reduce our future work.
Could you find ways for this?

src/otx/algo/instance_segmentation/maskdino.py
eunwoosh: If so, how about using `gradient_clip_val` instead of `_add_grad_clipping` function and tuning the value in the next phase?

eunwoosh:
> > Thanks for applying my comments. I added a few comments. Please check it. And could you align all docstring to OTX style and add unit test for all added modules?
> 
> Module unit test covered under: `tests/unit/algo/instance_segmentation/test_maskdino.py`

I think an objective of unit test is to check each component is executed as we expected and one of principles to make good unit test is testing one thing ([refer](https://stackoverflow.com/questions/61400/what-makes-a-good-unit-test)). "tests/unit/algo/instance_segmentation/test_maskdino.py" may run most of code added but It is hard to say that it checks every component works as we expected. If it's too time consuming to implement all unit tests, then how about skipping a code copied and not changed at all but implementing a unit test for code implemented or changed by our side?

src/otx/algo/instance_segmentation/heads/pixel_decoder/position_encoding.py
```diff
@@ -0,0 +1,60 @@
...
+"""Various positional encodings for the transformer."""
+from __future__ import annotations
+
+import math
+
+import torch
+from torch import Tensor, nn
+
+
+class PositionEmbeddingSine(nn.Module):
```
sungchul1, line 14: It looks similar with below, could you check if it can be used instead?
https://github.com/openvinotoolkit/training_extensions/blob/7773abb5bf6c44e9d180a91b861fc895ecd6e6d2/src/otx/algo/object_detection_3d/backbones/monodetr_resnet.py#L19-L67

src/otx/algo/instance_segmentation/layers/batch_norm.py
```diff
@@ -0,0 +1,109 @@
...
+"""Detectron2 batch norm modules."""
+
+from __future__ import annotations
+
+import torch
+from torch import nn
+from torch.nn import functional as f
+
+
+class FrozenBatchNorm2d(nn.Module):
```
sungchul1, line 13: It looks similar with below, could you check if it can be used instead?
https://github.com/openvinotoolkit/training_extensions/blob/7773abb5bf6c44e9d180a91b861fc895ecd6e6d2/src/otx/algo/modules/norm.py#L20-L83

src/otx/algo/instance_segmentation/losses/hungarian_matcher.py
```diff
@@ -0,0 +1,202 @@
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
+
+"""Modules to compute the matching cost and solve the corresponding LSAP."""
```
sungchul1, line 5: Just a question, what is LSAP?

src/otx/algo/instance_segmentation/losses/maskdino_criterion.py
```diff
@@ -0,0 +1,425 @@
...
+    loss = ce_loss * ((1 - p_t) ** gamma)
+
+    if alpha >= 0:
+        alpha_t = alpha * targets + (1 - alpha) * (1 - targets)
+        loss = alpha_t * loss
+
+    return loss.mean(1).sum() / num_boxes
+
+
+def dice_loss(
```
sungchul1, line 49: It looks same with below
https://github.com/openvinotoolkit/training_extensions/blob/7773abb5bf6c44e9d180a91b861fc895ecd6e6d2/src/otx/algo/instance_segmentation/losses/dice_loss.py#L17-L74

src/otx/algo/instance_segmentation/maskdino.py
```diff
@@ -0,0 +1,537 @@
...
+from otx.algo.instance_segmentation.utils.utils import ShapeSpec
+from otx.core.data.entity.base import ImageInfo, OTXBatchLossEntity
+from otx.core.data.entity.instance_segmentation import InstanceSegBatchDataEntity, InstanceSegBatchPredEntity
+from otx.core.exporter.base import OTXModelExporter
+from otx.core.exporter.native import OTXNativeModelExporter
+from otx.core.model.instance_segmentation import ExplainableOTXInstanceSegModel
+from otx.core.utils.mask_util import polygon_to_bitmap
+
+
+class _MaskDINO(nn.Module):
```
sungchul1: Why do you use _ as a prefix?

src/otx/algo/instance_segmentation/maskdino.py
```diff
@@ -0,0 +1,537 @@
...
+        self.size_divisibility = size_divisibility
+
+        # additional args
+        self.test_topk_per_image = test_topk_per_image
+
+        self.focus_on_box = focus_on_box
+        self.transform_eval = transform_eval
+
+    @classmethod
+    def from_config(cls, num_classes: int) -> _MaskDINO:
```
sungchul1: Please add builders in each module, for example, `MaskDINOHead` as a builder of `MaskDINOHeadModule`, instead of hard-coding configs here.

src/otx/algo/utils/mmengine_utils.py
sungchul1: I think here is a common part that can quietly affect other models as well.
I tried model-dependent module like below, could you check if MaskDINO can also be adapted?
https://github.com/openvinotoolkit/training_extensions/blob/7773abb5bf6c44e9d180a91b861fc895ecd6e6d2/src/otx/algo/detection/detectors/single_stage_detector.py#L290-L338

src/otx/algo/instance_segmentation/maskdino.py
```diff
@@ -0,0 +1,537 @@
...
+            images=inputs.images,
+            imgs_info=inputs.imgs_info,
+            scores=scores,
+            bboxes=bboxes,
+            masks=masks,
+            polygons=[],
+            labels=labels,
+        )
+
+    def post_process_instance_segmentation(
```
sungchul1, line 452: What about inserting this in head?
I think lit module is inappropriate to do this operation and the model should support this.

sungchul1 (commented):
Thanks for this update, I'd like to request few things:
1. there seems to be a lot of duplicate implementations with the otx implementation. Could you check it and if they can be used instead or updated for your implementation?
2. could you update all docstrings in all files you worked?
3. could you consider to modify your implementations to align it with otx instead of adding them as is?

src/otx/algo/instance_segmentation/heads/pixel_decoder/position_encoding.py
eugene123tw: There are a few modules that could be shared, but I don't plan to refactor them in this PR. Doing so would introduce many unrelated changes to overlapping modules, which is not my intent for this PR. I will address this in the next optimization phase.

src/otx/algo/instance_segmentation/layers/batch_norm.py
eugene123tw: I think it can, that said, I plan to reduce the duplication in the next phase.

src/otx/algo/instance_segmentation/losses/hungarian_matcher.py
eugene123tw: It's Linear Sum Assignment Problem.

src/otx/algo/instance_segmentation/losses/maskdino_criterion.py
eugene123tw: I will remove the duplication in the next phase.

src/otx/algo/instance_segmentation/maskdino.py
eugene123tw: yeah, I will remove this function and only specify `gradient_clip_val` in config yaml

src/otx/algo/instance_segmentation/backbones/detectron_resnet.py
eugene123tw: yeah, reuse is possible, but it's not exactly a striaghtforward task. It'd be great if everything is ready in one PR—but alas, I can only do it in stages. Rome wasn’t built in a day, yeah?

src/otx/algo/instance_segmentation/maskdino.py
eugene123tw: I’ll keep it as-is for now, since ResNet50 is the only supported version. I’ll add the module builder in the next PR. Either way, everything will be hard-coded, just split across different modules.

src/otx/algo/instance_segmentation/maskdino.py
eugene123tw: Updated

src/otx/algo/instance_segmentation/backbones/detectron_resnet.py
sungchul1: Yes, it's annoying. But if we work on it while considering it, the new model will naturally fit into the library structure, the code will be cleaner, the PR size will be smaller, the person in charge will have less to consider, and the reviewers will have an easier time reviewing.
And what I would like to suggest is that the work of adding a new model should include performance reproduction and refactoring to fit the library structure.

eugene123tw:
> Thanks for this update, I'd like to request few things:
> 
> 1. there seems to be a lot of duplicate implementations with the otx implementation. Could you check it and if they can be used instead or updated for your implementation?
> 2. could you update all docstrings in all files you worked?
> 3. could you consider to modify your implementations to align it with otx instead of adding them as is?

@sungchul2 Thanks for reviewing! The first and third questions are pretty similar—essentially, there are just too many modules to refactor all at once. My intent with this PR is to make the candidate algorithm compatible with OTX (training, testing, prediction, export, optimizer), rather than solving everything in one go. Testing and refactoring will happen in the next stage.

As for the second question, I’m not sure I fully understand. Are you asking me to update "ALL" the docstrings for MaskDINO? I don’t think it’s necessary unless they’re missing. Afaik, all the docstrings are already in place. Could you clarify?

src/otx/algo/instance_segmentation/maskdino.py
eugene123tw: I think this part still needs further optimization and fine-tuning for both PyTorch and OpenVINO export. I'm currently experimenting with fine-tuning the post-processing. In short, refactoring and optimization are in progress, and changes are expected in the next phase, so I’ll leave it as is for now.

src/otx/algo/instance_segmentation/backbones/detectron_resnet.py
eugene123tw: I mean, using the model module directly from an existing module isn't straightforward. For instance, this PR is created before detection 3D, but after merging 3D PR first, we end up with duplicate code.

I think at this stage, the reviewer can be selective and focus on areas that impact OTX's core functionality, such as testing, exporting, prediction, and post-processing. Since much of the code is being migrated, it’s impossible to review every line in detail. Instead, we can leave the migrated code as is, as long as there are no major issues, and as implementer focusing on further refactoring.

This doesn’t mean we should migrate code without consideration, but rather that we can approach it in stages. I’m not sure a full model integration with complete refactoring is necessary all at once in one PR, and I’d be fine refactoring in a separate PR like we did for other PRs.

src/otx/algo/instance_segmentation/backbones/detectron_resnet.py
eugene123tw: Updated.

sungchul1:
> @sungchul2 Thanks for reviewing! The first and third questions are pretty similar—essentially, there are just too many modules to refactor all at once. My intent with this PR is to make the candidate algorithm compatible with OTX (training, testing, prediction, export, optimizer), rather than solving everything in one go. Testing and refactoring will happen in the next stage.
> 
> As for the second question, I’m not sure I fully understand. Are you asking me to update "ALL" the docstrings for MaskDINO? I don’t think it’s necessary unless they’re missing. Afaik, all the docstrings are already in place. Could you clarify?

As far as I know, otx is following google style docstrings but I found some parts aren't (of course, I didn't check all parts).
https://github.com/openvinotoolkit/training_extensions/blob/develop/for_developers/contribution_guide.md#docstrings

For examples, it is not following the style.
https://github.com/openvinotoolkit/training_extensions/blob/86654867bb3a4f7b754207e32a25535bbe0dcaf3/src/otx/algo/instance_segmentation/utils/box_ops.py#L44-L51

About arguments, I don't know `memory` argument's role because there is no description. I think it needs to be updated.
https://github.com/openvinotoolkit/training_extensions/blob/86654867bb3a4f7b754207e32a25535bbe0dcaf3/src/otx/algo/instance_segmentation/heads/transformer_decoder/dino_decoder.py#L51-L61

Some parts are missed.
https://github.com/openvinotoolkit/training_extensions/blob/86654867bb3a4f7b754207e32a25535bbe0dcaf3/src/otx/algo/instance_segmentation/layers/batch_norm.py#L66-L77

Usually, args in docstrings have type hint, but they aren't. It needs to follow others.
https://github.com/openvinotoolkit/training_extensions/blob/86654867bb3a4f7b754207e32a25535bbe0dcaf3/src/otx/algo/instance_segmentation/losses/hungarian_matcher.py#L18-L27

I think it is your job to find these things and looking for them is a waste of reviewer's time.
Please let reviewers focus on what they need to focus on, just like you said 😊

eugene123tw:
> > @sungchul2 Thanks for reviewing! The first and third questions are pretty similar—essentially, there are just too many modules to refactor all at once. My intent with this PR is to make the candidate algorithm compatible with OTX (training, testing, prediction, export, optimizer), rather than solving everything in one go. Testing and refactoring will happen in the next stage.
> > As for the second question, I’m not sure I fully understand. Are you asking me to update "ALL" the docstrings for MaskDINO? I don’t think it’s necessary unless they’re missing. Afaik, all the docstrings are already in place. Could you clarify?
> 
> As far as I know, otx is following google style docstrings but I found some parts aren't (of course, I didn't check all parts). https://github.com/openvinotoolkit/training_extensions/blob/develop/for_developers/contribution_guide.md#docstrings
> 
> For examples, it is not following the style.
> 
> https://github.com/openvinotoolkit/training_extensions/blob/86654867bb3a4f7b754207e32a25535bbe0dcaf3/src/otx/algo/instance_segmentation/utils/box_ops.py#L44-L51
> 
> About arguments, I don't know `memory` argument's role because there is no description. I think it needs to be updated.
> 
> https://github.com/openvinotoolkit/training_extensions/blob/86654867bb3a4f7b754207e32a25535bbe0dcaf3/src/otx/algo/instance_segmentation/heads/transformer_decoder/dino_decoder.py#L51-L61
> 
> Some parts are missed.
> 
> https://github.com/openvinotoolkit/training_extensions/blob/86654867bb3a4f7b754207e32a25535bbe0dcaf3/src/otx/algo/instance_segmentation/layers/batch_norm.py#L66-L77
> 
> Usually, args in docstrings have type hint, but they aren't. It needs to follow others.
> 
> https://github.com/openvinotoolkit/training_extensions/blob/86654867bb3a4f7b754207e32a25535bbe0dcaf3/src/otx/algo/instance_segmentation/losses/hungarian_matcher.py#L18-L27
> 
> I think it is your job to find these things and looking for them is a waste of reviewer's time. Please let reviewers focus on what they need to focus on, just like you said 😊

@sungchul2 Thanks for the suggestion! I will make sure all docstrings comply with the current style. That said, there are a few areas where I have doubts. A lot of the existing code doesn’t align with the standards you mentioned. While I understand your point, the current state of the repository has inconsistent documentation policies, making it unclear how to fully address your concerns.

For example, in [rtdetr_decoder.py](https://github.com/openvinotoolkit/training_extensions/blob/develop/src/otx/algo/detection/heads/rtdetr_decoder.py#L360), there’s no clear documentation on memory handling, and `variable memory_level_start_index` isn’t even used. 

Similarly, in [_build_model](https://github.com/openvinotoolkit/training_extensions/blob/develop/src/otx/algo/detection/yolov9.py#L76), the function lacks complete documentation.

I’m not pointing fingers; it’s just that unclear code and sparse documentation exist throughout the codebase. If we aim to uphold a certain standard, we should apply it uniformly, whether it’s for past, present, or future PRs. The process isn’t perfect right now but progress will come.

src/otx/algo/instance_segmentation/maskdino.py
```diff
@@ -0,0 +1,517 @@
...
+from otx.algo.instance_segmentation.utils.utils import ShapeSpec
+from otx.core.data.entity.base import ImageInfo, OTXBatchLossEntity
+from otx.core.data.entity.instance_segmentation import InstanceSegBatchDataEntity, InstanceSegBatchPredEntity
+from otx.core.exporter.base import OTXModelExporter
+from otx.core.exporter.native import OTXNativeModelExporter
+from otx.core.model.instance_segmentation import ExplainableOTXInstanceSegModel
+from otx.core.utils.mask_util import polygon_to_bitmap
+
+
+class MaskDINO(nn.Module):
```
kprokofi, line 32: We need to move it into "segmentor" folder

src/otx/algo/instance_segmentation/maskdino.py
```diff
@@ -0,0 +1,517 @@
...
+
+
+class MaskDINOR50(ExplainableOTXInstanceSegModel):
+    """OTX MaskDINO model with ResNet50 backbone."""
+
+    load_from = "https://github.com/IDEA-Research/detrex-storage/releases/download/maskdino-v0.1.0/maskdino_r50_50ep_300q_hid2048_3sd1_instance_maskenhanced_mask46.3ap_box51.7ap.pth"
+    mean = (123.675, 116.28, 103.53)
+    std = (58.395, 57.12, 57.375)
+
+    def _build_model(self, num_classes: int) -> nn.Module:
```
kprokofi, line 312: Can we use common "model_name" + builder approach?

src/otx/algo/instance_segmentation/utils/box_ops.py
```diff
@@ -0,0 +1,96 @@
...
+
+"""Utilities for bounding box manipulation and GIoU."""
+from __future__ import annotations
+
+import torch
+from torch import Tensor
+from torchvision.ops.boxes import box_area
+
+
+def box_cxcywh_to_xyxy(x: Tensor) -> Tensor:
```
kprokofi, line 13: Can we use torchvision ops?

src/otx/algo/instance_segmentation/utils/box_ops.py
```diff
@@ -0,0 +1,96 @@
...
+    wh = (rb - lt).clamp(min=0)  # [N,M,2]
+    inter = wh[:, :, 0] * wh[:, :, 1]  # [N,M]
+
+    union = area1[:, None] + area2 - inter
+
+    iou = inter / (union + 1e-6)
+    return iou, union
+
+
+def generalized_box_iou(boxes1: Tensor, boxes2: Tensor) -> Tensor:
```
kprokofi, line 44: We also have implementation in OTX for that. Please, consider in the next PR

kprokofi (dismissed):
Thank you for your work. Let's make a refactoring in the next PRs

src/otx/algo/instance_segmentation/utils/box_ops.py
eugene123tw: yeah, I will check and remove duplication.

src/otx/algo/instance_segmentation/utils/box_ops.py
eugene123tw: Maybe, I will take a look the differences.

eunwoosh (approved):

sovrasov (approved):

sovrasov:
@eunwoosh @sungchul2 thanks for the discussion, let's wait for the next PR to continue there

PR #3985, opened 2024-09-26
Add mAP for multilabel classifciation

sovrasov:
### Summary

This PR adds publicly adopted metric to match results from papers


tests/unit/core/metrics/test_mlc_map.py
```diff
@@ -0,0 +1,31 @@
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
+"""Unit tests for PCK metric."""
```
kprokofi: Could you rename? And also the class name

kprokofi: I think it was for key point detection

sovrasov: Done

kprokofi (approved):

eunwoosh (approved):

PR #4009, opened 2024-10-10
Relieve memory usage criteria on batch size 2 during adaptive_bs

eunwoosh:
### Summary
`adpative_bs` decides whether batch size is available conservatively. So even if some batch size doesn't make OOM, `adaptive_bs` can refuse to use it if it uses most of memory.
It's same to batch size 2, but it's better to try it if it doesn't make OOM because the only left thing to do is just raising an error if not.
So, this PR changes `adaptive_bs` to allow to return batch size 2 if it doesn't make OOM and use most of memory.

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/engine/adaptive_bs/bs_search_algo.py
+++ b/src/otx/engine/adaptive_bs/bs_search_algo.py
@@ -112,8 +112,14 @@ class BsSearchAlgo:
                 break
 
         if available_bs == 0:
-            msg = "Current device can't train model even with 2."
-            raise RuntimeError(msg)
+            if oom:
+                msg = "Current device can't train model even with 2."
+                raise RuntimeError(msg)
+            logger.warning(
+                "Even with a batch size of 2, most of the memory is used, "
+                "which could cause the training to fail midway.",
+            )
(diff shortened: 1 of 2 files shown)


chuneuny-emily (dismissed):

kprokofi (approved):

sungchul1 (approved):
LGTM, I have a question.
Why is the minimum batch size set to 2? Is it because batch normalization which can be in the model is considered? If so, it also looks okay to set batch size = 1 for the models that don't have batch normalization, what do you think?

yunchu:
@eunwoosh could you make same PR for the releases/2.3.0?

eunwoosh:
> LGTM, I have a question. Why is the minimum batch size set to 2? Is it because batch normalization which can be in the model is considered? If so, it also looks okay to set batch size = 1 for the models that don't have batch normalization, what do you think?

Adaptive batch size only uses even number, and this is why `adapt_bs` uses 2 as minimum value. I think we can consider to use batch size 1 as you said, if there is no problem. I'll check it later.

PR #4042, opened 2024-10-17
Fix multilabel_accuracy of MixedHLabelAccuracy

harimkang:
### Summary

https://jira.devtools.intel.com/browse/CVS-155228

https://github.com/Lightning-AI/torchmetrics/blob/6377aa5b6fe2863761839e6b8b5a857ef1b8acfa/src/torchmetrics/functional/classification/stat_scores.py#L583-L584

Due to the above code, an error is thrown if `num_multilabel_classes` is 1. To prevent this. (if num_multilabel_classes == 1, then use binary classification for that)

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/core/metrics/accuracy.py
+++ b/src/otx/core/metrics/accuracy.py
@@ -288,7 +288,9 @@ class MixedHLabelAccuracy(Metric):
         ]
 
         # Multilabel classification accuracy metrics
-        if self.num_multilabel_classes > 0:
+        # https://github.com/Lightning-AI/torchmetrics/blob/6377aa5b6fe2863761839e6b8b5a857ef1b8acfa/src/torchmetrics/functional/classification/stat_scores.py#L583-L584
+        # MultilabelAccuracy is available when num_multilabel_classes is greater than 2.
+        if self.num_multilabel_classes > 1:
             self.multilabel_accuracy = TorchmetricMultilabelAcc(
                 num_labels=self.num_multilabel_classes,
                 threshold=0.5,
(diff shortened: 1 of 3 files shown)


src/otx/core/metrics/accuracy.py
```diff
@@ -303,7 +305,7 @@ def _apply(self, fn: Callable, exclude_state: Sequence[str] = "") -> nn.Module:
             )
             for acc in self.multiclass_head_accuracy
         ]
-        if self.num_multilabel_classes > 0:
+        if hasattr(self, "multilabel_accuracy"):
```
eunwoosh: How about using `self.multilabel_accuracy == None` instead of it?

eunwoosh (commented):
Thanks for your work :) Could you add unit test for this case?

src/otx/core/metrics/accuracy.py
harimkang: Is it okay to ask what the reason for this is? Is that a better way?

eunwoosh: Almost same but a little bit better thing to use it is that `self.multilabel_accuracy == None` is more explicit so it can deal with a case that attribute 'num_multilabel_classes' is set my mistake. If you want to leave it, you can ignore this comment :)

harimkang: i'll update it. you meant `if self.multilabel_accuracy is not None` right?

eunwoosh: Thanks. yes and also need to add `self.multilabel_accuracy = None` in the `__init__()`.

harimkang: Thanks! Done - https://github.com/openvinotoolkit/training_extensions/pull/4042/commits/f055880f5c8ce01d604e1922e93f8563c3a0f904

eunwoosh (approved):

chuneuny-emily (approved):

PR #4046, opened 2024-10-17
Move transforms from 3d object detection dataset

kprokofi:
### Summary

Move 

- [x] Move affine transforms including resize from the dataset to torchvision transforms
- [x] Remove kitti utils
- [x] PIL -> OpenCV, align image ratio
- [x]  Fix OV IR Inference, make it works without OTX pre-processing 

### Checklist

### License

```python
```


src/otx/core/model/detection_3d.py
```diff
@@ -397,7 +423,7 @@ def _customize_inputs(
         self,
         entity: Det3DBatchDataEntity,
     ) -> dict[str, Any]:
-        img_sizes = np.array([img_info.ori_shape for img_info in entity.imgs_info])
+        img_sizes = np.array([img_info.ori_shape[::-1] for img_info in entity.imgs_info])
```
eunwoosh: Is there any reason why setting img_size to WxH?

src/otx/core/data/dataset/object_detection_3d.py
```diff
@@ -68,238 +55,55 @@ def __init__(
...
-            boxes_3d=torch.as_tensor(targets["boxes_3d"], dtype=torch.float32),
-            size_2d=torch.as_tensor(targets["size_2d"], dtype=torch.float32),
-            size_3d=torch.as_tensor(targets["size_3d"], dtype=torch.float32),
-            depth=torch.as_tensor(targets["depth"], dtype=torch.float32),
-            heading_angle=torch.as_tensor(
-                np.concatenate([targets["heading_bin"], targets["heading_res"]], axis=1),
-                dtype=torch.float32,
-            ),
+            boxes=np.zeros((self.max_objects, 4), dtype=np.float32),
+            labels=np.zeros((self.max_objects), dtype=np.int8),
```
eunwoosh, line 88: `Det3DDataEntity` get torch but numpy as passed. If numpy needs to be passed, then could you update type hint of `Det3DDataEntity`? And could you revmoe `# mypy: ignore-errors` in this file?

src/otx/core/data/transform_libs/torchvision.py
```diff
@@ -3449,6 +3449,304 @@ def __repr__(self) -> str:
         return repr_str
 
 
+class Decode3DInputsAffineTransforms(TopdownAffine):
```
eunwoosh, line 3452: Could you add unit test for this class?

eunwoosh (commented):

src/otx/core/model/detection_3d.py
kprokofi: I will change it to use HxW

kprokofi: forward function and outputs post processing expect WxH image ratio. I will try to change this though

src/otx/core/data/dataset/object_detection_3d.py
kprokofi: Agree, but the problem that the decoding transform and all related functions expect numpy passed. However, the Det3DDataEntity intended to be with torch tensors for training. I will try to change this behavior somehow

src/otx/core/data/transform_libs/torchvision.py
kprokofi: Sure, will do

kprokofi: Done https://github.com/openvinotoolkit/training_extensions/pull/4046/commits/ae60b71090bdc8f7e00776055de3f2c7e9684831

src/otx/core/data/dataset/object_detection_3d.py
kprokofi: Done https://github.com/openvinotoolkit/training_extensions/pull/4046/commits/ae60b71090bdc8f7e00776055de3f2c7e9684831

src/otx/core/model/detection_3d.py
kprokofi: Aligned to HxW everywhere https://github.com/openvinotoolkit/training_extensions/pull/4046/commits/ae60b71090bdc8f7e00776055de3f2c7e9684831

src/otx/core/model/detection_3d.py
```diff
@@ -583,7 +609,7 @@ def _convert_pred_entity_to_compute_metric(
         .numpy()
     )
 
-    img_sizes = np.array([img_info.ori_shape for img_info in inputs.imgs_info])
+    img_sizes = np.array([img_info.ori_shape for img_info in inputs.imgs_info])  # HxW -> WxH
```
eunwoosh:
```suggestion
    img_sizes = np.array([img_info.ori_shape for img_info in inputs.imgs_info])
```

eunwoosh (commented):
Thanks for applying comment :) I left a minor comment. Could you check it?

src/otx/core/data/transform_libs/torchvision.py
```diff
@@ -3449,6 +3449,333 @@ def __repr__(self) -> str:
...
+    @staticmethod
+    def rect_to_img(p2: np.ndarray, pts_rect: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
+        """Convert camera coordinates to image coordinates.
+
+        Args:
+            p2 (np.ndarray): Projection matrix with shape (3, 4).
+            pts_rect (np.ndarray): Rectangular coordinates with shape (N, 4).
+
+        Returns:
+            np.ndarray: Image coordinates with shape (N, 2).
```
sungchul1:
```suggestion
            tuple[np.ndarray, np.ndarray]: Image coordinates with shape (N, 2).
```

src/otx/core/data/transform_libs/torchvision.py
sungchul1: Nevertheless, most edge cases are not converted. Aren't unit tests needed to cover them?

src/otx/core/data/transform_libs/torchvision.py
```diff
@@ -3449,6 +3449,333 @@ def __repr__(self) -> str:
...
+            msg = "angle not in 0 ~ 2pi"
+            raise ValueError(msg)
+
+        angle_per_class = 2 * np.pi / float(num_heading_bin)
+        shifted_angle = (angle + angle_per_class / 2) % (2 * np.pi)
+        class_id = int(shifted_angle / angle_per_class)
+        residual_angle = shifted_angle - (class_id * angle_per_class + angle_per_class / 2)
+        return class_id, residual_angle
+
+    def convert(self, inputs: T_OTXDataEntity, mask_2d: np.ndarray, image_to_tensor: bool = False) -> T_OTXDataEntity:  # type: ignore[override]
```
sungchul1, line 3765: Isn't it needed to be adapted in mixin for other transforms for 3d detection can be added in the future?
Or leaving a comment looks good.

src/otx/core/data/transform_libs/torchvision.py
sungchul1: And this transform seems to have multiple functionality, I think splitting it (decode and affine transform) looks better if possible. What do you think?

sungchul1: And could you explain the reason for inheriting `TopdownAffine`? I didn't find the clue.
Why I ask it is that if `convert` is updated here and there is no reason to inherit `TopdownAffine`, it can inherit `tvt_v2.Transform` only.

src/otx/core/data/transform_libs/torchvision.py
kprokofi: This transform is used only for 3d detection task, I don't think it should be updated in mixin

src/otx/core/data/transform_libs/torchvision.py
kprokofi: This transform has multiple functionality because it is hard to divide transforms from decoding KITTI annotations for 3d detection training. The annotations should use info from affine transforms. I created this unified transform to move resizing, cropping and flipping from __get_item__ in the Dataset. 

This transform inherits from TopdownAffine, because it uses 2 its main methods, "_get_warp_matrix" and " _get_warp_image"

sungchul1: Thanks for the explanation. If so, I agree with your decision.

kprokofi: Added more unit tests: https://github.com/openvinotoolkit/training_extensions/pull/4046/commits/23d767dcc23942efb027e24f3872010542530abb

eunwoosh (approved):

sovrasov (approved):

PR #4039, opened 2024-10-16
Turn on/off classification augmentations

goodsong81:
### Summary

* Add 'enable' flag for classification recipe transforms (default=True)
* Add common augmentations (disabled by default for backward compatibility)
```python
configurable_augs = [
    "PhotoMetricDistortion",
    "RandomAffine",
    "RandomVerticalFlip",
    "GaussianBlur",
    "GaussianNoise",
]
```
* Add aug on/off combination tests for all classification recipes

### How to test

```console
pytest -vs tests/integration/api/test_augmentation.py
```

### Checklist

- [x] I have added unit tests to cover my changes.​
- [x] I have added integration tests to cover my changes.​
- [ ] I have ran e2e tests and there is no issues.
- [x] I have added the description of my changes into CHANGELOG in my target branch (e.g., [CHANGELOG](https://github.com/openvinotoolkit/training_extensions/blob/develop/CHANGELOG.md) in develop).​
- [ ] I have updated the documentation in my target branch accordingly (e.g., [documentation](https://github.com/openvinotoolkit/training_extensions/tree/develop/docs) in develop).
- [ ] I have [linked related issues](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

### License

- [x] I submit _my code changes_ under the same [Apache License](https://github.com/openvinotoolkit/training_extensions/blob/develop/LICENSE) that covers the project.
      Feel free to contact the maintainers if that's a concern.
- [x] I have updated the license header for each file (see an example below).

```python
# Copyright (C) 2024 Intel Corporation
# SPDX-License-Identifier: Apache-2.0
```

PR #4066, opened 2024-10-24
Turn on/off augmentations for detection and instance segmentation

GalyaZalesskaya:
### Summary

2nd PR with implementation for Detection and Instance Segmentation.
1st PR for classification is here [PR#4039](https://github.com/openvinotoolkit/training_extensions/pull/4039).

- Replaced `torchvision.transforms.v2.RandomPhotometricDistort` for `rtdetr` models with configurable `otx.core.data.transform_libs.torchvision.PhotoMetricDistortion` (to switch on and off it via UI) with changed `hue_delta` parameter to align implementation
- Moved heavy operations to run on resized or resized-cropped resolution
- Add common augmentations (disabled by default for backward compatibility)
```
configurable_augs = [
    "PhotoMetricDistortion",
    "RandomAffine",
    "RandomVerticalFlip",
    "GaussianBlur",
    "GaussianNoise",
]
```
- Currently some YOLOX models have both configurable `PhotoMetricDistortion` (to switch on/off via UI) and original `YOLOXHSVRandomAug` with the close implementation -> to be fixed later

### Checklist

### License

```python
```


src/otx/recipe/_base_/data/detection.yaml
```diff
@@ -16,6 +16,10 @@ train_subset:
   to_tv_image: false
   transforms:
     - class_path: otx.core.data.transform_libs.torchvision.MinIoURandomCrop
+    - class_path: otx.core.data.transform_libs.torchvision.PhotoMetricDistortion
+      enable: false
+    - class_path: otx.core.data.transform_libs.torchvision.RandomAffine
+      enable: false
```
goodsong81: What do you think of moving these down to the spot just after the "Resize"?
Even after crop, working in the original resolution (like 4096) might be inefficient.
Same for other recipes.

src/otx/recipe/detection/yolov9_c.yaml
```diff
@@ -45,6 +45,8 @@ overrides:
     train_subset:
       batch_size: 10
       transforms:
+        - class_path: otx.core.data.transform_libs.torchvision.PhotoMetricDistortion
+          enable: false
```
goodsong81: This might replace YOLOXHSV with appropriate params. Let's put it close to each other, and replace later.

src/otx/recipe/detection/yolox_tiny.yaml
```diff
@@ -61,6 +61,13 @@ overrides:
         - class_path: otx.core.data.transform_libs.torchvision.RandomFlip
           init_args:
             prob: 0.5
+            is_numpy_to_tvtensor: true
+        - class_path: torchvision.transforms.v2.RandomVerticalFlip
+          enable: false
+        - class_path: torchvision.transforms.v2.GaussianBlur
+          enable: false
+          init_args:
+            kernel_size: 5
```
goodsong81: This will result in additional numpy->tvtensor conversion.
Why don't you insert these tv2 transforms after "Pad"?

src/otx/recipe/detection/rtdetr_101.yaml
```diff
@@ -58,15 +58,25 @@ overrides:
...
         - class_path: otx.core.data.transform_libs.torchvision.RandomFlip
           init_args:
             prob: 0.5
+            is_numpy_to_tvtensor: true
+        - class_path: torchvision.transforms.v2.RandomVerticalFlip
+          enable: false
+        - class_path: torchvision.transforms.v2.GaussianBlur
+          enable: false
+          init_args:
+            kernel_size: 5
```
goodsong81: We can move these to after "Resize".

src/otx/recipe/detection/rtmdet_tiny.yaml
```diff
@@ -44,6 +44,10 @@ overrides:
     train_subset:
       batch_size: 8
       transforms:
+        - class_path: otx.core.data.transform_libs.torchvision.PhotoMetricDistortion
+          enable: false
+        - class_path: otx.core.data.transform_libs.torchvision.RandomAffine
+          enable: false
```
goodsong81: Maybe after "RandomCrop"?

goodsong81 (commented):
Thank you for the work! Please see my comments below.
In overall, a little bit of reordering might be needed for efficiency.
* Minimize numpy <-> tvtensor conversion
* Run heavy operations in resized or resized-cropped resolution

src/otx/recipe/_base_/data/detection.yaml
GalyaZalesskaya: Thank you for your suggestions, applied them

harimkang (approved):
LGTM!
@sungchul2 @kprokofi @eugene123tw Could you review this PR?

src/otx/recipe/_base_/data/detection.yaml
```diff
@@ -20,17 +20,29 @@ train_subset:
...
+      init_args:
+        kernel_size: 5
     - class_path: torchvision.transforms.v2.ToDtype
       init_args:
         dtype: ${as_torch_dtype:torch.float32}
     - class_path: torchvision.transforms.v2.Normalize
       init_args:
         mean: [0.0, 0.0, 0.0]
         std: [255.0, 255.0, 255.0]
+    - class_path: torchvision.transforms.v2.GaussianNoise
```
sungchul1, line 44: Just a question, is it okay that it is the next of `Normalize` because of N(0, 1)?
https://pytorch.org/vision/stable/generated/torchvision.transforms.v2.GaussianNoise.html#torchvision.transforms.v2.GaussianNoise

src/otx/recipe/detection/yolov9_c.yaml
```diff
@@ -65,15 +65,24 @@ overrides:
...
             max_cached_images: 10
-        - class_path: otx.core.data.transform_libs.torchvision.YOLOXHSVRandomAug
         - class_path: otx.core.data.transform_libs.torchvision.Resize
           init_args:
             scale: $(input_size)
             keep_ratio: true
             transform_bbox: true
+        - class_path: otx.core.data.transform_libs.torchvision.PhotoMetricDistortion
+          enable: false
+        - class_path: otx.core.data.transform_libs.torchvision.YOLOXHSVRandomAug
```
sungchul1, line 75: I think this order change won't affect performance, but did you check it?

sungchul1 (commented):
LGTM

sungchul1 (approved):

eugene123tw (approved):

src/otx/recipe/_base_/data/detection.yaml
GalyaZalesskaya: I followed the pattern from the classification task ([PR#4039](https://github.com/openvinotoolkit/training_extensions/pull/4039)) where we add noise after normalization. 
It seems to me that gaussian noise with (0,0.1) after normalization will add the additional regularization. While the other way around it seems less meaningful.
@goodsong81 Do you have an opinion about the order of normalization-noise augmentations?

src/otx/recipe/detection/yolov9_c.yaml
GalyaZalesskaya: No, I didn't conduct the performance measurements. I just followed the idea of applying transformations on resized images to decrease (theoretically and on average) the performance overhead

src/otx/recipe/detection/rtdetr_101.yaml
```diff
@@ -58,26 +58,36 @@ overrides:
       batch_size: 4
       to_tv_image: true
       transforms:
-        - class_path: torchvision.transforms.v2.RandomPhotometricDistort
-          init_args:
-            p: 0.5
         - class_path: torchvision.transforms.v2.RandomZoomOut
           init_args:
             fill: 0
-        - class_path: otx.core.data.transform_libs.torchvision.RandomFlip
```
kprokofi, line 67: Why was it deleted?

src/otx/recipe/detection/rtmdet_tiny.yaml
```diff
@@ -60,10 +60,21 @@ overrides:
...
+        - class_path: otx.core.data.transform_libs.torchvision.PhotoMetricDistortion
+          enable: false
         - class_path: otx.core.data.transform_libs.torchvision.YOLOXHSVRandomAug
         - class_path: otx.core.data.transform_libs.torchvision.RandomFlip
           init_args:
             prob: 0.5
+            is_numpy_to_tvtensor: true
+        - class_path: torchvision.transforms.v2.RandomVerticalFlip
```
kprokofi, line 72: I still think that it is better to use horizontal flip, since vertical is not natural and not common in the CV field

src/otx/recipe/_base_/data/detection.yaml
goodsong81: Actually, there is an "clip 0~1" option which is default in GaussianNoise op.
So, it should be false for standardized (0-mean 1-var) and true for normalized (0-min 1-max) images. I'll fix in the next PRs. (If we are allowed to do it! ;])

src/otx/recipe/detection/rtmdet_tiny.yaml
goodsong81: It's a misunderstanding. We keep h-flip as default as-is, and adding v-flip as optional aug,
and it's been included both in Roboflow & Landing.AI apps as well.

src/otx/recipe/detection/rtdetr_101.yaml
GalyaZalesskaya: It wasn't deleted, it was moved to L74 to minimize the number of numpy (otx augs) <-> tvtensor (torchvision augs) conversions

kprokofi (approved):

harimkang (approved):

PR #4068, opened 2024-10-25
Update benchmark datasets

chuneuny-emily:
### Summary
The purpose of this PR is for aligning perf test code to the dataset really used.
Current change has been used for OTX benchmark on CUDA. Because CI has been broken, these datasets have been used locally.

### Checklist

### License

```python
```

### DIFF
--- a/tests/perf/test_detection.py
+++ b/tests/perf/test_detection.py
@@ -47,8 +47,8 @@ class TestPerfObjectDetection(PerfTestBase):
 
     DATASET_TEST_CASES = [
         Benchmark.Dataset(
-            name=f"pothole_small_{idx}",
-            path=Path("detection/pothole_small") / f"{idx}",
+            name=f"bdd_small_{idx}",
+            path=Path("detection/bdd_small") / f"{idx}",
             group="small",
             num_repeat=5,
             extra_overrides={
@@ -60,8 +60,8 @@ class TestPerfObjectDetection(PerfTestBase):
         for idx in (1, 2, 3)
     ] + [
         Benchmark.Dataset(
-            name="pothole_medium",
-            path=Path("detection/pothole_medium"),
+            name="bdd_medium",
+            path=Path("detection/bdd_medium"),
             group="medium",
             num_repeat=5,
             extra_overrides={
@@ -71,8 +71,8 @@ class TestPerfObjectDetection(PerfTestBase):
             },
         ),
         Benchmark.Dataset(
-            name="vitens_large",
-            path=Path("detection/vitens_large"),
+            name="bdd_large",
+            path=Path("detection/bdd_large"),
             group="large",
             num_repeat=5,
             extra_overrides={
--- a/tests/perf/test_semantic_segmentation.py
+++ b/tests/perf/test_semantic_segmentation.py
@@ -37,15 +37,15 @@ class TestPerfSemanticSegmentation(PerfTestBase):
         for idx in (1, 2, 3)
     ] + [
         Benchmark.Dataset(
-            name="kvasir_medium",
-            path=Path("semantic_seg/kvasir_medium"),
+            name="cityscapes_185_70_medium",
+            path=Path("semantic_seg/cityscapes_185_70_medium"),
             group="medium",
             num_repeat=5,
             extra_overrides={},
         ),
         Benchmark.Dataset(
-            name="kvasir_large",
-            path=Path("semantic_seg/kvasir_large"),
+            name="voc_2012_cut_large",
+            path=Path("semantic_seg/voc_2012_cut_large"),
             group="large",
             num_repeat=5,
             extra_overrides={},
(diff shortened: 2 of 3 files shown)


tests/perf/test_semantic_segmentation.py
```diff
@@ -102,46 +102,46 @@ class TestPerfSemanticSegmentationSemiSL(TestPerfSemanticSegmentation):
...
 
     DATASET_TEST_CASES = [  # noqa: RUF012
         Benchmark.Dataset(
             name="kvasir",
-            path=Path("semantic_seg/semisl/kvasir_24"),
+            path=Path("semantic_seg/semi-sl/kvasir_24"),
             group="small",
-            num_repeat=5,
-            unlabeled_data_path=Path("semantic_seg/semisl/unlabeled_images/kvasir"),
+            num_repeat=1,
```
eunwoosh, line 119: Why is it set to 1?

eunwoosh (commented):
Thanks for update. I left a minor comment. Please take a look.

tests/perf/test_semantic_segmentation.py
chuneuny-emily: All semisl has run_repeat=1, if not, only semisl testing will take over 10 days.

tests/perf/test_detection.py
sungchul1, line 1: pothole and vitens are single class datasets, but bdd is a multi class dataset.
Don't we need to metric performance on single class datasets?

tests/perf/test_semantic_segmentation.py
sungchul1: I think, at least small datasets, increasing num_repeat is needed to check stable performance even though it takes a time.

sungchul1 (commented):
Isn't there any way to do parallel tests per e2e per model? If so, it can reduce the time effectively. Of course e2e time can increase a little, but other tasks are on the same circumstance and we can compare the ratio between e2e times.

tests/perf/test_semantic_segmentation.py
chuneuny-emily: Even small datasets take a long time. There are 3 splits for small dataset, so if we go around 5 times, it will actually go around 15 times, and the execution time for semi-sl will already exceed the general benchmark test execution time.

tests/perf/test_detection.py
chuneuny-emily: Based on Mark's previous feedback that detection performance benchmarks should be avoided as much as possible with a single-class dataset, the detection dataset has been replaced with bdd.

chuneuny-emily:
> Isn't there any way to do parallel tests per e2e per model? If so, it can reduce the time effectively. Of course e2e time can increase a little, but other tasks are on the same circumstance and we can compare the ratio between e2e times.

Since the benchmark looks not only at the accuracy but also at the e2e time, it is difficult to perform performance tests on multiple models at the same time. In principle, we are conducting single model tests on a single GPU.

chuneuny-emily:
The purpose of this PR is to aim to be measured on the same dataset/conditions as the existing benchmarks when performing benchmarks with develop branch on arc GPUs, so if you have any other comments that don't meet this purpose, please let us know via Teams chat.

tests/perf/test_detection.py
kprokofi: I don't think that measuring on one dataset is a good idea. We need to consider other datasets and use cases. Ideally, we need 3 different datasets. We cannot obtain fair validation result of the model. We can only compare some numbers from one version to another. I know that it is the main purpose of the perf tests, but we use this benchmark system also for new models assessment. It will be better to use versatile or diverse datasets

kprokofi (approved):

tests/perf/test_semantic_segmentation.py
kprokofi: Can we keep only 1 dataset then?  It is not our priority and it uses same models as in supervised.

eunwoosh (commented):
As Emily said, it's for aligning perf code to real dataset used, so let's discuss about how to improve our perf test later.

eunwoosh (approved):

sungchul1:
> The purpose of this PR is to aim to be measured on the same dataset/conditions as the existing benchmarks when performing benchmarks with develop branch on arc GPUs, so if you have any other comments that don't meet this purpose, please let us know via Teams chat.

Could you update this purpose in the summary above?
And I think if it is just for Arc GPU, it must be set as an option and original tests should remain. If removing them, we cannot compare the performance with the past and the present anymore.

eunwoosh:
> > The purpose of this PR is to aim to be measured on the same dataset/conditions as the existing benchmarks when performing benchmarks with develop branch on arc GPUs, so if you have any other comments that don't meet this purpose, please let us know via Teams chat.
> 
> Could you update this purpose in the summary above? And I think if it is just for Arc GPU, it must be set as an option and original tests should remain. If removing them, we cannot compare the performance with the past and the present anymore.

Current change has been used for OTX benchmark on CUDA. Because CI has been broken, these datasets have been used locally. So, the purpose of this PR is for aligning perf test code to the dataset really used.

tests/perf/test_semantic_segmentation.py
chuneuny-emily: @kprokofi - Isn't it better to measure performance once on 3 different splits of a dataset and average the results, rather than measuring performance 3 times on a single split, for the sake of diversity? When I previously discussed classification semi-sl datasets with Harim, I believe Harim is on the same page. Can you discuss this topic again with Harim in the OTX production group and let me know which approach is more efficient?

Issue #3971, opened 2024-09-20
Feature Request:  Add support for EfficientAD from Anomalib


j99ca:
Currently there are only 2 supported anomaly models in OTX:  PADIM and STFPM.

EfficientAD is a newer model that that Anomalib supports that has very high accuracy according to papers with code as well as my own internal testing for industrial use-cases. This would be a great model to include in OTX's anomaly detection flow, or perhaps tighter integration between OTX and Anomalib in general would be ideal.

harimkang:
I personally would like to see all of Anomalib's models evolve to be commonly used in OTX, just like Timm or hugging-face models are supported by wrapper classes. But as far as I know, this needs some cleanup at the moment. But I think it's achievable.
@ashwinvaidya17

ashwinvaidya17:
Yeah. That's the plan. We are going to slowly all support for all the algorithms available in Anomalib

PR #4102, opened 2024-11-06
🧹 Refactor Anomaly Models

ashwinvaidya17:
### Checklist

### License

```python
```

### DIFF
--- a/src/otx/core/model/anomaly.py
+++ b/src/otx/core/model/anomaly.py
@@ -14,6 +14,7 @@ from anomalib.callbacks.post_processor import _PostProcessorCallback
 from anomalib.callbacks.thresholding import _ThresholdCallback
 from torch import nn
 
+from otx import __version__
 from otx.core.data.entity.anomaly import (
     AnomalyClassificationBatchPrediction,
     AnomalyClassificationDataBatch,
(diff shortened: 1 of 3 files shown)


sovrasov:
@ashwinvaidya17 would that be compatible with Geti?

harimkang:
> @ashwinvaidya17 would that be compatible with Geti?

As far as I can see, there are no input changes, it's mainly a refactoring inside the model, so it probably won't affect Geti.
@ashwinvaidya17 Could you confirm?

harimkang (commented):
Looks good, please take a look at the unit-test fail.

harimkang:
+ @ashwinvaidya17 Another question, when will we be able to upgrade to the latest version of anomalib in otx?

We also have the following request from a user, Could you please check it out?
https://github.com/openvinotoolkit/training_extensions/issues/3971

ashwinvaidya17:
Yeah. Only the methods have been moved around a bit.

src/otx/core/model/anomaly.py
```diff
@@ -157,6 +155,22 @@ def configure_callbacks(self) -> list[Callback]:
             ),
         ]
 
+    def on_validation_epoch_start(self) -> None:
+        """Don't call OTXModel's ``on_validation_epoch_start``."""
+        return
```
sovrasov, line 160: Does it make sense to raise an error here, since this method should never be called?

ashwinvaidya17: Some callbacks in check if `on_validation_epoch_start` has been overridden, and if so, run the required step. But in this case it might not be necessary. I'll revisit again.

harimkang (approved):

sovrasov (approved):

PR #4106, opened 2024-11-08
Add dummy XAI to RTDETR (export mode) & disable strong aug

eugene123tw:
### Summary

* Disable strong aug in RTDETR recipes
* Add dummy tensors XAI for RTDETR export mode (PyTorch mode still broken)

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algo/detection/base_models/detection_transformer.py
+++ b/src/otx/algo/detection/base_models/detection_transformer.py
@@ -5,6 +5,7 @@
 
 from __future__ import annotations
 
+import warnings
 from typing import Any
 
 import numpy as np
@@ -95,15 +96,21 @@ class DETR(BaseModule):
         explain_mode: bool = False,
     ) -> dict[str, Any] | tuple[list[Any], list[Any], list[Any]]:
         """Exports the model."""
-        if explain_mode:
-            msg = "Explain mode is not supported for DETR models yet."
-            raise NotImplementedError(msg)
-
-        return self.postprocess(
+        results = self.postprocess(
             self._forward_features(batch_inputs),
             [meta["img_shape"] for meta in batch_img_metas],
             deploy_mode=True,
         )
+
+        if explain_mode:
+            # TODO(Eugene): Implement explain mode for DETR model.
+            warnings.warn("Explain mode is not supported for DETR model. Return dummy values.", stacklevel=2)
+            xai_output = {
+                "feature_vector": torch.zeros(1, 1),
+                "saliency_map": torch.zeros(1),
+            }
+            results.update(xai_output)  # type: ignore[union-attr]
+        return results
 
     def postprocess(
         self,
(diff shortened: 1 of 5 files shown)


src/otx/algo/detection/base_models/detection_transformer.py
```diff
@@ -95,16 +96,22 @@ def export(
...
-            raise NotImplementedError(msg)
-
-        return self.postprocess(
+        results = self.postprocess(
             self._forward_features(batch_inputs),
             [meta["img_shape"] for meta in batch_img_metas],
             deploy_mode=True,
         )
 
+        if explain_mode:
```
kprokofi, line 105: Is it agreed with IAI team? On the Geti side model will return dummy values

src/otx/recipe/detection/rtdetr_50.yaml
```diff
@@ -53,10 +53,11 @@ overrides:
       transforms:
         - class_path: torchvision.transforms.v2.RandomPhotometricDistort
           init_args:
-            p: 0.5
+            p: 0.0
```
kprokofi: Why did you switch off the RandomPhotometricDistort and RandomZoomOut?

kprokofi: I think it can hurt the generalization of the model. We don't have augmentations now except for Flip.

src/otx/algo/detection/base_models/detection_transformer.py
sovrasov: That's fine for now, XAI for the new models is not expected, however, we have to generate stubs to make sure Geti inference won't fail.

src/otx/recipe/detection/rtdetr_50.yaml
eugene123tw: @kprokofi The accuracy is very bad (almost 0%) on several dataset I benchmarked. This is due to strong augmentation applied after the augmentation PR.

eugene123tw: I think we can keep RandomPhotometricDistort, but RandomZoomOut might over-augment the image.

sovrasov: @eugene123tw did you run a quick training on a medium to check is the new setting are ok?

kprokofi: I agree, we can remove RandomZoomOut, but RandomPhotometricDistort should not influence so much, it is very common augmentation, almost basic one. We can reduce probability though. Also, if you want to switch off the augmentation, could you remove it at all from data recipe? I don't think it makes sense to keep it in recipe with probability = 0

eugene123tw: @sovrasov I ran it on pothole_medium
test/f1-score with both aug = 0.5822
test/f1-score disable both   =   0.4100
test/f1-score with RandomPhotometricDistort = 0.6271.

I think it's safer just to keep `RandomPhotometricDistort`, and disable zoom in and out

sovrasov (approved):

kprokofi (approved):

PR #4113, opened 2024-11-12
Export static shape for detection/instance-segmentation models

### LINKED ISSUE #4111: ONNX dynamic shaping error in OpenCV embedding shape interpolate
**Describe the bug**
ONNX dynamic shaping not working with Model API `inference_adapter.embed_preprocessing`.

```
[CPU] Interpolate node with name 'Interpolate_65093' Check 'cmp::le(m_min, u) && cmp::le(u, m_max)' failed at src/core/shape_inference/include/shape_infer_type_utils.hpp:54:
Value -1 not in range [0:18446744073709551615]
```

**Steps to Reproduce**

1. `otx train --config src/otx/recipe/detection/atss_mobilenetv2.yaml --data_root tests/assets/car_tree_bug --work_dir otx-workspace/atss_mobilenetv2 `
2. `otx export  --work_dir otx-workspace/atss_mobilenetv2 --export_format ONNX`
3. ` otx test --work_dir otx-workspace/atss_mobilenetv2  --checkpoint otx-workspace/atss_mobilenetv2/.latest/export/exported_model.onnx`

**Environment:**

- OS: 
- Framework version: 
- Python version:
- OpenVINO version:
- CUDA/cuDNN version:
- GPU model and memory:

eugene123tw:
### Summary

Fix ONNX inference issue by using static shape for detection/inst-seg models.

### How to test

Added `test_onnx_export` unittest.

### Checklist

### License

```python
```


pyproject.toml
```diff
@@ -98,6 +99,7 @@ base = [
     "openvino-model-api==0.2.4",
     "onnx==1.17.0",
     "onnxconverter-common==1.14.0",
+    "onnxruntime==1.20.0",
```
sovrasov: Why do we need onnxruntime in basic OTX installation? It seems like it's only used in tests

tests/unit/algo/instance_segmentation/test_maskrcnn.py
```diff
@@ -82,3 +83,27 @@ def test_export(self, model):
...
+            output_dir=tmp_path,
+            base_name="model",
+            export_format=OTXExportFormatType.ONNX,
+        )
+
+        assert model_path.exists()
+
+        # TODO(Eugene): ONNX issue with ExperimentalDetectronROIFeatureExtractor
+        try:
+            ort.InferenceSession(model_path)
```
sovrasov: Other yet unknown failures can also be caught here, a targeted pytest.skip call would look more accurate here

sovrasov: Upd: I see that this is a MRCNN-specific test

src/otx/algo/detection/atss.py
```diff
@@ -183,7 +183,6 @@ def _exporter(self) -> OTXModelExporter:
                 "input_names": ["image"],
                 "output_names": ["boxes", "labels"],
                 "dynamic_axes": {
-                    "image": {0: "batch", 2: "height", 3: "width"},
```
sovrasov, line 186: I looks a bit weird that the batch dim is static on input, but dynamic on output. What is the actual outcome in terms of i/o shapes?

eugene123tw: OV does support dynamic input shapes; however, we have performance issues I think, which is why we currently fix the input shape. This could be made configurable in the future, there is no immediate use case afaik. For now, the static input shape is determined based on the input tensor during tracing. Regarding the output, it remains dynamic, upto the number of max pred.

pyproject.toml
eugene123tw: This is to ensure that ONNX model can be successfully loaded. Most detection models work with ONNX, so it makes sense to test it since we support ONNX export. Would you recommend adding an ONNX inference tests as well?

sovrasov: I see, but in general adding tests deps to the base installation is questionable. Could you move onnxrt to the test deps section?

src/otx/algo/detection/atss.py
sovrasov: https://github.com/openvinotoolkit/training_extensions/issues/4111 was related to OV preprocessor insertion failure: it could not deduct h w required to initialize the resize.
I think dynamic batch is ok here, since OV can set it to 1 on it's side.
Now, logically, the export config doesn't make sense to me: it says the model has fixed input batch dim, but dynamic output batch. OV eats that, because it does shape inference on it's side, but overall the configuration looks inconsistent: the batch axis should be either kept everywhere, or deleted from the config.

sovrasov (approved):

kprokofi (approved):

PR #4127, opened 2024-11-21
Update Label Info handling

sovrasov:
### Summary

- Fix multilabel data parsing
- Introduce label ids to OTX
- Recover label names from arrow (doesn't work for tiling)

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/core/data/dataset/base.py
+++ b/src/otx/core/data/dataset/base.py
@@ -70,6 +70,7 @@ class OTXDataset(Dataset, Generic[T_OTXDataEntity]):
         max_refetch: Maximum number of images to fetch in cache
         image_color_channel: Color channel of images
         stack_images: Whether or not to stack images in collate function in OTXBatchData entity.
+        data_format: Source data format, which was originally passed to datumaro (could be arrow for instance).
 
     """
 
--- a/src/otx/core/data/dataset/base.py
+++ b/src/otx/core/data/dataset/base.py
@@ -83,6 +83,7 @@ class OTXDataset(Dataset, Generic[T_OTXDataEntity]):
         image_color_channel: ImageColorChannel = ImageColorChannel.RGB,
         stack_images: bool = True,
         to_tv_image: bool = True,
+        data_format: str = "",
     ) -> None:
         self.dm_subset = dm_subset
         self.transforms = transforms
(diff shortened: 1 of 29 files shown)


src/otx/core/types/label.py
```diff
@@ -79,6 +85,38 @@ def from_dm_label_groups(cls, dm_label_categories: LabelCategories) -> LabelInfo
         return LabelInfo(
             label_names=label_names,
             label_groups=label_groups,
+            label_ids=[str(i) for i in range(len(label_names))],
+        )
+
+    @classmethod
+    def from_dm_label_groups_arrow(cls, dm_label_categories: LabelCategories) -> LabelInfo:
```
kprokofi, line 92: Could you please add unit test for this method?

src/otx/core/types/label.py
```diff
@@ -279,8 +317,60 @@ def convert_labels_if_needed(
...
+    @classmethod
+    def from_dm_label_groups_arrow(cls, dm_label_categories: LabelCategories) -> HLabelInfo:
+        """Generate HLabelData from the Datumaro LabelCategories. Arrow-specific implementation.
+
+        Args:
+            dm_label_categories (LabelCategories): the label categories of datumaro.
+        """
+        dm_label_categories = copy.deepcopy(dm_label_categories)
+
+        empty_label_name = None
```
kprokofi, line 332: I think we need to remove empty_label for other tasks also. Can you add similar approach to the base LabelInfo?

sovrasov: It's too dangerous for 2.6, I'm even going to reduce the set of changes in this PR. The idea of the PR is to delegate further filtering to task-specific datasets instead of trying to filter in one place. However, now I'm not sure if this approach is the best option. H-cls can not be filtered by datumaro's standard methods, so the related changes for h-labels would stay.

src/otx/core/data/dataset/base.py
```diff
@@ -83,6 +83,7 @@ def __init__(
         image_color_channel: ImageColorChannel = ImageColorChannel.RGB,
         stack_images: bool = True,
         to_tv_image: bool = True,
+        data_format: str = "",
```
eugene123tw, line 87: Could you add a docstring for `data_format`? (sorry for the nitpicking)

sovrasov: No problem, will add it when starting finalization of the PR

sovrasov: Done

src/otx/core/types/label.py
sovrasov: Added

eugene123tw (dismissed):

kprokofi (dismissed):

sovrasov:
Classification tests have passed locally:
pytest -v -s tests/integration/ -k cls
image
Had to apply some tricks in conftest.py as the memory consumption is huge. Iseg and unit tests are passed on CI indicating that the general LabelInfo changes don't break other components.

kprokofi (approved):

eugene123tw (approved):

PR #4131, opened 2024-11-25
Fix early stopping in converter patching + fix lr warmup for all tasks

kprokofi:
### Summary
Fix found bugs in OTX2.2 for OTX2.2.1 and Geti2.6 releases.

- [x] Correct EarlyStoppingWithWarmup overriding in converter instead of lightning early stopping. (I checked with Geti config that now it overrides early stopping correctly)
- [x] Add an interface for warming up scheduler in configs for all models. It should be possible to set warmup for every model. Next, warmup with 3 steps looks like a bug. This PR set it to 0 for now. (we need to search for better schedule later)
- [x] Change patience for ReduceLROnPlateue for classification task from 1 epoch to 5 (seems reasonable for me, 1 epoch looks very strange)
- [x] Increase patience for classification and align this default number across all current Geti tasks. (patience=3 are not reasonable and may hurt model convergence/performance due to random fluctuations.)
From Geti it is clearly seen, that sometimes more epochs are required to converge
image

- [x] Template update

image

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/recipe/rotated_detection/maskrcnn_efficientnetb2b.yaml
+++ b/src/otx/recipe/rotated_detection/maskrcnn_efficientnetb2b.yaml
@@ -13,7 +13,7 @@ model:
     scheduler:
       class_path: otx.core.schedulers.LinearWarmupSchedulerCallable
       init_args:
-        num_warmup_steps: 10
+        num_warmup_steps: 100
         main_scheduler_callable:
           class_path: lightning.pytorch.cli.ReduceLROnPlateau
           init_args:
--- a/src/otx/recipe/rotated_detection/maskrcnn_efficientnetb2b.yaml
+++ b/src/otx/recipe/rotated_detection/maskrcnn_efficientnetb2b.yaml
@@ -13,7 +13,7 @@ model:
     scheduler:
       class_path: otx.core.schedulers.LinearWarmupSchedulerCallable
       init_args:
-        num_warmup_steps: 100
+        num_warmup_steps: 10
         main_scheduler_callable:
           class_path: lightning.pytorch.cli.ReduceLROnPlateau
           init_args:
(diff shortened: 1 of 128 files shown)


sovrasov (dismissed):

sovrasov:
@kprokofi do you have any experiments to compare the modified schedule for classification?

eugene123tw (dismissed):

src/otx/recipe/rotated_detection/maskrcnn_efficientnetb2b.yaml
```diff
@@ -13,7 +13,7 @@ model:
     scheduler:
       class_path: otx.core.schedulers.LinearWarmupSchedulerCallable
       init_args:
-        num_warmup_steps: 100
```
eugene123tw: Hi @kprokofi could you revert the changes for rotated detection as well? Thanks

kprokofi: Done https://github.com/openvinotoolkit/training_extensions/pull/4131/commits/570014762261d4d02fa3e4667c4c1e587a28669c

kprokofi:
After benchmarking all models, I found, that indeed updated schedule can show up to 2% of accuracy gain. However, the training time has grown drastically. For large datasets, I see 2-2.5x training time. So, I will decrease patience for classification from 10 to 5 epochs and patience for scheduler from 5 to 3.

Unfortunately, we don't have small datasets to test it on the edge cases like Geti, we have only one small dataset used for CI, but the accuracy becomes 1 after only few epochs.

sovrasov:
@kprokofi could you update the changelog + also mention Dino V2 PR there?

kprokofi:
> @kprokofi could you update the changelog + also mention Dino V2 PR there?

Done

sovrasov (dismissed):

kprokofi:
Classification tests:
image

eugene123tw (dismissed):
Thanks, Kirill. This PR increases the early stopping patience at project level, which will affect all tasks. Let’s keep in mind that this might lead to increased accuracy but also training times in Geti, so we should monitor this closely.

kprokofi:
> Thanks, Kirill. This PR increases the early stopping patience at project level, which will affect all tasks. Let’s keep in mind that this might lead to increased accuracy but also training times in Geti, so we should monitor this closely.

Eventually, It increases early stopping for classification only
patience=10 was always a default number in train.yaml
I also doesn't change it in templates for other tasks

image

sovrasov (approved):

PR #4142, opened 2024-12-04
OTX D-Fine Detection Algorithm Integration

eugene123tw:
# Summary

## OTX D-Fine Detection Algorithm Integration: https://github.com/Peterande/D-FINE
- Introduced five variants of the D-Fine detection algorithm.
- Integrated the HGNetv2 backbone from PaddleDetection.
- Cleaned and optimized the original codebase by:
  - Reducing code duplication where possible.
  - Adding docstrings for all methods and functions.
  - Benchmarking OpenVINO/PyTorch detection results for accuracy and performance.

## Cross-Reference Comparison

|                          | Geti DFine-X  | OTX DFine-X |
|--------------------------|---------------|-------------|
| BlueBerry                | 88%           | 88%         |
| Vitens Coliform          | 56%           | 48%         |
| Vitens Coliform (Tiling) | 86%           | 74%         |
| Vitens Aeromonas         | 92%           | 94%         |
| WGISD                    | 80%           | 86%         |

## Next phase
- Validate potential module combinations that could be unified in future iterations, such as:
  - D-Fine Decoder and RT-DETR Decoder.
  - D-Fine Hybrid Encoder and RT-DETR Decoder.
  - D-Fine Criterion and RT-DETR Criterion.
- Validate Post-Training Optimization Tool (POT) results and assess potential accuracy drops.
- Validate XAI feature.

## How to test

* `otx train --config src/otx/recipe/detection/dfine_x.yaml --data_root DATA_ROOT`
* `pytest tests/unit/algo/detection/test_dfine.py`

### Checklist

- [x] I have updated the documentation in my target branch accordingly (e.g., [documentation](https://github.com/openvinotoolkit/training_extensions/tree/develop/docs) in develop).

### License

```python
```


src/otx/algo/detection/heads/dfine_decoder.py
```diff
@@ -0,0 +1,1165 @@
...
+        sampling_value_list.append(sampling_value_l)
+
+    attn_weights = attention_weights.permute(0, 2, 1, 3).reshape(bs * n_head, 1, len_q, sum(num_points_list))
+    weighted_sample_locs = torch.concat(sampling_value_list, dim=-1) * attn_weights
+    output = weighted_sample_locs.sum(-1).reshape(bs, n_head * c, len_q)
+
+    return output.permute(0, 2, 1)
+
+
+class MSDeformableAttentionV2(nn.Module):
```
kprokofi: Can we use this for RTDetr as well? Maybe it will be upgrade for RTDetrV2

src/otx/algo/detection/d_fine.py
```diff
@@ -0,0 +1,312 @@
...
+    from lightning.pytorch.cli import LRSchedulerCallable, OptimizerCallable
+
+    from otx.core.metrics import MetricCallable
+    from otx.core.schedulers import LRSchedulerListCallable
+    from otx.core.types.label import LabelInfoTypes
+
+
+PRETRAINED_ROOT: str = "https://github.com/Peterande/storage/releases/download/dfinev1.0/"
+
+PRETRAINED_WEIGHTS: dict[str, str] = {
```
kprokofi, line 41: I wonder whether we need all of these variants? We are currently overwhelmed with detection recipes. Could we choose maybe 2 models to expose and omit others? The largest one shows the best performance and it is a candidate for Geti largest template revamp, but other templates seems to be not so beneficial comparing with already introduced models.
So, I would consider cleaning some model versions here (same concerns RTDetr and YOLOX, but it is another story)

src/otx/algo/detection/heads/dfine_decoder.py
```diff
@@ -0,0 +1,1165 @@
...
+            torch.cat(left_values),
+            torch.zeros_like(up[0][None]),
+            torch.cat(right_values),
+            upper_bound2,
+        ],
+        0,
+    )
+
+
+def distance2bbox(points: Tensor, distance: Tensor, reg_scale: Tensor) -> Tensor:
```
kprokofi: maybe put this to utils?

src/otx/algo/detection/heads/dfine_decoder.py
kprokofi: Secondly, I would rather put it to otx/src/otx/algo/common/layers/transformer_layers.py as done for RTDetr.

src/otx/algo/detection/heads/dfine_decoder.py
```diff
@@ -0,0 +1,1165 @@
...
+    x1 = points[..., 0] - (0.5 * reg_scale + distance[..., 0]) * (points[..., 2] / reg_scale)
+    y1 = points[..., 1] - (0.5 * reg_scale + distance[..., 1]) * (points[..., 3] / reg_scale)
+    x2 = points[..., 0] + (0.5 * reg_scale + distance[..., 2]) * (points[..., 2] / reg_scale)
+    y2 = points[..., 1] + (0.5 * reg_scale + distance[..., 3]) * (points[..., 3] / reg_scale)
+
+    bboxes = torch.stack([x1, y1, x2, y2], -1)
+    return box_convert(bboxes, in_fmt="xyxy", out_fmt="cxcywh")
+
+
+def deformable_attention_core_func_v2(
```
kprokofi: Same comment about the location, maybe: otx/src/otx/algo/modules/transformer.py?

src/otx/algo/detection/necks/dfine_hybrid_encoder.py
```diff
@@ -0,0 +1,438 @@
...
+        """Forward pass."""
+        y = list(self.cv1(x).split((self.c, self.c), 1))
+        y.extend(m(y[-1]) for m in [self.cv2, self.cv3])
+        return self.cv4(torch.cat(y, 1))
+
+
+class HybridEncoderModule(nn.Module):
+    """HybridEncoder for DFine.
+
+    TODO(Eugene): Merge with current rtdetr.HybridEncoderModule in next PR.
```
kprokofi, line 174: 👍

src/otx/core/data/transform_libs/torchvision.py
```diff
@@ -3921,3 +3921,44 @@ def _dispatch_transform(cls, cfg_transform: DictConfig | dict | tvt_v2.Transform
             raise TypeError(msg)
 
         return transform
+
+
+class RandomIoUCrop(tvt_v2.RandomIoUCrop):
```
kprokofi, line 3926: If we use already defined RandomIOUCrop in this file, the performance issues occur?
image

kprokofi (commented):
Thank you, Eugene for your great contribution! 
I will try D-Fine from your branch with Intel GPUs

src/otx/algo/detection/d_fine.py
eugene123tw: I suggest removing the three recipes (D-Fine tiny/small/medium) but keeping their configurations in d_fine.py. This way we can reintroduce those models base on user requests, or if there are future improvements to the pre-trained models. Also, removing the recipes will reduce the load on our CI pipeline.

src/otx/algo/detection/heads/dfine_decoder.py
eugene123tw: I moved D-Fine utility functions under: `src/otx/algo/detection/utils/utils.py`

src/otx/algo/detection/heads/dfine_decoder.py
eugene123tw: I removed this fucntion and added to MSDeformableAttentionV2

src/otx/algo/detection/heads/dfine_decoder.py
eugene123tw: @kprokofi yes, we can use it for RTDetrV2. I moved it to otx/src/otx/algo/common/layers/transformer_layers.py

src/otx/core/data/transform_libs/torchvision.py
eugene123tw: I used `torchvision.RandomIOUCrop` to align with the original implementation. I also tested it with `mmdet.MinIoURandomCrop` and observed no significant differences in accuracy or performance. 

I suggest removing `mmdet.MinIoURandomCrop` and using torchvision.RandomIOUCrop to reduce the code maintenance overhead.

sovrasov:
@eugene123tw it looks like detection integration tests failed. Could you have a look?

src/otx/recipe/detection/dfine_l.yaml
```diff
@@ -74,10 +74,26 @@ overrides:
             transform_bbox: true
             keep_ratio: false
             is_numpy_to_tvtensor: true
+        - class_path: otx.core.data.transform_libs.torchvision.PhotoMetricDistortion
+          enable: false
```
samet-akcay: question for more general otx structure: `enable` means this transform will not be part of the transforms? If yes, why is it included here?

src/otx/recipe/detection/dfine_l.yaml
```diff
@@ -74,10 +74,26 @@ overrides:
...
+          init_args:
+            is_numpy_to_tvtensor: true
+        - class_path: otx.core.data.transform_libs.torchvision.RandomAffine
+          enable: false
+          init_args:
+            is_numpy_to_tvtensor: true
```
samet-akcay: Again more OTX-related question: what does `is_numpy_to_tv_tensor` do? Why do we need to set it to `True`

src/otx/core/data/transform_libs/torchvision.py
kprokofi: In general, I agree. If there is no difference, it is better to remove mmdet implementation (We could do it in the next PRs, I plan to review transforms, as there are a lot of them and their amount can be reduced.)

kprokofi (dismissed):

kprokofi:
My experiments with XPU devices have shown that training with D-Fine model shows numerical errors in the Hungarian algorithm during assignments on random iteration. Training speed is similar to RTDetr (slow). If at the end we decide to expose this model to the Geti users, I will create a related ticket.

docs/source/guide/explanation/algorithms/object_detection/object_detection.rst
```diff
@@ -73,6 +73,16 @@ We support the following ready-to-use model recipes:
...
+| `D-Fine S Detection <https://github.com/openvinotoolkit/training_extensions/blob/develop/src/otx/recipe/detection/dfine_s.yaml>`                           |   D-Fine S          | 25.1714             | 39.9            |
++------------------------------------------------------------------------------------------------------------------------------------------------------------+---------------------+---------------------+-----------------+
+| `D-Fine M Detection <https://github.com/openvinotoolkit/training_extensions/blob/develop/src/otx/recipe/detection/dfine_m.yaml>`                           |   D-Fine M          | 56.6844             | 75.4            |
++------------------------------------------------------------------------------------------------------------------------------------------------------------+---------------------+---------------------+-----------------+
+| `D-Fine L Detection <https://github.com/openvinotoolkit/training_extensions/blob/develop/src/otx/recipe/detection/dfine_l.yaml>`                           |   D-Fine L          | 91.0344             | 120.0            |
```
samet-akcay: Not sure if markdown lints are happy with this though.
```suggestion
| `D-Fine L Detection <https://github.com/openvinotoolkit/training_extensions/blob/develop/src/otx/recipe/detection/dfine_l.yaml>`                           |   D-Fine L          | 91.0344             | 120.0           |
```

src/otx/algo/detection/d_fine.py
```diff
@@ -0,0 +1,320 @@
+# Copyright (C) 2024 Intel Corporation
```
samet-akcay: maybe 2025? :)

src/otx/algo/detection/d_fine.py
```diff
@@ -0,0 +1,320 @@
...
+            label_info=label_info,
+            input_size=input_size,
+            optimizer=optimizer,
+            scheduler=scheduler,
+            metric=metric,
+            torch_compile=torch_compile,
+            tile_config=tile_config,
+        )
+
+    def _create_model(self) -> nn.Module:
```
samet-akcay: what's the difference between create_model and build_model. I understand what this code achieves but, conceptually, create-model and build-model are sort of similar, no?

tox.ini
```diff
@@ -48,7 +48,19 @@ deps =
...
+        --cov-report=term-missing \
+        --cov-fail-under=0 \
+        {posargs}
+
+[testenv:heavy-unit-test-{py310, py311}]
+deps =
+    .[base,dev]
+commands =
+    ; Run Unit-Test with coverage report.
+    pytest -m "heavy" tests/unit \
```
samet-akcay: Does heavy mean resource intensive? Is `heavy` common term for such tests? I feel `resource-intensive` or something similar would be more descriptive and common?

samet-akcay: Also, this brings another question: why these tests are resource-intensive? Like the nature of the architecture or something else? Can it be simplified or converted into an integration test?

samet-akcay (commented):
thanks a ton for this huge effort @eugene123tw! 

One suggestion, perhaps, would be to change the license to 2025

Other than that, I only have minor questions/comments.

.github/workflows/pre_merge.yaml
```diff
@@ -84,6 +84,55 @@ jobs:
           curl -Os https://uploader.codecov.io/latest/linux/codecov
           chmod +x codecov
           ./codecov -t ${{ secrets.CODECOV_TOKEN }} --sha $COMMIT_ID -U $HTTP_PROXY -f .tox/coverage_unit-test-${{ matrix.tox-env }}.xml -F ${{ matrix.tox-env }}
+  Heavy-Unit-Test:
```
sovrasov: @negvet please give your opinion on that stuff, can unit tests be so gpu-intensive, or is the builder that weak?

eugene123tw:
image

@samet-akcay @sovrasov @negvet @atwinand Unit test keeps failing due to RAM OOM (not GPU OOM). We think there's a CPU memory leak in dataloader. CPU memory usage grows liearly with the amount of tests for some reason and this affects all tests. I already tried to reduce CPU memory usage by switching off memory cache in tiling unit test case in my PR. But unit test still keeps failing due to lack of CPU memory in free tier github CI machine. 

The only way for me to fix this is to use a self-hosted machine in Korea with more CPU memory (ie. `otx-gpu-a10g-1`). 
We can also upgrade `ubuntu-20.04` to `ubuntu-20.04-8-cores` but this would induce extra cost which we need to seek approval.

The current free tier github action CI has 16GB
image

For now, my only viable solution is to use a more powerful self-hosted machine for resource-intensive unit tests that require higher RAM.

src/otx/recipe/detection/dfine_l.yaml
eugene123tw: **Configurable Augmentation** is introduced in https://github.com/openvinotoolkit/training_extensions/pull/4066. All detection/inst-seg models from now on need to have all the following augmentations in recipes, I guess for Geti integration purpose:

configurable_augs = [
    "PhotoMetricDistortion",
    "RandomAffine",
    "RandomVerticalFlip",
    "GaussianBlur",
    "GaussianNoise",
]

If not, tests/integration/api/test_augmentation.py would fail. I personally think this unit test should be relaxed as it doesn't make sense to include all the configurable augmentations to every detection models. 

@kprokofi @atwinand could you give your opinions?

.github/workflows/pre_merge.yaml
negvet: From my perspective, we should try to avoid making unit tests to be intensive. But as I got Eugene's opinion, as for now, there are no options to fix it, so maybe we might revisit it later on when doing the whole test redesign.

src/otx/recipe/detection/dfine_l.yaml
kprokofi: It is a part of configurable augmentations feature. Every model exposed to Geti should support it. Geti is still interested in that, but with lower priority, I guess. 
It has not been yet finalized. Let's keep it as is right now. We get back to it later.

src/otx/recipe/detection/dfine_l.yaml
kprokofi: Because some augmentations work with numpy, some of them work with torch.Tensor. This flag means that we are explicitly converting numpy to tensors.

src/otx/algo/detection/d_fine.py
kprokofi: Yes, the same. It was needed to hack some piece of code common for all detection models. We call one method inside another. Will need to change it in the future.

tox.ini
eugene123tw: I think this is a project wise issue from pytest and not an issue particularly from this PR. Evgeny will be looking this issue further.

src/otx/algo/detection/d_fine.py
```diff
@@ -0,0 +1,320 @@
...
+            load_checkpoint(detector, self.load_from, map_location="cpu")
+        return detector
+
+    def _build_model(self, num_classes: int) -> DETR:
+        backbone = HGNetv2(model_name=self.model_name)
+        encoder = HybridEncoder(model_name=self.model_name)
+        decoder = DFINETransformer(
+            model_name=self.model_name,
+            num_classes=num_classes,
+        )
```
Daankrol, line 94: Can we merge this code into _create_model()?

src/otx/algo/detection/d_fine.py
```diff
@@ -0,0 +1,320 @@
...
+            imgs_info=inputs.imgs_info,
+            scores=scores,
+            bboxes=bboxes,
+            labels=labels,
+        )
+
+    def configure_optimizers(self) -> tuple[list[torch.optim.Optimizer], list[dict[str, Any]]]:
+        """Configure an optimizer and learning-rate schedulers.
+
+        Configure an optimizer and learning-rate schedulers
```
Daankrol: can remove the first line here.

tests/integration/api/test_xai.py
```diff
@@ -51,6 +51,10 @@ def test_forward_explain(
         # TODO(Eugene): maskdino not support yet.
         pytest.skip(f"There's issue with inst-seg: {model_name}. Skip for now.")
 
+    if "dfine" in model_name:
+        # TODO(Eugene): dfine not support yet.
+        pytest.skip(f"There's issue with dfine: {model_name}. Skip for now.")
+
```
Daankrol: Maybe add a ticket in the todo here for future tracking

Daankrol (dismissed):
Looks really good! Some very small notes, can be merged if tests are fixed.

src/otx/algo/detection/d_fine.py
eugene123tw: I kept _build_model but removed _create_model as I found we don't actually need it.

src/otx/algo/detection/d_fine.py
eugene123tw: Updated

tests/integration/api/test_xai.py
eugene123tw: Updated

Daankrol (approved):

.github/workflows/pre_merge.yaml
```diff
@@ -84,6 +84,38 @@ jobs:
           curl -Os https://uploader.codecov.io/latest/linux/codecov
           chmod +x codecov
           ./codecov -t ${{ secrets.CODECOV_TOKEN }} --sha $COMMIT_ID -U $HTTP_PROXY -f .tox/coverage_unit-test-${{ matrix.tox-env }}.xml -F ${{ matrix.tox-env }}
+  Intense-Unit-Test:
```
sovrasov, line 87: Perhaps that can be removed one we downsize our models portfolio. 
cc @kprokofi @ashwinvaidya17

sovrasov (approved):

.github/workflows/pre_merge.yaml
ashwinvaidya17: Maybe we can also remove tox as it is slower. @samet-akcay saw some speedup by using direct pytest call. Also, he is working on a new design for testing.

PR #4188, opened 2025-01-27
Test Geti integration

negvet:
### Summary

Test Geti-OTX interaction (2 way).

Currently tests cover three classification tasks (multi-class, multi label, h label), effnet_b0 is used for all tasks.
3 tests in total.

### Checklist

### License

```python
```


tests/integration/api/test_geti_interaction.py
```diff
@@ -0,0 +1,209 @@
...
+            if self.mapi_model_class is not None:
+                ov_export_dir = self.tmp_path / "ov_export"
+                ov_export_dir.mkdir(parents=True, exist_ok=True)
+                unzip_exportable_code(
+                    work_dir=self.tmp_path,
+                    exported_path=exported_path,
+                    dst_dir=ov_export_dir,
+                )
+                xml_path = ov_export_dir / "exported_model.xml"
+                mapi_model = self.mapi_model_class.create_model(xml_path)
```
sovrasov: There's no need to specify a particular class: `Model.create(...)` should work

tests/integration/api/test_geti_interaction.py
```diff
@@ -0,0 +1,209 @@
...
+
+                predictions = mapi_model(self.image)
+                assert predictions is not None
+                assert len(predictions.top_labels) > 0
+
+            exported_path.unlink(missing_ok=True)
+
+    def test_export_and_infer_openvino(self):
+        """Test exporting the model to OpenVINO."""
+        for precision in [OTXPrecisionType.FP16, OTXPrecisionType.FP32]:
```
sovrasov, line 126: Can be done via pytest parametrization for verbosity

tests/integration/api/test_geti_interaction.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def test_optimize_openvino_fp32(self):
+        """Test optimizing the OpenVINO model with FP32 precision."""
+        fp32_export_dir = self.tmp_path / "fp32_export"
+        fp32_export_dir.mkdir(parents=True, exist_ok=True)
+        exported_path = self.engine.export(
+            export_format=OTXExportFormatType.OPENVINO,
+            export_precision=OTXPrecisionType.FP32,
+            explain=True,
+            export_demo_package=True,
```
sovrasov, line 162: Is there something special about exporting models as a demo package? Normal export is enough to run in the scenario declared here

tests/integration/api/test_geti_interaction.py
```diff
@@ -0,0 +1,209 @@
...
+            checkpoint=fp32_export_dir / "exported_model.xml",
+            export_demo_package=True,
+        )
+        assert optimized_path.exists()
+
+        # Test Model API model
+        if self.mapi_model_class is not None:
+            ov_optimized_dir = self.tmp_path / "ov_optimize"
+            ov_optimized_dir.mkdir(parents=True, exist_ok=True)
+            unzip_exportable_code(
```
sovrasov: Doesn't `optimized_path` already point to the compressed model?

tests/integration/api/test_geti_interaction.py
eugene123tw: @negvet @sovrasov Similar question. I think Geti support ONNX export? Do we need to test it as well?

tests/integration/api/test_geti_interaction.py
```diff
@@ -0,0 +1,223 @@
...
+            xml_path = ov_optimized_dir / "exported_model.xml"
+            mapi_model = self.mapi_model_class.create_model(xml_path)
+            assert mapi_model is not None
+
+            predictions = mapi_model(self.image)
+            assert predictions is not None
+            assert len(predictions.top_labels) > 0
+
+
+@pytest.mark.parametrize("task", pytest.TASK_LIST)
```
eugene123tw, line 192: I think we don't need all models from list for API test, just 3 models (fast/balance/accurate) should suffic?

tests/integration/api/test_geti_interaction.py
negvet: how will it differentiate between e.g. cls and det models?

tests/integration/api/test_geti_interaction.py
negvet: in this case I still need to manually loop through precision types in the `test_engine_api`.
The goal is also to do one training per both types check.

tests/integration/api/test_geti_interaction.py
negvet: You are right. But I just followed the same scenario as Geti. In Geti, `export_demo_package` is always True for OV.

tests/integration/api/test_geti_interaction.py
negvet: Fair point, currently I implemented only one model per task, it is effnet_b0 (I guess it is "fast").
But this can be extended to fast/balance/accurate.

tests/integration/api/test_geti_interaction.py
negvet: `optimized_path` is `.zip`

tests/integration/api/test_geti_interaction.py
sovrasov: That information is already embedded to `.xml` files

tests/integration/api/test_geti_interaction.py
eugene123tw: @sovrasov @negvet yeah, not sure how we'll address this. But currently config.json is tied to one speific model. To be able to train all 3, we need to decide if we wanna have 3 config.json or overwrite model entry in current config.json.

eugene123tw (approved):

sovrasov (approved):

PR #4214, opened 2025-02-12
Don't filter empty label from kp arrow

sovrasov:
### Summary
Pre-filtering was causing one missing keypoint label. Related workaround in kp dataset was deleted

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/core/data/dataset/keypoint_detection.py
+++ b/src/otx/core/data/dataset/keypoint_detection.py
@@ -19,7 +19,6 @@ from otx.core.data.entity.keypoint_detection import KeypointDetBatchDataEntity, 
 from otx.core.data.mem_cache import NULL_MEM_CACHE_HANDLER, MemCacheHandlerBase
 from otx.core.data.transform_libs.torchvision import Compose
 from otx.core.types.image import ImageColorChannel
-from otx.core.types.label import LabelInfo, NullLabelInfo
 
 from .base import OTXDataset
 
@@ -54,16 +53,6 @@ class OTXKeypointDetectionDataset(OTXDataset[KeypointDetDataEntity]):
         )
 
         self.dm_subset = self._get_single_bbox_dataset(dm_subset)
-
-        if self.dm_subset.categories():
-            kp_labels = self.dm_subset.categories()[AnnotationType.points][0].labels
-            self.label_info = LabelInfo(
-                label_names=kp_labels,
-                label_groups=[],
-                label_ids=[str(i) for i in range(len(kp_labels))],
-            )
-        else:
-            self.label_info = NullLabelInfo()
 
     def _get_single_bbox_dataset(self, dm_subset: DatasetSubset) -> Dataset:
         """Method for splitting dataset items into multiple items for each bbox/keypoint."""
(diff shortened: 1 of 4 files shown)


src/otx/core/data/dataset/keypoint_detection.py
```diff
@@ -55,15 +55,14 @@ def __init__(
 
         self.dm_subset = self._get_single_bbox_dataset(dm_subset)
 
-        if self.dm_subset.categories():
+        # arrow doesn't follow common coco convention, no need to fetch kp-specific labels
+        if self.dm_subset.categories() and data_format != "arrow":
```
atwinand, line 59: What would be set for self.label_info if this condition is not fulfilled?

src/otx/core/data/dataset/keypoint_detection.py
```diff
@@ -55,15 +55,14 @@ def __init__(
...
             kp_labels = self.dm_subset.categories()[AnnotationType.points][0].labels
             self.label_info = LabelInfo(
                 label_names=kp_labels,
                 label_groups=[],
                 label_ids=[str(i) for i in range(len(kp_labels))],
             )
-        else:
```
kprokofi, line 65: I think we should explicitly assign self.label_info in case of 'arrow' format. Overwise -> raise an error, since NullLabelInfo will break the training at the end.

src/otx/core/data/pre_filtering.py
```diff
@@ -94,7 +95,12 @@ def remove_unused_labels(
         raise ValueError(msg)
     if len(used_labels) == len(original_categories):
         return dataset
-    if data_format == "arrow" and max(used_labels) != len(original_categories) - 1:
+    # this section removes an empty label from arrow dataset. Not applicable for keypoints data
```
kprokofi: We don't have "Empty label" in case of KP detection, right? Relying on the AnnotationType is not a good choice in my opinion. Can we make it in the same way as for Hierarchical classification? Just skip pre-filtering in case of KP.

src/otx/core/data/dataset/keypoint_detection.py
sovrasov: In that case, label info generated by parent class is used. For kp arrow common labels extraction procedure works correctly, while for kp filesystem coco dataset is doesn't work, and we need another one here

src/otx/core/data/dataset/keypoint_detection.py
sovrasov: By entering this if/else label info (Null or not) was already generated. The culprit is that it is correct by this moment for arrow, but for coco it is not, and therefore we have an extra labels extraction procedure in this constructor

src/otx/core/data/pre_filtering.py
sovrasov: Pre-filtering removes not just empty labels, but any other unused labels, so it should be done for kp filesystem data as well. Moved checks to the upper level

src/otx/core/data/dataset/keypoint_detection.py
sovrasov: The reference implementation (which is used for arrow) is here: https://github.com/openvinotoolkit/training_extensions/blob/3c9c606589a0053471e17810a500d6fb62eebcad/src/otx/core/data/dataset/base.py#L99
If a change in `NullLabelInfo` is required, it should be done there to cover all the tasks

PR #4227, opened 2025-02-14
Remove Semi-SL algorithm + MM transforms

kprokofi:
### Summary

- [x] delete Semi-SL algo + tests
- [x] delete unlabeled data
- [x] remove TrainType
- [x] delete MM transforms files

### Checklist

### License

```python
```


ashwinvaidya17 (dismissed):
Thanks for the efforts! Maybe in the future we should consider splitting files like `classification.py` into multiple files.

eugene123tw (approved):

src/otx/core/model/classification.py
```diff
@@ -60,10 +61,31 @@ def __init__(
             scheduler=scheduler,
             metric=metric,
             torch_compile=torch_compile,
-            train_type=train_type,
         )
         self.input_size: tuple[int, int]
 
+    def _create_model(self) -> nn.Module:
```
sovrasov, line 67: Where was this method before and why do we need to replicate it 3 times now?

kprokofi: this method was in each specific model implementation. Actually, there was a code duplication, that's why I put it in the base class.

kprokofi: _build_model is called with different number of classes to identify classification layers. Not sure if it is required. Need to revisit it. For now, I just copied it.

sovrasov (approved):

PR #4241, opened 2025-02-22
Refactor OTX Models

kprokofi:
### Summary

- [x] Combine Explainable models with base models
- [x] divide classification tasks for better maintenance and cleaner code
- [x] Remove get_classification_layers and necessity to build models 3 time every time. Move this to incremental checkpoint loading
- [x] Introduce DataInputParams, incapsulate input_size, mean and scale. Derive mean and std values from Normilize transform and automatically pass  it to the model
- [x] Align all models with same interface, introduce model_name for every model. Note: didn't pay much attention to Anomaly models since we will move them to separate Engine. Same for OV models (we will remove them). Aligned between models pretrained weights handling.

### Checklist

### License

```python
```


src/otx/algo/anomaly/padim.py
```diff
@@ -47,13 +49,13 @@ def __init__(
             OTXTaskType.ANOMALY_DETECTION,
             OTXTaskType.ANOMALY_SEGMENTATION,
         ] = OTXTaskType.ANOMALY_CLASSIFICATION,
-        input_size: tuple[int, int] = (256, 256),
     ) -> None:
-        self.input_size = input_size
```
ashwinvaidya17, line 52: Um so I am not sure if this will work. The entire anomaly task is a hack to get around the OTX design. The input size needs to be set before super is called otherwise the task breaks :(

src/otx/recipe/rotated_detection/openvino_model.yaml
```diff
@@ -1,5 +1,5 @@
 model:
-  class_path: otx.core.model.instance_segmentation.OVRotatedDetectionModel
+  class_path: otx.core.model.instance_segmentation.OVInstanceSegmentationModel
```
ashwinvaidya17, line 2: So the plan is to do rotated detection using instance segmentation models?

ashwinvaidya17 (commented):
Thanks for the large refactor! I have one minor comment and one concern related to the anomaly models.

src/otx/core/exporter/native.py
```diff
@@ -118,7 +115,8 @@ def to_onnx(
         Returns:
             Path: The path to the saved ONNX model.
         """
-        dummy_tensor = torch.rand(self.input_size).to(next(model.parameters()).device)
+        input_size = (1, 3, *self.data_input_params.input_size)
```
sovrasov: I'd opt to move this operation out from the exporter, as it becomes not that generic. Once can add a method to the input params to generate `NCHW` shape for instance

sovrasov (commented):
Thanks for the PR!
Is MaskDINO model also deleted?
Can we separate cleaning PRs and refactoring PRs to avoid huge changesets which are hard to review?

src/otx/algo/anomaly/padim.py
kprokofi: Actually, I did this in different way, I assign some default value to OTXAnomaly and then reload the attribute in the derived class. However, I decided to change it in same way as it is now. I pass self.data_input_params to OTXAnomaly.

src/otx/recipe/rotated_detection/openvino_model.yaml
kprokofi: It is right now, rotated detection is the instance segmentation algorithm with post-processing on the predict stage. I've just removed unnecessary separate file and moved RotatedMaskRCNN to MaskRCNN file. OVRotatedDetectionModel == OVInstanceSegmentationModel

src/otx/core/exporter/native.py
kprokofi: I added a method to DataInputParams to return input_size as NCHW format.

kprokofi:
> Thanks for the PR! Is MaskDINO model also deleted? Can we separate cleaning PRs and refactoring PRs to avoid huge changesets which are hard to review?

I understand that there are a lot of changes. However, I decided to delete models, that we rejected and those which are not exposed to Geti. My changes with input size affected all the models and it is better to delete them at all rather than fixing them and tests.

What has been deleted: YOLOV9, MaskDINO, HuggingFace models (we will move HuggingFace functionality to the separate engine) + CLI install (There was no need to delete this file in this PR, but if it is OK, I would not return it back)

sovrasov:
> > Thanks for the PR! Is MaskDINO model also deleted? Can we separate cleaning PRs and refactoring PRs to avoid huge changesets which are hard to review?
> 
> I understand that there are a lot of changes. However, I decided to delete models, that we rejected and those which are not exposed to Geti. My changes with input size affected all the models and it is better to delete them at all rather than fixing them and tests.
> 
> What has been deleted: YOLOV9, MaskDINO, HuggingFace models (we will move HuggingFace functionality to the separate engine) + CLI install (There was no need to delete this file in this PR, but if it is OK, I would not return it back)

Unused models could have been deleted in advance, before the actual refactoring, like Ashwin did with a number of other models, or even tasks

kprokofi:
> > > Thanks for the PR! Is MaskDINO model also deleted? Can we separate cleaning PRs and refactoring PRs to avoid huge changesets which are hard to review?
> > 
> > 
> > I understand that there are a lot of changes. However, I decided to delete models, that we rejected and those which are not exposed to Geti. My changes with input size affected all the models and it is better to delete them at all rather than fixing them and tests.
> > What has been deleted: YOLOV9, MaskDINO, HuggingFace models (we will move HuggingFace functionality to the separate engine) + CLI install (There was no need to delete this file in this PR, but if it is OK, I would not return it back)
> 
> Unused models could have been deleted in advance, before the actual refactoring, like Ashwin did with a number of other models, or even tasks

Agree, I will create a separate PR with deleting things and update this one after

kprokofi:
Cleaning PR: https://github.com/openvinotoolkit/training_extensions/pull/4247

kprokofi:
PR is updated and ready for review

src/otx/algo/detection/atss.py
```diff
@@ -108,8 +120,8 @@ def _build_model(self, num_classes: int) -> SingleStageDetector:
                 target_means=(0.0, 0.0, 0.0, 0.0),
                 target_stds=(0.1, 0.1, 0.2, 0.2),
             ),
-            train_cfg=train_cfg,  # TODO (sungchul, kirill): remove
-            test_cfg=test_cfg,  # TODO (sungchul, kirill): remove
+            train_cfg=train_cfg,  # TODO (Kirill): remove
+            test_cfg=test_cfg,  # TODO (Kirill): remove
```
Daankrol, line 124: can we create a ticket/issue for this and add a reference here?

src/otx/core/model/base.py
```diff
@@ -593,26 +626,31 @@ def load_state_dict_pre_hook(self, state_dict: dict[str, torch.Tensor], prefix:
...
             model_param = self.state_dict()[param_name].clone()
             ckpt_param = state_dict[prefix + param_name]
-            stride = info.get("stride", 1)
-            num_extra_classes = info.get("num_extra_classes", 0)
             for model_dst, ckpt_dst in enumerate(model2ckpt):
                 if ckpt_dst >= 0:
-                    model_param[(model_dst) * stride : (model_dst + 1) * stride].copy_(
-                        ckpt_param[(ckpt_dst) * stride : (ckpt_dst + 1) * stride],
+                    model_param[model_dst : model_dst + 1].copy_(
+                        ckpt_param[ckpt_dst : ckpt_dst + 1],
```
Daankrol, line 636: Why are we removing all strides?

Daankrol (dismissed):
Thanks for this great PR! It would have been easier to review if this was split up into seperate PR's though.

src/otx/core/model/base.py
kprokofi: This parameter didn't make sense. It was always 1 for every model. Previously this code was copied from OTX1.6 or even MMLab. I simplified it now. I changed "get classification layers" and removed unused info like strides and etc.

src/otx/algo/detection/atss.py
kprokofi: I do have a ticket: CVS-163772
However, not sure if it is OK to put Jira ticket in the code.

tests/unit/core/data/utils/test_utils.py
```diff
@@ -176,8 +173,8 @@ def test_adapt_input_size_to_dataset(mocker):
         },
         "annotation": {"size_of_shape": {"robust_min": 64}},
     }
-    input_size = adapt_input_size_to_dataset(dataset=MagicMock(), base_input_size=512)
-    assert input_size == (512, 512)
+    input_size = adapt_input_size_to_dataset(dataset=MagicMock())
+    assert input_size == (512, 600)
```
ashwinvaidya17, line 177: Seems like the functionality has changed here. Does it break anything? Do we not expect square resizing anymore?

src/otx/algo/instance_segmentation/maskrcnn.py
```diff
@@ -350,3 +392,83 @@ def _optimization_config(self) -> dict[str, Any]:
             return {"model_type": "transformer"}
 
         return {}
+
+
+class RotatedMaskRCNNModel(MaskRCNN):
```
ashwinvaidya17, line 397: For the next refactor; It would be nice to split these files as well

ashwinvaidya17 (dismissed):
I see that there are two refactors here. 1. Split the classification models. 2. Introduce input parameters. Maybe for next time we can split the two tasks

tests/unit/core/data/utils/test_utils.py
kprokofi: Actually we don't use this functionality anywhere. It is an optional feature. Some algorithm colects statistic abouth the given dataset and outputs "optimal average input size". I doubt whether it is useful. 
In that test case algorithm preserves ratio between height and width. Previously, we could use "base_input_size", which meaning was misleading, parameter was redundant in my opinion and it required additional parameter in OTXDataModule.

I didn't delete this "auto input size" for now. But we could consider doing that in the nearest future.

src/otx/algo/instance_segmentation/maskrcnn.py
kprokofi: RotatedMaskRCNNModel is actually MaskRCNN, but with an additional postprocessing on the predict stage. We discussed that with Eugene and decided to put this model alongside with MaskRCNN.

ashwinvaidya17 (approved):

Daankrol (approved):

PR #4270, opened 2025-03-04
Fix KP Detection performance

kprokofi:
### Summary

- [x] Filter invisible/occluded keypoints based on the dataset info (Geti usecase)
- [x] Change resize for the export and align resize with Model API
- [x] Tune augmentations and training schedule (lr + lr scheduler)
- [x] Delete second configuration for RTMPose model since we don't use this configuration right now
- [x] Change decoding of the scores (use normalized values) to fix scores > 1 issue

### Checklist

### License

```python
```


sovrasov (commented):
Thanks for the PR! Could you provide some experiments results for both the templates we have to check the gain obtained?

kprokofi:
> Thanks for the PR! Could you provide some experiments results for both the templates we have to check the gain obtained?

Sure!
I sampled data from full COCO dataset with 4 different sizes 3 times (12 datasets). I also included 2 datasets from Geti project in Arrow format. I run my experiments with 3 repeats. I recorded min accuracy during runs and average for each data size. Please note, that data size means full dataset (training + val + test). Splits were done as in Geti. It means, 50% train, 50% Val + Test. For example, data size == 36 means: 18 images for train, 9 for val, 9 for test. Test dataset differs from validation. 
Also note, that on the validation and test in the develop branch we use AffineTransform, however on my branch we use OpenCV resize. The results on the develop branch with Resize may be even worse.

develop branch (Before):
Average PCK for data size 12: 0.0488, Min PCK: 0.0000
Average PCK for data size 24: 0.0577, Min PCK: 0.0000
Average PCK for data size 36: 0.1342, Min PCK: 0.0267
Average PCK for data size 48: 0.2730, Min PCK: 0.0671
coco_person_small (64 training images), val score: 53%

___
Current branch (After):
Average PCK for data size 12: 0.2048, Min PCK: 0.0594
Average PCK for data size 24: 0.3341, Min PCK: 0.1706
Average PCK for data size 36: 0.5254, Min PCK: 0.3854
Average PCK for data size 48: 0.5124, Min PCK: 0.4247
coco_person_small (64 training images), val score: 72%

It is clear, that results became more reliable starting already from 24 images.

src/otx/recipe/_base_/data/visual_prompting.yaml
```diff
@@ -23,7 +23,7 @@ train_subset:
         scale: $(input_size)
         keep_ratio: true
         transform_bbox: true
-        transform_point: true
```
Daankrol, line 26: It feels weird that we need to specify these "transform_bbox" and "transform_point" params for tasks where we do not even have bbox or points. Can't we use default values for this in the function itself?

src/otx/recipe/keypoint_detection/rtmpose_tiny_single_obj.yaml
```diff
@@ -10,19 +10,17 @@ model:
...
+        num_warmup_steps: 100
         main_scheduler_callable:
-          class_path: lightning.pytorch.cli.ReduceLROnPlateau
+          class_path: torch.optim.lr_scheduler.CosineAnnealingLR
           init_args:
-            mode: max
-            factor: 0.1
-            patience: 9
-            monitor: val/PCK
+            T_max: 10000
```
Daankrol: Do we have any experimental results on the parameters for the CosineAnnealing scheduler? A Tmax value of 10000 is quite high for the amount of epochs (<10000) we train the KP detection models for.

kprokofi: It is not epochs, it is iterations. This value is suboptimal, I would agree. The best way is to design an automatic mechanism to patch scheduler with number of iterations for each specific dataset. Since this algo in Geti experimental, I just set some value to fit at least 10k iterations. I tried to set this value to 100000 as it done for classification, it also works, but the result is slightly worse on the small data

src/otx/recipe/_base_/data/visual_prompting.yaml
kprokofi: I just set these parameter as they were before. However, I renamed points to bbox_points to not confuse with keypoints meaning. If we set these value to True by default, it will affect other algorithms that use Resize().  Actually we decode some info in visual prompting task:

image

src/otx/recipe/_base_/data/keypoint_detection.yaml
```diff
@@ -6,20 +6,18 @@ data_format: coco_person_keypoints
...
 image_color_channel: RGB
 input_size:
-  - 256
-  - 192
+  - 512
+  - 512
 train_subset:
   subset_name: train
   batch_size: 32
+  num_workers: 2
```
sovrasov, line 14: Is there a problem when num workers is not 2?

kprokofi: No problem, I just aligned it with other tasks and added the explicit value in config

sovrasov (approved):

samet-akcay (approved):
lgtm, thanks

PR #4330, opened 2025-04-07
📝 Improve Geti Integration and Streamline Pre-Merge Tests

eugene123tw:
### Summary

This PR introduces improvements to Geti integration pipeline and optimizes test coverage for pre-merge CI. 

Updates include:

🚀 Features
* Added OTXConfig class: Converts Geti templates into OTXConfig for Geti workflows using the Engine API.
* Added Arrow-based integration dataset

🛠️ Fixes
* Resolved missing values in several task configuration YAMLs 

✅ Testing & CI Improvements
* Pre-merge CLI integration tests now run only on the balance model to minimize CI resource usage.
* Pre-merge Geti integration tests are expanded to cover balance, accuracy, and speed model variants across all supported tasks (excluding KP detection fow now due to missing arrow)
* Deprecated and removed unused tests/integration/cli/test_export_inference.py 
### Checklist

### License

```python
```


tests/integration/api/test_geti_interaction.py
```diff
@@ -187,17 +228,54 @@ def test_optimize_and_infer_openvino_fp32(self):
         assert predictions is not None
 
 
-@pytest.mark.parametrize("task", pytest.TASK_LIST)
-def test_engine_api(task: OTXTaskType, tmp_path: Path):
+@pytest.fixture()
+def fxt_engine_api(
+    task_template: tuple[OTXTaskType, str, bool],
```
ashwinvaidya17: What's the advantage of splitting this into a separate fixture? Is the plan to use `fxt_engine_api` elsewhere as well?

tests/integration/api/geti_otx_config_utils.py
```diff
@@ -0,0 +1,238 @@
...
+UI_RULES = "ui_rules"
+TYPE = "type"
+OPTIONS = "options"
+ENUM_NAME = "enum_name"
+
+
+def allows_model_template_override(keyword: str) -> bool:
+    """Returns True if the metadata element described by `keyword` can be overridden in a model template file.
+
+    :param keyword: Name of the metadata key to check.
```
ashwinvaidya17: Minor comment but this file uses a different docstring convention

ashwinvaidya17 (commented):
I have a few minor comments. However, my biggest concern with the current tests is OTX is that they are hard to follow. We have a lot of pre-processing/setup logic for running the tests. It might be a good idea to revisit the tests in the future.

eugene123tw:
@ashwinvaidya17 I think the complexity mainly comes from the number of testing scenarios we need to handle, as well as the fact that Geti uses OTX in a very different way compared to a typical OTX user would.

On one hand, we need to support Geti-specific workflows—meaning OTX has to convert config.json files from Geti into a format it can consume. But we also want to optimize runtime by reducing the number of test cases run during PR merges.

At the same time, we don’t want to compromise on integration test coverage. So we introduced a daily test job that runs across all recipes. On PR merges, we still want to run a subset of the CLI tests, not just the Geti integration ones.

All of this combined means there’s quite a bit of logic needed to balance speed, coverage, and flexibility. Right now, I don’t see a super clean or elegant way to manage it all, but it’s definitely something we can revisit and improve in the future.

tests/integration/api/test_geti_interaction.py
eugene123tw: I removed this part

tests/integration/api/geti_otx_config_utils.py
eugene123tw: Modified

ashwinvaidya17 (dismissed):

tests/conftest.py
```diff
@@ -1,10 +1,15 @@
-# Copyright (C) 2023 Intel Corporation
+# Copyright (C) 2023-2025 Intel Corporation
```
Daankrol, line 1: shouldn't this just be 2025 instead of 2023-2025?
```suggestion
# Copyright (C) 2025 Intel Corporation
```

tests/integration/api/geti_otx_config_utils.py
```diff
@@ -0,0 +1,249 @@
...
+                },
+                fp,
+            )
+
+    def to_otx_config(self, work_dir: Path) -> dict[str, dict]:
+        fpath = work_dir / "tmp_config.json"
+        self.to_json_file(fpath)
+
+        with self.monkeypatch_cls_task_type(override_cls_task_type=self.sub_task_type):
+            otx2_config = ConfigConverter.convert(fpath)
```
Daankrol: this variable name is a bit unclear to me. Why not just "otx_config"

tests/integration/api/geti_otx_config_utils.py
```diff
@@ -0,0 +1,249 @@
...
+        otx2_config["data"]["train_subset"]["subset_name"] = "TRAINING"
+        otx2_config["data"]["val_subset"]["subset_name"] = "VALIDATION"
+        otx2_config["data"]["test_subset"]["subset_name"] = "TESTING"
+
+        return otx2_config
+
+    @staticmethod
+    @contextmanager
+    def monkeypatch_cls_task_type(override_cls_task_type: OTXTaskType | None = None) -> Iterator[None]:
+        """Monkeypath classification task type which is fixed as `MULTI_CLASS_CLS` in OTX side.
```
Daankrol:
```suggestion
        """Monkeypatch classification task type which is fixed as `MULTI_CLASS_CLS` in OTX side.
```

Daankrol (dismissed):
Great PR! I have some minor comments

tests/integration/api/test_geti_interaction.py
```diff
@@ -187,17 +228,36 @@ def test_optimize_and_infer_openvino_fp32(self):
...
+    # Unpack the fixture
+    task, template_path, tiling = task_template
+
     if task not in DEFAULT_GETI_CONFIG_PER_TASK:
         pytest.skip("Only the Geti Tasks are tested to reduce unnecessary resource waste.")
 
     config_arrow_path = DEFAULT_GETI_CONFIG_PER_TASK[task]
-    geti_config_path = config_arrow_path / "config.json"
-    arrow_file_path = config_arrow_path / "datum-0-of-1.arrow"
+    arrow_file_path = (
```
rajeshgangireddy, line 248: Would be better to use os.path.join for paths

tox.ini
```diff
@@ -56,8 +56,10 @@ commands =
     python -m pytest tests/integration -ra --showlocals --csv={toxworkdir}/{envname}.csv --task {[testenv]task} --open-subprocess {posargs}
 
 
-# Pre-merge integration tests only run on model recipe/template labelled with `model-category`
-[testenv:integration-test-{all, classification, multi_class_cls, multi_label_cls, h_label_cls, detection, rotated_detection, keypoint_detection, instance_segmentation, semantic_segmentation, anomaly}]
+# Pre-merge integration tests run 
+#   * CLI tests on "BALANCE" category for each task
+#   * Geti tests on "SPEED/BALANCE/ACCURACY" category for each task
+[testenv:integration-test-{all, classification, multi_class_cls, multi_label_cls, h_label_cls, detection, rotated_detection, keypoint_detection, instance_segmentation, semantic_segmentation, anomaly_classification}]
```
rajeshgangireddy, line 62: what's the difference between classification and multi_class_cls? Why do we need both ?

tests/integration/api/test_geti_interaction.py
```diff
@@ -19,15 +19,29 @@
...
+    PrecisionType,
+    load_hyper_parameters,
+)
 
 if TYPE_CHECKING:
     from otx.engine.engine import Engine
 
 TEST_PATH = Path(__file__).parent.parent.parent
 DEFAULT_GETI_CONFIG_PER_TASK = {
+    # OTXTaskType.KEYPOINT_DETECTION: Not supported yet as we can't import KP dataset to Geti
```
rajeshgangireddy, line 36: is it possible to make this cleaner as TEST_PATH / "assets" / "geti_config_arrow" is common for all tasks

rajeshgangireddy (commented):
Nice! Thanks for the effort. minor comments for my side as well

tests/conftest.py
eugene123tw: If new file is introduced then it'd be 2025 only, however, if edited we still keep the original year but extended to current year. I knew this from Samet 😃

tox.ini
eugene123tw: Currently, `classification` task runs all three variants: multi_class_cls, multi_label_cls, and h_label_cls. It is still used by the legacy performance benchmark workflow (`.github/workflows/perf_benchmark.yaml`), but not by the newer `perf_benchmark_v2.yaml`. Once all benchmark datasets are available and migrated to the new workflow, we can safely remove both the old performance benchmark and the classification task. Until then, it's best to keep it for compatibility.

tests/integration/api/test_geti_interaction.py
eugene123tw: I changed it to TEST_ARROW_PATH = Path(__file__).parent.parent.parent / "assets" / "geti_config_arrow"

tests/integration/api/test_geti_interaction.py
eugene123tw: I think it's best to continue using pathlib, as it's the most commonly used approach in OTX.

sovrasov: @rajeshgangireddy we've previously agreed to prefer pathlib over os.path in OTX

sovrasov (approved):

rajeshgangireddy (approved):

PR #4336, opened 2025-04-10
Bug Fix : Convert labels for multi-label Classification into LongTensor before one_hot encoding

rajeshgangireddy:
### Summary

For multilabel classification, we use torch' one_hot to get one hot encoding of the labels. This method requires labels to be for dtype LongTensor. When we have empty labels, then the tensor is of type float32. This gives an error.  This PR manually changes the labels to be LongTensor if not already before calling the one_hot function. 

[one_hot](https://pytorch.org/docs/stable/generated/torch.nn.functional.one_hot.html#torch-nn-functional-one-hot)

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/core/data/dataset/classification.py
+++ b/src/otx/core/data/dataset/classification.py
@@ -111,6 +111,11 @@ class OTXMultilabelClsDataset(OTXDataset):
 
     def _convert_to_onehot(self, labels: torch.tensor, ignored_labels: list[int]) -> torch.tensor:
         """Convert label to one-hot vector format."""
+
+        # Torch's one_hot() expects the input to be of type long
+        # However, when labels are empty, they are of type float32
+        labels = labels.to(dtype=torch.long, copy=False)
+
         onehot = functional.one_hot(labels, self.num_classes).sum(0).clamp_max_(1)
         if ignored_labels:
             for ignore_label in ignored_labels:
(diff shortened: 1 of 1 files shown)


sovrasov (dismissed):

sovrasov:
Does the problem occur only on samples with no labels?

rajeshgangireddy:
@sovrasov 
> Does the problem occur only on samples with no labels?


Yes.

eugene123tw:
Thanks Rajesh fixing this issue :100:  Would it be possible to change labels.to(dtype=torch.long, copy=False) to labels.long() instead? I think it's mainly for brevity and since performance or memory usage may not be a concern in this case as it's a small tensor.

rajeshgangireddy:
> Thanks Rajesh fixing this issue 💯 Would it be possible to change labels.to(dtype=torch.long, copy=False) to labels.long() instead? I think it's mainly for brevity and since performance or memory usage may not be a concern in this case as it's a small tensor.

Done.

src/otx/core/data/dataset/classification.py
```diff
@@ -111,6 +111,9 @@ def _get_item_impl(self, index: int) -> TorchDataItem | None:
 
     def _convert_to_onehot(self, labels: torch.tensor, ignored_labels: list[int]) -> torch.tensor:
         """Convert label to one-hot vector format."""
+        # Torch's one_hot() expects the input to be of type long
+        # However, when labels are empty, they are of type float32
+        labels = labels.long()
         onehot = functional.one_hot(labels, self.num_classes).sum(0).clamp_max_(1)
```
Daankrol: not sure if the comments are necessary. 
```suggestion
        onehot = functional.one_hot(labels.long(), self.num_classes).sum(0).clamp_max_(1)
```

Daankrol (approved):

sovrasov (approved):

PR #4339, opened 2025-04-14
🛠️ Engine Refactor | Native Engine

ashwinvaidya17:
### Summary

- Part 1 of the refactor. Moves current engine to native under backend

### Checklist

### License

```python
```


src/otx/backend/native/engine/engine.py
```diff
@@ -0,0 +1,1183 @@
+"""OTX Engine."""
```
eugene123tw: Could we streamline the folder structure a bit? I’m wondering if moving src/otx/backend/native/engine/engine.py to src/otx/backend/native/engine.py would make things clearer.

src/otx/engine/__init__.py
```diff
@@ -1,8 +1,38 @@
...
+from typing import TYPE_CHECKING
+
 from .engine import Engine
 
-__all__ = ["Engine"]
+if TYPE_CHECKING:
+    from otx.types import DATA, MODEL
+
+
+def create_engine(model: MODEL, data: DATA) -> Engine:
```
eugene123tw, line 16: Since this folder is now empty, could we move create_engine to backend/__init__.py and remove the folder? Or is there a dependency reason for keeping it?

eugene123tw (commented):
Hi Ashwin, thanks for your effort on this PR . It's a great start and pave the way for adding more engines. I’ve left two minor comments regarding the folder structure.

src/otx/backend/native/engine/engine.py
ashwinvaidya17: I'll move it a level higher. The reason I didn't put engine.py in native was because current OTX has engine.py nested in a folder also contains utis like adaptive batch size finder.

src/otx/engine/__init__.py
ashwinvaidya17: The folder structure is based on this design
image
It is in turn based on the Keras design as they also support multiple backends where the base classes/front-facing API remain in `src/` and the actual backend implementations mimicking the root but inside `backend`

ashwinvaidya17: Also, it isn't really empty as there is `engine.py` in the same folder

eugene123tw (approved):

kprokofi (commented):
@ashwinvaidya17, thank you for this PR. I have a question. Many things like algo/ and models from core folder have not been migrated to native backend. Will they be considered in the next PR?

ashwinvaidya17:
> @ashwinvaidya17, thank you for this PR. I have a question. Many things like algo/ and models from core folder have not been migrated to native backend. Will they be considered in the next PR?

Yeah. I want to move those files piecemeal. Some of the files will be deleted when we move to anomalib engine. However, others could use a refactor.

kprokofi (approved):

PR #4347, opened 2025-04-15
Numpy DataItem

eugene123tw:
### Checklist

### License

```python
```


src/otx/core/data/dataset/base.py
```diff
@@ -86,6 +90,7 @@ def __init__(
         stack_images: bool = True,
         to_tv_image: bool = True,
         data_format: str = "",
+        collate_mode: Literal["torch", "numpy"] = "torch",
```
ashwinvaidya17: I am in favour of using enumerations as strings are quite susceptible to typos thus introducing errors. We can also just type this as 
```python
collate_mode: CollateMode | str = CollateMode.Torch,
```

And below,
```python
self.collate_mode = CollateMode(collate_mode)
```

src/otx/data/numpy/numpy.py
```diff
@@ -0,0 +1,79 @@
...
+            masks=[item.masks for item in items],
+            polygons=[item.polygons for item in items],  # type: ignore[misc]
+            imgs_info=[item.img_info for item in items],
+        )
+
+    def __iter__(self) -> Iterator[str]:
+        for field_ in fields(self):
+            yield field_.name
+
+    def __getitem__(self, key: str) -> Any:  # noqa: ANN401
```
ashwinvaidya17, line 68: Is there a reason we want to allow dictionary access?

src/otx/core/data/dataset/base.py
```diff
@@ -257,7 +263,97 @@ def _cache_img(self, key: str | int, img_data: np.ndarray, meta: dict[str, Any]
...
+    def postprocess(
+        self,
+        image: np.ndarray,
+        img_info: ImageInfo,
+        label: np.ndarray | None = None,
+        bboxes: np.ndarray | None = None,
+        masks: np.ndarray | None = None,
+        polygons: list[dict[str, Any]] | None = None,
+        keypoints: np.ndarray | None = None,
+    ) -> T_OTXDataEntity:
```
ashwinvaidya17:
```suggestion
    ) -> TorchDataItem | NumpyDataItem:
```

src/otx/core/data/dataset/base.py
```diff
@@ -257,7 +263,97 @@ def _cache_img(self, key: str | int, img_data: np.ndarray, meta: dict[str, Any]
...
+
+            entity = TorchDataItem(
+                image=image,
+                label=label,
+                masks=masks,
+                bboxes=bboxes,
+                keypoints=keypoints,
+                polygons=polygons,
+                img_info=img_info,
+            )
```
ashwinvaidya17: Can we split this and the NumpyDataItem conversion to two separate methods?

src/otx/core/data/dataset/base.py
```diff
@@ -257,7 +263,97 @@ def _cache_img(self, key: str | int, img_data: np.ndarray, meta: dict[str, Any]
...
+            image: Image data.
+            label: Label data.
+            masks: Mask data.
+            bboxes: Bounding box data.
+            keypoints: Keypoint data.
+            polygons: Polygon data.
+            img_info: Image information.
+
+        Returns:
+            A data item with the provided data.
```
ashwinvaidya17: Maybe we should mention here that `postprocess` does not apply transforms when using numpy data item.

ashwinvaidya17 (commented):
Thanks for the efforts. It looks good overall. I have a few minor comments

src/otx/data/numpy/numpy.py
eugene123tw: I'm not sure as it's almost the same as `TorchDataItem`. Do we need dictionary access? If we don't I can remove both TorchDataItem and NumpyDataItem.

ashwinvaidya17: Um yeah I missed it. Let's keep it for now

ashwinvaidya17 (dismissed):
Thanks! I am fine with the changes

kprokofi (changes requested):
Thanks for the update, Eugene. That said, I feel now the NumpyDataItem is starting to overcomplicate things. We're now duplicating logic by preprocessing and collating everything separately into NumPy arrays. On top of that, for IR validation, we still need the annotations as PyTorch tensors since we're using torchmetrics for evaluation.

Would it make sense to support only NumPy images with optional transforms, and keep the annotations in torch format? From the perspective of the OV Engine and model implementations, it seems more consistent that way.
What do you think?

eugene123tw:
> Thanks for the update, Eugene. That said, I feel now the NumpyDataItem is starting to overcomplicate things. We're now duplicating logic by preprocessing and collating everything separately into NumPy arrays. On top of that, for IR validation, we still need the annotations as PyTorch tensors since we're using torchmetrics for evaluation.
> 
> Would it make sense to support only NumPy images with optional transforms, and keep the annotations in torch format? From the perspective of the OV Engine and model implementations, it seems more consistent that way. What do you think?

@kprokofi I think we should avoid depending on torchmetrics for IR inference/validation. Using torchmetrics requires converting both predictions and annotations into PyTorch tensors, which doesn’t make sense for IR tasks — the outputs are typically generic (e.g., lists, NumPy arrays) and shouldn't be tied to a specific deep learning framework.

I also feel mixing NumPy arrays and torch tensors within a single data entity isn’t clean and could introduce bugs or unnecessary complexity.

A more framework-agnostic alternative could be HuggingFace’s evaluate library (https://huggingface.co/docs/evaluate/index), which supports segmentation, object detection, and other tasks without depending on any particular DL framework.

As for the duplication concern: if we remove `torchmetrics` dependency for IR, we wouldn’t need to collate annotations separately into tensors — we could simply pass datasets with NumPy data items directly to the evaluator, without needing to wrap them in a PyTorch DataLoader at all. This would simplify the flow.

What do you think?

PR #4369, opened 2025-05-01
🧹 Refactor base train and test recipe

ashwinvaidya17:
### Summary

- 🗑️Removes base train and test yaml in favour of API based initialization
- ✨ Default callbacks and loggers are set within the Engine.

### Checklist

### License

```python
```


src/otx/engine/engine.py
```diff
@@ -1104,31 +1114,88 @@ def trainer(self) -> Trainer:
...
+            callbacks.append(RichProgressBar(refresh_rate=1, leave=False))
+        if not has_callback(RichModelSummary):
+            callbacks.append(RichModelSummary(max_depth=1))
+        if not has_callback(IterationTimer):
+            callbacks.append(IterationTimer(prog_bar=True, on_step=False, on_epoch=True))
+        if not has_callback(LearningRateMonitor):
+            callbacks.append(LearningRateMonitor(logging_interval="epoch", log_momentum=True))
+        if not has_callback(AdaptiveTrainScheduling):
+            callbacks.append(
+                AdaptiveTrainScheduling(
```
kprokofi, line 1197: This AdaptiveTrainScheduling is conflicting with "check_val_every_n_epoch" that you introduced. Inside this callback we rewrite intervals for validation based on the dataset size. We also rewrite lr schedule patience and early stopping patience.  IMO, this callback introduces sometimes not optimal heuristic. How do you think, can we avoid using it?

src/otx/recipe/detection/atss_mobilenetv2_tile.yaml
```diff
@@ -32,6 +32,15 @@ callback_monitor: val/map_50
 data: ../_base_/data/detection_tile.yaml
 overrides:
```
kprokofi: Do we still need this "overrides" section? Won't it work without putting this under overrides?

src/otx/recipe/detection/atss_mobilenetv2.yaml
```diff
@@ -38,6 +38,14 @@ overrides:
         max_interval: 5
         decay: -0.025
         min_lrschedule_patience: 3
+    - class_path: otx.algo.callbacks.adaptive_early_stopping.EarlyStoppingWithWarmup
```
kprokofi: Can we register this callback in the Engine as default one?

src/otx/tools/converter.py
```diff
@@ -174,41 +174,9 @@
...
-    "ote_anomaly_segmentation_padim": {
-        "task": OTXTaskType.ANOMALY_SEGMENTATION,
-        "model_name": "padim",
-    },
-    "ote_anomaly_segmentation_stfpm": {
-        "task": OTXTaskType.ANOMALY_SEGMENTATION,
-        "model_name": "stfpm",
-    },
-    # KEYPOINT_DETECTION
-    "Keypoint_Detection_RTMPose_Tiny": {
```
kprokofi: Why did you delete KP Detection ?

ashwinvaidya17: Ah that is a mistake 🙃

src/otx/recipe/detection/atss_mobilenetv2_tile.yaml
ashwinvaidya17: I guess it should work for callbacks. But we still need it for data right?

src/otx/engine/engine.py
ashwinvaidya17: Hmm I am not sure how it worked earlier. The `train.py` had check_val_every_n_epoch` and `AdaptiveTrainScheduling`.

ashwinvaidya17: If the team decides that we can remove it then I am all for it

src/otx/recipe/detection/atss_mobilenetv2.yaml
ashwinvaidya17: Yeah I wanted to discuss this. I feel this should be set from the `configure_callbacks` methods of the lightning module (base otx module?). Not every model would require early stopping so maybe we should not add it via engine.

ashwinvaidya17: Actually, I had forgotten why I didn't move it to the lightning model. It is because `configure_callbacks` gets priority over CLI. So, if we move this to lightning module, we won't be able to override it. Maybe we can explore
```yaml
trainer: ../__base__/trainer.yaml
data: ../__base__/data.yaml
```

src/otx/engine/engine.py
```diff
@@ -1104,31 +1114,88 @@ def trainer(self) -> Trainer:
...
+                    save_dir=self.work_dir,
+                    name="tensorboard/",
+                    log_graph=False,
+                    default_hp_metric=True,
+                    prefix="",
+                ),
+            ]
+        return logger
+
+    def _setup_otx_callbacks(self) -> None:
```
ashwinvaidya17: change to `configure_callbacks`

kprokofi (approved):

kprokofi:
Hey @ashwinvaidya17 is it ready to merge?

src/otx/recipe/anomaly/padim.yaml
```diff
@@ -14,10 +14,18 @@ engine:
...
+callbacks:
+  - class_path: otx.algo.callbacks.adaptive_train_scheduling.AdaptiveTrainScheduling
+    init_args:
+      max_interval: 1
+  - class_path: lightning.pytorch.callbacks.ModelCheckpoint
+    init_args:
+      dirpath: "" # use engine.work_dir
+      save_last: true
+      auto_insert_metric_name: false
+      filename: "checkpoints/epoch_{epoch:03d}"
```
Daankrol, line 27: why are we removing the default train/test.yaml model files? This now results in a lot of duplicate code/config in each model file.

ashwinvaidya17: Because we rely on train and test yaml for everything. There are some parameters set in the base yaml files that are not needed by some of the models. Additionally, it sets the parameters silently. If I remember correctly this caused some issues with integrating the uflow model. Imo configs should not be mandatory. Each model should have the default parameters set so that we should be able to just use three lines to create the entities.
```python
model = ModelA()
dataloader = dataloaderX()
engine = create_engine(model, dataloader)
engine.train()
```
Configs should be optional rather than mandatory. Originally all this is based on the lightning CLI design. Anomalib also follows it. The configs are more of an example https://github.com/open-edge-platform/anomalib/blob/main/examples/configs/model/uflow.yaml
If we keep train and test yaml then the CLI becomes mandatory rather than optional.

Daankrol (approved):

PR #4414, opened 2025-06-11
[feature-v3] Integrate "create_engine" to tests and polish engine interface

kprokofi:
### Summary

- [x]  Integrate get_engine into Geti interaction tests
- [x]  Refactor Engine class: add more abstract methods such as datamodule and model
- [x]  Update OTXEngine interface to align with the abstract Engine and OVEngine — accept data and instantiate a checkpoint immediately if provided via the constructor, remove "task" parameter. We now derive task info from config or OTXModel directly.
- [x]  Add tests for create_engine
- [x] Simplify AutoConfigurator and Engine interface. The Engine now requires model and data as mandatory parameters.
Instead of using an arbitrary name for the native engine, we now pass a path to the configuration file. This makes the logic clearer and eliminates unnecessary heuristics like "default tasks and models."
It is also now possible to pass any configuration file—not just those located in the recipe folder.

NOTE:

- create_engine will support custom engines, as long as the user defines them by inheriting from the base Engine class.

- I manually registered OTXEngine and OVEngine to avoid issues with Engine.__subclasses__(). Since __subclasses__() only returns classes that have already been imported, relying on it alone would miss any engines not yet loaded at runtime. Therefore, we need to explicitly maintain the list of supported engines.

### Checklist

### License

```python
```


tests/unit/backend/native/test_engine.py
```diff
@@ -0,0 +1,328 @@
+# Copyright (C) 2024 Intel Corporation
```
sovrasov: 2025 here and in other new files

kprokofi: Done

src/otx/backend/native/engine.py
```diff
@@ -130,12 +128,12 @@ def __init__(
         self.device = device  # type: ignore[assignment]
         self.num_devices = num_devices
         self._auto_configurator = AutoConfigurator(
-            data_root=data_root,
-            task=datamodule.task if datamodule is not None else task,
+            data_root=data if isinstance(data, (str, os.PathLike)) else None,
+            task=data.task if isinstance(data, OTXDataModule) else task,
```
sovrasov: We need to mention this behavior in docstring: if dataloader is passed as data, `task` argument is ignored.

src/otx/backend/native/engine.py
```diff
@@ -100,11 +100,9 @@ class OTXEngine(Engine):
 
     def __init__(
         self,
-        *,
-        data_root: PathLike | None = None,
+        data: OTXDataModule | PathLike | None = None,
```
sovrasov: Is it possible to drop support of `PathLike` here? If we're switching to dataloader, passing a path doesn't make sense

src/otx/cli/cli.py
```diff
@@ -175,7 +175,7 @@ def engine_subcommand_parser(subcommand: str, **kwargs) -> tuple[ArgumentParser,
         parser.add_class_arguments(Workspace, "workspace")
         parser.link_arguments("work_dir", "workspace.work_dir")
 
-        parser.link_arguments("data_root", "engine.data_root")
+        # parser.link_arguments("data_root", "engine.data")
```
sovrasov: Pls del the dead code

src/otx/backend/native/engine.py
```diff
@@ -1041,6 +1042,28 @@ def datamodule(self) -> OTXDataModule:
             raise RuntimeError(msg)
         return self._datamodule
 
+    @datamodule.setter
+    def datamodule(self, data: OTXDataModule | PathLike) -> None:
+        """Sets the datamodule for the engine.
```
sovrasov: Is such setter safe if engine has already been constructed with another dataloader?

src/otx/types/types.py
```diff
@@ -6,10 +6,12 @@
...
 from otx.data.entity import OTXDataItem
 from otx.data.module import OTXDataModule
+from otx.types import PathLike
 
 METRICS = dict[str, float]
 ANNOTATIONS = list[OTXDataItem]
-MODEL = OTXModel  # TODO(ashwinvaidya17, kprokofi): Temporary till model is properly defined
-DATA = OTXDataModule  # TODO(ashwinvaidya17, kprokofi): Temporary till data is properly defined
+MODEL = OTXModel | OVModel | PathLike
+DATA = OTXDataModule | PathLike
```
sovrasov, line 17: Why do we need this alias if it is not used for instance in `src/otx/backend/native/engine.py`?

src/otx/backend/native/engine.py
kprokofi: Path is also an option here. If we pass a path to a dataset, Engine will construct the datamodule automatically

src/otx/types/types.py
kprokofi: It is used in the base Engine to define types. Each specific Engine, on the other hand, returns specific class

src/otx/backend/native/engine.py
sovrasov: I see, but the question is can we drop the path type? It makes the API unclear, as there is a complicated logic to handle both types. It is also related to my previous comment: the logic we have is even hard to describe in the docs

src/otx/backend/native/engine.py
```diff
@@ -153,6 +151,11 @@ def __init__(
         self._model: OTXModel = (
             model if isinstance(model, OTXModel) else self._auto_configurator.get_model(**get_model_args)
         )
+        if self.checkpoint:
+            if not isinstance(self.checkpoint, (Path, str)) and not Path(self.checkpoint).exists():
+                msg = f"Checkpoint {self.checkpoint} does not exist."
+                raise FileNotFoundError(msg)
+            self.model.load_state_dict_incrementally(torch.load(self.checkpoint))
```
eugene123tw: Could you move the checkpoint logic closer to the top, where the other variables are initialised? I think something like this would be nice:
```py
if checkpoint:
    if not isinstance(checkpoint, (str, Path)) and not Path(checkpoint).exists():
        raise FileNotFoundError(f"Checkpoint file not found: {checkpoint_path}")
    self.checkpoint = checkpoint
    self.model.load_state_dict_incrementally(torch.load(checkpoint))
```

tests/unit/backend/native/test_engine.py
```diff
@@ -0,0 +1,325 @@
...
+        assert isinstance(fxt_engine.model, EfficientNetMulticlassCls)
+
+    def test_training_with_override_args(self, fxt_engine, mocker) -> None:
+        mocker.patch("pathlib.Path.symlink_to")
+        mocker.patch("otx.backend.native.engine.Trainer.fit")
+        mock_seed_everything = mocker.patch("otx.backend.native.engine.seed_everything")
+
+        assert fxt_engine._cache.args["max_epochs"] == 90
+
+        fxt_engine.train(max_epochs=100, seed=1234)
```
eugene123tw: Is it necessary to train for such a high number of epochs (e.g., 90 or 100)? Could a smaller number suffice for unit test (e.g 5-10)?

eugene123tw (commented):
Thanks for the updates! I’ve left two minor comments. Overall, the changes look good, just a couple of small clarifications to consider before merging. Let me know if you have any questions!

src/otx/backend/native/engine.py
kprokofi: I removed "task" parameter and made "model" and "data" mandatory parameters. 
Data can be a path to a dataset or OTXDatamodule itself
Model can be a path to OTX config or OTXModel itself

src/otx/backend/native/engine.py
kprokofi: I removed setters for datamodule and model. To avoid any issues with messed up data and models.

src/otx/backend/native/engine.py
kprokofi: I moved the assignment of the checkpoint closer to where it is loaded into the model. However, it’s not possible to move it further up, since the model depends on the AutoConfigurator and DataModule being instantiated first. The model requires information from the DataModule, so those components must be initialized before the model is created.

tests/unit/backend/native/test_engine.py
kprokofi: We mock the "fit" function, so there is no actual training.

kprokofi: Anyway, I decreased it

sovrasov (approved):

PR #4417, opened 2025-06-19
Update module docstrings for consistency across various files

### LINKED ISSUE #4415: Inconsistent license headers + docstrings
## Problem

The repo does not have a consistent format for license headers and docstrings. Here are some examples from various files that follows different approaches. 

Example - 1:
https://github.com/open-edge-platform/training_extensions/blob/02f32fe40e81e057361aaef4c7e8cdecec394eaf/src/otx/data/torch/torch.py#L1-L4

Example - 2: 
https://github.com/open-edge-platform/training_extensions/blob/02f32fe40e81e057361aaef4c7e8cdecec394eaf/src/otx/core/config/data.py#L1-L6

Example - 3: 
https://github.com/open-edge-platform/training_extensions/blob/02f32fe40e81e057361aaef4c7e8cdecec394eaf/src/otx/core/config/explain.py#L1-L4

And there are some others as well...

## Solution
Ideally these should be standardized throughout the repo such that each file has the same format. My personal preference would be docstring followed by a license header. Docstring would define what the file is about, which is followed by its license headers. Of course this is subjective, and people might have their own preferences.

```python
"""Some explanation about what the module is about."""

from package import module
```

eugene123tw:
### Summary

Update module docstrings for consistency across various files

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/algo/__init__.py
+++ b/src/otx/algo/__init__.py
@@ -1,6 +1,6 @@
 # Copyright (C) 2023 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
-#
+
 """Module for OTX custom algorithms, e.g., model, losses, hook, etc..."""
 
 from . import (
--- a/src/otx/algo/modules/scale.py
+++ b/src/otx/algo/modules/scale.py
@@ -1,7 +1,8 @@
 # Copyright (C) 2024 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
+
 # Copyright (c) OpenMMLab. All rights reserved.
-# https://github.com/open-mmlab/mmcv/blob/98e7b3ab6088117600773245b8c6541301bfab95/mmcv/cnn/bricks/scale.py
+
 """Utils from mmcv."""
 
 import torch
--- a/src/otx/__init__.py
+++ b/src/otx/__init__.py
@@ -1,7 +1,7 @@
-"""OpenVINO Training Extensions."""
-
 # Copyright (C) 2024-2025 Intel Corporation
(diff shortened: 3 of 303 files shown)


samet-akcay (dismissed):
Thanks a lot!

src/otx/algo/__init__.py
```diff
@@ -1,6 +1,6 @@
 # Copyright (C) 2023 Intel Corporation
```
rajeshgangireddy, line 1: Should we update this to 2024-2025 (or 2025) ?

src/otx/algo/modules/scale.py
```diff
@@ -1,7 +1,8 @@
 # Copyright (C) 2024 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
+
 # Copyright (c) OpenMMLab. All rights reserved.
-# https://github.com/open-mmlab/mmcv/blob/98e7b3ab6088117600773245b8c6541301bfab95/mmcv/cnn/bricks/scale.py
```
rajeshgangireddy, line 4: Don't we have to put this info somewhere else?

src/otx/algo/__init__.py
eugene123tw: I don't think we need to update it as it's the year the code was created, and we haven't change any code. We'd change the year only when there's code changes.

samet-akcay (approved):
Thanks for the effort, Eugene

sovrasov (approved):

rajeshgangireddy (approved):

PR #4433, opened 2025-06-26
Fix loading/saving checkpoints in OTX

kprokofi:
### Summary

ITEP-70170

## Problem
OTX checkpoints store many internal entities, making it difficult to safely load them using torch.load. After recent refactoring (e.g., removal of TrainType and UnlabeledDataset), loading old checkpoints can completely fail due to missing modules and classes.

## Solution
Several approaches were explored, including writing a custom pickle Unpickler, but most had significant limitations and error cases.
The final solution enables safe checkpoint loading using weights_only=True, which prevents deserialization of unsafe or deprecated objects, reducing security and stability risks.

To support backward compatibility with older snapshots (e.g., for Geti), the solution also includes creating fake modules for entities that no longer exist in the codebase.

## Additional Fixes

- [x]  Prevented direct saving of internal OTX entities into the checkpoint. Now only safe, allowed types are saved.

- [x]  Omitted saving the DataModule hyperparameters, since they redundantly include all OTXDataModule components such as configs and augmentations.
- [x] Removed unnecessary and duplicated hyperparameters (e.g., TilingConfig and LabelInfo were previously saved twice).
- [x] Changed the way checkpoints are loaded for export, test, predict, and explain. Previously, we relied on entities embedded in the checkpoint itself. As the codebase evolves rapidly, this approach is brittle. The new approach loads only the model's state dict and applies it to an explicitly constructed model instance, avoiding dynamic recreation of classes from the checkpoint.

## Outcome
Starting from v2.4.5, checkpoints can be safely and cleanly loaded using the default torch.load(..., weights_only=True), with improved backward compatibility and reduced snapshot size.

### Checklist

### License

```python
```


sovrasov:
Thanks @kprokofi ! I think we need to add at least one real-life snapshot loafing from 2.2.2 to the tests. That'd set compatibility requirement. I can host a snapshot on MAPI's s3 if that'd help

kprokofi:
> Thanks @kprokofi ! I think we need to add at least one real-life snapshot loafing from 2.2.2 to the tests. That'd set compatibility requirement. I can host a snapshot on MAPI's s3 if that'd help

Done

sovrasov (approved):

sovrasov (approved):

PR #4441, opened 2025-07-07
Provide workarounds for the XPU training

kprokofi:
### Summary
Resolves [ITEP-71393](https://jira.devtools.intel.com/browse/ITEP-71393), [ITEP-71392](https://jira.devtools.intel.com/browse/ITEP-71392)

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/__init__.py
+++ b/src/otx/__init__.py
@@ -3,7 +3,7 @@
 
 """OpenVINO Training Extensions."""
 
-__version__ = "2.5.0dev"
+__version__ = "2.6.0dev"
 
 import os
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


src/otx/__init__.py
```diff
@@ -17,6 +17,9 @@
     "HF_HUB_CACHE",
     str(Path.home() / ".cache" / "torch" / "hub" / "checkpoints"),
 )
+# Set the value of ONEDNN_PRIMITIVE_CACHE_CAPACITY to set the cache capacity for oneDNN primitives.
+# It will be ignored if no XPU devices are available.
+os.environ["ONEDNN_PRIMITIVE_CACHE_CAPACITY"] = "10000"
```
sovrasov, line 22: We could be not the only one oneDNN clients. I'd propose not to set this var without a reason to avoid side-possible effects for non-xpu users: the line can be protected with `hasattr(torch, "xpu") and torch.xpu.is_available()`

kprokofi: The problem is that we need to set env variable before importing torch

kprokofi: the second option will be to add setting env variable directly in Docker image in Geti. But it means for OTX usage we always need to set this variable manually to run some models like RTDetr, DFine, etc.

sovrasov: I see, let's have this env setter here, since it's hard to fine system-agnostic way of finding Intel GPU. But we need to warn users somehow: it can be a note in the installation section, or one can print a warning when importing OTX. The second option is annoying though

sovrasov (approved):

eugene123tw (approved):

PR #4446, opened 2025-07-09
DEIM-DFine

eugene123tw:
### Summary
This PR introduces DEIM-DFine object detection model and implements a dynamic data augmentation switching mechanism that adapts augmentation policies during training based on epoch progression.

New Features
1. DEIM-DFine Model Implementation
* New Model: Added DEIMDFine class in `src/otx/backend/native/models/detection/deim.py`
* Model Variants: Support for multiple 3 DEIM-DFine variants (x, l, m)
* Configuration Files: Added recipe configurations:
  * `src/otx/recipe/detection/deim_dfine_l.yaml`
  * `src/otx/recipe/detection/deim_dfine_x.yaml`

2. Dynamic Data Augmentation Switch
  * Adaptive Training: Implements epoch-based augmentation policy switching 
  * 3-Stage Training Pipeline (DEIM-DFine-X for instance):
        Stage 1 (epochs 0-3): No augmentation (no_aug)
        Stage 2 (epochs 4-28): Strong augmentation with random selection (strong_aug_1, strong_aug_2)
        Stage 3 (epochs 29+): Light augmentation (light_aug)
  * Multiprocessing Safe: Uses multiprocessing.Value for shared epoch state across dataloaders

#### Core Components
* `DataAugSwitch` Class:
```py
class DataAugSwitch:
    """Manages dynamic switching between augmentation policies based on training epochs."""
    
    def __init__(self, policy_epochs: list[int], policies: dict[str, dict[str, Any]]):
        # Expects 3 policy epochs for 3-stage scheduler (e.g., [4, 29, 50])
        # Policies define transforms for each augmentation stage
```
  * Epoch-based Policy Selection: Automatically selects appropriate augmentation policy
  * Shared Memory Integration: Thread-safe epoch tracking via multiprocessing.Value
* `AugmentationSchedulerCallback`:
```python
class AugmentationSchedulerCallback(Callback):
    """PyTorch Lightning callback that updates epoch information for DataAugSwitch."""
    
    def on_train_epoch_start(self, trainer: Trainer, pl_module: LightningModule) -> None:
        self.data_aug_switch.epoch = trainer.current_epoch
```
* `DataAugSwitchMixin` with `OTXDetectionDataset` integration
```py
class DataAugSwitchMixin:
    """Mixin providing dynamic augmentation switching functionality to datasets."""
    
    def _apply_augmentation_switch(self, entity: OTXDataItem) -> OTXDataItem:
        if self.data_aug_switch is not None:
            self.to_tv_image, self.transforms = self.data_aug_switch.current_transforms
        return entity
```

```py
class OTXDetectionDataset(OTXDataset, DataAugSwitchMixin):  # type: ignore[misc]
    def _get_item_impl(self, index: int) -> OTXDataItem | None:
        # ... data loading logic ...
        if self.has_dynamic_augmentation:
          self._apply_augmentation_switch()  # ← Dynamic augmentation
        return self._apply_transforms(entity)
```

* `OTXEngine` Integration
  * Engine automatically detects `AugmentationSchedulerCallback` and initialises shared memory in `configure_callbacks`
  * Automatically configures datasets with DataAugSwitch if callback is present

```py
def configure_callbacks(self) -> None:
    # ... existing callback setup ...
    
    # Setup shared memory for dynamic augmentation
    if aug_scheduler_callback is not None and aug_scheduler_callback.data_aug_switch is not None:
        shared_epoch = Value("i", 0)
        aug_scheduler_callback.data_aug_switch.set_shared_epoch(shared_epoch)

def _configure_datasets_with_data_aug_switch(self) -> None:
    """Configure datasets with DataAugSwitch if available."""
    if hasattr(self.datamodule, "subsets") and "train" in self.datamodule.subsets:
        if isinstance(self.datamodule.subsets["train"], DataAugSwitchMixin):
            self.datamodule.subsets["train"].set_data_aug_switch(data_aug_switch)
```

### How to test

* test_aug_scheduler.py: DataAugSwitch and AugmentationSchedulerCallback 
* test_mixins.py: DataAugSwitchMixin functionality 
* test_detection_aug_switch.py: Integration tests 

### Checklist

### License

```python
```


src/otx/backend/native/models/detection/losses/deim_loss.py
```diff
@@ -0,0 +1,73 @@
+# Copyright (C) 2025 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+#
```
samet-akcay:
```suggestion

```

src/otx/backend/native/callbacks/aug_scheduler.py
```diff
@@ -0,0 +1,105 @@
...
+from otx.data.transform_libs.torchvision import Compose, TorchVisionTransformLib
+
+if TYPE_CHECKING:
+    from multiprocessing import Value
+
+    from lightning.pytorch import LightningModule, Trainer
+
+
+class DataAugSwitch:
+    """Data augmentation switch."""
```
samet-akcay: can you expand the docstring about what this callback does with examples section in the docstring etc.?

src/otx/backend/native/callbacks/aug_scheduler.py
```diff
@@ -0,0 +1,105 @@
...
+    @property
+    def current_transforms(self) -> tuple[bool, Compose]:
+        """Get the current transforms."""
+        name = self.current_policy_name
+        policy = self.policies.get(name)
+        return policy["to_tv_image"], policy["transforms"]  # type: ignore[index]
+
+
+class AugmentationSchedulerCallback(Callback):
+    """Callback for data augmentation scheduler."""
```
samet-akcay: Same here. it would be good to have a docstring with examples section to really understand the effect.

samet-akcay: Is this callback deim/dfine specific, or can it be applied to any model?

src/otx/backend/native/engine.py
```diff
@@ -1058,6 +1064,61 @@ def configure_callbacks(self) -> None:
 
         self._cache.args["callbacks"] = callbacks + config_callbacks
 
+        # Setup DataAugSwitch with shared multiprocessing.Value
+        self._setup_augmentation_scheduler()
+
+    def _setup_augmentation_scheduler(self) -> None:
+        """Set up shared memory for DataAugSwitch and AugmentationSchedulerCallback."""
```
samet-akcay: can you expand the docstring a bit more to explain why this method is needed to share memory?

src/otx/backend/native/engine.py
```diff
@@ -1058,6 +1064,61 @@ def configure_callbacks(self) -> None:
...
+                aug_scheduler_callback = callback
+                break
+
+        # If AugmentationSchedulerCallback exists and has a data_aug_switch, set up shared memory
+        if aug_scheduler_callback is not None and aug_scheduler_callback.data_aug_switch is not None:
+            # Create shared multiprocessing.Value for epoch tracking
+            shared_epoch = Value("i", 0)
+            aug_scheduler_callback.data_aug_switch.set_shared_epoch(shared_epoch)
+
+    def _setup_data_aug_switch_for_datasets(self) -> None:
```
samet-akcay, line 1123: same as above, can you expand the docstring here as well?

src/otx/backend/native/engine.py
```diff
@@ -1058,6 +1064,61 @@ def configure_callbacks(self) -> None:
 
         self._cache.args["callbacks"] = callbacks + config_callbacks
 
+        # Setup DataAugSwitch with shared multiprocessing.Value
+        self._setup_augmentation_scheduler()
+
+    def _setup_augmentation_scheduler(self) -> None:
```
samet-akcay, line 1070: i'm not crystal clear on why data augmentation stuff is needed to be handled here in the engine.

This might be the same for future readers as well. Maybe we could explicitly explain here why it is really needed here to be setup?

src/otx/backend/native/models/detection/deim.py
```diff
@@ -0,0 +1,151 @@
...
+
+if TYPE_CHECKING:
+    from lightning.pytorch.cli import LRSchedulerCallable, OptimizerCallable
+
+    from otx.backend.native.schedulers import LRSchedulerListCallable
+    from otx.metrics import MetricCallable
+    from otx.types.label import LabelInfoTypes
+
+
+class DEIMDFine(RTDETR):
```
samet-akcay, line 30: can you explain why we call it `DEIMDFine`?

src/otx/backend/native/models/detection/deim.py
```diff
@@ -0,0 +1,151 @@
...
+        "deim_dfine_hgnetv2_m": "https://github.com/eugene123tw/DEIM/releases/download/poc/deim_dfine_hgnetv2_m_coco_90e.pth",
+        "deim_dfine_hgnetv2_l": "https://github.com/eugene123tw/DEIM/releases/download/poc/deim_dfine_hgnetv2_l_coco_50e.pth",
+        "deim_dfine_hgnetv2_x": "https://github.com/eugene123tw/DEIM/releases/download/poc/deim_dfine_hgnetv2_x_coco_50e.pth",
+    }
+    input_size_multiplier = 32
+
+    def __init__(
+        self,
+        label_info: LabelInfoTypes,
+        data_input_params: DataInputParams,
```
samet-akcay, line 72: I mention this to @kprokofi, and am repeating myself, but it would be great to be able to just instantiate this model without passing any arguments. This would instantiate the model with the default args that are reported in the paper ?

src/otx/backend/native/models/detection/deim.py
samet-akcay: let's say I'm a user who is explicitly interested in one of the models, say just Deim? Won't I be able to import `from otx.models import Deim`, or `from otx.models import Dfine`?

src/otx/data/dataset/detection.py
```diff
@@ -14,9 +14,10 @@
 from otx.data.entity.torch import OTXDataItem
 
 from .base import OTXDataset
+from .mixins import DataAugSwitchMixin
 
 
-class OTXDetectionDataset(OTXDataset):
+class OTXDetectionDataset(OTXDataset, DataAugSwitchMixin):  # type: ignore[misc]
```
samet-akcay, line 20: is this augmentation switch mixing needed here because otx cannot handle preprocessing within the model pipeline?

samet-akcay (commented):
Thanks for the great effort @eugene123tw. Overall, it looks great. I've got some questions/comments in some parts especially the name of the model, and data augmentation switch implementation. 

let's discuss. thanks again

src/otx/data/dataset/detection.py
eugene123tw: yeah, OTXDataset processes items by retrieving a single sample and applying a list of transforms within each DataLoader. Multiple DataLoaders run in parallel for efficiency, so applying transforms in parallel is significantly faster than iterating over samples sequentially. `DataAugSwitchMixin` is here so we could dynamically switch between different transform pipelines before transform happens.

src/otx/backend/native/callbacks/aug_scheduler.py
eugene123tw: I added docstring and example for DataAugSwitch, not sure if it's too verbose or clear enough. Please let me know!

src/otx/backend/native/callbacks/aug_scheduler.py
eugene123tw: Added docstring and example. This can be use for every detection model, but have to be used with `OTXEngine`

src/otx/backend/native/engine.py
eugene123tw: I added more description to it.

src/otx/backend/native/models/detection/deim.py
eugene123tw: It's called DEIM, short for DETR with Improved Matching. We refer to our variant as DEIM-DFine, since there are other versions like DEIM-RTDETR. Essentially, DEIM-DFine means applying the DEIM framework on the DFine architecture with DFine loss, while DEIM-RTDETR applies it to the RT-DETR architecture using RT-DETR's loss functions.

eugene123tw: > let's say I'm a user who is explicitly interested in one of the models, say just Deim? Won't I be able to import `from otx.models import Deim`, or `from otx.models import Dfine`?

If users are interested in the DEIM variants, such as the DEIM-DFine series (5 variants), you would import like this:

```py
from otx.models import DEIMDFine
model = DEIMDFine(variant="x")  # where "x" refers to the xlarge architecture
```

If you're only interested in the standard DFine-X (non-DEIM), you'd use:

```py
from otx.models import DFine
model = DFine(variant="x")
```

src/otx/backend/native/engine.py
eugene123tw: Added

src/otx/data/dataset/mixins.py
```diff
@@ -0,0 +1,68 @@
...
+                # ... get your data ...
+                self._apply_augmentation_switch()
+                return self._apply_transforms(entity)
+    """
+
+    def _ensure_data_aug_switch_initialized(self) -> None:
+        """Ensure data_aug_switch attribute is initialized.
+
+        This method is called lazily since __init__ may not be called
+        due to multiple inheritance MRO in some dataset classes.
```
ashwinvaidya17, line 32: Maybe we should add a TODO here. This looks like a workaround for the current design. We should fix it and return to this again.

ashwinvaidya17 (approved):
Thanks for the efforts! One very minor comment

src/otx/data/dataset/mixins.py
eugene123tw: Thanks. We can sort that out in the future.

src/otx/backend/native/engine.py
eugene123tw: Added

src/otx/backend/native/callbacks/aug_scheduler.py
```diff
@@ -0,0 +1,195 @@
...
+    This class manages multiple data augmentation policies and switches between them
+    based on the current training epoch. It is designed to support multi-stage augmentation
+    strategies, such as starting with no augmentation, then applying strong augmentations,
+    and finally switching to lighter augmentations as training progresses.
+
+    The switch is typically used in conjunction with a callback (e.g., AugmentationSchedulerCallback)
+    that updates the current epoch, allowing the augmentation policy to change automatically
+    as training advances.
+
+    Args:
```
Daankrol, line 34: If we always expect three values, maybe we should use a 3-value tuple for better type checking. If we support more stages we might want to update the docs here.

Daankrol (approved):
Great PR, thanks! 🔥

src/otx/backend/native/models/detection/deim.py
samet-akcay: ok thanks. Lets resolve this for now, and focus on the refactoring later on as we agreed.

samet-akcay (approved):
thanks for the great effort

src/otx/backend/native/callbacks/aug_scheduler.py
eugene123tw: Yeah, I agree. I’ll try to make some improvements for this in a separate PR. Right now, it’s mostly tailored for DEIM, but other detectors could benefit from it too. We just need to find an elegant way to support them as well.

PR #4466, opened 2025-07-16
Add DoRA support for ViT classification model.

gyuilLim:
### Summary

Added DoRA support and replaced `lora` falg with `peft` argument using `Literal["lora", "dora"]`.

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/backend/native/models/classification/backbones/vision_transformer.py
+++ b/src/otx/backend/native/models/classification/backbones/vision_transformer.py
@@ -215,7 +215,7 @@ class VisionTransformer(BaseModule):
         act_layer: LayerType | None = None,
         norm_layer: LayerType | None = None,
         interpolate_offset: float = 0.1,
-        lora: bool = False,
+        peft: Literal["lora", "dora"] | None = None,
     ) -> None:
         super().__init__()
         if isinstance(model_name, str):
@@ -301,24 +301,43 @@ class VisionTransformer(BaseModule):
 
         self.norm = norm_layer(self.embed_dim)
 
-        self.lora = lora
-        if self.lora:
+        self.peft = peft
+        if self.peft is None:
+            pass
+        elif self.peft in ["lora", "dora"]:
             lora_rank = 8
             lora_alpha = 1.0
-            assign_lora = partial(AttentionWithLoRA, rank=lora_rank, alpha=lora_alpha)
-            for block in self.blocks:
-                block.attn.qkv = assign_lora(block.attn.qkv)
 
-            # Freeze all params
+            peft_config: dict[str, dict[str, Any]] = {
+                "lora": {
+                    "assign_func": partial(AttentionWithLoRA, rank=lora_rank, alpha=lora_alpha),
+                    "target_attrs": ["lora_q", "lora_v"],
+                },
+                "dora": {
+                    "assign_func": partial(AttentionWithDoRA, rank=lora_rank, alpha=lora_alpha),
+                    "target_attrs": ["dora_q", "dora_v"],
+                },
+            }
+
+            config = peft_config[self.peft]
+
+            # Apply PEFT to all blocks
+            for block in self.blocks:
+                block.attn.qkv = config["assign_func"](block.attn.qkv)
+
+            # Freeze all parameters
             for param in self.parameters():
                 param.requires_grad = False
 
-            # Unfreeze LoRA layers
+            # Unfreeze PEFT layers
             for block in self.blocks:
-                for param in block.attn.qkv.lora_q.parameters():
-                    param.requires_grad = True
-                for param in block.attn.qkv.lora_v.parameters():
-                    param.requires_grad = True
+                for attr_name in config["target_attrs"]:
+                    peft_layer = getattr(block.attn.qkv, attr_name)
+                    for param in peft_layer.parameters():
+                        param.requires_grad = True
+        else:
+            msg = f"Unsupported `peft` value '{self.peft}', please choose from 'lora' or 'dora'."
+            raise ValueError(msg)
 
     def init_weights(self) -> None:
         """Initializes the weights of the VisionTransformer."""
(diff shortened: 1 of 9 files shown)


src/otx/backend/native/models/classification/backbones/vision_transformer.py
```diff
@@ -301,24 +301,43 @@ def __init__(  # noqa: PLR0913
 
         self.norm = norm_layer(self.embed_dim)
 
-        self.lora = lora
-        if self.lora:
+        self.peft = peft
+        if self.peft is None:
```
kprokofi: PLease, rewrite it in a more clear way. I'd propose to not store possible configuration in a dict. We don't need it. Maybe simpler approach  would look better.

```Python
if peft is not None:
   if not peft in ["lora", "dora"]:
        raise ValueError ...

   attention_func = AttentionWithLoRA if peft == "lora" else AttentionWithDoRA
   for block in self.blocks:
         block.attn.qkv = attention_func(block.attn.qkv, rank = 8, alpha = 1.0)
 
   ....
   for attr_name in [f"{peft}_q, f"{peft}_v"]:
        ....

```

src/otx/backend/native/models/classification/backbones/vision_transformer.py
```diff
@@ -773,10 +792,86 @@ def __init__(self, qkv: Attention, rank: int, alpha: float):
...
+        self.bias = qkv.bias
 
     def forward(self, x: torch.Tensor) -> torch.Tensor:
         """Forward pass of the AttentionWithLoRA."""
         qkv = self.qkv(x)
         qkv[:, :, : self.dim] += self.lora_q(x)
         qkv[:, :, -self.dim :] += self.lora_v(x)
         return qkv
+
+
```
kprokofi: I would propose to move all PEFT related implementation in different folder. src/otx/backend/native/models/classification/utils/peft.py may work for that purpose.

src/otx/backend/native/models/classification/multilabel_models/vit.py
```diff
@@ -96,7 +96,7 @@ def _create_model(self, num_classes: int | None = None) -> nn.Module:
         vit_backbone = VisionTransformer(
             model_name=self.model_name,
             img_size=self.data_input_params.input_size,
-            lora=self.lora,
+            peft=self.peft,
```
kprokofi: You forgot to change parameter name in class constructor. Same for h_label. Also, since we haven't tested it for other types of classification tasks, I would be careful here.  
@sovrasov should we expose possibility to train with lora/dora for other classification tasks? I'd propose to remove it or test to be sure at least that it can be launched.

gyuilLim: @kprokofi  Thank you, I missed that.
Would it be better to remove the peft argument from the constructors in `hlabel` and `multilabel`?

src/otx/backend/native/models/classification/utils/peft.py
```diff
@@ -0,0 +1,116 @@
...
+"""PEFT modules for Vision Transformer attention."""
+
+import torch
+from timm.models.vision_transformer import Attention
+
+
+class LoRALayer(torch.nn.Module):
+    """LoRA layer implementation for computing A, B composition."""
+
+    def __init__(self, in_dim: int, out_dim: int, rank: int, alpha: float):
```
sovrasov: It'd be helpful to have init parameters documented, the same for other classes in the file

src/otx/recipe/classification/multi_class_cls/deit_tiny.yaml
```diff
@@ -4,7 +4,6 @@ model:
   init_args:
     label_info: 1000
     model_name: "vit-tiny"
-    lora: false
```
sovrasov, line 7: Can peft: None be used in DINOv2 recipe to highlight we have this parameter?

src/otx/backend/native/models/classification/multilabel_models/vit.py
sovrasov: I agree we should limit ourselves to multiclass classification for now. The feature is straightforward to scale to other classification tasks. This can be done in future PRs

sovrasov (commented):
@gyuilLim thanks for your efforts! Could you add a brief version of the table with results of the experiments? That'd be useful in the future, so everyone can see the impact of the PR by simply looking at description. 
Also, the PR doesn't contain any tests. Newly added modules should be covered with unit tests. It'd be good to check that peft parameter actually works, like Kirill said. It might be enough to check if a multiclass cls ViT model can be created and forwarded with peft not None. Actual training seems to be a bit redundant and we have don't have a redy-to-use location to store that kind of tests

src/otx/backend/native/models/classification/multilabel_models/vit.py
gyuilLim: I got it. Then, should we keep the `lora` argument as it is?

gyuilLim:
@sovrasov Thank you! I'll add unit tests.
and here is the table that shows exprimental results using `DINOv2` on six datasets

- GPU memory usage(GB)

| Method           | FGVC_aircraft | FOOD_101 | Stanford_Cars | CUB-200 | HAM10000 | RESISC45 | Avg Memory |
|------------------|---------------|----------|---------------|---------|----------|----------|---------------|
| Full Fine-Tuning | 4.050         | 4.070    | 4.010         | 4.020   | 3.990    | 4.050    | 4.032         |
| Freeze backbone  | 0.550         | 0.560    | 0.550         | 0.560   | 0.530    | 0.540    | 0.548         |
| LoRA(q, v)       | 2.580         | 2.640    | 2.580         | 2.590   | 2.560    | 2.570    | 2.587         |
| DoRA(q, v)       | 2.680         | 2.700    | 2.680         | 2.730   | 2.670    | 2.670    | 2.688         |

- Test Accuracy(%)

| Method           | FGVC_aircraft | FOOD_101 | Stanford_Cars | CUB-200 | HAM10000 | RESISC45 | Avg Top-1 Acc |
|------------------|---------------|----------|---------------|---------|----------|----------|---------------|
| Full Fine-Tuning | 47.11         | 83.87    | 75.0          | 75.41   | 78.16    | 92.14    | 75.28         |
| Freeze backbone  | 38.52         | 87.12    | 66.24         | 81.98   | 59.76    | 80.87    | 69.08         |
| LoRA(q, v)       | 49.12         | 88.53    | 76.51         | 80.36   | 78.62    | 93.05    | 77.70         |
| DoRA(q, v)       | 50.04         | 88.87    | 75.70         | 81.05   | 79.02    | 92.76    | 77.90         |

src/otx/backend/native/models/classification/multilabel_models/vit.py
sovrasov: I think we can just delete it for now, since the next version of OTX is going to contain breaking changes anyway

tests/unit/backend/native/models/classification/backbones/test_vision_transformer.py
```diff
@@ -61,3 +61,16 @@ def test_forward(self, config):
         assert len(outs) == 1
         feat = outs[-1]
         assert feat.shape == (1, 197, model.embed_dim)
+
+    def test_peft(self, config):
+        # test peft methods
```
sovrasov: Does this comment actually help to understand what's going on?

tests/unit/backend/native/models/classification/backbones/test_vision_transformer.py
```diff
@@ -61,3 +61,16 @@ def test_forward(self, config):
         assert len(outs) == 1
         feat = outs[-1]
         assert feat.shape == (1, 197, model.embed_dim)
+
+    def test_peft(self, config):
```
sovrasov, line 65: It's also beneficial to have separate unit tests for `AttentionWith*` in a standalone file as well, since this test is tied to a particular backbone

sovrasov (approved):

kprokofi (approved):

PR #4470, opened 2025-07-18
🐞 Minor Bugfixes for benchmarking

ashwinvaidya17:
- Max epochs in train overrides the max_epochs value loaded from config when creating the engine
- Other fixes for benchmarking script

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/backend/native/engine.py
+++ b/src/otx/backend/native/engine.py
@@ -147,11 +147,11 @@ class OTXEngine(Engine):
 
     def train(
         self,
-        max_epochs: int = 200,
+        max_epochs: int | None = None,
         min_epochs: int = 1,
         seed: int | None = None,
         deterministic: bool | Literal["warn"] = False,
-        precision: _PRECISION_INPUT | None = "16",
+        precision: _PRECISION_INPUT | None = None,
         val_check_interval: int | float | None = None,
         callbacks: list[Callback] | Callback | None = None,
         logger: Logger | Iterable[Logger] | bool | None = None,
--- a/src/otx/data/dataset/anomaly.py
+++ b/src/otx/data/dataset/anomaly.py
@@ -80,7 +80,7 @@ class OTXAnomalyDataset(OTXDataset):
                 ori_shape=img_shape,
                 image_color_channel=self.image_color_channel,
             ),
-            label=torch.tensor(label, dtype=torch.long),
+            label=label.to(dtype=torch.long),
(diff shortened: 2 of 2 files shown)


eugene123tw (approved):
Looks good to me! Thanks for founding the bugs and making the fixes :)

src/otx/backend/native/engine.py
```diff
@@ -147,11 +147,11 @@ def __init__(
 
     def train(
         self,
-        max_epochs: int = 200,
+        max_epochs: int | None = None,
```
kprokofi, line 150: What then the default value? Why does it override the config? I think, we need to keep default value and fix this problem in other way

Previously it was possible to run engine.train() w/o any parameters

kprokofi (changes requested):
Thanks Ashwin for this PR, could take a look at my comment?
  Also, all relevant bug fixes should be applied to 2.5 first

src/otx/backend/native/engine.py
ashwinvaidya17: Yeah this it tricky. Currently this
```python
engine = OTXEngine(model="padim.yaml", data="/data/mvtec/bottle")
engine.train()
```
Runs for 200 epochs even though Padim has 1 epoch set in the config.
This is because the `max_epochs` is 200 and overrides the trainer args set in `__init__`. I noticed this when running benchmarking. In the old version https://github.com/open-edge-platform/training_extensions/blob/49bf19f63eab86a0d600e7b906b92de092dfd7f2/src/otx/engine/engine.py#L158 the default value of max_epochs was 10. I guess benchmarking ran for 10 epochs for Padim but probably went unnoticed.
Maybe a solution to this might be to get the max_epochs from the trainer args and re-patch it in engine.train in the benchmark script. Or introduce something like 
https://github.com/open-edge-platform/anomalib/blob/5966423da7fe9896500d0595f41c86f3f7bb1f41/src/anomalib/models/components/base/anomalib_module.py#L248
which overrides all the arguments passed to the trainer as it is specific to the model.

kprokofi: I changed the behavior of the OTXEngine.from_config 
https://github.com/open-edge-platform/training_extensions/pull/4486

kprokofi:
We don't need this PR, since this problem was fixed in other [PR](https://github.com/open-edge-platform/training_extensions/pull/4496)

PR #4473, opened 2025-07-21
Improve EarlyStoppingWithWarmup docs and set check_on_train_epoch_end to False as default

eugene123tw:
## Summary

This PR improves `EarlyStoppingWithWarmup` callback documentation and sets `check_on_train_epoch_end` to `False` by default to avoid double counting on both train and validation epoch ends, preventing unexpectedly early training termination.

-  Added unit test for `EarlyStoppingWithWarmup` callback

### How to test

- Tests cover both unit functionality and integration scenarios
### Checklist

### License

```python
```

### DIFF
--- a/src/otx/backend/native/callbacks/adaptive_early_stopping.py
+++ b/src/otx/backend/native/callbacks/adaptive_early_stopping.py
@@ -27,7 +27,7 @@ class EarlyStoppingWithWarmup(EarlyStopping):
         check_finite: bool = True,
         stopping_threshold: float | None = None,
         divergence_threshold: float | None = None,
-        check_on_train_epoch_end: bool | None = None,
+        check_on_train_epoch_end: bool | None = False,
         log_rank_zero_only: bool = False,
         warmup_iters: int = 100,
         warmup_epochs: int = 3,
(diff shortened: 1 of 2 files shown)


src/otx/backend/native/callbacks/adaptive_early_stopping.py
```diff
@@ -27,7 +27,7 @@ def __init__(
         check_finite: bool = True,
         stopping_threshold: float | None = None,
         divergence_threshold: float | None = None,
-        check_on_train_epoch_end: bool | None = None,
+        check_on_train_epoch_end: bool | None = False,
```
kprokofi, line 30: Are there any use cases for this parameter? Maybe just delete it?

rajeshgangireddy (approved):

src/otx/backend/native/callbacks/adaptive_early_stopping.py
eugene123tw: Even if we remove this parameter, we still need to explicitly set `check_on_train_epoch_end=False` in the `super().__init__()` call. Otherwise, the parent class's `setup()` method will automatically determine this setting based on the validation schedule:

```python
@override
def setup(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", stage: str) -> None:
    if self._check_on_train_epoch_end is None:
        # if the user runs validation multiple times per training epoch or multiple training epochs without
        # validation, then we run after validation instead of on train epoch end
        self._check_on_train_epoch_end = trainer.val_check_interval == 1.0 and trainer.check_val_every_n_epoch == 1
```

This auto-detection logic would enable train epoch end checks when `check_val_every_n_epoch == 1`, causing double counting issues that we're trying to avoid.

kprokofi (approved):

PR #4475, opened 2025-07-21
Merge 2.4.6 to 2.5.0 (Fix checkpoint loading update)

### LINKED ISSUE #4460: Implement Model Checkpoint Compatibility Testing
## Problem Statement
OTX releases may introduce breaking changes that affect model checkpoint compatibility. We need a system to save lightweight model checkpoints for every release and test their compatibility with newer versions, ensuring backward compatibility and providing early warning of breaking changes.

### Functional Requirements
1. Checkpoint Generation: Create lightweight checkpoints for all models in each release
2. Weight Reduction: Remove or zero-out weights to minimize storage while preserving model structure
3. Compatibility Testing: Load previous release checkpoints in current OTX version
4. Version Matrix: Test compatibility across multiple version combinations
5. Automated Collection: Integrate with release pipeline to automatically generate checkpoints

### Non-Functional Requirements
1. Reduce checkpoint size 
2. Detect breaking changes 
3. Minimal manual intervention required

kprokofi:
### Summary

- [x]  Merge changes from 2.4.6
- [x] Update checkpoint loading fix

### Checklist

### License

```python
```


src/otx/backend/native/models/instance_segmentation/base.py
```diff
@@ -251,6 +251,7 @@ def _export_parameters(self) -> TaskLevelExportParameters:
         # Instance segmentation needs to add empty label to satisfy MAPI wrapper requirements
         modified_label_info.label_names.insert(0, "otx_empty_lbl")
         modified_label_info.label_ids.insert(0, "None")
+        modified_label_info.label_groups[0].insert(0, "otx_empty_lbl")
```
eugene123tw, line 256: Could you explain why we need to insert an empty label into the label groups? Is it possible for the label group to not include an empty label? If there is a specific reason for this requirement, could you please add a comment in the code to clarify it?

src/otx/backend/native/utils/utils.py
```diff
@@ -94,3 +94,34 @@ def ensure_callable(func: Callable[[_T], _V]) -> Callable[[_T], _V]:
     if not callable(func):
         raise TypeError(func)
     return func
+
+
+def mock_modules_for_chkpt() -> None:
```
eugene123tw: Would it be possible to refactor `mock_modules_for_chkpt()` into a context manager? I'd suggest naming it something like `legacy_otx_compatibility_context()` for clarity.
The current approach patches `sys.modules` globally and these changes stay in memory, so other parts of OTX might unexpectedly use the patched modules.

I think a context manager would yield the patched modules only during loading and automatically revert sys.modules to its original state, removing those risks.

This would change the usage from:
```python
mock_modules_for_chkpt()
ckpt = torch.load(checkpoint, map_location=map_location, weights_only=False)
```
to:
```python
with legacy_otx_compatibility_context():
    ckpt = torch.load(checkpoint, map_location=map_location, weights_only=False)
```

```python
@contextmanager
def legacy_otx_compatibility_context():
  try:
    # storw and patch existing module attributes
    # ...
    yield
  finally:
   # restore original state

**Comment by eugene123tw** [comment]
Thanks @kprokofi! Great work on handling backward checkpoint compatibility! :+1:   To make this even more robust, could you consider refactoring `mock_modules_for_chkpt()` to a context manager like `legacy_otx_compatibility_context()`? Happy to help if you need any help on the implementation.

File: src/otx/backend/native/models/instance_segmentation/base.py (hunk 1 above)
Comment (kprokofi) [reply src/otx/backend/native/models/instance_segmentation/base.py:256]: because after loading checkpoint, length of the label groups is different comparing to label names. I saw an error loading snapshot. I don't know exactly why we need otx_empty_label, but it was already there, I changed the label groups only.

File: src/otx/backend/native/utils/utils.py (hunk 2 above)
Comment (kprokofi) [reply src/otx/backend/native/utils/utils.py]: Done

**Review by eugene123tw — APPROVED** [review]
Thanks Kirill!

File: CHANGELOG.md [hunk 3]
```diff
@@ -2,7 +2,49 @@
...
+
+- Fix torchmetrics to 1.6.0
+
+## \[2.4.1\]
+
+### Bug fixes
+
+- Update Datumaro from 1.10.0rc0 to 1.10.0
+
+## \[2.4.0\]
```
sovrasov, line 96: Is it actually changelog from 2.5?

kprokofi: Updated

sovrasov (approved):

PR #4486, opened 2025-07-25
Fix XPU training and optimization from Geti2.5

kprokofi:
### Summary

- [x]  Apply fixes to run training on XPU
- [x]  Add labels_id to be able to load old OV models
- [x]  Fix auto batch size for XPU devices and make "safe" by default value when using Geti

### Checklist

### License

```python
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,6 +71,7 @@ cuda = ["torch==2.7.0"]
 xpu = [
     "torch==2.7.0+xpu",
     "pytorch-triton-xpu==3.3.0",
+    "torchvision==0.22.0+xpu"
 ]
 
 docs = [
--- a/src/otx/backend/native/engine.py
+++ b/src/otx/backend/native/engine.py
@@ -853,18 +853,19 @@ class OTXEngine(Engine):
             dict[str, Any]: Training arguments for OTXEngine.train method.
 
         Example:
-            >>> engine = OTXEngine.from_model_name(
+            >>> engine, train_kwargs = OTXEngine.from_model_name(
             ...     model_name="atss_mobilenetv2",
             ...     task="DETECTION",
             ...     data_root=<dataset/path>,
             ... )
+            ... engine.train(**train_kwargs)
 
             If you want to override configuration from default config:
                 >>> overriding = {
                 ...     "data.train_subset.batch_size": 2,
                 ...     "data.test_subset.subset_name": "TESTING",
                 ... }
-                >>> engine = OTXEngine(
+                >>> engine, train_kwargs = OTXEngine(
                 ...     model_name="atss_mobilenetv2",
                 ...     task="DETECTION",
                 ...     data_root=<dataset/path>,
--- a/src/otx/backend/native/engine.py
+++ b/src/otx/backend/native/engine.py
@@ -185,6 +185,7 @@ class OTXEngine(Engine):
             adaptive_bs (Literal["None", "Safe", "Full"]):
                 Change the actual batch size depending on the current GPU status.
                 Safe => Prevent GPU out of memory. Full => Find a batch size using most of GPU memory.
+                Defaults to "None".
             check_val_every_n_epoch (int | None, optional): How often to check validation. Defaults to 1.
             num_sanity_val_steps (int | None, optional): Number of validation steps to run before training starts.
             **kwargs: Additional keyword arguments for pl.Trainer configuration.
--- a/src/otx/backend/native/engine.py
+++ b/src/otx/backend/native/engine.py
@@ -755,7 +755,7 @@ class OTXEngine(Engine):
         data_root: PathLike | None = None,
         work_dir: PathLike | None = None,
         **kwargs,
-    ) -> tuple(Engine, dict[str, Any]):
+    ) -> tuple[Engine, dict[str, Any]]:
         """Builds the engine from a configuration file.
 
         Args:
@@ -836,7 +836,7 @@ class OTXEngine(Engine):
         data_root: PathLike | None = None,
         work_dir: PathLike | None = None,
         **kwargs,
-    ) -> tuple(Engine, dict[str, Any]):
+    ) -> tuple[Engine, dict[str, Any]]:
         """Builds the engine from a model name.
 
         Args:
@@ -963,7 +963,7 @@ class OTXEngine(Engine):
             self.work_dir = self._trainer.default_root_dir
 
     def configure_accelerator(self) -> None:
-        """Updates the cache arguments based on the device type.""" 
+        """Updates the cache arguments based on the device type."""
         if self._device.accelerator == DeviceType.xpu:
             self._cache.update(strategy="xpu_single")
             # add plugin for Automatic Mixed Precision on XPU
--- a/src/otx/backend/native/engine.py
+++ b/src/otx/backend/native/engine.py
@@ -149,7 +149,7 @@ class OTXEngine(Engine):
 
     def train(
         self,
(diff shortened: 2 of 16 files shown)


src/otx/backend/native/engine.py
```diff
@@ -825,7 +826,7 @@ def from_config(
             data=datamodule,
             model=model,
             **engine_kwargs,
-        )
+        ), train_kwargs
```
eugene123tw: I'm not sure it's better to return kwargs from `from_engine`. Could we keep it cleaner by simply passing `max_epochs` directly to `OTXEngine.from_config(...)` to override it, instead of returning and modifying the returned kwargs and then passing them to `engine.train`? I think the workflow would be smoother if we override parameters within `from_config` itself, rather than handling it externally.

eugene123tw (commented):
Thanks Kirill! I left a comment.

src/otx/backend/native/engine.py
kprokofi: This is the problem. In `from_config` we add `max_epoch` to `enigine_kwargs` and store them in the cache.
But in train method, we override it by passing number of epochs from train method. Argument in train method has priority. To avoid this, I propose to return `train_kwargs`. I didn't find a solution how we can override it directly in `OTXEngine.from_config(...)`.

We return `train_kwargs` similarly in ConfigConverter.

eugene123tw: @kprokofi Is it possible to create an empty `self._train_overrides` for storing  training parameter overrides from config?

And in `from_config` we can set `engine._train_overrides = train_kwargs` before returning `engine`. 
In `train` we can find signatures and replace them:

```py
    def _apply_param_overrides(self, current_locals: dict[str, Any]) -> dict[str, Any]:
        result = {}
        sig = inspect.signature(self.train)
        
        for param_name, param in sig.parameters.items():
            if param_name not in {'self', 'kwargs'} and param_name in current_locals:
                current_value = current_locals[param_name]
                
                # Apply override if current value matches default and we have an override
                # NOTE: basically to check the default value and set value, if its not the same, use the set value
                if (param.default != inspect.Parameter.empty and 
                    current_value == param.default and
                    param_name in self._train_overrides):
                    result[param_name] = self._train_overrides[param_name]
                else:
                    result[param_name] = current_value        
        return result
```

in `train()`
```py
params = self._apply_param_overrides(locals())
max_epochs = params['max_epochs']
min_epochs = params['min_epochs']  
seed = params['seed']
deterministic = params['deterministic']
precision = params['precision']
val_check_interval = params['val_check_interval']
callbacks = params['callbacks']
logger = params['logger']
resume = params['resume']
metric = params['metric']
checkpoint = params['checkpoint']
adaptive_bs = params['adaptive_bs']
check_val_every_n_epoch = params['check_val_every_n_epoch']
num_sanity_val_steps = params['num_sanity_val_steps']
```

kprokofi: Do we really need so much complexity for the train function?
And also,
"current_value == param.default" - user may want to  specifically use default parameter, but it will be overridden with self._train_overrides.

eugene123tw: It won't be overridden, `self._train_overrides` is empty if the user does not set configurable parameters with `from_config`. And it will use param.default.

eugene123tw: I find it a bit awkward that `from_config` returns a `train_kwargs` dictionary, which then requires the user to manually override values before passing them to `engine.train`.

For example:
Option 1:
```py
engine, train_kwargs = OTXEngine.from_config(...)
train_kwargs["max_epochs"] = epoch
engine.train(**train_kwargs)
```
vs.

Option 2:
```py
engine = OTXEngine.from_config(..., max_epochs=epoch)
engine.train()
```
I think Option 2 looks cleaner and more intuitive, as it avoids the extra indirection and keeps the override logic within `from_config`.

eugene123tw (dismissed):

pyproject.toml
```diff
@@ -71,6 +71,7 @@ cuda = ["torch==2.7.0"]
 xpu = [
     "torch==2.7.0+xpu",
     "pytorch-triton-xpu==3.3.0",
+    "torchvision==0.22.0+xpu"
```
leoll2, line 74: In Geti `xpu/Dockerfile` we have custom instructions to install torch with XPU support (see below). Now that OTX provides the `[xpu]` extra, I think we can simplify it by running `uv sync` with an `xpu` extra that pulls `otx[xpu]`.

CC @maxxgx

https://github.com/open-edge-platform/geti/blob/1a57d6f8926b19231a6c883fd2e748a27b24d6da/interactive_ai/workflows/train/trainer/xpu/Dockerfile#L55

src/otx/backend/native/models/modules/conv_module.py
```diff
@@ -392,7 +394,7 @@ class Conv2dModule(ConvModule):
     # Use the patched Conv2d if XPU is available.
     # This is to avoid issues with XPU performance.
     # TODO(kprokofi): Remove this when XPU performance is fixed.
```
rajeshgangireddy, line 396: I think this comment can be removed now ?

src/otx/tools/converter.py
```diff
@@ -511,6 +511,8 @@ def instantiate(
         instantiated_kwargs = engine_parser.instantiate_classes(Namespace(**config))
 
         train_kwargs = {k: v for k, v in instantiated_kwargs.items() if k in train_arguments}
+        # enable auto batch size for training
+        train_kwargs["adaptive_bs"] = "Safe"
```
rajeshgangireddy, line 515: Don't we have to perform any check whether `train_kwargs["adaptive_bs"]` is already populated with another type of `adaptive_bs` method?

src/otx/backend/native/models/detection/heads/atss_head.py
```diff
@@ -25,7 +25,7 @@
 )
 from otx.backend.native.models.detection.utils.prior_generators.utils import anchor_inside_flags
 from otx.backend.native.models.detection.utils.utils import unmap
-from otx.backend.native.models.modules.conv_module import Conv2dModule
+from otx.backend.native.models.modules.conv_module import Conv2dModule, PatchedConv2d
```
rajeshgangireddy: I see that in [models/modules/__init__.py](https://github.com/open-edge-platform/training_extensions/blob/develop/src/otx/backend/native/models/modules/__init__.py) we already have a list of modules. Can we move the PatchedConv2D to that list and import from there?

src/otx/tools/converter.py
kprokofi: It is made for Converter only (Geti use case). In Geti we agreed to use Safe mode by default and do not expose it to the users

pyproject.toml
kprokofi: Yes, now we can install otx like: uv pip install otx[xpu]

src/otx/backend/native/models/modules/conv_module.py
kprokofi: No, we still need to use PatchedConv2d. When IPEX fix this problem with upstream version of the PyTorch, we can remove this TODO

src/otx/backend/native/models/detection/heads/atss_head.py
kprokofi: Done https://github.com/open-edge-platform/training_extensions/pull/4486/commits/fa5c0e47089bb78f30de4601e860c2b5537f84ed

eugene123tw (approved):
Thanks Kirill!

PR #4498, opened 2025-08-01
Fix confidence threshold cache invalidation and filtering logic

eugene123tw:
### Summary

Fixing this issue: https://github.com/open-edge-platform/geti/issues/940

🐛 Problem
`best_confidence_threshold` property in detection and instance segmentation models had two major issues:
* Cache invalidation bug: The property cached `_best_confidence_threshold` on first access, but when `hparams["best_confidence_threshold"]` was updated during validation, the cached value (`_best_confidence_threshold`) was never invalidated, causing stale thresholds to be used.

* Incorrect filtering location: Confidence threshold filtering was applied in validation_step() and test_step(), which meant:
  * Validation metrics were computed on pre-filtered data 
  * FMeasure couldn't properly compute optimal thresholds from the full prediction distribution 
  * This created a chicken-and-egg problem where the metric needed unfiltered data to find the best threshold

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/backend/native/models/base.py
+++ b/src/otx/backend/native/models/base.py
@@ -221,7 +221,7 @@ class OTXModel(LightningModule):
         if isinstance(preds, OTXBatchLossEntity):
             raise TypeError(preds)
 
-        metric_inputs = self._convert_pred_entity_to_compute_metric(preds, batch, "val")
+        metric_inputs = self._convert_pred_entity_to_compute_metric(preds, batch)
 
         if isinstance(metric_inputs, dict):
             self.metric.update(**metric_inputs)
@@ -246,7 +246,7 @@ class OTXModel(LightningModule):
         if isinstance(preds, OTXBatchLossEntity):
             raise TypeError(preds)
 
-        metric_inputs = self._convert_pred_entity_to_compute_metric(preds, batch, "test")
+        metric_inputs = self._convert_pred_entity_to_compute_metric(preds, batch)
 
         if isinstance(metric_inputs, dict):
             self.metric.update(**metric_inputs)
@@ -372,7 +372,6 @@ class OTXModel(LightningModule):
         self,
         preds: OTXPredBatch,
         inputs: OTXDataBatch,
-        stage: Literal["val", "test"],
     ) -> MetricInput:
         """Convert given inputs to a Python dictionary for the metric computation."""
         raise NotImplementedError
--- a/src/otx/backend/native/models/base.py
+++ b/src/otx/backend/native/models/base.py
@@ -410,8 +410,6 @@ class OTXModel(LightningModule):
             checkpoint["state_dict"] = remove_state_dict_prefix(compiled_state_dict, "_orig_mod.")
         super().on_save_checkpoint(checkpoint)
 
-        if "best_confidence_threshold_list" in checkpoint["hyper_parameters"]:
-            checkpoint["hyper_parameters"].pop("best_confidence_threshold_list")
         checkpoint["hyper_parameters"]["label_info"] = asdict(self.label_info)
         checkpoint["otx_version"] = __version__
         checkpoint["hyper_parameters"]["tile_config"] = asdict(self.tile_config)
--- a/src/otx/backend/native/models/base.py
+++ b/src/otx/backend/native/models/base.py
@@ -409,6 +409,9 @@ class OTXModel(LightningModule):
             compiled_state_dict = checkpoint["state_dict"]
             checkpoint["state_dict"] = remove_state_dict_prefix(compiled_state_dict, "_orig_mod.")
         super().on_save_checkpoint(checkpoint)
+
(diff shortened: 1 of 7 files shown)


src/otx/backend/native/models/instance_segmentation/base.py
```diff
@@ -301,9 +377,37 @@ def _log_metrics(self, meter: Metric, key: Literal["val", "test"], **compute_kwa
                 {"best_confidence_threshold": best_confidence_threshold} if best_confidence_threshold else {}
             )
 
-            return super()._log_metrics(meter, key, **compute_kwargs)
-
-        raise ValueError(key)
+            super()._log_metrics(meter, key, **compute_kwargs)
+
+    def _filter_outputs_by_threshold(self, outputs: OTXPredBatch) -> OTXPredBatch:
```
kprokofi, line 373: Didn't we filter outputs by threshold before?

src/otx/backend/native/models/instance_segmentation/base.py
```diff
@@ -264,35 +271,104 @@ def _export_parameters(self) -> TaskLevelExportParameters:
...
+            elif isinstance(meter, FMeasure):
+                fmeasure = meter
+
+            if fmeasure is not None and hasattr(fmeasure, "best_confidence_threshold"):
+                # Instead of using only the last best_confidence_threshold,
+                # we now average the last N values to get a more stable estimate,
+                # especially helpful for small datasets where it can vary greatly across epochs.
+                if "best_confidence_threshold_list" not in self.hparams:
+                    self.hparams["best_confidence_threshold_list"] = []
+                self.hparams["best_confidence_threshold_list"].append(fmeasure.best_confidence_threshold)
```
kprokofi: Could you explain why do we need `best_confidence_threshold_list`? Do you have some experiments to prove necessity of this average threshold? Also, could we avoid storing this list in hparams and save there only final one?

eugene123tw: This is explained in the comments.

Instead of relying on just the last best_confidence_threshold, we now average the last N values to obtain a more stable estimate. This is particularly useful for small datasets, where the threshold can fluctuate significantly between epochs.

The issue originates from: `https://github.com/open-edge-platform/geti/issues/940`.

src/otx/backend/native/models/instance_segmentation/base.py
eugene123tw: we do, but the ordering of metric.compute and filtering is wrong.

src/otx/backend/native/models/base.py
```diff
@@ -409,6 +409,9 @@ def on_save_checkpoint(self, checkpoint: dict[str, Any]) -> None:
             compiled_state_dict = checkpoint["state_dict"]
             checkpoint["state_dict"] = remove_state_dict_prefix(compiled_state_dict, "_orig_mod.")
         super().on_save_checkpoint(checkpoint)
+
+        if "best_confidence_threshold_list" in checkpoint["hyper_parameters"]:
+            checkpoint["hyper_parameters"].pop("best_confidence_threshold_list")
```
kprokofi: checkpoint["hyper_parameters"].pop("best_confidence_threshold_list", None) -> no need to check key

kprokofi (commented):
Could you please update the Changelog? Seems like a new feature - heuristic. Worth to mention

src/otx/backend/native/models/instance_segmentation/base.py
kprokofi: I see, how much does it increase the f1-score estimation?

eugene123tw:
@kprokofi 
I’ve decided to keep this PR simple and not implement averaging over `best_confidence_threshold_list` for now.

### Reasoning:
I experimented with the mean-based approach, but test scores still frequently drop to 0%. The core issue is high variance in confidence thresholds across validation epochs.

Here’s an example from 10 evaluation cycles: `[0.175, 0.1, 0.075, 0.075, 0.275, 0.2, 0.175, 0.675, 0.125, 0.45]`
* Mean: 0.234
* Std: 0.193
* CV (Coefficient of Variation): 82% → Extremely high variance

This level of variance makes the mean unreliable, and outliers like 0.675 significantly skew the result. As a result, the averaged threshold often doesn’t correspond to a good-performing value.

### Proposal:
Let’s address robust averaging separately in a future PR.
Proper handling would require:
* Coefficient of Variation analysis
* Median-based thresholdding for moderate variance
* Fallback/default thresholds for high variance
* Minimum evaluation size checks

However, these additions introduce a lot of complexity, which is out of scope for a bug-fix PR.

### So for this PR I will include:
* Fix the caching bug in best_confidence_threshold
* Ensure threshold updates only on best F1 scores
*  Skip threshold list averaging logic

### Next Steps:
Let’s revisit the robust thresholdding logic in a dedicated feature PR, where we can design, test, and validate the statistical approach more thoroughly.

kprokofi (approved):
Make sense, thank you for investigation!

PR #4508, opened 2025-08-06
Fix gaussian noise augmentation and add random gaussian blur

kprokofi:
### Summary

Problem:

1. The Gaussian Noise augmentation from TorchVision can only be applied to images scaled to the [0, 1] range. This is not feasible for some of our models. Additionally, applying augmentations after normalization is generally discouraged.
2. Both Gaussian Noise and Gaussian Blur are always applied without any probability control, making it impossible to adjust the frequency of these augmentations.

### Checklist

### License

```python
```

### DIFF
--- a/src/otx/data/transform_libs/torchvision.py
+++ b/src/otx/data/transform_libs/torchvision.py
@@ -26,6 +26,7 @@ from omegaconf import DictConfig
 from scipy.stats import truncnorm
 from torchvision import tv_tensors
 from torchvision._utils import sequence_to_str
+from torchvision.transforms.v2 import GaussianNoise
 from torchvision.transforms.v2 import functional as F  # noqa: N812
 
 from otx.data.entity.base import (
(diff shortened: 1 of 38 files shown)


kprokofi:
image

image
image

sovrasov (commented):
Is that correct the fix doesn't change the default OTX trainings?

kprokofi:
> Is that correct the fix doesn't change the default OTX trainings?

Yes, this fix needed for Geti training. In Geti we will have configurable augmentations for classification task

src/otx/data/transform_libs/torchvision.py
```diff
@@ -903,6 +904,58 @@ def __repr__(self) -> str:
...
+    This augmentation allows to add gaussian noise to unscaled image.
+    Only float32 images are supported for this augmentation.
+    """
+
+    def __init__(self, mean: float = 0.0, sigma: float = 0.1, clip: bool = True, prob: float = 0.5) -> None:
+        super().__init__(mean=mean, sigma=sigma, clip=clip)
+        self.prob = prob
+
+    def _is_scaled(self, tensor: torch.Tensor) -> bool:
+        return torch.all((tensor >= 0) & (tensor <= 1))
```
eugene123tw: I'd recommend using `torch.min(tensor) >= 0 and torch.max(tensor) <= 1` instead. It's generally more efficient, especially for large tensors, since it avoids per-element comparisons and doesn't require creating a full boolean mask where `torch.all` does.

If you're dealing with float rounding issues, adding a small tolerance might help:
```
eps = 1e-6
torch.min(tensor) >= -eps and torch.max(tensor) <= 1 + eps
```
Original approach is totally valid, this is just an optimization suggestion.

eugene123tw (commented):
Thanks Kirill! Left a comment for optimization!

src/otx/data/transform_libs/torchvision.py
kprokofi: Done. Basically we just need to check the max value

sovrasov (approved):

eugene123tw (approved):

kprokofi:
I merged pending PRs to release branch and ran our integration tests. 
All tests are passed:

Anomaly:
image
inst_seg:
image
detection:
image
multi_class_cls:
image
multi_labe:
image
h-label:
image
key point:
image

PR #4571, opened 2025-08-19
Geti tune: fix precommit and CI workflows

leoll2:
Changes:
- adjust pre-commit configuration and resolve small issues
- update CI runners for the migrated workflows


leoll2 (commented):
LGTM

itallix (approved):

ui/package.json
```diff
@@ -45,6 +45,7 @@
         "recharts": "^2.15.3",
         "sonner": "^2.0.6",
         "static-path": "^0.0.4",
+        "usehooks-ts": "^3.1.1",
```
jpggvilaca, line 48: where is this change coming from?

itallix: Added to fix https://github.com/open-edge-platform/training_extensions/actions/runs/17067218189/job/48387137480

kprokofi (approved):

ui/package.json
jpggvilaca: The update should come from the @geti/ui package but we can also changes that later. This is ok for now.

jpggvilaca (approved):

PR #4577, opened 2025-08-20
Configurable augmentations for OD, IS, KP, SemSeg tasks

kprokofi:
### Summary
- [x] Add configurable augmentations for OD, IS, Sem_Seg, KP_DET. Please, note, not all of them are available for Tiling.
- [x] Align PhotoMetricDistortion (use from torchvision for all tasks). However, due to prevent accuracy degradation, our current implementation will be used for Semantic Seg task, later we can change it with more experiments.
- [x] Enable "is_numpy_to_tvtensor" as True by default: Since we now can enable/disable almost any augmentation, we need to be sure that all combinations will be compatible 
- [x] Add unit and integration tests with augmentations for all tasks

### Checklist

### License

```python
```


kprokofi:
image
Unit tests

kprokofi:
Anomaly:
image

Detection:
image

Instance seg:
image

Keypoint:
image

Multi-class-cls:
image

Multi-label-cls:
image

Semantic_segmentations:
image

src/otx/data/transform_libs/torchvision.py
```diff
@@ -815,7 +816,7 @@ def __init__(
         self,
         prob: float | Iterable[float] | None = None,
         direction: str | Sequence[str | None] = "horizontal",
-        is_numpy_to_tvtensor: bool = False,
+        is_numpy_to_tvtensor: bool = True,
```
eugene123tw: Will this change affect speed or accuracy? For example, in `src/otx/recipe/detection/dfine_x.yaml`, `is_numpy_to_tvtensor` was previously set to `False` as default in `RandomFlip`, but it’s now True. Will this have any impact?

eugene123tw (commented):
Thanks, Kirill, for the PR. I have a few questions regarding is_numpy_to_tvtensor and its impact.

If all transforms now have is_numpy_to_tvtensor = true, it seems like we might not need this option anymore. That said, I assume it was introduced for a reason. Do you have any data comparing accuracy and training time before and after this change? It would be great to attach those results to the PR for future reference.

src/otx/recipe/detection/yolox_x.yaml
```diff
@@ -64,6 +64,13 @@ overrides:
...
       batch_size: 4
       transforms:
+        - class_path: otx.data.transform_libs.torchvision.MinIoURandomCrop
+          enable: false
+        - class_path: otx.data.transform_libs.torchvision.Resize
+          init_args:
+            scale: $(input_size)
+            keep_ratio: true
+            transform_bbox: true
         - class_path: otx.data.transform_libs.torchvision.CachedMosaic
```
eugene123tw: I also have a related question about YOLOX. Previously, the transforms were defined as:

```
- class_path: otx.data.transform_libs.torchvision.CachedMosaic
  init_args:
    is_numpy_to_tvtensor: false
- class_path: otx.data.transform_libs.torchvision.RandomAffine
  init_args:
    is_numpy_to_tvtensor: false
- class_path: otx.data.transform_libs.torchvision.CachedMixUp
  init_args:
    is_numpy_to_tvtensor: false
- class_path: otx.data.transform_libs.torchvision.Resize
  init_args:
    is_numpy_to_tvtensor: false
- class_path: otx.data.transform_libs.torchvision.YOLOXHSVRandomAug
  init_args:
    is_numpy_to_tvtensor: false
- class_path: otx.data.transform_libs.torchvision.RandomFlip
  init_args:
    is_numpy_to_tvtensor: false
- class_path: otx.data.transform_libs.torchvision.Pad
  init_args:
    is_numpy_to_tvtensor: true
```

Now they are:
```
- class_path: otx.data.transform_libs.torchvision.CachedMosaic
  init_args:
    is_numpy_to_tvtensor: true
- class_path: otx.data.transform_libs.torchvision.RandomAffine
  init_args:
    is_numpy_to_tvtensor: true
- class_path: otx.data.transform_libs.torchvision.CachedMixUp
  init_args:
    is_numpy_to_tvtensor: true
- class_path: otx.data.transform_libs.torchvision.Resize
  init_args:
    is_numpy_to_tvtensor: true
- class_path: otx.data.transform_libs.torchvision.YOLOXHSVRandomAug
  init_args:
    is_numpy_to_tvtensor: true
- class_path: otx.data.transform_libs.torchvision.RandomFlip
  init_args:
    is_numpy_to_tvtensor: true
- class_path: otx.data.transform_libs.torchvision.Pad
  init_args:
    is_numpy_to_tvtensor: true
```

Will setting `is_numpy_to_tvtensor` to true for all transforms have any impact on performance or accuracy?

src/otx/data/transform_libs/torchvision.py
kprokofi: I think it can give some impact, maybe not so huge. But I don't see the solution right now how to support configurable augmentations on the OTX side without this change. The problem is that in Geti we can disable some augmentations, but the whole pipeline should be compatible.

src/otx/recipe/detection/yolox_x.yaml
kprokofi: I’ve already discussed this issue with Radwan, and we decided to create a ticket considering a possible migration to Albumentations, for example. This should resolve the problem. For now, however, we should continue working with the current setup. Yes, probably it will give us a bit slowness because of the numpy -> torch conversion

src/otx/data/transform_libs/torchvision.py
eugene123tw: Could you elaborate on why the original `is_numpy_to_tvtensor` wouldn't work when certain augmentations are enabled while others are disabled?

kprokofi: Yes, in Geti we can switch off augmentation before torchvision and the pipeline will fail since torchvision requires always torch tensor as input, while our internal augmentations require numpy arrays. This inconsistency leads to problems. To avoid this I suggest always convert numpy arrays to torch tensors for safety.

eugene123tw (dismissed):

Issue #4569, opened 2025-08-19
Migrate services from Singleton pattern to FastAPI DI


leoll2:
Refactor services to use FastAPI’s dependency injection (DI) mechanism instead of the current singleton/metaclass approach.

**Problems with the current approach:**
- Constructor Semantics: Any change to a service’s constructor requires updating all direct instantiations across the codebase, increasing maintenance overhead and risk of errors.
- Tight Coupling: Endpoints and other consumers are tightly coupled to concrete implementations, making the code less modular and harder to refactor.
- Testing Complexity: Testing requires import patching or monkey-patching singletons, which is error-prone and less transparent than using DI.

**Benefits of FastAPI DI:**
- Centralized Configuration: Service instantiation and configuration are managed in one place, simplifying updates and environment-specific changes. (`dependencies.py`)
- Loose Coupling: Dependent components will be injected via constructor and not instantiated directly (see examples of coupling in `ConfigurationService._notify_*` methods)
- Lifecycle Management: DI allows for flexible service lifecycles (per-request, singleton, etc.) as needed.
- Consistency: Aligns with FastAPI best practices and makes the codebase more consistent and idiomatic.

**Acceptance Criteria:**
- All services (except `Scheduler`) currently using singleton/metaclass patterns are refactored [to use DI](https://fastapi.tiangolo.com/tutorial/dependencies/).
- Service instantiation is centralized in dependency provider functions (`dependencies.py`).
- All endpoints and consumers use [FastAPI’s Depends](https://fastapi.tiangolo.com/reference/dependencies/?h=depends#fastapi.Depends) for service injection.
- Tests are updated to use DI overrides instead of import patching.

PR #4599, opened 2025-08-25
Add model latency endpoint

### LINKED ISSUE #4537: Endpoint to monitor model latency
Proposal: endpoint `/api/projects/<id>/pipeline/metrics`

Response like:
```json
{
  "time_window": {
    "start": <timestamp>,
    "end": <timestamp>,
    "duration_seconds": 60
  },
  "inference": {
    "latency": {
       "avg_ms": 15.1,
       "min_ms": 12.3,
       "max_ms": 30.4,
       "p95_ms": 25.4,
       "latest_ms": 15.6
    }
  }
}
```

A-Artemis:
### Summary

Added a MetricsService which collects and retrieves the model's inference latencies. It uses shared memory so that it is process safe when accessing the recorded measurements.
The latencies are stored in memory and only a total of 1024 entries are kept.

### How to test

```http://[geti-tune.localhost/api/pipelines/ace3f1da-fdd9-4048-a95e-a647ed969442/metrics](http://geti-tune.localhost/api/pipelines/ace3f1da-fdd9-4048-a95e-a647ed969442/metrics)```

image

### Checklist

### License

```python
```


backend/app/api/endpoints/pipelines.py
```diff
@@ -197,6 +198,36 @@ async def disable_pipeline(
...
+    responses={
+        status.HTTP_200_OK: {"description": "Pipeline metrics successfully calculated", "model": PipelineMetrics},
+        status.HTTP_400_BAD_REQUEST: {"description": "Invalid pipeline ID or duration parameter"},
+        status.HTTP_404_NOT_FOUND: {"description": "Pipeline not found"},
+    },
+)
+async def get_pipeline_metrics(
+    pipeline_id: Annotated[UUID, Depends(get_pipeline_id)],
+    pipeline_service: Annotated[PipelineService, Depends(get_pipeline_service)],
+    duration_seconds: int = 10,
```
leoll2: Rename suggestion `duration_seconds` -> `time_window`

backend/tests/integration/services/test_pipeline_service.py
```diff
@@ -319,3 +319,87 @@ def test_delete_active_pipeline(
...
+            name="Test Pipeline",
+            model_id=model_id,
+        )
+        db_session.add(pipeline_db)
+        db_session.flush()
+
+        # Mock metrics collector
+        with patch(
+            "app.services.metrics_collector.MetricsCollector.get_latency_measurements"
+        ) as mock_metrics_collector:
```
leoll2: This looks like a unit test. Integration tests shouldn't mock core parts of the components being tested.

backend/app/services/pipeline_service.py
```diff
@@ -90,3 +94,59 @@ def delete_pipeline_by_id(self, pipeline_id: UUID) -> None:
                 raise ResourceInUseError(ResourceType.PIPELINE, str(pipeline_id), MSG_ERR_DELETE_RUNNING_PIPELINE)
             self._persistence.delete_by_id(pipeline_id, db)
             db.commit()
+
+    def get_pipeline_metrics(self, pipeline_id: UUID, duration_seconds: int = 60) -> PipelineMetrics:
+        """Calculate metrics for a pipeline over a specified time window."""
+        # First check if pipeline exists
+        pipeline = self.get_pipeline_by_id(pipeline_id)
+        if not pipeline.model_id:
+            raise ValueError("Cannot get metrics for a pipeline without an associated model.")
```
leoll2: I would change this check to `if pipeline.status != PipelineStatus.RUNNING` and raise a more specific exception. Then make sure to handle that exception in the endpoint (remap to 400 bad request).

backend/app/services/pipeline_service.py
```diff
@@ -90,3 +94,59 @@ def delete_pipeline_by_id(self, pipeline_id: UUID) -> None:
...
+
+        # Calculate time window
+        end_time = datetime.now(UTC)
+        start_time = end_time - timedelta(seconds=duration_seconds)
+
+        # Get actual latency measurements from the metrics collector
+        metrics_collector = MetricsCollector()
```
leoll2: Can you move the metrics collector initialization to the constructor? Then you don't have to make the class Singleton, I guess.

backend/app/services/pipeline_service.py
```diff
@@ -90,3 +94,59 @@ def delete_pipeline_by_id(self, pipeline_id: UUID) -> None:
...
+        # Calculate latency metrics
+        if latency_samples:
+            avg_ms = statistics.mean(latency_samples)
+            min_ms = min(latency_samples)
+            max_ms = max(latency_samples)
+            p95_ms = self._calculate_percentile(latency_samples, 95)
+            latest_ms = latency_samples[-1]
+        else:
+            # No data available
+            avg_ms = min_ms = max_ms = p95_ms = latest_ms = 0.0
```
leoll2: Why 0.0 instead of None?

backend/app/workers/inference.py
```diff
@@ -71,10 +76,14 @@ def on_inference_completed(inf_result: DetectionResult, userdata: dict[str, Any]
                     queue_data = frame_queue.get(timeout=1)
                 except queue.Empty:
                     continue
+
+                inference_start_time = metrics_collector.record_inference_start()
                 model.infer_async(
                     queue_data.frame_data,
                     user_data={"stream_data": queue_data, "model_name": model_service.get_active_model_name()},
                 )
+                model_service.get_loaded_inference_model()
```
leoll2: Why this call?

backend/app/workers/inference.py
```diff
@@ -71,10 +76,14 @@ def on_inference_completed(inf_result: DetectionResult, userdata: dict[str, Any]
...
+                metrics_collector.record_inference_end(model_id=loaded_model.id, start_time=inference_start_time)
```
leoll2: This is not the right place to call `record_inference_end`: the inference is asynchronous, so at this point the inference request has been submitted but you have no guarantee that the inference itself is complete. Instead, you should compute the duration in the callback (although not very accurate due to scheduling overheads); you can pass the start time to the callback via `infer_async`.

backend/app/services/pipeline_service.py
A-Artemis: There is a good chance the model collector will be used again and with similar metrics. There is `throughput` coming which will be used in other places, keeping it as a singleton with a shared queue would make further calculations easier.

backend/app/services/metrics_collector.py
```diff
@@ -0,0 +1,79 @@
...
+    latency_ms: float
+    timestamp: datetime
+
+
+class MetricsCollector(metaclass=Singleton):
+    """Thread-safe in-memory metrics collector for model latency data"""
+
+    def __init__(self, max_age_seconds: int = 60):
+        self._max_age_seconds = max_age_seconds
+        self._measurements: deque[LatencyMeasurement] = deque()
```
itallix: The inference worker runs in a separate process, meaning a different Python interpreter instance with isolated memory, and will have its own Singleton instance of `MetricsCollector`. As a result, the measurements deque will be different from the one in the main process where the FastAPI server is started. Using an SQLite database would be the easiest way to persist the measurements. Are there requirements to keep it in memory?

A-Artemis: The metrics collector now uses shared memory with pre-allocated size https://docs.python.org/3/library/multiprocessing.shared_memory.html

backend/app/services/metrics_collector.py
```diff
@@ -0,0 +1,95 @@
...
+    SIZE = DTYPE.itemsize * MAX_MEASUREMENTS  # 160 * 1024 = 163840 bytes (160KB) allocated
+
+    def __init__(self, max_age_seconds: int = 60):
+        self._max_age_seconds = max_age_seconds
+        self._lock = Lock()
+        try:
+            self._shm = shared_memory.SharedMemory(name=self.SHM_NAME)
+            self._array: np.ndarray = np.ndarray((self.MAX_MEASUREMENTS,), dtype=self.DTYPE, buffer=self._shm.buf)
+        except FileNotFoundError:
+            self._shm = shared_memory.SharedMemory(name=self.SHM_NAME, create=True, size=self.SIZE)
```
itallix: Shared memory should be created centrally in the scheduler and passed to the process via arguments, similar to other shared resources. This way, we can cleanup it properly during application shutdown. Currently, shared memory is removed in a singleton that will be garbage collected twice, causing exception on the second attempt.

backend/app/services/metrics_collector.py
```diff
@@ -0,0 +1,95 @@
...
+
+class LatencyMeasurement(NamedTuple):
+    """Individual latency measurement"""
+
+    model_id: str  # UUID as 36 character string "00000000-0000-0000-0000-000000000000"
+    latency_ms: float
+    timestamp: float
+
+
+class MetricsCollector(metaclass=Singleton):
```
itallix: We are working to minimize the use of the Singleton pattern (see related issues [here](#4569)). Since metrics are collected within the same process, we can consider separating metric collection from retrieval. Probably, the `PipelineService` can access shared memory for metrics directly and collector don't need to be a Singleton in that case?

itallix (commented):
Great progress! I've added a few suggestions for further improvement.

backend/app/services/pipeline_service.py
```diff
@@ -90,3 +94,55 @@ def delete_pipeline_by_id(self, pipeline_id: UUID) -> None:
...
             db.commit()
+
+    def get_pipeline_metrics(self, pipeline_id: UUID, time_window: int = 60) -> PipelineMetrics:
+        """Calculate metrics for a pipeline over a specified time window."""
+        # First check if pipeline exists
+        pipeline = self.get_pipeline_by_id(pipeline_id)
+        if pipeline.status != PipelineStatus.RUNNING:
+            raise ValueError("Cannot get metrics for a pipeline that is not running.")
+        if pipeline.model_id is None:
+            raise ValueError("Cannot get metrics for a pipeline with no model, please check a model is assigned.")
```
leoll2: This check can be removed, since you already check that the pipeline is running at L102. A running pipeline always has a model.

backend/app/services/metrics_collector.py
```diff
@@ -0,0 +1,94 @@
...
+        timestamp = datetime.now(UTC).timestamp()
+
+        measurement = LatencyMeasurement(str(model_id), latency_ms, timestamp)
+        with self._lock:
+            idx = self._head % self.MAX_MEASUREMENTS
+            self._array[idx] = (measurement.model_id, measurement.latency_ms, measurement.timestamp)
+            self._head += 1
+            logger.debug(f"Latency measurement recorded for model {model_id}: {latency_ms:.2f} ms")
+
+    def get_latency_measurements(self, model_id: UUID, time_window: int = 60) -> list[float]:
```
leoll2: Please add docstrings to the public methods

backend/app/core/scheduler.py
```diff
@@ -36,6 +38,10 @@ def __init__(self) -> None:
         # Condition variable to notify processes about configuration updates
         self.mp_config_changed_condition = mp.Condition()
 
+        # Shared memory for metrics collector
+        self.shm_metrics_collector: SharedMemory = SharedMemory(name=SHM_NAME, create=True, size=SIZE)
+        print(self.shm_metrics_collector.name, SIZE)
```
itallix: `print` stmt

backend/tests/unit/services/test_metrics_collector.py
```diff
@@ -0,0 +1,186 @@
...
+
+import time
+from datetime import UTC, datetime, timedelta
+from multiprocessing.shared_memory import SharedMemory
+from unittest.mock import patch
+from uuid import uuid4
+
+from app.services.metrics_collector import MAX_MEASUREMENTS, SHM_NAME, SIZE, MetricsCollector
+
+SHM = SharedMemory(name=SHM_NAME, create=True, size=SIZE)
```
itallix: This should be cleaned up explicitly.
We can use `scope=module` fixture in autouse mode, so that cleanup can be handled after `yield`, e.g.:

```python
@pytest.fixture(scope="module", autouse=True)
def fxt_metrics_mem():
  shm = SharedMemory(name=SHM_NAME, create=True, size=SIZE)
  yield
  shm.close()
  shm.unlink()
```

backend/app/services/pipeline_service.py
itallix: > Can you move the metrics collector initialization to the constructor? Then you don't have to make the class Singleton, I guess.

The original comment is still relevant. There are two possible approaches to avoid instantiating extra objects within this method (both options help prevent monkey-patching in tests and allow for better composition of fixtures):

- **Option 1:** Create a shared memory reference in the constructor and access it directly to retrieve metrics.
- **Option 2:** Pass a `MetricsCollector` instance via dependencies (see `dependencies.py`), ensuring a single instance exists throughout the web server lifecycle.

I believe the retrieval logic should be decoupled from `MetricsCollector`, as its name suggests a focus on collection. Retrieval is only required within `PipelineService`.

leoll2 (dismissed):

itallix (approved):
LGTM!

leoll2 (approved):

PR #4606, opened 2025-08-27
Classification Dataset refactor

AlbertvanHouten:
This pull request introduces significant updates to the OTX data pipeline, focusing on integrating the new experimental Datumaro `Dataset` and related APIs for classification tasks. The changes add new base and classification dataset classes, update the dataset factory and sampler logic to support both legacy and experimental Datumaro datasets, and refactor sample entities for better compatibility. 

**Experimental Datumaro integration and new dataset classes:**

* Added `OTXDataset` and `OTXMulticlassClsDataset` classes in `base_new.py` and `classification_new.py`, respectively, to support the new Datumaro experimental `Dataset` API for classification tasks. These classes include new batching and sample handling logic. [[1]](diffhunk://#diff-53f178ea7a8ca9d9b33acc604531494be51a0878874cf77e18e50b6f605f4ab1R1-R161) [[2]](diffhunk://#diff-4b31185466d7508d334b7249c28440f86a51c99182b8c62cc2b9007ea78d08a4R1-R33)
* Introduced new sample entity classes (`OTXSample`, `ClassificationSample`) in `sample.py` to standardize sample structure and conversion from Datumaro items for classification.

**Factory and sampler logic updates:**

* Refactored `OTXDatasetFactory` in `factory.py` to instantiate datasets using either legacy or new Datumaro datasets, including conversion logic and label category extraction for the experimental API. [[1]](diffhunk://#diff-535a232efb8b265495f79f79fe09b2dc476c6cf5b18efb25c176ba0550d9c015R10-L18) [[2]](diffhunk://#diff-535a232efb8b265495f79f79fe09b2dc476c6cf5b18efb25c176ba0550d9c015L44-R57) [[3]](diffhunk://#diff-535a232efb8b265495f79f79fe09b2dc476c6cf5b18efb25c176ba0550d9c015L74-R91) [[4]](diffhunk://#diff-535a232efb8b265495f79f79fe09b2dc476c6cf5b18efb25c176ba0550d9c015R121-R130)
* Updated `BalancedSampler` in `balanced_sampler.py` to support both legacy and new dataset types, with logic to extract class indices appropriately. [[1]](diffhunk://#diff-349966a5ebd0a8cd149ab936e2b2f9fa73bbe685592b68f23543fb111bb988b5R12-R20) [[2]](diffhunk://#diff-349966a5ebd0a8cd149ab936e2b2f9fa73bbe685592b68f23543fb111bb988b5L46-R49) [[3]](diffhunk://#diff-349966a5ebd0a8cd149ab936e2b2f9fa73bbe685592b68f23543fb111bb988b5R67-R72)

**Compatibility and bug fixes:**

* Updated `_dispatch_label_info` in `models/base.py` to handle both legacy and new `LabelCategories` types for label info dispatching. [[1]](diffhunk://#diff-240c8860881c8a0937277a0673e2695f542567f0506a9d35182de969410d75b9R19) [[2]](diffhunk://#diff-240c8860881c8a0937277a0673e2695f542567f0506a9d35182de969410d75b9R888-R895)
* Added support for the `mps` device type in GPU memory monitoring callbacks for broader hardware compatibility.

**Testing and minor fixes:**

* Updated classification dataset tests to use the new sample entity (`ClassificationSample`) and dataset class (`OTXMulticlassClsDataset`). [[1]](diffhunk://#diff-614d0d8a62ab8af268e46c7d3eb4dfc31819ccc20c4fba573898d829dbf206c8L11-R14) [[2]](diffhunk://#diff-614d0d8a62ab8af268e46c7d3eb4dfc31819ccc20c4fba573898d829dbf206c8L27-R28) [[3]](diffhunk://#diff-614d0d8a62ab8af268e46c7d3eb4dfc31819ccc20c4fba573898d829dbf206c8L38-R39)

**Other minor changes:**

* Changed the `datumaro` dependency in `pyproject.toml` to the develop branch for OTX integration.
* Minor update to image conversion utility and class statistics logic for improved robustness. [[1]](diffhunk://#diff-79cb32270ba6193a200ab38fbab16ae45a34a752e8ce3123dcf88276b6921dc9R132) [[2]](diffhunk://#diff-f29e3ef832fa9ab1f32358061431d30bf261265823142f6093bc460f3effe2a8L328-R331)

### Checklist

### License

```python
```


lib/src/otx/data/utils/utils.py
```diff
@@ -325,10 +325,10 @@ def get_adaptive_num_workers(num_dataloader: int = 1) -> int | None:
 def get_idx_list_per_classes(dm_dataset: DmDataset, use_string_label: bool = False) -> dict[int | str, list[int]]:
     """Compute class statistics."""
     stats: dict[int | str, list[int]] = defaultdict(list)
-    labels = dm_dataset.categories().get(AnnotationType.label, LabelCategories())
     for item_idx, item in enumerate(dm_dataset):
         for ann in item.annotations:
             if use_string_label:
+                labels = dm_dataset.categories().get(AnnotationType.label, LabelCategories())
```
gdlg: Why moving this?

lib/src/otx/data/samplers/balanced_sampler.py
```diff
@@ -61,7 +64,11 @@ def __init__(
         super().__init__(dataset)
 
         # img_indices: dict[label: list[idx]]
-        ann_stats = get_idx_list_per_classes(dataset.dm_subset)
+        if isinstance(dataset.dm_subset, DatasetSubset):
+            ann_stats = get_idx_list_per_classes(dataset.dm_subset)
```
gdlg: If possible, I would suggest to add a `get_idx_list_per_classes()` method to the old `OTXDataset`. That way, we don’t need to differentiate here between OTXDataset and OTXDatasetNew.

lib/src/otx/data/factory.py
```diff
@@ -71,12 +76,19 @@ def create(
...
+                dataset = DatasetNew(ClassificationSample, categories={"label": categories})
+                for item in dm_subset:
+                    if len(item.media.data.shape) == 3:  # TODO: Account for grayscale images
+                        dataset.append(ClassificationSample.from_dm_item(item))
+                common_kwargs["dm_subset"] = dataset
             return OTXMulticlassClsDataset(**common_kwargs)
 
         if task == OTXTaskType.MULTI_LABEL_CLS:
-            from .dataset.classification import OTXMultilabelClsDataset
+            from otx.data.dataset.classification import OTXMultilabelClsDataset
```
gdlg: Why this change to the import path?

lib/src/otx/data/factory.py
```diff
@@ -71,12 +76,19 @@ def create(
...
-
+            from .dataset.classification_new import ClassificationSample, OTXMulticlassClsDataset
+
+            if isinstance(dm_subset, DmDataset):
+                categories = cls._get_label_categories(dm_subset, data_format)
+                dataset = DatasetNew(ClassificationSample, categories={"label": categories})
+                for item in dm_subset:
+                    if len(item.media.data.shape) == 3:  # TODO: Account for grayscale images
+                        dataset.append(ClassificationSample.from_dm_item(item))
+                common_kwargs["dm_subset"] = dataset
```
gdlg, line 87: Could you add an `else` here? If it’s not supported, we can raise a RuntimeError.

lib/src/otx/backend/native/models/base.py
```diff
@@ -884,6 +885,14 @@ def _dispatch_label_info(label_info: LabelInfoTypes) -> LabelInfo:
...
                 label_ids=[str(i) for i in range(len(label_info))],
             )
+        if isinstance(label_info, NewLabelCategories):
+            # Handle LabelCategories objects from datumaro
+            labels = label_info.labels
+            return LabelInfo(
+                label_names=labels,
+                label_groups=[labels],
+                label_ids=[str(i) for i in range(len(labels))],
+            )
```
gdlg: OTX label_info is used in many places and I think that replacing it with NewLabelCategories might have unintended consequences. We can think about refactoring it but for now, I would prefer to avoid refactoring the label_info and focus on updating the dataset class. So I would suggest to do the conversion of NewlabelCategories into LabelInfo in the constructor of the new OTXDataset.

Later when we have finished migrating to the new dataset class, we can consider streamlining the code by removing the LabelInfo and also improve the OTXSample.

lib/tests/unit/data/conftest.py
```diff
@@ -41,7 +41,7 @@
 
 
 @pytest.fixture(params=["bytes", "file"])
-def fxt_dm_item(request, tmpdir) -> DatasetItem:
+def fxt_dm_item(requeset, tmpdir) -> DatasetItem:
```
gdlg: Typo

lib/src/otx/backend/native/models/base.py
AlbertvanHouten: OTXModel is initialized with label info from the dataset. This is where the NewLabelCategories isinstance check comes from. It is then converted back to the OTX LabelInfo class before being used within the model. Indeed, in the future it would be best to remove this dispatching step.

lib/src/otx/data/factory.py
AlbertvanHouten: This might raise some problems if used with Geti. In my dataset of car images, I (unknowingly) had a grayscale image. This image has been exported in arrow format and reloaded with datumaro, but it only has 2 dimensions. Raising an error means we completely fail training. I could convert those image to be in RGB format for now or log a warning, but I don't think we should raise an error as of now.

lib/src/otx/data/utils/utils.py
AlbertvanHouten: labels are only needed in this if block. It's unnecessary to create this object for the other path.

lib/src/otx/data/samplers/balanced_sampler.py
AlbertvanHouten: Good point, fixed.

lib/src/otx/data/factory.py
AlbertvanHouten: reverted

lib/src/otx/data/factory.py
gdlg: Sorry for the confusion, I mean for the outer if: `if isinstance(dm_subset, DmDataset):`.

lib/src/otx/backend/native/models/base.py
gdlg: Ok, thanks for the explanation. Let’s leave it like that.

lib/pyproject.toml
```diff
@@ -27,7 +27,7 @@ classifiers = [
     "Programming Language :: Python :: 3.12",
 ]
 dependencies = [
-    "datumaro==1.10.0",
+    "datumaro @ git+https://github.com/open-edge-platform/datumaro.git@develop",
```
leoll2, line 30: Note: before the feature branch (`feature/datumaro`) can be merged to `develop`, this needs to be updated to point to an official datumaro release.

gdlg (dismissed):
LGTM, thanks Albert!

lib/src/otx/data/entity/sample.py
```diff
@@ -0,0 +1,118 @@
...
+        if isinstance(self.image, (np.ndarray, torch.Tensor)):
+            self.image = tv_tensors.Image(self.image)
+            return
+        msg = "OTXSample must have an image"
+        raise ValueError(msg)
+
+    @property
+    def masks(self) -> Mask | None:
+        """Get masks for the sample."""
+        return None
```
leoll2, line 44: What's the semantics of these properties `masks`, `bboxes`, ... ? Why do they `return None`?

lib/src/otx/data/entity/sample.py
```diff
@@ -0,0 +1,118 @@
...
+
+    @property
+    def img_info(self) -> ImageInfo | None:
+        """Get image information for the sample."""
+        if getattr(self, "_img_info", None) is None:
+            image = getattr(self, "image", None)
+            if image is not None and hasattr(image, "shape") and len(image.shape) == 3:
+                img_shape = image.shape[:2]
+            else:
+                return None
```
leoll2, line 74: In what situations is this property expected to return None? Wouldn't it be better to raise an exception if it's not possible to extract image information from the sample?

lib/src/otx/data/entity/sample.py
```diff
@@ -0,0 +1,118 @@
...
+    @img_info.setter
+    def img_info(self, value: ImageInfo) -> None:
+        self._img_info = value
+
+
+class ClassificationSample(OTXSample):
+    """OTXDataItemSample is a base class for OTX data items."""
+
+    image: np.ndarray | tv_tensors.Image = image_field(dtype=pl.UInt8)
+    label: torch.Tensor = label_field(pl.Int32())
```
leoll2, line 91: If the `label` is always a torch type, why can image be a numpy type? The double type `np.ndarray | tv_tensors.Image` may be problematic for the code that consumes this class, because it needs to handle both cases (image as `np.ndarray` and image as `tv_tensors.Image`).

tests/unit/data/entity/test_sample.py
```diff
@@ -0,0 +1,173 @@
+# Copyright (C) 2023-2024 Intel Corporation
```
leoll2: Outdated copyright header

tests/unit/data/dataset/test_classification_new.py
```diff
@@ -0,0 +1,66 @@
+# Copyright (C) 2023-2024 Intel Corporation
```
leoll2: Outdated copyright header

tests/unit/data/dataset/test_base_new.py
```diff
@@ -0,0 +1,261 @@
+# Copyright (C) 2023-2024 Intel Corporation
```
leoll2: Outdated copyright header

lib/src/otx/data/samplers/class_incremental_sampler.py
```diff
@@ -65,7 +64,7 @@ def __init__(
         super().__init__(dataset)
 
         # Need to split new classes dataset indices & old classses dataset indices
-        ann_stats = get_idx_list_per_classes(dataset.dm_subset, True)
+        ann_stats = dataset.get_idx_list_per_classes(True)
```
leoll2:
```suggestion
        ann_stats = dataset.get_idx_list_per_classes(use_string_label=True)
```
Pass arguments by keyword when the meaning of the value is not obvious

lib/src/otx/data/dataset/base_new.py
```diff
@@ -0,0 +1,161 @@
+# Copyright (C) 2023 Intel Corporation
```
leoll2: Outdated copyright header

lib/src/otx/data/dataset/base.py
```diff
@@ -196,3 +197,18 @@ def _get_item_impl(self, idx: int) -> OTXDataItem | None:
     def collate_fn(self) -> Callable:
         """Collection function to collect KeypointDetDataEntity into KeypointDetBatchDataEntity in data loader."""
         return OTXDataItem.collate_fn
+
+    def get_idx_list_per_classes(self, use_string_label: bool = False) -> dict[int | str, list[int]]:
+        """Compute class statistics."""
```
leoll2: Please add a better docstring to explain what the method does and what's the structure of the returned dictionary

lib/src/otx/data/dataset/classification_new.py
```diff
@@ -0,0 +1,35 @@
...
+    def __init__(self, **kwargs) -> None:
+        """Initialize OTXMulticlassClsDataset.
+
+        Args:
+            **kwargs: Keyword arguments to pass to OTXDataset
+        """
+        kwargs["sample_type"] = ClassificationSample
+        super().__init__(**kwargs)
+
+    def get_idx_list_per_classes(self, use_string_label: bool = False) -> dict[int, list[int]]:
```
leoll2, line 24: This method is already implemented in the parent class `OTXDataset`. Why does this class need to override it?

lib/src/otx/data/dataset/base_new.py
```diff
@@ -0,0 +1,161 @@
...
+        bboxes=[item.bboxes for item in items] if any(item.bboxes is not None for item in items) else None,
+        keypoints=[item.keypoints for item in items] if any(item.keypoints is not None for item in items) else None,
+        polygons=[item.polygons for item in items if item.polygons is not None]
+        if any(item.polygons is not None for item in items)
+        else None,
+        imgs_info=[item.img_info for item in items] if any(item.img_info is not None for item in items) else None,
+    )
+
+
+class OTXDataset(TorchDataset):
```
leoll2, line 69: If the old and new `OTXDataset` have the same API, I suggest to add some black-box functional tests that instantiate both classes, apply a sequence of operations to each, then finally assert that the respective results are identical.

lib/src/otx/data/entity/sample.py
AlbertvanHouten: A large portion of the OTX codebase depends on the [OTXDataItem](https://github.com/open-edge-platform/training_extensions/blob/6dbad4d62d25d995c088f7d67e67b4e7c84b81ac/lib/src/otx/data/entity/torch/torch.py#L35). For example, all the transforms in https://github.com/open-edge-platform/training_extensions/blob/8418221b4e065d14b6f2c223f1ee297b80fd64c8/lib/src/otx/data/transform_libs/torchvision.py. 

For now, we will mimic that functionality with the OTXSample to not break any logic. In the future, OTXDataItem should be replaced with OTXSample but this will require a larger refactor, including updating all the transforms to not take OTXDataItems as input.

lib/src/otx/data/dataset/classification_new.py
AlbertvanHouten: I believe the implementation of this method will depend on the sample type. If it will be uniform across samples, it can be moved back to the new OTXDataset. The new OTXDataset does not implement this method at the moment.

lib/src/otx/data/entity/sample.py
AlbertvanHouten: It is very hard to understand all usages of OTXDataItem.img_info and if they would work with None or not. For now, it is better to maintain the existing logic until we start to factor out OTXDataItem.

lib/src/otx/data/entity/sample.py
AlbertvanHouten: This is again existing OTX logic from OTXDataItem. The OTX codebase is already taking the different types into account.

lib/src/otx/data/dataset/base.py
AlbertvanHouten: This is copied and pasted from https://github.com/open-edge-platform/training_extensions/pull/4606/files#diff-f29e3ef832fa9ab1f32358061431d30bf261265823142f6093bc460f3effe2a8 

I'll update this function doc, but I am not really planning to address all these previously existing issues in the OTX codebase within this feature branch. We should create a task to update the linter ruleset and enforce this throughout the codebase.

lib/src/otx/data/dataset/base.py
```diff
@@ -196,3 +197,18 @@ def _get_item_impl(self, idx: int) -> OTXDataItem | None:
     def collate_fn(self) -> Callable:
         """Collection function to collect KeypointDetDataEntity into KeypointDetBatchDataEntity in data loader."""
         return OTXDataItem.collate_fn
+
+    def get_idx_list_per_classes(self, use_string_label: bool = False) -> dict[int | str, list[int]]:
+        """Get a dictionary with class labels (string/int) as keys and lists of sample indices as values."""
```
eugene123tw: What’s the intended use case for use_string_label? Could you also add a Google-style docstring explaining it? For example:


```python
"""Get a dictionary mapping class labels (string or int) to lists of samples.

Args:
    use_string_label (bool): If True, use string class labels as keys.  
        If False, use integer indices as keys.
"""

```

lib/src/otx/data/factory.py
```diff
@@ -71,8 +76,18 @@ def create(
             return OTXAnomalyDataset(task_type=task, **common_kwargs)
 
         if task == OTXTaskType.MULTI_CLASS_CLS:
-            from .dataset.classification import OTXMulticlassClsDataset
-
+            from .dataset.classification_new import ClassificationSample, OTXMulticlassClsDataset
+
+            if isinstance(dm_subset, DmDataset):
+                categories = cls._get_label_categories(dm_subset, data_format)
+                dataset = DatasetNew(ClassificationSample, categories={"label": categories})
```
eugene123tw, line 83: I don’t fully understand the logic here. At first glance, it seems like `dm.DatasetNew` is meant to replace `dm.Dataset`. But the code still relies on the original `dm.Dataset`, which makes the workflow a bit unclear.

The current approach looks like this:

```python
original_dm_dataset = dm.Dataset.import_from(data_root, format=data_format)
new_dm_dataset = dm.DatasetNew(ClassificationSample, categories={"label": categories})
for item in original_dm_dataset:
    if len(item.media.data.shape) == 3:  # TODO(albert): Handle grayscale images
        new_dm_dataset.append(ClassificationSample.from_dm_item(item))
```

Why do we need to create a new `DatasetNew` and then manually append `ClassificationSample` instances to it?

Wouldn’t it be cleaner to write something like this instead?

```python
dataset = dm.DatasetNew.import_from(data_root, format=data_format)
classification_dataset = dataset.to(ClassificationSample)
```

eugene123tw (commented):
Thanks, Albert, for the refactoring! I just have a few comments, but overall it looks great.

lib/src/otx/data/factory.py
gdlg: That’s indeed what we intended but there are a couple of issues:
* Dataset import is not yet implemented in dm.DatasetNew. Hopefully we can resolve that soon
* To entirely replace dm.Dataset with dm.DatasetNew in OTX, we need to do a more significant refactoring of OTX since this factory takes a dm.Dataset which has been loaded elsewhere. It’s also difficult to change that now because we haven’t migrated all the tasks so some of them still expect the old dm.Dataset.

We are also trying to break the work into smaller PRs, so we need this kind of temporary glue code. Without it, we would have to refactor many files and the PR would become unwieldy.

So this is temporary solution until the import is fully supported in dm.DatasetNew and we have finished migrated the tasks to dm.Dataset. Once that’s completed, we can update OTX to directly import the data as dm.DatasetNew.

eugene123tw: Sounds good! Thanks Greg!

eugene123tw (dismissed):

gdlg (approved):

PR #4624, opened 2025-09-01
Geti Tune new entities design

### LINKED ISSUE #4573: Design API for label management
The model of the pipeline should predict classes (labels) specified by the user at configuration stage.

Write a design to define:
- an API to configure the task (classification, detection, segmentation) and relative labels
- how to persist the above information

leoll2:
### Summary

This PR introduces a new design proposal for Geti Tune API and server architecture, providing an overview of both the high-level entities and the storage-level representation.

- [x] Add ER diagram (entity-relationship) of all entities and their relationships
- [x] Add diagram of all database tables and objects
- [x] Write document to briefly describe all entities
- [x] Update design document about pipeline management
- [x] Update design document about dataset management
- [x] Add design document about labels


docs/geti_tune/api.md
```diff
@@ -0,0 +1,123 @@
...
+| `GET`    | `/api/projects/<id>/dataset/views`            | -         | list of views | List the dataset views           |
+| `GET`    | `/api/projects/<id>/dataset/views/<id>`       | -         | view info     | Get info about a dataset view    |
+| `GET`    | `/api/projects/<id>/dataset/views/<id>/items` | -         | list of items | List the items in a dataset view |
+| `POST`   | `/api/projects/<id>/dataset/views/<id>/items` | items ids | -             | Add items to a dataset view      |
+| `DELETE` | `/api/projects/<id>/dataset/views/<id>/items` | items ids | -             | Remove items from a dataset view |
+| `DELETE` | `/api/projects/<id>/dataset/views/<id>`       | -         | -             | Delete a dataset view            |
+
+## Models
+
+| Method   | Path                                          | Payload                         | Return             | Description                              |
```
MarkRedeman, line 95: Do we not allow users to upload their own models? (Either trained from Geti, externally, or from a different project)

docs/geti_tune/entities.md
```diff
@@ -0,0 +1,45 @@
...
+The collected frames are stored in a _dataset_ (`Dataset`), These frames are saved in a dataset, so that the user can
+examine them and eventually use them, for example, to make the model more robust via fine-tuning.
+It is also possible for the user to manually upload images to a dataset, and annotate them using the built-in
+annotation tool. A dataset is therefore a collection of image-like items (`DatasetItem`), each of which may have
+some _annotation_ data (`Annotation`) that is either a _user annotation_ or a _model prediction_.
+
+In addition to annotations, which are used when training and evaluating models, it is also possible to assign
+custom _tags_ to each media. A tag is a simple label that has no effect on model training or evaluation, but can be
+used to organize and filter the dataset items. Filtered data, based on tags and/or other properties (e.g. timestamp),
+can be used to create a _view_ in the dataset (`DatasetView`). A dataset view basically represents a subset of the
```
MarkRedeman, line 28: Can you elaborate on what the intention of a `DatasetView` is? Is this similar to the `POST dataset:query` endpoint we have at Geti?

docs/geti_tune/api.md
leoll2: No, it won't be supported for now. As the name suggests, Geti Tune focuses on models that can be fine-tuned. If we allow uploading models from other applications, then we have to deal with challenges and limitations:

1. Models converted to OpenVINO format can be used for inference, but they can't be fine-tuned
2. The models may have an incompatible input/output layout
3. The `rt_info` section of the model XML may not contain the necessary information about labels

docs/geti_tune/entities.md
leoll2: This PR is still a draft, I'm rewriting the datasets document.

In short, the dataset view is a custom subset of the dataset. The user can create views with any items he wants, and he can use views as input for other workflows (evaluation, export, ...). Views can be created through a filtering query, but it is also possible to assign/remove items by id.

docs/geti_tune/media/er/full.svg
jpggvilaca, line 1: I think **DataCollectionPolicy** should belong to either **Dataset** or **Project**. Perhaps project since it can possibly be abstracted or reused to serve a different purpose, but definitely outside of pipeline.

Another minor remarks regarding the naming:
1) InferencePipeline can just be Pipeline, as we dont support any other types.
2) LabelSchema can be Task

jpggvilaca (commented):
The same way we wont add project/dataset Import/export, i think Dataset Views should also be only taken care "later".

docs/geti_tune/media/er/full.svg
leoll2: _Data collection_ is one of the stages of the pipeline, responsible for selecting the frames to add to the dataset based on one or more policy, that take into account various information (prediction confidence, detected anomalies in the data/model, ...). That is the reason why I put the data collection policy configuration in the pipeline.
Since the pipeline is 1:1 with the project, data collection is indirectly linked to the project too. Logically, it also has a connection with the dataset, in the sense that it affects what items enter the dataset. I think the following question may help us decide whether the dataset collection policy belongs to the pipeline or not: if you export a pipeline, do you want to include the dataset collection configuration in addition to the source/sink? Imho the answer could be yes, but I acknowledge it's subjective. @Radwan-Ibrahim do you have any preference?

EDIT: I checked with Radwan, it should be part of the pipeline.

> InferencePipeline can just be Pipeline, as we dont support any other types.

I agree, indeed in the REST API it appears as "pipeline". The full name "inference pipeline" may be helpful in the documentation to clarify the purpose of this pipeline; in the future, the term "pipeline" may generate some ambiguity with training pipelines.

> LabelSchema can be Task

Also agree, Radwan prefers LabelSchema. I don't have a strong preference. Note that in the REST API we never use the term "label_schema".

leoll2:
> The same way we wont add project/dataset Import/export, i think Dataset Views should also be only taken care "later".

Absolutely, they can be implemented at a later moment. I just wanted to include them in the design to show that we still have flexibility to run certain workflows on part of the dataset, despite having a single dataset per project.

docs/geti_tune/diagrams/storage/full.puml
```diff
@@ -0,0 +1,169 @@
...
+  updated_at : TIMESTAMP NOT NULL
+  --
+  'Task-related details
+  task_type : TEXT NOT NULL
+  exclusive_labels : BOOLEAN NOT NULL
+  --
+  'Pipeline-related details
+  running : BOOLEAN NOT NULL
+  fk(source_id) : UUID
+  fk(sink_id) : UUID
```
leoll2: @itallix observed that it may be convenient to store the pipeline-related information in a separate table, for better extensibility if we ever change the project-pipeline relationship from 1:1 to 1:N (i.e. allow multiple pipelines per project). Implementation-wise, it doesn't add much complexity thanks to SQLalchemy ORM auto-join.

docs/geti_tune/dataset.md
```diff
@@ -1,55 +1,106 @@
...
-analytical workflows (training, evaluation, data exploration, ...) and other batch operations (conversion, export, ...).
-In Geti Tune, the same media may be assigned to multiple datasets: this is a simple yet powerful property,
-that gives the user a lot of flexibility in choosing which subset of the data to use for a given workflow.
+In Geti Tune, the _dataset_ is the container for all the media and related annotations within a project.
+New items can be added to the dataset from:
 
-![Diagram showing the dataset manager](media/dataset-management.jpg)
+- the inference pipeline, whose data collection stage implements policies to select which frames to save,
+  along with the predictions generated by the model.
+- the user, who manually uploads images through the web interface or the REST API.
```
itallix, line 10: Could it also be useful to create a dataset directly from a folder in the filesystem?  
This way, we wouldn't need to copy binaries across; we could simply create the corresponding metadata in the database for the existing images in the specified folder.

leoll2: It is an interesting idea, avoiding copies can be faster and storage-friendly. However, there are a few aspects that may be tricky: 
1. Ambiguous ownership of the data. If the user deletes one of these dataset item through the UI, should we delete the media file from the folder? Conversely, If the user deletes an image directly from the folder, what happens to the dataset item?
2. We have to handle situations and errors when some of the files or folders don't have the right r/w permissions, or the permissions change at a later time
3. The folder must be mounted on the container when starting it
4. We still have to validate that the images are readable and generate a thumbnail for them (in the same folder?). Also, we currently assume that the media are stored under `projects/<id>/dataset/<id>.<extension>`, which allows direct access without looking up the path in the DB. For media in external folders, we could maybe create a symlink to replicate the same layout?

Given the extra complexity, I'd suggest to revisit this idea at a later time. Btw, we already have a mechanism to select a folder as pipeline input. It's a bit different - and it copies data - but it also generates predictions for each item. Perhaps we can enhance the pipeline to create a reference (symlink) for images that are acquired from a 'folder' source type, instead of a making a hard-copy of the file.

docs/geti_tune/labels.md
```diff
@@ -0,0 +1,114 @@
...
+
+Labels are uniquely identified by a UUID, which is immutable. They also have a name and other frontend-related
+attributes (color, hotkey, etc.) that can be modified by the user.
+
+## Storage
+
+### DB schema
+
+```
+,-----------------------------------.
```
warrkan, line 27: Is `LabelSchema` omitted here for the sake of simplicity? Information here and below contradicts a bit with ER diagram

docs/geti_tune/entities.md
```diff
@@ -0,0 +1,45 @@
...
+An example of a task is "object detection" with labels "car", "person" and "bicycle".
+
+Each project has an `InferencePipeline`, that is a sequence of processing steps that acquires a stream of
+data from a _source_ (`Source`), runs inference using a _model_ (`ModelRevision`) like YOLO, processes the results
+and eventually sends them to a _sink_ (`Sink`). Examples of sources are video files or IP cameras, while examples of
+sinks are folders on the local filesystem or messaging frameworks like MQTT or ROS2.
+After inference, the frames and their predictions are analyzed to identify potential anomalies or out-of-distribution
+samples (_model monitoring_), and these may be collected based on one or more policies (`DataCollectionPolicy`).
+
+The collected frames are stored in a _dataset_ (`Dataset`), These frames are saved in a dataset, so that the user can
```
itallix:
```suggestion
The collected frames are stored in a _dataset_ (`Dataset`), so that the user can
```

docs/geti_tune/entities.md
```diff
@@ -0,0 +1,45 @@
...
+When the user decides to train a new model - optionally fine-tuning a previous model - Geti Tune creates a snapshot
+of the dataset, namely a _dataset revision_ (`DatasetRevision`), that captures the exact state of the dataset when
+that model is trained. A dataset revision is therefore an immutable entity, and is not affected by subsequent changes
+to the main dataset, for example when images are added or removed. In other words, it is always possible to access the
+exact data used to train a specific model, even if an image has been deleted or the annotations have been modified.
+
+A _model revision_ (`ModelRevision`) is a model instance obtained by fine-tuning the weights of another model revision,
+or a set of publicly available pre-trained weights. For each model, it is important to store not only the training
+dataset revision, but also the exact label schema used to train it (`LabelSchemaRevision`). This is because the user
+may later choose to add, remove or rename labels, while the stricture of the model output is fixed at training time.
```
itallix:
```suggestion
may later choose to add, remove or rename labels, while the structure of the model output is fixed at training time.
```

docs/geti_tune/pipeline.md
```diff
@@ -69,10 +82,10 @@ Each sink is identified by a unique id, and optionally a friendly name chosen by
 
 ### Models
 
-Geti Tune supports models exported from Geti in OpenVINO IR format (.xml + .bin). All tasks are supported, including
-classification, detection, segmentation, anomaly and keypoint detection.
+A pipeline performs inference using models OpenVINO IR format (.xml + .bin), converted after fine-tuning.
```
itallix:
```suggestion
A pipeline performs inference using models in OpenVINO IR format (.xml + .bin), converted after fine-tuning.
```

docs/geti_tune/pipeline.md
```diff
@@ -69,10 +82,10 @@ Each sink is identified by a unique id, and optionally a friendly name chosen by
...
+The supported task types include image classification, object detection and instance segmentation.
```
itallix, line 86: This can be probably omitted because it duplicates the section from labels.md:

> In Geti Tune, every _project_ addresses a specific _task_. The currently supported task types are:
> 
> - **Classification**: Categorize images into one (_multiclass_) or more (_multilabel_) classes.
> - **Detection**: Identify and localize objects in images using bounding boxes.
> - **Instance Segmentation**: Identify and segment objects in images using polygons.

leoll2: I'd keep it, it is a minor redundancy that provides context to the reader, without him having to jump to another file.

docs/geti_tune/labels.md
leoll2: The ER diagram (`er/full.puml`) represents the high-level entities and their relationships. The other diagram (`storage/full.puml`) shows it from the database perspective. They are similar, although entities don't always correspond 1:1 with tables. In case of LabelSchema, the static information about the task is in the `projects` table, because there is really no point to have a separate `label_schemas` table for it. While the labels, whose properties can change dynamically (name, color, ...) are stored in `labels`.

jpggvilaca (approved):

itallix (approved):

PR #4648, opened 2025-09-04
Monitor model throughput

### LINKED ISSUE #4538: Endpoint to monitor model throughput
Endpoint: `GET /api/pipelines/ace3f1da-fdd9-4048-a95e-a647ed969442/metrics`

Response:
```json
{
  "time_window": {...},
  "inference": {...},
  "throughput": {
    "avg_requests_per_second": 66.7,
    "total_requests": 4000,
    "max_requests_per_second": 85.2
  }
}
```

A-Artemis:
### Summary

This PR adds model throughput calculations and retrieval to the metrics service.

### How to test
```http://geti-tune.localhost/api/projects/9d6af8e8-6017-4ebe-9126-33aae739c5fa/pipeline/metrics```

image

### Checklist

### License

```python
```


backend/app/schemas/metrics.py
```diff
@@ -41,7 +50,12 @@ class PipelineMetrics(BaseModel):
             "example": {
                 "time_window": {"start": "2025-08-25T10:00:00Z", "end": "2025-08-25T10:01:00Z", "time_window": 60},
                 "inference": {
-                    "latency": {"avg_ms": 15.1, "min_ms": 12.3, "max_ms": 30.4, "p95_ms": 25.4, "latest_ms": 15.6}
+                    "latency": {"avg_ms": 15.1, "min_ms": 12.3, "max_ms": 30.4, "p95_ms": 25.4, "latest_ms": 15.6},
+                    "throughput": {
+                        "avg_inferences_per_second": 66.7,
+                        "total_inferences": 4000,
+                        "peak_inferences_per_second": 85.2,
```
leoll2:
```suggestion
                        "avg_requests_per_second": 66.7,
                        "total_requests": 4000,
                        "max_requests_per_second": 85.2,
```
Small renaming suggestions. "requests" instead of "inferences" (which I think is a contraption of inference requests), and "max" instead of "peak" for consistency with the latency attributes ("max_ms").

backend/app/services/metrics_service.py
```diff
@@ -99,6 +160,12 @@ def reset(self) -> None:
...
     def __del__(self):
         try:
-            self._shm.close()
+            with MetricsService._lock_class:
+                MetricsService._instance_count -= 1
+                if MetricsService._instance_count <= 0 and MetricsService._shared_memory:
+                    MetricsService._shared_memory.close()
+                    MetricsService._shared_memory.unlink()
+                    MetricsService._shared_memory = None
+                    logger.info("Cleaned up shared memory")
```
leoll2: Why this complex cleanup with reference-counting instead of letting the Scheduler allocate and release the shared memory?

backend/app/services/metrics_service.py
```diff
@@ -36,13 +37,40 @@ class LatencyMeasurement(NamedTuple):
 class MetricsService:
     """Process-safe metrics service using shared memory for model metric data"""
 
+    _shared_memory: ClassVar[SharedMemory | None] = None
+    _instance_count: ClassVar[int] = 0
+    _lock_class: ClassVar[LockType] = Lock()
```
itallix: This design will not work reliably in a multiprocess setting because each process (`main` and `inference`) gets its own separate instance of `MetricsService`, including separate instances of the lock and shared memory variables. For safe and correct IPC, primitives like shared memory and locks should be instantiated once in a central place (e.g., the `Scheduler`) and then explicitly passed down to downstream processes. This ensures all processes coordinate on the same shared resources and synchronization objects, preventing race conditions and data inconsistencies.

backend/app/services/metrics_service.py
```diff
@@ -90,6 +91,39 @@ def get_latency_measurements(self, model_id: UUID, time_window: int = 60) -> lis
...
+
+        Args:
+            model_id: UUID of the model to filter measurements
+            time_window: Time window in seconds to look back for measurements (default 60s)
+
+        Returns: Tuple of (total_requests, list of (timestamp, inference_count) per second)
+        """
+        cutoff_time = datetime.now(UTC).timestamp() - time_window
+        with self._lock:
+            arr = self._array.copy()
```
itallix, line 108: It's safe to perform calculations outside of the lock as we are working with the copy:

```python
with self._lock:
  arr = self._array.copy()

calculations_on_array(arr)
```

backend/app/services/metrics_service.py
```diff
@@ -90,6 +91,39 @@ def get_latency_measurements(self, model_id: UUID, time_window: int = 60) -> lis
...
+
+            # Count inferences per second
+            inferences_per_second: dict[int, int] = defaultdict(int)
+            total_requests = 0
+
+            for entry in arr:
+                if entry["timestamp"] < cutoff_time or entry["timestamp"] == 0.0:
+                    continue
+                if entry["model_id"] == str(model_id):
```
itallix: I recommend to perform string conversion once outside of `for loop`.

backend/app/services/metrics_service.py
```diff
@@ -90,6 +91,39 @@ def get_latency_measurements(self, model_id: UUID, time_window: int = 60) -> lis
...
+        cutoff_time = datetime.now(UTC).timestamp() - time_window
+        with self._lock:
+            arr = self._array.copy()
+
+            # Count inferences per second
+            inferences_per_second: dict[int, int] = defaultdict(int)
+            total_requests = 0
+
+            for entry in arr:
+                if entry["timestamp"] < cutoff_time or entry["timestamp"] == 0.0:
```
itallix: `if` can be extended with the condition below?

A-Artemis: Yes it can be, but because I don't see it being a major performance increase, I find that splitting them shows the two cases clearly. 

Or are you suggesting a different `if`?

itallix: I was considering using a single `if` condition so that the code below wouldn’t require nested formatting. But separating them is also fine.

itallix (approved):
LGTM!

Issue #4630, opened 2025-09-01
Implement endpoints to (de)activate a pipeline


leoll2:
### Activate a pipeline

Endpoint: `/api/projects/<id>/pipeline:enable`

Response: empty (204)

### Deactivate a pipeline

Endpoint: `/api/projects/<id>/pipeline:disable`

Response: empty (204)

Issue #4629, opened 2025-09-01
Implement endpoints to get and reconfigure a pipeline


leoll2:
### Get pipeline info

Endpoint: `GET /api/projects/<id>/pipeline`

Response:
```json
{
  "source": {"id": "550e8400-e29b-41d4-a716-446655440007", "name": "my webcam", ...},
  "sink": {"id": "550e8400-e29b-41d4-a716-446655440008", ...},
  "model": {"id": "550e8400-e29b-41d4-a716-446655440009", "architecture": "YOLO-X" ...},
  "status": "idle",
}
```

If the source, sink or model are not configured, return `null` in the respective field.

### Reconfigure pipeline

Endpoint: `PATCH /api/projects/<id>/pipeline`

Payload:
```json
{
  "source_id": "550e8400-e29b-41d4-a716-446655440007",
  "sink_id": ...,
  "model_id": ...,
}
```

Response: same as `GET /api/projects/<id>/pipeline`

It's possible to specify one or more keys between `source_id`, `sink_id` and `model_id` (partial updates).

Issue #4627, opened 2025-09-01
Implement endpoint to edit labels


leoll2:
Endpoint: `PATCH /api/projects/<id>/labels`

Payload:
```json
{
  "labels_to_edit": [{"id": "550e8400-e29b-41d4-a716-446655440003", "new_name": "horse", "new_color": "#8090C1", "new_hotkey": "h"}],
  "labels_to_add": [{"name": "Archbishop", "color": "#A070C1", "hotkey": "a"}],
  "labels_to_remove": [{"id": "550e8400-e29b-41d4-a716-446655440004"}]
}
```

Response: new labels after the update, in the same format as the response of `GET /api/projects/<id>/labels`.

Validate the request feasibility before executing it. Example: if the id of the label to remove doesn't exist, the request should be rejected with error 404 and none of the other actions (rename, add labels) should be committed.

PR #4658, opened 2025-09-05
Projects endpoints with Pipelines refactor

### LINKED ISSUE #4626: Implement endpoints to create and get projects
Based on the design in #4624, implement the following endpoints

### Create a new project

Endpoint: `POST /api/projects`
Payload:
```json
{
  "name": "animals",
  "task": {
    "type": "classification",
    "exclusive_labels": true,
    "labels": [{"name": "cat"}, {"name": "dog"}]
  }
}
```
Response:
```json
{
  "id": "7b073838-99d3-42ff-9018-4e901eb047fc"
  "name": "animals",
  "task": {
    "task_type": "classification",
    "exclusive_labels": true,
    "labels": [{"name": "cat"}, {"name": "dog"}]
  }
}
```

### List projects

Endpoint: `GET /api/projects`

Response:
```json
[{"id": "7b073838-99d3-42ff-9018-4e901eb047fc", "name": "animals", ...}, ...]
```

### Get project info

Endpoint: `GET /api/projects/{id}`

Response:
```json
{"id": "7b073838-99d3-42ff-9018-4e901eb047fc", "name": "animals", ...}
```

### Delete a project

Endpoint: `DELETE /api/projects/{id}`

Return 409 Conflict if the pipeline is active

### Get project labels

Endpoint: `DELETE /api/projects/{id}`

```json
{
  "task_type": "classification",
  "exclusive_labels": true,
  "labels": [{"name": "cat"}, {"name": "dog"}]
}
```

## Database

Image

itallix:
### Summary

Resolves #4626, #4629 and #4630 

This PR introduces the following changes:

- Adds new projects endpoints to the API.
- Modifies existing pipeline endpoints and updates the logic behind to accommodate the new model relationships between projects and pipelines.

### How to test

1. Remove the old database file (if it exists).
2. Run the server in seed mode:

```bash
SEED_DB=true ./run.sh
```

3. The server will start, and inference will produce output in the configured sink.
4. List projects using:

```bash
curl http://localhost:7860/api/projects  # This should return the list of available projects
```
5. Try to enable/disable pipeline:

```bash
# IDs remain the same as they are fixed in seed mode
curl -X POST localhost:7860/api/projects/9d6af8e8-6017-4ebe-9126-33aae739c5fa/pipeline:disable
# The pipeline will stop. Try reloading the output folder - no new predictions will be generated.

curl -X POST localhost:7860/api/projects/9d6af8e8-6017-4ebe-9126-33aae739c5fa/pipeline:enable
# The pipeline will resume. Try reloading the output folder - new predictions will be generated.
```

### Checklist

### License

```python
```


backend/app/core/scheduler.py
```diff
@@ -10,14 +10,13 @@
...
-from app.services.metrics_service import SHM_NAME, SIZE
-from app.utils.singleton import Singleton
+from app.services.metrics_service import SIZE
 from app.workers import dispatching_routine, frame_acquisition_routine, inference_routine
 
 logger = logging.getLogger(__name__)
 
 
-class Scheduler(metaclass=Singleton):
+class Scheduler:
```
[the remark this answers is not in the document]
itallix: We can omit `Singleton` because `Scheduler` is only instantiated once at app startup.

backend/app/main.py
```diff
@@ -47,13 +49,12 @@
...
-app.include_router(datasets.router)
-app.include_router(models.router)
-app.include_router(system.router)
-app.include_router(webrtc.router)
+# Include all API routers from the endpoints package
+for module_info in pkgutil.iter_modules(endpoints.__path__):
+    module_name = module_info.name
+    module = importlib.import_module(f"app.api.endpoints.{module_name}")
+    if hasattr(module, "router"):
+        app.include_router(module.router)
```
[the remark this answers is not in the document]
itallix: The `endpoints` package is structured so that each module contains a single router. Using dynamic inclusion with `importlib` allows us to automate the setup process, reducing manual effort and saving lines of code as more endpoints are added.

backend/app/api/endpoints/projects.py
```diff
@@ -0,0 +1,136 @@
...
+}
+
+
+@router.post(
+    "",
+    status_code=status.HTTP_201_CREATED,
+    response_model=Project,
+    responses={
+        status.HTTP_201_CREATED: {"description": "Project successfully created"},
+        status.HTTP_409_CONFLICT: {"description": "Project already exists"},
```
A-Artemis, line 65: A few questions here.

What are the constraints for a project already existing? Is it matched on name or ID? Can the user create with an ID? Can projects exist with the same name?

backend/app/services/parent_process_guard.py
```diff
@@ -26,7 +28,11 @@ def parent_process_only[T, **P, R](func: Callable[Concatenate[T, P], R]) -> Call
 
     @wraps(func)
     def wrapper(self: T, *args: P.args, **kwargs: P.kwargs) -> R:
-        if mp.parent_process() is not None:
+        settings = get_settings()
+        # Check conditions for the watcher process. In this case, the web server will be not in the main process.
+        # We can ignore the guard in this setup.
+        is_watcher_running = settings.debug and settings.environment == "dev"
+        if not is_watcher_running and mp.parent_process() is not None:
```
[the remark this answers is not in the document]
itallix: I will remove it for now. Should be addressed with reload mode fix.

backend/app/api/endpoints/projects.py
itallix: - The `id` field is the primary key for projects. If a user attempts to create a new project with an existing id, the operation will fail and we return a `409 Conflict` response.
- The `name` field for projects is **not** unique (unlike for sinks and sources). This allows multiple projects to share the same name, which can be useful if we want to reuse names across different task_types.
- Users can specify the `id` when creating a project, but it must be unique (This might be useful for the UI, as it allows pre-generating IDs upfront).

@leoll2 Please confirm if the non-unique name is intentional.

backend/tests/unit/services/test_pipeline_service.py
```diff
@@ -1,45 +1,11 @@
...
 from app.services import MetricsService, PipelineService
-from app.services.metrics_service import SHM_NAME, SIZE
-
-
-@pytest.fixture(scope="module", autouse=True)
-def fxt_metrics_mem():
-    shm = SharedMemory(name=SHM_NAME, create=True, size=SIZE)
-    yield
-    shm.close()
-    shm.unlink()
```
leoll2, line 20: Did you remove this fixture intentionally? I don't know if and where it is needed, asking because `autouse=True` so removing it may silently break some tests.

backend/app/api/endpoints/projects.py
leoll2: I don't have a strong opinion about the uniqueness constraint on the project name. Allowing duplicate names (like in Geti) is convenient if we implement a project export/import functionality, because we can import a project with its original name without having to check for collisions and/or renaming it.

backend/app/db/schema.py
```diff
@@ -24,19 +24,37 @@ class SourceDB(Base):
     updated_at: Mapped[datetime] = mapped_column(DateTime, server_default=func.current_timestamp())
 
 
+class ProjectDB(Base):
+    __tablename__ = "projects"
+
+    id: Mapped[str] = mapped_column(Text, primary_key=True, default=lambda: str(uuid4()))
+    name: Mapped[str] = mapped_column(String(255), nullable=False)
+    task_type: Mapped[str] = mapped_column(String(20), nullable=False)
```
leoll2: 20 characters seems too short, the string "instance_segmentation" is already longer than that. iirc SQLite doesn't care about this specifiers, still I think it would be better to allow more space in the task type string.

backend/app/db/schema.py
```diff
@@ -24,19 +24,37 @@ class SourceDB(Base):
...
+    exclusive_labels: Mapped[bool] = mapped_column(Boolean, default=False)
+    labels: Mapped[list[str]] = mapped_column(JSON, nullable=False)
```
leoll2, line 34: In the design, labels are first-class entities, stored in a separate table `labels`. This makes it easier to get and update labels individually.

image

backend/app/repositories/base.py
```diff
@@ -32,9 +32,10 @@ def save(self, item: ModelType) -> ModelType:
 
     def update(self, item: ModelType) -> ModelType:
         item.updated_at = datetime.now()  # type: ignore[attr-defined]
-        self.db.merge(item)
+        updated = self.db.merge(item)
         self.db.flush()
-        return item
+        self.db.refresh(updated)
+        return updated
```
leoll2, line 38: Can you explain this change?

backend/app/api/endpoints/pipelines.py
```diff
@@ -69,69 +44,38 @@
...
-    """List the available pipelines"""
-    return pipeline_service.list_pipelines()
-
-
-@router.get(
-    "/{pipeline_id}",
-    responses={
-        status.HTTP_200_OK: {"description": "Pipeline found", "model": Pipeline},
+        status.HTTP_200_OK: {"description": "Pipeline found"},
         status.HTTP_400_BAD_REQUEST: {"description": "Invalid pipeline ID"},
```
leoll2: "Invalid pipeline ID" -> "Invalid project ID". Also in the other endpoints.

backend/tests/unit/services/test_pipeline_service.py
itallix: Yes, the fixture is required for `MetricsService` to work, but since we use `MagicMock` here, shm object is not needed.

backend/app/db/schema.py
itallix: This is planned for #4627

backend/app/db/schema.py
itallix: The declared lengths are advisory indeed. I will increase to 50.

backend/app/repositories/base.py
itallix: Before introducing dependent objects via `sqlalchemy.orm.relationship`, we didn't care about the actual state of the updated object (that was inferred from `merge` directly). Now we need to refresh explicitly, otherwise dependent objects will not be updated, e.g.: `PipelineDB(sink_id="new_id", sink=(id="old_id"))`.

In general, the new code is more correct, as it returns the actual, refreshed database object rather than the possibly detached or stale input object.

backend/app/api/endpoints/projects.py
itallix: Let's keep non-unique.

leoll2 (approved):

PR #4661, opened 2025-09-08
EditBounding box tool integration

jpggvilaca:
### Summary

First part of https://github.com/open-edge-platform/training_extensions/issues/4649

* Add AnnotatorProvider, AnnotationProvider & SelectionProvider
* Add EditBoundingBox for translation/resize functionality
* Cleanup/refactors (tidy up and move things to their own folders)

```
// Canvas (annotations.component.tsx)
<svg width={width} height={height} style={DEFAULT_ANNOTATION_STYLES} onClick={handleClickOutside}>
    <MaskAnnotations annotations={orderedAnnotations} width={width} height={height} isEnabled={isFocussed}>
          {annotations.map((annotation) => (
             <Annotation annotation={annotation} key={annotation.id} />
         )}
    </MaskAnnotations>
</svg>

export const Annotation = ({ annotation }: AnnotationProps) => {
    return (
        <AnnotationContext.Provider value={annotation}>
            <SelectableAnnotation> // Handles select behavior
                <EditableAnnotation> // Handles edit behavior (translate/resize)
                    <AnnotationShape annotation={annotation} /> // Draws the actual annotation
                </EditableAnnotation>
            </SelectableAnnotation>
        </AnnotationContext.Provider>
    );
};
```

https://github.com/user-attachments/assets/f2358339-8e6e-41c3-ae9e-46d16e6aa51a

### Checklist

### License

```python
```


ui/src/features/annotator/annotation-shape.tsx
```diff
@@ -14,19 +16,38 @@ const getFormattedPoints = (points: Point[]): string => {
 export const AnnotationShape = ({ annotation }: AnnotationShapeProps) => {
     const shape = annotation.shape;
     const color = annotation.labels[0].color;
+    const { setSelectedAnnotation, selectedAnnotation } = useAnnotator();
```
MarkRedeman: We should try to avoid adding behaviour on top of the `AnnotationShape` as it's only purpose should be rendering its shape. This way we can independantly use it for our drawing tools as well as render it in the media grid.

Instead we should compose behaviour by adding a `<SelectableAnnotation >` component that we talked about here: https://github.com/open-edge-platform/training_extensions/pull/4621#issuecomment-3242052575

ui/src/features/annotator/annotation-shape.tsx
```diff
@@ -14,19 +16,38 @@ const getFormattedPoints = (points: Point[]): string => {
...
+    const { scale } = useZoom();
+
+    const isSelected = selectedAnnotation?.id === annotation.id;
+
+    const selectedStyles = {
+        strokeWidth: 2 / scale,
```
MarkRedeman: The current zoom level should be exposed as a css variable (I believe the `ZoomProvider` does this)
That way we can avoid having to rely on any zoom hook here:

```suggestion
        strokeWidth: `calc(2 / var(--zoom-level))`
```

ui/src/features/annotator/annotation-shape.tsx
jpggvilaca: Totally agree. I opened this PR to get feedback on this because once this tool is integrated, all others will follow the same approach, so i want to get this as right as possible

ui/src/features/annotator/annotations.tsx
```diff
@@ -25,11 +24,16 @@ const DEFAULT_ANNOTATION_STYLES = {
     strokeOpacity: 'var(--annotation-border-opacity, 1)',
 } satisfies CSSProperties;
 
-export const Annotations = ({ annotations, width, height, isFocussed }: AnnotationsProps) => {
+export const Annotations = ({ width, height, isFocussed }: AnnotationsProps) => {
+    const { annotations, selectedAnnotation } = useAnnotator();
+
+    // Hide the ones being edited (resized or translated)
+    const staticAnnotations = annotations.filter((a) => a.id !== selectedAnnotation?.id);
```
MarkRedeman: The method (https://github.com/open-edge-platform/training_extensions/pull/4621#issuecomment-3242052575) we talked about in the previous tech talk should allow us to avoid having to do these type of filters.
What we can do instead is compose all annotations with multiple small components that can individually decide how annotations should be displayed. (so that for instance we no longer need to render the annotations again in the tool related component)

ui/src/features/annotator/mask-annotations.tsx
```diff
@@ -3,40 +3,22 @@
...
+export const MaskAnnotations = ({ children, width, height, isEnabled }: MaskAnnotationsProps) => {
     const id = useId();
+
     const maskOpacity = isEnabled ? 0.8 : 0.0;
 
     return (
         <>
             <mask id={`mask-${id}`}>
                 <rect x='0' y='0' width={width} height={height} style={{ fill: 'white', fillOpacity: 1.0 }} />
-                {annotations.map((annotation, idx) => (
```
MarkRedeman: Can you clarify this change?

jpggvilaca: The annotations were being rendered here and below on this file. I dont know why we needed this duplication? The visual effect is the same. Maybe i missed something? I made this change on a single commit. Perhaps @camiloHimura  will know more?

ui/src/features/annotator/annotations.tsx
jpggvilaca: agree!

ui/src/features/annotator/mask-annotations.tsx
MarkRedeman: It probably looks the same because you only had a single annotation, _and_ it was rendered on top of all other components.
This component does two things: 
1. It renders a black rectangle
2. It removes all the given annotations from that rectangle

What this allows us to do is highlight / focus certain annotations (for instance the selected annotations) by passing `<MaskAnnotations annotations={annotations.filter(isSelected} />`.
We could for instance combine this with a "hover" or "selected" state:
```tsx
import { ReactNode } from 'react';

import { Annotation } from './../types';
import { AnnotationShape } from './annotation-shape';
import { AnnotationContext } from './annotations';
import { useHoveredAnnotation } from './hovered';

export function MaskAnnotations({
    annotations,
    children,
    width,
    height,
}: {
    annotations: Array<Annotation>;
    children: ReactNode;
    width: number;
    height: number;
}) {
    const hoveredAnnotatoinId = useHoveredAnnotation();
    const isHovered = hoveredAnnotatoinId !== null;
    const maskOpacity = isHovered ? 0.8 : 0.0;

    return (
        <>
            <mask id={'mask'}>
                <rect x='0' y='0' width={width} height={height} style={{ fill: 'white', fillOpacity: 1.0 }} />
                {annotations.map((annotation) => (
                    <g
                        key={annotation.id}
                        style={{
                            fill: 'black',
                            fillOpacity: annotation.id === hoveredAnnotatoinId ? 1.0 : 0.0,
                            transitionProperty: 'fill-opacity',
                            transitionTimingFunction: 'ease-in-out',
                            transitionDuration: annotation.id === hoveredAnnotatoinId ? '0.2s' : '0.1s',
                            transitionDelay: annotation.id === hoveredAnnotatoinId ? '0s' : '.25s',
                        }}
                    >
                        <AnnotationContext.Provider value={annotation}>
                            <AnnotationShape />
                        </AnnotationContext.Provider>
                    </g>
                ))}
            </mask>
            <rect
                x={0}
                y={0}
                width={width}
                height={height}
                mask={`url(#mask)`}
                style={{
                    fillOpacity: maskOpacity,
                    fill: 'black',
                    strokeWidth: 0,
                    transition: 'fill-opacity 0.1s ease-in-out',
                    transitionDelay: isHovered ? '0s' : '.25s',
                    transitionDuration: isHovered ? '0.2s' : '0.1s',
                }}
            />
            <g>{children}</g>
        </>
    );
}
```

Where this hovered state can come from `hovered.tsx`,

```tsx
import { createContext, Dispatch, ReactNode, SetStateAction, useContext, useState } from 'react';

import { Annotation } from '../types';
import { useAnnotation } from './annotations';

const noop = () => {};
const HoveredAnnotation = createContext<string | null>(null);
const SetHoveredAnnotation = createContext<Dispatch<SetStateAction<string | null>>>(noop);

export function useSetHoveredAnnotation() {
    return useContext(SetHoveredAnnotation);
}

export function useHoveredAnnotation() {
    return useContext(HoveredAnnotation);
}

// Composes the annotation component
export function HoverableAnnotation({ children }: { children: ReactNode }) {
    const annotation = useAnnotation();
    const isHovered = useHoveredAnnotation() === annotation.id;
    const setHoveredAnnotation = useSetHoveredAnnotation();

    return (
        <g
            onPointerEnter={() => setHoveredAnnotation(annotation.id)}
            onPointerLeave={() => isHovered && setHoveredAnnotation(null)}
            style={{
                ...(isHovered ? { transitionDelay: '0' } : {}),
                transitionProperty: 'fill-opacity',
                transitionTimingFunction: 'ease-in-out',
                transitionDuration: '0.1s',
                transitionDelay: isHovered ? '0s' : '0.25s',

                fill: isHovered ? 'yellow' : undefined,
            }}
        >
            {children}
        </g>
    );
}

// Render along with the `AnnotatorProvider`
export function HoveredProvider({ children }: { children: ReactNode }) {
    const [hoveredAnnotation, setHoveredAnnotation] = useState<string | null>(null);

    return (
        <HoveredAnnotation.Provider value={hoveredAnnotation}>
            <SetHoveredAnnotation.Provider value={setHoveredAnnotation}>{children}</SetHoveredAnnotation.Provider>
        </HoveredAnnotation.Provider>
    );
}

```

ui/src/features/annotator/annotation-shape.tsx
jpggvilaca: Updated

ui/src/features/annotator/hooks/use-translate.hook.ts
```diff
@@ -0,0 +1,62 @@
...
+    const onPointerDown = (event: PointerEvent<SVGSVGElement>): void => {
+        if (dragFromPoint.current !== null) {
+            return;
+        }
+
+        if (event.pointerType === 'touch' || !isLeftButton(event)) {
+            return;
+        }
+
+        const mouse = { x: Math.round(event.clientX / zoom), y: Math.round(event.clientY / zoom) };
```
MarkRedeman, line 26: Perhaps we can extract this to a `getMousePosition(event. zoomState): Point` function since we will need to do this for all tools.
It might also need to take into account `zoomState.translate` for the other tools. (here it's not necessary as when moving the shape we only care about delta x.

Related: I'm wondering if we can use a different property like `offsetX` instead of `clientX` here. I believe we tested this before in Geti but ended up having to compute this value based on `clientX` due to not all browsers having the same behaviour. Perhaps that's fixed now?

ui/src/features/annotator/annotator-provider.component.tsx
```diff
@@ -0,0 +1,83 @@
...
+}) => {
+    const { activeTool, selectTool } = useActiveTool();
+    const [selectedAnnotation, setSelectedAnnotation] = useState<Annotation | null>(null);
+    const [annotations, setAnnotations] = useState<Annotation[]>(initialAnnotations);
+
+    const handleSelectTool = (tool: ToolType) => {
+        selectTool(tool);
+    };
+
+    const handleSelectAnnotation = (newAnnotation: Annotation) => {
```
MarkRedeman: If we don't have any logic inside of the current provider that depends on the selection state then I'd move all the selection related state into its own provider. I.e.
```tsx
type SelectedAnnotations = Set<string>;

const SelectedAnnotations = createContext<SelectedAnnotations | null>(null);
const SetSelectedAnnotation = createContext<Dispatch<SetStateAction<SelectedAnnotations | null>> | null>(null);

export function SelectedProvider({ children }: { children: ReactNode }) {
    const [selectedAnnotation, setSelectedAnnotation] = useState<SelectedAnnotations | null>(new Set());

    return (
        <SelectedAnnotation.Provider value={selectedAnnotation}>
            <SetSelectedAnnotation.Provider value={setSelectedAnnotation}>{children}</SetSelectedAnnotation.Provider>
        </SelectedAnnotation.Provider>
    );
}

export function useSetSelectedAnnotations() {
    const ctx =  useContext(SetSelectedAnnotation);

    if (ctx === null) {
        throw new Error('No contex');
    }

    return ctx;
}

export function useSelectedAnnotation() {
    const ctx = useContext(SelectedAnnotation);

    if (ctx === null) {
        throw new Error('No contex');
    }

    // note: we could also do something like `return annotations.filter(({ id }) => ctx.has(id));` so that the suer gets annotations and not a set
    return ctx;
}

```

which can then be rendered by the annotator provider.
The motivation for this is that it scales better wrt performance: we can use the specific `useSelectedAnnotations` independantly from `useSetSelectedAnnotations`.

ui/src/features/dataset/gallery/gallery.component.tsx
```diff
@@ -85,7 +86,9 @@ export const Gallery = () => {
 
             <DialogContainer onDismiss={() => setSelectedMediaItem(null)}>
                 {selectedMediaItem !== null && (
-                    <MediaPreview mediaItem={selectedMediaItem} close={() => setSelectedMediaItem(null)} />
+                    <AnnotatorProvider initialAnnotations={selectedMediaItem.annotations}>
+                        <MediaPreview mediaItem={selectedMediaItem} close={() => setSelectedMediaItem(null)} />
```
MarkRedeman: I wonder if we should put the `AnnotatorProvider` inside of `MediaPreview` instead of rendering it here. It already has access to `selectedMediaItem.annotations` via its `mediaItem` so it might be more convenient if we render it there?

jpggvilaca: i dont see any inconvenient either way, Ill move it

ui/src/features/annotator/hooks/use-translate.hook.ts
jpggvilaca: Im already doing way more than i intended to in this PR. There is also drawing box + anchor + others tools. And you also `translate` on editBoundingbox. I intend on refactor/extract this behavior, but for now this PR is too convoluted so i'd rather do it on next PRs

ui/src/features/annotator/annotator-provider.component.tsx
jpggvilaca: Was planning on doing it as well, but on the next PR if that's ok.

ui/src/features/annotator/tools/bounding-box-tool/bounding-box-tool.component.tsx
```diff
@@ -44,8 +47,8 @@ export const EditBoundingBox = ({ annotation, roi, image, zoom, updateAnnotation
     return (
         <>
             <svg
-                width={image.width}
-                height={image.height}
+                width={roi.width}
+                height={roi.height}
```
MarkRedeman: I'd prefer to use `image` here and remove `roi` from the component's props as we will no longer need to deal with an `roi` in Geti Tune.

ui/src/features/annotator/tools/bounding-box-tool/translate-shape.component.tsx
```diff
@@ -16,64 +15,24 @@ interface TranslateShapeProps {
...
+    const { onPointerDown, onPointerMove, onPointerUp } = useTranslate({
+        zoom,
+        onTranslate: translateShape,
+        onComplete,
+    });
 
     return (
         <g
-            id={`translate-annotation-${annotation.id}`}
+            id={`translate-annotation-${annotation}`}
```
MarkRedeman: Why this change? (this could result in a very messy id since the annotation shape and labels won't result in a clean string)

ui/src/components/tool-selection-bar/icon-wrapper.component.tsx
```diff
@@ -0,0 +1,24 @@
+// Copyright (C) 2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+import { ReactNode } from 'react';
+
+import cls from 'clsx';
```
dwesolow: np. typo, `clsx`

ui/src/components/tool-selection-bar/zoom/zoom-selector.component.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { IconWrapper } from '../icon-wrapper.component';
+
+export const ZoomSelector = () => {
+    return (
+        <>
+            <IconWrapper>
+                <Add />
+            </IconWrapper>
+
+            <Flex>110%</Flex>
```
dwesolow, line 16: leftover?

jpggvilaca: No, just a placeholder. Zoom functionality (buttons) will be added later

ui/src/features/annotator/annotations/selectable-annotation.component.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { useSelectedAnnotations, useSetSelectedAnnotations } from '../select-annotation-provider.component';
+import { useAnnotation } from './annotation.component';
+
+export const SelectableAnnotation = ({ children }: { children: ReactNode }) => {
+    const annotation = useAnnotation();
+    const isSelected = useSelectedAnnotations().has(annotation.id);
+    const setSelectedAnnotations = useSetSelectedAnnotations();
+    const { activeTool } = useAnnotator();
+
+    const hasSelectionToolActive = activeTool === 'selection';
```
dwesolow: Do we need to render this component if the `activeTool !== 'selection'`?

jpggvilaca: Yes, this check is just to avoid misclicking on an annotation if you have any other tool selected. This will definitely change once i added all other tools (Drawing box, Polygon and SAM)

ui/src/features/annotator/select-annotation-provider.component.tsx
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    return ctx;
+};
+
+export const SelectAnnotationProvider = ({ children }: { children: ReactNode }) => {
+    const [selectedAnnotations, setSelectedAnnotations] = useState<SelectedAnnotations | null>(new Set());
+
+    return (
+        <SelectedAnnotation.Provider value={selectedAnnotations}>
```
dwesolow: Isn't it premature optimization?

ui/src/features/annotator/tools/bounding-box-tool/bounding-box-tool.component.tsx
```diff
@@ -13,16 +13,19 @@ import classes from './bounding-box-tool.module.scss';
...
 const ANCHOR_SIZE = 8;
 
-export const EditBoundingBox = ({ annotation, roi, image, zoom, updateAnnotation }: EditBoundingBoxProps) => {
+export const EditBoundingBox = ({ annotation, roi, zoom, updateAnnotation }: EditBoundingBoxProps) => {
     const [shape, setShape] = useState(annotation.shape);
 
+    useEffect(() => {
+        setShape(annotation.shape);
+    }, [annotation]);
+
```
dwesolow, line 25: Do u remember my last comment how we can replace `useEffect` with `prevShape` etc?

ui/src/features/annotator/tools/selection-tool/selection-tool.component.tsx
```diff
@@ -0,0 +1,22 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import { AnnotationShape } from '../../annotations/annotation-shape.component';
+import { Annotation } from '../../types';
+
+type SelectionToolProps = {
+    annotation: Annotation & { shape: { shapeType: 'rect' } };
+};
+
+export const SelectionTool = ({ annotation }: SelectionToolProps) => {
```
dwesolow: I see it's not used anywhere. What will be the intention of this component?

ui/src/features/annotator/annotations/annotation.component.tsx
```diff
@@ -0,0 +1,32 @@
...
+    return ctx!;
+};
+
+interface AnnotationProps {
+    annotation: AnnotationInterface;
+}
+export const Annotation = ({ annotation }: AnnotationProps) => {
+    return (
+        <AnnotationContext.Provider value={annotation}>
+            <EditAnnotation>
```
dwesolow: I see that `EditAnnotation` also checks if the annotation is selected and conditionally render either `children` or `AnnotationShape`. It means that when it does not render `children` then `SelectableAnnotation` is not rendered. Is it intended behavior?

ui/src/features/annotator/select-annotation-provider.component.tsx
jpggvilaca: Used most of the code from Mark's branch from the Geti talks presentation. I do agree it might be premature opt

ui/src/features/annotator/tools/bounding-box-tool/bounding-box-tool.component.tsx
jpggvilaca: reminder needed ^^

ui/src/features/annotator/tools/selection-tool/selection-tool.component.tsx
jpggvilaca: I am still unsure. Now it is indeed not being used but i believe i will need it to handle all selection logic

ui/src/features/annotator/annotations/annotation.component.tsx
jpggvilaca: Removed EditAnnotation for now.

ui/src/features/annotator/tools/selection-tool/selection-tool.component.tsx
jpggvilaca: Removed

ui/src/features/annotator/tools/bounding-box-tool/bounding-box-tool.component.tsx
jpggvilaca: Updated

ui/src/components/tool-selection-bar/icon-wrapper.component.tsx
jpggvilaca: updated

dwesolow (approved):

dwesolow (dismissed):

ui/src/features/annotator/annotations/annotation-shape.component.test.tsx
```diff
@@ -4,17 +4,25 @@
...
 type AnnotationPolygon = Annotation & { shape: Polygon };
 
+const mockMediaItem = response.items[0];
+
 describe('AnnotationShape', () => {
     it('bounding box as a rect', () => {
         const annotation = getMockedAnnotation() as AnnotationRect;
-        render(<AnnotationShape annotation={annotation} />);
+        render(
+            <AnnotatorProvider mediaItem={mockMediaItem}>
```
MarkRedeman: np: `AnnotationShape` does not rely on `AnnotatorProvider` anymore, so it can be removed here

ui/src/features/annotator/hooks/use-active-tool.hook.ts
```diff
@@ -0,0 +1,16 @@
...
+import { ToolType } from '../../../components/tool-selection-bar/tools/interface';
+
+export const useActiveTool = () => {
+    const [activeTool, setActiveTool] = useState<ToolType>('selection');
+
+    const handleSelectTool = (tool: ToolType) => {
+        setActiveTool(tool);
+    };
+
+    return { activeTool, selectTool: handleSelectTool };
```
MarkRedeman: np:

```suggestion
    const [activeTool, setActiveTool] = useState<ToolType>('selection');

    return { activeTool, selectTool: setActiveTool };
```

(perhaps rename `selectTool` to `setActiveTool`? I don't see a huge benefit atm of abstracting this)

ui/src/features/annotator/tools/bounding-box-tool/bounding-box-tool.component.tsx
```diff
@@ -43,23 +43,18 @@ export const EditBoundingBox = ({ annotation, roi, image, zoom, updateAnnotation
...
-                    onComplete={onComplete}
-                />
-            </svg>
+                <AnnotationShape annotation={{ ...annotation, shape }} />
+            </TranslateShape>
 
             <svg
-                width={image.width}
-                height={image.height}
+                width={roi.width}
```
MarkRedeman: From my earlier testing: I believe we can remove this `svg`, it didn't seem to contribute to anything

ui/src/features/annotator/annotator-provider.component.tsx
```diff
@@ -0,0 +1,65 @@
...
+        <AnnotatorProviderContext.Provider
+            value={{
+                activeTool,
+                selectTool: handleSelectTool,
+
+                updateAnnotation: handleUpdateAnnotation,
+
+                annotations,
+
+                roi,
```
MarkRedeman: Perhaps instead of exposing only `roi` we could expose `mediaItem`.

MarkRedeman (dismissed):
Some nit picks and questions, but I'm fine with merging as is.

ui/src/features/annotator/annotations/annotation-shape.component.test.tsx
jpggvilaca: Updated

ui/src/features/annotator/hooks/use-active-tool.hook.ts
jpggvilaca: Updated

ui/src/features/annotator/tools/bounding-box-tool/bounding-box-tool.component.tsx
jpggvilaca: Updated

ui/src/features/annotator/annotator-provider.component.tsx
jpggvilaca: Updated

MarkRedeman (approved):

PR #4668, opened 2025-09-09
simplify workers shutdown

leoll2:
### Summary

Changes:
- Remove unnecessary logic to flush queue in the workers
  - Queues are eventually flushed by the scheduler, and the workers can still exit thanks to `cancel_join_thread`
- Add warning when a thread does not terminate properly within timeout
- Fix the `log_threads` method - previous implementation attempted to list the threads of other processes, which is not possible

### How to test

Run the application as usual and shutdown it with ctrl+C (keyboard interrupt). Repeat multiple times.


itallix (dismissed):
LGTM!

itallix:
For failing unit test in `TestFrameAcquisition`, we can simply remove it as we do not expect cleanup to empty the queue anymore.

itallix (approved):

PR #4678, opened 2025-09-10
Implement endpoint to edit labels

### LINKED ISSUE #4627: Implement endpoint to edit labels

itallix:
### Summary

This PR implements the `PATCH` endpoint `/api/projects/<id>/labels` for updating project labels. The endpoint allows clients to:

- Edit existing labels (change name, color, or hotkey)
- Add new labels
- Remove labels by ID

Returns the updated list of labels for project.

**Validation & Atomicity:**
- The request is validated for feasibility before any changes are made.
- If any label to remove or edit does not exist in the project, the request is rejected with a 404 Not Found error and no changes are committed.
- All operations (add, edit, remove) are performed atomically—either all succeed or none are applied.

### How to test

1. Remove the old database file (if it exists).
2. Run the server in seed mode:
```bash
SEED_DB=true ./run.sh
```
3. Fetch existing labels (two options):
```bash
# As part of the project details:
curl localhost:7860/api/projects/9d6af8e8-6017-4ebe-9126-33aae739c5f
# Get just the labels (PATCH with empty body returns current labels)
curl -X PATCH -H "Content-Type: application/json"  localhost:7860/api/projects/9d6af8e8-6017-4ebe-9126-33aae739c5fa/labels -d '{}' # or PATCH with empty body
```
4. Test label update:
```bash
# Add a label
curl -X PATCH -H "Content-Type: application/json"  localhost:7860/api/projects/9d6af8e8-6017-4ebe-9126-33aae739c5fa/labels -d '{ "labels_to_add": [ {"name":"animal","color":"#0000FF"}] }'
# Edit existing label
curl -X PATCH -H "Content-Type: application/json"  localhost:7860/api/projects/9d6af8e8-6017-4ebe-9126-33aae739c5fa/labels -d '{ "labels_to_edit": [ {"id":"84dfad1a-e7f5-4aba-b08a-b2ec6bcce543", "new_name":"car"}] }'
# Remove existing label
curl -X PATCH -H "Content-Type: application/json"  localhost:7860/api/projects/9d6af8e8-6017-4ebe-9126-33aae739c5fa/labels -d '{ "labels_to_remove": [ {"id":"84dfad1a-e7f5-4aba-b08a-b2ec6bcce543" }] }'
```

### Checklist

### License

```python
```


backend/app/db/schema.py
```diff
@@ -99,3 +99,22 @@ class DatasetItemDB(Base):
     subset_assigned_at: Mapped[datetime | None] = mapped_column(DateTime, nullable=True)
     created_at: Mapped[datetime] = mapped_column(DateTime, server_default=func.current_timestamp())
     updated_at: Mapped[datetime] = mapped_column(DateTime, server_default=func.current_timestamp())
+
+
+class LabelDB(Base):
+    __tablename__ = "labels"
+    __table_args__ = (
+        UniqueConstraint("project_id", "name", name="uq_project_label_name"),
+        UniqueConstraint("project_id", "color", name="uq_project_label_color"),
```
leoll2: The unique constraint on the label color is arguably unnecessary. For name and hotkey it makes sense.

backend/app/repositories/label_repo.py
```diff
@@ -0,0 +1,25 @@
...
+from app.repositories.base import BaseRepository
+
+
+class LabelRepository(BaseRepository[LabelDB]):
+    """Repository for label-related database operations."""
+
+    def __init__(self, db: Session):
+        super().__init__(db, LabelDB)
+
+    def list_by_project_id(self, project_id: str) -> list[LabelDB]:
```
leoll2: What do you think about passing `project_id` as argument to the repo constructor? Similar to Geti project-based repos, where you don't have to pass `project_id` to all methods. Then the method could be renamed `get_all()`, since the scope is limited to a given project.

backend/app/services/label_service.py
```diff
@@ -0,0 +1,99 @@
...
+        """
+        Add, update, and remove labels in a project transactionally.
+
+        This method performs label operations in a specific, deterministic order:
+        1. First removes labels (by their IDs)
+        2. Then updates existing labels
+        3. Finally adds new labels
+
+        This order ensures data integrity and prevents conflicts when operations
+        might affect the same labels in different ways.
```
leoll2: What happens if the client asks to edit and rename the same label? If the removal is executed first, does the second operation (edit) fail?

backend/app/db/schema.py
itallix: Does it make sense to render different classes with the same color?

backend/app/services/label_service.py
itallix: Missed test for that.
Currently:
```
sqlalchemy.orm.exc.StaleDataError: UPDATE statement on table 'labels' expected to update 1 row(s); 0 were matched.
```
What would be the desired behavior? We can run updates first, and then deletes, so we don't need to deal with it as an error case or prevent it even earlier during request payload validation (to prevent users from specifying multiple operations (edit and remove) for the same label in a single request).

backend/app/repositories/label_repo.py
itallix: It's a good idea!

backend/app/services/label_service.py
itallix: I don't think it makes sense to allow this operation. I suggest tightening the payload validation.

backend/app/db/schema.py
leoll2: Using the same color for two labels is probably not useful, but it doesn't break any feature. We shouldn't be too strict in this case. I imagine two situations where this unique constraint could be annoying:
- a user writes a script to automate project creation, and for convenience assigns color `#ffffff` to all labels -> request fails
- a user creates a new label without specifying color -> server must pick a random color while avoiding collisions with existing labels -> complex logic

The last scenario actually applies to hotkeys too: there can be two labels without hotkey, i.e. with the same `null` value: would that violate the unique constraint?

itallix: > The last scenario actually applies to hotkeys too: there can be two labels without hotkey, i.e. with the same null value: would that violate the unique constraint?

No issues with that. Database unique constraints treat NULL values as distinct, so multiple rows with NULL in the hotkey or color columns will not violate the unique constraint.

backend/app/services/label_service.py
leoll2: Yeah, I think it's safer to run the operations in the following order: edit, remove, add. It also prevents errors in the rare scenario where you add a new label with the same name of a removed one. I don't think we need extra validation - these are corner cases that probably never happen in real projects, so we can leave the behavior unspecified and implement it in the most simple and convenient way, as long as the server doesn't crash (5xx).

leoll2 (dismissed):

leoll2 (approved):

PR #4683, opened 2025-09-11
Refactor workers package to use class-based inheritance

### LINKED ISSUE #4667: Refactor workers package to use class inheritance-based design
### Background

Currently, the `workers` package uses standalone functions and ad-hoc functions from `utils` package to manage both process-based and thread-based workers. This design is not ideal as it results in repeated patterns, making the code harder to maintain and extend.

### What to Do

- **Introduce Base Classes:**
    Create a `BaseWorker` class (and, if needed, `BaseProcessWorker` and `BaseThreadWorker` subclasses) that encapsulate shared logic for worker lifecycle, signal handling, queue management, and diagnostics.
- **Refactor Existing Workers:**  
    Refactor functions into corresponding classes (e.g. `StreamLoader`, `Inferencer`, `Dispatcher`) and inherit from new base classes. Process/thread creation semantics in `Scheduler` will change as well, e.g.:
```python
   stream_loader_proc = StreamLoader(self.frame_queue, self.mp_stop_event, self.mp_config_changed_condition)
```
- **Centralize Utilities:**  
    Move process/thread worker utilities (e.g., signal suppression, queue flushing, thread logging) from `utils` into the relevant base class(es).
- **Preserve Functionality:**  
    Ensure all existing features and tests continue to work as before.

### Why

- **Maintainability:**  Easier to update repeated worker behavior.
- **Extensibility:**  New worker types can be added by subclassing.
- **Clarity:**  OO structure makes intent and flow clearer.

### Acceptance Criteria

- All worker logic is encapsulated in classes using inheritance.
- No duplicated logic between process and thread workers.
- All existing functionality and tests pass.

itallix:
- [x] New base classes `BaseProcessWorker` and `BaseThreadWorker` with clean lifecycle: `setup` -> `run_loop` -> `teardown`
- [x] Replaced routine functions with classes
- [x] All sleep/wait paths are stop-aware
- [x] Producer-side queue `cancel_join_thread()` is handled centrally by `BaseProcessWorker` when declared
- [x] Code review has been done


backend/app/workers/dispatching.py
```diff
@@ -5,47 +5,59 @@
...
+        self._active_pipeline_service: ActivePipelineService | None = None
+        self._prev_sink_config: Sink | None = None
+        self._destinations: list[Dispatcher] = []
+
+    def setup(self) -> None:
+        from app.api.dependencies import get_active_pipeline_service  # Avoid circular import
+
+        self._active_pipeline_service = get_active_pipeline_service()
+
+    def _reset_sink_if_needed(self, sink_config: Sink) -> None:
```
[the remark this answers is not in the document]
itallix: False alarm, same below

backend/app/core/scheduler.py
```diff
@@ -50,30 +50,18 @@ def start_workers(self) -> None:
...
-            args=(self.frame_queue, self.mp_stop_event, self.mp_config_changed_condition),
+        stream_loader_proc = StreamLoader(self.frame_queue, self.mp_stop_event, self.mp_config_changed_condition)
+
+        inference_server_proc = InferenceWorker(
+            self.frame_queue,
+            self.pred_queue,
+            self.mp_stop_event,
+            self.mp_model_reload_event,
+            self.shm_metrics.name,
+            self.shm_metrics_lock,
```
leoll2: I suggest to pass the arguments by keyword: the list is quite long and some parameters have the same type too, so it's easy to accidentally swap them.

backend/app/workers/base.py
```diff
@@ -0,0 +1,164 @@
...
+    thread_list_msg = (
+        f"Alive threads for process with pid '{pid}': "
+        f"{', '.join([str((thread.name, thread.ident)) for thread in alive_threads])}"
+    )
+    logger.log(level=log_level, msg=thread_list_msg)
+
+
+class StopMixin:
+    def should_stop(self) -> bool:
+        raise NotImplementedError
```
leoll2: Can you please add a docstring to this class?

backend/app/workers/dispatching.py
```diff
@@ -5,47 +5,59 @@
...
+            self._prev_sink_config = copy.deepcopy(sink_config)
+
+    def run_loop(self) -> None:
+        while not self.should_stop():
+            sink_config = self._active_pipeline_service.get_sink_config()  # type: ignore
 
             if sink_config.sink_type == SinkType.DISCONNECTED:
                 logger.debug("No sink available... retrying in 1 second")
-                time.sleep(1)
+                self.stop_aware_sleep(1)
```
leoll2, line 53: The new implementation with `stop_aware_sleep` is slightly more responsive on shutdown, since it sleeps only 0.1 instead of 1s before checking the signal. However, I'm concerned about the context switching overhead, now that the thread has to go sleep and wake up 10 times per second. Probably it's not that bad, but is it worth it? Did the old `sleep(1)` cause any significant problem?

leoll2 (commented):
A few minor comments, very nice improvement. The code of the workers is now much cleaner.

backend/app/workers/dispatching.py
itallix: That's a good point. While the old sleep approach wasn't causing any issues, we ideally want to avoid unnecessary delays where possible. If we need to increase the current timeout (2sec) for exception handling, those extra seconds could become noticeable during shutdown.
I found that the Event object has a built-in [wait()](https://docs.python.org/3/library/threading.html#threading.Event.wait) method with timeout support that works exactly as we need. Unlike regular sleep, it's implemented using conditions that allow it to wake up immediately when the stop signal is received.

leoll2 (approved):

PR #4688, opened 2025-09-15
Geti Tune model management and training docs

### LINKED ISSUE #4563: Design for fine-tuning and training dataset management
Design the API to fine-tune models, view model information and browse training data.

leoll2:
### Summary

Added new document to describe how models are managed in Geti Tune.
- [x] High-level overview of the model lifecycle
- [x] Explanation of how models are versioned and linked to each other
- [x] Detailed description of the model storage (DB, weights, manifests, training artifacts)
- [x] Breakdown of the training and evaluation job
- [ ] ~Overview of the job scheduling framework~ (will be addressed separately)
- [x] Model deletion workflow


application/docs/models.md
```diff
@@ -0,0 +1,207 @@
...
+   available for the project.
+
+2. **Selection of architecture, training configuration, and base model**: Before training, the user chooses a suitable
+   model architecture (such as YOLO, EfficientNet) and configures training parameters (e.g., learning rate, augmentations).
+   The user also selects a base model for fine-tuning, which can be a public pre-trained model or an existing model
+   revision from the same project.
+
+3. **Training**: The selected model is fine-tuned using OTX (OpenVINO Training Extensions), the open-source framework
+   which implements the actual model and relative training pipeline in PyTorch. During training, the model learns from
+   the dataset annotations to to predict the specific classes of the project. The training process produces a bunch of
```
itallix:
```suggestion
   the dataset annotations to predict the specific classes of the project. The training process produces a bunch of
```

application/docs/models.md
```diff
@@ -0,0 +1,207 @@
...
+Training is a lengthy process that involves multiple epochs of forward and backward passes, optimization steps,
+and periodic evaluation on the validation set. All these operations are handled internally by OTX, which works
+seamlessly with different devices including CPU and GPUs (Intel and Nvidia).
+Throughout the training, various metrics (e.g. accuracy and loss) are logged to a JSON file for later analysis.
+After training, the resulting PyTorch weights are converted to OpenVINO IR format using the
+[OpenVINO model converter](https://docs.openvino.ai/2024/notebooks/convert-to-openvino-with-output.html).
+Finally, the training job updates the model revision record in the database with the training status, timestamps,
+and other relevant metadata.
+
+The final step is the evaluation of the trained model on the testing subset, using standard metrics such as accuracy
```
itallix: I think we are lacking additional context here since the term "standard metrics" is vague and only mentions accuracy for classification tasks. This can be improved with:

- Specific metrics used for different task types
- A table showing which metrics are computed for each task type

application/docs/models.md
```diff
@@ -0,0 +1,207 @@
...
+pipeline. Implementation-wise, the evaluation step relies on ModelAPI to generate the predictions which are then
+compared to the ground truth annotations to compute the metrics.
+
+![Training data flow](media/training-data-flow.jpg)
+
+> [!NOTE]
+> For training purposes, the Datumaro `Dataset` can be kept in-memory and consumed as is by OTX.
+> However, persisting it to disk is still necessary if, at a later time, the user wants to re-train another model
+> on the same dataset revision, or to simply inspect the dataset contents. For this reason, it is recommended to
+> store the dataset in a format like Parquet that allows efficient loading and querying, with or without Datumaro.
```
itallix, line 193: This might be an interesting alternative to Parquet: https://github.com/lancedb/lance

[LanceDB](https://github.com/lancedb/lancedb) which is built on top of this format has quite powerful features for dataset exploration. Could be worth exploring for the local setting.

application/docs/models.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Model deletion
+
+Model artifacts, especially the weights, can consume a significant amount of disk space.
+To manage storage effectively, Geti Tune allows users to delete model revisions or just their weights when they
+are no longer needed. Deleting a model revision removes its metadata from the database and all associated files
+from the filesystem, while deleting just the weights retains the metadata for tracking purposes but frees up
+the disk space used by the weight files.
+After removing the weights, it is no longer possible to use the model for inference or as a base for further training.
```
itallix, line 214: What are the benefits of having model revision metadata without weights from the user perspective?

application/docs/models.md
leoll2: Good point, I have added a small section with a table containing a list of the metrics for each task.

application/docs/models.md
leoll2: Thanks for the suggestion, I will definitely look into it

application/docs/models.md
leoll2: Even without the weights, the metadata is still useful for various purposes:

- Reconstruct the exact lineage of a model, i.e. the sequence of fine-tuned revisions that led to that model.
- View the training statistics / logs. This can be useful after upgrading to a new OTX release, to assess if the performance has improved or not, or to spot regressions.
- Explore the training dataset (assuming its files are not deleted)

itallix (approved):

application/docs/api.md
```diff
@@ -28,22 +28,26 @@
...
+| Method   | Path                                        | Payload            | Return           | Description                       |
+| -------- | ------------------------------------------- | ------------------ | ---------------- | --------------------------------- |
+| `POST`   | `/api/projects`                             | name, task, labels | project info     | Create a new project              |
+| `GET`    | `/api/projects`                             | -                  | list of projects | List the available projects       |
+| `GET`    | `/api/projects/<id>`                        | -                  | project info     | Get info about a project          |
+| `PATCH`  | `/api/projects/<id>`                        | name               | project info     | Rename a project                  |
+| `DELETE` | `/api/projects/<id>`                        | -                  | -                | Delete a project                  |
+| `PATCH`  | `/api/projects/<id>/labels`                 | labels to change   | task and labels  | Add, remove or edit labels        |
```
jpggvilaca, line 38: do we need to return `task` here?

application/docs/models.md
```diff
@@ -0,0 +1,220 @@
...
+4. **Model conversion**: After training, the model weights are converted from PyTorch format to OpenVINO Intermediate
+   Representation (IR) format, for efficient inference on Intel hardware.
+
+5. **Evaluation**: The converted model is evaluated on validation data to assess its accuracy and performance.
+   Evaluation results help determine whether the model is ready for deployment or requires further fine-tuning.
+
+6. **Enablement in the inference pipeline**: Once validated, the model revision can be manually enabled in the project's
+   inference pipeline. The selected model is then used to generate real-time predictions on the incoming data stream.
+
+7. **Model or weight removal**: Eventually, after the model is no longer needed or a more recent and accurate revision
```
jpggvilaca, line 57: could you elaborate on this one? Did we have this option in Geti to remove/add weights?

application/docs/models.md
```diff
@@ -0,0 +1,220 @@
...
+### Evaluation metrics
+
+The following table lists the evaluation metrics supported by Geti Tune for each task type.
+Most of the metrics commonly used in research and industry are included.
+
+| Task Type                 | Metrics                                                   |
+| ------------------------- | --------------------------------------------------------- |
+| Multiclass Classification | Accuracy, Precision, Recall, F1                           |
+| Multilabel Classification | Subset Accuracy, Hamming Loss, Precision, Recall, F1, mAP |
+| Object Detection          | mAP@0.5-0.95 (and single mAP at the various thresholds)   |
```
jpggvilaca, line 204: is this instance, semantic, or neither?

jpggvilaca (approved):

application/docs/api.md
leoll2: Actually not, good catch. The current implementation only returns a list of labels; I'll fix this line in my next PR.

application/docs/models.md
leoll2: Yes, it is called _model archival_ in Geti.

application/docs/models.md
leoll2: The segmentation task is _instance segmentation_. For object detection there is no distinction between _instance_ or _semantic_, in fact it is always instance.

PR #4697, opened 2025-09-15
Integrate /projects api

jpggvilaca:
### Summary

* Add CreateProjectForm
* Tidy up (move stuff into their own folders, renamed model to task)
* Update interfaces/tests
* Removed labels route (deprecated)

### Checklist

### License

```python
```


ui/src/features/project/project-provider.component.tsx
```diff
@@ -0,0 +1,77 @@
...
+    name: string;
+};
+
+const ProjectContext = createContext<ProjectContextType | null>(null);
+
+type ProjectProviderProps = {
+    children: ReactNode;
+};
+
+export const ProjectProvider = ({ children }: ProjectProviderProps) => {
```
MarkRedeman: I'd rename this to `NewProjectFormProvider` or something similar to emphasize that this is used to create new projects.

ui/src/features/project/project-provider.component.tsx
```diff
@@ -0,0 +1,77 @@
...
+    const [selectedTask, setSelectedTask] = useState<TaskType>('detection');
+    const [labels, setLabels] = useState<{ name: string }[]>([]);
+    const [name, setName] = useState<string>('Project #1');
+
+    const projectId = uuid();
```
MarkRedeman: This would become a new id on every render, since it's only used inside of the mutation perhaps you can move it into its body?

ui/src/features/project/label-selection/label-selection.component.test.tsx
```diff
@@ -4,26 +4,32 @@
...
-const renderLabelSelection = () => {
-    return render(
-        <>
-            <LabelSelection />
-            <Toast />
-        </>
+const App = () => {
+    return (
+        <TestProviders>
+            <ProjectProvider>
```
MarkRedeman: The testing library library recommends to use custom renders for this kind of setup (https://testing-library.com/docs/react-testing-library/setup). Could you look into that?

One specific advantage of this is that in the testing code you can still do `<LabelSelection prop1={prop1} />`, i.e. pass the props per test without having to change the providers.

ui/src/features/project/label-selection/label-selection.component.tsx
```diff
@@ -88,6 +90,11 @@ export const LabelSelection = () => {
     const [items, setItems] = useState<Omit<LabelItemProps, 'onDelete'>[]>([
         { id: uuid(), colorValue: '#F20004', nameValue: 'Car' },
     ]);
```
MarkRedeman: Why not use the label state from the project context directly?

ui/src/features/project/project-name.tsx
```diff
@@ -18,15 +16,17 @@ import {
 } from '@geti/ui';
 import { Edit } from '@geti/ui/icons';
 
+import { useProject } from './project-provider.component';
```
MarkRedeman: Perhaps try to make this component not rely on the project provider. If we do rely on it then we must make sure to always render its provider before tendering the `ProjectName` component.

I think it would be fine to use `$api.useMutation` directly here if all we want to do is update the name. (vs the provider that also allows to update the labels as well as its tasks (which should only be possible when creating a new project)

ui/src/features/project/project-provider.component.tsx
```diff
@@ -0,0 +1,77 @@
...
+    createProject: ({ onSuccess }: { onSuccess?: () => void }) => void;
+    setSelectedTask: Dispatch<SetStateAction<TaskType>>;
+    setLabels: Dispatch<SetStateAction<{ name: string }[]>>;
+    setName: Dispatch<SetStateAction<string>>;
+    selectedTask: TaskType;
+    labels: { name: string }[];
+    name: string;
+};
+
+const ProjectContext = createContext<ProjectContextType | null>(null);
```
dwesolow: Why do we need context for creating project? Would not it be enough to have just state in the parent and pass it down the tree (asking because I see only `CreateProject` component uses it)

ui/src/features/project/project-provider.component.tsx
jpggvilaca: It was being used by the navigate, to add the id to the path, but that is a leftover from previous implementation

ui/src/features/project/project-provider.component.tsx
jpggvilaca: True, and being kind of a form, i was torn between a provider or just a component. I think i'll revert and just use a form component

ui/src/features/project/label-selection/label-selection.component.tsx
jpggvilaca: Because the local state is quite complicated, but i think i'll revert to a single form component + pass item/setItem props to the children

ui/src/features/project/project-name.tsx
jpggvilaca: It's a tradeoff right? I wanted to keep everything tidy in its own place, with the tradeoff of adding this dependency. If i move $api here then the reusability decreases a bit, but in the end i dont think it's better or worse

ui/src/features/project/project-provider.component.tsx
jpggvilaca: Commented above, i was torn between that. Between having
```
<Provider>
    <Child1>
    <Child2>
</Provider

vs

const [value, setValue] = useState(...);
const [value2, setValue2] = useState(...);


<Form>
    <Child1 value={value} setValue={setValue} />
     <Child1 value={value2} setValue={setValue2} />
```

So i think i'll go with the second approach. Any objections?

ui/src/features/project/project-name.tsx
MarkRedeman: > If i move $api here then the reusability decreases a bit

To me it increases. This component's goal is to show the project name and to potentially update it. With the current code it is also aware of the other stuff from the project provider, which increases the "risk" of reusing it since you'd need to make sure to also add the provider.
While if we use `$api` directly then the only dependency it has is the api.

ui/src/features/project/project-provider.component.tsx
MarkRedeman: Yeah I think for this it would be fine to do a bit of (local) prop drilling.

ui/src/features/project/label-selection/label-selection.component.test.tsx
jpggvilaca: I think ill address this on a different PR

MarkRedeman (approved):

PR #4707, opened 2025-09-16
Update tune routes

### LINKED ISSUE #4709: Update routes to reflect the project-centric navigation
/entitity should be /project/{projectId}/entity

jpggvilaca:
### Summary

* Routes should be project centric (/project/project-id/entity)

### Checklist

### License

```python
```

### DIFF
--- a/ui/src/hooks/use-project-identifier.hook.ts
+++ b/ui/src/hooks/use-project-identifier.hook.ts
@@ -1,0 +1,14 @@
+// Copyright (C) 2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+import { useParams } from 'react-router';
+
+export const useProjectIdentifier = () => {
+    const { projectId } = useParams<{ projectId: string }>();
+
+    if (!projectId) {
+        throw new Error('No projectId found in the route');
+    }
+
+    return projectId;
+};
--- a/ui/src/layout.tsx
+++ b/ui/src/layout.tsx
@@ -2,11 +2,12 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { Flex, Grid, Item, TabList, TabPanels, Tabs, View } from '@geti/ui';
-import { Outlet, useLocation, useParams } from 'react-router';
+import { Outlet, useLocation } from 'react-router';
 
 import { ReactComponent as BuildIcon } from './assets/icons/build-icon.svg';
 import { ReactComponent as LiveFeedIcon } from './assets/icons/live-feed-icon.svg';
 import { ReactComponent as Webhook } from './assets/icons/webhook.svg';
+import { useProjectIdentifier } from './hooks/use-project-identifier.hook';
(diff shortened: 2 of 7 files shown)


ui/src/layout.tsx
```diff
@@ -16,6 +15,8 @@ const iconStyles = {
 };
 
 const Header = () => {
+    const { projectId = '' } = useParams<{ projectId: string }>();
```
MarkRedeman: Perhaps extract these into a `useProjectIdentifier()` hook like we do in Geti.
We could add an extra guard/assert there to make sure `projectId` is not empty.

ui/src/router.tsx
```diff
@@ -22,29 +22,19 @@ import { EditProject } from './routes/project/edit-project';
 
 const root = path('/');
 const project = root.path('/project');
-const inference = root.path('/inference');
-const dataset = root.path('/dataset');
-const models = root.path('/models');
-const labels = root.path('/labels');
+const inference = project.path('/:projectId/inference');
```
MarkRedeman: General suggestion: make the route be `/projects` and not `/project`, then you can use a `project` path here and down below

```suggestion
const projects = root.path('/projects');
const project = projects.path('/:projectId');
const inference = project.path('/:projectId/inference');
```

ui/tests/data-collection.spec.ts
```diff
@@ -5,7 +5,7 @@ import { expect, test } from './fixtures';
 
 test.describe('Dataset', () => {
     test('list items', async ({ page }) => {
-        await page.goto('/dataset');
+        await page.goto('project/some-id/dataset');
```
MarkRedeman: Assuming you'd go with `/projects/some-id/dataset` then you'd need to update these

MarkRedeman (dismissed):

ui/src/router.tsx
jpggvilaca: Agree. Updated

ui/tests/data-collection.spec.ts
jpggvilaca: Updated

ui/src/layout.tsx
jpggvilaca: Updated

ui/src/hooks/use-project-identifier.hook.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright (C) 2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+import { useParams } from 'react-router';
+
+export const useProjectIdentifier = () => {
+    const { projectId } = useParams<{ projectId: string }>();
+
+    if (!projectId) {
+        throw new Error('No projectId found in the route');
```
jpggvilaca, line 10: for now added an error but later maybe we can redirect or show a notification.

pplaskie (approved):

PR #4720, opened 2025-09-17
Update project list

jpggvilaca:
### Summary

* Update project card
* Update mocked projects
* Integrate GET /projects replacing the mocks
* Style updates
* Add route for project listing instead of modal

Screenshot 2025-09-17 at 09 59 44

Menu actions will be implemented later

### Checklist

### License

```python
```


ui/src/features/project/interface.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright (C) 2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+export type Project = {
```
MarkRedeman: Can this be taken from the OpenAPI spec? I.e. use the exports from `openapi-spec.d.ts` that look like,
```
export type SchemaProjectConfig = components['schemas']['ProjectConfig'];
```

ui/src/features/project/list/project-list.component.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { $api } from '../../../api/client';
+import Background from '../../../assets/background.png';
+import { mockedProjects } from './mocked-projects';
+import { NewProjectLink } from './new-project-link.component';
+import { ProjectCard } from './project-card';
+
+import classes from './project-list.module.scss';
+
+export const ProjectList = () => {
+    const projects = $api.useQuery('get', '/api/projects');
```
MarkRedeman: Something to consider: we could use `useSuspenseQuery` more often and have a suspense boundary above this component. That way you could do `const { data: projects } = $api.useSuspenseQuery(...)` which guarantees that `data` exists.

To me the main benefit is that this would force us more to add error and data loading handling (i.e. error and suspense boundaries) as with `useQuery` you could do `query?.data ?? []` and get a way with using default values and not show any data loading indicators.

This isn't specific to this PR though, mainly mentioning it so that we can think about using this in the future.

ui/src/features/project/list/project-list.module.scss
```diff
@@ -28,6 +28,7 @@
     border-color: var(--spectrum-global-color-gray-200);
     border-radius: var(--spectrum-global-dimension-size-50);
     background-color: var(--spectrum-global-color-gray-50);
+    cursor: pointer;
```
MarkRedeman: np/reminder: this rule won't be necessary when changing the cards to be links.

ui/src/features/project/interface.ts
jpggvilaca: Updated

ui/src/features/project/list/project-list.component.tsx
jpggvilaca: Updated

ui/src/features/project/list/project-list.module.scss
jpggvilaca: Updated

camiloHimura (approved):

Issue #4699, opened 2025-09-15
Integrate /datasets api


jpggvilaca:
* Get media items
* Delete an item

PR #4725, opened 2025-09-17
4699 - render dataset items from api - P1

camiloHimura:
### Summary #4699

This PR renders items from the dataset endpoint, including thumbnails and a full-size image in the preview modal. The next PR will add functionality for deleting items.

https://github.com/user-attachments/assets/01132ddf-1528-4c23-9f11-559408babe02

Note: Since we can't call the add endpoint directly from the UI yet, we need to run this command in order to get data on the dataset list page
`curl 'http://localhost:7860/api/projects/{project_id}/dataset/items' \
  --request POST \
  --header 'Content-Type: multipart/form-data' -F "file=@${file_path}"`
  
 Replace `project_id` with the project ID on your local setup, and set `file_path` to the absolute path of the image you want to load. Only JPG and PNG images are supported
  

### Checklist

### License

```python
```


ui/src/features/dataset/gallery/use-get-items.hook.ts
```diff
@@ -0,0 +1,42 @@
+// Copyright (C) 2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+import { $api } from '../../../api/client';
+import { useProjectIdentifier } from '../../../hooks/use-project-identifier.hook';
+
+const datasetItemsLimit = 5;
+
+export const useGetItems = () => {
```
jpggvilaca:
```suggestion
export const useGetMediaItems
or
export const useGetDatasetItems
```

ui/src/features/dataset/toolbar/toolbar.component.tsx
```diff
@@ -4,19 +4,23 @@
...
     const { selectedKeys, setSelectedKeys, setMediaState } = useSelectedData();
     const totalSelectedElements = selectedKeys instanceof Set ? selectedKeys.size : 0;
     const hasSelectedElements = totalSelectedElements > 0;
 
-    const message = hasSelectedElements ? `${totalSelectedElements} selected` : `${response.items.length} images`;
+    const message = hasSelectedElements ? `${totalSelectedElements} selected` : `${items.length} images`;
 
     const handleToggleManyItemSelection = () => {
-        const images = response.items.map((item) => item.id);
+        const images = items.map((item) => String(item.id));
```
jpggvilaca, line 23: why do we need to `String` if the `id` is already a string? Also, i see that the DatasetItem has `id` optional as well. I already informed the team to fix. this for all entities. `id` should be required.

ui/src/features/dataset/gallery/use-get-items.hook.ts
camiloHimura: update with `useGetDatasetItems`

ui/src/features/dataset/toolbar/toolbar.component.tsx
camiloHimura: We need this because it is currently optional, and TS complains about it. We can leave it like this for now and remove it once the endpoint definition is updated.

jpggvilaca (approved):

PR #4735, opened 2025-09-19
Move segment-anything tool

### LINKED ISSUE #4612: Extract interactive segmentation tool

jpggvilaca:
### Summary

* Update lockfile
* Move SAM files from Geti Classic & update imports/interfaces

Most files are fully copy-paste, with the only change on the `mediaItem`, that we get from `useAnnotator`

Video:
https://github.com/user-attachments/assets/977dfc56-48c6-43f5-86f2-5a59d4bb4914

### Checklist

### License

```python
```


ui/src/features/annotator/annotator-canvas.tsx
```diff
@@ -46,20 +46,19 @@ export const AnnotatorCanvas = ({ mediaItem, isFocussed }: { mediaItem: Item; is
...
 
-                    {!isEmpty(mediaItem.annotations) && (
-                        <View gridArea={'innercanvas'}>
-                            <svg
-                                width={size.width}
-                                height={size.height}
-                                style={DEFAULT_ANNOTATION_STYLES}
-                                onClick={handleClickOutside}
-                            >
+                    <View gridArea={'innercanvas'}>
```
jpggvilaca, line 44: render the toolmanager even if we dont have annotations

ui/rsbuild.config.ts
```diff
@@ -54,4 +53,16 @@ export default defineConfig({
             },
         },
     },
+    server: {
```
jpggvilaca, line 56: needed for webworkers

ui/src/features/annotator/tools/segment-anything-tool/segment-anything-state-provider.component.tsx
```diff
@@ -0,0 +1,138 @@
+// Copyright (C) 2025 Intel Corporation
```
jpggvilaca: copy paste

ui/src/features/annotator/tools/segment-anything-tool/segment-anything-tool.component.tsx
```diff
@@ -0,0 +1,214 @@
+// Copyright (C) 2025 Intel Corporation
```
jpggvilaca, line 1: copy paste

ui/src/features/annotator/tools/segment-anything-tool/segment-anything.module.scss
```diff
@@ -0,0 +1,24 @@
+.stroke {
```
jpggvilaca, line 1: copy paste

ui/src/features/annotator/tools/segment-anything-tool/segment-anything.interface.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright (C) 2025 Intel Corporation
```
jpggvilaca, line 1: copy paste

ui/src/features/annotator/tools/segment-anything-tool/use-decoding-query.hook.ts
```diff
@@ -0,0 +1,74 @@
+// Copyright (C) 2025 Intel Corporation
```
jpggvilaca, line 1: copy paste

ui/src/features/annotator/tools/segment-anything-tool/use-single-stack-fn.hook.ts
```diff
@@ -0,0 +1,65 @@
+// Copyright (C) 2025 Intel Corporation
```
jpggvilaca, line 1: copy paste

ui/src/features/annotator/tools/segment-anything-tool/use-throttle-callback.hook.ts
```diff
@@ -0,0 +1,26 @@
+// Copyright (C) 2025 Intel Corporation
```
jpggvilaca, line 1: copy paste

ui/src/features/annotator/hooks/use-load-image-query.hook.ts
```diff
@@ -0,0 +1,30 @@
...
+    const projectId = useProjectIdentifier();
+
+    return useQuery({
+        queryKey: ['mediaItem', mediaItem?.id, projectId],
+        queryFn: async () => {
+            if (mediaItem === undefined) {
+                throw new Error("Can't fetch undefined media item");
+            }
+
+            const imageUrl = `http://localhost:7860/api/projects/${projectId}/dataset/items/${mediaItem.id}/binary`;
```
camiloHimura:
```suggestion
            const imageUrl = `${API_BASE_URL}/api/projects/${projectId}/dataset/items/${mediaItem.id}/binary`;
```

ui/src/features/annotator/hooks/use-load-image-query.hook.ts
```diff
@@ -0,0 +1,30 @@
...
+            const image = await loadImage(imageUrl);
+
+            return getImageData(image);
+        },
+        enabled: mediaItem !== undefined && !!projectId,
```
camiloHimura: `boolean` makes the code more readable

```suggestion
        enabled: mediaItem !== undefined && Boolean(projectId),
```

ui/src/features/annotator/loading.component.tsx
```diff
@@ -0,0 +1,44 @@
+// Copyright (C) 2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+import { Flex, Heading, View } from '@geti/ui';
+
+import IntelBrandedLoadingGif from '../../assets/intel-loading.webp';
+
+export const Loading = ({ isLoading }: { isLoading: boolean }) => {
```
camiloHimura: This is a pretty generic name, and from the code, I understand that this component specifically applies to sam images only.  `ImageProcessingSpinner` or `ImageProcessingLoader`

ui/src/features/annotator/tools/segment-anything-tool/interactive-segmentation-point.component.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+import { InteractiveAnnotationPoint } from './segment-anything.interface';
+
+interface InteractiveSegmentationPointProps extends InteractiveAnnotationPoint {
+    isLoading: boolean;
+}
+
+export const InteractiveSegmentationPoint = ({ x, y, positive, isLoading }: InteractiveSegmentationPointProps) => {
+    const fill = positive ? 'var(--brand-moss)' : 'var(--brand-coral-cobalt)';
+    const radius = `calc(1 / var(--zoom-scale))`;
```
camiloHimura: Line 19 also has a radius calculation, but with a different value. Could you use more descriptive variable names so it’s clear what each one is used for?

ui/src/features/annotator/tools/segment-anything-tool/use-segment-anything.hook.ts
```diff
@@ -0,0 +1,119 @@
...
+
+            setModelIsLoading(false);
+        };
+
+        if (data && modelRef.current === undefined && !modelIsLoading) {
+            loadWorker();
+        }
+    }, [data, modelIsLoading, algorithmType]);
+
+    return modelRef.current;
```
camiloHimura, line 53: Wouldn’t it be useful to return `modelIsLoading` as well?

ui/src/features/annotator/tools/svg-tool-canvas.component.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+import { roiFromImage } from '@geti/smart-tools/utils';
+
+import { allowPanning } from '../utils';
+
+type CanvasProps = SVGProps<SVGSVGElement> & { image: ImageData } & { canvasRef?: RefObject<SVGRectElement | null> };
+// This svg component is used to by tools that need to add local listeners that work in
+// a given region of interest.
+// An invisible rect is rendered to guarantee that the svg gets a width and height.
+export const SvgToolCanvas: FC<PropsWithChildren<CanvasProps>> = ({
```
camiloHimura, line 14: Will this only be used by SAM?

jpggvilaca: No, the plan is to add this to the drawingbox tool and the polygon tool later on

ui/src/features/annotator/loading.component.tsx
jpggvilaca: It will be used by any tool or event that requires the annotator to be in a loading state. The description is pretty generic as well. Basically everytime we call opencv we'll use this loading

ui/src/features/annotator/tools/segment-anything-tool/use-segment-anything.hook.ts
jpggvilaca: It wasnt exposed before, and it makes sense because that var is only used to control when we call `loadWorker`

ui/src/features/annotator/hooks/use-load-image-query.hook.ts
jpggvilaca: Updated

ui/src/features/annotator/hooks/use-load-image-query.hook.ts
jpggvilaca: Updated

ui/src/features/annotator/tools/segment-anything-tool/interactive-segmentation-point.component.tsx
jpggvilaca: Updated

ui/src/features/annotator/tools/segment-anything-tool/use-segment-anything.hook.ts
MarkRedeman: We could move the initialization into the initial query function and then return the full query. I think we didn't do this before because loading the worker itself was abstracted away so we couldn't combine that with other loading steps.

Now though we could combine building the worker and initializing it into the same query.

ui/src/features/annotator/annotator-provider.component.tsx
```diff
@@ -25,6 +35,9 @@ export const AnnotatorProvider = ({ mediaItem, children }: { mediaItem: DatasetI
     const [activeTool, setActiveTool] = useState<ToolType>('selection');
     // todo: pass media annotations
     const [annotations, setAnnotations] = useState<Annotation[]>([]);
+    const [isLoading, setIsLoading] = useState(false);
```
MarkRedeman: I'm not a fan of adding a loading state to the annotator to show a loading indicator for the tool.
Can we make the design of the tool so that it can render its own loading indicator? (that way we can avoid having to use `useEffect` to sync these states)

ui/src/features/annotator/loading.component.tsx
```diff
@@ -0,0 +1,44 @@
...
+            bottom={0}
+            UNSAFE_style={{
+                backgroundColor: 'var(--spectrum-alias-background-color-modal-overlay)',
+                zIndex: 10,
+            }}
+        >
+            <Flex direction={'column'} alignItems={'center'} justifyContent={'center'} height='100%' gap='size-100'>
+                {/*  eslint-disable-next-line jsx-a11y/img-redundant-alt */}
+                <img
+                    src={IntelBrandedLoadingGif}
```
MarkRedeman, line 24: Perhaps we can add a `variant={"spinner" | "intel-blocks"}` to the `Loading` component of `@geti/ui`? If we can then combine that with the fullscreen, inline etc modes of that loading component then that gives us some nice flexibility.

ui/rsbuild.config.ts
```diff
@@ -54,4 +53,17 @@ export default defineConfig({
...
+        headers: {
+            'Cross-Origin-Embedder-Policy': 'credentialless',
```
MarkRedeman, line 58: In Geti we use `require-corp` [here](https://github.com/open-edge-platform/geti/blob/main/web_ui/dev-proxy.ts#L92-L106), why did you pick `credentialless` here?

jpggvilaca: On firefox it wasnt loading media thumbnails. I think the backend might also need to update the response headers?

ui/src/features/annotator/annotator-provider.component.tsx
jpggvilaca: One of the reasons for this is that the loading needs to happen OUTSIDE of the tool component (svgs inside svgs, etc), and it also made sense to me to give the annotator the control to show its loading. What's the main issue?

ui/src/features/annotator/loading.component.tsx
jpggvilaca: I like the idea. I'll a task to make the loading more flexible

ui/src/features/annotator/annotator-provider.component.tsx
MarkRedeman: Why does it need to be outside? (In Geti the tools themselves were deciding if they needed to render svg or html, we can do that here too)

> What's the main issue?

To me it makes it more complicated because you're adding an explicit loading state of the tool (e.g. "downloading models", "initializing models", "encoding image") as a generic loading state of the annotator ("is loading"). So it will be hard to make any extra UX improvements and/or error handling.
We didn't do this well in Geti either but suppose initializing the models ended up errroring, then we'd need to show the user some info about this. If we keep this state in the annotator provider then it's responsibility will start to blow up.
Generally I think keeping the loading state as local as possible will make it easier for us to improve the UX.

ui/rsbuild.config.ts
MarkRedeman: > On firefox it wasnt loading media thumbnails. I think the backend might also need to update the response headers?

Ah yeah that makes sense. I think the best option would be to add back the devproxy configuration that we used before. That way the backend and UI will be on the same host and we don't have to deal with CORS and less secure headers.
For production we expect the backend and UI to be on the same host as well so that should make the dev environment be more in line with the prod environment.

jpggvilaca: Agree. Ill create a task for adding the proxy

ui/src/features/annotator/annotator-provider.component.tsx
jpggvilaca: ForeigObject did the trick

ui/src/features/annotator/tools/annotator-tools.component.tsx
```diff
@@ -17,7 +17,8 @@ const TASK_TOOL_CONFIG: Record<string, ToolConfig[]> = {
     ],
     segmentation: [
         { type: 'selection', icon: Selector },
-        // TODO: Add 'polygon' and 'sam' tools later
+        { type: 'sam', icon: BoundingBox },
```
MarkRedeman: np: do we have an up to date icon for this?

jpggvilaca: i forgot the name so i just added this one. Radwan doesnt want the SAM tool to be on the tool list, both tools (Detection and Polygon) will have an "auto" mode that will trigger SAM, so we wont need the icon at all. I just added for now so i can work on it and debug.

ui/src/features/annotator/tools/segment-anything-tool/segment-anything-tool.component.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+// Whenever the user moves their mouse over the canvas  we compute a preview of
+// SAM being applied to the user's mouse position.
+// The decoding step of SAM takes on average 100ms with 150-250ms being a high
+// exception. We throttle the mouse update based on this so that we don't overload
+// the user's cpu with too many decoding requests
+const THROTTLE_TIME = 150;
+
+const toolSettings = {
+    interactiveMode: true,
```
MarkRedeman: This should be false if we don't have a secondary toolbar. as the "interactive move" requires interactivity through the toolbar (e.g. accepting/rejecting the annotations, and configuring right click mode)

jpggvilaca: Updated

ui/src/features/annotator/tools/segment-anything-tool/interactive-segmentation-point.component.tsx
```diff
@@ -11,7 +11,7 @@ interface InteractiveSegmentationPointProps extends InteractiveAnnotationPoint {
 export const InteractiveSegmentationPoint = ({ x, y, positive, isLoading }: InteractiveSegmentationPointProps) => {
     const { scale } = useZoom();
     const fill = positive ? 'var(--brand-moss)' : 'var(--brand-coral-cobalt)';
-    const animationScale = `calc(1 / var(--zoom-scale))`;
+    const animationScale = 1 / scale;
     const pointRadius = 5 / scale;
```
MarkRedeman: Could we keep using css variables here instead? That will give a much smoother zoom in / out experience. (edit nvm: we can remove this component since we don't have interactive mode in this tool)

ui/src/features/annotator/tools/segment-anything-tool/segment-anything-tool.component.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+            {previewShapes.length > 0 &&
+                previewShapes.map((shape, idx) => (
+                    <g
+                        key={idx}
+                        aria-label='Segment anything preview'
+                        {...SELECT_ANNOTATION_STYLES}
+                        strokeWidth={'calc(3px / var(--zoom-scale))'}
+                        fillOpacity={0.0}
+                        className={classes.stroke}
```
MarkRedeman, line 137: This shape should include the `classes.animateStroke` class so that the border animates, giving the impression that this is not yet a finalized annotation

ui/src/assets/icons/pencil-minus.svg
MarkRedeman, line 1: This won't be needed without interactive mode.

ui/src/features/annotator/tools/segment-anything-tool/interactive-segmentation-point.component.tsx
MarkRedeman, line 1: This won't be needed without interactive mode.

ui/src/features/annotator/tools/segment-anything-tool/segment-anything-state-provider.component.tsx
```diff
@@ -0,0 +1,139 @@
...
+            if (!decodingQuery.isPlaceholderData) {
+                toast({
+                    message: `Unable to segment object from the selected point${
+                        state.points.length > 1 ? 's' : ''
+                    }. Press ESC to reset points.`,
+                    type: 'warning',
+                });
+            }
+        }
+    }, [decodingQuery.data, decodingQuery.isPlaceholderData, state.points]);
```
MarkRedeman: This won't be needed without interactive mode.

ui/src/features/annotator/tools/segment-anything-tool/segment-anything-state-provider.component.tsx
```diff
@@ -0,0 +1,139 @@
...
+        reset();
+    };
+
+    const handleCancelAnnotation = () => {
+        if (!isProcessing) {
+            reset();
+        }
+    };
+
+    const addPoint = (point: InteractiveAnnotationPoint, keepPreviousData = false) => {
```
MarkRedeman: This won't be needed without interactive mode.

ui/src/features/annotator/tools/segment-anything-tool/segment-anything-state-provider.component.tsx
MarkRedeman, line 1: I think all of the contents from the provider can be moved into the tool component since we don't have interactive mode (this was adding a lot of complexity to the tool's code in Geti because we needed to sync states between both versions)

ui/src/features/annotator/tools/segment-anything-tool/segment-anything-tool.component.tsx
```diff
@@ -0,0 +1,227 @@
...
+import { getRelativePoint, removeOffLimitPoints } from '../utils';
+import { InteractiveSegmentationPoint } from './interactive-segmentation-point.component';
+import { useSegmentAnything } from './segment-anything-state-provider.component';
+import { InteractiveAnnotationPoint } from './segment-anything.interface';
+import { useSingleStackFn } from './use-single-stack-fn.hook';
+import { useThrottledCallback } from './use-throttle-callback.hook';
+
+import classes from './segment-anything.module.scss';
+
+const isPositivePoint = (point: Point, shapes: Shape[], isRightClick: boolean, rightClickMode: boolean) => {
```
MarkRedeman: This won't be needed without interactive mode. We only support adding a single point, which will be positive

ui/src/features/annotator/tools/segment-anything-tool/segment-anything-tool.component.tsx
```diff
@@ -0,0 +1,227 @@
...
+        }
+
+        if (!rightClickMode && isRightButton(event)) {
+            return;
+        }
+
+        // The user must first place a positive point as otherwise we can't show a preview
+        if (rightClickMode && isRightButton(event) && points.length === 0) {
+            return;
+        }
```
MarkRedeman: Right click mode can be removed (only supported in interactive mode)

ui/src/features/annotator/tools/segment-anything-tool/segment-anything-tool.component.tsx
```diff
@@ -0,0 +1,227 @@
...
+        // In task chain don't allow the user to place a point outside the ROI
+        if (!pointInRectangle(roi, point)) {
+            return;
+        }
+
+        const positive = isPositivePoint(point, result.shapes, isRightButton(event), rightClickMode);
+
+        const shouldKeepPreviousPoints = interactiveMode === true || isRightButton(event);
+
+        addPoint({ x: point.x, y: point.y, positive }, shouldKeepPreviousPoints);
```
MarkRedeman: This can be a direct evocation of the decoder, whose result we add the annotation scene.
Before with interactive mode this was more complicated because we either had to add the point to our internal state and keep track of the result, or use a single point and then add the result to the annotation scene.

ui/src/features/annotator/tools/segment-anything-tool/segment-anything-tool.component.tsx
```diff
@@ -0,0 +1,227 @@
...
+    };
+
+    const showPreviewShapes = result.shapes.length === 0 && mousePosition !== undefined;
+    const annotations = (showPreviewShapes ? previewShapes : result.shapes).map((shape, idx): Annotation => {
+        return {
+            shape,
+            labels: [{ id: 'id', color: 'red', name: 'Segment Anything', isPrediction: false }],
+            id: `${idx}`,
+        };
+    });
```
MarkRedeman: This can be simplified since there should only be one source of the preview shapes. (before we had both the hover effect, and the interactive mode's result)

ui/src/features/annotator/tools/segment-anything-tool/segment-anything-tool.component.tsx
```diff
@@ -0,0 +1,227 @@
...
+            onPointerUp={onPointerUp}
+            onPointerLeave={() => {
+                throttleSetMousePosition.cancel();
+                setMousePosition(undefined);
+                setPreviewShapes([]);
+            }}
+            style={{
+                cursor: interactiveMode
+                    ? `url("/icons/pencil-plus.svg") 16 16, auto`
+                    : `url("/icons/selection.svg") 8 8, auto`,
```
MarkRedeman:
```suggestion
                cursor: `url("/icons/selection.svg") 8 8, auto`,
```

ui/src/features/annotator/tools/segment-anything-tool/segment-anything-tool.component.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+            {points.map((point, index) => (
+                <InteractiveSegmentationPoint
+                    key={index}
+                    x={point.x}
+                    y={point.y}
+                    positive={point.positive}
+                    isLoading={isProcessing}
+                />
+            ))}
```
MarkRedeman: This can be removed

ui/src/features/annotator/annotator-canvas.tsx
```diff
@@ -51,20 +41,16 @@ export const AnnotatorCanvas = ({ mediaItem, isFocussed }: AnnotatorCanvasProps)
...
-                            >
-                                <Annotations width={size.width} height={size.height} isFocussed={isFocussed} />
-
-                                <ToolManager />
-                            </svg>
-                        </View>
-                    )}
+                    <View gridArea={'innercanvas'}>
+                        <>
+                            {!isEmpty(annotations) && (
```
MarkRedeman: np: I think we can omit this empty check. Always rendering the `Annotations` component might help with tests as well: we could do an assertion that the contents of this svg is empty.

ui/src/features/annotator/annotator-provider.component.tsx
```diff
@@ -57,6 +65,8 @@ export const AnnotatorProvider = ({ mediaItem, children }: { mediaItem: DatasetI
                 annotations,
 
                 mediaItem,
+                image: imageQuery.data || new ImageData(mediaItem.width, mediaItem.height),
```
MarkRedeman: We should avoid the `|| new ImageData` here as it could result in very weird behaviour (e.g. any smart tool using image data would work on an empty image). Perhaps we can use a suspense query to load the image?

camiloHimura (approved):

ui/src/features/annotator/hooks/use-load-image-query.hook.ts
```diff
@@ -0,0 +1,30 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import { useSuspenseQuery, UseSuspenseQueryResult } from '@tanstack/react-query';
+
+import { API_BASE_URL } from '../../../api/client';
+import { useProjectIdentifier } from '../../../hooks/use-project-identifier.hook';
+import { getImageData, loadImage } from '../tools/utils';
+import { DatasetItem } from '../types';
+
+export const useLoadImageQuery = (mediaItem: DatasetItem | undefined): UseSuspenseQueryResult<ImageData, unknown> => {
```
dwesolow, line 11: Do we want to use `$api.useSuspenseQuery(...)`?

jpggvilaca: I dont know yet. We'll see as we go

jpggvilaca: But for now i added so we know that we always have an image and there's no need for `new ImageData(...)`

dwesolow (approved):

PR #4745, opened 2025-09-22
4699 - dataset api - p3

camiloHimura:
### Summary 

Implements removing multiple elements #4699

https://github.com/user-attachments/assets/e73492b9-1b17-42cb-a18a-359b064e01cc

### Checklist

### License

```python
```

### DIFF
--- a/ui/src/features/dataset/gallery/delete-media-item/alert-dialog-content.component.tsx
+++ b/ui/src/features/dataset/gallery/delete-media-item/alert-dialog-content.component.tsx
@@ -6,11 +6,11 @@ import { AlertDialog, Flex, Text } from '@geti/ui';
 import { useEventListener } from '../../../../hooks/event-listener/event-listener.hook';
 
 type AlertDialogContentProps = {
-    itemId: string;
+    itemsIds: string[];
     onPrimaryAction: () => void;
 };
 
-export const AlertDialogContent = ({ itemId, onPrimaryAction }: AlertDialogContentProps) => {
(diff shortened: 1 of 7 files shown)


MarkRedeman:
Quick comment: I feel like the multiple toasts are a bit too much, especially since they don't show a name (i.e. UUIDs) that is nice to read.
Is it possible to make it so that there is 1 notification that says" 10 images removed" or something like that?
Alternatively the user can already see that the images have been removed after we refresh the media list so maybe we don't even need a toast notification for this?

jpggvilaca:
> Quick comment: I feel like the multiple toasts are a bit too much, especially since they don't show a name (i.e. UUIDs) that is nice to read. Is it possible to make it so that there is 1 notification that says" 10 images removed" or something like that? Alternatively the user can already see that the images have been removed after we refresh the media list so maybe we don't even need a toast notification for this?

Im ok with the notification but like i commented before, it's better to display a quantity, or else this won't scale. I dont want to see 50 ids if i remove 50 images. The number is more than enough for me to confirm imho

camiloHimura:
> > Quick comment: I feel like the multiple toasts are a bit too much, especially since they don't show a name (i.e. UUIDs) that is nice to read. Is it possible to make it so that there is 1 notification that says" 10 images removed" or something like that? Alternatively the user can already see that the images have been removed after we refresh the media list so maybe we don't even need a toast notification for this?
> 
> Im ok with the notification but like i commented before, it's better to display a quantity, or else this won't scale. I dont want to see 50 ids if i remove 50 images. The number is more than enough for me to confirm imho

Goos Point, updated

jpggvilaca (approved):

PR #4747, opened 2025-09-23
Added project thumbnail endpoint

### LINKED ISSUE #4702: Implement endpoint for project thumbnail
Endpoint: `GET /api/projects/{project_id}/thumbnail`.

The endpoint should return the thumbnail of the first media of the project, so that the response is consistent over time and not affected e.g. by the addition of new items. If the dataset item is deleted, it is okay if the thumbnail changes to the next available media. Note that this endpoint is not responsible for generating thumbnails, just returning it.
If the project dataset is empty, return 404.
 
For reference, check the dataset item thumbnail endpoints: https://github.com/open-edge-platform/training_extensions/blob/85ed8df26eefbf44fc71867bdc6ff17403cf1d18/backend/app/api/endpoints/datasets.py#L135

A-Artemis:
### Summary

This PR adds the GET project thumbnail endpoint. It also links the thumbnail's ID to the project for consistency.

### How to test

http://geti-tune.localhost/api/projects/9d6af8e8-6017-4ebe-9126-33aae739c5fa/thumbnail
image

### Checklist

### License

```python
```


backend/app/services/project_service.py
```diff
@@ -47,3 +49,33 @@ def delete_project_by_id(self, project_id: UUID) -> None:
...
+                return thumbnail_path
+        else:
+            with get_db_session() as db:
+                project_repo = ProjectRepository(db)
+                dataset_repo = DatasetItemRepository(str(project_id), db)
+
+                dataset_items = dataset_repo.list_items(limit=10, offset=0)
+                for item in dataset_items:
+                    thumbnail_path = self._get_thumbnail_path_for_item(project_id=project_id, thumbnail_id=item.id)
+                    if thumbnail_path.exists():
```
leoll2: Why do you need to check if the path exists? Every dataset item is guaranteed to have a thumbnail, no?

A-Artemis: It is possible that the media is too large to crop, or corrupt media, or the thumbnail is simply yet generated. It is a good assumption to make that a thumbnail is present but I don't think it is a 100% guarantee.

backend/app/db/schema.py
```diff
@@ -31,6 +31,7 @@ class ProjectDB(Base):
     name: Mapped[str] = mapped_column(String(255), nullable=False)
     task_type: Mapped[str] = mapped_column(String(50), nullable=False)
     exclusive_labels: Mapped[bool] = mapped_column(Boolean, default=False)
+    thumbnail_id: Mapped[str | None] = mapped_column(String(255), nullable=True)
```
itallix: Would it make sense to add foreign key here to reference `DatasetItemDB` with `ondelete="SET NULL"`?

backend/app/schemas/project.py
```diff
@@ -22,12 +22,14 @@ class Task(BaseModel):
 
 class Project(BaseIDNameModel):
     task: Task
+    thumbnail_id: str | None = Field(default=None)
```
itallix: Should be `UUID`

backend/app/services/project_service.py
```diff
@@ -47,3 +49,33 @@ def delete_project_by_id(self, project_id: UUID) -> None:
...
+
+    def get_project_thumbnail_path(self, project_id: UUID) -> Path | None:
+        """Get the path to the project's thumbnail image, selecting one if none is set"""
+        project = self.get_project_by_id(project_id)
+
+        if project.thumbnail_id:
+            thumbnail_path = self._get_thumbnail_path_for_item(project_id=project_id, thumbnail_id=project.thumbnail_id)
+            if thumbnail_path.exists():
+                return thumbnail_path
+        else:
```
itallix: `else` is not required here

backend/app/services/project_service.py
```diff
@@ -47,3 +49,33 @@ def delete_project_by_id(self, project_id: UUID) -> None:
...
+            with get_db_session() as db:
+                project_repo = ProjectRepository(db)
+                dataset_repo = DatasetItemRepository(str(project_id), db)
+
+                dataset_items = dataset_repo.list_items(limit=10, offset=0)
```
itallix:
```suggestion
                # Note: In theory, all items on the first page of 10 could be invalid.
                # However, this is extremely rare, so we assume the happy path for simplicity.
                dataset_items = dataset_repo.list_items(limit=10, offset=0)
```

backend/app/services/project_service.py
leoll2: Thumbnails are generated synchronously wrt dataset item creation ([source](https://github.com/open-edge-platform/training_extensions/blob/42585cf4a56ae200fc1bd3cbefa882990ad1739f/backend/app/services/dataset_service.py#L80)), and before the record is committed to the DB, therefore you can always assume that the thumbnail is present. If the media is corrupt, it will raise an error earlier at validation ([source](https://github.com/open-edge-platform/training_extensions/blob/42585cf4a56ae200fc1bd3cbefa882990ad1739f/backend/app/services/dataset_service.py#L59)). In practice, you can assume that the thumbnail is always available, and if it's not then we should treat it as a bug.

What I'd like to avoid is storing the project thumbnail_id in the database, because it adds unnecessary complexity. To get the project thumbnail, you can simply query the dataset item with the earliest `created_at`, and return its thumbnail.

backend/app/alembic/versions/da385d690aae_schema.py
```diff
@@ -27,6 +27,7 @@ def upgrade() -> None:
         sa.Column("name", sa.String(length=255), nullable=False),
         sa.Column("task_type", sa.String(length=50), nullable=False),
         sa.Column("exclusive_labels", sa.Boolean(), nullable=False),
+        sa.Column("thumbnail_id", sa.Text(), nullable=True),
```
leoll2: This change should probably be reverted

backend/app/api/dependencies.py
```diff
@@ -166,9 +166,9 @@ def get_webrtc_manager(request: Request) -> WebRTCManager:
 
 
 @lru_cache
-def get_project_service() -> ProjectService:
+def get_project_service(request: Request) -> ProjectService:
```
leoll2: Adding a parameter like `request` breaks the LRU cache behavior, since all the requests are different, leading to a cache-miss every time. The `data_dir` is a static setting, why do you want to pass it as argument?

A-Artemis: I copied the dataset and model service. I change them as well now.

leoll2 (approved):

PR #4770, opened 2025-09-26
Migrate anomaly task to use the new Datumaro dataset

gdlg:
### Summary

This PR adds support for the new dataset class with the anomaly task. Note that the format of the data has changed a bit: there is now a global normal/anomalous label available with clear semantic. Compared to the old format, the logic is much simplified as we don’t to distinguish between different data types like Geti or mvtec.

For more details about the format, see the PR in Datumaro: https://github.com/open-edge-platform/datumaro/pull/1901

### Checklist

### License

```python
```


library/src/otx/data/dataset/anomaly_new.py
```diff
@@ -0,0 +1,36 @@
...
+from otx.types.task import OTXTaskType
+
+if TYPE_CHECKING:
+    from datumaro.experimental import Dataset
+
+
+class OTXAnomalyDataset(OTXDataset):
+    """OTXDataset class for anomaly task."""
+
+    def __init__(self, task_type: OTXTaskType, dm_subset: Dataset, **kwargs) -> None:
```
samet-akcay, line 24: not related to this pr, but looking for your suggestion in general..

here `dm_subset` arg name does not make sense to me as a reader. Is this really a subset? or dataset? Or something else? What are your thoughts?

samet-akcay (approved):

ashwinvaidya17 (approved):

library/src/otx/data/dataset/anomaly_new.py
gdlg: Yes, I agree. I think that we could just rename dm_subset to “dataset”. If you look at the overall code, then this is a subset of the overall dataset but I prefer to name variables based on what they do rather than how it’s used based the rest of the code and I am not sure that it makes to call it a subset.

Besides this class is called OTXAnomalyDataset, and the base class in PyTorch is also called torch.utils.data.Dataset. So it’s not very consistent to call the parameter “subset”.

PR #4775, opened 2025-09-26
Implement fixed rate data collection policy

### LINKED ISSUE #4690: Implement fixed rate data collection policy
The goal is to implement a simple policy that collects samples from the inference pipeline at a configured rate. 

The inference pipeline outputs new frames and predictions at a certain frequency, e.g. 10 fps. With the "fixed rate" policy, configured for example with a rate of 0.2 fps, the data collection stage should pick one frame and its prediction every 5 seconds, create a new dataset item out of it, then persist it. The new item becomes then part of the project dataset; the user can later review its prediction and optionally edit it before accepting.

In scope of this ticket:

- Implement a method to create a new dataset item from the frame + prediction information contained in `StreamData` (in practice, the image is a `np.ndarray` and the prediction is a `model_api.Result`.
  - this method could be in `DatasetService`, or somewhere else
    - if possible, unify the code paths of data collection and manual upload when it comes to thumbnail generation, to avoid logic duplication.
  - the `annotation_data` should be initialized from the prediction, following the same JSON schema as user-generated annotations. The only difference is that predictions also have a confidence score associated to each shape.
  -  `user_reviewed` should be set to False, since it's a prediction
  - `prediction_model_id` should be set to the ID of the pipeline model, while `source_id` should refer to the pipeline source
  - `subset` can be left unassigned
- Run the data collection policy in the dispatching worker
  - At each iteration, the worker should: 1) loop through the data collection policies enabled for the pipeline and evaluate them; if any returns true, create and store a new item in the dataset; 2) loop through the sinks and dispatch the results (this logic is already implemented, and does not depend on data collection).

warrkan:
### Summary

Resolves #4690

This PR introduces data collector module which is responsible for data collection and adding it to a project dataset based on the pipeline policies.
The only policy implemented at the moment is Fixed rate policy which allows to collect images with predefined frequency.

In case if pipeline is changed, policies are automatically reloaded.

### Checklist

### License

```python
```

### DIFF
--- a/application/backend/app/schemas/dataset_item.py
+++ b/application/backend/app/schemas/dataset_item.py
@@ -58,6 +58,7 @@ class DatasetItemAnnotation(BaseModel):
 
     labels: list[LabelReference]
     shape: Shape
+    probability: float | None = None
 
     model_config = {
         "json_schema_extra": {
--- a/application/backend/app/schemas/dataset_item.py
+++ b/application/backend/app/schemas/dataset_item.py
@@ -58,7 +58,7 @@ class DatasetItemAnnotation(BaseModel):
 
     labels: list[LabelReference]
     shape: Shape
-    probability: float | None = None
+    confidence: float | None = None
 
     model_config = {
         "json_schema_extra": {
--- a/application/backend/app/schemas/shape.py
+++ b/application/backend/app/schemas/shape.py
@@ -6,8 +6,8 @@ from pydantic import BaseModel, Field
 
 
 class Point(BaseModel):
-    x: float = Field(..., description="Point x coordinate", ge=0)
-    y: float = Field(..., description="Point y coordinate", ge=0)
+    x: int = Field(..., description="Point x coordinate", ge=0)
+    y: int = Field(..., description="Point y coordinate", ge=0)
 
 
 # Base class with discriminator field
@@ -17,10 +17,10 @@ class ShapeBase(BaseModel):
 
 class Rectangle(ShapeBase):
     type: Literal["rectangle"] = "rectangle"
-    x: float = Field(..., description="Rectangle x coordinate", ge=0)
-    y: float = Field(..., description="Rectangle y coordinate", ge=0)
-    width: float = Field(..., description="Rectangle width", ge=0)
-    height: float = Field(..., description="Rectangle height", ge=0)
+    x: int = Field(..., description="Rectangle x coordinate", ge=0)
+    y: int = Field(..., description="Rectangle y coordinate", ge=0)
+    width: int = Field(..., description="Rectangle width", ge=0)
+    height: int = Field(..., description="Rectangle height", ge=0)
 
     model_config = {
         "json_schema_extra": {"example": {"type": "rectangle", "x": 10, "y": 20, "width": 100, "height": 200}}
--- a/application/backend/app/schemas/shape.py
+++ b/application/backend/app/schemas/shape.py
@@ -6,8 +6,8 @@ from pydantic import BaseModel, Field
 
 
 class Point(BaseModel):
-    x: int = Field(..., description="Point x coordinate", ge=0)
-    y: int = Field(..., description="Point y coordinate", ge=0)
+    x: float = Field(..., description="Point x coordinate", ge=0)
+    y: float = Field(..., description="Point y coordinate", ge=0)
 
 
 # Base class with discriminator field
@@ -16,11 +16,11 @@ class ShapeBase(BaseModel):
 
 
 class Rectangle(ShapeBase):
-    type: Literal["rectangle"]
-    x: int = Field(..., description="Rectangle x coordinate", ge=0)
-    y: int = Field(..., description="Rectangle y coordinate", ge=0)
-    width: int = Field(..., description="Rectangle width", ge=0)
-    height: int = Field(..., description="Rectangle height", ge=0)
+    type: Literal["rectangle"] = "rectangle"
+    x: float = Field(..., description="Rectangle x coordinate", ge=0)
+    y: float = Field(..., description="Rectangle y coordinate", ge=0)
+    width: float = Field(..., description="Rectangle width", ge=0)
+    height: float = Field(..., description="Rectangle height", ge=0)
 
     model_config = {
         "json_schema_extra": {"example": {"type": "rectangle", "x": 10, "y": 20, "width": 100, "height": 200}}
@@ -28,14 +28,14 @@ class Rectangle(ShapeBase):
 
 
 class Polygon(ShapeBase):
-    type: Literal["polygon"]
+    type: Literal["polygon"] = "polygon"
     points: list[Point] = Field(..., description="Polygon points")
 
     model_config = {"json_schema_extra": {"example": {"type": "polygon", "points": [[10, 20], [20, 60], [30, 40]]}}}
 
 
 class FullImage(ShapeBase):
-    type: Literal["full_image"]
+    type: Literal["full_image"] = "full_image"
 
     model_config = {
         "json_schema_extra": {
--- a/application/backend/app/services/data_collect/data_collector.py
+++ b/application/backend/app/services/data_collect/data_collector.py
@@ -81,7 +81,7 @@ class DataCollector(metaclass=Singleton):
         )
         self.dataset_service.create_dataset_item(
             project_id=project.id,
-            name=str(int(timestamp)),
+            name=f"{timestamp:.4f}".replace(".", "_"),
             format=DatasetItemFormat.JPG,
             data=frame_data,
             user_reviewed=False,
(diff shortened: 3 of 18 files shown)


application/backend/app/services/data_collect/data_collector.py
```diff
@@ -0,0 +1,102 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class PolicyChecker(metaclass=ABCMeta):
+    @abstractmethod
+    def should_collect(self, timestamp: float) -> bool:
+        """
+        Checks if certain dispatched image should be collected to the project dataset
+        :param timestamp: dispatched image timestamp
```
leoll2, line 26:
```suggestion
        Checks if certain dispatched image should be collected to the project dataset
        
        :param timestamp: dispatched image timestamp
```
Please leave a blank line between the description and the parameters, otherwise PyCharm and other editors don't render it correctly

application/backend/app/schemas/shape.py
```diff
@@ -16,26 +16,26 @@ class ShapeBase(BaseModel):
...
-    type: Literal["rectangle"]
-    x: int = Field(..., description="Rectangle x coordinate", ge=0)
-    y: int = Field(..., description="Rectangle y coordinate", ge=0)
-    width: int = Field(..., description="Rectangle width", ge=0)
-    height: int = Field(..., description="Rectangle height", ge=0)
+    type: Literal["rectangle"] = "rectangle"
+    x: float = Field(..., description="Rectangle x coordinate", ge=0)
+    y: float = Field(..., description="Rectangle y coordinate", ge=0)
+    width: float = Field(..., description="Rectangle width", ge=0)
+    height: float = Field(..., description="Rectangle height", ge=0)
```
leoll2: Why this change? All coordinates should be integer (pixel values)

application/backend/app/schemas/shape.py
```diff
@@ -6,8 +6,8 @@
 
 
 class Point(BaseModel):
-    x: int = Field(..., description="Point x coordinate", ge=0)
-    y: int = Field(..., description="Point y coordinate", ge=0)
+    x: float = Field(..., description="Point x coordinate", ge=0)
+    y: float = Field(..., description="Point y coordinate", ge=0)
```
leoll2: Same as the other comment, should be int

application/backend/app/services/data_collect/prediction_converter.py
```diff
@@ -0,0 +1,105 @@
...
+                continue
+            if len(contour) <= 2 or cv2.contourArea(contour) < 1.0:
+                continue
+            points = [
+                Point(
+                    x=point[0][0] / width,
+                    y=point[0][1] / height,
+                )
+                for point in list(contour)
+            ]
```
leoll2: Why do you normalize the point coordinates? It's inconsistent to use relative coords for points and absolute coords for rectangles, let's just use absolute values everywhere. The endpoint spec also uses absolute coords.

application/backend/app/services/data_collect/data_collector.py
```diff
@@ -0,0 +1,102 @@
...
+        """
+        should_collect = any(checker.should_collect(timestamp) for checker in self.policy_checkers)
+        if not should_collect:
+            return
+        annotations = convert_prediction(
+            labels=project.task.labels, frame_data=frame_data, prediction=inference_data.prediction
+        )
+        self.dataset_service.create_dataset_item(
+            project_id=project.id,
+            name=str(int(timestamp)),
```
leoll2: The integer part of the timestamp has precision limited to the second, which means that you could have two dataset items with the same name if they're all collected within a time window of 1s. Maybe increase the precision down to milliseconds?
For example: `timestamp = 1759131683.823` -> `name = "1759131683-823"` (I used dash to keep the integer part easy to recognize)

application/backend/app/schemas/dataset_item.py
```diff
@@ -58,6 +58,7 @@ class DatasetItemAnnotation(BaseModel):
 
     labels: list[LabelReference]
     shape: Shape
+    probability: float | None = None
```
leoll2: One more thing: could you rename this attribute "probability" to "confidence"? It's more correct because, from a mathematical perspective, this value does not represent a probability.

application/backend/app/services/data_collect/prediction_converter.py
warrkan: Removed the normalization

application/backend/app/schemas/shape.py
warrkan: It comes as legacy from Geti classic, reverted it to int's

application/backend/app/services/data_collect/data_collector.py
warrkan: Changed the naming of dataset items to include fractional part as well

leoll2 (dismissed):

application/backend/app/services/data_collect/data_collector.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def should_collect(self, timestamp: float) -> bool:
+        time_since_last = timestamp - self.last_collect_time
+        if time_since_last < self.min_interval:
+            return False
+        self.last_collect_time = timestamp
+        return True
+
+
+class DataCollector(metaclass=Singleton):
```
itallix: Is there a need to define it as a `Singleton`?

application/backend/app/services/data_collect/prediction_converter.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def _convert_detection_prediction(labels: list[Label], prediction: DetectionResult) -> list[DatasetItemAnnotation]:
+    result = []
+    for idx, box in enumerate(prediction.bboxes):
+        label_name = prediction.label_names[idx]
+        confidence = prediction.scores.tolist()[idx]
+        label = next((label for label in labels if label.name == label_name), None)
+        if not label:
+            logger.warning(f"Prediction label {label_name} cannot be found in the project")
```
itallix: Let's employ the rule to have deferred string interpolation for logging? - [link](https://docs.astral.sh/ruff/rules/logging-f-string/)

application/backend/app/services/pipeline_service.py
```diff
@@ -45,6 +46,7 @@ def _notify_sink_changed(self) -> None:
     def _notify_pipeline_changed(self) -> None:
         self._notify_source_changed()
         self._notify_sink_changed()
+        DataCollector(self._active_pipeline_service).reload_policies()
```
itallix: I’d suggest avoiding this pattern. Consider passing dependencies through the constructor - [link](https://github.com/open-edge-platform/training_extensions/issues/4569). Let's reduce loose coupling between components.

application/backend/tests/integration/services/test_dataset_service.py
```diff
@@ -82,43 +82,71 @@ def _create_annotations(label_id: str) -> list[DatasetItemAnnotation]:
...
-        img = Image.new("RGB", (1024, 768))
-        img.save(img_byte_arr, "PNG")
+        project_id = fxt_stored_projects[0].id
+        image = Image.new("RGB", (1024, 768))
+        fxt_db_models[0].project_id = project_id
+        db_session.add_all([fxt_db_sources[0], fxt_db_models[0]])
+        db_session.flush()
+        stored_source_id = fxt_db_sources[0].id
+        model_revision_id = fxt_db_models[0].id
+        db_session.commit()
```
itallix: Why commiting? We rollback transaction after each test to keep the DB state always clean.

application/backend/app/services/data_collect/prediction_converter.py
warrkan: Good one, thanks!

application/backend/app/services/pipeline_service.py
warrkan: Agree, Refactored it to use constructor DI.

application/backend/tests/integration/services/test_dataset_service.py
warrkan: I think it accidentally leaked into test code. Removed it

application/backend/app/services/data_collect/data_collector.py
warrkan: DataCollector is a stateful component, it keeps for example data when latest frame has been collected and whether new one must be collected again. That's why it needs to be singleton

itallix: We can safely remove the `Singleton` now.

For stateful components, we have several options:

- Use lazy instantiation with `@lru_cache` in the dependency function.
- Instantiate the component during the application's lifespan and persist its state in an `app.state` attribute.
- Initialize it in a separate process during the `setup` phase.

Each approach ensures proper state management without relying on the Singleton pattern.

warrkan: Nice. It already has @lru_cache, so I suppose we can just drop Singleton.

itallix (approved):
LGTM!

Issue #4755, opened 2025-09-25
Hierarchical folder layout for model binaries


leoll2:
All model binaries are currently stored under `<BASE_DATA_DIR>/models`, a single folder for all projects. To better isolate the data of different projects and avoid potential conflicts, we'd like to store the models as follows:

```
BASE_DATA_DIR/
├─ projects/
│  ├─ 0cd6adee-8887-45d9-ae46-b90a214bb1ea/
│  │  ├─ models/
│  │  │  ├─ 6b7bb928-5d6f-46ea-8fd2-5ce80dd1e12b/
│  │  │  │  ├─ model.xml               # OpenVINO IR model (XML), converted from pytorch
│  │  │  │  ├─ model.bin               # OpenVINO IR weights (BIN)
```

Later, the model folders may also be used to store the respective training artifacts (e.g. the logs used to train that specific model).

PR #4782, opened 2025-09-29
Add BaseWeightsService

### LINKED ISSUE #4761: Implement logic to download model weights
Create the class `BaseWeightsService`, responsible for downloading pretrained weights from an external archive and managing a local cache of weight files.

Proposed methods:

- `get_remote_weights_path(task: TaskType, model_manifest_id: str) -> str`: return the remote location of the weights, as configured in the manifest
- `get_local_weights_path(task: TaskType, model_manifest_id: str, allow_download : bool = True) -> Path`: returns the location of the weights (.pt file) in the local storage. If not already present and `allow_download` is enabled, then it should first pull the weights from the remote. Before download, the function should check if there is enough space on disk.
- `remove_local_weights(task: TaskType, model_manifest_id: str)`: delete the local base weights of a specific model archicture to free space on disk.
- `remove_all_local_weights()`: same as above, but for all models

The location of the model weights is defined for each model architecture in its manifest (#4695).

A-Artemis:
### Summary

This PR introduces a new `BaseWeightsService` which handles downloading and fetching of model's pretrained weights.

### Checklist

### License

```python
```

### DIFF
--- a/application/backend/app/services/base_weights_service.py
+++ b/application/backend/app/services/base_weights_service.py
@@ -1,0 +1,261 @@
+# Copyright (C) 2025 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+import hashlib
+import logging
+import shutil
+from pathlib import Path
+
+import requests
+
+from app.supported_models import ModelManifest
+from app.supported_models.default_models import TaskType
+from app.supported_models.supported_models import SupportedModels
+
+logger = logging.getLogger(__name__)
+
+
+class BaseWeightsService:
+    """Service for downloading and managing pretrained model weights from external archives."""
+
+    def __init__(self, data_dir: Path) -> None:
+        self._weights_cache_dir = data_dir / "models" / "pretrained_weights"
+        self._weights_cache_dir.mkdir(parents=True, exist_ok=True)
+
+    def get_remote_weights_path(self, task: TaskType, model_manifest_id: str) -> str:
+        """
+        Return the remote location of the weights as configured in the manifest.
+
+        Args:
+            task: The task type (used for validation)
+            model_manifest_id: The unique identifier of the model architecture
+
+        Returns:
+            str: The remote URL of the pretrained weights
+        """
+        manifest = self._get_and_validate_model_manifest(task, model_manifest_id)
+        return manifest.pretrained_weights.url
+
+    def get_local_weights_path(self, task: TaskType, model_manifest_id: str, allow_download: bool = True) -> Path:
+        """
+        Return the location of the weights (.pt file) in local storage.
+
+        If not already present and allow_download is enabled, downloads the weights from remote.
+
+        Args:
+            task: The task type
+            model_manifest_id: The unique identifier of the model architecture
+            allow_download: Whether to download weights if not present locally
+
+        Returns:
+            Path: Path to the local weights file
+
+        Raises:
+            FileNotFoundError: If weights are not found locally and allow_download is False
+        """
+        manifest = self._get_and_validate_model_manifest(task, model_manifest_id)
+
+        task_dir = self._weights_cache_dir / task.name.lower()
+        task_dir.mkdir(parents=True, exist_ok=True)
+        local_path = task_dir / Path(manifest.pretrained_weights.url).name
+
+        if local_path.exists():
+            if self._verify_file_integrity(file_path=local_path, sha_sum=manifest.pretrained_weights.sha_sum):
+                logger.info(f"Using cached weights for {model_manifest_id}: {local_path}")
+                return local_path
+
+            logger.warning(f"Cached weights for {model_manifest_id} failed integrity check, will re-download")
+            local_path.unlink()
+
+        if not allow_download:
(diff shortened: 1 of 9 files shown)


application/backend/app/services/base_weights_service.py
```diff
@@ -0,0 +1,261 @@
...
+                temp_path.unlink()
+                raise ValueError(f"Downloaded file failed integrity check for {remote_url}")
+
+            # Move temporary file to final location
+            temp_path.rename(local_path)
+            logger.info(f"Successfully downloaded and verified weights: {local_path}")
+        except requests.RequestException as e:
+            logger.error(f"Failed to download weights from {remote_url}: {e}")
+        except Exception as e:
+            logger.error(f"Unexpected error downloading weights from {remote_url}: {e}")
```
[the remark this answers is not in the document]
A-Artemis: @leoll2 What do you think the behaviour should be here?

application/backend/app/services/base_weights_service.py
```diff
@@ -0,0 +1,262 @@
...
+from app.supported_models.supported_models import SupportedModels
+
+logger = logging.getLogger(__name__)
+
+
+class BaseWeightsService:
+    """Service for downloading and managing pretrained model weights from external archives."""
+
+    def __init__(self, data_dir: Path) -> None:
+        self._weights_cache_dir = data_dir / "models" / "pretrained_weights"
```
leoll2:
```suggestion
        self._weights_cache_dir = data_dir / "pretrained_weights"
```
The root `models` folder is going to be removed, instead models will be stored inside the respective project folder (#4755).

The `pretrained_weights` dir is supposed to be directly under `data_dir`: https://github.com/open-edge-platform/training_extensions/blob/6ec7c7d6ae419eca1e2683db1e8db6def3e3939f/docs/geti_tune/models.md?plain=1#L126-L127

application/backend/app/services/base_weights_service.py
```diff
@@ -0,0 +1,262 @@
...
+            model_manifest_id: The provided model manifest ID, or None to use default
+
+        Returns:
+            ModelManifest: The validated model manifest
+
+        Raises:
+            ValueError: If the model manifest is not found, task type mismatch, or doesn't have pretrained weights
+        """
+        manifest = SupportedModels.get_model_manifest_by_id(model_manifest_id)
+        if manifest.id == "null":
```
leoll2: My bad for not pointing out in the previous PR, but I don't like the `NullModelManifest` class. A more Pythonic and maintainable approach is to raise a `ManifestNotFoundException` in `SupportedModels.get_model_manifest_by_id` instead of returning a null-like object.

application/backend/app/services/base_weights_service.py
```diff
@@ -0,0 +1,262 @@
...
+            raise ValueError(f"Model manifest '{model_manifest_id}' not found")
+        if manifest.task.lower() != task.name.lower():
```
leoll2: I suggest to update the `ModelManifest.task` attribute to be `TaskType` instead of string, so you can do a direct comparison. It's also more robust because a ValidationError is raised when loading a manifest whose YAML references an unrecognized task type.

application/backend/app/services/base_weights_service.py
```diff
@@ -0,0 +1,262 @@
...
+            raise ValueError(f"Task mismatch: expected '{task.name.lower()}', got '{manifest.task.lower()}'")
+        if not manifest.pretrained_weights.url or manifest.pretrained_weights.url == "null":
+            raise ValueError(f"No pretrained weights configured for model '{model_manifest_id}'")
```
leoll2: This type of validation is not responsibility of `BaseWeightsService`. Since `ModelManifest.pretrained_weights.url` is not optional, this check is already performed by Pydantic when loading the manifest from YAML.

application/backend/app/services/base_weights_service.py
```diff
@@ -0,0 +1,262 @@
...
+        Returns:
+            Path: Path to the local weights file
+
+        Raises:
+            FileNotFoundError: If weights are not found locally and allow_download is False
+        """
+        manifest = self._get_and_validate_model_manifest(task, model_manifest_id)
+
+        task_dir = self._weights_cache_dir / task.name.lower()
+        task_dir.mkdir(parents=True, exist_ok=True)
```
leoll2: I suggest to move the logic to create the task directories to the constructor. In other words, in the constructor you should iterate over the task types and create the missing folders. This way we can detect much earlier issues like incorrect write permissions on the data_dir, already when the application is launched.

application/backend/app/services/base_weights_service.py
```diff
@@ -0,0 +1,262 @@
...
+        Verify the integrity of a downloaded file using SHA256 checksum.
+
+        Args:
+            file_path: Path to the file to verify
+            sha_sum: Expected SHA256 checksum
+
+        Returns:
+            bool: True if the file integrity is valid, False otherwise
+        """
+        if sha_sum == "null" or not sha_sum:
```
leoll2: Should never be "null" or None

leoll2 (approved):

PR #4803, opened 2025-09-30
Add redirect logic

jpggvilaca:
### Summary

* Add initial redirect depending on the number of projects

### Checklist

### License

```python
```

### DIFF
--- a/application/ui/src/router.tsx
+++ b/application/ui/src/router.tsx
@@ -91,7 +91,7 @@ export const router = createBrowserRouter([
                 ),
             },
             {
-                path: 'dataset',
+                path: paths.project.dataset.pattern,
                 element: (
                     <ZoomProvider>
                         <SelectedDataProvider>
@@ -101,7 +101,7 @@ export const router = createBrowserRouter([
                 ),
             },
             {
-                path: 'models',
+                path: paths.project.models.pattern,
                 element: <Models />,
             },
         ],
--- a/application/ui/src/router.tsx
+++ b/application/ui/src/router.tsx
@@ -1,10 +1,10 @@
 // Copyright (C) 2025 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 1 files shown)


application/ui/src/router.tsx
```diff
@@ -20,7 +20,43 @@ import { Models } from './routes/models/models';
...
+            } else {
+                navigate(paths.project.new({}), { replace: true });
+            }
+        } else {
+            // More than 1 project -> Load index page (/projects)
+            navigate(paths.project.index({}), { replace: true });
+        }
+    }, [projects, navigate]);
+
+    return <Loading />;
```
MarkRedeman: Instead of using `useEffect` and rendering `Loading` you could render a `<Navigate>` component. This way we could make sure that the user always ends up navigating to a different page instead of potentially getting stuck at the `<Loading>` screen (if for some reason the `useEffect` doesn't trigger).

Since you're using a `useSuspenseQuery` this should already trigger the loading component at the top of the router.

application/ui/src/router.tsx
```diff
@@ -48,14 +80,10 @@ export const router = createBrowserRouter([
...
-                loader: () => {
-                    // TODO: If there is no project configured then redirect to new project creation
-                    // else redirect to inference
-                    return redirect(paths.project.index({}));
-                },
+                element: <ViewProject />,
             },
             {
-                path: paths.project.inference.pattern,
+                path: 'inference',
```
MarkRedeman: Are these paths changes intentional?

MarkRedeman (dismissed):

application/ui/src/router.tsx
jpggvilaca: Agree, updated

application/ui/src/router.tsx
jpggvilaca: Nope ^^ updated

camiloHimura (approved):

PR #4805, opened 2025-10-01
Collected dataset items have inverted color channels

### LINKED ISSUE #4793: Collected dataset items have inverted color channels
**Steps to reproduce**

1. In Geti Tune, configure a dataset collection policy. For example `curl -s -X PATCH localhost:7860/api/projects/9d6af8e8-6017-4ebe-9126-33aae739c5fa/pipeline -H "Content-Type: application/json" --data '{"data_collection_policies": [{"type": "fixed_rate", "enabled": true, "rate": 0.2}]}'`
2. After a few seconds, view the dataset items in the UI or the browser (e.g. `localhost:7860/api/projects/9d6af8e8-6017-4ebe-9126-33aae739c5fa/dataset/items/35d016ec-0220-419e-bd78-2831b0de86ff/binary`)

**Expected**

The collected image looks identical to the frame processed by the inference pipeline

**Actual**

The image appears blue-ish, as if the color channels are inverted.

Image

warrkan:
### Summary

Resolves #4793 

Fixes a bug with incorrect RGB channels, Data Collector performs BGR -> RGB conversion if necessary.

### Checklist

### License

```python
```

### DIFF
--- a/application/backend/app/services/data_collect/data_collector.py
+++ b/application/backend/app/services/data_collect/data_collector.py
@@ -4,6 +4,7 @@ import logging
 from abc import ABCMeta, abstractmethod
 from uuid import UUID
 
+import cv2
 import numpy as np
 
 from app.entities.stream_data import InferenceData
@@ -78,6 +79,7 @@ class DataCollector:
         )
         if not should_collect:
             return
+        frame_data = cv2.cvtColor(frame_data, cv2.COLOR_BGR2RGB)  # Convert BGR to RGB
         annotations = convert_prediction(
             labels=project.task.labels, frame_data=frame_data, prediction=inference_data.prediction
         )
--- a/application/backend/tests/unit/services/data_collect/test_data_collector.py
+++ b/application/backend/tests/unit/services/data_collect/test_data_collector.py
@@ -1,7 +1,7 @@
 # Copyright (C) 2025 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 from datetime import datetime
-from unittest.mock import MagicMock, patch
+from unittest.mock import ANY, MagicMock, patch
 from uuid import uuid4
 
 import numpy as np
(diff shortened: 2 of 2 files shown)


application/backend/app/services/data_collect/data_collector.py
```diff
@@ -78,6 +79,7 @@ def collect(
         )
         if not should_collect:
             return
+        frame_data = cv2.cvtColor(frame_data, cv2.COLOR_BGR2RGB)  # Convert BGR to RGB
```
leoll2, line 82: I tried to look a bit deeper in the code to understand why this BGR->RGB conversion is necessary here for dataset collection, but not when writing to the folder sink. Here are my findings:

- the image traverses the inference pipeline as BGR. In other words, `StreamData.frame_data` is BGR.
- the dataset collection stage calls `DatasetService.create_dataset_item`, which calls `_read_image_from_ndarray`. The latter uses `PIL.Image.fromarray`, which assumes `RGB` if no mode is specified.
- the folder dispatcher instead uses `cv2.imencode` to encode the image as PNG. OpenCV assumes BGR by default.

Conclusion: instead of adding a conversion step here, I think we should fix `_read_image_from_ndarray`. Try using `Image.fromarray(data, mode="BGR")`: if that's not supported, then do the channel swap right before. Later, I think we should update the implementation to use OpenCV instead of Pillow.

warrkan: Unfortunately Pillow doesn't support `mode="BGR"`:

```
  File "/root/training_extensions/application/backend/app/services/dataset_service.py", line 58, in _read_image_from_ndarray
    return Image.fromarray(data, mode="BGR")
```

so we need to convert it before. I don't want to put it in dataset_service to keep there only Pillow dependency

leoll2 (approved):

PR #4811, opened 2025-10-01
Improve db session mgmt & upgrade sqlalchemy API

### LINKED ISSUE #4749: Improve SQLAlchemy API usage and db session management
Refactor the project’s SQLAlchemy integration to align with best practices and the latest API standards. 

**Acceptance Criteria:**
- Session factory improvements: update `SessionLocal` to use `autoflush=True` behavior by default to reduce boilerplate.
- Session DI: define `db_session` as a dependency and inject into services. (this will help to avoid monkeypatching in tests)
- No caching for stateless services: most of the services are stateless, we can avoid using `@lru_cache` with it.
- Upgrade to the latest 2.0 `SQLAlchemy API`:
  - Identify and refactor all usages of legacy SQLAlchemy APIs (e.g., `db.query`) to use the recommended modern API (e.g., `db.execute(select(...))`)
  - Update repository code and ensure all queries use the latest patterns.
- Testing improvements: review unit tests and replace monkeypatching of database/session logic with dependency injection or fixtures where possible.

itallix:
### Summary

This PR modernizes database session handling, upgrades to SQLAlchemy 2.0 API, and refactors `ModelService` to better separate stateful and stateless concerns:

- [x] Database sessions are now managed through FastAPI's dependency injection system, eliminating manual session handling
- [x] Sessions automatically commit on successful request completion
- [x] `SQLAlchemy`: Migrated from legacy `query()` API to modern `select()`, `update()`, and `delete()` constructs
- [x] Extracted `ActiveModelService` to separate stateful model operations from the stateless `ModelService`
- [x] Removed `@lru_cache` from stateless services
- [x] All datetime operations now explicitly use UTC to eliminate timezone-related bugs and ambiguity

*Note! Keeping repository instantiation in the service layer to avoid an already-large changeset from growing further. Ideally, services shouldn't know about the DB at all and should receive fully-configured repositories via DI. We can address this in a follow-up ticket.*

### How to test

- [x] Inference pipeline from seed setup should work as expected. 

### Checklist

### License

```python
```

### DIFF
--- a/application/backend/app/repositories/base.py
+++ b/application/backend/app/repositories/base.py
@@ -1,10 +1,11 @@
 # Copyright (C) 2025 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 
-from datetime import datetime
+from collections.abc import Sequence
+from datetime import UTC, datetime
 from typing import TypeVar
 
-from sqlalchemy import exists
+from sqlalchemy import delete, exists, select
 from sqlalchemy.orm import Session
 
 from app.db.schema import Base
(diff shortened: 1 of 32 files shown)


application/backend/app/repositories/base.py
```diff
@@ -23,40 +24,46 @@ def get_by_id(self, obj_id: str) -> ModelType | None:
         return self.db.get(self.model, obj_id)
 
     def exists(self, obj_id: str) -> bool:
-        return self.db.query(exists().where(self.model.id == obj_id)).scalar()  # type: ignore[attr-defined]
+        stmt = select(exists().where(self.model.id == obj_id))  # type: ignore[attr-defined]
```
leoll2: I wonder if it's possible to remove the `type: ignore[attr-defined]` here and in the other functions, by defining a `BaseWithId` class like this:

```python
class BaseWithId(DeclarativeBase):
    id: Mapped[str] = mapped_column(Text, primary_key=True, default=lambda: str(uuid4()))

class ProjectDB(BaseWithId):
    ...
```

Then if you declare `ModelType = TypeVar("ModelType", bound=BaseWithId)`, all the models should have the `id` attribute defined.

application/backend/app/services/base.py
```diff
@@ -125,6 +125,8 @@ def update(self, item: S, partial_config: dict, db: Session | None = None) -> S:
         with self._get_repo(db) as repo:
             to_update = item.model_copy(update=partial_config)
             updated = repo.update(self.config.mapper_class.from_schema(to_update))
+            if db:
+                db.commit()  # Ensure changes are committed for notification purposes
```
leoll2: What do you mean by "notification purposes"?

application/backend/app/services/configuration_service.py
```diff
@@ -45,69 +47,60 @@ def _notify_source_changed(self) -> None:
...
         Notification triggered only when the configuration is used by the active pipeline."""
-        pipeline_repo = PipelineRepository(db)
+        pipeline_repo = PipelineRepository(self._db_session)
         active_pipeline = pipeline_repo.get_active_pipeline()
         if active_pipeline and getattr(active_pipeline, field) == str(config_id):
             notify_fn()
 
     def list_sources(self) -> list[Source]:
-        return self._source_service.list_all()
+        return self._source_service.list_all(self._db_session)
```
leoll2: The source and sink services receive the session through the methods (`list_all`, `get_by_id`, ...), while other services like `ConfigurationService` and `DatasetService` get the session through the constructor (`__init__`). Is it by design? Can't we pass the session to the `GenericPersistenceService` when instantiating it?

leoll2 (dismissed):

application/backend/app/services/configuration_service.py
itallix: Good point, we can pass in the constructor indeed.

application/backend/app/services/base.py
itallix: Since the transaction is aligned with the HTTP request lifecycle now, it will be committed automatically at the end. However, when configuration changes occur, sometimes we need to notify other services that attempt to read the updated data in a separate transaction and will not see the changes until the current transaction is committed.

Alternatively, we could propagate updates directly in an `Event` payload, eliminating the need to query the database in dependent places. This can be explored in a separate ticket.

application/backend/app/repositories/base.py
itallix: Currently, `PipelineDB` does not have an `ID`, but the base repository should be generalizable to all database schema entities.

application/backend/app/services/base.py
itallix: I have updated the comment. Hopefully, it now better captures the context.

leoll2 (dismissed):

application/backend/app/repositories/base.py
leoll2: If a DB entity doesn't have the ID, then some repo methods inherited by BaseRepository are broken. For example `PipelineRepository.exists(obj_id)`, this method shouldn't exist in the first place.

 Would it make sense to define two base repo classes, one inheriting from the other, where the first base implements all methods that do NOT require an id in the model, and the second one implements all the other methods?

Iterating on the previous example (I use the term "first class" to denote the entities with an id):

```python
class Base(DeclarativeBase):
    pass

class FirstClassBase(Base): 
    id: Mapped[str] = mapped_column(Text, primary_key=True, default=lambda: str(uuid4()))

ModelType = TypeVar("ModelType", bound=Base)
FirstClassModelType = TypeVar("ModelType", bound=FirstClassBase)

class BaseRepository[ModelType]:
    """All base methods that don't require an ID go here"""
    def save(...):
        ...

    def list_all(...):
        ...

class FirstClassBaseRepository(BaseRepository[FirstClassModelType]):
    """All base methods that require an ID go here"""
    def get_by_id(...):
        ...

    def exists(...):
        ...
```
Note: I don't want to add too many abstractions, but at the same time I would like to have correct type-annotations.

itallix: Added the `BaseID` class for DB schema entities and redesigned `BaseRepository`. Notably, mypy behaves differently with `BaseRepository(Generic[ModelType])` versus `BaseRepository[ModelType]`; in the latter case, using `type[ModelType]` can result in attribute errors.

I’ve also decoupled `PipelineRepository` - currently, it’s the only model lacking an ID. If more such entities appear, we can introduce a separate base class for them.

leoll2 (approved):

PR #4812, opened 2025-10-01
Improved integrity errors handling

### LINKED ISSUE #4777: Incorrect exception message when attempting to create project with duplicate labels
**Steps to reproduce**

1. Try to create a Geti Tune project with two labels that have the same name

This is equivalent to submitting this request (note the duplicate labels):
```sh
curl -s -x POST localhost:7860/api/projects -H "Content-Type: application/json" --data '{ "name": "animals", "task": { "task_type": "classification", "exclusive_labels": true, "labels": [ { "name": "cat" }, { "name": "cat" } ] } }'
```

**Expected behavior**

The request is rejected with error 404 or 409, with a message that indicates the duplicate label

**Actual behavior**

The request is rejected with a misleading message:

> `{"detail":"Project with name 'animals' already exists."}`

The project name is not duplicated, the label names are (within the request). This issue is probably caused by how integrity errors raised by children entities are handled and propagated to the parent: https://github.com/open-edge-platform/training_extensions/blob/2c55cc54d7b7290f44668e1fdd4773e0be6effdd/application/backend/app/services/base.py#L121

warrkan:
### Summary

Resolves #4777

### Checklist


application/backend/app/services/project_service.py
```diff
@@ -29,7 +31,10 @@ def __init__(self, data_dir: Path) -> None:
 
     @parent_process_only
     def create_project(self, project: Project) -> Project:
-        return self._persistence.create(project)
+        try:
+            return self._persistence.create(project)
+        except UniqueConstraintIntegrityError:
+            raise DuplicateLabelsError
```
leoll2: The project creation endpoint allows the client to provide an id. It's a bit of an edge case, but it may happen that the client-provided id conflicts with an existing project and a `PrimaryKeyIntegrityError` is raised here. In such case, does the API return 409 with a clear error message? You can test it by calling `POST /projects` twice with the same id in the payload.

leoll2 (dismissed):
LGTM, one minor comment

application/backend/app/api/endpoints/sinks.py
```diff
@@ -97,7 +102,7 @@ async def create_sink(
 
     try:
         return configuration_service.create_sink(sink_config)
-    except ResourceAlreadyExistsError as e:
+    except ResourceWithNameAlreadyExistsError as e:
```
itallix: We should handle the case where ids are provided as well, since all resources allow passing an ID during creation.

warrkan: It will be handled by https://github.com/open-edge-platform/training_extensions/pull/4812/files#diff-831489a76e6762f2886866f1bdbffe2953a1fc9dbd49ee5085ab59c2176d74f5R122

application/backend/app/services/project_service.py
warrkan: Yes, if you try to create a project with the existing ID, you'll get a proper error:
image

application/backend/app/services/data_collect/data_collector.py
```diff
@@ -107,7 +107,11 @@ def collect(
             labels=project.task.labels, frame_data=frame_data, prediction=inference_data.prediction
         )
         with get_db_session() as session:
-            dataset_service = DatasetService(self.data_dir, session)
+            label_service = LabelService(db_session=session)
+            project_service = ProjectService(data_dir=self.data_dir, db_session=session, label_service=label_service)
+            dataset_service = DatasetService(
+                data_dir=self.data_dir, db_session=session, project_service=project_service
+            )
```
leoll2, line 119: Would it be possible to simplify this code by letting the `DatasetService` instantiate `ProjectService` if not provided by argument? Similarly, `ProjectService` could instantiate a `LabelService` if not passed explicitly.

application/backend/app/services/__init__.py
```diff
@@ -1,16 +1,24 @@
 # Copyright (C) 2025 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
-from .active_model_service import ActiveModelService
+
+from .base import (  # isort: skip
```
leoll2: Sorting the imports manually through `# isort: skip` allows to avoid circular imports, but it becomes harder to maintain as the number of imports grows. A possible alternative, imho better, is to use _relative_ imports within the module (`app.services` in this case), as in the following example:

image

If you agree with this approach, I can push a commit to fix all the imports in the services module.

leoll2: Pushed fix -> [02a43b4](https://github.com/open-edge-platform/training_extensions/pull/4812/commits/02a43b409145bb7e663cac09bc32c4a258e609d5)

application/backend/app/services/active_pipeline_service.py
```diff
@@ -75,7 +78,7 @@ def _load_app_config(self) -> None:
 
             project = pipeline.project
             if project is not None:
-                self._project = ProjectMapper.to_schema(project)
+                self._project = ProjectMapper.to_schema(project, LabelService(db).list_all(UUID(project.id)))
```
itallix, line 81: I believe we've reached a point where view objects no longer align well with our internal service needs. While these objects work fine when used by routers, they're not ideal for internal operations - particularly in stateful services like `ActiveModelService` and `ActivePipelineService` (e.g., labels are meaningless within the scope of `ActivePipelineService`).

For services that maintain runtime state, we should consider designing proper domain objects tailored to their specific operational requirements. However, this is probably something for future tasks rather than the current PR.

application/backend/app/services/data_collect/data_collector.py
warrkan: I wouldn't do it, let's keep instances instantiation a responsibility of DI component. Otherwise it can become a mess

warrkan: @itallix what do you think?

application/backend/app/services/active_pipeline_service.py
warrkan: I completely agree, and afaik I've raised this topic at some point. In general it's a good idea to separate presentation layer objects and business layer objects.

application/backend/app/services/__init__.py
warrkan: Thanks @leoll2 !

application/backend/app/api/endpoints/sinks.py
itallix:
```suggestion
    except (ResourceWithNameAlreadyExistsError, ResourceWithIdAlreadyExistsError) as e
```

application/backend/app/api/endpoints/sources.py
```diff
@@ -108,7 +114,7 @@ def create_source(
     """Create and configure a new source"""
     try:
         return configuration_service.create_source(source_create)
-    except ResourceAlreadyExistsError as e:
+    except ResourceWithNameAlreadyExistsError as e:
```
itallix:
```suggestion
    except (ResourceWithNameAlreadyExistsError, ResourceWithIdAlreadyExistsError) as e:
```

application/backend/app/services/data_collect/data_collector.py
itallix: The current approach with DI is the correct pattern and should be maintained. Services should focus on their business logic, not instantiating their dependencies. 
We can extract factories when a number of dependencies is required like here.

BTW, we still violate this pattern when it comes to repositories - possible task for a future refactor.

application/backend/app/services/dispatchers/filesystem.py
```diff
@@ -10,7 +10,8 @@
 from model_api.models.result import Result
 
 from app.schemas.sink import FolderSinkConfig, OutputFormat
-from app.services.dispatchers.base import BaseDispatcher
+
+from .base import BaseDispatcher
```
itallix, line 14: Is there a reason to change this to relative import.

Currently we have three groups of imports:
1. sdk
2. third party libs
3. app deps

This is introducing unnecessary group within app deps (abosolute + relative). I find it convenient for `__init__.py` to expose inners, but probably unnecessary here?

leoll2: Related comment: https://github.com/open-edge-platform/training_extensions/pull/4812#discussion_r2418911821

When Python modules import classes from sibling modules and these classes are exposed with `__init__.py`, it's very common to run into circular imports. To avoid that, here we use relative imports. It's important to note that relative imports are used only to import other _services_ (i.e. classes defined within this module `app.services`), while for everything else it is still better to use absolute imports.

application/backend/app/services/mappers/label_mapper.py
```diff
@@ -15,7 +16,9 @@ def to_schema(label_db: LabelDB) -> Label:
         return Label.model_validate(label_db, from_attributes=True)
 
     @staticmethod
-    def from_schema(label: Label) -> LabelDB:
+    def from_schema(project_id: UUID, label: Label) -> LabelDB:
         """Convert Label schema to db model."""
 
-        return LabelDB(**label.model_dump(mode="json"))
+        label_db = LabelDB(**label.model_dump(mode="json"))
```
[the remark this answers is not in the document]
warrkan: Label doesn't have `project_id` field, so it cannot be initialized via constructor

application/backend/app/services/dataset_service.py
```diff
@@ -162,36 +164,40 @@ def list_dataset_items(
...
         if not dataset_item:
             raise ResourceNotFoundError(ResourceType.DATASET_ITEM, str(dataset_item_id))
-        return self.projects_dir / f"{project_id}/dataset/{dataset_item.id}-thumb.jpg"
+        return self.projects_dir / f"{project.id}/dataset/{dataset_item.id}-thumb.jpg"
 
     def delete_dataset_item(self, project_id: UUID, dataset_item_id: UUID) -> None:
         """Delete a dataset item by its ID"""
-        repo = DatasetItemRepository(project_id=str(project_id), db=self._db_session)
+        project = self._project_service.get_project_by_id(project_id)
+        repo = DatasetItemRepository(project_id=str(project.id), db=self._db_session)
```
leoll2, line 205: Why do you need to load the `project` here? What's wrong with the old implementation that used `project_id` directly?

warrkan: Because I think it's more correct to indicate that project doesn't exist rather than dataset item is missing

leoll2 (approved):

application/backend/app/services/configuration_service.py
```diff
@@ -26,16 +34,46 @@ class PipelineField(StrEnum):
...
+        try:
+            return super().create(item)
+        except UniqueConstraintIntegrityError:
+            raise ResourceWithNameAlreadyExistsError(ResourceType.SOURCE, item.name)
+
+    def update(self, item: Source, partial_config: dict) -> Source:
+        try:
+            return super().update(item, partial_config)
+        except UniqueConstraintIntegrityError:
+            raise ResourceWithNameAlreadyExistsError(ResourceType.SOURCE, partial_config["name"])
```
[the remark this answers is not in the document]
warrkan: If the error is raised, it means that there was an attempt to change source name to existing one. It means that `name` in this case must be present in `partial_config`

application/backend/app/services/configuration_service.py
```diff
@@ -26,16 +34,46 @@ class PipelineField(StrEnum):
...
+        try:
+            return super().create(item)
+        except UniqueConstraintIntegrityError:
+            raise ResourceWithNameAlreadyExistsError(ResourceType.SINK, item.name)
+
+    def update(self, item: Sink, partial_config: dict) -> Sink:
+        try:
+            return super().update(item, partial_config)
+        except UniqueConstraintIntegrityError:
+            raise ResourceWithNameAlreadyExistsError(ResourceType.SINK, partial_config["name"])
```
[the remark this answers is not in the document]
warrkan: The same here

PR #4815, opened 2025-10-02
Use new datumaro for multi label and hierarchical classification

AlbertvanHouten:
This pull request introduces support for multi-label classification tasks and hierachical classification tasks to the OTX data pipeline. The main changes include the addition of new sample types and dataset classes for multi/hierarchical label classification, updates to the dataset factory to use these new components, and utility functions for proper label handling.

### Checklist


ashwinvaidya17 (approved):

gdlg (approved):
LGTM, thanks Albert!

library/src/otx/data/entity/sample.py
```diff
@@ -99,6 +99,13 @@ class ClassificationSample(OTXSample):
     label: torch.Tensor = label_field(pl.Int32())
 
 
+class ClassificationMultiLabelSample(OTXSample):
+    """ClassificationMultiLabelSample is a base class for OTX multi label classification items."""
+
+    image: np.ndarray | tv_tensors.Image = image_field(dtype=pl.UInt8)
+    label: np.ndarray | torch.Tensor = label_field(pl.Int32(), multi_label=True)
```
leoll2, line 142: Should it be named `labels` instead of `label`? By the way, can you show an example how to instantiate a `ClassificationMultiLabelSample` with two labels? How does the `multi_label` parameter work?

AlbertvanHouten: Labels would make more sense, however OTX shares sample structure through the OTXSample class between different task types, so renaming it is not possible at the moment unless we remove the inheritance from OTXSample. This will require a larger refactoring, which I don't want to do as part of this PR.

leoll2: Do you mean the `label` property of the `OTXSample`?

https://github.com/open-edge-platform/training_extensions/blob/4751cb27fb049df2113e33bb6ffa530cf9d67f4d/library/src/otx/data/entity/sample.py#L71

This class has a bunch of properties with no implementation (`masks`, `bboxes`, ...) and all these properties are inherited by the specialized samples even when not meaningful. It seems a bit weird and dangerous antipattern to me. For example, why does `classification_sample.polygons` silently return `None` instead of raising an exception (missing method / not implemented)? The name `label` could be misleading for the code that consumes this property, since it may not expect more than 1 label to be returned.

AlbertvanHouten: Yes, it is very much a problem however in the feature branch we are simply mimicking OTXDataItem https://github.com/open-edge-platform/training_extensions/blob/d27ff0271c1b491aa3b89e8d47e29026882623de/library/src/otx/data/entity/torch/torch.py

It's not so easy to refactor this since OTX is passing this whole object around in multiple places instead of only passing around what is needed. For example, a lot of the [transforms](https://github.com/open-edge-platform/training_extensions/blob/d27ff0271c1b491aa3b89e8d47e29026882623de/library/src/otx/data/transform_libs/torchvision.py) depend on the entire OTXDataItem object, which makes it quite a big effort to refactor.

leoll2 (approved):

PR #4828, opened 2025-10-02
Add E2E tests infrastructure

jpggvilaca:
### Summary

* Add E2E frontend config
* Add E2E GHA job
* Minor copy updates
* Update eslint to include the correct test paths

### Checklist


.github/workflows/ui-lint-and-test.yaml
```diff
@@ -353,6 +353,83 @@ jobs:
...
+        run: npm run test:e2e
+
+      - name: Upload E2E results
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        if: always()
+        with:
+          name: e2e-test-results
+          path: application/ui/playwright-report/
+
   required_check:
```
AlexanderBarabanov, line 435: I think, we need to add `e2e-tests` into `needs:` section of this job.

AlexanderBarabanov (dismissed):

.github/workflows/ui-lint-and-test.yaml
```diff
@@ -353,6 +353,80 @@ jobs:
...
+        uses: actions/download-artifact@634f93cb2916e3fdff6788551b99b062d0335ce0 # v5.0.0
+        with:
+          name: openapi-spec
+          path: application/ui/src/api
+
+      - name: Prepare UI
+        working-directory: application/ui
+        run: |
+          tar -xzf dist.tar.gz
+          npm ci
```
dwesolow: Could we extract dependency installation as a separate step?

application/ui/src/components/project-panel/projects-list-panel.component.tsx
```diff
@@ -59,7 +59,7 @@ const AddProjectButton = () => {
             onPress={addProject}
         >
             <AddCircle />
-            <Text marginX='size-50'>Add project</Text>
+            <Text marginX='size-50'>Create project</Text>
```
dwesolow, line 62: Why this change? If we change here, we should change it in all the apps.

jpggvilaca: To make it consistent within Tune

application/ui/playwright.config.ts
```diff
@@ -10,6 +10,7 @@ const CI = !!process.env.CI;
  */
 export default defineConfig({
     testDir: './tests',
+    testIgnore: '**/e2e/**',
```
dwesolow, line 13: WDYT of having `tests/component` so we don't need to ignore e2e? (we'd have two top directories, `e2e` and `component`)

dwesolow (approved):

application/ui/playwright.config.ts
jpggvilaca: `tests` is for component/e2e, the unit ones will live alongside the components themselves. It's similar on classic, but in here I used 2 configs instead of having if/else's everywhere

AlexanderBarabanov (approved):

application/ui/playwright.config.ts
jpggvilaca: I dont mind that approach tho. I will merge this for now and once we have e2e on Prompt perhaps we can reach a consensus

Issue #4743, opened 2025-09-22
Add proxy to UI development


jpggvilaca:
Use a simplified version of Geti Classic's UI proxy to make sure we dont need to worry about headers or other browser issues (e.g cors)

PR #4837, opened 2025-10-03
#4743 Use dev proxy for development

MarkRedeman:
### Summary

When deploying Tune we expect the frontend and backend to be hosted on the same domain.
By using a proxy to talk to the backend API we can simplify our development setup as this will mimic the production setup. This way we don't have to consider some security measures like CORS and simplify our CSP headers.

*Note*: currently the target is hardcoded. Perhaps we can change this to depend on the server's settings.

### Checklist


jpggvilaca (approved):

application/ui/src/api/client.ts
```diff
@@ -6,7 +6,7 @@ import createClient from 'openapi-react-query';
 
 import type { paths } from './openapi-spec'; // generated by openapi-typescript
 
-export const API_BASE_URL = import.meta.env.PUBLIC_API_BASE_URL || '';
+export const API_BASE_URL = '/';
```
MarkRedeman, line 9: I think this change might break the tests. Will try to fix this in a cleaner way

application/ui/rsbuild.config.ts
```diff
@@ -61,9 +57,16 @@ export default defineConfig({
...
                 "worker-src 'self' blob:; " +
-                "connect-src 'self' http://localhost:7860 data:; " +
-                "img-src 'self' http://localhost:7860 data: blob:; " +
+                "connect-src 'self' data:; " +
+                "img-src 'self' data: blob:; " +
                 "style-src 'self' 'unsafe-inline';",
         },
+        proxy: {
+            '/api': {
+                target: 'http://localhost:7861',
```
jpggvilaca, line 66: shouldnt it be `7860` here?

PR #4842, opened 2025-10-06
Rmt det inst mask head fix

### LINKED ISSUE #4823: RTMDet Inst poor performance
When comparing instance segmentation models RTMDet scores way lower on validation metrics compared to other models on my custom dataset. 

I am running on the latest commit of origin/develop: 472120a397ee1520b9ebee519579b5efa325c36c

With the command: 
`otx train --config library/src/otx/recipe/instance_segmentation/rtmdet_inst_tiny.yaml --data_root dataset_path/
`
I get these results;
Image

I've compared it to MaskRCNN TV:
`otx train --config library/src/otx/recipe/instance_segmentation/maskrcnn_r50_tv.yaml --data_root dataset_path/
`
Image

rtmdet inst has significantly lower val/map: 0.889 vs 0.125

LuukvandenBent:
self.mask_postprocess expects H,W but was given W,H. This PR fixes that along with properly using reg_convs instead of cls_convs for the self.reg_convs.

Solves https://github.com/open-edge-platform/training_extensions/issues/4823


kprokofi (approved):
Thanks for catching that!

leoll2 (approved):

PR #4853, opened 2025-10-08
Update OTXDataModule

### LINKED ISSUE #4448: Simplify API for OTXDatamodule
## Problem

Currently, creating an OTXDatamodule requires specifying a large number of parameters manually, including detailed SubsetConfig objects for training, validation, and test splits.

Example:
```
train_subset_config = SubsetConfig(
    batch_size=64,
    subset_name="train",
    transforms=v2.Compose([
        v2.RandomResizedCrop(size=(224, 224), antialias=True),
        v2.RandomHorizontalFlip(p=0.5),
        v2.ToDtype(torch.float32, scale=True),
        v2.Normalize(mean=[0.485, 0.456, 0.406], std=[0.229, 0.224, 0.225]),
    ]),
    transform_lib_type=TransformLibType.TORCHVISION,
    num_workers=2,
)

val_subset_config = SubsetConfig(
    batch_size=64,
    subset_name="val",
    transforms=v2.Compose([
        v2.Resize(size=(224, 224)),
        v2.ToDtype(torch.float32, scale=True),
        v2.Normalize(mean=[0.485, 0.456, 0.406], std=[0.229, 0.224, 0.225]),
    ]),
    transform_lib_type=TransformLibType.TORCHVISION,
    num_workers=2,
    sampler_config=SamplerConfig(init_args={...}),
)

test_subset_config = SubsetConfig(
    batch_size=64,
    subset_name="test",
    transforms=v2.Compose([
        v2.Resize(size=(224, 224)),
        v2.ToDtype(torch.float32, scale=True),
        v2.Normalize(mean=[0.485, 0.456, 0.406], std=[0.229, 0.224, 0.225]),
    ]),
    transform_lib_type=TransformLibType.TORCHVISION,
    num_workers=2,
)

datamodule = OTXDatamodule(
    task=...,
    data_root=...,
    train_subset=train_subset_config,
    val_subset=val_subset_config,
    test_subset=test_subset_config,
)
```

This level of boilerplate makes the API verbose and less user-friendly, especially for common use cases.

## Proposal

Proposing simplifying the API by allowing OTXDatamodule to be initialized with default configurations inferred from the task and data_root. The full configuration should only be necessary when overriding specific defaults.

Desired usage:

```
# Minimal usage
datamodule = OTXDatamodule(data_root=..., task=...)
```

Default transforms, batch size, workers, etc., would be automatically populated based on the task and dataset.

Optional overrides (after initialization):

```
datamodule.train_subset.transforms = v2.Compose([
    v2.Resize(size=(224, 224)),
    v2.RandomHorizontalFlip(p=0.5),
    SomeCustomTorchvisionTransform(),
    v2.ToDtype(torch.float32, scale=True),
    v2.Normalize(mean=[0.485, 0.456, 0.406], std=[0.229, 0.224, 0.225]),
])

datamodule.train_subset.batch_size = 8
```
Or overrides during initialization:

```
custom_transforms = v2.Compose([
    v2.Resize(size=(224, 224)),
    v2.RandomHorizontalFlip(p=0.5),
    SomeCustomTorchvisionTransform(),
    v2.ToDtype(torch.float32, scale=True),
    v2.Normalize(mean=[0.485, 0.456, 0.406], std=[0.229, 0.224, 0.225]),
])

datamodule = OTXDatamodule(
    task=...,
    data_root=...,
    train_subset={"batch_size": 8, "transforms": custom_transforms},
)
```
Internally, train_subset would be merged with the default configuration.

## Benefits

- Greatly improves usability and readability
- Reduces repetitive boilerplate
- Makes the API more approachable for newcomers
- Encourages consistent configuration defaults while supporting customization

### LINKED ISSUE #4781: Update datumaro to 1.12
In the `otx` package, upgrade the `datumaro` dependencies to the latest 1.12.* .

With respect to the currently [pinned](https://github.com/open-edge-platform/training_extensions/blob/56c3e2346ef6ddbe5f28e7f4e3638fac8cb4d36d/library/pyproject.toml#L30) version 1.10, the new Datumaro releases provide support for Python 3.13, remove several unnecessary dependencies and add some experimental dataset classes.

### LINKED ISSUE #4827: Support direct instantiation of OTXDataModule with Dataset
The [OTXDataModule](https://github.com/open-edge-platform/training_extensions/blob/c1b80845999cf47f9a2b6725caafea47c3dcc3dd/library/src/otx/data/module.py#L34) constructor requires a `data_format` and `data_root` parameters, therefore it assumes that the data is stored on disk. However, if the `Dataset` is already loaded in-memory, it should be possible to instantiate `OTXDataModule` by directly passing the dataset to it. Then the `DmDataset.import_from(...)` step at [L100](https://github.com/open-edge-platform/training_extensions/blob/c1b80845999cf47f9a2b6725caafea47c3dcc3dd/library/src/otx/data/module.py#L100) can be skipped.

kprokofi:
### Summary

Resolves: https://github.com/open-edge-platform/training_extensions/issues/4448, resolves https://github.com/open-edge-platform/training_extensions/issues/4827, resolves https://github.com/open-edge-platform/training_extensions/issues/4781

- [x] Add additional constructor method "from_otx_dataset" to enable possibility to instantiate DataModule with already constructed datasets (main reason of this PR)
- [x] Update Datumaro

- [x]  Remove redundant adaptive input size. We don't use this functionality and arbitrary input size can lead to training issues. 

- [x] Additionally addressed a problem in https://github.com/open-edge-platform/training_extensions/issues/4448. It is not ideal solution, since we should rewrite Data handling completely. IMO, we need to divide dataset and data loader creating, remove SubsetConfig entity 
        -  Remove mandatory SubsetConfig, instead constructing it from base yaml configs from recipe folder per task
        -  Add task parameter to each dataset to get this info in DataModule and remove mandatory parameter pass to     "from_otx_dataset" method

 
 New possible data creation:
 
```
 datamodule = OTXDatamodule(
    task=...,
    data_root=...,
)

```
It will create a default dataset with default augmentations and parameters from base data recipe

Advanced OTXDataModule construction:

```
train_subset_config = SubsetConfig(
    batch_size=64,
    transforms=v2.Compose([
        v2.RandomResizedCrop(size=(224, 224), antialias=True),
        v2.RandomHorizontalFlip(p=0.5),
        v2.ToDtype(torch.float32, scale=True),
        v2.Normalize(mean=[0.485, 0.456, 0.406], std=[0.229, 0.224, 0.225]),
    ]),
    num_workers=4,
)

 datamodule = OTXDatamodule(
    task=...,
    data_root=...,
    train_subset= train_subset_config
)
```

Since val pipeline is barely needed to be modified, we can pass only train_subset_config for training part. 

To control image_size for each pipeline we can add this parameter to SubsetConfig (train_subset_config, val_subset_config ...)

If you need to use the same input_size for all subsets: just pass a "input_size" parameter to datamodule

New way to create OTXDataModule:

```
# create datasets from datumaro dataset
dm_dataset = Datumaro.Dataset(data_root = ...)
train_dataset = OTXClassificationDataset(...)
val_dataset = OTXClassificationDataset(...)
test_dataset = OTXClassificationDataset(...)

# create simple SubsetConfig
train_subset = SubsetConfig(batch_size=64) # we don't need pass transforms again, we already constructed dataset

# construct datamodule
datamodule = OTXDatamodule.from_otx_dataset(
    train_dataset= train_dataset,
    val_dataset = val_dataset,
    test_dataset = test_dataset,
    train_subset= train_subset_config
)
```

### Checklist


library/pyproject.toml
```diff
@@ -27,7 +27,7 @@ classifiers = [
     "Programming Language :: Python :: 3.12",
 ]
 dependencies = [
-    "datumaro==1.10.0",
+    "datumaro==1.12.0",
```
leoll2:
```suggestion
    "datumaro~=1.12.0",
```
Let's keep flexibility on the patch version, if Datumaro releases any bugfix (e.g. 1.12.1)

library/pyproject.toml
```diff
@@ -51,6 +51,7 @@ dependencies = [
     "onnxconverter-common==1.14.0",
     "nncf==2.17.0",
     "anomalib[core]==1.1.3",
+    "numpy<2.0"
```
leoll2: We should upgrade numpy from 1.x to 2.x at some point, since 1.x is now quite old.
Thanks to ruff, the migration should be relatively straightforward: https://numpy.org/devdocs/numpy_2_0_migration_guide.html#ruff-plugin

library/src/otx/data/dataset/base.py
```diff
@@ -73,13 +75,15 @@ def __init__(
         self,
         dm_subset: DatasetSubset,
         transforms: Transforms,
+        task_type: OTXTaskType,
```
leoll2: Adding a new, non-optional parameter to a public class is technically an API-breaking change. Now, I don't know if Geti or any other framework instantiates this class directly, so I can't tell whether the change has impact on external projects. Just want to point it out.

library/src/otx/config/data.py
```diff
@@ -46,21 +43,18 @@ class SubsetConfig:
...
-
-    # TODO (vinnamki): Revisit data configuration objects to support a union type in structured config
-    # Omegaconf does not allow to have a union type, https://github.com/omry/omegaconf/issues/144
-    transforms: list[dict[str, Any]]
+    batch_size: int = 6
+    subset_name: str = "train"
+    transforms: list[dict[str, Any]] = field(default_factory=list)
     transform_lib_type: TransformLibType = TransformLibType.TORCHVISION
-    num_workers: int = 2
+    num_workers: int = 4
```
leoll2: What's the reason for this default value change?

library/src/otx/data/dataset/classification.py
```diff
@@ -13,18 +13,44 @@
...
+        self,
+        dm_subset: DatasetSubset,
+        transforms: Transforms,
+        task_type: OTXTaskType = OTXTaskType.MULTI_CLASS_CLS,
+        max_refetch: int = 1000,
+        image_color_channel: ImageColorChannel = ImageColorChannel.RGB,
+        stack_images: bool = True,
+        to_tv_image: bool = True,
+        data_format: str = "",
+    ) -> None:
```
leoll2, line 65: CC @AlbertvanHouten , you may need to backport these changes to your feature branch.

library/src/otx/data/dataset/segmentation.py
```diff
@@ -158,20 +158,22 @@ class OTXSegmentationDataset(OTXDataset):
 
     def __init__(
         self,
-        dm_subset: DmDataset,
+        dm_subset: DatasetSubset,
```
leoll2: What's the difference between `Dataset` and `DatasetSubset`? I see both implement the `IDataset` interface, but `DatasetSubset` doesn't have a docstring to explain its purpose ([source](https://github.com/open-edge-platform/datumaro/blob/4f3bed0245348abf49bd542283c4500a00846626/src/datumaro/components/dataset.py#L76)). CC @AlbertvanHouten

library/src/otx/data/module.py
```diff
@@ -199,6 +192,262 @@ def __init__(
...
+
+    @classmethod
+    def from_otx_datasets(
+        cls,
+        train_dataset: OTXDataset,
+        val_dataset: OTXDataset,
+        test_dataset: OTXDataset | None = None,
+        train_subset: SubsetConfig | None = None,
+        val_subset: SubsetConfig | None = None,
+        test_subset: SubsetConfig | None = None,
```
leoll2, line 232: I find the name of these three parameters a bit confusing. At first I thought that `train_subset` is the actual training subset data (i.e. all the dataset samples that belong to the training subset), in fact it refers to the subset configuration. Perhaps `train_subset_config` would be more clear.

leoll2 (commented):
Thanks for the changes Kirill, I left a few comments but LGTM. Please let Albert and some OTX maintainers review it too.

library/pyproject.toml
kprokofi: Current Anomalib doesn't allow us to upgrade, that why I set numpy<2.0

Let me look at this problem in other PR

library/src/otx/data/dataset/base.py
kprokofi: No it shouldn't break. This non-optional parameter in base dataset only. This base dataset serves as base class for our per task datasets. In child classes I made it optional

library/src/otx/config/data.py
kprokofi: No reason, just for more threads when preprocess dataset by default. I can revert it to 2. But for Geti this parameter is taken from config.

library/src/otx/data/dataset/segmentation.py
kprokofi: Don't know the exact difference TBH, I aligned it with types in other datasets.

library/src/otx/data/module.py
kprokofi: Agree, however, we need to rewrite all configs if we want to change this naming. I can change it only for "from_otx_dataset" method, but then we lose consistency with "init" constructor.

library/src/otx/config/data.py
kprokofi: reverted

library/src/otx/data/dataset/base.py
```diff
@@ -73,13 +75,15 @@ def __init__(
...
         max_refetch: int = 1000,
         image_color_channel: ImageColorChannel = ImageColorChannel.RGB,
         stack_images: bool = True,
         to_tv_image: bool = True,
         data_format: str = "",
     ) -> None:
         self.dm_subset = dm_subset
+        self.task_type = task_type
```
AlbertvanHouten: Why does the dataset need to know about its task type? Isn't it already clear by which dataset was initialized for which tasks it can be used? I know instance seg and rotated detection use the same dataset, but I still don't see the need to make it a part of the dataset. It is also not being used in this PR as far as I can see.

library/src/otx/data/dataset/segmentation.py
AlbertvanHouten: This should be Dataset. You are passing it from here https://github.com/open-edge-platform/training_extensions/pull/4853/files#diff-6423701d3ddfc89bd27817762495a1ed7c5ff19f58b13524b080b0482590b5cfR172

library/src/otx/data/dataset/base.py
kprokofi: The task is needed to properly construct configuration for SubsetConfigs during OTXDataModule instantiation. We may go other way: introduce property for the task, for the 100% know cases, jut return proper task. If it is IS, Rotated OD or anomaly, pass task parameter to constructor.

One more solution: pass a task to "from_otx_dataset" method. I wanted to avoid it, since it should be already clear from dataset we pass to that constructor method. However, for the base init we require a task parameter.

I can implement the 1st option, it will remove task_type in base class, introducing it only for IS dataset

library/src/otx/data/dataset/segmentation.py
kprokofi: Will change it then, thanks

library/src/otx/data/dataset/base.py
kprokofi: updated https://github.com/open-edge-platform/training_extensions/pull/4853/commits/3d48d250040aab594439ddf23321fb8394f9012b

library/src/otx/data/dataset/segmentation.py
kprokofi: changed https://github.com/open-edge-platform/training_extensions/pull/4853/commits/3d48d250040aab594439ddf23321fb8394f9012b

library/pyproject.toml
```diff
@@ -51,6 +51,8 @@ dependencies = [
     "onnxconverter-common==1.14.0",
     "nncf==2.17.0",
     "anomalib[core]==1.1.3",
+    "numpy<2.0",
+    "tensorboard==2.14.0"
```
leoll2: Is this dependency really necessary? I see that the engine can log in TensorBoard format via `TensorBoardLogger`, but that comes by default with Pytorch Lightning and it doesn't need to run tensorboard.

kprokofi: I added this requirement explicitly since removing it leads to an error:

image

leoll2: Looks like it could be made an extra dependency with minimal code changes to only instantiate `TensorBoardLogger` if the package is installed. Just an idea to minimize the dependencies, out of scope for this PR.

leoll2 (approved):

AlbertvanHouten (approved):

PR #4864, opened 2025-10-09
Implement Datumaro dataset converter

### LINKED ISSUE #4801: Implement Datumaro dataset converter
The goal is to implement a method `DatasetService.get_dm_dataset(project_id) -> dm.Dataset` that can be used to prepare the training dataset to fine-tune a model in Geti Tune.

The method should:
- query the `dataset_items` DB table to find the dataset item of a specific project
- for each record (dataset item), initialize a datumaro [Sample](https://github.com/open-edge-platform/datumaro/blob/d19fe1d4b530bf726a4f035752d8906ecd3e154c/src/datumaro/experimental/dataset.py#L38) with a reference to the image (`ImagePathField`) and the annotation data. Note that each task type requires a specialized sample type, e.g. `DetectionSample`.
- Create a datumaro [Dataset](https://github.com/open-edge-platform/datumaro/blob/d19fe1d4b530bf726a4f035752d8906ecd3e154c/src/datumaro/experimental/dataset.py#L190), append all the samples to it and return it.

warrkan:
### Summary

Resolves #4801 

This PR introduces datumaro_converter.py module, which can convert dataset to Datumaro format, which later will be used by OTX training

### Checklist


application/backend/app/services/dataset_service.py
```diff
@@ -310,3 +312,22 @@ def delete_dataset_item_annotations(self, project_id: UUID, dataset_item_id: UUI
...
+            return str(self.get_dataset_item_binary_path(project_id=project_id, dataset_item=item))
+
+        project_repo = ProjectRepository(db=self._db_session)
+        project = project_repo.get_by_id(str(project_id))
+        if project is None:
+            raise ResourceNotFoundError(ResourceType.PROJECT, str(project_id))
+
+        repo = DatasetItemRepository(project_id=str(project_id), db=self._db_session)
+        count = repo.count()
+        dataset_items = repo.list_items(offset=0, limit=count)
```
leoll2: Can you make the `limit` parameter of `list_items` optional, so that you can avoid the `repo.count()` step?

application/backend/app/services/dataset_service.py
```diff
@@ -310,3 +312,22 @@ def delete_dataset_item_annotations(self, project_id: UUID, dataset_item_id: UUI
         updated = repo.set_annotation_data(obj_id=str(dataset_item_id), annotation_data=[])
         if not updated:
             raise ResourceNotFoundError(ResourceType.DATASET_ITEM, str(dataset_item_id))
+
+    def get_dataset_item_binary_path(self, project_id: UUID, dataset_item: DatasetItemDB) -> Path:
+        dataset_dir = self.projects_dir / f"{project_id}/dataset"
+        return dataset_dir / f"{dataset_item.id}.{dataset_item.format}"
```
leoll2: We already have a similar function `DatasetService.get_dataset_item_binary_path_by_id`, which instead takes `dataset_item_id` as input and loads the item from the DB.

To avoid unnecessary duplication, I propose to refactor them as follows:
```python
def get_dataset_item_binary_path(self, project_id: UUID, dataset_item: DatasetItemDB) -> Path:
        dataset_dir = self.projects_dir / f"{project_id}/dataset"
        return dataset_dir / f"{dataset_item.id}.{dataset_item.format}"

def get_dataset_item_binary_path_by_id(self, project_id: UUID, dataset_item_id: UUID) -> Path:
    """Get a dataset item binary content by its ID"""
    repo = DatasetItemRepository(project_id=str(project_id), db=self._db_session)
    dataset_item = repo.get_by_id(str(dataset_item_id))
    if not dataset_item:
        raise ResourceNotFoundError(ResourceType.DATASET_ITEM, str(dataset_item_id))
    return self.get_dataset_item_binary_path(project_id, dataset_item)
```

application/backend/app/services/datumaro_converter.py
```diff
@@ -0,0 +1,217 @@
...
+from datumaro.experimental.categories import LabelCategories
+from datumaro.experimental.fields import polygon_field
+
+from app.db.schema import DatasetItemDB, LabelDB, ProjectDB
+from app.schemas.dataset_item import DatasetItemAnnotation
+from app.schemas.project import TaskType
+from app.schemas.shape import Polygon, Rectangle
+
+
+class DetectionSample(Sample):
```
leoll2, line 24: After the `feature/datumaro` branch is merged, we can take these classes from https://github.com/open-edge-platform/training_extensions/blob/feature/datumaro/library/src/otx/data/entity/sample.py right?

application/backend/app/services/datumaro_converter.py
```diff
@@ -0,0 +1,217 @@
...
+    polygons: np.ndarray[Any, Any] = polygon_field(dtype=pl.Float32)
+    label: np.ndarray[Any, Any] = label_field(dtype=pl.Int32, is_list=True)
+
+
+def convert_rectangle(r: Rectangle) -> list[int]:
+    """
+    Convert Rectangle shape coordinates to a list
+
+    :param r: Rectangle shape
+    :return: list of coordinates in x1y1x2y2 format
```
leoll2: We recently decided to adopt Google-style docstrings, please update these to the new syntax.

https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

application/backend/app/services/datumaro_converter.py
leoll2, line 1: The file is called `services/datumaro_converter.py`, but there's no service class in it. What do you think about defining a class `DatumaroConversionService` and moving all the functions under its scope? A standalone function named `convert_dataset` is not self-describing, while `DatumaroConversionService.convert_dataset` would be more clear imho.

application/backend/app/services/dataset_service.py
```diff
@@ -310,3 +312,22 @@ def delete_dataset_item_annotations(self, project_id: UUID, dataset_item_id: UUI
...
+        project_repo = ProjectRepository(db=self._db_session)
+        project = project_repo.get_by_id(str(project_id))
+        if project is None:
+            raise ResourceNotFoundError(ResourceType.PROJECT, str(project_id))
+
+        repo = DatasetItemRepository(project_id=str(project_id), db=self._db_session)
+        count = repo.count()
+        dataset_items = repo.list_items(offset=0, limit=count)
+
+        return convert_dataset(project=project, dataset_items=dataset_items, get_image_path=_get_image_path)
```
leoll2: This implementation requires loading all dataset item DB records in-memory before constructing the datumaro dataset. In practice, if the dataset has N items, you have 2*N objects loaded in memory at the same time (one `DatasetItemDB` and one `Sample` per item). A more memory-efficient implementation, that only keeps N objects in-memory at any time, would be to load and convert the DB items in small batches, and finally merge the partial results.
It's a bit more complex, but it can potentially save a lot of memory for larger datasets. What do you think? Maybe I can create a separate ticket to address it later.

application/backend/app/services/datumaro_converter.py
```diff
@@ -0,0 +1,217 @@
...
+    for dataset_item in dataset_items:
+        image_path = get_image_path(dataset_item)
+        annotations = [DatasetItemAnnotation.model_validate(annotation) for annotation in dataset_item.annotation_data]
+        coords = [
+            convert_rectangle(annotation.shape)
+            for annotation in annotations
+            if (isinstance(annotation.shape, Rectangle))
+        ]
+        labels_indexes = [
+            project_labels_ids.index(annotation.labels[0].id)
```
leoll2: I suggest to handle the exception of label not found, logging a warning/error and skipping this dataset item.
Currently, such an exception could be raised following the removal of a label, because we haven't implemented any logic to filter out the removed label from the DB annotations.

Same for classification and segmentation.

leoll2 (commented):
The approach LGTM!

leoll2:
> ## Docker Image Sizes
> Image	Size
> geti-tune-pr-4864	3.1G
> geti-tune-sha-0217063	3.1G

Wow, the size increased more than I expected (1.2G -> 3.1G). Can you measure how much of the increase is attributable to the new system dependencies (git, build-essentials, curl, rust) and how much to datumaro?

The easiest way to measure it is probably to disable the datumaro dependency, build and compare.

warrkan:
> > ## Docker Image Sizes
> > Image	Size
> > geti-tune-pr-4864	3.1G
> > geti-tune-sha-0217063	3.1G
> 
> Wow, the size increased more than I expected (1.2G -> 3.1G). Can you measure how much of the increase is attributable to the new system dependencies (git, build-essentials, curl, rust) and how much to datumaro?
> 
> The easiest way to measure it is probably to disable the datumaro dependency, build and compare.

This is not optimized Docker image indeed, it has Rust build toolchain in the layer. I'm re-working it at the moment

application/backend/app/services/dataset_service.py
warrkan: Let's keep it for now, I've added batching instead.

application/backend/app/services/dataset_service.py
warrkan: Agree, done. Thanks for spotting

application/backend/app/services/datumaro_converter.py
warrkan: Not really, we can keep our samples definition and Datumaro will perform the conversion to OTX internally on the fly

application/backend/app/services/datumaro_converter.py
warrkan: Indeed, changed the docstrings

application/backend/app/services/datumaro_converter.py
warrkan: To be honest, I don't think that every class in `service` module should end with `Service` suffix. There are also nested modules where it already doesn't work. I think it's more about responsibility rather than naming

application/backend/app/services/dataset_service.py
warrkan: Added batching here, thanks for the idea

application/backend/app/services/datumaro_converter.py
warrkan: Indeed! Thanks, added exception handling

leoll2 (dismissed):

leoll2 (approved):

PR #4873, opened 2025-10-10
Training Configurations Add/Update implementation

### LINKED ISSUE #4799: Implement the training configuration endpoints

A-Artemis:
### Summary

Add the implementation for GET and POST of the training configuration

### How to test

http://geti-tune.localhost/api/projects/165f79f9-fba5-415b-adda-7e6548ed2c40/training_configuration

### Checklist


application/backend/app/repositories/training_configuration_repo.py
```diff
@@ -0,0 +1,69 @@
...
+    ) -> TrainingConfigurationDB | None:
+        """
+        Get training configuration by project ID and optional model architecture ID.
+
+        Args:
+            project_id (str): The ID of the project.
+            model_architecture_id (str | None): The ID of the model architecture.
+        """
+        return (
+            self.db.query(TrainingConfigurationDB)
```
itallix: Let's use SQLAlchemy 2.0 API, they marked `session.query` as [deprecated](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session.query) in 2.x

See other repos as an example

application/backend/tests/integration/services/test_training_configuration_service.py
```diff
@@ -0,0 +1,195 @@
...
+        )
+        db_session.add(project)
+        training_configuration = TrainingConfigurationDB(
+            id=str(uuid4()),
+            project_id=project.id,
+            model_architecture_id="Custom_Object_Detection_YOLOX",
+            configuration_data=fxt_training_configuration.model_dump(),
+        )
+        db_session.add(training_configuration)
+        db_session.commit()
```
itallix: There is no need to commit db changes in integration tests since we rollback changes after each test, `flush` is enough. Otherwise, they will persist and will be visible in other tests.

application/backend/tests/integration/services/test_training_configuration_service.py
```diff
@@ -0,0 +1,195 @@
...
```
itallix:
```suggestion
        db_session.flush()
```

application/backend/app/repositories/training_configuration_repo.py
```diff
@@ -0,0 +1,69 @@
...
+            project_id (str): The ID of the project.
+            model_architecture_id (str | None): The ID of the model architecture.
+            configuration_data (dict): The configuration data to store.
+
+        Returns:
+            TrainingConfigurationDB: The created or updated training configuration.
+        """
+        existing = self.get_by_project_and_model_architecture(project_id, model_architecture_id)
+
+        if existing:
```
itallix, line 58: Do we need `existing` check - will it not be able to detect whether to generate `update/insert` based on provided ids by default?

application/backend/app/services/training_configuration_service.py
```diff
@@ -0,0 +1,116 @@
...
+        project = ProjectRepository(self._db_session).get_by_id(str(project_id))
+        if not project:
+            raise ResourceNotFoundError(ResourceType.PROJECT, str(project_id))
+
+        current_config = self.get_training_configuration(
+            project_id=project_id,
+            model_architecture_id=model_architecture_id,
+        )
+        current_data = current_config.model_dump()
+        self._deep_update_dict(base_dict=current_data, updates=training_config_update)
```
itallix: Can we replace this with pydantic methods, e.g.: [`model_copy(update=obj, deep=True)`](https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_copy)?

application/backend/app/db/schema.py
```diff
@@ -122,3 +122,14 @@ class LabelDB(BaseID):
     hotkey: Mapped[str | None] = mapped_column(String(10), nullable=True)
 
     project = relationship("ProjectDB", back_populates="labels")
+
+
+class TrainingConfigurationDB(BaseID):
+    __tablename__ = "training_configurations"
```
leoll2, line 140: Could you add this table to `docs/diagrams/storage/full.puml`? If you use PyCharm, the diagram can be rendered by the plantuml4idea plugin.

application/backend/app/alembic/versions/da385d690aae_schema.py
```diff
@@ -152,4 +165,5 @@ def downgrade() -> None:
     op.drop_table("sources")
     op.drop_table("sinks")
     op.drop_table("projects")
+    op.drop_table("training_configurations")
```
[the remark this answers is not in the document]
leoll2: I think this comment is valid: it's better to delete tables in reverse order wrt creation.

application/backend/app/schemas/training_configuration.py
```diff
@@ -27,20 +27,62 @@ class TrainingConfiguration(BaseModel):
...
+
     model_config = {
         "json_schema_extra": {
             "example": {
-                "dataset_preparation": {"config": {"train_val_split_ratio": 0.8, "augmentation_enabled": True}},
-                "training": {"config": {"epochs": 100, "batch_size": 32, "learning_rate": 0.001}},
-                "evaluation": {"config": {"metrics": ["accuracy", "precision", "recall"], "validation_split": 0.2}},
+                "dataset_preparation": {
+                    "augmentation": {
+                        "topdown_affine": None,
```
leoll2: Null-valued parameters should be filtered out from the response, because in this context `None` represents a parameter that's not applicable to the algorithm.

I believe Geti already implements a similar feature. This is an example of response, as you can see there are some empty sections (-> this algorithm doesn't support any augmentation)

image

application/backend/app/schemas/training_configuration.py
```diff
@@ -27,20 +27,62 @@ class TrainingConfiguration(BaseModel):
...
+                            "saturation": [0.5, 1.5],
+                            "hue": [-0.05, 0.05],
+                            "probability": 0.5,
+                        },
+                        "gaussian_blur": {"enable": False, "kernel_size": 5, "sigma": [0.1, 2], "probability": 0.5},
+                        "photometric_distort": None,
+                        "gaussian_noise": {"enable": False, "mean": 0, "sigma": 0.1, "probability": 0.5},
+                        "tiling": {"enable": False, "adaptive_tiling": True, "tile_size": 400, "tile_overlap": 0.2},
+                    }
+                },
```
leoll2: I think the `subset_split` category is missing?

application/backend/app/schemas/training_configuration.py
```diff
@@ -27,20 +27,62 @@ class TrainingConfiguration(BaseModel):
...
+                "training": {
+                    "max_epochs": 200,
```
leoll2: The REST endpoint should return a richer representation of each parameter, including the name, description, type, min/max and default values, etc... These details are necessary to the UI to correctly display the parameter.

For reference, here is how the `max_epochs` parameter looks like in Geti response:

image

application/backend/app/alembic/versions/2786b50eb5a4_schema.py
```diff
@@ -151,5 +164,6 @@ def downgrade() -> None:
     op.drop_table("dataset_revisions")
     op.drop_table("sources")
     op.drop_table("sinks")
+    op.drop_table("training_configurations")
```
leoll2, line 193: Move to L160 (begin of list), that's where Alembic would put this line if this code was auto-generated (I assume you edited it manually).

application/backend/app/api/routers/training_configurations.py
```diff
@@ -0,0 +1,98 @@
...
+    Note: model_architecture_id and model_revision_id cannot be used together.
+
+    Args:
+        training_configuration_service (TrainingConfigurationService): The training configuration service.
+        project_id (UUID): The unique identifier of the project.
+        model_architecture_id (Optional[str]): The model architecture ID for specific configuration retrieval.
+        model_revision_id (Optional[UUID]): The model revision ID for specific configuration retrieval.
+
+    Returns:
+        TrainingConfiguration: The training configuration details.
```
leoll2:
```suggestion
```
Remove this Args/Returns section, endpoint parameters and response should be documented through pydantic.

FYI this docstring shows in the API docs

image

application/backend/app/api/routers/training_configurations.py
```diff
@@ -0,0 +1,98 @@
...
+logger = logging.getLogger(__name__)
+router = APIRouter(prefix="/api/projects/{project_id}/training_configuration", tags=["Training Configuration"])
+
+
+@router.get("")
+def get_training_configuration(
+    training_configuration_service: Annotated[
+        TrainingConfigurationService, Depends(get_training_configuration_service)
+    ],
+    project_id: UUID,
```
leoll2:
```suggestion
    project_id: Annotated[UUID, Depends(get_project_id)],
```
Use dependencies to make sure the parameter is properly validated.

application/backend/app/api/routers/training_configurations.py
```diff
@@ -0,0 +1,98 @@
...
+    model_architecture_id: str | None = None,
+    model_revision_id: UUID | None = None,
+) -> dict:
```
leoll2, line 26: Should return a `TrainingConfiguration` or similar pydantic model. This is important to ensure that the response schema is well-defined in the OpenAPI spec. The schema is necessary for the UI to validate the response and extract the attributes they need; conversely, a `dict` is very generic and carries no information about its internal structure.

image

image

application/backend/app/api/routers/training_configurations.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+@router.get("")
+def get_training_configuration(
+    training_configuration_service: Annotated[
+        TrainingConfigurationService, Depends(get_training_configuration_service)
+    ],
+    project_id: UUID,
+    model_architecture_id: str | None = None,
+    model_revision_id: UUID | None = None,
```
leoll2: Use a dependency to validate the ID. I suggest to reuse the `get_model_id` implementation as follows:

```python
def get_model_revision_id(model_revision_id: str) -> UUID:
    return get_model_id(model_id=model_revision_id)

File: application/backend/app/api/routers/training_configurations.py [hunk 16]
```diff
@@ -0,0 +1,98 @@
...
+router = APIRouter(prefix="/api/projects/{project_id}/training_configuration", tags=["Training Configuration"])
+
+
+@router.get("")
+def get_training_configuration(
+    training_configuration_service: Annotated[
+        TrainingConfigurationService, Depends(get_training_configuration_service)
+    ],
+    project_id: UUID,
+    model_architecture_id: str | None = None,
```
leoll2: This string should be validated before it's used to query the DB. You can define a dependency `get_model_architecture_id` that checks if a manifest exists with the given id (`SupportedModels.get_model_manifest_by_id`), else returns 400.

application/backend/app/api/routers/training_configurations.py
```diff
@@ -0,0 +1,98 @@
...
+
+@router.patch("", status_code=status.HTTP_204_NO_CONTENT)
+def update_training_configuration(
+    training_configuration_service: Annotated[
+        TrainingConfigurationService, Depends(get_training_configuration_service)
+    ],
+    project_id: UUID,
+    training_config_update: dict,
+    model_architecture_id: str | None = None,
+) -> None:
```
leoll2: Maybe we should return 200 + the updated configuration, for convenience. Can be in a separate PR.

application/backend/app/api/routers/training_configurations.py
```diff
@@ -0,0 +1,98 @@
...
+      "training": {...},
+      "evaluation": {...}
+    }
+    ```
+
+    Args:
+        training_configuration_service (TrainingConfigurationService): The training configuration service.
+        project_id (UUID): The unique identifier of the project.
+        training_config_update (dict): The configuration updates to apply.
+        model_architecture_id (Optional[str]): The model architecture ID for specific configuration update.
```
leoll2: Remove (see the other comment)

application/backend/app/api/routers/training_configurations.py
```diff
@@ -0,0 +1,98 @@
...
+    Note: model_architecture_id cannot be used with model_revision_id for updates.
+
+    Request body should contain elements of the configuration hierarchy to update:
+    ```json
+    {
+      "dataset_augmentation_parameters": {...},
+      "training": {...},
+      "evaluation": {...}
+    }
+    ```
```
leoll2: Should not be here, but in the `Body.description`. Check how it's done in the `update_pipeline` endpoint.

application/backend/app/supported_models/hyperparameters/base_model_no_extra.py
```diff
@@ -1,11 +0,0 @@
...
-# SPDX-License-Identifier: Apache-2.0
-
-
-from pydantic import BaseModel, ConfigDict
-
-
-class BaseModelNoExtra(BaseModel):
-    """Base model class that does not allow for extra fields."""
-
-    model_config = ConfigDict(extra="forbid")
```
leoll2, line 11: If I'm not wrong, Geti uses `BaseModelNoExtra` to validate that the configuration elements do not contain unrecognized keys. For example, if the UI/client provides a parameter key with a typo, it's better to be strict and raise an error instead of silently ignoring that parameter.

Why are you removing this base model? Did you find a different solution to address the above problem?

application/backend/app/repositories/training_configuration_repo.py
```diff
@@ -0,0 +1,68 @@
...
+    ) -> TrainingConfigurationDB:
+        """
+        Create or update a training configuration.
+
+        If a configuration for the given project and model architecture exists, it is updated.
+        Otherwise, a new configuration is created.
+
+        Args:
+            project_id (str): The ID of the project.
+            model_architecture_id (str | None): The ID of the model architecture.
```
leoll2, line 47:
```suggestion
            model_architecture_id (str | None): The ID of the model architecture.
                If None, the operation targets the project-level configuration.
```

application/backend/app/services/training_configuration_service.py
```diff
@@ -0,0 +1,158 @@
+# Copyright (C) 2025 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+from uuid import UUID
+
+from sqlalchemy.orm import Session
+
+from app.configuration_tools.training_configuration import TrainingConfiguration
+from app.repositories import ModelRevisionRepository, ProjectRepository
+from app.repositories.training_configuration_repo import TrainingConfigurationRepository
+from app.services import ResourceNotFoundError, ResourceType
```
leoll2, line 10:
```suggestion
from .base import ResourceNotFoundError, ResourceType
```
Prefer relative imports when at the same level, to prevent circular imports between services.

application/backend/app/configuration_tools/parameters/hyperparameters.py
```diff
@@ -115,21 +105,21 @@ def validate_input_size(self) -> "TrainingHyperParameters":
...
 
-class Hyperparameters(BaseModelNoExtra):
+class Hyperparameters(BaseModel):
     """Complete set of configurable parameters for model training and evaluation."""
 
-    dataset_preparation: DatasetPreparationParameters = Field(
-        default_factory=DatasetPreparationParameters,
-        title="Dataset preparation",
-        description="Parameters for preparing the dataset before training",
+    dataset_augmentation_parameters: DatasetAugmentationParameters = Field(
```
leoll2: Please revert to the previous name `dataset_preparation`. It's better for 3 reasons:
- Consistent API schema with Geti Classic -> easier to reuse UI components
- Consistent with the name of the parameter section displayed in the UI
- It may include other dataset-related parameters that are not augmentations (e.g. filtering or subset split)

application/backend/app/services/training_configuration_service.py
```diff
@@ -0,0 +1,158 @@
...
+    def __init__(self, db_session: Session) -> None:
+        self._db_session = db_session
+        self._training_config_repo = TrainingConfigurationRepository(db_session)
+
+    def get_training_configuration(
+        self,
+        project_id: UUID,
+        model_architecture_id: str | None = None,
+        model_revision_id: UUID | None = None,
+    ) -> TrainingConfiguration:
```
leoll2, line 26: I noticed that the endpoint response contains the field `"model_manifest_id"`. I'd suggest to remove it because it's either not applicable (like in my screenshot, where the endpoint is called without specifying any model) or redundant (if the model is already in the request).

image

application/backend/app/api/routers/training_configurations.py
```diff
@@ -0,0 +1,98 @@
...
+from app.configuration_tools.training_configuration_converter import convert_training_configuration_to_rest
+from app.services import ResourceNotFoundError
+from app.services.training_configuration_service import TrainingConfigurationService
+
+logger = logging.getLogger(__name__)
+router = APIRouter(prefix="/api/projects/{project_id}/training_configuration", tags=["Training Configuration"])
+
+
+@router.get("")
+def get_training_configuration(
```
leoll2, line 19: This response is malformed, because all training parameters are flattened under "training".

image

See how it differs from the response in Geti Classic:

image

Here "enable" and "patience" are correctly nested under "early_stopping".

application/backend/app/api/routers/training_configurations.py
A-Artemis: This will need to be addressed in a separate PR. Currently, the `TrainingConfiguration` is nothing like the desired `dict` output, this will need to be revisited.

application/backend/app/alembic/versions/2786b50eb5a4_schema.py
```diff
@@ -156,6 +156,19 @@ def upgrade() -> None:
         sa.ForeignKeyConstraint(["label_id"], ["labels.id"], ondelete="CASCADE"),
         sa.PrimaryKeyConstraint("dataset_item_id", "label_id"),
     )
+    op.create_table(
```
leoll2, line 159: It looks like the file was manually updated, please use `uv run alembic -c app/alembic.ini revision --autogenerate -m "schema"` instead

application/backend/app/api/routers/training_configurations.py
```diff
@@ -0,0 +1,80 @@
...
+    training_configuration_service: Annotated[
+        TrainingConfigurationService, Depends(get_training_configuration_service)
+    ],
+    project_id: Annotated[UUID, Depends(get_project_id)],
+    model_architecture_id: Annotated[str, Depends(get_model_architecture_id)],
+    model_revision_id: Annotated[UUID, Depends(get_model_revision_id)],
```
leoll2: These are marked as mandatory, but should be optional.

application/backend/app/api/routers/training_configurations.py
```diff
@@ -0,0 +1,80 @@
...
+    project_id: Annotated[UUID, Depends(get_project_id)],
+    model_architecture_id: Annotated[str, Depends(get_model_architecture_id)],
+    training_config_update: dict,
+) -> dict:
+    """
+    Update the training configuration for a project.
+
+    - If model_architecture_id is provided, updates configuration for that specific model architecture.
+    - If not provided, updates the general task-related configuration.
+    Note: model_architecture_id cannot be used with model_revision_id for updates.
```
leoll2: This comment is invalid, the PATCH endpoint does not have a "model_revision_id" parameter.

application/backend/app/api/routers/training_configurations.py
```diff
@@ -0,0 +1,80 @@
...
+        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(e))
+
+
+@router.patch("", status_code=status.HTTP_204_NO_CONTENT)
+def update_training_configuration(
+    training_configuration_service: Annotated[
+        TrainingConfigurationService, Depends(get_training_configuration_service)
+    ],
+    project_id: Annotated[UUID, Depends(get_project_id)],
+    model_architecture_id: Annotated[str, Depends(get_model_architecture_id)],
```
leoll2: Should be optional

application/backend/app/models/training_configuration/__init__.py
leoll2, line 1: In line with Evgeny's refactoring of entities, I suggest to move these classes under `app/models`:

```
app/models/training_configuration/
├─ augmentation.py
├─ hyperparameters.py
├─ configuration.py   # your 'training_configuration.py'
```

application/backend/app/configuration_tools/hyperparameters/base_model_no_extra.py
leoll2, line 1: Merge into `app/models/base.py`

application/backend/app/configuration_tools/training_configuration.py
```diff
@@ -0,0 +1,181 @@
...
+
+class MinAnnotationPixels(BaseModel):
+    """Parameters for minimum annotation pixels."""
+
+    enable: bool = Field(
+        default=False,
+        title="Enable minimum annotation pixels filtering",
+        description="Whether to apply minimum annotation pixels filtering",
+    )
+    min_annotation_pixels: int = Field(
```
leoll2:
```suggestion
class MinAnnotationPixels(BaseModel):
    """Parameters for minimum annotation pixels."""

    enable: bool = Field(
        default=False,
        title="Enable minimum annotation pixels filtering",
        description="Whether to apply minimum annotation pixels filtering",
    )
    value: int = Field(
```
`Filtering` already has a property `min_annotation_pixels` which maps to this class. Therefore we want this attribute to resolve as `filtering.min_annotation_pixels.value` instead of `filtering.min_annotation_pixels.min_annotation_pixels`.

application/backend/app/configuration_tools/training_configuration.py
```diff
@@ -0,0 +1,181 @@
...
+
+class MaxAnnotationPixels(BaseModel):
+    """Parameters for maximum annotation pixels."""
+
+    enable: bool = Field(
+        default=False,
+        title="Enable maximum annotation pixels filtering",
+        description="Whether to apply maximum annotation pixels filtering",
+    )
+    max_annotation_pixels: int = Field(
```
leoll2:
```suggestion
    value: int = Field(
```

application/backend/app/configuration_tools/training_configuration.py
```diff
@@ -0,0 +1,181 @@
...
+
+class MinAnnotationObjects(BaseModel):
+    """Parameters for maximum annotation objects."""
+
+    enable: bool = Field(
+        default=False,
+        title="Enable minimum annotation objects filtering",
+        description="Whether to apply minimum annotation objects filtering",
+    )
+    min_annotation_objects: int = Field(
```
leoll2:
```suggestion
    value: int = Field(
```

application/backend/app/configuration_tools/training_configuration.py
```diff
@@ -0,0 +1,181 @@
...
+
+class MaxAnnotationObjects(BaseModel):
+    """Parameters for maximum annotation objects."""
+
+    enable: bool = Field(
+        default=False,
+        title="Enable maximum annotation objects filtering",
+        description="Whether to apply maximum annotation objects filtering",
+    )
+    max_annotation_objects: int = Field(
```
leoll2:
```suggestion
    value: int = Field(
```

application/backend/app/services/tools/configuration_overlay.py
leoll2, line 1: In line with the recent code refactoring, I suggest to move this file to `services`.

- `app/configuration_tools/configurable_parameters_converter.py` -> `app/services/configuration_overlay.py`

application/backend/app/api/serializers/configurable_parameters.py
leoll2, line 1: This file contains mappers for the REST API schema. Therefore I suggest to move it under `app/api`:

- `app/configuration_tools/configurable_parameters_converter.py` -> `app/api/serializers/configurable_parameters.py`

application/backend/app/api/serializers/training_configuration.py
leoll2, line 1: Would also move under `app/api`:

`app/configuration_tools/training_configuration_converter.py` -> `app/api/serializers/training_configuration.py`

application/backend/app/alembic/versions/2786b50eb5a4_schema.py
A-Artemis: I believe it is correct. I ran the command and there was no change in the generated file.

application/backend/app/api/routers/training_configurations.py
```diff
@@ -0,0 +1,79 @@
...
+    get_model_architecture_id,
+    get_model_revision_id,
+    get_project_id,
+    get_training_configuration_service,
+)
+from app.api.serializers.training_configuration import TrainingConfigurationConverter
+from app.services import ResourceNotFoundError
+from app.services.training_configuration_service import TrainingConfigurationService
+
+logger = logging.getLogger(__name__)
```
leoll2: #4991 is merged, please replace with `from loguru import logger` after merging latest develop.

application/backend/app/services/tools/configuration_overlay.py
```diff
@@ -0,0 +1,118 @@
+# Copyright (C) 2025 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+from copy import deepcopy
+
+from app.models.training_configuration.configuration import PartialTrainingConfiguration, TrainingConfiguration
+
+
+class ConfigurationOverlayTools:
```
leoll2, line 9: Is this class used anywhere?

application/backend/app/models/training_configuration/configuration.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+class TrainingConfiguration(BaseModel):
+    """Configuration for model training"""
+
+    model_manifest_id: str | None = Field(
+        default=None,
+        title="Model manifest ID",
+        description="ID for the model manifest that defines the supported parameters and capabilities for training",
+    )
```
leoll2, line 157: No need to include `model_manifest_id` in the training configuration

application/backend/tests/unit/supported_models/test_model_manifest.py
```diff
@@ -125,8 +123,8 @@ def test_relative_path_parsing(self):
             },
             "support_status": "active",
             "supported_gpus": {"intel": True},
-            "hyperparameters": {
-                "dataset_preparation": {
+            "parameters": {
+                "dataset_augmentation_parameters": {
```
leoll2: This change should be reverted, right?

application/backend/tests/unit/routers/test_training_configurations.py
```diff
@@ -0,0 +1,56 @@
...
+    ):
+        """Test successful retrieval of training configuration."""
+        project_id = uuid4()
+        fxt_training_configuration_service.get_training_configuration.return_value = fxt_training_configuration
+
+        response = fxt_client.get(f"/api/projects/{project_id}/training_configuration")
+
+        assert response.status_code == 200
+        data = response.json()
+        assert "dataset_augmentation_parameters" in data
```
leoll2: Should be "dataset_preparation"

application/backend/app/services/tools/configuration_overlay.py
A-Artemis: Not yet, it will be used in the Training Configuration Update logic.

A-Artemis: It is now being used.

application/backend/tests/unit/configuration_tools/test_configuration_overlay_tools.py
```diff
@@ -0,0 +1,308 @@
...
+    def test_merge_deep_dict_common_parameters_only(self, a, b, expected) -> None:
+        result = ConfigurationOverlayTools.merge_deep_dict(a, b, common_parameters_only=True)
+        assert result == expected
+        # Ensure the function modifies the first dict in-place
+        assert result is a
+
+    ###########################
+    # Commented out tests for overlaying configurations, as they may depend on specific implementation details
+    ###########################
+    # def test_overlay_configurations(self, fxt_training_configuration_task_level) -> None:
```
leoll2: Should be uncommented?

application/backend/app/services/training_configuration_service.py
```diff
@@ -0,0 +1,161 @@
...
+        project = ProjectRepository(self._db_session).get_by_id(str(project_id))
+        if not project:
+            raise ResourceNotFoundError(ResourceType.PROJECT, str(project_id))
+
+        current_config = self.get_training_configuration(
+            project_id=project_id,
+            model_architecture_id=model_architecture_id,
+        )
+
+        converted_config = ConfigurableParametersConverter.configurable_parameters_from_rest(training_config_update)
```
leoll2: `ConfigurableParametersConverter` belongs to the API layer (`app.api`), ideally the services shouldn't depend on it. You can easily remove this inverse dependency by moving this line to the endpoint (optional: wrap it in a fastapi dependency) and pass the already `converted_config` to this function `TrainingConfigurationService.update_training_configuration`.

application/backend/app/models/training_configuration/configuration.py
leoll2: ^ this comment was marked as resolved but is still valid

leoll2 (commented):
LGTM, thanks for the changes! Only a few small comments left before it can be merged.

leoll2 (approved):

application/backend/app/models/training_configuration/configuration.py
A-Artemis: For now, I would like to keep the model_manifest_id as part of the `TrainingConfiguration` until the final correct pydantic format is decided.

application/backend/app/repositories/training_configuration_repo.py
A-Artemis: No, not in its current implementation. We allow it so that the user can create a new config with given IDs. 
If project/model_architecture ID is given, it is not guaranteed that that a document already exists.

PR #4875, opened 2025-10-10
E2E infra

### LINKED ISSUE #4850: E2E infrastructure

jpggvilaca:
### Summary

* Update `run.sh`
  - Add `DATABASE_FILE` as a param allowing us a to have different DB's, one for e2e tests and other for development
  - Improved checked and added more explanations
* Update `docker-compose` to include E2E steps
  - Add backend/frontend-e2e service + playwright service
  - We can now run e2e tests locally with or without docker
* Update GHA actions to leverage our docker-compose
* Update playwright config
* Add README's

### Checklist

### DIFF
--- a/.github/workflows/ui-lint-and-test.yaml
+++ b/.github/workflows/ui-lint-and-test.yaml
@@ -126,6 +126,8 @@ jobs:
 
       - name: Build UI
         working-directory: "application/ui"
+        env:
+          PUBLIC_API_BASE_URL: ""
         run: npm run build
 
       - name: Compress build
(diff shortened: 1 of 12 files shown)


.github/workflows/ui-lint-and-test.yaml
```diff
@@ -126,6 +126,8 @@ jobs:
 
       - name: Build UI
         working-directory: "application/ui"
+        env:
+          PUBLIC_API_BASE_URL: ""
```
jpggvilaca, line 130: to avoid setting `http://localhost:7680/api/...` to our baseUrl, so the UI use relative urls

.github/workflows/ui-lint-and-test.yaml
```diff
@@ -353,84 +355,64 @@ jobs:
...
+      - name: Install dependencies
+        working-directory: application/ui
+        run: npm ci
+
+      - name: Run E2E tests with Docker Compose
+        working-directory: application/docker
+        env:
+          E2E_ASSETS_S3_URL: ${{ vars.E2E_ASSETS_S3_URL }}
+        run: |
+          docker compose --profile e2e up \
```
jpggvilaca, line 394: the important change is this one.

application/.dockerignore
```diff
@@ -2,6 +2,7 @@
 
 # Backend files
 !backend/app
+!backend/run.sh
```
jpggvilaca, line 5: because we need to copy this bad boy now

application/docker/docker-compose.yaml
```diff
@@ -49,3 +50,89 @@ services:
       - mosquitto
     ports:
       - "8081:80"
+
+  backend-e2e:
```
jpggvilaca, line 54: not sure if this is overkill

application/docker/nginx-e2e.conf
```diff
@@ -0,0 +1,46 @@
+server {
```
jpggvilaca, line 1: Not sure if we need a second nginx but the flow goes like this:
* build:e2e builds UI with PUBLIC_API_BASE_URL=""
* Docker Compose starts
* backend-e2e: Runs run.sh, seeds DB, starts FastAPI on port 7860
frontend-e2e: Nginx serves dist/, proxies /api/* to backend
playwright-e2e: Runs tests against http://frontend-e2e:3000
Tests make API calls: http://frontend-e2e:3000/api/projects → nginx proxies → http://backend-e2e:7860/api/projects

application/ui/package.json
```diff
@@ -9,6 +9,7 @@
     },
     "scripts": {
         "build": "rsbuild build",
+        "build:e2e": "PUBLIC_API_BASE_URL='' rsbuild build",
```
jpggvilaca, line 12: running the build here and not on docker-compose avoids having to install aaaaaaaalllll node dependencies to just build the UI

application/ui/tests/e2e/existing-project.spec.ts
```diff
@@ -3,7 +3,7 @@
 
 import { expect, test } from '@playwright/test';
 
-test('[E2E] Existing project', async ({ page }) => {
+test.skip('[E2E] Existing project', async ({ page }) => {
```
jpggvilaca, line 6: skipping for now since the test step with the STREAM is not working.

application/docker/Dockerfile
```diff
@@ -25,6 +26,8 @@ RUN --mount=type=cache,target=/root/.cache/uv \
     uv sync --frozen --no-editable
 
 COPY backend/app ./app
+COPY backend/run.sh ./run.sh
```
leoll2, line 29: `run.sh` should not be copied in the Docker image, it is not meant for production.

application/.dockerignore
leoll2: No, we shouldn't copy it. Can you motivate the need for the copy?

application/backend/readme.md
leoll2, line 1: With the current pace of development, I'm afraid this `readme.md` will be largely obsolete in a couple of weeks. I intentionally kept it short so that we don't have to constantly update it.

.github/workflows/ui-lint-and-test.yaml
```diff
@@ -353,84 +355,64 @@ jobs:
...
+          name: ui-dist
+          path: application/ui
+
+      - name: Unpack build
+        working-directory: application/ui
+        run: tar -xzf dist.tar.gz
+
+      - name: Install dependencies
+        working-directory: application/ui
+        run: npm ci
```
leoll2, line 387: All this is unnecessary imho. We already have a `Dockerfile` that includes both the server and the UI, and a GHA workflow `build.yaml` to build it. I would expect the e2e workflow to use that image to spin up a container and run the tests on it.

application/docker/docker-compose.yaml
```diff
@@ -49,3 +50,89 @@ services:
...
+    develop:
+      watch:
+        - path: ../backend/app
+          action: sync+restart
+          target: /app/app
+        - path: ../backend/pyproject.toml
+          action: rebuild
+
+  frontend-e2e:
+    profiles:
```
mpryahin, line 95: I'm trying to understand the architecture choice here. we're using a separate `frontend-e2e` container with nginx to serve the UI, but in production we have a single `geti-tune` container that includes both the backend and frontend (nginx + FastAPI).

A couple of questions:

Since e2e tests are meant to validate our production setup, wouldn't it be more valuable to test against the actual production container that includes both services? This would catch any integration issues between nginx and FastAPI in the same container.
  
The current approach requires 3 containers + health checks + inter-service networking. Have you considered using the existing production image from the `build.yaml` workflow instead and running playwright tests directly in the runner's container?

application/backend/readme.md
```diff
@@ -1,3 +1,103 @@
...
+```bash
+# Start the server (development mode)
+./run.sh
+```
+
+### E2E Testing Setup
+
+```bash
+# Full E2E setup with database seeding and test file downloads
+DATABASE_FILE=geti_tune_e2e.db SEED_DB=true DOWNLOAD_FILES=true ./run.sh
```
mpryahin, line 18: Just wondering why a db_file name should be different for e2e tests setup?

jpggvilaca: The goal was to also be able to run these locally, and since some tests require seeding, we need a temporary DB that we can setup/destroy after every run

application/docker/docker-compose.yaml
jpggvilaca: Totally agree, we should leverage the existing `geti-tune`. The issues i had were related with the networking inside docker containers. We want to test webrtc streams and it wasnt working.

.github/workflows/ui-lint-and-test.yaml
jpggvilaca: Agree

jpggvilaca:
Stale PR, closing

PR #4882, opened 2025-10-13
Implement confidence based data collection policy

### LINKED ISSUE #4860: Implement confidence based data collection policy
Implement a policy that collects the frames with low-confidence predictions.

Example of configuration:
```json
{
  "type": "confidence_threshold",
  "enabled": true,
  "confidence_threshold": 0.2,
  "min_sampling_interval": 2.5
}
```

The policy evaluates true if the frame prediction contains _any_ shape with confidence lower (`<`) than the `confidence_threshold` value. To prevent the collection of multiple, almost identical frames in rapid succession, that may be undesirable in most scenarios, the optional parameter `min_sampling_interval` enforces a minimum gap between two consecutive samples. For example, if `"min_sampling_interval": 2.5`, when the policy evaluates true then it also triggers a cooldown of 2.5s: during the cooldown period, the policy always evaluates false, regardless of the confidence score.

warrkan:
Resolves #4860
This PR introduces new data collection policy - Confidence based one.

The policy evaluates true if the frame prediction contains any shape with confidence lower (<) than the confidence_threshold value. To prevent the collection of multiple, almost identical frames in rapid succession, that may be undesirable in most scenarios, the optional parameter min_sampling_interval enforces a minimum gap between two consecutive samples. For example, if "min_sampling_interval": 2.5, when the policy evaluates true then it also triggers a cooldown of 2.5s: during the cooldown period, the policy always evaluates false, regardless of the confidence score.

### Checklist


application/backend/app/api/endpoints/pipelines.py
```diff
@@ -110,7 +109,8 @@ def update_pipeline(
     try:
         if "data_collection_policies" in pipeline_config:
             pipeline_config["data_collection_policies"] = [
-                DataCollectionPolicy.model_validate(policy) for policy in pipeline_config["data_collection_policies"]
+                DataCollectionPolicyAdapter.validate_python(policy)
+                for policy in pipeline_config["data_collection_policies"]
```
leoll2, line 113: Was it a bug? I see that `PipelineService.update_pipeline` expects `partial_config: dict`, that is indeed the type returned by `validate_python`, while `model_validate` returns a `DataCollectionPolicyAdapter`.

leoll2 (dismissed):

application/backend/app/api/endpoints/pipelines.py
warrkan: That's because https://github.com/open-edge-platform/training_extensions/blob/68193b5193f61c7bcd3baad36eaba22a3553d967/application/backend/app/services/pipeline_service.py#L65 expects not `dict`, but already deserialized policies

leoll2 (approved):

PR #4885, opened 2025-10-14
Add secondary toolbar & label picker

### LINKED ISSUE #4886: Add label logic to annotator
* View labels on annotator on top of annotations
* Assign/unassign labels

jpggvilaca:
### Summary

* Add media preview's secondary toolbar
* Update tests
* Add hook to get project labels
* Add label picker

Screenshot 2025-10-14 at 09 35 30
Screenshot 2025-10-14 at 09 35 33

Next PR:
* Show labels on the canvas (shown on the screenshot partially)

### Checklist


application/ui/src/features/dataset/media-preview/secondary-toolbar/label-picker.component.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { components } from 'src/api/openapi-spec';
+
+type ServerLabel = components['schemas']['Label'];
+
+type LabelPickerProps = {
+    labels: ServerLabel[];
+    onSelect: (value: Key | null) => void;
+    selectedLabel: ServerLabel | null;
+};
+export const LabelPicker = ({ labels, onSelect, selectedLabel }: LabelPickerProps) => {
```
MarkRedeman, line 14: Later: let's check if we can reuse the label tree component from Geti. Even if we don't use any hierarchical labels, it's `prefix` and `suffix` design allows us to have a flexibility like showing the label color thumb next to the name, and an hover indicator to show if the label will be added / removed to the annotation(s).

I'd expect to also reuse this same component for showing the labels on the canvas.

application/ui/src/features/annotator/hooks/use-project-labels.hook.ts
```diff
@@ -0,0 +1,18 @@
...
+import { useProjectIdentifier } from 'hooks/use-project-identifier.hook';
+import { get } from 'lodash-es';
+import { $api } from 'src/api/client';
+import { components } from 'src/api/openapi-spec';
+
+type ServerLabel = components['schemas']['Label'];
+
+export const useProjectLabels = (): ServerLabel[] => {
+    const projectId = useProjectIdentifier();
+    const { data: project } = $api.useQuery('get', '/api/projects/{project_id}', {
```
MarkRedeman: Could this be changed to a suspense query? If we do that then we don't have to use lodash, and can do `project.task.labels` directly. (now I feel like we are sneaking in a type assertion)

application/ui/src/features/dataset/media-preview/secondary-toolbar/use-secondary-toolbar-state.hook.ts
```diff
@@ -0,0 +1,53 @@
...
+import { Label } from 'src/features/annotator/types';
+
+export const useSecondaryToolbarState = () => {
+    const { activeTool } = useAnnotator();
+    const { selectedAnnotations } = useSelectedAnnotations();
+    const { annotations, updateAnnotation } = useAnnotationActions();
+
+    const projectLabels = useProjectLabels();
+
+    const isHidden = selectedAnnotations.size === 0 && activeTool === 'selection';
```
MarkRedeman, line 19: Do we not have a "default label" like we have in Geti?

I.e. what happens if the user draws a new polygon or bounding box, will that new annotation not have any labels? To me the default label allowed me to very quickly annotate a lot of images.

application/ui/src/features/dataset/media-preview/secondary-toolbar/secondary-toolbar.component.tsx
```diff
@@ -2,13 +2,26 @@
...
     return (
-        <Flex height={'100%'} alignItems={'center'} justifyContent={'center'} margin={'size-100'}>
-            <Grid>
-                <Flex>Something</Flex>
-                <Flex>Something else</Flex>
+        <Flex height={'100%'} alignItems={'center'} margin={'size-100'} minHeight={'size-1200'}>
+            <Grid UNSAFE_className={classes.toolbarGrid} isHidden={isHidden}>
+                <Flex UNSAFE_className={classes.toolbarSection}>
+                    <LabelPicker
+                        selectedLabel={get(annotationsToUpdate, '[0].labels[0]', null)}
```
MarkRedeman: I personally really dislike using `'[0].labels[0]'` here. To me this is asking for type errors.

Can we not do something like,
```
const label = annotationsToUpdate.at(0)?.labels?.at(0) ?? 0;
```

application/ui/src/features/dataset/media-preview/secondary-toolbar/use-secondary-toolbar-state.hook.ts
```diff
@@ -0,0 +1,53 @@
...
+                updateAnnotation({
+                    ...annotation,
+                    labels: annotation.labels?.filter((label) => label.id !== labelId) as Label[],
+                });
+            } else {
+                updateAnnotation({
+                    ...annotation,
+                    labels: [...(annotation.labels || []), selectedLabel as Label],
+                });
+            }
```
MarkRedeman, line 34: This might lead to a weird issue if:
1. Annotation 1 has the label "Dog"
2. Annotation 2 does not have the label "Dog"
3. Annotation 1 and 2 are selected
4. User selects "Dog"

This will result in the label being removed from Annotation 1, but added to Annotation 2.

In Geti the bulk assign labels component had a different behaviour that I think is more user friendly:
```typescript
    const onToggleLabel = runWhenTruthy((label: Label) => {
        const annotationIds = getIds([...selectedAnnotations]);

        if (labelIsAssignedToEveryAnnotation(label)) {
            removeLabels([label], annotationIds);
        } else {
            addLabel(label, annotationIds);
        }

        setIsOpened(false);
    });
```

application/ui/src/features/dataset/media-preview/secondary-toolbar/label-picker.component.tsx
jpggvilaca: I will update this component to include color and more importantly, to allow multiple selection. I can take a peek at classic and see what i can steal from there

MarkRedeman:
One other thing: in the screenshot I see the "No object" label being referenced. How are we planning to deal with this?

application/ui/src/features/dataset/media-preview/secondary-toolbar/use-secondary-toolbar-state.hook.ts
jpggvilaca: There is no concept of default label yet

jpggvilaca:
> One other thing: in the screenshot I see the "No object" label being referenced. How are we planning to deal with this?

For now idk for sure, but we'll either map it to a user-friendly string or remove it (by filtering or asking the backend to not send it)

application/ui/src/features/dataset/media-preview/secondary-toolbar/use-secondary-toolbar-state.hook.ts
jpggvilaca: Agree, updated

application/ui/src/features/dataset/media-preview/secondary-toolbar/secondary-toolbar.component.tsx
jpggvilaca: Yeah ofc, updated

application/ui/src/features/annotator/hooks/use-project-labels.hook.ts
jpggvilaca: It was not the intent, but agree, updated

MarkRedeman:
> > One other thing: in the screenshot I see the "No object" label being referenced. How are we planning to deal with this?
> 
> For now idk for sure, but we'll either map it to a user-friendly string or remove it (by filtering or asking the backend to not send it)

I was more hinting to how we should apply the empty label if a user selects it. I.e. we should then remove all of the annotations and replace it with a global annotation that classifies the whole image as empty. Is this already supported by the backend?

leoll2:
> I was more hinting to how we should apply the empty label if a user selects it. I.e. we should then remove all of the annotations and replace it with a global annotation that classifies the whole image as empty. Is this already supported by the backend?

To apply the empty label, UI should POST on the `/annotations` endpoint with the payload `{"annotations": []}`

jpggvilaca:
> > > One other thing: in the screenshot I see the "No object" label being referenced. How are we planning to deal with this?
> > 
> > 
> > For now idk for sure, but we'll either map it to a user-friendly string or remove it (by filtering or asking the backend to not send it)
> 
> I was more hinting to how we should apply the empty label if a user selects it. I.e. we should then remove all of the annotations and replace it with a global annotation that classifies the whole image as empty. Is this already supported by the backend?

"No_Object" has no meaning , it was just a string added from seeding the db. To avoid making changes to annotations through the label selection i suggest having an explicit button to "REMOVE ALL ANNOTATIONS", for the usecase of none of them apply. WDYT? @MarkRedeman  @leoll2

leoll2:
Removal of annotations and assignment of "empty" label are semantically different operations; the server supports both.

- Remove an annotation = delete any annotation-related information for the item. After the operation, the annotation state is basically equivalent to a newly uploaded image: we don't know if it contains any object or not.
- Assign the empty label = explicitly state that the image has no object.

Unannotated images should be reviewed by the user, who can annotate them appropriately if they contain any object.
Empty-labeled images are already reviewed, no further processing required.

jpggvilaca:
> Removal of annotations and assignment of "empty" label are semantically different operations; the server supports both.
> 
> * Remove an annotation = delete any annotation-related information for the item. After the operation, the annotation state is basically equivalent to a newly uploaded image: we don't know if it contains any object or not.
> * Assign the empty label = explicitly state that the image has no object.
> 
> Unannotated images should be reviewed by the user, who can annotate them appropriately if they contain any object. Empty-labeled images are already reviewed, no further processing required.

ok so how we do we control this Empty label? I mean, on the UI we assign a label with that name, to the annotation, but how does the backend handle this? you mentioned sending an empty array of annotations, but isnt this already the case for unannotated images>?

MarkRedeman:
> > I was more hinting to how we should apply the empty label if a user selects it. I.e. we should then remove all of the annotations and replace it with a global annotation that classifies the whole image as empty. Is this already supported by the backend?
> 
> To apply the empty label, UI should POST on the `/annotations` endpoint with the payload `{"annotations": []}`

But how do we then distinguish between:
- The image explicitly not having any objects the user is interested in
- The image has not been annotated yet, OR the user removed annotations that were wrong, ending up with zero annotations




> explicit button to "REMOVE ALL ANNOTATIONS", for the usecase of none of them applyThat sounds weird to me? Why not just select all and click the delete button?

> Removal of annotations and assignment of "empty" label are semantically different operations; the server supports both.
> 
>     * Remove an annotation = delete any annotation-related information for the item. After the operation, the annotation state is basically equivalent to a newly uploaded image: we don't know if it contains any object or not.
> 
>     * Assign the empty label = explicitly state that the image has no object.
> 
> 
> Unannotated images should be reviewed by the user, who can annotate them appropriately if they contain any object. Empty-labeled images are already reviewed, no further processing required.

This seems to contradict your previous comment about posting with `{ annotations: [] }`. I remember seeing a design before where we would have `{ annotations: [{ shape: { type: 'full' }, labels: [{ id: emtpy_label_id }] ] }`

application/ui/src/features/dataset/media-preview/secondary-toolbar/use-secondary-toolbar-state.hook.ts
```diff
@@ -0,0 +1,71 @@
...
+    const annotationsToUpdate = annotations.filter((annotation) => selectedAnnotations.has(annotation.id));
+
+    const addLabels = (labelId: Key | null) => {
+        const selectedLabel = projectLabels.find((label) => label.id === labelId);
+
+        annotationsToUpdate.forEach((annotation) => {
+            const hasLabel = annotation.labels?.some((label) => label.id === labelId);
+
+            if (!hasLabel) {
+                updateAnnotation({
```
MarkRedeman, line 30: Is it possible to do this in a single state update, instead of calling it N times for N annotations?
If we do it in N updates then the undo / redo stack will also get N updates while the user only performed a single action.

application/ui/src/features/annotator/annotations/selectable-annotation.component.tsx
```diff
@@ -72,15 +72,13 @@ export const SelectableAnnotation = ({ children }: { children: ReactNode }) => {
                 ...(isSelected
                     ? {
                           fillOpacity: 0.7,
-                          ['--annotation-fill']: annotation.labels[0].color,
+                          ['--annotation-fill']: annotation.labels.length
+                              ? annotation.labels[0].color
+                              : 'var(--energy-blue-light)',
```
MarkRedeman, line 77: I'd prefer to change this to use `at()` because it gives a better indication of having undefined values, i.e.:
```
['--annotation-fill']: annotation.labels.at(0)?.color ?? 'var(--energy-blue-light)',
```

MarkRedeman (approved):
I tried it out locally but it feels very buggy. Approving to unblock.

application/ui/src/features/annotator/annotations/selectable-annotation.component.tsx
jpggvilaca: TS doesnt like that, i'll keep [0] for now as `labels` will always exist and `labels[0]` will return `undefined` so it's ok

application/ui/src/features/dataset/media-preview/secondary-toolbar/use-secondary-toolbar-state.hook.ts
jpggvilaca: Absolutely! I will make all annotator actions plural, to batch when needed. Thanks

leoll2:
> ok so how we do we control this Empty label? I mean, on the UI we assign a label with that name, to the annotation, but how does the backend handle this? you mentioned sending an empty array of annotations, but isnt this already the case for unannotated images>?

To assign the empty label, use `POST /annotations` with `{"annotations": []}`. To fully remove the annotation from an image that you previously annotated, namely reset it to unannotated state, use `DELETE /annotations`.

> But how do we then distinguish between:
> - The image explicitly not having any objects the user is interested in
> - The image has not been annotated yet, OR the user removed annotations that were wrong, ending up with zero annotations

In the first case, `GET /annotations` should return `{"annotations": [], "user_reviewed": true, "prediction_model_id": null}`. In the second case, the same endpoint should return 404 Not found. No need to distinguish between the case of new unannotated image and image with removed annotations; only the current state matters, not the history.

However, I've just verified that the implementation doesn't match the design, due to a bug: for newly uploaded images, GET incorrectly returns the empty list instead of 404. We'll fix it soon.

>  I remember seeing a design before where we would have { annotations: [{ shape: { type: 'full' }, labels: [{ id: emtpy_label_id }] ] }

I don't remind such a design. It's not ideal imho because it forces the empty label to have an explicit ID, and also forces detection/segmentation to accept the full-image shape type, which normally isn't allowed.

PR #4908, opened 2025-10-15
Label name and color should not be optional in endpoint response

### LINKED ISSUE #4892: Label name and color should not be optional in endpoint response
**Describe the bug**

There are some fields in the REST API, like [labels' name and color](https://github.com/open-edge-platform/training_extensions/blob/develop/application/backend/app/schemas/label.py#L89-L92) that are designed to be optional.
For labels specifically both the name and color should be required as otherwise we can't visualize these.

Note: I saw that the `id` is also optional. To me making this required would allow us to both simplify the client and server code.

**Steps to Reproduce**

1. Check the OpenAPI spec that all fields that are required are not nullable

warrkan:
### Summary

Resolves #4892

### Checklist


application/backend/app/schemas/label.py
```diff
@@ -39,66 +31,43 @@ def to_label(self) -> "Label":
...
-class PatchLabels(BaseModel):
-    labels_to_edit: list[LabelToEdit] = Field(default_factory=list, description="List of labels to edit")
-    labels_to_add: list[LabelToAdd] = Field(default_factory=list, description="List of labels to add")
-    labels_to_remove: list[LabelToRemove] = Field(default_factory=list, description="List of labels to remove")
+class LabelBase(BaseIDModel):
+    name: str = Field(..., min_length=1, max_length=50, description="Label name")
+    hotkey: str | None = Field(None, description="Hotkey assigned to the label")
+
+
+class LabelView(LabelBase):
```
itallix: For the response types `id` should be required too. This can be achieved with inheritance: `LabelView(RequiresID, LabelBase)`

application/backend/app/schemas/label.py
```diff
@@ -13,20 +13,12 @@
 
 class LabelToEdit(BaseModel):
     id: UUID = Field(..., description="UUID of the label to edit")
-    new_name: str | None = Field(None, min_length=1, max_length=50, description="New label name")
-    new_color: Annotated[str | None, StringConstraints(pattern=COLOR_REGEX)] = Field(
-        None, description="New hex color code, e.g. #RRGGBB or #RGB"
+    new_name: str = Field(..., min_length=1, max_length=50, description="New label name")
+    new_color: Annotated[str, StringConstraints(pattern=COLOR_REGEX)] = Field(
+        ..., description="New hex color code, e.g. #RRGGBB or #RGB"
```
itallix: Why making all new attributes required for edit schema? Don't we let the user decide whether to  change `color` or `name` or both?

application/backend/app/schemas/label.py
```diff
@@ -112,3 +81,26 @@ class LabelReference(BaseRequiredIDModel):
             }
         }
     }
+
+
+class PatchLabels(BaseModel):
+    labels_to_edit: list[LabelToEdit] = Field(default_factory=list, description="List of labels to edit")
+    labels_to_add: list[LabelCreate] = Field(default_factory=list, description="List of labels to add")
+    labels_to_remove: list[LabelToRemove] = Field(default_factory=list, description="List of labels to remove")
```
itallix: Shall we follow the same naming convention for request schemas and remove `To` suffix, e.g.: `LabelEdit` instead of `LabelToEdit`?

application/backend/app/services/label_service.py
```diff
@@ -34,28 +27,36 @@ class LabelService:
...
+    def create_label(self, project_id: UUID, label: LabelCreate) -> LabelView:
         label_repo = LabelRepository(str(project_id), self._db_session)
         try:
-            saved = label_repo.save(LabelMapper.from_schema(project_id=project_id, label=label))
+            saved = label_repo.save(
+                LabelDB(
+                    id=str(label.id) if label.id is not None else None,
+                    project_id=str(project_id),
+                    name=label.name,
+                    color=label.color if label.color is not None else random_color(),
```
itallix, line 38: I'm wondering whether we should require users to always provide the color instead of using an implicit approach with random generation?

application/backend/app/schemas/label.py
leoll2: True, the edit schema should allow partial updates, i.e. let the attribute be optional

application/backend/app/services/label_service.py
leoll2: For clients that interact directly with the REST API (without UI), the color is not important. To simplify the API usage, we only require the fields that are always required.

camiloHimura (approved):

itallix (approved):
LGTM!

PR #4913, opened 2025-10-16
Add annotation labels

jpggvilaca:
### Summary

* Visualisation of labels on the annotator
* Addition & deletion logic
* Pluralise annotation actions, allowing us to do bulk actions (add/delete/update)
* Add "deselect" logic to deselect all annotations
* Other fixes (project list layout, query infinite retry, virtualizer duplicate keys, project creation)

https://github.com/user-attachments/assets/341e952c-867d-4e9a-b01c-5adbd165eced

### Checklist


application/ui/src/features/annotator/annotations/annotation-shape-with-labels.component.tsx
```diff
@@ -0,0 +1,45 @@
...
+            </g>
+        );
+    }
+
+    // For polygon, find the top-left point to use as the group's origin
+    const polygonPoints = (shape as Polygon).points;
+    const minX = Math.min(...polygonPoints.map((point) => point.x));
+    const minY = Math.min(...polygonPoints.map((point) => point.y));
+
+    // Adjust polygon points to be relative to the group's origin
```
jpggvilaca: this was one of the important changes. To position the labels relative to the annotation, we wrap everything in a group (<g />) and use those X/Y coordinates to position the label

application/ui/src/features/annotator/annotations/annotation-shape-with-labels.component.tsx
```diff
@@ -0,0 +1,45 @@
...
+    if (shape.type === 'rectangle') {
+        return (
+            <g transform={`translate(${shape.x}, ${shape.y})`}>
+                <rect aria-label='annotation rect' x={0} y={0} width={shape.width} height={shape.height} fill={color} />
+                <AnnotationLabels labels={labels} onRemove={removeLabels} />
+            </g>
+        );
+    }
+
+    // For polygon, find the top-left point to use as the group's origin
```
jpggvilaca: might not be perfect. Haven't fully tested with segments

application/ui/src/features/annotator/annotations/annotations.component.tsx
```diff
@@ -16,13 +19,30 @@ type AnnotationsProps = {
 };
 
 export const Annotations = ({ annotations, width, height, isFocussed }: AnnotationsProps) => {
+    const { setSelectedAnnotations } = useSelectedAnnotations();
+
+    // If the user clicks on an empty spot on the canvas, we want to deselect
```
jpggvilaca, line 24: Added missing behavior to DESELECT. Important because when using the label picker we want to assign the label to the currently selected annotations.

application/ui/src/features/annotator/annotations/annotation-shape-with-labels.component.tsx
MarkRedeman: In Geti we use [polylabel](https://github.com/mapbox/polylabel) why not use that?

I think in the latest design from Marisuz he put the labels in the middle of the polygons. This library would help a lot with this

jpggvilaca: Ah good one! I will

jpggvilaca: Updated

application/ui/src/features/annotator/annotations/annotation-labels.component.tsx
```diff
@@ -0,0 +1,139 @@
...
+const placeholderLabel = { id: uuid(), name: 'No label', color: 'var(--annotation-fill)', isPrediction: false };
+
+interface AnnotationLabelsProps {
+    labels: Label[];
+    onRemove: (labelId: string) => void;
+}
+export const AnnotationLabels = ({ labels, onRemove }: AnnotationLabelsProps) => {
+    const { scale } = useZoom();
+
+    const styles = labelStyles(scale);
```
camiloHimura, line 53:
```suggestion
    const { height, padding, gap, closeButtonWidth, fontSize, yOffset, charWidth } = labelStyles(scale);
```

application/ui/src/features/annotator/annotations/annotation-labels.component.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+const labelStyles = (scale: number) => {
+    // We need the actual values for calculations:
+    // spectrum-global-dimension-size-50 = 4px
+    // spectrum-global-dimension-size-100 = 8px
+    // spectrum-global-dimension-size-175 = 14px
+    // spectrum-global-dimension-size-250 = 20px
+    // spectrum-global-dimension-size-300 = 24px
+
+    const height = 24 / scale;
```
MarkRedeman, line 19: I think Spectrum also has a utility function that allows you to get these sizes based on the variable's name and then get a javascript number out. @camiloHimura I remember you using this often, can you check that?

application/ui/src/features/annotator/annotations/annotation-shape-with-labels.component.tsx
```diff
@@ -0,0 +1,43 @@
...
+export const AnnotationShapeWithLabels = ({ annotation }: AnnotationShapeProps) => {
+    const { shape, labels } = annotation;
+    const color = labels.length ? labels[0].color : '--annotation-fill';
+    const { removeLabels } = useSecondaryToolbarState();
+
+    if (shape.type === 'rectangle') {
+        return (
+            <g transform={`translate(${shape.x}, ${shape.y})`}>
+                <rect aria-label='annotation rect' x={0} y={0} width={shape.width} height={shape.height} fill={color} />
+                <AnnotationLabels labels={labels} onRemove={removeLabels} />
```
MarkRedeman, line 23: In Geti when the user removes the label we remove the shape as well. I think this would make sense here as well.

If a user would want to remove the label without removing the shape, then that would actually mean that they want to change the label into another label. We can implement this the same way that we did with Geti: show the label tree picker that allows the user to add / remove labels the the annotation.

application/ui/src/features/annotator/annotations/annotation-labels.component.tsx
camiloHimura: I think you’re referring to `dimensionValue('size-125')`, but what that actually does is return the full CSS variable  for example, `var(--spectrum-global-dimension-size-125, var(--spectrum-alias-size-125))`  not its numeric value.

application/ui/src/features/annotator/annotations/annotation-shape-with-labels.component.tsx
```diff
@@ -0,0 +1,43 @@
...
+};
+export const AnnotationShapeWithLabels = ({ annotation }: AnnotationShapeProps) => {
+    const { shape, labels } = annotation;
+    const color = labels.length ? labels[0].color : '--annotation-fill';
+    const { removeLabels } = useSecondaryToolbarState();
+
+    if (shape.type === 'rectangle') {
+        return (
+            <g transform={`translate(${shape.x}, ${shape.y})`}>
+                <rect aria-label='annotation rect' x={0} y={0} width={shape.width} height={shape.height} fill={color} />
```
MarkRedeman, line 22: Can we use the `AnnotationShape` for the acutal shape? (i.e. for both the rect and polygon)
That way we avoid styling differences.

MarkRedeman: Perhaps for naming instead of `AnnotationShapeWithLabels` just call it `Annotation` (to me an `annotation = shape + labels`)

application/ui/src/features/annotator/annotations/annotation-labels.component.tsx
MarkRedeman: Ah yeah that's the one I was thinking of.

Alternatively I would wonder if you can't use these values as style values? Border radius etc are all usable as styles (for the zoom values just do `calc(var(spectrum-global-dimension-size-50) / var(--zoom-scale))`. That should also make the zoom transitions smoother as we could potentially avoid rerenders by not depending on `scale` directly.

application/ui/src/features/annotator/annotations/annotation-shape-with-labels.component.tsx
jpggvilaca: https://github.com/open-edge-platform/training_extensions/pull/4915 . As for the naming...i think it's more clear with the suffix imho...

application/ui/src/features/annotator/annotations/annotation-shape-with-labels.component.tsx
jpggvilaca: The user can have multiple labels and only wanting to remove one. We can obviously check the #labels but I think shape being separate from labels is a plus

application/ui/src/features/annotator/annotations/annotation-labels.component.tsx
jpggvilaca: svg doesnt support `calc` :/

application/ui/src/features/annotator/annotations/annotation-shape-with-labels.component.tsx
MarkRedeman: > The user can have multiple labels and only wanting to remove one

That's why the label picker is very useful. They can click the picker and we show a tree where when you hover over the labels it gives a hint on if the label gets added or removed.

jpggvilaca: i dont follow. What is the actual issue? Not removing the shape?

camiloHimura (approved):

application/ui/src/features/annotator/annotations/annotation-shape-with-labels.component.tsx
jpggvilaca: Merging this right now because it's blocking, but we can iterate on this.

PR #4920, opened 2025-10-20
Experimental datumaro implementation

### LINKED ISSUE #1802: Integrate OTX with the new dataset class
Use the new Datumaro dataset class (#1785) instead of the old one in the implementation of the OTX dataset classes and data loaders. Also replace the OTXDataItem class with the new Sample class.

The following dataset types and features must be supported:
- [x] Classification
- [x] Detection
- [x] Instance segmentation
- [x] Keypoint detection
- [x] Segmentation
- [x] Tiling
- [x] Multi-label and hierachical classification
- [x] Ellipse

AlbertvanHouten:
This pull request introduces the new experimental dataset into the dataset classes of OTX. A lot of logic has been migrated to datumaro such as management of image color channels and the tiling implementation. 

### Checklist


.github/workflows/lib-lint-and-test.yaml
```diff
@@ -116,8 +116,9 @@ jobs:
         with:
           persist-credentials: false
 
-      - name: Install Rust toolchain
-        uses: dtolnay/rust-toolchain@b9ed5a8fb8afb645c9cfaffecc60fc946d28f857 # stable
+      # TODO remove after switching to an official Datumaro release with pre-built wheels
+      - name: Installing Rust toolchain
+        uses: dtolnay/rust-toolchain@6d653acede28d24f02e3cd41383119e8b1b35921
```
mramotowski, line 121: Why this sha? What version is it? Pls add info in comment. I checked latest versions and can't see this commit:
commit 0d231036f5e097127fd969f80fb8e44d7394f7f9 (HEAD -> 1.91.1, origin/1.91.1, origin/1.91) 
commit a4a18948872904a1ea9bf180c73ce0a3ce1ceaab (HEAD -> 1.91.0, origin/1.91.0)
The commit you provide looks like commit from master branch:
commit 6d653acede28d24f02e3cd41383119e8b1b35921
Shouldn't we use some of release versions?

leoll2: It's probably taken from #5054, which I copied from the Datumaro repo [[ref](https://github.com/open-edge-platform/datumaro/blob/42138a84b48c1b6505a527600302d2b1acfbe81a/.github/workflows/pr_check.yml#L48)].

The maintainer of this action `dtolnay/rust-toolchain` keeps updating his project but doesn't make official GH releases (actually there's one release - [v1](https://github.com/dtolnay/rust-toolchain/releases) - but it's from 2022).

library/src/otx/data/__init__.py
leoll2, line 1: Why is this file removed? Could you restore it?

library/src/otx/recipe/semantic_segmentation/dino_v2_tile.yaml
AlbertvanHouten, line 1: @kprokofi please review if these changes make sense. The old logic only worked when tiling with square images.

.github/workflows/lib-lint-and-test.yaml
```diff
@@ -129,9 +135,19 @@ jobs:
...
+        run: uv run pytest tests/unit --ignore=tests/unit/backend/native/models --cov --cov-report=xml --cov-append
+
+      # Unit test models separately using --forked flag so that each is run in a separate process.
+      # Running these models after each other from the same process can cause segfaults
+      - name: Unit testing models
         working-directory: library
-        run: uv run pytest tests/unit --cov --cov-report=xml
+        env:
+          OMP_NUM_THREADS: "1"
+        run: uv run pytest tests/unit/backend/native/models --cov --cov-report=xml --cov-append --forked
```
leoll2, line 150: Do you have any idea what causes the segfaults?

library/src/otx/backend/native/callbacks/gpu_mem_monitor.py
```diff
@@ -29,7 +29,7 @@ def _get_and_log_device_stats(
             batch_size (int): batch size.
         """
         device = trainer.strategy.root_device
-        if device.type in ["cpu", "xpu"]:
+        if device.type in ["cpu", "xpu", "mps"]:
```
leoll2, line 32: Can you explain this change? Does OTX even support MPS?

library/src/otx/data/factory.py
```diff
@@ -41,58 +44,71 @@ class OTXDatasetFactory:
...
-        dm_subset: DmDataset,
+        dm_subset: DmDataset | DatasetNew,
         cfg_subset: SubsetConfig,
         data_format: str,
-        image_color_channel: ImageColorChannel = ImageColorChannel.RGB,
         include_polygons: bool = False,
-        ignore_index: int = 255,
-    ) -> OTXDataset:
+        # TODO(gdlg): Add support for ignore_index again
+        ignore_index: int = 255,  # noqa: ARG003
```
leoll2, line 52: There is a TODO here. Do you know what's the purpose of this `ignore_index` parameter?

library/src/otx/recipe/detection/rtmdet_tiny.yaml
```diff
@@ -61,7 +61,6 @@ overrides:
     input_size:
       - 640
       - 640
-    image_color_channel: BGR
```
leoll2, line 64: Some models expect BGR data as input, others RGB. If you remove this line from the recipes, how do you know how the right color format to use to prepare the input data? Note that training on RGB and BGR is not equivalent because the pre-trained weights are generated w.r.t. input data formatted and normalized in a specific way.

library/src/otx/backend/native/callbacks/gpu_mem_monitor.py
AlbertvanHouten: Yes, I've been doing most of the testing on my MacBook. I don't see a reason why it shouldn't be supported.

library/src/otx/recipe/detection/rtmdet_tiny.yaml
AlbertvanHouten: I removed this after you removed it from OTXDataModule as this broke the recipes https://github.com/open-edge-platform/training_extensions/pull/4920/commits/eafdce1ab2c6b1aa0091c05e3307101a355c53af

I did not look into it further. Was there a specific reason why you removed it from OTXDataModule? If not, I'll look into reverting it.

AlbertvanHouten: After further inspection, only the detection tasks seem to use BGR. This is already defined in the sample class.

leoll2: > Was there a specific reason why you removed it from OTXDataModule?

I removed `image_color_channel` from `OTXDataModule` because it was already removed from `OTXDataset` in this earlier commit: a7d0c5129924edc47a5501bfa77e293de6c10495, so that data module would try to read an undefined `train_dataset.image_color_channel` attribute.

AlbertvanHouten: I see, makes sense :D

.github/workflows/lib-lint-and-test.yaml
AlbertvanHouten: Unclear to me. I've tried a few different things to see if it would fix it, like clearing up pytorch memory usage, but this was the only way that I could fix it.

library/src/otx/data/factory.py
kprokofi: Ignore index is necessary for Semantic Segmentation tasks. This index helps to ignore marked pixels.

library/src/otx/backend/native/models/detection/rtdetr.py
```diff
@@ -131,6 +132,15 @@ def _customize_inputs(
         pad_size_divisor: int = 32,
         pad_value: int = 0,
     ) -> dict[str, Any]:
+        # Stack images if they're in list format
+        if isinstance(entity.images, list):
```
kprokofi, line 136: Why does this change apply only for RTDetr? Can we ensure that images are always stacked on the dataset level?

library/src/otx/backend/native/tools/tile_merge.py
```diff
@@ -508,7 +499,7 @@ def __init__(
     def merge(
         self,
         batch_tile_preds: list[OTXPredBatch],
-        batch_tile_attrs: list[list[dict]],
+        batch_tile_attrs: list[list[TileInfo]],
```
kprokofi: Since you are changing name tile_attrs to tile_info, could you change it everywhere for consistency?

library/src/otx/backend/openvino/models/segmentation.py
```diff
@@ -137,13 +137,34 @@ def prepare_metric_inputs(
...
 
-        return [
-            {
-                "preds": pred_mask,
-                "target": target_mask,
-            }
-            for pred_mask, target_mask in zip(preds.masks, inputs.masks)
-        ]
+        aligned: list[dict[str, Any]] = []
+        for pred_mask, target_mask in zip(preds.masks, inputs.masks):
```
kprokofi: Why not just pm, tm then? (for pm, tm in ..)

library/src/otx/data/factory.py
AlbertvanHouten: Is there any use case for the ignore index being different from the value used for background pixels? Or is it  always the same?

library/src/otx/backend/openvino/models/segmentation.py
```diff
@@ -137,13 +137,34 @@ def prepare_metric_inputs(
...
+            tm = target_mask
+
+            # Ensure both are 3D with channel dim first: [1, H, W]
+            # Some backends may return [H, W] — add channel dim if needed.
+            if pm.ndim == 2:
+                pm = pm.unsqueeze(0)
+            if tm.ndim == 2:
+                tm = tm.unsqueeze(0)
+
+            # Align spatial dimensions if swapped (e.g., [1, W, H] vs [1, H, W])
```
kprokofi: Why spatial dimensions can be misaligned and swapped? Code here looks as a hack for me. We need to align targets and predictions on the dataset preparation / output preprocessing stage, not here.

library/src/otx/data/entity/sample.py
```diff
@@ -0,0 +1,298 @@
...
+
+        self.img_info = ImageInfo(
+            img_idx=0,
+            img_shape=shape,
+            ori_shape=shape,
+        )
+
+
+@register_pytree_node
+class InstanceSegmentationSampleWithMask(OTXSample):
```
kprokofi: 1) Can't we combine InstanceSegmentationSample and InstanceSegmentationSampleWithMask? The difference in mask field only. 
2) Why do we still include polygons in InstanceSegmentationSampleWithMask?

library/src/otx/data/factory.py
```diff
@@ -41,58 +44,71 @@ class OTXDatasetFactory:
...
             "transforms": transforms,
             "data_format": data_format,
-            "image_color_channel": image_color_channel,
             "to_tv_image": cfg_subset.to_tv_image,
         }
 
         if task == OTXTaskType.MULTI_CLASS_CLS:
             from .dataset.classification import OTXMulticlassClsDataset
 
+            dataset = convert_from_legacy(dm_subset)
```
kprokofi, line 66: Could you explain for me, why do we need "convert from legacy"?

library/src/otx/data/factory.py
```diff
@@ -41,58 +44,71 @@ class OTXDatasetFactory:
...
+        dm_subset: DmDataset | DatasetNew,
         cfg_subset: SubsetConfig,
         data_format: str,
-        image_color_channel: ImageColorChannel = ImageColorChannel.RGB,
         include_polygons: bool = False,
-        ignore_index: int = 255,
-    ) -> OTXDataset:
+        # TODO(gdlg): Add support for ignore_index again
+        ignore_index: int = 255,  # noqa: ARG003
+    ) -> OTXDataset | OTXDatasetNew:
```
kprokofi: they are the same objects. Is there any reason to define OTXDatasetNew?

library/src/otx/data/factory.py
```diff
@@ -41,58 +44,71 @@ class OTXDatasetFactory:
 
     @classmethod
     def create(
-        cls: type[OTXDatasetFactory],
+        cls,
         task: OTXTaskType,
-        dm_subset: DmDataset,
+        dm_subset: DmDataset | DatasetNew,
```
kprokofi, line 48: Is it for backward compatibility? I don't think we need that.  cc @leoll2 
Same for "convert_from_legacy".

GetiTune will depend on the new Datumaro Dataset, Geti classic has no dependency on the develop branch. OTX uses file system, not datumaro directly. I propose to support only the latest version if no stoppers.

library/src/otx/data/transform_libs/torchvision.py
```diff
@@ -112,8 +112,20 @@ class NumpytoTVTensorMixin:
...
+                if isinstance(image, np.ndarray):
+                    # Ensure the image is in HWC format before converting to tv_tensor
+                    # If the image is in CHW format (first dim <= 4 and smaller than other dims), transpose it
+                    if image.ndim == 3 and image.shape[0] <= 4 and image.shape[0] < min(image.shape[1:]):
+                        image = image.transpose(1, 2, 0)
+                    inputs.image = F.to_image(image.copy())
+                elif isinstance(image, torch.Tensor):
+                    # Handle tensor in HWC format - convert to CHW
+                    # If the last dimension is <= 4 and smaller than other dims, it's likely HWC format
+                    if image.ndim == 3 and image.shape[-1] <= 4 and image.shape[-1] < min(image.shape[:-1]):
```
kprokofi: Why do we need to convert tensors from CHW (what is actually needed for training and transforms) to HWC?

library/src/otx/data/transform_libs/torchvision.py
```diff
@@ -2725,18 +2741,65 @@ class Compose(tvt_v2.Compose):
...
+    transformable fields (image, masks, bboxes) and applying transforms to them,
+    avoiding transforms being applied to non-image tensors like labels.
     """
 
+    def _is_native_torchvision_transform(self, transform: tvt_v2.Transform) -> bool:
+        """Check if the transform is a native torchvision transform."""
+        module = type(transform).__module__
+        return module.startswith("torchvision.")
+
+    def _apply_native_transform(self, transform: tvt_v2.Transform, inputs: OTXDataItem) -> OTXDataItem:
```
kprokofi, line 2747: I don't uderstand why do we need this function? We create a copy of entity with fields (BTW, you forgot about Keypoints) and then reassign transformed back. Torchvision can recognize and apply transformations to the needed fields. We define pytree_node for that. It worked before correctly

library/src/otx/data/transform_libs/utils.py
```diff
@@ -120,15 +118,27 @@ def get_image_shape(img: np.ndarray | Tensor | list) -> tuple[int, int]:
...
-    """Convert torch.Tensor 3D image to numpy 3D image.
+    """Convert torch.Tensor 3D image to numpy 3D image in HWC format.
 
     TODO (sungchul): move it into base data entity?
 
     """
     if isinstance(img, np.ndarray):
+        # Check if the numpy array is in CHW format (channels should be <= 4 typically)
+        # If the first dimension is small (<=4) and smaller than other dimensions, it's likely CHW format
+        if img.ndim == 3 and img.shape[0] <= 4 and img.shape[0] < min(img.shape[1:]):
```
kprokofi, line 129: Same comment as above. I think we should always stick to CHW format.

library/src/otx/data/transform_libs/utils.py
```diff
@@ -178,28 +177,37 @@ def rescale_masks(
...
+    rescaled_polygons = np.empty_like(polygons)
+    for i, poly_points in enumerate(polygons):
+        if poly_points.size > 0:
+            scaled_points = poly_points.astype(np.float32)
+            scaled_points[:, 0] *= w_scale  # x coordinates
+            scaled_points[:, 1] *= h_scale  # y coordinates
+            rescaled_polygons[i] = scaled_points
+        else:
+            # Handle empty or invalid polygons
+            rescaled_polygons[i] = np.array([[0, 0], [0, 0], [0, 0]], dtype=np.float32)
```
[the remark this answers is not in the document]
kprokofi: why exactly 3 points BTW? Why not just empty tensor or 0 tensor?

kprokofi (commented):
First round of comments

library/src/otx/recipe/semantic_segmentation/dino_v2_tile.yaml
kprokofi: I think padding do nothing here. We resize and pad to the same size.

library/src/otx/data/transform_libs/torchvision.py
AlbertvanHouten: Torchvision applies transformations to any Tensor. Typing throughout OTX is a huge mess of converting from numpy to tensors back and forth and with the new Datumaro typing has become a lot stricter now. Labels for some tasks are expected as a pytorch Tensor which causes torchvision to try and apply transformations to the labels which will throw an error.

library/src/otx/data/transform_libs/utils.py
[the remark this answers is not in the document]
AlbertvanHouten: Not entirely sure why Greg applied it in this manner, but I guess there is a validation step somewhere that requires 3 points for a valid polygon.

library/src/otx/types/label.py
```diff
@@ -63,7 +74,7 @@ def from_num_classes(cls, num_classes: int) -> LabelInfo:
         )
 
     @classmethod
-    def from_dm_label_groups(cls, dm_label_categories: LabelCategories) -> LabelInfo:
+    def from_dm_label_groups(cls, dm_label_categories: HierarchicalLabelCategories) -> LabelInfo:
```
kprokofi: Since we deleted this in OTXDataset and we define LabelInfo directly, could you clean this class? Specifically, we don't need "from_dm_label_groups", "from_dm_label_groups_arrow", etc.

library/tests/conftest.py
```diff
@@ -22,6 +23,51 @@
...
+    # Restore original working directory
+    os.chdir(original_cwd)
+
+    # Force cleanup to prevent segfaults during pytest teardown
+    try:
+        torch._C._jit_clear_class_registry()
+        if torch.cuda.is_available():
+            torch.cuda.empty_cache()
+        if is_xpu_available():
+            import intel_extension_for_pytorch as ipex
```
kprokofi: We don't use IPEX. It is outdated

library/src/otx/backend/native/models/detection/rtdetr.py
AlbertvanHouten: Other tasks do not implement a `_customize_inputs` function. The existing logic did not account properly for OTXDataBatch as it has images as `torch.Tensor | list[torch.Tensor]`. If OTXDataBatch then actually contains a list this code would break. 

I am not sure how you envision the stacking at the dataset level; this is currently done on the OTX side.

library/tests/integration/api/test_augmentation.py
```diff
@@ -109,10 +109,8 @@ def test_augmentation_kp_det(
     fxt_target_dataset_per_task: dict,
 ):
     configurable_augs = [
-        "TopdownAffine",
         "RandomPhotometricDistort",
         "RandomGaussianBlur",
         "RandomGaussianNoise",
```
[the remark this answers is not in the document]
kprokofi: Why did you remove TopdownAffine from configurable augs list? We actually provide possibility to turn on/off this augmentation in Geti

kprokofi (commented):
Second round of comments

library/src/otx/data/factory.py
AlbertvanHouten: Legacy datumaro datasets can still be passed here since the conversion to the new datumaro dataset schema happens in this factory. We will remove this in the future, but as long as we still support the legacy datasets in datumaro it also makes sense to keep it here. If you already have a new datumaro Dataset, you can initialize the tasks dataset immediately. 

It also reduces the scope for this PR a bit.

library/src/otx/backend/native/models/detection/rtdetr.py
kprokofi: I mean in  the OTX Dataset (in the OTXDataBatch). We need to stack images always for training

kprokofi: Anyway, let me look at this later.

library/src/otx/data/transform_libs/torchvision.py
kprokofi: hmm, labels shouldn't be transformed.  
Ok, let me handle this in my PR with augmentations. I plan to remove numpy and will look at torchvision.Compose problem

library/src/otx/data/factory.py
kprokofi: It is always the same if we don't count background as training class. Sometimes (in some datasets), ignore_index = -1 is used to indicate invalid pixels instead of 255. 

However in the datasets like Pascal VOC we use background in training and mark it as 0 class

library/tests/integration/api/test_augmentation.py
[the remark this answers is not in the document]
AlbertvanHouten: iirc, TopdownAffine can't be an optional augmentation since it is responsible for resizing the image. Otherwise, at least for this test, the images were in the wrong size.

library/src/otx/recipe/semantic_segmentation/dino_v2_tile.yaml
AlbertvanHouten: Without the padding, tiling only works if the image is square (which it was previously in the integration test). I have replaced it with another dataset, as the polygons in the dataset were self-intersecting, which isn't supported, and this new dataset does have rectangle images, which broke specifically this tiling recipe.

library/src/otx/backend/openvino/models/segmentation.py
AlbertvanHouten: Let me recheck this after my break. I have made many changes to get training to work, and probably some of these checks can be removed. I do think it's still good to make some of this code more robust against channel orders.

library/src/otx/backend/openvino/models/segmentation.py
AlbertvanHouten: It is necessary due to a linter rule that does not allow overwriting variables in a loop.

library/src/otx/data/factory.py
AlbertvanHouten: It converts a legacy datumaro dataset to the new datumaro dataset. It will eventually be removed.

library/src/otx/data/entity/sample.py
AlbertvanHouten: Merged together and removed polygons.

library/src/otx/data/factory.py
AlbertvanHouten: Let me fix this in a separate PR. This ignore index should be handled on datumaro's side as well.

AlbertvanHouten:
I believe all comments should be handled now @kprokofi. Please recheck the PR when you have time.

library/src/otx/data/entity/sample.py
```diff
@@ -0,0 +1,266 @@
...
+            ori_shape=shape,
+        )
+
+
+@register_pytree_node
+class SegmentationSample(OTXSample):
+    """OTXDataItemSample is a base class for OTX data items."""
+
+    subset: Subset = subset_field()
+    image: np.ndarray | tv_tensors.Image | torch.Tensor = image_field(dtype=pl.UInt8(), channels_first=False)
```
leoll2, line 206: All other samples have `channels_first=False`. Why does `SegmentationSample` differ?

library/src/otx/backend/native/models/detection/rtdetr.py
leoll2: Please create a follow-up ticket if you plan to address it later

leoll2 (dismissed):

library/src/otx/data/entity/sample.py
AlbertvanHouten: I think its due to different transformations being called for the segmentation task. Without channels first, the predicted masks will have H/W channels swapped.

library/src/otx/backend/native/models/detection/rtdetr.py
kprokofi: https://github.com/open-edge-platform/training_extensions/issues/5162

kprokofi (approved):
Thank you for your work!

library/src/otx/data/factory.py
AlbertvanHouten: https://github.com/open-edge-platform/datumaro/issues/2004

mramotowski (approved):

PR #4932, opened 2025-10-20
Switch lib-lint-and-test workflow to uv

ivanzati:
Switching lib-lint-and-test.yaml workflow to uv from tox

## 🚀 Integration Run Summary

- **Last Commit:** Merge branch 'develop' into ivanzati/uv-lib-lint-workflow
- **Result:** ⚠️ SKIPPED
- **Workflow Run:** [View Run](https://github.com/open-edge-platform/training_extensions/actions/runs/18938961226)


.github/workflows/lib-lint-and-test.yaml
```diff
@@ -4,14 +4,14 @@ on:
   push:
     branches:
       - develop
-      - releases/**
+      - release/**
```
AlexanderBarabanov, line 7: In other workflows we use `releases/**`, although in branch names we use `release` (e.g. https://github.com/open-edge-platform/training_extensions/tree/release/2.6)

Leonardo, could you please clarify what we plan to use? @leoll2

.github/workflows/lib-lint-and-test.yaml
```diff
@@ -89,52 +96,41 @@ jobs:
...
-            COMMIT_ID=${{ github.event.pull_request.head.sha }}
-          else
-            COMMIT_ID=${{ github.sha }}
-          fi
-          # current version of codecov-action does not support uploading reports through the proxy
-          # so we use the latest version of codecov uploader binary
-          curl -Os https://uploader.codecov.io/latest/linux/codecov
-          chmod +x codecov
-          ./codecov -t ${{ secrets.CODECOV_TOKEN }} --sha $COMMIT_ID -U $HTTP_PROXY -f .tox/coverage_unit-test-${{ matrix.tox-env }}.xml -F ${{ matrix.tox-env }}
+        uses: codecov/codecov-action@5a1091511ad55cbe89839c7260b706298ca349f7 # v5.5.1
```
AlexanderBarabanov, line 145: This step will fail for PR from fork (by design, GH token for `pull_request` trigger has read-only permissions).
I would propose to add condition to skip this step for PR from forks (something like https://github.com/open-edge-platform/training_extensions/blob/develop/.github/workflows/ui-lint-and-test.yaml#L259)

library/pyproject.toml
```diff
@@ -12,7 +12,7 @@ name = "otx"
 dynamic = ["version"]
 description = "OpenVINO™ Training Extensions: Train, Evaluate, Optimize, Deploy Computer Vision Models via OpenVINO™"
 readme = "README.md"
-requires-python = ">=3.10"
+requires-python = ">=3.10,<3.13"
```
AlexanderBarabanov: In app we use `3.13` https://github.com/open-edge-platform/training_extensions/blob/develop/application/docker/Dockerfile#L6
Will it somehow affect lib usage in application?

library/pyproject.toml
```diff
@@ -39,8 +39,7 @@ dependencies = [
     "einops==0.8.1",
     "decord==0.6.0",
     "typeguard>=4.3,<4.5",
-    # TODO(ashwinvaidya17): https://github.com/openvinotoolkit/anomalib/issues/2126
-    "setuptools<70",
+    "setuptools",
```
AlexanderBarabanov: I propose to firstly merge this PR https://github.com/open-edge-platform/training_extensions/pull/4903, as there are some chages in toml as well, e.g. 
`"setuptools==78.1.1"`

.github/workflows/lib-lint-and-test.yaml
ivanzati: I have intention to change this to "release/**" for other workflows as well

library/pyproject.toml
ivanzati: Yep, indeed. Also this would help to integration test with adaptive batch size to succeed I reckon.

.github/workflows/lib-lint-and-test.yaml
leoll2: Good catch. I'd opt for `release` for compatibility with the existing release branches. The other workflows have to be updated accordingly.

.github/workflows/lib-lint-and-test.yaml
```diff
@@ -66,21 +66,28 @@ jobs:
...
           python-version: "3.12"
-      - name: Install tox
+          version: "0.8.15"
+
+      - name: Installing dependencies
         working-directory: library
-        run: |
-          pip install '.[dev]'
-      - name: Code quality checks
+        run: uv sync --locked --extra dev --extra cuda --no-extra xpu
```
leoll2: Is the extra `cuda` necessary for the tests to pass? Does it even work on the standard GPU-less Github runners?

.github/workflows/lib-lint-and-test.yaml
```diff
@@ -66,21 +66,28 @@ jobs:
...
+
+      - name: Run pre-commit checks
         working-directory: library
-        run: tox r -vv -e pre-commit
+        run: uv run --frozen --extra dev --extra cuda --no-extra xpu pre-commit run --all-files
```
leoll2: You already install the dependencies in the previous step, I think `--frozen --extra dev --extra cuda --no-extra xpu` is redundant.

.github/workflows/lib-lint-and-test.yaml
ivanzati: Nice catch, indeed the cuda is not required to pass the quality check. Wanted to keep it the same as next tests. Though would revert back. Thanks!

library/pyproject.toml
leoll2: OTX supports Python 3.13, so `<3.13` is wrong here. I would leave this line unchanged: `">=3.10"`.

.github/workflows/lib-lint-and-test.yaml
```diff
@@ -146,30 +144,47 @@ jobs:
...
+          python-version: ${{ matrix.python-version }}
+          version: "0.8.15"
+
+      - name: Installing dependencies
         working-directory: library
         run: |
           pip install '.[dev]'
-      - name: Run Integration Test
+          uv lock --check
+          uv sync --frozen --extra cuda --no-extra xpu
```
leoll2: FYI I have just merged a CPU only mode (`--extra cpu`, #4972). I don't know if the integration tests require a GPU to pass, probably yes; if not, you may consider switching to this CPU installation.

.github/workflows/lib-lint-and-test.yaml
```diff
@@ -146,30 +144,47 @@ jobs:
       matrix:
         include:
           - task: "multi_class_cls"
+            python-version: "3.12"
```
mramotowski: We are setting `python-version` to `3.12` for all of matrix builds wouldn't it be better to set single env for whole job?

library/pyproject.toml
```diff
@@ -46,29 +46,46 @@ dependencies = [
     "openvino-model-api==0.3.0.4",
     "onnx==1.19.1",
     "onnxconverter-common==1.16.0",
-    "onnxscript==0.5.3",
+    "onnxscript",
     "nncf==2.17.0",
     "anomalib[core]==1.1.3",
     "imgaug==0.4.0",  # for anomalib
     "numpy>=2.0",
+    "tensorboard==2.20.0",
```
leoll2, line 52: Where does this dependency come from?

library/pyproject.toml
```diff
@@ -46,29 +46,46 @@ dependencies = [
...
     "pytest-mock",
-    "pytest-csv",
-    "pytest-cov",
-    "onnxruntime==1.21.1",
+    "dill",
+    "tifffile",
+]
+
+lint = [
+    "ruff==0.9.3",
```
leoll2: FYI ruff upgraded to 0.11.13 in #5003

library/pyproject.toml
```diff
@@ -290,6 +306,9 @@ ignore = [
     "D107",     # Missing docstring in `__init__`
     "D105",     # Missing docstring in magic method
 
+    # NOTE: Disable the following rules for now.
+    "A005", # Module is shadowing a Python built-in
```
leoll2: Can be removed, this is no longer a problem since ruff 0.10 made this rule less strict

library/pyproject.toml
```diff
@@ -300,8 +319,15 @@ ignore = [
 
     "COM812",   # Missing trailing comma -> conflicts with ISC001
 
+    "PT001",    # Use @pytest.fixture over @pytest.fixture()
+
+    "PERF403",  # Use a dictionary comprehension instead of a for-loop
+
+    "RUF022",   # `__all__` is not sorted
+    "RUF024",   # Do not pass mutable objects as values to `dict.fromkeys`
```
leoll2: I think all these have been resolved in #5003. Can remove.

library/pyproject.toml
```diff
@@ -119,12 +127,20 @@ Documentation = "https://open-edge-platform.github.io/training_extensions/"
...
         { extra = "xpu" },
         { extra = "cuda" },
     ],
+    # redeclare library conflicts to workaround bug with uv dependency resolution
+    # For details, see: https://github.com/astral-sh/uv/issues/9942
+    [
+        { package = "otx", extra = "cpu" },
+        { package = "otx", extra = "xpu" },
+        { package = "otx", extra = "cuda" },
+    ],
```
leoll2: This looks copy-pasted from `backend`, it is not needed in the library.

library/pyproject.toml
ivanzati: Thank you, Leonardo!

library/pyproject.toml
ivanzati: Thank you, Leonardo!

library/pyproject.toml
ivanzati: Thank you, Leonardo!

library/pyproject.toml
ivanzati: Thank you, Leonardo!

.github/workflows/lib-lint-and-test.yaml
ivanzati: yep, will do 3.10 and 3.13 in a matrix

.github/workflows/lib-lint-and-test.yaml
```diff
@@ -153,23 +170,49 @@ jobs:
...
         working-directory: library
         run: |
-          pip install '.[dev]'
-      - name: Run Integration Test
+          uv lock --check
+          uv sync --frozen --extra cuda
+
+      - name: Integration testing
         working-directory: library
-        run: tox r -vv -e integration-test-${{ matrix.task }} -- --task ${{ matrix.task }} --run-category-only
```
leoll2, line 171: Could you please remove `library/tox.ini` if it is no longer used?

.github/workflows/lib-lint-and-test.yaml
```diff
@@ -66,21 +66,42 @@ jobs:
...
+          uv sync --frozen
+
+      - name: Check formatting with ruff
         working-directory: library
-        run: tox r -vv -e pre-commit
+        run: |
+          uv run ruff check --output-format=github .
+          uv run ruff format --check .
+
+      # TODO: Disable mypy for now due to numerous typing issues that need to be resolved
```
leoll2:
```suggestion
      # TODO: Disable mypy for now due to numerous typing issues that need to be resolved
      # https://github.com/open-edge-platform/training_extensions/issues/5006
```

.github/workflows/lib-lint-and-test.yaml
leoll2: Fixed by #5010

.github/workflows/lib-lint-and-test.yaml
```diff
@@ -66,21 +66,43 @@ jobs:
...
+      # https://github.com/open-edge-platform/training_extensions/issues/5006
+      # - name: Check source code with mypy
+      #   working-directory: library
+      #   run: |
+      #     uv run mypy src/otx/ --config-file=pyproject.toml
```
leoll2:
```suggestion
      # TODO: remove 'continue-on-error' after addressing all mypy issues
      # https://github.com/open-edge-platform/training_extensions/issues/5006
      - name: Check source code with mypy
        continue-on-error: true
        working-directory: library
        run: |
          uv run mypy src/otx/ --config-file=pyproject.toml
```
On a second thought, I think it's better to keep this check enabled but allow it to fail.

leoll2 (approved):

.github/workflows/lib-lint-and-test.yaml
leoll2: I pushed a commit to remove it [4939ba5](https://github.com/open-edge-platform/training_extensions/pull/4932/commits/4939ba5a86e02318a54dfb7402dd0cd8cfe6cf76)

AlexanderBarabanov (approved):

PR #4938, opened 2025-10-21
Decouple project feature

jpggvilaca:
### Summary

* Remove unnecessary code
* Move files to more appropriate folder
* Move Label & DatasetItem to shared types
* Refactor project background

### Checklist


application/ui/src/features/project/create/create-project-form.tsx
```diff
@@ -74,8 +74,6 @@ export const CreateProjectForm = ({ numberOfProjects }: { numberOfProjects: numb
             <Flex direction='column' gap='size-300' UNSAFE_style={{ overflow: 'auto', margin: '0 auto' }}>
                 <TaskSelection selectedTask={selectedTask} setSelectedTask={setSelectedTask} />
 
-                <Divider size='S' />
-
                 <LabelSelection labels={labels} setLabels={setLabels} />
```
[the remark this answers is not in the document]
jpggvilaca: No, it looks weird so i removed it

application/ui/mocks/mock-labels.ts
```diff
@@ -1,7 +1,7 @@
 // Copyright (C) 2025 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 
-import { Label } from 'src/features/annotator/types';
+import { Label } from 'src/constants/shared-types';
```
dwesolow, line 4: Wouldn't be better to put it in the "origin/context" where it's the most relevant? For instance imagine having `type ZoomValue` that we'd like to reuse in different places, in that case it makes sense to put it close to the main context. Thoughts?

Edit: I thought it's type from component. In the prompt we decided to export API types from `@geti-prompt/api` (we have ts paths defined + index.ts that is responsible to decide what is exposed from that module).

application/ui/src/features/annotator/annotation-actions-provider.component.tsx
```diff
@@ -8,9 +8,10 @@ import { useProjectIdentifier } from 'hooks/use-project-identifier.hook';
 import { get, isEmpty, isObject } from 'lodash-es';
 import { $api } from 'src/api/client';
 import { components } from 'src/api/openapi-spec';
+import { DatasetItem } from 'src/constants/shared-types';
 import { v4 as uuid } from 'uuid';
 
-import { Annotation, DatasetItem, Shape } from './types';
+import { Annotation, Shape } from './types';
```
dwesolow, line 14: We can help compiler a bit: `import type { ... } from './types';

application/ui/mocks/mock-labels.ts
jpggvilaca: We could go an extra mile and import directly from the api spec. But in this case, this `Label` is shared, so it made sense to me adding it to a "shared-types". It's the usual approach to decouple (or get rid of cyclic dependencies). My goal is to make it so that you can copy paste the feature folder without needing much, maybe a type or a hook, which can be easily copied/replaced.

application/ui/src/features/annotator/annotations/annotation-labels.component.test.tsx
```diff
@@ -2,10 +2,9 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { getMockedLabel } from 'mocks/mock-labels';
+import { Label } from 'src/constants/shared-types';
```
dwesolow, line 5: `import { type ... }`

application/ui/src/features/annotator/annotation-actions-provider.component.tsx
jpggvilaca: i can open a PR to replace all occurrences on the folder. Good point

application/ui/src/features/annotator/annotations/annotation-labels.component.test.tsx
jpggvilaca: same as above

application/ui/src/features/annotator/types.ts
```diff
@@ -16,11 +16,9 @@ export type Polygon = components['schemas']['Polygon'];
 
 export type Shape = Rect | Polygon;
 
-export type Label = components['schemas']['LabelView'];
-
 export type Annotation = components['schemas']['DatasetItemAnnotation-Input'] & {
     id: string;
-    labels: Label[];
+    labels: components['schemas']['LabelView'][];
```
dwesolow, line 21: We can use `Label` type here.

jpggvilaca: mixed feelings, then `types` depends on `shared-types` but ill think about it

application/ui/src/routes/project/sink.tsx
```diff
@@ -1,328 +0,0 @@
-// Copyright (C) 2025 Intel Corporation
```
dwesolow, line 1: Why this file is removed? I don't see it being copied anywhere.

jpggvilaca: not being used

application/ui/src/features/annotator/annotation-actions-provider.component.tsx
dwesolow: Ok, whenever there is a type import u can add `type`. Thanks.

dwesolow (approved):

camiloHimura (approved):

PR #4948, opened 2025-10-22
Introduce `dataset_items_labels` table for efficient label-centric queries

### LINKED ISSUE #4921: Add DB redundancy to query labels efficiently
Geti Tune has a table `dataset_items` to store all metadata about each element of the dataset, including their annotations. The latter is saved as JSON in the `annotation_data` field and it embeds the ids of the labels attached to the annotation.

There are several usecases which need to query the `dataset_items` table and process the list of labels of each item, for example:

- _Dataset filtering_: return the items that have the given label(s)
- _Label removal_: find the items that have the removed label and update their annotation data accordingly
- _Subset assignment_: assign a subset to each item trying to balance the classes

Extracting the label info from the JSON is not only impractical, but also inefficient if the dataset has thousands or more items. As an alternative approach, I propose to add a redundancy in the database schema, namely a collection `dataset_items_labels` where each record is a tuple `(dataset_item_id, label_id)` that indicates that `label_id` is present in `dataset_item_id`. Of course this table must be kept in sync with `dataset_items`, and that can be achieved in two ways: 1) extend the "set/delete annotation" operations to also update this table in the same transaction, or 2) implement a SQLite trigger that syncs `dataset_items_labels` when `annotation_data` changes.

Thanks to this redundancy, the aforementioned usecases would become straightforward to implement.

itallix:
### Summary

- [x] Added `dataset_items_labels` table with `(dataset_item_id, label_id)`
- [x] Updated `DatasetService` to perform inserts/deletes to the new table when required
- [x] Updated integration tests for `DatasetService`

### How to test

1. Start the server with `run.sh`
2. The `dataset_items_labels` table will be created and will populate based on the default data collection policy from the seed setup

### Checklist


application/backend/app/db/schema.py
```diff
@@ -119,3 +120,12 @@ class LabelDB(BaseID):
...
     hotkey: Mapped[str | None] = mapped_column(String(10), nullable=True)
+
+
+class DatasetItemLabelDB(Base):
+    __tablename__ = "dataset_items_labels"
+
+    dataset_item_id: Mapped[str] = mapped_column(
+        Text, ForeignKey("dataset_items.id", ondelete="CASCADE"), primary_key=True
+    )
+    label_id: Mapped[str] = mapped_column(Text, ForeignKey("labels.id", ondelete="CASCADE"), primary_key=True)
```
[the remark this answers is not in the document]
itallix: False

application/backend/app/repositories/dataset_item_repo.py
```diff
@@ -140,3 +140,13 @@ def set_subset(self, obj_id: str, subset: str) -> None:
             )
         )
         self.db.execute(stmt)
+
+    def save_labels(self, dataset_item_id: str, label_ids: set[str]) -> None:
+        dataset_item_label_db = [
+            DatasetItemLabelDB(dataset_item_id=dataset_item_id, label_id=label_id) for label_id in label_ids
+        ]
+        self.db.add_all(dataset_item_label_db)
```
leoll2: On annotation update (`set_dataset_item_annotations`), the `dataset_items_labels` table may already contain records for this dataset item. If I'm not wrong, `add_all` will create duplicate records for such labels. I think you should first call `delete_labels` to remove all existing records associated to this dataset item.

Please also add a test to verify this scenario.

itallix: Handled with `insert().on_conflict_do_nothing(index_elements=["dataset_item_id", "label_id"])`

leoll2: What about this scenario:
1. Annotate item X with labels A, B
2. Update annotation of item X with labels B, C

After step (1), the`dataset_items_labels`  should contain two records: `[(X, A), (X, B)]`. After step (2), the desired state is `[(X, B), (X, C)]`, but the current implementation would result in `[(X, A), (X, B), (X, C)]` because it doesn't remove the label A.

itallix: Indeed, the code was trying to perform the update rather then setting from the clean state.

I've renamed the method from `save_labels` to `set_labels` and fixed this behavior. I think existing tests already capture it.

leoll2 (approved):

PR #4956, opened 2025-10-24
Refactoring: Event Bus & services decoupling

warrkan:
This is the first PR to refactor services code. This PR addresses following addresses following issues:

1. Decouple `ProjectDB` and `PipelineDB`. Similar to Labels, it's better to manage independent entities separately. Moved creation of pipeline for new project to the service layer instead of repo, because it's business logic.
2. Reduce number of dependencies for `PipelineService`. Metrics functionality has been moved to a separate `PipelineMetricsService`. 
3. Introduced `EventBus` to decouple `PipelineService` and `ConfigurationService` from services\components which are interested in configuration and pipeline updates.
4. Removed `ActivePipelineService` because each `EventBus` listener doesn't need all the data from pipeline. Each listener re-reads only that chunk which is needed (i.e. `StreamLoader` only reloads source information). Additionally this service had multiple usage scenarios (in main process or in child process) and ways how it's being initialized quite differ. 
5. Moved source change condition to event bus. This condition is used to notify child processes (`StreamLoader`) about source change. Additionally added `sink_changed_condition` which is not used at the moment.

### Checklist


application/backend/app/lifecycle.py
```diff
@@ -65,14 +67,21 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None]:
         logger.error("Failed to initialize database. Application cannot start.")
         raise RuntimeError("Database initialization failed")
 
-    active_pipeline_service = ActivePipelineService()
-    app.state.active_pipeline_service = active_pipeline_service
+    # Condition to notify processes about source updates
+    source_changed_condition: Condition = mp.Condition()
 
-    data_collector = DataCollector(data_dir=settings.data_dir, active_pipeline_service=active_pipeline_service)
+    event_bus = EventBus(source_changed_condition=source_changed_condition)
```
leoll2, line 73: Missing `sink_changed_condition`?

application/backend/app/lifecycle.py
```diff
@@ -65,14 +67,21 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None]:
...
+    app.state.event_bus = event_bus
+
+    data_collector = DataCollector(data_dir=settings.data_dir, event_bus=event_bus)
     app.state.data_collector = data_collector
 
     # Initialize Scheduler
-    app_scheduler = Scheduler(active_pipeline_service=active_pipeline_service, data_collector=data_collector)
+    app_scheduler = Scheduler(
+        source_changed_condition=source_changed_condition,
+        event_bus=event_bus,
```
leoll2: The `source_changed_condition` is already available through `event_bus`. Can you reduce the parameter list of `Scheduler` to `event_bus` and `data_collector` only, so that we don't have to modify its signature everytime we add a new condition variable?

application/backend/app/workers/dispatching.py
```diff
@@ -27,48 +29,43 @@ class DispatchingWorker(BaseThreadWorker):
...
 
-    def _reset_sink_if_needed(self, sink_config: Sink) -> None:
-        if not self._prev_sink_config or sink_config != self._prev_sink_config:
-            logger.debug("Sink config changed from %s to %s", self._prev_sink_config, sink_config)
-            self._destinations = DispatchService.get_destinations(output_configs=[sink_config])
-            self._prev_sink_config = copy.deepcopy(sink_config)
+    def _load_sink(self):
+        with get_db_session() as db:
+            sink = SinkService(db).get_active_sink()
+        self._sink = sink if sink is not None else DisconnectedSinkConfig()
```
leoll2: The attribute `self._sink` does not appear explicitly in the constructor (although this function is called from the constructor). For readability, I suggest to slightly refactor like this:
```python
class DispatchingWorker(BaseThreadWorker, Listener):
    def __init__(self):
        ...
        self._sink: Sink
        self._destinations: list[Dispatcher]
        self._sink, self._destinations = self._load_sink()

    def _load_sink() -> tuple[Sink, list[Dispatcher]]:
        ...
        return sink, destinations

    def on_sink_changed(self) -> None:
        self._sink, self._destinations = self._load_sink()
```

application/backend/app/lifecycle.py
warrkan: We do not use it at the moment, it was added to `event_bus` in sake of consistency with `source_changed_condition`

leoll2 (dismissed):

application/backend/app/services/event/event_bus.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+        Callback method for pipeline status changes.
+        """
+
+
+class EventBus:
+    def __init__(
+        self, source_changed_condition: Condition | None = None, sink_changed_condition: Condition | None = None
+    ) -> None:
+        self._listeners: list[Listener] = []
```
itallix: The current `Listener` and `EventBus` design is tightly coupled to specific domain events. Every new event type will require modifying both the `Listener` base class and `EventBus`.

I would propose the following changes here:
1. Replace `Listener` class with event-handler mapping: Instead of requiring listeners to implement specific callback methods, allow subscribers to register handlers for specific event types using `dict[EventType, Callable]`.
2. Define events as `EventType` enum values (e.g., `SOURCE_CHANGED`, `SINK_CHANGED`) rather than hardcoded method names. New events can be added by extending the enum without modifying core classes.
3. Add thread safety for `_listeners` with `RLock` since it will be modified from dispatching thread as well.

Example:

```python
bus = EventBus()
bus.subscribe(EventType.SOURCE_CHANGED, handle_source_change)
bus.emit(EventType.SOURCE_CHANGED)
```

warrkan: That's a good point, thanks! I've replaced Listener with event types

application/backend/app/services/event/event_bus.py
```diff
@@ -1,42 +1,28 @@
...
+    PIPELINE_DATASET_COLLECTION_POLICIES_CHANGED = "PIPELINE_DATASET_COLLECTION_POLICIES_CHANGED"
+    PIPELINE_STATUS_CHANGED = "PIPELINE_STATUS_CHANGED"
 
 
 class EventBus:
     def __init__(
         self, source_changed_condition: Condition | None = None, sink_changed_condition: Condition | None = None
     ) -> None:
-        self._listeners: list[Listener] = []
+        self._event_handlers: dict[EventType, list[Callable]] = {}
```
itallix: If we init it as `defaultdict(list)`, there won't be need to use `.setdefault(event_type, [])` in `subscribe`

application/backend/app/services/event/event_bus.py
```diff
@@ -45,33 +31,21 @@ def source_changed_condition(self) -> Condition | None:
...
+    def subscribe(self, event_types: Sequence[EventType], handler: Callable) -> None:
+        with self._lock:
+            for event_type in event_types:
+                self._event_handlers.setdefault(event_type, []).append(handler)
+
+    def emit_event(self, event_type: EventType) -> None:
+        with self._lock:
+            for handler in self._event_handlers.get(event_type, []):
+                handler()
+        if (
```
itallix, line 44: Shall we extract condition-specific logic bounded to our domain into child class?

warrkan: Let's keep it as is for now, if the class grows further, we can figure out new design for the messaging

itallix (approved):

PR #4957, opened 2025-10-24
Evaluation metrics

### LINKED ISSUE #4865: Implement evaluation metrics
Implement methods to compute the necessary metrics to evaluate a trained model on a dataset.

The metrics to expose depend on the task type:

- **Multiclass Classification**: Accuracy, Precision, Recall, F1
- **Multilabel Classification**: Subset Accuracy, Hamming Loss, Precision, Recall, F1, mAP
- **Detection**: mAP@0.5-0.95 (COCO style), mAP@0.5
- **Instance segmentation**: mAP@0.5-0.95 (COCO style), mAP@0.5

leoll2:
### Summary

Changes:
- New `EvaluationService` that returns an appropriate `Evaluator` object depending on the dataset type.
  - Evaluators compute (lazily) a set of metrics; the available metrics depend on the evaluator itself, hence on the task:
    - **Multiclass classification**: precision, recall, accuracy, F1, confusion matrix
    - **Multilabel classification**: precision, recall, subset accuracy, F1
    - **Detection**: mAP (including `@0.5` and `@0.5-0.95`)
    - **Instance segmentation**: mAP (including `@0.5` and `@0.5-0.95`)
- Extend dataset `Sample` classes with an `image_info` field, to include the width/height properties of the image.
- Bugfix: `MultilabelClassificationSample` was incorrectly using `is_list` instead of `multi_label`.

_Note_: multilabel mAP will be implemented later, after a score field is added to prediction datasets (depends on https://github.com/open-edge-platform/datumaro/pull/1931).

Resolves #4865

### Checklist


application/backend/app/services/evaluation_service.py
```diff
@@ -0,0 +1,298 @@
...
+
+class DetectionEvaluator(MeanAveragePrecisionEvaluator):
+    """Evaluator for object detection tasks."""
+
+
+class InstanceSegmentationEvaluator(MeanAveragePrecisionEvaluator):
+    """Evaluator for instance segmentation tasks."""
+
+
+EvaluatorT = (
```
itallix: Let's omit `T` prefix for union types - just `Evaluator` is good enough here.

application/backend/app/services/evaluation_service.py
```diff
@@ -0,0 +1,298 @@
...
+    MultiClassClassificationEvaluator
+    | MultiLabelClassificationEvaluator
+    | DetectionEvaluator
+    | InstanceSegmentationEvaluator
+)
+
+
+class EvaluationService:
```
itallix: This looks more like a factory to me.

We can split it into 2 modules: `evaluators.py` and `factory.py` under `evaluation` package.
`EvaluatorFactory` can hold a registry `dict[dtype, type[Evaluator]]` to make factory method more compact.
Optionally, the 3rd module can be `base.py`.

leoll2: Thanks for the great suggestions, I have applied them

application/backend/app/services/evaluation/factory.py
```diff
@@ -0,0 +1,39 @@
...
+    """Factory to get a suitable evaluator for a given set of ground truth and predictions datasets."""
+
+    _registry: dict[type[Sample], type[Evaluator]] = {
+        ClassificationSample: MultiClassClassificationEvaluator,
+        MultilabelClassificationSample: MultiLabelClassificationEvaluator,
+        DetectionSample: DetectionEvaluator,
+        InstanceSegmentationSample: InstanceSegmentationEvaluator,
+    }
+
+    @staticmethod
```
itallix: Can we change it to `classmethod` and access class attribute without hardcoding the class name?

itallix (approved):

application/backend/app/services/evaluation/factory.py
```diff
@@ -0,0 +1,39 @@
...
+    _registry: dict[type[Sample], type[Evaluator]] = {
+        ClassificationSample: MultiClassClassificationEvaluator,
+        MultilabelClassificationSample: MultiLabelClassificationEvaluator,
+        DetectionSample: DetectionEvaluator,
+        InstanceSegmentationSample: InstanceSegmentationEvaluator,
+    }
+
+    @classmethod
+    def get_evaluator(cls, predictions_dataset: Dataset, ground_truth_dataset: Dataset) -> Evaluator:
+        if predictions_dataset.dtype != ground_truth_dataset.dtype:
```
A-Artemis, line 34: Comment for Datumaro (@AlbertvanHouten ), having it called `.dtype` can be confusing as this can be seen as `np.dtype` which are the classic `int8, float32, ...` and not what you would think as `ClassificationSample`.

`.stype`?
`.ptype`?

application/backend/app/services/evaluation/evaluators.py
```diff
@@ -0,0 +1,254 @@
...
+        coco_dataset_dict["images"].append(
+            {
+                "id": image_id,
+                "file_name": sample.image,
+                "width": sample.image_info.width,
+                "height": sample.image_info.height,
+            }
+        )
+
+        # Detection
```
A-Artemis, line 51: I assume this is for classification as well for now?

A-Artemis (approved):

application/backend/app/services/evaluation/evaluators.py
leoll2: The COCO format is not designed for the classification task. In a later PR, we'll need to implement mAP for multi-label classification, but for such task it can be easily done with sklearn `average_precision_score` (without `faster_coco_eval`).

PR #4963, opened 2025-10-28
 Extend EfficientNet pretrained support to B2-B8

itrushkin:
### Summary

Adds support for downloading and using pretrained EfficientNet weights for models larger than B1 (B2-B8).

In pytorchcv, the “c” denotes the community re-trained EfficientNet variants (ported via https://www.github.com/rwightman/timm) that use PyTorch-native ops and shapes, distinct from the original TF-trained baselines.

### Checklist

> [!WARNING]  
> This PR may increase test runtime because it involves downloading large models.

### DIFF
--- a/library/src/otx/backend/native/models/classification/backbones/efficientnet.py
+++ b/library/src/otx/backend/native/models/classification/backbones/efficientnet.py
@@ -672,6 +672,6 @@ class EfficientNetBackbone:
 
         if pretrained:
             cache_dir = Path.home() / ".cache" / "torch" / "hub" / "checkpoints"
-            download_model(net=model, model_name=f"{model_name}", local_model_store_dir_path=str(cache_dir))
+            download_model(net=model, model_name=f"{model_name}c", local_model_store_dir_path=str(cache_dir))
             print(f"Download model weight in {cache_dir!s}")
         return model
--- a/library/tests/unit/backend/native/models/classification/backbones/test_otx_efficientnet.py
+++ b/library/tests/unit/backend/native/models/classification/backbones/test_otx_efficientnet.py
@@ -23,8 +23,9 @@ class TestOTXEfficientNet:
             "efficientnet_b8",
         ],
     )
-    def test_forward(self, model_name):
-        model = EfficientNetBackbone(model_name, pretrained=None)
+    @pytest.mark.parametrize("pretrained", [True, False])
+    def test_forward(self, model_name, pretrained):
+        model = EfficientNetBackbone(model_name, pretrained=pretrained)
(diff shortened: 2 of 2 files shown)


kprokofi:
Since we used base models, we need to be sure that "c" variants are backward compatible and the performance of the b0 backbone remains same or better.
Could you provide these evidence?

itrushkin:
Hi @kprokofi,

To confirm backward compatibility: all existing unit tests continue to pass, and model calls with dummy inputs execute successfully, indicating that the integration of the "c" variants doesn’t break existing functionality.

The performance improvements for the "c" variants come from the [AdvProp paper](https://arxiv.org/abs/1911.09665). I haven't independently benchmarked the b0 backbone myself, but the original authors state that these models match or exceed the baseline performance.


<p align="center">
  image
</p>

library/src/otx/backend/native/models/classification/backbones/efficientnet.py
```diff
@@ -672,6 +672,6 @@ def __new__(
 
         if pretrained:
             cache_dir = Path.home() / ".cache" / "torch" / "hub" / "checkpoints"
-            download_model(net=model, model_name=f"{model_name}", local_model_store_dir_path=str(cache_dir))
+            download_model(net=model, model_name=f"{model_name}c", local_model_store_dir_path=str(cache_dir))
```
sovrasov, line 675: Thanks Ilya for the contribution. I think here we can not just replace this weights without a set of comprehensive experiments. 100% positive correlation with IN-top1 is not given, there could be sudden outlier which we'd like to avoid. 

To tackle that, I'd propose moving this version parameter to `init` of `EfficientNetBackbone` and forwarding it from the endpoint classes (`EfficientNetMulticlassCls` etc). Once done, the model version is configurable from model recipe yaml file. Corner case of b0 and b1 should be handled.

library/tests/unit/backend/native/models/classification/backbones/test_otx_efficientnet.py
```diff
@@ -23,8 +23,9 @@ class TestOTXEfficientNet:
             "efficientnet_b8",
         ],
     )
-    def test_forward(self, model_name):
-        model = EfficientNetBackbone(model_name, pretrained=None)
+    @pytest.mark.parametrize("pretrained", [True, False])
```
sovrasov, line 26: Please, don't download anything here to avoid increase of UTs duration and reliability. You could mock `download_model` if the idea is to check that the URL is correct somehow.

kprokofi:
Not a part of our roadmap

PR #4965, opened 2025-10-28
Implement subset splitting

### LINKED ISSUE #4925: Training job: subset splitting
In the data preparation step of the training job, implement a logic to assign a subset to every item of the dataset that doesn't already have one, trying to ensure a balanced representation of each label in every subset. The algorithm should behave as follows:

1. Get the list of dataset items that don't have a subset (subset = unassigned), along with their labels, from the database
2. Compute the split ratio to use for these items, taking into account both the current split (of the already assigned items) and the target ratio defined by the training configuration.
3. Assign each unannotated item to a subset. You can use [skmultilearn.iterative_stratification](http://scikit.ml/api/skmultilearn.model_selection.iterative_stratification.html) to determine the optimal assignment; notably, this algorithm works for multilabel problems too.
4. Persist the subset assignment in the database.

itallix:
### Summary

This PR implements subset splitting based on iterative stratification algorithm from [skmultilearn](https://github.com/scikit-multilearn/scikit-multilearn)

Additionally: 
- introduced `step` decorator and refactored the `OTXTrainer` class.
- introduced `ProjectTestDataFactory` to perform project setup in the database based on predefined fixtures (currently used only in `SubsetService` & `ModelService` integration tests, but should be adopted further)

### How to test

1. **Start the server.** Run `./run.sh` and wait for the collection policy to capture dataset items in the database (algorithm expects number of samples at least the same as number of folds = 3).
2. **Trigger a training job.** Execute the following curl command:
  ```bash
  curl -X POST -H "Content-Type: application/json" http://localhost:7860/api/jobs -d '{"job_type": "train", "project_id": "9d6af8e8-6017-4ebe-9126-33aae739c5fa", "parameters": { "model_architecture_id":
   "Custom_Object_Detection_YOLOX" }}'
  ```
3. **Verify training outcomes.** Check that dataset items are assigned to subsets (train/val/test) based on the target split.

### Checklist


application/backend/app/core/models/dataset_item.py
```diff
@@ -0,0 +1,16 @@
...
+class DatasetItemFormat(StrEnum):
+    JPG = "jpg"
+    PNG = "png"
+
+
+class DatasetItemSubset(StrEnum):
+    UNASSIGNED = "unassigned"
+    TRAINING = "training"
+    VALIDATION = "validation"
+    TESTING = "testing"
```
leoll2: These are domain-specific definitions. As you mentioned in [this comment](https://github.com/open-edge-platform/training_extensions/pull/4964#pullrequestreview-3389597724), shouldn't we keep `app/core` domain-agnostic?

application/backend/pyproject.toml
```diff
@@ -24,6 +24,7 @@ dependencies = [
     "datumaro[experimental] @ git+https://github.com/open-edge-platform/datumaro.git@develop",
     "scikit-learn~=1.7.0",
     "faster-coco-eval~=1.6.8",
+    "scikit-multilearn>=0.2.0",
```
leoll2:
```suggestion
    "scikit-multilearn~=0.2.0",
```
Even if this package hasn't been updated for a long time, it's better to use `~=` than `>=` to avoid problems if they release an API-breaking 1.0 version.

application/backend/app/services/training/base.py
```diff
@@ -2,33 +2,89 @@
...
+    """Composable pipeline of training steps."""
+
+    def __init__(self, steps: list[TrainingStep]):
+        self._steps = steps
+
+    def execute(self, ctx: ExecutionContext, params: TrainingParams) -> None:
+        pipeline_ctx = PipelineContext()
+        for step in self._steps:
+            ctx.report_progress(f"Starting: {step.get_name()}")
+            step.execute(ctx, params, pipeline_ctx)
```
leoll2: How do the steps exchange data with each other, e.g. if one step needs to consume the output of a previous step? Do they push the result to the pipeline context? I see a few possible problems with this approach:

1. Weak type validation: the pipeline context is an unstructured dictionary, mypy can't verify that steps push/pull data from the context in the right format.
2. Devs mental load: developers need to be aware of what's stored in the context at any moment
3. Not memory friendly: the context dict holds a reference to every object stored in the context and keeps them alive (in memory) until the job finishes, even if those objects are no longer necessary to later steps. In other words, memory keeps growing step after step, increasing the risk of OOM.

I would suggest a different approach to model the steps: each step is a function decorated with `@step(name=...)`. As a regular function, each step can have a custom signature. The decorator helps with progress reporting and other cross-cutting functionalities.

Example:
```python
@step(name="Prepare dataset")
def prepare_dataset() -> Dataset:
    ...
    return dataset

@step(name="Train")
def train(train_dataset: Dataset) -> Model
    ...
    return model

class OTXTrainer(Trainer):
    def run(self, ...) -> None:
        train_dataset = prepare_dataset()
        trained_model = train(train_dataset)
        ...
```

application/backend/app/core/models/dataset_item.py
itallix: Yes, I will wait for #4964 and rebase the code.

application/backend/app/services/training/base.py
itallix: I agree. And it's a very good use case for decorators. Thanks for suggestion!

itallix: @leoll2 I'm considering defining steps within the concrete trainer class to reduce the number of explicitly passed dependencies (currently passed via constructor).
Two possible approaches:
1. **Explicit composition in concrete classes**
Each concrete trainer implements `run()` and explicitly composes its steps (similar to what you mentioned in the comment above)
Pros: Simple and straightforward
Cons: Will lead to duplication if we introduce multiple trainer variants
2. **Convention-based step discovery in base class**
The base `Trainer` class discovers and orchestrates steps defined in child classes. Steps would be marked with a decorator like `@step(name: str, output: str | None)`, and the base class would handle ordering and dependency injection.
Pros: Steps have direct access to trainer state, eliminating explicit dependency passing; reduces duplication across trainer variants
Cons: Adds complexity to the base class through reflection/introspection

I'm leaning toward approach 2, but would appreciate your input.

leoll2: Personally I prefer option (1) to keep it simple for now. We don't trainer variants in the current roadmap; we can always revisit option (2) later if it becomes worth it.

itallix: Ok, let's keep it simple.

application/backend/app/repositories/dataset_item_repo.py
```diff
@@ -153,3 +155,24 @@ def set_labels(self, dataset_item_id: str, label_ids: set[str]) -> None:
...
+
+    def list_unassigned_items(self) -> list[DatasetItemLabelDB]:
+        stmt = (
+            select(DatasetItemLabelDB)
+            .join(DatasetItemDB)
+            .where(
+                DatasetItemDB.project_id == self.project_id,
+                DatasetItemDB.subset == DatasetItemSubset.UNASSIGNED,
+            )
+            .order_by(DatasetItemDB.created_at.asc())
```
leoll2: What's the purpose of `order_by` in this query?

application/backend/uv.lock
```diff
@@ -11,9 +11,9 @@ resolution-markers = [
 name = "aiofiles"
 version = "24.1.0"
 source = { registry = "https://pypi.org/simple" }
-sdist = { url = "https://files.pythonhosted.org/packages/0b/03/a88171e277e8caa88a4c77808c20ebb04ba74cc4681bf1e9416c862de237/aiofiles-24.1.0.tar.gz", hash = "sha256:22a075c9e5a3810f0c2e48f3008c94d68c65d763b9b03857924c99e57355166c", size = 30247, upload-time = "2024-06-24T11:02:03.584Z" }
```
leoll2, line 14: What version of `uv` do you use that causes `upload-time` to be removed from all lines? Perhaps we should set `required-version` as suggested [here](https://github.com/astral-sh/uv/issues/15220#issuecomment-3175133991).

itallix:
```sh
❯ uv version
uv 0.6.12 (e4e03833f 2025-04-02)
```

which one is recommended to use?

application/backend/app/repositories/dataset_item_repo.py
itallix: When it comes to listings, it's a good idea to apply default sorting to ensure deterministic results on the client side. However, the current use case doesn't require deterministic ordering. I can remove it, but keeping it wouldn't be harmful.

application/backend/uv.lock
leoll2: CI uses 0.8.8: https://github.com/open-edge-platform/training_extensions/blob/develop/.github/workflows/backend-lint-and-test.yaml#L77

itallix: Added `required-version = "~=0.8.8"` to `pyproject.toml` and updated lock file.

application/backend/app/services/training/subset_assignment/distribution.py
```diff
@@ -0,0 +1,54 @@
...
+
+        # Calculate how many more items are needed in each subset
+        needed = {
+            subset: max(0, target_counts[subset] - self.get_count(subset))
+            for subset in DatasetItemSubset
+            if subset != DatasetItemSubset.UNASSIGNED
+        }
+
+        total_needed = sum(needed.values())
+        if total_needed == 0:
```
leoll2: In what scenario does this condition evaluate true? Isn't `total_needed` always equal to `unassigned_count` by construction, since we want to assign a subset to every unassigned item? If this property holds true, I suggest to assert it and raise an error in case of mismatch, that indicates a bug in the splitting logic.

application/backend/app/services/training/subset_assignment/distribution.py
```diff
@@ -0,0 +1,54 @@
...
+        return self._counts.get(subset, 0)
+
+    def compute_adjusted_ratios(self, target_ratios: SplitRatios, unassigned_count: int) -> SplitRatios:
+        """
+        Compute adjusted ratios for unassigned items to reach target distribution.
+
+        This method calculates how unassigned items should be distributed to achieve the overall target ratios.
+        """
+        if unassigned_count == 0:
+            return target_ratios
```
leoll2, line 31:
```suggestion
        if unassigned_count < 0:
            raise ValueError(f"Parameter 'unassigned_count' cannot be negative (got {unassigned_count})")

        if unassigned_count == 0:
            return target_ratios
```
Just some extra validation, for safety

application/backend/app/services/training/subset_assignment/assigner.py
```diff
@@ -0,0 +1,72 @@
...
+        splits = list(stratifier.split(X, y))
+        train_idx = splits[0][1]
+        val_idx = splits[1][1]
+        test_idx = splits[2][1]
+
+        indices_by_subset = {
+            DatasetItemSubset.TRAINING: train_idx,
+            DatasetItemSubset.VALIDATION: val_idx,
+            DatasetItemSubset.TESTING: test_idx,
+        }
```
leoll2: Can you add an explicit type-hint to this dictionary? I think it's `dict[DatasetItemSubset, list[int]]`, although the variable name `train_idx` sounds like an int rather than a list.

application/backend/tests/integration/project_factory.py
```diff
@@ -0,0 +1,155 @@
...
+        >>> project = factory.with_project(ProjectDB(name="Test")).build()
+
+        Project with pipeline and models:
+        >>> project = (
+        ...     ProjectTestDataFactory(db_session)
+        ...     .with_project(ProjectDB(name="Test"))
+        ...     .with_pipeline(is_running=True)
+        ...     .with_models([ModelRevisionDB(...)])
+        ...     .build()
+        ... )
```
leoll2, line 35: I like this factory class, however I'm not sure if I like the fact that `with_xyz` methods take complex objects/fixtures as argument. For example, `ModelRevisionDB` has a `project_id` which may differ from the one in `ProjectDB`, leading to internal inconsistency in the final project. It's powerful, we just need to be careful I guess. A possible way to mitigate inconsistencies between fixtures is to create global fixtures like `fxt_project_id` with a static UUID, and build the other fixtures on top of these predefined ids.

application/backend/tests/integration/services/training/steps/subset_assignment/test_subset_service.py
```diff
@@ -0,0 +1,118 @@
...
+        assert (
+            db_session.scalar(
+                select(func.count())
+                .select_from(DatasetItemDB)
+                .where(DatasetItemDB.subset == DatasetItemSubset.UNASSIGNED)
+            )
+            == 0
+        )
+
+        # Verify that all items have been assigned to TRAINING subset
```
leoll2: To make the test more robust, you could use an assignment with at least two distinct subsets.

leoll2 (commented):
A few minor comments, LGTM

application/backend/app/services/training/subset_assignment/distribution.py
itallix: It's just a safeguard from `ZeroDivision` when calculating normalized values below, but indeed it shouldn't be 0 since we've already verified `unassigned_count` above. I think we are safe to remove this check.

application/backend/tests/integration/project_factory.py
itallix: We've started to produce some duplication in our integration tests around project setup, so I thought it would be helpful to introduce a factory pattern that can be reused. The `build` call should return the project along with all related entities, since they inherit the `project_id` of the project that's being built by the client.

I agree we should be cautious about extending this pattern further - it's a convention without any automated enforcement.

application/backend/app/repositories/dataset_item_repo.py
itallix: The intention was to order `DatasetItemLabelDB`, not `DatasetItemDB`, however this would require additional index. Let's remove it as the client doesn't care here.

leoll2 (approved):

PR #4973, opened 2025-10-31
Refactoring: Entities - Sinks

### LINKED ISSUE #4983: Introduce Sink service layer entities

warrkan:
### Checklist


application/backend/app/api/routers/sinks.py
```diff
@@ -95,50 +95,60 @@ def create_sink(
...
 )
 def get_sink(
     sink_id: Annotated[UUID, Depends(get_sink_id)],
     configuration_service: Annotated[ConfigurationService, Depends(get_configuration_service)],
-) -> Sink:
+) -> SinkView:
     """Get info about a sink"""
     try:
-        return configuration_service.get_sink_by_id(sink_id)
+        sink = configuration_service.get_sink_by_id(sink_id)
```
itallix: 1. Shall we define a dependency `get_sink` similar to `get_project` and reuse in routes?
2. `export_sink` still performs check `if not sink` which is not necessary anymore.

application/backend/app/api/routers/sinks.py
```diff
@@ -154,12 +164,22 @@ def update_sink(
...
-) -> Sink:
+) -> SinkView:
     """Reconfigure an existing sink"""
     if "sink_type" in sink_config:
         raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail="The 'sink_type' field cannot be changed")
     try:
-        return configuration_service.update_sink(sink_id, sink_config)
+        sink = configuration_service.get_sink_by_id(sink_id)
+        updated_sink: Sink = sink.model_copy(update=sink_config)
+        updated_sink.config_data = updated_sink.config_data.model_copy(update=sink_config)
```
itallix, line 165: 1. Shall we introduce a separate schema for update operations? Now, the client can request any payload in `PATCH` and it will be silently ignored, as `model_copy` would not complain. 
2. We check resource existence twice now: on service layer and here in routes. This can be probably done only once on the service layer.

application/backend/app/services/dispatchers/mqtt.py
```diff
@@ -63,6 +66,20 @@ def __init__(
         self.client = mqtt_client or self._create_default_client()
         self._connect()
 
+    @staticmethod
+    def _get_credentials(auth_required: bool) -> tuple[str | None, str | None]:
```
itallix: Why not keeping this in the model `MqttConfig` as a property? In theory, it can be reused in other places.

application/backend/app/api/routers/sinks.py
warrkan: Nice idea! Added

application/backend/app/api/routers/sinks.py
warrkan: I'd add it as a future improvement. Since we have a separate schema for each sink type, it can be not very straightforward.

application/backend/app/services/dispatchers/mqtt.py
warrkan: I'd keep model as simple as possible and not add extra responsibilities there, like getting credentials from env vars.

itallix: It still treated as a part of configuration, rather than dispatcher. Env vars context is available from everywhere. 
In case we keep it here, I suggest we do not hardcode a class name.

itallix (approved):

jpggvilaca (approved):

PR #4978, opened 2025-11-03
Refactoring: Entities - Sources

### LINKED ISSUE #4984: Introduce Source service layer entities

warrkan:
### Checklist


application/backend/app/api/routers/sources.py
```diff
@@ -169,13 +169,20 @@ def update_source(
...
+) -> SourceView:
     """Reconfigure an existing source"""
     if "source_type" in source_config:
         raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail="The 'source_type' field cannot be changed")
+
     try:
-        source = source_service.get_by_id(source_id)
-        return source_service.update(source, source_config)
+        updated_source: Source = source.model_copy(update=source_config)
+        updated_source.config_data = updated_source.config_data.model_copy(update=source_config)
```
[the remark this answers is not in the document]
warrkan: For now it's an acceptable approach, we'll introduce more strict update schema later on.

leoll2 (dismissed):

application/backend/app/api/schemas/sink.py
```diff
@@ -33,7 +42,8 @@ class DisconnectedSinkConfigView(DisconnectedSinkConfig):
 class FolderSinkConfigView(FolderSinkConfig):
     config_data: FolderConfig = Field(exclude=True)
 
-    @computed_field
+    @computed_field  # type: ignore[prop-decorator]
```
leoll2: Assuming this `# type: ignore[prop-decorator]` is due to [this mypy issue](https://github.com/python/mypy/issues/14461), which I think is solved in newer mypy versions, I suggest to blacklist this error in `pyproject.toml` instead of inline in so many places, with a comment explaining that it can be re-enabled once upgrading to a newer mypy version.

application/backend/app/services/source_service.py
```diff
@@ -2,58 +2,102 @@
...
+            raise ResourceWithIdAlreadyExistsError(ResourceType.SOURCE, str(source_id))
         except UniqueConstraintIntegrityError:
-            raise ResourceWithNameAlreadyExistsError(ResourceType.SOURCE, item.name)
+            raise ResourceWithNameAlreadyExistsError(ResourceType.SOURCE, name)
 
     @parent_process_only
-    def update(self, source: Source, partial_config: dict) -> Source:
+    def update_source(
+        self,
+        source: Source,
```
itallix: Is there a reason why we don't pass just `source_id` here similar to `create_source` method?

application/backend/app/services/source_service.py
```diff
@@ -2,58 +2,102 @@
...
+    def list_all(self) -> list[Source]:
+        return [
+            SourceAdapter.validate_python(db_source, from_attributes=True)
+            for db_source in SourceRepository(self._db_session).list_all()
+        ]
 
     @parent_process_only
-    def delete_by_id(self, item_id: UUID) -> None:
-        super().delete_by_id(item_id)
+    def delete_source(self, source: Source) -> None:
```
itallix, line 65: Same here, any reason not to use just an `ID` like in original signature?

application/backend/app/services/source_service.py
```diff
@@ -2,58 +2,102 @@
...
+    ResourceWithIdAlreadyExistsError,
     ResourceWithNameAlreadyExistsError,
-    ServiceConfig,
 )
 from .event.event_bus import EventBus, EventType
-from .mappers import SourceMapper
 from .parent_process_guard import parent_process_only
 
 
-class SourceService(GenericPersistenceService[Source, SourceRepository]):
```
itallix, line 23: We can probably drop `GenericPersistenceService` now?
The idea was to reduce duplication, but it doesn't generalize well and adds some unnecessary cognitive load.

application/backend/app/services/source_service.py
warrkan: Idea here is that you need to pass `Source` object here which should be obtained first from the DB. It also ensures that such a source exists and raises `ResourceNotFoundError` otherwise.

application/backend/app/services/source_service.py
warrkan: Ah, right. It's not used anymore, deleted

leoll2 (approved):

itallix (approved):

jpggvilaca (approved):

application/backend/app/models/source.py
```diff
@@ -0,0 +1,112 @@
...
+class IPCameraSourceConfig(BaseSourceConfig):
+    source_type: Literal[SourceType.IP_CAMERA]
+    config_data: IPCameraConfig
+
+
+class VideoFileConfig(SourceConfig):
+    video_path: str
+
+
+class VideoFileSourceConfig(BaseRequiredIDNameModel):
```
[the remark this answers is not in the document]
warrkan: Fixed

PR #4998, opened 2025-11-11
Create training dataset

### LINKED ISSUE #4927: Training job: create training dataset
In the training job, create the training `Dataset` and the relative training/validation/testing subsets.

The full `Dataset` can be obtained via `DatasetService.get_dm_dataset`. Make sure to only include items with `user_reviewed=True`. To extract the subsets from a `dm.Dataset`, you can use the `filter_by_subset` method implemented in https://github.com/open-edge-platform/datumaro/pull/1933.

In another PR, each of the subset `dm.Dataset` will be used to initialize an [OTXDataset](https://github.com/open-edge-platform/training_extensions/blob/develop/library/src/otx/data/dataset/base.py#L59) and finally an `OTXDataModule` through [from_otx_datasets](https://github.com/open-edge-platform/training_extensions/blob/715bbb5e1d3a72fb43b431b212676229a99f9b35/library/src/otx/data/module.py#L224).

itallix:
### Summary

This PR adds dataset creation step to the training pipeline using the new Datumaro experimental API.

Additionally:
- [x] Introduced a new `BaseSessionManagedService` with db session setter for late db initialization (used in otx trainer)
- [x] Added a subset information to type-based Sample classes to allow filtering

### Checklist


application/backend/app/api/routers/jobs.py
```diff
@@ -66,6 +66,7 @@ async def submit_job(
                         model_architecture_id=job_request.parameters.model_architecture_id,
                         parent_model_revision_id=job_request.parameters.parent_model_revision_id,
                         task_type=project.task.task_type,
+                        exclusive_labels=project.task.exclusive_labels,
```
leoll2: It's probably more convenient to pass the full task instead of its individual properties (`task=project.task`)

application/backend/app/services/datumaro_converter.py
```diff
@@ -381,7 +410,8 @@ def _convert_sample(
 
 
 def convert_dataset(
-    project: ProjectBase,
+    task_type: TaskType,
+    exclusive_labels: bool,
```
leoll2: Same as the other comment, passing the task is more compact and maintainable

application/backend/app/services/datumaro_converter.py
```diff
@@ -40,6 +39,7 @@ class ClassificationSample(Sample):
     image_info: ImageInfo = image_info_field()
     label: int = label_field(dtype=pl.Int32(), is_list=False)
     confidence: float | None = score_field(dtype=pl.Float32())
+    subset: Subset = subset_field()
```
leoll2, line 43:
```suggestion
    subset | None: Subset = subset_field()
```
We can use the `None` value to mean "unassigned". Note that these Sample classes will be later reused for dataset import/export too, where it makes sense to have samples without a subset.

application/backend/app/services/datumaro_converter.py
```diff
@@ -98,6 +100,29 @@ class InstanceSegmentationSample(Sample):
...
+    """
+    match subset:
+        case DatasetItemSubset.TRAINING:
+            return Subset.TRAINING
+        case DatasetItemSubset.VALIDATION:
+            return Subset.VALIDATION
+        case DatasetItemSubset.TESTING:
+            return Subset.TESTING
+        case _:
+            raise ValueError(f"Unknown subset type: {subset}")
```
leoll2, line 130:
```suggestion
    match subset:
        case DatasetItemSubset.TRAINING:
            return Subset.TRAINING
        case DatasetItemSubset.VALIDATION:
            return Subset.VALIDATION
        case DatasetItemSubset.TESTING:
            return Subset.TESTING
        case DatasetItemSubset.UNASSIGNED:
            return None
        case _:
            raise ValueError(f"Unknown subset type: {subset}")
```

application/backend/app/services/datumaro_converter.py
```diff
@@ -173,7 +198,7 @@ def convert_classification_dataset(
     def _convert_sample(
         dataset_item: DatasetItem, image_path: str, project_labels_ids: list[UUID]
     ) -> ClassificationSample | None:
-        if dataset_item.annotation_data is None:
+        if dataset_item.annotation_data is None or dataset_item.subset is None:
             return None
```
leoll2, line 207: Skipping an item if it doesn't have a subset is unexpected behavior for this function; its responsibility should be limited to conversion, not filtering. For the training usecase, unassigned items should be filtered out elsewhere, by calling `DatasetService.get_dm_dataset(task=project.task, annotation_status=DatasetItemAnnotationStatus.REVIEWED)`.

application/backend/tests/unit/services/evaluation/conftest.py
```diff
@@ -21,11 +21,21 @@ def fxt_multiclass_classification_dataset_gt() -> Dataset:
     dataset = Dataset(ClassificationSample, categories={"label": LabelCategories(labels=("cat", "dog", "bird"))})
     img_info = ImageInfo(width=100, height=100)
     samples = (
-        ClassificationSample(image="/dummy/path/A.jpg", image_info=img_info, label=0, confidence=None),
-        ClassificationSample(image="/dummy/path/B.jpg", image_info=img_info, label=1, confidence=None),
-        ClassificationSample(image="/dummy/path/C.jpg", image_info=img_info, label=2, confidence=None),
-        ClassificationSample(image="/dummy/path/D.jpg", image_info=img_info, label=1, confidence=None),
-        ClassificationSample(image="/dummy/path/E.jpg", image_info=img_info, label=2, confidence=None),
+        ClassificationSample(
+            image="/dummy/path/A.jpg", image_info=img_info, label=0, confidence=None, subset=Subset.TRAINING
```
leoll2: Nitpicky: these are evaluation unit tests, which is done on the validation/testing subset but never on the training one. I suggest to update to `Subset.VALIDATION` (or `TESTING`) for logical consistency.

application/backend/app/services/datumaro_converter.py
itallix: @AlbertvanHouten

When defining the field as `subset: Subset | None = subset_field()`, an error occurs during sample retrieval with `dataset[0]`. Should the `Subset` be extended with an `UNASSIGNED` type on the Datumaro side, or are there other proposals?

```
target_type = datumaro.experimental.fields.Subset | None

    def from_polars(self, name: str, row_index: int, df: pl.DataFrame, target_type: type[T]) -> T:
        """Reconstruct subset value from Polars data.
    
        If target_type is an Enum, converts string back to enum value.
        Otherwise, returns the string value directly.
        """
        value = df[name][row_index]
    
        if value is None:
            return None  # type: ignore
    
>       if issubclass(target_type, Enum):
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       TypeError: issubclass() arg 1 must be a class

application/backend/.venv/lib/python3.13/site-packages/datumaro/experimental/fields.py:1174: TypeError
```

application/backend/app/api/routers/jobs.py
itallix: Indeed, makes sense, replaced with `task` attribute.

application/backend/app/services/datumaro_converter.py
itallix: Upgraded `datumaro` and using `UNASSIGNED` in conversions.

application/backend/app/services/dataset_service.py
```diff
@@ -60,13 +63,28 @@ def __init__(self, message: str | None = None):
...
+        label_service: LabelService,
+        db_session: Session | None = None,
+    ) -> None:
+        super().__init__(db_session)
 
         self.projects_dir = data_dir / "projects"
-        self._db_session = db_session
-        self._label_service = LabelService(db_session=db_session)
+        self._label_service = label_service
+        self._session_managed_services.append(label_service)
```
leoll2: This step looks a bit hacky, the attribute `self._session_managed_services` is hidden in `BaseSessionManagedService` and not documented. To make it slightly more dev-friendly, I suggest to add a public method called `BaseSessionManagedService.register_child_service` or similar, to wrap this operation.

application/backend/app/services/dataset_service.py
```diff
@@ -309,39 +324,44 @@ def set_dataset_item_annotations(
...
+        return self.get_dataset_item_by_id(project_id=project_id, dataset_item_id=dataset_item_id)
 
-    def get_dm_dataset(self, project: ProjectView) -> dm.Dataset:
+    def get_dm_dataset(self, project_id: UUID, task: TaskBase) -> dm.Dataset:
         def _get_dataset_items(offset: int, limit: int) -> list[DatasetItem]:
-            return self.list_dataset_items(project=project, limit=limit, offset=offset)
+            return self.list_dataset_items(
+                project_id=project_id,
+                filters=DatasetItemFilters(
+                    limit=limit, offset=offset, annotation_status=DatasetItemAnnotationStatus.REVIEWED
```
leoll2: Please expose `annotation_status` as an optional parameter of `get_dm_dataset` (default `None`) and pass the value `DatasetItemAnnotationStatus.REVIEWED` explicitly when you call it in `OTXTrainer`.

leoll2 (approved):

PR #5013, opened 2025-11-14
Simplify OTXModel API

### LINKED ISSUE #4449: Simplify OTXModel API
### 🚩 Problem

Currently, initializing an `OTXModel` (e.g., `MobileNetV3MulticlassCls`) requires manually passing verbose configuration objects like `label_info` and `data_input_params`.

#### Example

```python
model = MobileNetV3MulticlassCls(
    label_info={
        "label_names": ["daisy", "dandelion", "roses", "sunflowers", "tulips"],
        "label_id": [0, 1, 2, 3, 4],
        "label_groups": [["daisy", "dandelion", "roses", "sunflowers", "tulips"]],
    },
    data_input_params={
        "input_size": [224, 224],
        "mean": [123.675, 116.28, 103.53],
        "std": [58.395, 57.12, 57.375],
    },
)
```

The `label_info` argument is particularly inconvenient when the label set is unknown in advance or contains a large number of classes (e.g., 100+). In such cases, it becomes unnecessarily difficult to obtain a standalone model for simple use cases.

## Proposal

Introduce a simplified API:

```python
model = MobileNetV3(model_name="mobilenet_v3_large")
```

- `label_info` can be inferred directly from the dataset once it's available, or initialized with a placeholder if the labels are not yet known.
- To eliminate the need for manually specifying `data_input_params`, propose moving the `Exporter` logic to the `Engine` level. This way, the necessary metadata can be derived directly from the `datamodule`. The model itself would only require `input_size`, which can default to a sensible value based on the chosen architecture.

kprokofi:
### Summary

resolves https://github.com/open-edge-platform/training_extensions/issues/4449

Now, we can create any model using only number of classes and infer on arbitrary tensor:

```
from otx.models import YOLOX
import torch

model = YOLOX(5) # 5 classes
image = torch.rand(1,3,640,640)
outputs = model(image)
```

We can init Engine with the model name as well:

```
from otx.backend.native.engine import OTXEngine

engine = OTXEngine(model="efficientnet_b0", data="path/to/dataset", task="MULTI_CLASS_CLS")
engine.train()

```

- [x] Make DataInputParams - optional
- [x] Set default values for each model (input size, mean and std)
- [x] Make possible to forward a tensor (no OTXDataItem required)
- [x] Return possibility to instantiate Engine with model/task only
- [x] Remove V1 support

### Checklist


library/tests/unit/backend/native/models/detection/test_base.py
```diff
@@ -45,7 +45,7 @@ def mock_scheduler(self):
                 "state_dict": {},
             },
         ],
-        ids=["v1", "v2"],
+        ids=["v2"],
```
leoll2: I think `ids` and `params` should have the same length, but now `ids` has 1 element while `params` has 2.

library/src/otx/backend/native/models/segmentation/dino_v2_seg.py
```diff
@@ -52,7 +52,7 @@ class DinoV2Seg(OTXSegmentationModel):
     def __init__(
         self,
         label_info: LabelInfoTypes,
-        data_input_params: DataInputParams,
+        data_input_params: DataInputParams | None = None,
```
leoll2, line 55: Does this and other segmentation models need no `_default_preprocessing_params`?

library/src/otx/backend/native/models/base.py
```diff
@@ -114,11 +129,14 @@ class OTXModel(LightningModule):
 
     _OPTIMIZED_MODEL_BASE_NAME: str = "optimized_model"
     input_size_multiplier: int = 1
+    _default_preprocessing_params: ClassVar[dict[str, DataInputParams] | DataInputParams] = DataInputParams(
+        input_size=(240, 240), mean=(0.0, 0.0, 0.0), std=(1.0, 1.0, 1.0)
+    )
```
leoll2: What do these values mean for the base class? These parameters are architecture specific, I don't think there's a reasonable global default value; all model implementations shall explicitly provide their defaults.

I would slightly refactor the code to make it an abstract variable:
```python

class OTXModel(LightningModule):
    @property
    def _default_preprocessing_params(self) -> DataInputParams:
        """
        Parameters for image preprocessing.

        Each model architecture must implement this property, returning a DataInputParams
        containing the image input size, mean, and std, that is used to preprocess the input image.
        """

class OTXHlabelClsModel(OTXModel):
    _default_preprocessing_params = DataInputParams(
        input_size=(224, 224), mean=(123.675, 116.28, 103.53), std=(58.395, 57.12, 57.375)
    )
```

This way, `_default_preprocessing_params` becomes required for every architecture - an error will be raised if the developer forgets to specify it.

_Note: this approach leverages a Python [workaround](https://discuss.python.org/t/provide-a-canonical-way-to-declare-an-abstract-class-variable/69416) to declare abstract class variables._

library/src/otx/backend/native/models/classification/hlabel_models/base.py
```diff
@@ -48,11 +49,14 @@ class OTXHlabelClsModel(OTXModel):
     """
 
     label_info: HLabelInfo
+    _default_preprocessing_params: ClassVar[dict[str, DataInputParams] | DataInputParams] = DataInputParams(
+        input_size=(224, 224), mean=(123.675, 116.28, 103.53), std=(58.395, 57.12, 57.375)
```
leoll2: Where did you obtain these values? Before this PR, were they hardcoded somewhere else?

kprokofi: Yes, in the config

library/src/otx/backend/native/models/segmentation/dino_v2_seg.py
kprokofi: They are common and defined in the base segmentation class.

library/src/otx/backend/native/models/base.py
kprokofi: I made this parameter via property and set abstract property in the base OTXModel class

library/tests/unit/backend/native/models/detection/test_base.py
kprokofi: Done

library/src/otx/backend/native/cli/utils.py
```diff
@@ -34,7 +34,9 @@ def get_otx_root_path() -> Path:
 RECIPE_PATH = get_otx_root_path() / "recipe"
 
 
-def list_models(task: OTXTaskType | None = None, pattern: str | None = None, print_table: bool = False) -> list[str]:
+def list_models(
+    task: OTXTaskType | None = None, pattern: str | None = None, print_table: bool = False, return_recipes: bool = False
```
leoll2, line 38: Missing `return_recipes` in the docstring

library/src/otx/tools/auto_configurator.py
```diff
@@ -89,20 +89,40 @@ def __init__(
         self,
         data_root: PathLike | None = None,
         task: OTXTaskType | None = None,
-        model_config_path: PathLike | None = None,
+        model: PathLike | str | None = None,
```
leoll2, line 86: The name of the parameter (`model`) doesn't match the docstring (L71, `model_name`)

library/src/otx/backend/native/models/keypoint_detection/base.py
```diff
@@ -42,10 +43,14 @@ class OTXKeypointDetectionModel(OTXModel):
 
     """
 
+    _default_preprocessing_params: ClassVar[dict[str, DataInputParams] | DataInputParams] = {
+        "rtmpose_tiny": DataInputParams(input_size=(640, 640), mean=(0.0, 0.0, 0.0), std=(255.0, 255.0, 255.0)),
```
leoll2: Why do some models have a `dict[str, DataInputParams]` while other provide the `DataInputParams` directly?

library/src/otx/backend/native/models/detection/yolox.py
```diff
@@ -194,6 +197,14 @@ def export(
...
-    def load_from_otx_v1_ckpt(self, state_dict: dict, add_prefix: str = "model.") -> dict:
-        """Load the previous OTX ckpt according to OTX2.0."""
-        return OTXv1Helper.load_det_ckpt(state_dict, add_prefix)
+    @property
+    def _default_preprocessing_params(self) -> DataInputParams | dict[str, DataInputParams]:
+        return {
+            "yolox_tiny": DataInputParams(
+                input_size=(640, 640), mean=(123.675, 116.28, 103.53), std=(58.395, 57.12, 57.375)
+            ),
+            # TODO(@kprokofi): this looks like a bug. The image should be normalized before training.
```
leoll2, line 205: Please link a ticket to every TODO

library/src/otx/backend/native/models/instance_segmentation/maskrcnn.py
```diff
@@ -391,6 +386,22 @@ def _optimization_config(self) -> dict[str, Any]:
 
         return {}
 
+    @property
+    def _default_preprocessing_params(self) -> DataInputParams | dict[str, DataInputParams]:
+        return {
+            "maskrcnn_resnet_50": DataInputParams(
+                input_size=(1024, 1024), mean=(123.675, 116.28, 103.53), std=(58.395, 57.12, 57.375)
+            ),
+            # TODO(@kprokofi): The std values of (1.0, 1.0, 1.0) for maskrcnn_efficientnet_b2b
```
leoll2, line 395: Please link a ticket to every TODO

library/src/otx/backend/native/models/detection/base.py
```diff
@@ -41,25 +41,35 @@
...
         optimizer (OptimizerCallable, optional): Optimizer callable. Defaults to DefaultOptimizerCallable.
         scheduler (LRSchedulerCallable | LRSchedulerListCallable, optional): Scheduler callable.
         Defaults to DefaultSchedulerCallable.
         metric (MetricCallable, optional): Metric callable. Defaults to MeanAveragePrecisionFMeasureCallable.
         torch_compile (bool, optional): Whether to use torch compile. Defaults to False.
         tile_config (TileConfig, optional): Configuration for tiling. Defaults to TileConfig(enable_tiler=False).
-        explain_mode (bool, optional): Whether to enable explain mode. Defaults to False.
+        explain_mode (bool, optional): Whether to enable explain mode.
+            The model will return feature vectors need for XAI visualization.
+            Defaults to False.
```
leoll2, line 65: I don't see any `explain_mode` parameter in the constructor

library/src/otx/backend/native/models/base.py
```diff
@@ -594,9 +615,13 @@ def _customize_outputs(
...
     def forward(
         self,
-        inputs: OTXDataBatch,
-    ) -> OTXPredBatch | OTXBatchLossEntity:
+        inputs: OTXDataBatch | Tensor,
+    ) -> OTXPredBatch | OTXBatchLossEntity | Tensor:
         """Model forward function."""
+        # Simple forward
+        if isinstance(inputs, Tensor):
+            return self.forward_for_tracing(inputs)
```
leoll2, line 623: Can you help me understand this change?

leoll2 (commented):
A few minor comments, LGTM

library/src/otx/backend/native/models/base.py
kprokofi: Previously, we required to construct OTXDataItem first to infer a model. It complicates things. In OTX we already have functionality that is used for export to infer just a tensor. I added a simple line of code to propagate this functionality in forward method. It helps users and developers to test/infer model quickly. 

So, this allows us:
```
x = torch.tensor(...) # or load any image here
out = model(x) # no need of additional entities
```

Previously:
```
from otx.data.entity import OTXDataItem
input = OTXDataItem(images=x, bbox=None, ...)
model(input)
```

library/src/otx/backend/native/models/keypoint_detection/base.py
kprokofi: For certain models, such as YOLOX and MaskRCNN, it is needed to utilize distinct values for each model version. Consequently, I opt for a dictionary to achieve this. However, it is impractical to enumerate all models for every scenario. For instance, in classification tasks, we employ factory backbones with numerous options. These backbones consistently share ImageNet values. Therefore, the direct utilization of DataInputParams is more suitable in such cases.

kprokofi: I deleted dict when it is unnecessary

library/src/otx/backend/native/models/detection/yolox.py
kprokofi: Done

library/src/otx/backend/native/models/instance_segmentation/maskrcnn.py
kprokofi: Done

library/src/otx/backend/native/models/detection/base.py
kprokofi: Fixed

library/src/otx/tools/auto_configurator.py
kprokofi: Fixed

library/src/otx/backend/native/cli/utils.py
kprokofi: Fixed

leoll2 (approved):

PR #5021, opened 2025-11-14
Upgrade Augmentations Pipeline in OTX + 16bit images support

### LINKED ISSUE #5020: Optimize OTX Data Augmentations pipeline
### 📄 Description

**Current Problem:**

1. **Mixed augmentation pipelines**: OTX currently uses a combination of `torchvision` and OpenMMLab augmentations (some self-implemented in the repo). Later in the pipeline, operations rely on NumPy images, while `torchvision` expects PyTorch tensors. This causes constant conversion between tensors and NumPy arrays, introducing a **performance bottleneck** that slows down training.  

2. **Inconsistent interfaces**: Parameter names, formats, and APIs differ across the various augmentation implementations, making the codebase harder to maintain and extend.  

3. **Redundant self-implemented augmentations**: Several augmentations are re-implemented locally in OTX, even though robust third-party solutions exist and are actively maintained.  

---

**Proposed Solution:**

Evaluate **Kornia** as the primary augmentation library for OTX. Kornia is a differentiable computer vision library built on top of PyTorch, with an extensive set of augmentation operators.  

Key benefits of Kornia:  
- 🟢 **PyTorch-first design**: All augmentations operate directly on PyTorch tensors, removing unnecessary conversions from/to NumPy and improving training efficiency.  
- 🟢 **Rich functionality**: Provides a wide range of augmentations, including geometric, color, and intensity transformations, as well as advanced ones (e.g., motion blur, random perspective, cutout).  
- 🟢 **Differentiable & GPU-accelerated**: Augmentations are differentiable, making them suitable for integration with modern deep learning pipelines, and can be executed efficiently on GPUs.  
- 🟢 **Unified API & consistency**: Standardized function signatures simplify maintainability and reduce code complexity.  
- 🟢 **Extended support**: Handles not only images but also masks, bounding boxes, and keypoints, aligning with the requirements of detection and segmentation tasks.  

---

**Benchmarking Plan (Kornia vs torchvision.v2):**  
- Compare **throughput (images/sec)** during training with augmentation-heavy pipelines.  
- Measure **GPU utilization and memory footprint** when using Kornia vs torchvision v2.  
- Evaluate **coverage of transformations** (ensure Kornia includes all augmentations currently used in OTX).  
- Validate **annotation consistency** (masks, bounding boxes, keypoints remain synchronized after augmentation).  

---

### 🎯 Objective

- Unify OTX augmentation pipeline under a **PyTorch-first** approach.  
- Improve **training performance** by reducing tensor/NumPy conversions.  
- Increase **maintainability and clarity** of augmentation code by eliminating redundant implementations.  
- Provide **benchmark results** to decide between Kornia and torchvision v2.

### LINKED ISSUE #5641: Support 16-bit images in OTX
Ensure that OTX handles 16-bit images for:
- Data loading
- Augmentation
- Training

OTX should support this format natively, without forcing a conversion to a different bit depth.

kprokofi:
### Summary

resolves https://github.com/open-edge-platform/training_extensions/issues/5020

- [x] Provide batch augmentations support per model per task
- [x] Change augmentation pipeline to hybrid mode (batch augmentation + torchvision.v2). Operate only tensors
- [x] Remove all custom unnecessary augmentations
- [x] Provide DataAugmentationFactory to support configurable augmentation both for dataset based and batch based. This solution should work with Geti templates.
- [x] Provide 16bit images support for training via intensity transforms
- [ ] Provide benchmark results

### Checklist


kprokofi:
First validation results:

Batch based augmentations + tensor only operations can improve iteration time almost 2x. The larger batch the larger improvements


| Method   | Task           | Model           | iter_time |
|----------|----------------|-----------------|-----------|
| Current  | classification | efficientnet_b0 | 0.446     |
| Proposed | classification | efficientnet_b0 | 0.225     |
| Current  | detection      | YOLOX_X         | 0.369     |
| Proposed | detection      | YOLOX_X         | 0.180     |

library/src/otx/backend/native/callbacks/gpu_augmentation.py
```diff
@@ -0,0 +1,259 @@
...
+        """Apply GPU augmentation pipeline to batch (in-place).
+
+        Automatically determines which data to transform based on batch content.
+        Kornia decides whether to modify labels based on the augmentations used.
+
+        Args:
+            pipeline: GPUAugmentationPipeline to apply.
+            batch: OTXSampleBatch to transform.
+        """
+        from torchvision import tv_tensors
```
kprokofi: Move at the top

library/src/otx/backend/native/callbacks/gpu_augmentation.py
```diff
@@ -0,0 +1,259 @@
...
+        OTXTaskType.KEYPOINT_DETECTION: ["keypoints", "label"],
+        OTXTaskType.SEMANTIC_SEGMENTATION: ["mask"],
+    }
+
+    def __init__(
+        self,
+        train_config: SubsetConfig | None = None,
+        val_config: SubsetConfig | None = None,
+        test_config: SubsetConfig | None = None,
+        apply_on_val: bool = True,
```
kprokofi: remove these, use pipelines based on _val_pipeline / _test_pipeline availability

library/src/otx/backend/native/models/classification/hlabel_models/base.py
```diff
@@ -69,7 +68,6 @@ def __init__(
         super().__init__(
             label_info=label_info,
             data_input_params=data_input_params,
-            task=OTXTaskType.H_LABEL_CLS,
```
kprokofi, line 72: review why you deleted this

library/src/otx/backend/native/models/classification/multiclass_models/base.py
```diff
@@ -119,6 +120,15 @@ def _customize_outputs(
             scores=list(scores),
         )
 
+    @property
+    def _default_train_transforms(self):  # noqa: ANN202
```
kprokofi: Remove

library/src/otx/backend/native/models/classification/multiclass_models/base.py
```diff
@@ -182,6 +192,11 @@ def forward_explain(self, inputs: OTXSampleBatch) -> OTXPredictionBatch:
             feature_vector=[feature_vector.unsqueeze(0) for feature_vector in outputs["feature_vector"]],
         )
 
+    @property
+    def task(self) -> OTXTaskType:
```
kprokofi, line 188: Review, why do you set it ti separate property

library/src/otx/backend/native/models/common/rfdetr_mixin.py
```diff
@@ -206,7 +206,7 @@ def _customize_outputs(  # pyrefly: ignore[bad-override]
                     raise TypeError(msg)
             return losses
 
-        image_shapes = [img_info.img_shape for img_info in inputs.imgs_info]  # type: ignore[union-attr]
+        image_shapes = [img_info.ori_shape for img_info in inputs.imgs_info]  # type: ignore[union-attr]
```
kprokofi, line 209: We standardize validation for IS and OD. Everything should be validated on the original image shape.

library/src/otx/backend/native/models/detection/base.py
```diff
@@ -549,6 +534,24 @@ def get_num_anchors(self) -> list[int]:
 
         return [1] * 10
 
+    @staticmethod
+    @torch.no_grad()
+    def _apply_batch_augmentations(
```
kprokofi: Remove old implementation

library/src/otx/backend/native/models/detection/yolox.py
```diff
@@ -198,11 +197,32 @@ def export(
...
+                            "Use yolox_tiny or a model with normalization for 16-bit images."
+                        )
+                        raise RuntimeError(msg)
+
+
+            inputs = super()._customize_inputs(entity)
+            # The CPU pipeline always scales images to [0, 1] float.
+            # YOLOX-S/L/X pretrained weights expect [0, 255] float, so rescale here.
+            # We create a new entity so the original (with [0, 1] images) remains intact
+            inputs["entity"] = dataclasses.replace(inputs["entity"], images=entity.images.mul(255.0))
```
kprokofi: YOLOX TINY is affected here as well. Add additional check

library/src/otx/backend/native/models/instance_segmentation/maskrcnn.py
```diff
@@ -391,16 +391,16 @@ def _optimization_config(self) -> dict[str, Any]:
...
             "maskrcnn_resnet_50": DataInputParams(
-                input_size=(1024, 1024), mean=(123.675, 116.28, 103.53), std=(58.395, 57.12, 57.375)
+                input_size=(1024, 1024), mean=(0.485, 0.456, 0.406), std=(0.229, 0.224, 0.225)
             ),
             # TODO(@kprokofi): The std values of (1.0, 1.0, 1.0) for maskrcnn_efficientnet_b2b
             # differ from other variants which use (58.395, 57.12, 57.375), which may indicate missing normalization.
             # issue: https://github.com/open-edge-platform/training_extensions/issues/5023
             "maskrcnn_efficientnet_b2b": DataInputParams(
-                input_size=(1024, 1024), mean=(123.675, 116.28, 103.53), std=(1.0, 1.0, 1.0)
+                input_size=(1024, 1024), mean=(0.485, 0.456, 0.406), std=(1.0, 1.0, 1.0)
```
kprokofi, line 400: Efficient Net should be checked. Seems like again no normalization here

library/src/otx/backend/native/models/keypoint_detection/base.py
```diff
@@ -207,6 +209,41 @@ def _export_parameters(self) -> TaskLevelExportParameters:
             confidence_threshold=self.hparams.get("best_confidence_threshold", None),
         )
 
+    @staticmethod
+    @torch.no_grad()
+    def _apply_batch_augmentations(
```
kprokofi: Remove old implementation

library/src/otx/backend/native/models/keypoint_detection/base.py
```diff
@@ -207,6 +209,41 @@ def _export_parameters(self) -> TaskLevelExportParameters:
...
+            # Compute visible mask. Keypoints should be visible if they are inside the image (>=0, x<=w, y<=h)
+            visible_mask = (
+                (augmented_kps > 0).all(axis=2) * (augmented_kps[:, :, 0] <= w) * (augmented_kps[:, :, 1] <= h)
+            )
+            stacked_kps[:, :, 2] = stacked_kps[:, :, 2] * visible_mask
+            # Update visible keypoints with augmented values
+            batch.keypoints = list(stacked_kps)
+
+    @property
+    def _default_train_transforms(self):  # noqa: ANN202
```
kprokofi: Remove

library/src/otx/backend/native/models/keypoint_detection/base.py
```diff
@@ -207,6 +209,41 @@ def _export_parameters(self) -> TaskLevelExportParameters:
...
+        return AugmentationSequential(
+            kornia.augmentation.Normalize(self.data_input_params.mean, self.data_input_params.std),
+            data_keys=["input", "keypoints"],
+            keepdim=True,
+        )
+
     @property
     def _default_preprocessing_params(self) -> DataInputParams | dict[str, DataInputParams]:
-        return DataInputParams(input_size=(640, 640), mean=(0.0, 0.0, 0.0), std=(255.0, 255.0, 255.0))
+        return DataInputParams(input_size=(512, 512), mean=(0.485, 0.456, 0.406), std=(0.229, 0.224, 0.225))
```
kprokofi, line 217: Check normalization parameters. They are changed

library/src/otx/backend/native/models/base.py
```diff
@@ -65,17 +65,21 @@
 class DataInputParams:
     """Parameters of the input data such as input size, mean, and std."""
 
-    input_size: tuple[int, int]
-    mean: tuple[float, float, float]
-    std: tuple[float, float, float]
+    input_size: tuple[int, int] | None = None
```
kprokofi: None? Check this

library/src/otx/backend/native/engine.py
```diff
@@ -986,11 +987,43 @@ def configure_callbacks(self) -> None:
...
+        )
+        has_val_gpu_augs = val_config and hasattr(val_config, "augmentations_gpu") and val_config.augmentations_gpu
+
+        if not has_train_gpu_augs and not has_val_gpu_augs:
+            return None
+
+        logging.info("Building GPU augmentation callback with Kornia augmentations")
+        return GPUAugmentationCallback(
+            train_config=train_config if has_train_gpu_augs else None,
+            val_config=val_config if has_val_gpu_augs else None,
```
kprokofi, line 1028: where is the test?

library/src/otx/cli/utils/jsonargparse.py
```diff
@@ -230,18 +230,40 @@ def list_override(configs: Namespace, key: str, overrides: list, convert_dict_to
...
-        item = next((item for item in configs[key] if item["class_path"] == class_path), None)
+        item = next((item for item in base_list if item["class_path"] == class_path), None)
         if item is not None:
             Namespace(item).update(target)
         else:
             converted_target = dict_to_namespace(target) if convert_dict_to_namespace else target
-            configs[key].append(converted_target)
+            base_list.append(converted_target)
+
+    item_by_cp: dict[str, Any] = {}
```
kprokofi: please, check the order. It should be as: base_list first as a core, override if needed, everything else from source config

library/src/otx/config/data.py
```diff
@@ -38,39 +86,45 @@ class SubsetConfig:
...
+            Examples: Resize, RandomResizedCrop, intensity mapping transforms.
+        augmentations_gpu (list[dict[str, Any]]): GPU-stage augmentations using Kornia.
+            These run after batch transfer to GPU via Lightning Callback.
+            Examples: RandomHorizontalFlip, ColorJiggle, Normalize.
+        intensity (IntensityConfig): High-bit-depth intensity mapping configuration.
+        transform_lib_type (TransformLibType): [DEPRECATED] Specifies the transform library type used.
         num_workers (int): Number of worker processes for the dataloader.
         sampler (SamplerConfig): Sampler configuration for the dataloader.
-        to_tv_image (bool): Whether to convert images to torch tensors.
+        to_tv_image (bool): [DEPRECATED] Whether to convert images to torch tensors.
```
kprokofi: Remove deprecated things

library/src/otx/config/data.py
```diff
@@ -38,39 +86,45 @@ class SubsetConfig:
...
     batch_size: int = 6
     subset_name: str = "train"
-    transforms: list[dict[str, Any]] = field(default_factory=list)
-    transform_lib_type: TransformLibType = TransformLibType.TORCHVISION
+    augmentations_cpu: list[dict[str, Any]] = field(default_factory=list)
+    augmentations_gpu: list[dict[str, Any]] = field(default_factory=list)
+    intensity: IntensityConfig = field(default_factory=IntensityConfig)
     num_workers: int = 2
     sampler: SamplerConfig = field(default_factory=SamplerConfig)
     to_tv_image: bool = True
```
kprokofi: Remove

library/src/otx/data/augmentation/pipeline.py
```diff
@@ -0,0 +1,812 @@
...
+            CPUAugmentationPipeline ready for use in Dataset.
+        """
+        input_size = getattr(config, "input_size", None)
+        aug_configs = config.augmentations_cpu
+        intensity_config = getattr(config, "intensity", None)
+
+        augmentations: list[nn.Module] = []
+
+        # --- 1. Prepend intensity mapping transform ---------------------------
+        if intensity_config is not None:
```
kprokofi, line 312: It always should be applied to avoid different ranges for images

library/src/otx/data/augmentation/pipeline.py
```diff
@@ -0,0 +1,812 @@
...
+            intensity_transform = build_intensity_transform(intensity_config)
+            augmentations.append(_IntensityAdapter(intensity_transform, intensity_config.storage_dtype))
+
+        # --- 2. User-configured augmentations ---------------------------------
+        if aug_configs:
+            for aug_config in aug_configs:
+                cfg = copy(aug_config)
+                if isinstance(cfg, (dict, DictConfig)):
+                    # Skip disabled transforms
```
kprokofi: Remove this. I removed this flag.

library/src/otx/data/augmentation/pipeline.py
```diff
@@ -0,0 +1,812 @@
...
+        else:
+            result = transform(transformable)
+            if isinstance(result, dict):
+                for key, value in result.items():
+                    if key == "boxes":
+                        inputs.bboxes = value
+                    elif key == "labels":
+                        inputs.label = value
+                    else:
+                        setattr(inputs, key, value)
```
kprokofi: We can do that for boxes and labels as well

library/src/otx/data/augmentation/pipeline.py
```diff
@@ -0,0 +1,812 @@
...
+            else:
+                # Single result, assume it's the image
+                inputs.image = result
```
kprokofi, line 414: It should be checked above

library/src/otx/data/augmentation/pipeline.py
```diff
@@ -0,0 +1,812 @@
...
+        if not aug_configs:
+            return cls([], data_keys=data_keys)
+
+        augmentations = []
+
+        for aug_config in aug_configs:
+            cfg = copy(aug_config)
+            if isinstance(cfg, (dict, DictConfig)):
+                # Skip disabled transforms
+                if not cfg.get("enable", True):
```
kprokofi: Remove this

library/src/otx/data/augmentation/pipeline.py
```diff
@@ -0,0 +1,812 @@
...
+                    continue
+
+                # Handle input_size placeholder
+                cfg = CPUAugmentationPipeline._configure_input_size(dict(cfg), input_size)  # noqa: SLF001
```
kprokofi: Consider moving this functionality to common function instead of method

library/src/otx/data/augmentation/pipeline.py
```diff
@@ -0,0 +1,812 @@
...
+            return {"images": images, "labels": labels, "bboxes": bboxes, "masks": masks, "keypoints": keypoints}
+
+        # Handle instance segmentation masks: Kornia expects (N, C, H, W) but instance
+        # masks are (N_instances, H, W). We add a channel dim before and squeeze after.
+        # This allows Kornia to properly transform instance masks along with images.
+        if masks is not None and "mask" in self._data_keys:
+            # Instance seg masks: (N_instances, H, W) - 3D per sample
+            # Semantic seg masks: (C, H, W) where C is often 1 or num_classes
+            # We add channel dim to all masks for consistency with Kornia
+            masks = [m.unsqueeze(0) for m in masks]  # (N, H, W) -> (N, 1, H, W)
```
kprokofi, line 643: check semantic segmentation once again here

library/src/otx/data/augmentation/pipeline.py
```diff
@@ -0,0 +1,812 @@
...
+                mask_results = [m.squeeze(0) for m in mask_results]  # (1, N, H, W) -> (N, H, W)
+                output["masks"] = mask_results
+            elif key == "keypoints":
+                output["keypoints"] = results[i]
+
+        # Sanitize geometric annotations after Kornia transforms.
+        if output["images"] is not None:
+            s_bboxes, s_labels, s_masks, s_keypoints = self._sanitize_annotations(
+                typing.cast("torch.Tensor", output["images"]),
+                typing.cast("list[torch.Tensor] | None", output["bboxes"]),
```
kprokofi: check why it is needed

library/src/otx/data/augmentation/transforms.py
```diff
@@ -0,0 +1,951 @@
...
+
+        return valid
+
+    def _create_mosaic_canvas(self, dtype: torch.dtype, device: torch.device) -> torch.Tensor:
+        """Create empty mosaic canvas filled with pad_val."""
+        mosaic_h = self.img_scale[0] * 2
+        mosaic_w = self.img_scale[1] * 2
+
+        fill_val = self.pad_val if isinstance(self.pad_val, (int, float)) else self.pad_val[0]
+        if fill_val > 1.0:
```
kprokofi: It is dangerous. the value can be 1.00001

library/src/otx/data/dataset/base.py
```diff
@@ -73,19 +74,38 @@ def _default_collate_fn(items: list[OTXSample]) -> OTXSampleBatch:
...
+        # instead of silently producing wrong values.
+        if img.dtype != torch.float32:
+            if img.dtype in (torch.int32, torch.int16, torch.int64):
+                msg = (
+                    f"Image tensor has dtype {img.dtype} which looks like a high-bit-depth image "
+                    "that was not converted to float32. Please configure an intensity transform "
+                    "(IntensityConfig) in the recipe to map raw pixel values to [0, 1] float32."
+                )
+                raise TypeError(msg)
+            # uint8 → float32 (safe, values 0-255 → 0.0-255.0 — GPU pipeline normalizes)
```
kprokofi: No, no safe

kprokofi: Values should be from 0 to 1

library/src/otx/data/entity/sample.py
```diff
@@ -266,19 +355,20 @@ def collate_fn(samples: list[OTXSample]) -> OTXSampleBatch:
...
-    # Check if all images have the same size. TODO(kprokofi): remove this check once OV IR models are moved.
-    if all(sample.image.shape == samples[0].image.shape for sample in samples):
-        images = torch.stack([sample.image for sample in samples])
-    else:
-        # we need this only in case of OV inference, where no resize
-        images = [sample.image for sample in samples]
+    images = torch.stack([sample.image for sample in samples])
+
+    # Collect optional fields - use getattr to handle samples without these attributes
+    labels = [getattr(s, "label", None) for s in samples] if hasattr(samples[0], "label") else None
```
kprokofi: looks too complicated just to avoid pyrefly issues

library/src/otx/data/module.py
```diff
@@ -123,8 +121,13 @@ def __init__(
...
                     subset_cfg.input_size = input_size  # type: ignore[assignment]
 
-        # Extract mean and std from Normalize transform
-        self.input_mean, self.input_std = self.extract_normalization_params(self.train_subset.transforms)
+        # Derive mean/std from the CPU pipeline's Normalize transform.
+        # If no Normalize is present, leave as None so models fall back to their own defaults.
+        cpu_pipeline: CPUAugmentationPipeline | None = getattr(self.train_subset, "_cpu_pipeline", None)
+        if cpu_pipeline is None and getattr(self.train_subset, "augmentations_cpu", None):
+            cpu_pipeline = CPUAugmentationPipeline.from_config(self.train_subset)
+        self.input_mean: tuple[float, float, float] | None = cpu_pipeline.mean if cpu_pipeline is not None else None
```
kprokofi: check all use cases (CPU - None, GPU - not None, etc.)

library/tests/perf_v2/tasks/detection.py
```diff
@@ -19,74 +20,70 @@
 
 MODEL_TEST_CASES = [
     ModelInfo(task=TASK_TYPE.value, name="atss_mobilenetv2", category="default"),
-    ModelInfo(task=TASK_TYPE.value, name="yolox_s", category="speed"),
+    # ModelInfo(task=TASK_TYPE.value, name="yolox_s", category="speed"),
```
kprokofi: Return back

application/ui/src/features/dataset/import-export/export-jobs-list/export-job/cancel-job-confirmation/cancel-job-confirmation.component.tsx
```diff
@@ -0,0 +1,56 @@
...
+                    removeLsExportId(jobId);
+                },
+                onError: (error) => {
+                    isInvalidJob(error) && removeLsExportId(jobId);
+                },
+                onSettled: () => {
+                    dialogState.close();
+                },
+            }
+        );
```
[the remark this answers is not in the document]
kprokofi: UI changes will be reverted, they are accidentally made

library/pyproject.toml
```diff
@@ -51,6 +51,7 @@ dependencies = [
     "pyyaml==6.0.3",
     "tensorboard==2.20.0",
     "tensorboardx==2.6.4",
+    "kornia",
```
leoll2:
```suggestion
    "kornia~=0.8.2",
```

library/tests/unit/tools/test_transforms_updater.py
```diff
@@ -0,0 +1,529 @@
+# Copyright (C) 2024 Intel Corporation
```
leoll2: Stale copyright header

library/src/otx/recipe/_base_/data/classification.yaml
```diff
@@ -6,119 +6,100 @@ data_format: imagenet_with_subset_dirs
 unannotated_items_ratio: 0.0
 train_subset:
   subset_name: train
-  transform_lib_type: TORCHVISION
   batch_size: 64
-  num_workers: 2
-  to_tv_image: false
-  transforms:
-    - class_path: otx.data.transform_libs.torchvision.RandomResizedCrop
+  num_workers: 0
```
leoll2: Why `num_workers` changed from 2 to 0?

library/src/otx/recipe/_base_/data/classification.yaml
```diff
@@ -6,119 +6,100 @@ data_format: imagenet_with_subset_dirs
...
-        scaling_ratio_range:
-          - 0.5
-          - 1.5
-        max_shear_degree: 2.0
-    - class_path: otx.data.transform_libs.torchvision.RandomFlip
-      enable: true
-      init_args:
-        probability: 0.5
-    - class_path: torchvision.transforms.v2.RandomVerticalFlip
-      enable: false
```
leoll2, line 52: This transformation seems to be missing in the updated recipe. Is it no longer supported, or is it just omitted because `enable: false` by default?

library/src/otx/backend/native/callbacks/aug_scheduler.py
```diff
@@ -1,108 +1,112 @@
...
+    Manages multiple augmentation policies and switches between them based on the
+    current training epoch. Each policy contains separate CPU and GPU augmentation
+    lists that are built into ``CPUAugmentationPipeline`` and ``GPUAugmentationPipeline``
+    respectively.
 
+    Args:
+        policy_epochs (list[int]): List of 3 epoch boundaries ``[p0, p1, p2]``:
+            - ``epoch < p0``: ``no_aug``
+            - ``p0 <= epoch < p1``: ``strong_aug_1`` or ``strong_aug_2`` (random)
+            - ``epoch >= p1``: ``light_aug``
```
leoll2: What's the meaning of `p2`?

library/src/otx/backend/native/callbacks/aug_scheduler.py
```diff
@@ -1,108 +1,112 @@
...
+                "augmentations_cpu": [
+                    {"class_path": "otx.data.augmentation.transforms.Resize", ...},
+                ],
+                "augmentations_gpu": [
+                    {"class_path": "kornia.augmentation.Normalize", ...},
+                ],
+            },
+            "strong_aug_1": { ... },
+            "strong_aug_2": { ... },
+            "light_aug": { ... },
```
leoll2, line 68: The names `no_aug`, `strong_aug_1`, `strong_aug_2` and `light_aug` are hardcoded, right? We should perhaps refine the `DataAugSwitch` interface in a later PR, because it has some limitations:
- it only supports 3 policy epochs
- the strong augmentation phase only supports 2 configurations
- the light augmentation phase only supports 1 configuration

library/src/otx/backend/native/callbacks/aug_scheduler.py
```diff
@@ -1,108 +1,112 @@
...
+        # Build pipelines for each policy
+        self.policies: dict[str, dict[str, Any]] = {}
         for name, config in policies.items():
+            cpu_aug_configs = config.get("augmentations_cpu", [])
+            gpu_aug_configs = config.get("augmentations_gpu", [])
+
+            # Build CPU pipeline via SubsetConfig → CPUAugmentationPipeline
+            cpu_subset = SubsetConfig(
+                augmentations_cpu=cpu_aug_configs,
+                batch_size=1,
```
leoll2: Any reason to have `batch_size=1` here? Could we potentially support larger batch sizes?

library/src/otx/backend/native/models/base.py
```diff
@@ -65,17 +65,21 @@
 class DataInputParams:
     """Parameters of the input data such as input size, mean, and std."""
 
-    input_size: tuple[int, int]
-    mean: tuple[float, float, float]
-    std: tuple[float, float, float]
```
leoll2, line 70: Why are these now optional? `as_ncwh` throws an error if `input_size` is None, and other functions fallback to default mean/std values which doesn't seem a safe pattern to me.

library/src/otx/backend/openvino/models/base.py
```diff
@@ -421,16 +421,9 @@ def get_dummy_input(self, batch_size: int = 1) -> OTXSampleBatch:
...
-            infos.append(
-                ImageInfo(
-                    img_idx=i,
-                    img_shape=img.shape,
-                    ori_shape=img.shape,
-                ),
-            )
+        images = torch.stack([torch.rand(3, 224, 224) for _ in range(batch_size)])
+        img_shape = (224, 224)
+        infos = [ImageInfo(img_idx=i, img_shape=img_shape, ori_shape=img_shape) for i in range(batch_size)]
```
leoll2, line 426: What is this function `get_dummy_input` needed for?

library/src/otx/backend/native/models/common/rfdetr_mixin.py
```diff
@@ -206,7 +206,7 @@ def _customize_outputs(  # pyrefly: ignore[bad-override]
...
         scores_list, boxes_list, labels_list, masks_list = self.model.postprocess(  # type: ignore[attr-defined]  # pyrefly: ignore[not-callable]
```
leoll2, line 210: Not specific to this PR, just want to point out that the code quality of this `RFDETRMixin` class is below standard: it doesn't have a constructor, several attributes are not defined within the class, and it has countless `ignore` directives to suppress linters. Let's try to refactor it after release.

library/src/otx/recipe/_base_/data/classification.yaml
kprokofi: Sorry, a lot of debugging, forgot to revert

library/src/otx/recipe/_base_/data/classification.yaml
kprokofi: Yes, all augmentations with "enable: False" are now deleted. It was just a complication of things. Now, we rely on Model Manifests. If they provide additional augmentation that is not currently in recipe, it will be added based on internal name mapping. The order will be preserved as in the model manifest.

library/src/otx/backend/native/callbacks/aug_scheduler.py
kprokofi: Yes, I didn't refactor this from original implementation. We could refactor it later to handle more sophisticated schedule

library/src/otx/backend/openvino/models/base.py
kprokofi: For OTX benchmark, we can measure FPS, FLOPS with native OTXEngine. Only for that

library/src/otx/backend/native/models/common/rfdetr_mixin.py
kprokofi: It is a Mixin class to reduce code duplication for Instance Seg and OD RFDETR classes. So, missing some attributes is expected, I guess.

library/src/otx/backend/native/models/base.py
kprokofi: The logic is complex. However, I tend to make our API simple for a user. 
1) Making these parameters optional will no longer require to pass this to the model constructor. It will simply be taken from models defaults during model creation. 
2) We now derive this info from two sources: CPU augmentations and GPU. By default scenario we use GPU Normalization. This info can be obtained only in callback and rewritten in model there. So, making None at the beginning allows us to instantiate a model without this parameter. 

Yes, as_ncwh will through an error if all parameters are None. This functionality is used in export stage. On the export stage they must not be None -> if they are None, error is expected. It can happen only if a user created a model with explicit Nones and then call an export()

library/src/otx/backend/native/callbacks/aug_scheduler.py
kprokofi: Originally, p2 means no augs after that epoch. 
We don't use it though. We use light augs after p1. 

Let's remove p2 then.

library/src/otx/backend/native/callbacks/aug_scheduler.py
kprokofi: It doesn't play any role here. We need it to overwrite the augmentations in the dataset. Batch size here is just a dummy placeholder. I will remove it at all.

library/src/otx/data/dataset/base.py
```diff
@@ -120,40 +137,60 @@ class OTXDataset(TorchDataset):
...
+        sample_type: type = OTXSample,
+        storage_dtype: str = "uint8",
     ) -> None:
         self.transforms = transforms
         self.stack_images = stack_images
-        self.to_tv_image = to_tv_image
         self.sample_type = sample_type
         self.max_refetch = max_refetch
         self.data_format = data_format
+        self.storage_dtype = storage_dtype
```
AlbertvanHouten: This seems to be unused

library/src/otx/backend/native/models/common/rfdetr_mixin.py
leoll2: I see. Avoiding code duplication is generally a good idea, however there are more robust patterns to do so. This mixin class does not fit well into the class hierarchy; a base class that consumes attributes declared in the subclasses is an anti-pattern. Anyway, let's not focus too much on this class, I'll create a ticket to refactor it after release.

AlbertvanHouten:
Could you add test that verifies the 16 bit path with transformations?

library/src/otx/backend/native/models/base.py
leoll2: > Making these parameters optional will no longer require to pass this to the model constructor. It will simply be taken from models defaults during model creation.

If I'm not wrong, that's already the case. Parameter `data_input_params` is already optional for the `OTXModel` constructor, and if unspecified it's automatically determined from the model defaults.

https://github.com/open-edge-platform/training_extensions/blob/9571e2933cac6e5fdda4c74d45debc6a6e63fdf6/library/src/otx/backend/native/models/base.py#L135

If I understand correctly, `_configure_preprocessing_params` can resolve all three attributes (`input_size`, `mean` and `std`), so it outputs a `DataInputParams` with non-null attribute values. However, from a typing perspective, the attributes are still nullable because of the `| None`, so any code that consumes `model.data_input_params` must explicitly check the null condition.

Can you maybe share an example of how this change would make the API more user-friendly?

kprokofi:
@AlbertvanHouten I added new integration test with 16bit images.
I found that Datumaro doesn't support 16bit images fully. Converter is still not implemented, I temporary added it to OTX. Please, review and this should be removed after support on the datumaro side

library/src/otx/backend/native/models/base.py
kprokofi: I agree with your opinion. One additional reason to make those parameters optional is that now we may not be able to derive proper information from DataModule since normalization happens on GPU. Anyway, I agree that too many optional parameters introduced which complicates things. I made them required again, fixing a problem in different way.

library/src/otx/backend/native/callbacks/aug_scheduler.py
kprokofi: Done

AlbertvanHouten:
> @AlbertvanHouten I added new integration test with 16bit images. I found that Datumaro doesn't support 16bit images fully. Converter is still not implemented, I temporary added it to OTX. Please, review and this should be removed after support on the datumaro side

Ah I see, the samples are not updated to represent image_fields in uint16. This will double storage size in the polars dataframe so we should probably define another set of samples specifically for 16 bit images.

library/src/otx/data/dataset/base.py
```diff
@@ -6,24 +6,25 @@
...
-    from datumaro.experimental import Dataset
-
+from otx.data.augmentation.pipeline import CPUAugmentationPipeline
 from otx.data.entity.sample import OTXSample, OTXSampleBatch
-from otx.data.transform_libs.torchvision import Compose
 from otx.types import OTXTaskType
 
-Transforms = Union[Compose, Callable, List[Callable], dict[str, Compose | Callable | List[Callable]]]
+if TYPE_CHECKING:
+    from datumaro import Dataset as DmDataset
```
AlbertvanHouten: Please remove this old datumaro import. It has been completely removed from the codebase

kprokofi: Done

library/src/otx/data/dataset/base.py
kprokofi: Removed

AlbertvanHouten (dismissed):

library/tests/test_helpers.py
```diff
@@ -0,0 +1,175 @@
...
+    LabelCategories,
+    MaskCategories,
+)
+from datumaro.components.dataset import DatasetItem
+from datumaro.components.errors import MediaTypeError
+from datumaro.components.exporter import Exporter
+from datumaro.components.media import Image
+from datumaro.util.definitions import DEFAULT_SUBSET_NAME
+from datumaro.util.image import save_image
+from datumaro.util.meta_file_util import save_meta_file
```
leoll2: This file imports from the old Datumaro API, and we recently removed all such imports. To not block this PR from merging, it can be addressed in another PR.

CC @AlbertvanHouten

leoll2 (dismissed):

library/tests/test_helpers.py
AlbertvanHouten: Ah good catch, please replace this with logic from the experimental module or add the necessary logic to OTX

library/src/otx/data/augmentation/kernels.py
```diff
@@ -0,0 +1,90 @@
...
+@F.register_kernel(functional=F.pad, tv_tensor_cls=ImageInfo)
+def _pad_image_info(
+    image_info: ImageInfo,
+    padding: int | list[int],
+    **kwargs,  # noqa: ARG001
+) -> ImageInfo:
+    """Register ImageInfo to TorchVision v2 pad kernel."""
+    left, right, top, bottom = F._geometry._parse_pad_padding(padding)  # noqa: SLF001
+    height, width = image_info.img_shape
+    image_info.padding = (left, top, right, bottom)
```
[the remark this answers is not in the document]
kprokofi: [_parse_pad_padding](vscode-file://vscode-app/private/var/folders/5r/xw3kmtjx7494yn12zwvt02w00000gn/T/AppTranslocation/AD9E3F1B-6614-4C43-9E4B-1B3BF05FE4E2/d/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) returns [(left, right, top, bottom)](vscode-file://vscode-app/private/var/folders/5r/xw3kmtjx7494yn12zwvt02w00000gn/T/AppTranslocation/AD9E3F1B-6614-4C43-9E4B-1B3BF05FE4E2/d/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html). The kernel reorders to [(left, top, right, bottom)](vscode-file://vscode-app/private/var/folders/5r/xw3kmtjx7494yn12zwvt02w00000gn/T/AppTranslocation/AD9E3F1B-6614-4C43-9E4B-1B3BF05FE4E2/d/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) for storage.  [_crop_padded_map](vscode-file://vscode-app/private/var/folders/5r/xw3kmtjx7494yn12zwvt02w00000gn/T/AppTranslocation/AD9E3F1B-6614-4C43-9E4B-1B3BF05FE4E2/d/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) in [xai_utils.py](vscode-file://vscode-app/private/var/folders/5r/xw3kmtjx7494yn12zwvt02w00000gn/T/AppTranslocation/AD9E3F1B-6614-4C43-9E4B-1B3BF05FE4E2/d/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) unpacks as [left, top, right, bottom = paddings[i]](vscode-file://vscode-app/private/var/folders/5r/xw3kmtjx7494yn12zwvt02w00000gn/T/AppTranslocation/AD9E3F1B-6614-4C43-9E4B-1B3BF05FE4E2/d/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) — fully consistent. The test assert result.padding == (1, 2, 3, 4) for input [padding=[1,2,3,4]](vscode-file://vscode-app/private/var/folders/5r/xw3kmtjx7494yn12zwvt02w00000gn/T/AppTranslocation/AD9E3F1B-6614-4C43-9E4B-1B3BF05FE4E2/d/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is correct.

library/src/otx/backend/native/models/instance_segmentation/maskrcnn.py
kprokofi: On develop, EfficientNet uses [mean=(123.675,...), std=(1.0,1.0,1.0)](vscode-file://vscode-app/private/var/folders/5r/xw3kmtjx7494yn12zwvt02w00000gn/T/AppTranslocation/AD9E3F1B-6614-4C43-9E4B-1B3BF05FE4E2/d/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) in 0-255 space. On kp/kornia it's [mean=(0.485,...), std=(1.0,1.0,1.0)](vscode-file://vscode-app/private/var/folders/5r/xw3kmtjx7494yn12zwvt02w00000gn/T/AppTranslocation/AD9E3F1B-6614-4C43-9E4B-1B3BF05FE4E2/d/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) in 0-1 space. The [std=1.0](vscode-file://vscode-app/private/var/folders/5r/xw3kmtjx7494yn12zwvt02w00000gn/T/AppTranslocation/AD9E3F1B-6614-4C43-9E4B-1B3BF05FE4E2/d/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is wrong in 0-1 space (should be 1.0/255 ≈ 0.003922 to match develop's behavior), BUT this was already a known issue on develop — there's already a TODO + [issue #5023](vscode-file://vscode-app/private/var/folders/5r/xw3kmtjx7494yn12zwvt02w00000gn/T/AppTranslocation/AD9E3F1B-6614-4C43-9E4B-1B3BF05FE4E2/d/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) tracking it. The model's pretrained weights with [std=1.0](vscode-file://vscode-app/private/var/folders/5r/xw3kmtjx7494yn12zwvt02w00000gn/T/AppTranslocation/AD9E3F1B-6614-4C43-9E4B-1B3BF05FE4E2/d/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) may actually need full ImageNet std (0.229, 0.224, 0.225) anyway — this requires investigation in the dedicated issue, not a blind fix here

library/src/otx/cli/utils/jsonargparse.py
kprokofi: Fixed — removed the incorrect reordering logic. The first loop already updates base items in-place and appends new items at the end, preserving correct order: base items first (with overrides applied), then new items from override config

library/src/otx/data/factory.py
```diff
@@ -93,3 +111,42 @@ def create(
 
             case _:
                 raise NotImplementedError(task)
+
+    @staticmethod
+    def _detect_storage_dtype(dm_subset: Dataset) -> str:
+        """Detect image storage bit depth from file header or dataset schema.
```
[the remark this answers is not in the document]
kprokofi: Added unit test

library/src/otx/data/factory.py
```diff
@@ -93,3 +111,42 @@ def create(
...
+        from otx.data.entity.utils import detect_image_dtype
+
+        # 1. Try file-based detection
+        try:
+            first_item = next(iter(dm_subset))
+            path = getattr(first_item.media, "path", None) if hasattr(first_item, "media") else None
+            if path is not None:
+                return detect_image_dtype(path)
+        except (StopIteration, Exception):  # noqa: S110
+            pass
```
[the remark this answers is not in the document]
kprokofi: Fixed

library/tests/test_helpers.py
kprokofi: Removed

library/src/otx/data/factory.py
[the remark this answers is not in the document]
kprokofi: Added unit test

library/src/otx/data/entity/__init__.py
```diff
@@ -26,12 +25,10 @@
     "OTXPredictionBatch",
     "OTXSample",
     "OTXSampleBatch",
-    "Points",
     "TileBatchDetDataEntity",
     "TileBatchInstSegDataEntity",
     "TileBatchInstSegDataEntity",
```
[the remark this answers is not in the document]
kprokofi: Fixed

library/src/otx/data/augmentation/kernels.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+@F.register_kernel(functional=F.center_crop, tv_tensor_cls=ImageInfo)
+def _center_crop_image_info(
+    image_info: ImageInfo,
+    output_size: list[int],
+    **kwargs,  # noqa: ARG001
+) -> ImageInfo:
+    """Register ImageInfo to TorchVision v2 center_crop kernel."""
+    img_shape = F._geometry._center_crop_parse_output_size(output_size)  # noqa: SLF001
```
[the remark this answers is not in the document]
kprokofi: Skipped

AlbertvanHouten (dismissed):

leoll2 (approved):

AlbertvanHouten (approved):

PR #5026, opened 2025-11-18
Upgrade Geti Tune dependencies

leoll2:
### Summary

Upgrade Geti Tune dependencies to their latest version. I have personally reviewed the release notes of the upgraded packages, I didn't find any potential issue or incompatibility. I have also locked the Python version to 3.13, because not every package (notably, `openvino`) support 3.14 yet.

### How to test

Start the application and verify that it runs as expected.

### Checklist


application/backend/app/repositories/base.py
```diff
@@ -61,8 +61,8 @@ def update(self, item: ModelType) -> ModelType:
 
     def delete(self, obj_id: str) -> bool:
         stmt = delete(self.model).where(self.model.id == obj_id)
-        result = self.db.execute(stmt)
-        return result.rowcount > 0  # type: ignore[union-attr]
+        result = cast(CursorResult, self.db.execute(stmt))
```
leoll2, line 64: This cast is necessary and recommended by sqlalchemy [[source](https://github.com/sqlalchemy/sqlalchemy/releases/tag/rel_2_0_44)]:

> [typing] [bug] Fixed typing bug where the Session.execute() method advertised that
it would return a CursorResult if given an insert/update/delete
statement. This is not the general case as several flavors of ORM
insert/update do not actually yield a CursorResult which cannot
be differentiated at the typing overload level, so the method now yields
Result in all cases. For those cases where
CursorResult is known to be returned and the .rowcount
attribute is required, please use typing.cast().

itallix (approved):

PR #5027, opened 2025-11-18
Simplify UUID validation with type annotations

### LINKED ISSUE #4876: Simplify UUID validation with consistent type annotations
**Current State:**

- Multiple validation approaches:
  - Individual functions like `get_source_id`, `get_project_id` using `is_valid_uuid()`
  - New centralized `validate_uuid_param` function using direct `UUID()` constructor (introduced in #4871)
- Code duplication: Each ID type has its own validation function with nearly identical logic

**Proposed Changes:**

1. Replace individual validation functions with the new type-based annotations:

```python

# Remove these functions from dependencies.py:
# - get_source_id()
# - get_project_id()
# - etc.

# Replace with (validators.py):
SourceID = Annotated[UUID, Depends(lambda source_id: validate_uuid_param(source_id, "source_id"))]
ProjectID = Annotated[UUID, Depends(lambda project_id: validate_uuid_param(project_id, "project_id"))]
```

2. Update all endpoint parameters to use the new annotation types:

```python

# Before:
project_id: Annotated[UUID, Depends(get_project_id)]

# After:
project_id: ProjectID
```

3. Remove the `is_valid_uuid()` function if it's no longer needed elsewhere

itallix:
### Summary

- [x] Replaced validations functions with type-based annotations

### Checklist

### DIFF
--- a/application/backend/app/api/routers/models.py
+++ b/application/backend/app/api/routers/models.py
@@ -49,8 +49,7 @@ def get_model(
 ) -> Model:
     """Get a specific model by ID."""
     try:
-        _ = project  # Ensure project existence
-        return model_service.get_model_by_id(model_id=model_id)
+        return model_service.get_model(project_id=project.id, model_id=model_id)
     except ResourceNotFoundError as e:
         raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(e))
 
@@ -74,8 +73,7 @@ def delete_model(
 ) -> None:
     """Delete a model from a project."""
     try:
-        _ = project  # Ensure project existence
-        model_service.delete_model_by_id(model_id)
+        model_service.delete_model(project_id=project.id, model_id=model_id)
     except ResourceNotFoundError as e:
         raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(e))
     except ResourceInUseError as e:
--- a/application/backend/app/api/routers/models.py
+++ b/application/backend/app/api/routers/models.py
@@ -2,12 +2,12 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from typing import Annotated
-from uuid import UUID
 
 from fastapi import APIRouter, Depends, HTTPException, status
 
-from app.api.dependencies import get_model_id, get_model_service, get_project_id
-from app.schemas import Model
+from app.api.dependencies import get_model_service, get_project
+from app.api.validators import ModelID
+from app.schemas import Model, ProjectView
 from app.services import ModelService, ResourceInUseError, ResourceNotFoundError
 
 router = APIRouter(prefix="/api/projects/{project_id}/models", tags=["Models"])
@@ -23,12 +23,12 @@ router = APIRouter(prefix="/api/projects/{project_id}/models", tags=["Models"])
     },
 )
 def list_models(
-    project_id: Annotated[UUID, Depends(get_project_id)],
+    project: Annotated[ProjectView, Depends(get_project)],
     model_service: Annotated[ModelService, Depends(get_model_service)],
 ) -> list[Model]:
     """Get all models in a project."""
     try:
-        return model_service.list_models(project_id)
+        return model_service.list_models(project.id)
     except ResourceNotFoundError:
         raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Project not found")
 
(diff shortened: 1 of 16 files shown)


application/backend/app/api/routers/models.py
```diff
@@ -43,13 +43,14 @@ def list_models(
...
-    model_id: Annotated[UUID, Depends(get_model_id)],
+    project: Annotated[ProjectView, Depends(get_project)],
+    model_id: ModelID,
     model_service: Annotated[ModelService, Depends(get_model_service)],
 ) -> Model:
     """Get a specific model by ID."""
     try:
-        return model_service.get_model_by_id(project_id=project_id, model_id=model_id)
+        _ = project  # Ensure project existence
+        return model_service.get_model_by_id(model_id=model_id)
```
leoll2: To avoid this strange `_ = project`, the project existence could be checked in `ModelRevisionRepository`, by making this repo project-based (like `DatasetItemRepository` which has a `project_id` attribute, included in all queries).

itallix: Yeah, I agree. I think it's acceptable if we loose project-specific not found exception by including `project_id` into queries

itallix: However, the `project_id` in queries doesn't add any value in this case. We are fine just with the lookup by a primary key. Shall we just remove this check and get rid of `project_id` in API path?

MarkRedeman: In other projects I'd sometimes use a middleware around grouped routes (i.e. if a group of routes has the `/projects/{project_id}/`) which assert the project would be available. (or in Laravel I would use [explicit route model binding](https://laravel.com/docs/12.x/routing#explicit-binding)).
Maybe one of those ideas could be useful here?

The main benefit that I like of those approaches is that you can keep your repositories stateless (except for possibly taking your database connection via DI).

itallix: Thanks, Mark. I would suggest we slightly rearrange our API design around models if there are no objections:
1. Move the listing of project models under the projects router
2. Use `/api/models/{model_id}` for other endpoints since `model_id` uniquely identifies the model and there is no need to include the project ID.

leoll2: > Thanks, Mark. I would suggest we slightly rearrange our API design around models if there are no objections:
> 
> 1. Move the listing of project models under the projects router
> 2. Use `/api/models/{model_id}` for other endpoints since `model_id` uniquely identifies the model and there is no need to include the project ID.

I don't like this solution, it makes the API inconsistent. I prefer all model-related endpoints to stay under the `projects/<id>/models/` prefix for a few reasons:
1. It clearly shows the hierarchy between resources. A model is always part of a project, it cannot exist as a standalone resource (in Geti Tune).
2. If we ever need to implement permissions / access policies like in Geti, it will be much easier with a "hierarchical" endpoints layout than  a "flat" one.
3. It prevents semantic ambiguities and conflicts. If you have a certain resource _"foo"_ that may exist either within a project, or outside of it, then you can represent the two flavors with `api/projects/<id>/foos/<id>` and `api/foos/<id>`, respectively. Here the hierarchy clarifies the scope of the operation and when to validate the project.


Imho, the best approach is the one currently adopted in `[get_dataset_item](https://github.com/open-edge-platform/training_extensions/blob/bb080aeeb13d865506f5a8e1cc4649ec00423497/application/backend/app/api/routers/datasets.py#L160)`: the endpoint has the `get_project` dependency, which ensures that the project existence is validated and a clear error is raised when it doesn't; additionally, the project id is passed to the service and used for the DB query. Filtering by project_id in the repo may seem redundant in this case, but you have to consider that not all repo queries originate from an API request. If the operation is triggered by a background process (e.g. inference pipeline or job), then it makes sense to have this layer of protection in the repos, ensuring that all operations are always properly scoped within the parent entity.

itallix: > but you have to consider that not all repo queries originate from an API request. If the operation is triggered by a background process (e.g. inference pipeline or job), then it makes sense to have this layer of protection in the repos, ensuring that all operations are always properly scoped within the parent entity.

Could you please clarify this point. How exactly introducing redundant `project_id` helps here?

leoll2: > How exactly introducing redundant project_id helps here?

It is defensive programming. If an update/deletion query selects more objects than expected due to an implementation error, we can at least contain the failure within one project, preventing data corruption in other unrelated projects.

leoll2 (approved):

MarkRedeman (approved):
Approving to unblock, but on the UI we should check why `itemId` was inferred to be undefined.
If it can be undefined then we need to do a proper check and not do `String(undefined)` which will lead to bugs.

If `itemId` is not intended to be undefined then the contract from the backend should be changed so that it does not return undefined.

PR #5030, opened 2025-11-24
Delete model artifacts

### LINKED ISSUE #5029: Delete Geti Tune model artifacts
When deleting a model, `ModelService.delete_model` should also remove all artifacts stored in the model folder.

There is a related TODO comment here: https://github.com/open-edge-platform/training_extensions/blob/fadcddeeb1b5cb14217d2ea0f6ba4cee8da6909a/application/backend/app/services/model_service.py#L73

itallix:
### Summary

- [x] Extended `ModelService.delete_model` to remove model revision directory

### Checklist

### DIFF
--- a/application/backend/app/services/model_service.py
+++ b/application/backend/app/services/model_service.py
@@ -79,13 +79,15 @@ class ModelService(BaseSessionManagedService):
                 (e.g., the model is referenced by other entities).
         """
         model_rev_repo = ModelRevisionRepository(project_id=str(project_id), db=self.db_session)
+
+        path = self._projects_dir / str(project_id) / "models" / str(model_id)
+        if path.exists():
+            shutil.rmtree(path)
+
         try:
             deleted = model_rev_repo.delete(str(model_id))
             if not deleted:
                 raise ResourceNotFoundError(ResourceType.MODEL, str(model_id))
-            path = self._projects_dir / str(project_id) / "models" / str(model_id)
-            if path.exists():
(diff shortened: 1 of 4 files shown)


application/backend/app/services/model_service.py
```diff
@@ -70,10 +80,12 @@ def delete_model(self, project_id: UUID, model_id: UUID) -> None:
         """
         model_rev_repo = ModelRevisionRepository(project_id=str(project_id), db=self.db_session)
         try:
-            # TODO: delete model artifacts from filesystem when implemented
             deleted = model_rev_repo.delete(str(model_id))
             if not deleted:
                 raise ResourceNotFoundError(ResourceType.MODEL, str(model_id))
+            path = self._projects_dir / str(project_id) / "models" / str(model_id)
+            if path.exists():
+                shutil.rmtree(path)
```
leoll2: Can be moved outside of the `try...except` block. Alternatively, you can add an `except` clause to catch `OSError` and warn that the file cleanup failed and should be done manually.

itallix: I have moved it outside of the `try...except` but let `OSError` to propagate. 
Would it be still preferable to remove DB record with the warning when `OSError` occur?

leoll2: I think it's okay to let `OSError` propagate, it means that there's a problem in the user deployment. User should fix this issue first, then retry the deletion.

leoll2 (approved):

PR #5037, opened 2025-11-26
Switch from pre-commit to prek

### LINKED ISSUE #5036: Switch from pre-commit to prek
> [prek](https://github.com/j178/prek) is a reimagined version of pre-commit, built in Rust. It is designed to be a faster, dependency-free and drop-in alternative for it, while also providing some additional long-requested features.

For Geti Tune and OTX, prek represents a very interesting tool because it would allow us to address some of the pre-commit's shortcomings for monorepos. In particular, prek allows to define separate pre-commit configurations for the different subprojects, and to run the checks within the virtual envs of the respective subprojects.

leoll2:
### Summary

[`prek`](https://github.com/j178/prek) is a modern alternative to `pre-commit`, which offers better integration capabilities with monorepos (through the 'workspaces' feature) and with uv as a virtual environment manager. It's also implemented in Rust, so it's slightly faster.

This PR:
- Updates and splits the pre-commit configuration for the different subprojects, leveraging prek workspace mode
- Enables `uv-lock` pre-commit check
- Removes `shellcheck` pre-commit as we don't use any shell script for production

Resolves #5036 

### How to test

Install prek:
```sh
curl --proto '=https' --tlsv1.2 -LsSf https://github.com/j178/prek/releases/download/v0.2.19/prek-installer.sh | sh
```

Install and run prek:
```sh
prek install
prek run --all-files
```

image

### Checklist

### DIFF
--- a/application/backend/.pre-commit-config.yaml
+++ b/application/backend/.pre-commit-config.yaml
@@ -8,7 +8,7 @@ repos:
       - id: uv-lock
         name: uv lock
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: "v1.18.2"
+    rev: "v1.16.1"
     hooks:
       - id: mypy
         alias: mypy_backend
--- a/application/backend/.pre-commit-config.yaml
+++ b/application/backend/.pre-commit-config.yaml
@@ -1,0 +1,17 @@
+default_language_version:
+  python: python3.13 # Should match the Python version used by the app (set in backend/pyproject.toml)
+
+repos:
+  - repo: https://github.com/astral-sh/uv-pre-commit
+    rev: "0.9.7"
+    hooks:
+      - id: uv-lock
+        name: uv lock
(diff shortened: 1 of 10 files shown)


application/backend/.pre-commit-config.yaml
```diff
@@ -0,0 +1,17 @@
+default_language_version:
+  python: python3.13 # Should match the Python version used by the app (set in backend/pyproject.toml)
+
+repos:
+  - repo: https://github.com/astral-sh/uv-pre-commit
```
itallix, line 5: Is there a reason why this hook not placed one level above to avoid duplication between `application/backend` and `library`?

leoll2: It doesn't work if put at the top level, because `uv lock` needs to execute from the base folder of each subproject. If it runs at the repo root level, the check gets skipped because of "no files to check".

itallix (approved):

samet-akcay (approved):

PR #5045, opened 2025-11-28
Run Geti Tune container as non-root user

### LINKED ISSUE #5000: Run Geti Tune container as non-root user
The Geti Tune Docker image uses root as default user, which is not ideal from a security perspective.

A non-root user can be added to the Dockerfile as follows:
```dockerfile
RUN useradd -m -u 10001 non-root && chown -R non-root:non-root /application
USER non-root
```

Try to add the non-root user and verify that everything still works correctly. Some adjustments may be necessary, e.g. change nginx to run on a non-privileged port. Privileges may be also needed to attach devices (e.g. cameras).

For reference:
- the Dockerfile of Geti Prompt: https://github.com/open-edge-platform/geti-prompt/blob/main/application/docker/Dockerfile

itallix:
### Summary

- [x] Refactor layers: extract cpu/gpu/xpu base + runtime
- [x] Add `non-root` user

### Checklist

### DIFF
--- a/application/docker/Dockerfile
+++ b/application/docker/Dockerfile
@@ -4,13 +4,14 @@
 # Backend base
 ##############
 FROM python:3.13-slim@sha256:85dfbf1b566b7addfe645faea9938e81a0a01a83580b0ea05fb23706357d77fb AS backend-base
+
+# uv / uvx
 COPY --from=docker.io/astral/uv:0.9.7@sha256:ba4857bf2a068e9bc0e64eed8563b065908a4cd6bfb66b531a9c424c8e25e142 /uv /uvx /bin/
 
-# Set working directory
 WORKDIR /application
 
-# Install required build dependencies
-# TODO remove 'git' once all dependencies are available on PyPI
+# Build deps (for compiling / Rust-based crates)
+# TODO: remove 'git' once all dependencies are on PyPI
 RUN apt-get update  \
     && apt-get install -y --no-install-recommends \
         build-essential=12.12 \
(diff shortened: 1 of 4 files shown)


application/docker/Dockerfile
```diff
@@ -19,18 +20,40 @@ RUN apt-get update  \
...
-# Install dependencies
+##########
+# CPU Base
+##########
+FROM backend-base AS cpu-base
+
+RUN --mount=type=cache,target=/root/.cache/uv \
+    --mount=type=bind,source=backend/uv.lock,target=uv.lock \
+    --mount=type=bind,source=backend/pyproject.toml,target=pyproject.toml \
+    uv sync --frozen --no-dev --no-editable --inexact --extra mqtt --extra cpu
```
leoll2:
```suggestion
    uv sync --frozen --no-dev --no-editable --extra mqtt --extra cpu
```
You can remove `--inexact`. I added it because we had two calls to `uv sync` (first one with `--extra mqtt` to install base deps, second one with `--inexact --extra cpu` to install CPU-specific deps): the `--inexact` flag in the second call ensured that sync would not uninstall the extra packages specified in the first call (mqtt).

application/docker/Dockerfile
```diff
@@ -19,18 +20,40 @@ RUN apt-get update  \
...
+##########
+# GPU Base
+##########
+FROM backend-base AS gpu-base
+
 RUN --mount=type=cache,target=/root/.cache/uv \
     --mount=type=bind,source=backend/uv.lock,target=uv.lock \
     --mount=type=bind,source=backend/pyproject.toml,target=pyproject.toml \
-    uv sync --frozen --no-dev --no-editable --extra mqtt
+    uv sync --frozen --no-dev --no-editable --inexact --extra mqtt --extra cuda
```
leoll2:
```suggestion
    uv sync --frozen --no-dev --no-editable --extra mqtt --extra cuda
```

application/docker/Dockerfile
```diff
@@ -19,18 +20,40 @@ RUN apt-get update  \
...
+##########
+# XPU Base
+##########
+FROM backend-base AS xpu-base
 
-COPY backend/app ./app
+RUN --mount=type=cache,target=/root/.cache/uv \
+    --mount=type=bind,source=backend/uv.lock,target=uv.lock \
+    --mount=type=bind,source=backend/pyproject.toml,target=pyproject.toml \
+    uv sync --frozen --no-dev --no-editable --inexact --extra mqtt --extra xpu
```
leoll2:
```suggestion
    uv sync --frozen --no-dev --no-editable --extra mqtt --extra xpu
```

application/docker/Dockerfile
```diff
@@ -87,28 +113,45 @@ RUN apt-get update \
...
+
+EXPOSE 8080
+
+WORKDIR /application
+ENV PYTHONPATH=/application
+
+USER ${USER_NAME}
 
-EXPOSE 80
+CMD ["sh", "-c", "nginx && exec uv run app/main.py;"]
```
leoll2, line 139: `runtime-base` is an intermediate stage, does CMD have any effect here? Shouldn't it be in the final stages (`geti-tune-cpu`, etc...)?

itallix: It has effect since `runtime-base` is inherited by final stages. It makes sense to define it there since all 3 finals share the same command.

leoll2 (approved):

AlexanderBarabanov (approved):

PR #5065, opened 2025-12-08
Geti Tune training loop

### LINKED ISSUE #4868: Implement training stage of training job
The "training" step is responsible for executing the actual training loop to fine-tune the model (forward / backward passes and evaluation on validation set). At the end of the loop, this step should also convert the resulting model to OpenVINO IR.

To instantiate the `OTXDataModule`: https://github.com/open-edge-platform/training_extensions/blob/33f43c3dbb4c5d9b62293b8bc1999f6e91e550bb/library/src/otx/data/module.py#L225

To instantiate the `OTXEngine`: https://github.com/open-edge-platform/training_extensions/blob/33f43c3dbb4c5d9b62293b8bc1999f6e91e550bb/library/src/otx/backend/native/engine.py#L78

### Sample training script

```python
train_otx_dataset = OTXMulticlassClsDataset(
    dm_subset=train_dm_dataset,
    transforms=[...],  # Setup transforms based on augmentation params
)

data_module = OTXDataModule.from_otx_datasets(
    train_dataset=train_otx_dataset,
    val_dataset=val_otx_dataset,
    test_dataset=test_otx_dataset,
    train_subset=SubsetConfig(batch_size=2),
)

engine = OTXEngine(
    model="efficientnet_b0",
    data=data_module,
)

engine.train(max_epochs=1)
```

leoll2:
### Summary

This PR implements the training job steps that fine-tune the model on the training dataset, evaluate it and finally export it to OpenVINO format. The process to setup the training pipeline involves wrapping the Datumaro subsets into `OTXDataset` objects, using them to initialize an `OTXDataModule` and `OTXEngine`, then using the latter to orchestrate the training, evaluation and export. At the end of the training, artifacts (including the converted model, logs and metrics) are stored in the model folder.

Resolves #4868 

### How to test

Submit a training job with: `curl -X POST -H "Content-Type: application/json" localhost:7860/api/jobs -d '{"job_type": "train", "project_id": "a1b2c3d4-e5f6-7890-abcd-ef1234567890", "parameters": { "model_architecture_id": "Custom_Counting_Instance_Segmentation_MaskRCNN_EfficientNetB2B" }}'`

Monitor the job progress with: `curl localhost:7860/api/jobs`

_Note: the above steps assume a project with some dataset items, annotated and user-reviewed._

### Checklist


application/backend/app/models/partial.py
```diff
@@ -67,7 +67,7 @@ def make_field_optional(field: FieldInfo) -> tuple[Any, FieldInfo]:
         if type(new_field.annotation) is type(BaseModel):
             partial_inner_model = partial_model(cast("type[BaseModel]", new_field.annotation))  # type: ignore[arg-type]
             partial_fields[field_name] = (
-                partial_inner_model | None,
+                partial_inner_model | new_field.annotation | None,
```
leoll2, line 70: I made this change because to allow non-partial objects in the initialization of partial models nested fields.
For example, the following code would raise an error before, because `PartialGlobalParameters` only accepted `PartialGlobalDatasetPreparationParameters` but not `GlobalDatasetPreparationParameters` (the superclass).
```python
params = PartialGlobalParameters(
    dataset_preparation=GlobalDatasetPreparationParameters(
        subset_split=SubsetSplit()
    )
)
```

@A-Artemis please review this change.

application/backend/app/services/training_configuration_service.py
```diff
@@ -113,6 +122,9 @@ def _get_default_configuration(self, project_id: UUID) -> TrainingConfiguration:
         return PartialTrainingConfiguration(
             model_manifest_id=default_model_id,
             hyperparameters=default_model_manifest.hyperparameters.model_dump(),  # type: ignore[arg-type]
+            global_parameters=PartialGlobalParameters(
+                dataset_preparation=GlobalDatasetPreparationParameters(subset_split=SubsetSplit())
+            ),
```
leoll2, line 127: @A-Artemis Please review these changes. My goal is to return `global_parameters` with default values (e.g. the default 70-20-10 training split) instead of `None`.

application/backend/app/services/datumaro_converter.py
```diff
@@ -397,10 +413,13 @@ def _convert_sample(
             if all_with_confidence
             else []
         )
+        # Polygons array is ragged because the number of vertices is not fixed
```
A-Artemis: This looks like an important conversion for polygons. Is this only for instance segmentation, or could it end up being used somewhere else?

application/backend/app/models/partial.py
A-Artemis: Thanks for the fix, but I still think that we need to rework the `TrainingConfiguration` model. With our current implementation we have 2 definitions; a partial, and a complete. This makes it complicated in serializing the object for the endpoint, as well as defining a return type for the endpoint (and its API docs).

I think having `PartialTrainingConfiguration` is too much, instead we should just use the complete training configuration and set its values to None where not in use (different to default value).

application/backend/app/services/training/otx_trainer.py
```diff
@@ -177,17 +292,123 @@ def prepare_model(self, training_params: TrainingParams, dataset_revision_id: UU
...
+        otx_model: OTXModel = model_parser.instantiate_classes(Namespace(model=model_cfg)).get("model")
+        if hasattr(otx_model, "tile_config"):
+            otx_model.tile_config = otx_datamodule.tile_config
+
+        # Set up the OTXEngine
+        logger.info("Initializing the OTXEngine for training (model_id={})", model_id)
+        otx_engine = OTXEngine(
+            model=otx_model,
+            data=otx_datamodule,
+            device=DeviceType.cpu,  # TODO make device configurable (#5040)
```
AlbertvanHouten, line 324: Probably best to leave this as DeviceType.auto if the TODO isn't implemented uet.

application/backend/app/services/training/otx_trainer.py
```diff
@@ -137,34 +159,127 @@ def assign_subsets(self, project_id: UUID) -> None:
...
+            )
+            val_subset_config = SubsetConfig(
+                sampler=SamplerConfig(**val_subset_sampler_cfg_data), **val_subset_cfg_data
+            )
+            test_subset_config = SubsetConfig(
+                sampler=SamplerConfig(**test_subset_sampler_cfg_data), **test_subset_cfg_data
+            )
+            train_subset_config.transforms = TorchVisionTransformLib.generate(train_subset_config)
+            val_subset_config.transforms = TorchVisionTransformLib.generate(val_subset_config)
+            test_subset_config.transforms = TorchVisionTransformLib.generate(test_subset_config)
```
itallix: The repeated `subset config + sampler + transforms` with otx dataset construction below can be extracted into a small helper function parametrized by subset type.

application/backend/app/services/training/otx_trainer.py
```diff
@@ -137,34 +159,127 @@ def assign_subsets(self, project_id: UUID) -> None:
...
+                otx_training_dataset=otx_training_dataset,
+                otx_validation_dataset=otx_validation_dataset,
+                otx_testing_dataset=otx_testing_dataset,
+                otx_training_subset_config=train_subset_config,
+                otx_validation_subset_config=val_subset_config,
+                otx_testing_subset_config=test_subset_config,
+                revision_id=dataset_revision_id,
             )
 
     @step("Prepare Model and Training Configuration")
```
itallix:
```suggestion
    @step("Prepare Model")
```

application/backend/app/services/training/otx_trainer.py
```diff
@@ -137,34 +159,127 @@ def assign_subsets(self, project_id: UUID) -> None:
...
-    def prepare_model(self, training_params: TrainingParams, dataset_revision_id: UUID) -> None:
+    def prepare_model(
+        self, training_params: TrainingParams, dataset_revision_id: UUID, configuration: TrainingConfiguration
+    ) -> None:
         if training_params.project_id is None:
             raise ValueError("Project ID must be provided for model preparation")
```
itallix: Since `project_id` is already validated in `run` method, probably we can just cast it similar to L166:

```suggestion
        project_id = cast(UUID, training_params.project_id)
```

application/backend/tests/unit/services/training/test_otx_trainer.py
```diff
@@ -394,7 +542,284 @@ def test_prepare_model_no_project_id_raises_error(
...
+        fxt_otx_trainer: Callable[[], OTXTrainer],
+        tmp_path: Path,
+    ):
+        """Test successful storing of model artifacts and cleanup."""
+        # Arrange
+        project_id = uuid4()
+        model_id = uuid4()
+
+        # Create a separate data directory for this test
+        test_data_dir = tmp_path / "test_data"
```
itallix: Is there any benefit of creating a separate subdirectory `test_data` within `tmp_path` rather than reusing `tmp_path` directly in this test?

application/backend/app/services/training/otx_trainer.py
leoll2: Thanks for the suggestion, refactored

application/backend/tests/unit/services/training/test_otx_trainer.py
leoll2: No benefit, I've just removed this redundant `test_data` folder

application/backend/app/services/training/otx_trainer.py
leoll2: I prefer to keep `cpu` for now, it is more reliable for development purposes. Also `xpu` is sometimes very slow on the first training (due to the compilation of triton kernels), not the best choice for quick local testing with tiny datasets.

application/backend/app/services/datumaro_converter.py
leoll2: In Geti Tune, it's only needed for instance segmentation, that's the only type with polygon-like annotations.

The "issue" with polygons is that, unlike bounding boxes, they have a variable number of vertices. For example, an image could have a polygon with 45 points and another one with 13 points. The efficient numpy native datatypes (e.g. `np.int32`) can only be used for homogeneous, fixed-size arrays, because numpy allocates contiguous memory to store the array; conversely, if the array shape is irregular, numpy relies on the (less efficient) `object` dtype, that is basically a pointer. For instance segmentation, we have no choice but to use an object array (_ragged_) to stack the polygons on the dimension that varies in length (num vertices). While building the array, however, we must pay attention to some implicit type conversion attempted by numpy. For instance, if the polygons happen to have the same number of vertices, numpy would convert to the native type: that's a problem for us because it makes the array type unpredictable, and Datumaro indeed breaks due to the unexpected array layout.
The new logic makes sure that the outer array _always_ has `dtype=object`, no matter the shape of the inner arrays.

application/backend/app/models/partial.py
leoll2: Yeah, I agree we'll need to revisit it later and remove the duplication if possible. For this PR, I would keep it as it is.

leoll2:
e2e tests for this feature will be added in a separate PR later (https://github.com/open-edge-platform/training_extensions/issues/5104).

application/backend/tests/unit/services/training/test_otx_trainer.py
```diff
@@ -366,35 +526,275 @@ def test_prepare_model_success(
...
+            dataset_revision_service=Mock(spec=DatasetRevisionService),
+            model_service=Mock(spec=ModelService),
+            training_configuration_service=Mock(spec=TrainingConfigurationService),
+            db_session_factory=Mock(),
+        )
+        otx_trainer = OTXTrainer(training_deps)
+        execution_ctx = Mock(spec=ExecutionContext)
+        execution_ctx.report = Mock()
+        execution_ctx.heartbeat = Mock()
+        otx_trainer._ctx = execution_ctx
```
itallix: Is it replaceable with `fxt_otx_trainer` fixture? The setup looks very similar

application/backend/app/services/datumaro_converter.py
itallix: The original comment doesn't not reflect some important context. Can we extend it with this additional detail about implicit conversion and type mismatch?

application/backend/tests/unit/services/training/test_otx_trainer.py
leoll2: Done

application/backend/app/services/datumaro_converter.py
leoll2: Sure, I just added an extended comment in the code (6c7c4fabf0a481440e7a40fee0f54c4be7478fb7)

itallix (approved):

A-Artemis (approved):

PR #5070, opened 2025-12-09
Dataset revision endpoint implementation

### LINKED ISSUE #5052: Implement endpoints to list items from dataset revisions
After #4995, finalize the implementation of the endpoints to list the items of a dataset revision. Note that dataset revisions are persisted as parquet files (as opposed to Sqlite), so it's necessary to query the polars dataframe to find the records corresponding to the items.

This issue requires some exploration of the technical paths, to answer the following questions:
- Is it better to store the dataset revisions as zip or uncompressed? [(ref)](https://github.com/open-edge-platform/training_extensions/blob/62e7923caf4589ed2a10d3a45b6c96b3b8432bd4/application/backend/app/services/dataset_service.py#L407) How much disk space can we save on average by storing the datasets compressed? And how much overhead (latency) to decompress the dataset, wrt to the dataset size?
-  Is it better to implement our query logic with polars, loading the parquet file directly, or should we load the dataset through Datumaro and delegate the filtering to it?

A-Artemis:
### Summary

Adds the implementation for the Dataset Revision endpoints.
- get
- get binary/thumbnail
- list
- delete

BENCHMARK 1: Storage Format Comparison

Storage Format       Size            Export Time     Savings
#----------------------------------------------------------------------
Uncompressed         2.93 MB         0.210s                
Zip Compressed       2.93 MB         0.374s          0.2%

BENCHMARK 2: Decompression Overhead

Extraction time: 0.175s
Load from uncompressed: 0.165s
Total overhead (extract + load): 0.340s

BENCHMARK 3: Query Performance

Method 1: Direct Polars Query
  Select all                               2.84ms (100 rows)
  Filter by subset                         3.29ms (80 rows)
  Paginate (limit 10, offset 0)            1.77ms (10 rows)
  Paginate (limit 10, offset 50)           1.62ms (10 rows)

Method 2: Datumaro Dataset Query
  Dataset load time                        142.69ms
  Dataset length                           100 items

#--------------------------------------------------------------------------------
Query Performance Summary:
  Average Polars query time:               2.38ms
  Datumaro load time:                      142.69ms

Polars is significantly faster for querying (59.9x faster than Datumaro load)

### Checklist


application/backend/app/services/dataset_service.py
```diff
@@ -456,6 +459,184 @@ def delete_dataset_revision_files(self, project_id: UUID, revision_id: UUID) ->
...
+            # Extract to temp directory
+            temp_dir = Path(tempfile.mkdtemp(prefix=f"revision_{revision_id}_"))
+            with ZipFile(zip_path, "r") as zip_ref:
+                zip_ref.extractall(temp_dir)
+            return temp_dir / "data.parquet"
+
+        # Otherwise it's uncompressed (though this shouldn't happen with current implementation)
+        return revision_path / "data.parquet"
+
+    def list_dataset_revision_items(
```
leoll2: I suggest to move the new logic to a separate service `DatasetRevisionService`, it is easier to maintain and there is not really much in common with the other methods of `DatasetService`.

application/backend/app/api/routers/dataset_revisions.py
```diff
@@ -33,15 +36,45 @@
...
+
+    items = []
+    for item_data in items_data:
+        item_view = DatasetItemView(
+            id=UUID(item_data["id"]),
+            name=Path(item_data["image"]).stem,
+            format=DatasetItemFormat(Path(item_data["image"]).suffix.lstrip(".")),
+            width=item_data["image_info"]["width"],
+            height=item_data["image_info"]["height"],
+            size=0,  # Not available in parquet
```
leoll2: It's generally better to use `None` to represent "unknown" instead of 0. In this case, I'd suggest to define a new view model `DatasetRevisionItemView`, similar to `DatasetItemView` but without `size` and `source_id`.

application/backend/app/api/routers/dataset_revisions.py
```diff
@@ -89,13 +142,18 @@ def get_dataset_revision_item_binary(
...
-    _dataset_item_id: DatasetItemID,
-    _dataset_service: Annotated[DatasetService, Depends(get_dataset_service)],
+    project: Annotated[Project, Depends(get_project)],
+    dataset_revision: Annotated[DatasetRevision, Depends(get_dataset_revision)],
+    dataset_item_id: DatasetItemID,
+    dataset_service: Annotated[DatasetService, Depends(get_dataset_service)],
 ) -> FileResponse:
     """Get the thumbnail of an item in the dataset revision"""
-    raise NotImplementedError
+    binary_path = dataset_service.get_dataset_revision_item_binary_path(
```
leoll2: Returning the full image as thumbnail is not useful; we can address the thumbnail problem in another PR, for example we can generate the thumbnail lazily and cache it (next to the dataset revision) for subsequent requests.

application/backend/app/services/dataset_service.py
```diff
@@ -456,6 +459,184 @@ def delete_dataset_revision_files(self, project_id: UUID, revision_id: UUID) ->
...
+            revision_id: The UUID of the dataset revision.
+
+        Returns:
+            Path to the data.parquet file (may be in a temp directory if extracted from zip).
+        """
+        revision_path = self.projects_dir / str(project_id) / "dataset_revisions" / str(revision_id)
+
+        # When as_zip=True, datumaro creates dataset.zip inside the output directory
+        zip_path = revision_path / "dataset.zip"
+        if zip_path.exists():
```
leoll2: Before checking if the zip path exists, you could check if `data.parquet` already exists.

Pseudocode:
```
if parquet file exists:
  return parquet file path
elif zip file exist:
  uncompress
  return parquet file path
else:
  raise exception
```

application/backend/app/services/dataset_service.py
```diff
@@ -456,6 +459,184 @@ def delete_dataset_revision_files(self, project_id: UUID, revision_id: UUID) ->
...
+            # Apply pagination
+            df = df.slice(offset, limit)
+
+            # Convert to list of dicts
+            items = df.to_dicts()
+
+            return items, total_count
+        finally:
+            # Clean up temp directory if we extracted a zip
+            if "revision_" in str(parquet_path.parent.name) and parquet_path.parent.parent.name != "dataset_revisions":
```
leoll2: Cleanup should be ideally done in the same context where you allocate the resource (i.e. close to where you create the temp directory)

application/backend/app/api/schemas/dataset_item.py
```diff
@@ -87,7 +111,7 @@ class DatasetItemsWithPagination(BaseModel):
     Dataset Items list with pagination info
     """
 
-    items: list[DatasetItemView]
+    items: list[DatasetItemView | DatasetItemRevisionView]
```
leoll2: Please define a separate model `DatasetRevisionItemsWithPagination` for the revision endpoints, so that the UI doesn't have to deal with the union type (annoying for type-checking).

application/backend/app/services/dataset_revision_service.py
```diff
@@ -0,0 +1,282 @@
...
+        if revision_path.exists():
+            shutil.rmtree(revision_path)
+            logger.info("Deleted dataset revision files at '{}'", revision_path)
+
+        # Mark as deleted in the database
+        revision_repo = DatasetRevisionRepository(db=self.db_session)
+        revision_db = revision_repo.get_by_id(str(revision_id))
+        if revision_db:
+            revision_db.files_deleted = True
+            revision_repo.save(revision_db)
```
leoll2: To prevent race conditions, it's better to update the DB first (mark the revision as deleted -> inaccessible) and delete the files immediately after.

application/backend/app/services/dataset_revision_service.py
```diff
@@ -0,0 +1,282 @@
...
+            project_id: The UUID of the project.
+            revision_id: The UUID of the dataset revision.
+            limit: Maximum number of items to return.
+            offset: Number of items to skip.
+            subset: Optional subset filter.
+
+        Returns:
+            Tuple of (list of items as dicts, total count).
+        """
+        parquet_path, is_temp = self._get_revision_parquet_path(project_id, revision_id)
```
leoll2: In the current implementation, `list_dataset_revision_items` would uncompress the dataset everytime it is called. This doesn't play well with pagination: if e.g. you call the endpoints 3 times to get 150 items in pages of 50, then you uncompress the same dataset 3 times in a row, which is clearly inefficient.

Assuming that storing the dataset revisions in a compressed zip format would significantly reduce the disk usage (to be verified empirically), then we need a better approach to uncompress the dataset when needed and keep it available, uncompressed, for subsequent requests in a short time frame. For example, we could implement a mechanism that caches the uncompressed datasets somewhere, and automatically deletes them after some time (e.g. a few hours) or when the disk space gets low.

Anyway, to make a proper decision, we first need answers to the questions in the ticket.

application/backend/app/api/routers/dataset_revisions.py
```diff
@@ -33,33 +40,74 @@
...
+    for item_data in items_data:
+        item_view = DatasetItemRevisionView(
+            id=UUID(item_data["id"]),
+            name=Path(item_data["image"]).stem,
+            format=DatasetItemFormat(Path(item_data["image"]).suffix.lstrip(".")),
+            width=item_data["image_info"]["width"],
+            height=item_data["image_info"]["height"],
+            subset=DatasetItemSubset(item_data["subset"]),
+        )
+        items.append(item_view)
```
leoll2: Ideally, this logic should be in `DatasetRevisionService.list_dataset_revision_items`, which should return an object of type `app.models.dataset_item_revision.DatasetRevisionItem` (the domain entity) instead of a raw dictionary. The endpoint should then convert to the view type with `DatasetItemRevisionView.model_validate(dataset_revision_item, from_attributes=True)`.

For reference, you can check the implementation of `list_dataset_items` in `api/routers/datasets.py`.

leoll2 (commented):
LGTM, only one suggestion to improve the service interface through the use of business-layer entities, like the other services do.

leoll2 (approved):

PR #5076, opened 2025-12-11
Bugfixes in OTXDataModule

leoll2:
### Summary

This PR addresses a bunch of issues that I encountered when working with `OTXDataModule`.

Changes:
- Bugfix: `extract_normalization_params` no longer throws an error if the transforms are given in the form of a `Compose` object rather than a list. It now works fine with both `Compose` and iterables.
- Bugfix: the `image_size` attribute would assume an incorrect value when `OTXDataModule` is initialized from the new Datumaro datasets, because the deprecated `img_info.shape` attribute is not updated by the transforms (always refers to the original image size)
- Fix an issue where the transforms provided through the input dataset would be silently overridden by default transforms in the SubsetConfig
  - For clarify, transforms are now _always_ taken from the dataset; if the user attempts to pass them through SubsetConfig, they are ignored and a warning is raised.

### How to test

Manually instantiate a data module and train.

### Checklist

### DIFF
--- a/library/src/otx/data/module.py
+++ b/library/src/otx/data/module.py
@@ -7,7 +7,7 @@ from __future__ import annotations
 
 import logging as log
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, Sequence
 
 from datumaro import Dataset as DmDataset
 from lightning import LightningDataModule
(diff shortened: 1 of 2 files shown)


AlbertvanHouten (dismissed):

library/src/otx/data/module.py
```diff
@@ -316,27 +339,35 @@ def from_otx_datasets(
...
+                    default_subset_configs = instance.get_default_subset_configs(instance.input_size)
                 subset_to_assign = default_subset_configs[f"{name}_subset"]
-                # Override transforms with the ones from the pre-constructed dataset
-                subset_to_assign.transforms = instance.subsets[name].transforms  # type: ignore[assignment]
+
+            # Override transforms with the ones from the pre-constructed dataset
+            subset_to_assign.transforms = instance.subsets[name].transforms  # type: ignore[assignment]
 
             # Assign subset config and transforms to subset dataset
             setattr(instance, f"{name}_subset", subset_to_assign)
```
kprokofi, line 373: `instance.subsets[name].transforms = subset_to_assign.transforms` (line 374)  can be removed now, we don't change the transforms in datasets

leoll2: Removed

AlbertvanHouten (approved):

kprokofi (approved):

PR #5071, opened 2025-12-10
Fix integration tests not running

AlbertvanHouten:
### Summary

List of tasks are wrongly defined in the GHA workflow causing only semantic_segmentation tests to run.

### Checklist

- [ ] The PR title and description are clear and descriptive
- [ ] I have manually tested the changes
- [ ] All changes are covered by automated tests
- [ ] All related issues are linked to this PR (if applicable)
- [ ] Documentation has been updated (if applicable)

PR #5079, opened 2025-12-12
Fix export for problematic models & Integration tests

### LINKED ISSUE #5073: Export to ONNX fails for models with NMS stage
Models that have a NMS stage, such as Yolo and Mask-RCNN, fail to be exported to ONNX (or OpenVINO via ONNX).

The following operation fails:
```python
exported_model_path = otx_engine.export(
    checkpoint=trained_yolo_path,
    export_format=OTXExportFormatType.OPENVINO,
    export_precision=OTXPrecisionType.FP16,
)
```
with error:
```
torch.fx.experimental.symbolic_shapes.GuardOnDataDependentSymNode: Could not guard on data-dependent expression u1 < 200 (unhinted: u1 < 200).  (Size-like symbols: u1)                                                                                                                                                                                                                                                                                                                                        consider using data-dependent friendly APIs such as guard_or_false, guard_or_true and statically_known_trueCaused by: (_decomp/decompositions.py:757 in slice_forward)
For more information, run with TORCH_LOGS=\"dynamic\"
For extended logs when we create symbols, also add TORCHDYNAMO_EXTENDED_DEBUG_CREATE_SYMBOL=\"u1\"
If you suspect the guard was triggered from C++, add TORCHDYNAMO_EXTENDED_DEBUG_CPP=1
For more debugging help, see https://docs.google.com/document/d/1HSuTTVvYH1pTew89Rtpeu84Ht3nQEFTYhAX3Ypa_xJs/edit?usp=sharing
 
For C++ stack trace, run with TORCHDYNAMO_EXTENDED_DEBUG_CPP=1
 
The following call raised this error:
  File \"/home/leoll2/training_extensions/application/backend/.venv/lib/python3.13/site-packages/otx/backend/native/models/common/utils/nms.py\", line 53, in forward
    inds = inds[:max_num]
```
The full stacktrace can be also seen in the [integration tests logs](https://github.com/open-edge-platform/training_extensions/actions/runs/20094067885/job/57648974416?pr=5071).

For exporting to OpenVINO, a workaround is to bypass the intermediate ONNX conversion (set `via_onnx=False` [here](https://github.com/open-edge-platform/training_extensions/blob/8cbbbcc642d729708541d146393c5dae52d20ee0/library/src/otx/backend/native/models/detection/yolox.py#L153)), but it doesn't address the root cause.

kprokofi:
### Summary

resolves https://github.com/open-edge-platform/training_extensions/issues/5073

- [x] Fix issues with MaskRCNN and empty masks on inference
- [x] Fix ONNX and OV export for OD and IS models
- [x] Ensure all integration tests are passed

### Checklist


library/src/otx/backend/native/models/instance_segmentation/maskrcnn.py
```diff
@@ -363,8 +363,9 @@ def _exporter(self) -> OTXModelExporter:
                     "labels": {0: "batch", 1: "num_dets"},
                     "masks": {0: "batch", 1: "num_dets", 2: "height", 3: "width"},
                 },
-                "opset_version": 11,
+                "opset_version": 18,
                 "autograd_inlining": False,
+                "dynamo": False,
```
leoll2, line 368: What problem does "dynamo" have with this specific model (Mask RCNN)?

library/src/otx/backend/native/models/instance_segmentation/maskrcnn_tv.py
```diff
@@ -244,7 +244,7 @@ def _exporter(self) -> OTXModelExporter:
             resize_mode="fit_to_window",
             pad_value=0,
             swap_rgb=False,
-            via_onnx=True,
+            via_onnx=False,
```
leoll2, line 288: Since `via_onnx=False`, can we avoid passing `onnx_export_configuration` (L248-259)? Or do you prefer to keep it in case we switch back to `True` in the future?

library/src/otx/backend/native/models/detection/heads/yolox_head.py
```diff
@@ -369,8 +369,7 @@ def export_by_feat(  # type: ignore[override]
 
         if not with_nms:
             return bboxes, scores
-
-        return multiclass_nms(
+        nms = multiclass_nms(
```
leoll2: Any specific reason to use a variable here instead of `return` directly?

leoll2 (commented):
A couple of questions:

1. Should we set `via_onnx=False` for *all* models, to have a consistent export flow for all models?
2. Can you try to cherrypick the commit from this Albert's branch, to re-enable the integration tests that check the export? https://github.com/open-edge-platform/training_extensions/pull/5071

library/src/otx/backend/native/models/detection/heads/yolox_head.py
kprokofi: already removed

library/src/otx/backend/native/models/instance_segmentation/maskrcnn_tv.py
kprokofi: onnx_export_configuration is needed for ONNX export. We still need to support both.

library/src/otx/backend/native/models/instance_segmentation/maskrcnn.py
kprokofi: It cannot handle ROI extractor properly.
The error is in single_level_roi_extractor.py. With Dynamo-based export, the condition if inds.numel() > 0 creates a data-dependent guard that Dynamo can't handle because inds.numel() is a symbolic value at trace time.

The easiest fix - use standard export

kprokofi:
> A couple of questions:
> 
> 1. Should we set `via_onnx=False` for _all_ models, to have a consistent export flow for all models?
> 2. Can you try to cherrypick the commit from this Albert's branch, to re-enable the integration tests that check the export? #5071

1. I checked those models with via_onnx=True and some of them still cannot be exported directly to OV without code change. So, some of the models should be still using ONNX first.
2. Yes, I can

library/src/otx/backend/native/models/instance_segmentation/maskrcnn_tv.py
leoll2: Right, I thought this function was only for exporting to OpenVINO but it's actually for ONNX too.

library/src/otx/recipe/_base_/data/keypoint_detection.yaml
```diff
@@ -52,6 +52,9 @@ train_subset:
       init_args:
         mean: [123.675, 116.28, 103.53]
         std: [58.395, 57.12, 57.375]
+  sampler:
+    class_path: torch.utils.data.RandomSampler
+    init_args: null
```
leoll2, line 57: Is this change related to the PR?

kprokofi: I found raising an error from jsonargpass for this task

leoll2 (approved):

AlbertvanHouten (approved):

AlexanderBarabanov (approved):

PR #5081, opened 2025-12-13
Add RFDetr model for Object Detection and Instance Segmentation

### LINKED ISSUE #5011: Add RF-DETR Object Detection Model
**Is your feature request related to a problem? Please describe.**

Currently, OpenVINO Training Extensions offers excellent support for CNN-based object detectors like YOLO variants, but there's a gap in state-of-the-art transformer-based architectures that achieve both high accuracy and real-time performance. While RT-DETR is supported, the ecosystem would benefit from newer transformer architectures that push the boundaries of accuracy-speed trade-offs, especially for applications requiring strong domain adaptability and end-to-end deployment without complex post-processing like Non-Maximum Suppression (NMS).

**Describe the solution you'd like to propose.**

I propose integrating **RF-DETR (Roboflow Detection Transformer)** as a native object detection model in OTX. RF-DETR is the first real-time transformer-based detector to surpass 60 mAP on COCO benchmark while maintaining competitive inference speeds. Key benefits include:

- **State-of-the-art performance**: 60.5 mAP (Base) and 64.2 mAP (Large) on COCO, outperforming YOLOv11 and RT-DETR
- **Real-time inference**: 25 FPS (Base) on NVIDIA T4, with new Nano/Small/Medium variants scaling to 100+ FPS
- **No NMS required**: True end-to-end detection simplifies deployment and reduces latency
- **Apache 2.0 license**: Fully compatible with OTX's open-source model
- **Domain adaptability**: Superior performance on RF100-VL benchmark across 100+ diverse real-world domains (aerial, industrial, medical, etc.)
- **Multiple model sizes**: Nano (3.2M params) → Large (129M params) for flexible deployment from edge to cloud

The integration would leverage RF-DETR's existing Python package and follow OTX's pattern of supporting transformer-based architectures, similar to how RT-DETR was integrated.

**Describe alternatives you've considered.**

1. **RT-DETR**: Already supported in OTX, but RF-DETR offers higher accuracy (60+ vs ~54 mAP) and better domain generalization
2. **DEIM/DFINE**: While fast, RF-DETR offers faster inference and higher accuracy. 
3. **Custom implementation**: Re-implementing from scratch would be redundant given RF-DETR's mature, open-source package and active maintenance by Roboflow

**Additional context**

- **Paper**: "RF-DETR: Neural Architecture Search for Real-Time Detection Transformers" 
- **Repository**: https://github.com/roboflow/rf-detr
- **PyPI Package**: `rf-detr` (Apache 2.0 license)
- **Architecture**: Built on DINOv2 backbone + LW-DETR with Deformable Attention, offering excellent transfer learning capabilities
- **Industry adoption**: RF100-VL benchmark is used by Apple, Microsoft, Baidu for evaluating real-world detector performance
- **OTX alignment**: Fits perfectly with OTX's roadmap of integrating Transformers library and third-party backends while maintaining unified CLI/API

This would position OTX as the premier framework for both CNN and transformer-based real-time detection, giving users more options for accuracy-speed trade-offs without leaving the OTX ecosystem.

### LINKED ISSUE #5017: Add RF-DETR Instance Segmentation Model
Same as #5011, but for instance segmentation.

Reference implementation: https://github.com/roboflow/rf-detr

> RF-DETR is a real-time, transformer-based object detection and instance segmentation model architecture developed by Roboflow and released under the Apache 2.0 license.
RF-DETR is the first real-time model to exceed 60 AP on the Microsoft COCO object detection benchmark alongside competitive performance at base sizes. It also achieves state-of-the-art performance on RF100-VL, an object detection benchmark that measures model domain adaptability to real world problems. RF-DETR is fastest and most accurate for its size when compared current real-time objection models.
On image segmentation, RF-DETR Seg (Preview) is 3x faster and more accurate than the largest YOLO when evaluated on the Microsoft COCO Segmentation benchmark, defining a new real-time state-of-the-art for the industry-standard benchmark in segmentation model evaluation.

kprokofi:
### Summary

- [x] Add RFDetr model (OD)
- [x] Add RFDetr model (IS)
- [x] Add unit tests
- [x] Fine-tune hyper parameters (LR, Augmentations, EMA, Scheduling, BS, Grad accumulation, multi-scale training)
- [x] Resolve an issue with filtering predictions during test (filtering predictions before computing mAP -> destroyed the metric. Now we do filtering only in predict method and directly use threshold in F1-score metric)
- [x] Upgrade TorchMetrics and Lightning to use EMA. Fix Dice metric for segmentation
- [ ] Provide full benchmark

### Checklist


library/pyproject.toml
```diff
@@ -133,6 +140,9 @@ Repository = "https://github.com/open-edge-platform/training_extensions/"
 
 [tool.uv]
 default-groups = ["dev", "lint"]
+override-dependencies = [
+    "torch==2.10.0",
```
leoll2: I suggest to add a comment to explain this override. I assume it's because `rfdetr` torch is locked to 2.8.

library/pyproject.toml
```diff
@@ -53,7 +53,14 @@ dependencies = [
     "numpy>=2.0",
     "pyyaml==6.0.3",
     "tensorboard==2.20.0",
-    "tensorboardx==2.6.4"
+    "tensorboardx==2.6.4",
+    "rfdetr @ git+https://github.com/kprokofi/rf-detr.git@develop",
```
leoll2: Instead of depending on a fork, it would be better to open a PR to the main repo with the export fixes; it is more maintainable in the long run. Of course it may take a while to get it reviewed and merged, so the fork is okay as a temporary solution.

library/pyproject.toml
```diff
@@ -53,7 +53,14 @@ dependencies = [
...
+    "aiohttp==3.13.3",
+    "certifi>=2024.7.4",
+    "filelock>=3.20.3",
+    "idna>=3.7",
+    "urllib3>=2.6.3",
+    "Werkzeug>=3.1.5",
```
leoll2: Where do these dependencies come from? I don't see being imported in our source code.

library/src/otx/backend/native/callbacks/lr_monitor.py
```diff
@@ -0,0 +1,86 @@
+# Copyright (C) 2024-2025 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
```
leoll2: Stale copyright header

library/src/otx/backend/native/callbacks/gpu_mem_monitor.py
```diff
@@ -66,6 +81,26 @@ def on_train_batch_start(
...
 
+    def on_train_batch_end(
+        self,
+        trainer: Trainer,
+        pl_module: LightningModule,
+        outputs: Any,  # noqa: ANN401
+        batch: Any,  # noqa: ANN401
+        batch_idx: int,
+    ) -> None:
+        """Clean up CUDA cache every 50 train iterations if enabled.
```
leoll2: Why do you have to explicitly cleanup the cache? Also, the docstring says 50 iterations but the actual number is configured in the constructor.

library/src/otx/metrics/dice.py
```diff
@@ -30,45 +30,34 @@ def _segm_callable(label_info: SegLabelInfo) -> MetricCollection:
     )
 
 
-class OTXDice(Dice):
+class OTXDice(DiceScore):
```
leoll2, line 33: Can you add a unit test for this metric? The other metrics are validated in `tests/unit/metrics/`, but for some reason this metric (OTXDice) doesn't.

library/src/otx/data/entity/sample.py
```diff
@@ -173,16 +173,14 @@ class DetectionSample(OTXSample):
     dm_image_info: DmImageInfo = image_info_field()
 
     def __post_init__(self) -> None:
-        shape = (self.dm_image_info.height, self.dm_image_info.width)
-
         # Convert bboxes to tv_tensors format
-        if isinstance(self.bboxes, np.ndarray):
```
leoll2, line 179: Why are you removing this check? `bboxes` can  be initialized with a `tv_tensors.BoundingBoxes`, in such case the conversion shouldn't be performed. Also note that `InstanceSegmentationSample` performs the same check, but you didn't remove it.

library/src/otx/backend/native/engine.py
```diff
@@ -166,7 +167,7 @@ def train(
         min_epochs: int = 1,
         seed: int | None = None,
         deterministic: bool | Literal["warn"] = False,
-        precision: _PRECISION_INPUT | None = 16,
+        precision: _PRECISION_INPUT | None = "16-mixed",
```
leoll2, line 170: Is this change a bugfix, or should it be reverted?

library/src/otx/backend/native/engine.py
```diff
@@ -909,7 +910,7 @@ def configure_accelerator(self) -> None:
         if self._device.accelerator == DeviceType.xpu:
             self._cache.update(strategy="xpu_single")
             # add plugin for Automatic Mixed Precision on XPU
-            if self._cache.args.get("precision", 32) == 16:
+            if self._cache.args.get("precision", 32) in [16, "16-mixed", "bf16-mixed", "bf16"]:
```
leoll2: Is the precision cached always cached as a string or can it also be an integer? Allowing both seems unnecessary and fragile, I would limit it to string.

library/src/otx/backend/native/models/detection/backbones/hgnetv2.py
```diff
@@ -481,9 +481,9 @@ def __init__(
         self.use_lab = use_lab
         self.return_idx = return_idx
 
-        stem_channels = self.arch_configs[name]["stem_channels"]
-        stage_config = self.arch_configs[name]["stage_config"]
-        download_url = self.arch_configs[name]["url"]
+        stem_channels: list[int] = self.arch_configs[name]["stem_channels"]  # pyrefly: ignore[bad-assignment]
+        stage_config: dict[str, list] = self.arch_configs[name]["stage_config"]  # pyrefly: ignore[bad-assignment]
+        download_url: str = self.arch_configs[name]["url"]  # pyrefly: ignore[bad-assignment]
```
leoll2: What's the actual cause of these `bad-assignment` warnings? The `arch_configs` dicts are hardcoded, in theory Pyrefly should be able to infer the correct types.

library/src/otx/backend/native/models/detection/rfdetr.py
```diff
@@ -0,0 +1,389 @@
...
+    from pathlib import Path
+
+    from lightning.pytorch.cli import LRSchedulerCallable, OptimizerCallable
+
+    from otx.backend.native.schedulers import LRSchedulerListCallable
+    from otx.metrics import MetricCallable
+    from otx.types.label import LabelInfoTypes
+
+
+class RFDETR(DFine):
```
leoll2: Why does `RFDETR` subclass `DFine`? As far as I know, they are independent model architectures.

library/src/otx/backend/native/models/utils/utils.py
```diff
@@ -90,7 +90,7 @@ def load_checkpoint(
     if Path(checkpoint).exists():
         load_checkpoint_to_model(
             model,
-            torch.load(checkpoint, map_location),
+            torch.load(checkpoint, map_location, weights_only=False),
```
leoll2: `weights_only=False` is a potential security liability, especially since we download weights from 3rd party sources. Can you motivate your need to set this parameter to `False`?

library/pyproject.toml
kprokofi: I will remove it, my fork allows torch 2.10

library/pyproject.toml
kprokofi: Yes, that is my plan

library/pyproject.toml
kprokofi: I will check, I added this to resolve several env issues

library/src/otx/metrics/dice.py
kprokofi: Ok

library/src/otx/data/entity/sample.py
kprokofi: No, it cannot be, Datumaro doesn't support tv_tensors.BoundingBoxes. Anyway, I will add a check comparing to tv_tensors.BoundingBoxes class.

library/src/otx/backend/native/engine.py
kprokofi: precision = 16 is deprecated, "16-mixed" is recommended

library/src/otx/backend/native/engine.py
kprokofi: It can be integer as well. A user can pass an integer to the train() method.

library/src/otx/backend/native/models/detection/rfdetr.py
kprokofi: They shared some methods like prepare_outputs(). However, since I added object capping functionality, we can avoid this inheritance

library/src/otx/data/entity/sample.py
leoll2: CC @AlbertvanHouten

library/src/otx/backend/native/models/utils/utils.py
kprokofi: Agree, I think there was an issue and I added it to avoid it. I forgot to revert it

library/src/otx/data/entity/sample.py
AlbertvanHouten: There is an issue in the datumaro repo to track it. I can priotize it next week if needed https://github.com/open-edge-platform/datumaro/issues/1940

library/pyproject.toml
kprokofi: deleted

library/pyproject.toml
kprokofi: removed

library/src/otx/backend/native/callbacks/lr_monitor.py
kprokofi: Fixed

library/src/otx/backend/native/callbacks/gpu_mem_monitor.py
kprokofi: Removed

library/src/otx/metrics/dice.py
kprokofi: Added

library/src/otx/backend/native/models/detection/backbones/hgnetv2.py
kprokofi: image

kprokofi: It cannot recognize correct typing from nested dict

library/src/otx/backend/native/models/detection/rfdetr.py
kprokofi: Done

library/src/otx/data/entity/sample.py
kprokofi: I returned it back. I will remove numpy in my augmentations PR

leoll2 (dismissed):
Outstanding work, thank you!

kprokofi:
Instance Seg (Vs MaskRCNN-R50)

image

Det (vs DEIMV2)
image
image

kprokofi:
Model Manifests will be added in the next PR after final benchmark with all models

library/src/otx/backend/native/models/detection/rfdetr.py
```diff
@@ -0,0 +1,388 @@
...
+            rfdetr_args=self.rfdetr_args,  # pyrefly: ignore[bad-argument-type]
+            input_size=self.data_input_params.input_size[0],
+            multi_scale=self.multi_scale,
+        )
+
+    def _customize_inputs(
+        self,
+        entity: OTXSampleBatch,
+        pad_size_divisor: int = 32,
+        pad_value: int = 0,
```
eugene123tw: Is there any reason keeping `pad_size_divisor` and `pad_value`?

library/src/otx/backend/native/models/detection/rfdetr.py
```diff
@@ -0,0 +1,388 @@
...
+
+    def export(
+        self,
+        output_dir: Path,
+        base_name: str,
+        export_format: OTXExportFormatType,
+        precision: OTXPrecisionType = OTXPrecisionType.FP32,
+    ) -> Path:
+        """Export the model to the requested format."""
+        self.model.lwdetr.export()  # pyrefly: ignore[missing-attribute]
```
eugene123tw: `lwdetr.export()` permanently mutates the model with no rollback

LWDETR.export() replaces self.forward with self.forward_export in-place (and recursively for all submodules). There is no restore() / unexport() method — this is a one-way mutation. If super().export(...) fails (ONNX trace error, disk issue, etc.), the model is left in a corrupted state where forward points to the export path, which skips NestedTensor wrapping, ignores group_detr, and behaves differently for masks.

In practice this is probably fine since export is typically the last operation on a model instance, but it would be safer to guard against partial failure:

```python
        """Export the model to the requested format."""
        if self.explain_mode:
            msg = "Explain mode is not supported for RF-DETR model."
            raise ValueError(msg)
        self.model.lwdetr.export()  # pyrefly: ignore[missing-attribute]
        try:
            return super().export(output_dir, base_name, export_format, precision)
        except Exception:
            # Restore forward methods if export fails, so the model remains usable
            if hasattr(self.model.lwdetr, "_forward_origin"):
                self.model.lwdetr.forward = self.model.lwdetr._forward_origin
                self.model.lwdetr._export = False
            raise
```

Also worth noting: the same pattern exists in `rfdetr_inst.py`. Both should be updated together. Ideally your rfdetr fork could add a `restore()` method to LWDETR for proper cleanup.

library/src/otx/backend/native/models/detection/rfdetr.py
```diff
@@ -0,0 +1,388 @@
...
+
+    Args:
+        label_info: Information about the labels.
+        data_input_params: Parameters for image data preprocessing.
+        model_name: Name of the model variant to use.
+        optimizer: Callable for the optimizer.
+        scheduler: Callable for the learning rate scheduler.
+        metric: Callable for the metric.
+        multi_scale: Whether to use multi-scale training.
+        torch_compile: Whether to use torch compile.
```
eugene123tw, line 52: Could you add a docstring for `max_total_objects_per_batch`?

library/src/otx/backend/native/models/detection/rfdetr.py
```diff
@@ -0,0 +1,388 @@
...
+                        "boxes": scaled_bboxes,
+                        "labels": ll,
+                        "size": torch.tensor([h, w], device=device),
+                        "orig_size": torch.tensor([h, w], device=device),
+                    }
+                )
+
+        # Apply batch-level object limiting if configured (only during training)
+        if self.training and self.max_total_objects_per_batch is not None:
+            targets = limit_batch_objects(targets, self.max_total_objects_per_batch)
```
eugene123tw: I think it's worth noting a few trade-offs:

* Step 1 (cap at 300) isn't truly lossless — pre-filtering to the 300 largest GTs before matching changes the cost matrix. The optimal Hungarian assignment over all 500 GTs might match queries to smaller objects that get dropped, so this is a heuristic approximation.

* If the root cause is the matcher flattening (B×Q) × T — we should considier computing the cost matrix per-image instead of across the batch, this would avoid the OOM without touching GT data?

* As we are always keeping largest boxes biases against small objects — on dense datasets, the model never sees small objects during training. Have you tried random sampling or epoch-level rotation, and see if it would give more balanced coverage?

Could we document this trade-off (e.g., in the module docstring or a note in the recipe)? It would also be worth validating the strategy by benchmarking RF-DETR against other models on a dense dataset like VisDrone.

library/src/otx/backend/native/callbacks/__init__.py
```diff
@@ -4,5 +4,6 @@
 """Module for OTX custom callbacks."""
```
eugene123tw, line 4: nitpick: could you change license header to:

```
# Copyright (C) 2023-2026 Intel Corporation
```

library/src/otx/backend/native/callbacks/gpu_mem_monitor.py
```diff
@@ -7,14 +7,18 @@
```
eugene123tw, line 7: apply the same to license header:
```
# Copyright (C) 2023-2026 Intel Corporation
```

library/src/otx/backend/native/models/detection/backbones/hgnetv2.py
```diff
@@ -481,9 +481,9 @@ def __init__(
         self.use_lab = use_lab
         self.return_idx = return_idx
```
eugene123tw, line 483: same here: 

```
# Copyright (C) 2025-2026 Intel Corporation
```

library/src/otx/backend/native/models/detection/detectors/detection_transformer.py
```diff
@@ -14,6 +14,7 @@
 from torchvision.tv_tensors import BoundingBoxes
 
 from otx.backend.native.models.detection.losses import DetrCriterion
+from otx.backend.native.models.detection.utils.utils import generate_scales
```
eugene123tw:
```suggestion
from otx.backend.native.models.detection.utils import generate_scales
```

library/src/otx/metrics/dice.py
```diff
@@ -8,8 +8,8 @@
 from typing import TYPE_CHECKING, Literal
 
 from torchmetrics import JaccardIndex
-from torchmetrics.classification.dice import Dice
 from torchmetrics.collections import MetricCollection
+from torchmetrics.segmentation.dice import DiceScore
```
eugene123tw, line 12: As we changed the metric, would you also share benchmark results on semantic segmentation?

library/src/otx/backend/native/models/instance_segmentation/rfdetr_inst.py
```diff
@@ -0,0 +1,437 @@
...
+    from pathlib import Path
+
+    from lightning.pytorch.cli import LRSchedulerCallable, OptimizerCallable
+
+    from otx.backend.native.schedulers import LRSchedulerListCallable
+    from otx.metrics import MetricCallable
+    from otx.types.label import LabelInfoTypes
+
+
+class RFDETRInst(OTXInstanceSegModel):
```
eugene123tw: **Nitpick: significant code duplication between `RFDETR` and `RFDETRInst`**

`configure_optimizers()`, `forward_for_tracing()`, `export()`, `_optimization_config`, and most of `_create_model()` are near-identical copies across rfdetr.py and rfdetr_inst.py (~100 lines). Consider extracting the shared logic into a mixin, e.g.:

```python
class RFDETRMixin:
    """Shared logic for RF-DETR detection and instance segmentation models."""

    def configure_optimizers(self):
        ...

    def forward_for_tracing(self, inputs):
        return self.model.export(inputs)

    def export(self, output_dir, base_name, export_format, precision=OTXPrecisionType.FP32):
        self.model.lwdetr.export()
        if self.explain_mode:
            raise ValueError("Explain mode is not supported for RF-DETR model.")
        return super().export(output_dir, base_name, export_format, precision)

    @property
    def _optimization_config(self):
        return {"model_type": "transformer"}
```

The task-specific classes would then only override `_customize_inputs`, `_customize_outputs`, `_exporter`, and `_default_preprocessing_params`.

Not a blocker, but would reduce maintenance burden — any future fix to `configure_optimizers` or `export` would need to be applied in two places otherwise.

eugene123tw (commented):
Hi Kirill, thanks for this great PR! Really nice to have RF-DETR in OTX!

A few notes from my side:

1. **License headers** — Could you verify that all modified files have updated copyright years? Existing files should use `# Copyright (C) 20xx-2026 Intel Corporation` and new files should use `# Copyright (C) 2026 Intel Corporation`.

2. **`limit_batch_objects` trade-offs** — Nice workaround for the OOM issue. I'm wondering if alternative approaches were considered (e.g., computing the cost matrix per-image in a loop, or chunking the matcher). If those aren't practical, could you document the accuracy trade-off somewhere (e.g., module docstring or recipe notes)? It would also be great to share some numbers showing the effectiveness of batch object limiting — before/after memory usage and any mAP impact on a dense dataset like VisDrone.

3. **`lwdetr.export()` mutates the model permanently** — There's no rollback if export fails partway through. Not a blocker, but it would be cleaner to wrap it with restore logic (e.g., a context manager or try/finally) so the original forward gets restored after export.

4. **`torchmetrics` upgrade (1.6.0 → 1.8.2)** — The new `DiceScore` API has backward-incompatible changes. Could you verify we're still getting comparable Dice scores after the upgrade?

5. **Code duplication between `RFDETR` and `RFDETRInst`** — `configure_optimizers()`, `forward_for_tracing()`, `export()`, and `_optimization_config` are near-identical copies (~100 lines). Not a blocker for this PR, but worth extracting into a shared mixin in a follow-up.

library/src/otx/backend/native/models/detection/rfdetr.py
kprokofi: We should avoid using fork. I added it because source implementation doesn't support export for torch > 2.10 and doesn't support dynamo export at all. Also, they have conflicts in torch version. I'm considering to create a PR to roboflow repo and remove fork in the future. 
Anyway, thanks for suggesting that, I will implement your proposal

library/src/otx/backend/native/models/detection/rfdetr.py
kprokofi: I didn’t want to spend much on this heuristic. It helps mitigate OOM, but I’d consider it the last step.

The main problem isn’t in matcher. Matcher is the same as for D-Fine, simple Hungarian algorithm, 1-to-1 matching. The problem is in the decoder architecture. They use groups DETR, 13x, which becomes very sensitive to the number of objects.

For Geti-tune users, I suggest:
1) Lower batch and gradient checkpointing.
2) Lower the number of groups in the decoder (accuracy degradation? Probably yes, I didn’t experiment with that).
3) Cap objects. I don’t know the performance without this hack. I couldn't manage to train this model on 1 GPU with normal batch size on Visdrone dataset without targets cleaning. It probably hurts training in some degree, but I generally achieved good mAP metric compared to other DETR models.

Ablation study is required here, I agree. Unfortunately, I don't have much time right now for additional experiments. In this first iteration, I wouldn't expose this parameter for the user in UI, only explanation and warning, that this model is not suitable for very dense datasets. However, I would like to keep this functionality for further exploration and probably some enhancements.

kprokofi: Please, see the benchmark above vs DEIMV2 model, it includes Visdrone dataset.

library/src/otx/metrics/dice.py
kprokofi: I will check on few datasets. Thanks

library/src/otx/backend/native/models/instance_segmentation/rfdetr_inst.py
kprokofi: You are right. I will create a mixin class in common folder

library/src/otx/backend/native/models/detection/rfdetr.py
kprokofi: I don't think so. I will remove that

library/src/otx/backend/native/models/instance_segmentation/rfdetr_inst.py
kprokofi: Done

library/src/otx/metrics/dice.py
kprokofi: I checked. Metric is equal. However, I don't include background class now to match Geti expectations, so lower numbers are expected.

library/src/otx/backend/native/models/detection/rfdetr.py
kprokofi: Done

library/src/otx/backend/native/models/detection/rfdetr.py
kprokofi: Removed

eugene123tw (dismissed):
Thanks Kirill!

library/uv.lock
```diff
@@ -1369,62 +1358,32 @@ wheels = [
...
+sdist = { url = "https://files.pythonhosted.org/packages/7c/b7/8cb61d2eece5fb05a83271da168186721c450eb74e3c31f7ef3169fa475b/huggingface_hub-0.36.2.tar.gz", hash = "sha256:1934304d2fb224f8afa3b87007d58501acfda9215b334eed53072dd5e815ff7a", size = 649782, upload-time = "2026-02-06T09:24:13.098Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/d5/ae/2f6d96b4e6c5478d87d606a1934b5d436c4a2bce6bb7c6fdece891c128e3/huggingface_hub-1.4.1-py3-none-any.whl", hash = "sha256:9931d075fb7a79af5abc487106414ec5fba2c0ae86104c0c62fd6cae38873d18", size = 553326, upload-time = "2026-02-06T09:20:00.728Z" },
+    { url = "https://files.pythonhosted.org/packages/a8/af/48ac8483240de756d2438c380746e7130d1c6f75802ef22f3c6d49982787/huggingface_hub-0.36.2-py3-none-any.whl", hash = "sha256:48f0c8eac16145dfce371e9d2d7772854a4f591bcb56c9cf548accf531d54270", size = 566395, upload-time = "2026-02-06T09:24:11.133Z" },
 ]
 
 [[package]]
 name = "idna"
-version = "3.11"
+version = "2.10"
```
leoll2: Not sure why but this package is being downgraded, resulting in a vulnerability alert:

> library/uv.lock » idna@2.10 – Internationalized Domain Names in Applications (IDNA) vulnerable to denial of service from specially crafted inputs to idna.encode (moderate severity)

https://github.com/open-edge-platform/training_extensions/actions/runs/22174824340/job/64120737186?pr=5081

library/uv.lock
```diff
@@ -1369,62 +1358,32 @@ wheels = [
...
-]
-sdist = { url = "https://files.pythonhosted.org/packages/b1/df/48c586a5fe32a0f01324ee087459e112ebb7224f646c0b5023f5e79e9956/httpx-0.28.1.tar.gz", hash = "sha256:75e98c5f16b0f35b567856f597f06ff2270a374470a5c2392242528e3e3e42fc", size = 141406, upload-time = "2024-12-06T15:37:23.222Z" }
-wheels = [
-    { url = "https://files.pythonhosted.org/packages/2a/39/e50c7c3a983047577ee07d2a9e53faf5a69493943ec3f6a384bdc792deb2/httpx-0.28.1-py3-none-any.whl", hash = "sha256:d909fcccc110f8c7faf814ca82a9a4d816bc5a6dbfea25d6591d6985b8ba59ad", size = 73517, upload-time = "2024-12-06T15:37:21.509Z" },
-]
-
 [[package]]
 name = "huggingface-hub"
-version = "1.4.1"
+version = "0.36.2"
```
leoll2, line 1446: Another unexpected major version downgrade here

library/uv.lock
```diff
@@ -3861,16 +4076,25 @@ wheels = [
...
+source = { registry = "https://pypi.org/simple" }
+sdist = { url = "https://files.pythonhosted.org/packages/2f/fa/1d6a26185653e3b65f2ddcee7f716a9d41f79a39ffb0e9dd9bf0ce43e587/pydeprecate-0.4.0.tar.gz", hash = "sha256:7c07295ee6faf427686a458ddc3406014124ac0fb6ad8f90e50eb9d21adb86f3", size = 27674, upload-time = "2025-12-03T10:40:10.066Z" }
+wheels = [
+    { url = "https://files.pythonhosted.org/packages/20/e4/d618f8eec6bfbab64e2af8a5359a423470a2cad60ef8ea60e8a60be28517/pydeprecate-0.4.0-py3-none-any.whl", hash = "sha256:5f3a07fddb03829bb934163076853ce6ead460b2ec2ba7e0fdaad3e4fa93c186", size = 21798, upload-time = "2025-12-03T10:40:08.69Z" },
+]
+
 [[package]]
 name = "pydot"
-version = "3.0.4"
+version = "1.4.2"
```
leoll2: Another unexpected major version downgrade here

library/uv.lock
kprokofi: I regenerated uv.lock and it changed many of the dependencies. I will fix that

library/uv.lock
kprokofi: There is no solution to resolve this. 
Lower version of tokenizers cannot be built, newer version cannot work with huggingface-hub>1.0
transformers also have conflicts.

library/uv.lock
kprokofi: Fixed

library/uv.lock
kprokofi: Fixed

library/uv.lock
```diff
@@ -3006,6 +2965,24 @@ dependencies = [
     { name = "typing-extensions" },
 ]
 
+[[package]]
+name = "opencv-python"
```
leoll2: One last request, could you suppress this dependency, introduced by `supervision`? We already have `opencv-python-headless`, so we don't need two OpenCV packages.

I think you can do it `pyproject.toml` as follows:
```toml
[tool.uv]
override-dependencies = [
    "opencv-python; sys_platform == 'never'",
]
```

kprokofi: Should be ok

leoll2 (approved):

PR #5088, opened 2025-12-18
Fix model is not reloaded when pipeline status changes

### LINKED ISSUE #5083: Model is not reloaded when pipeline status changes
**Steps to reproduce**

(assuming default projects after seeding)

1. Start the server
2. Call the `pipeline:disable` endpoint
3. Stop the server
4. Restart the server
5. Call the `pipeline:enable` endpoint
6. Monitor the logs and output

**Expected**

After step 6, the inference worker logs should report that a model got loaded. Predictions should be visible after a few seconds in the `data/output` folder.

**Actual**

After enabling the pipeline, no model gets reloaded and no predictions are generated.

**Root cause**

The `mp_model_reload_event` condition is never notified, so the inference workers cannot react to the pipeline updates. The notification logic probably got lost during some refactoring.

itallix:
### Summary

- [x] Extracted app-agnostic event bus into a base class
- [x] Added reload model event to app-specific event bus implementation 

### How to test

Follow the steps from #5083 and ensure expected behavior is in place.

### Checklist

### DIFF
--- a/application/backend/app/workers/inference.py
+++ b/application/backend/app/workers/inference.py
@@ -25,7 +25,7 @@ class InferenceWorkerConfig:
     frame_queue: mp.Queue
     pred_queue: mp.Queue
     stop_event: EventClass
-    model_reload_event: EventClass
+    model_reload_event: EventClass | None
     shm_name: str
     shm_lock: Lock
     logger_: LoguruLogger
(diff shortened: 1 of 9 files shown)


application/backend/app/workers/inference.py
```diff
@@ -25,7 +25,7 @@ class InferenceWorkerConfig:
     frame_queue: mp.Queue
     pred_queue: mp.Queue
     stop_event: EventClass
-    model_reload_event: EventClass
+    model_reload_event: EventClass | None
```
leoll2, line 28: Why `None`?

itallix: This aligns with the `event_bus.py` property type: `Event | None` (Initial `EventBus` design has conditions as optional, so the event follows that approach)

leoll2 (approved):

PR #5095, opened 2025-12-22
Improve annotator Part 1

jpggvilaca:
### Summary

* Add label selection logic
* Make SAM tool consistent with Geti Prompt

Next PR: Add label UI's

### Checklist


application/ui/src/features/annotator/tools/segment-anything-tool/segment-anything-tool.component.tsx
```diff
@@ -1,87 +1,111 @@
...
-        if (!ref.current) {
-            return;
-        }
+    const handleAddAnnotations = (shapes: Shape[], label: Label) => {
+        addAnnotations(shapes, [label]);
+        setPreviewShapes([]);
+    };
 
-        const point = clampPoint(getRelativePoint(ref.current, { x: event.clientX, y: event.clientY }, zoom.scale));
+    // const handleAddAnnotationsCreateLabel = (label: Label) => {
```
MarkRedeman: I think for Tune we should not make SAM be responsible for creating new labels. I imagine that for Tune it will be more expensive to add labels (users would need to retrain their models before inference will return these labels) so having a separate more dedicated UI would feel more appropriate to me.

If we do want an easy way to add new labels from the annotator screen then I'd prefer us to think of a design that would allow us to reuse the same design for the other tools (bounding box, polygon etc) as otherwise that could get inconsistent.

application/ui/src/features/annotator/tools/segment-anything-tool/segment-anything-tool.component.tsx
```diff
@@ -90,61 +114,70 @@ export const SegmentAnythingTool = () => {
...
+                onPointerLeave={() => {
+                    setPreviewShapes([]);
+                }}
+                style={{
+                    cursor: `url("/icons/selection.svg") 8 8, auto`,
+                }}
+            >
+                <PreviewAnnotations previewAnnotations={previewAnnotations} image={image} />
+            </SvgToolCanvas>
+            {/* <CreateLabelPopover
```
MarkRedeman: Let's remove the commented code for now

application/ui/src-tauri/capabilities/default.json
MarkRedeman, line 1: In a separate PR can you check if these files can be included in the CI / prettier config?
I imagine that if we don't include them then these files will be updated every time someone makes manual changes to them based on their editor settings.

application/ui/src/features/annotator/tools/segment-anything-tool/segment-anything-tool.component.tsx
jpggvilaca: That is a valid point! I will rethink the architecture and see if it's possible to adjust without touching the UX

application/ui/src/features/annotator/tools/segment-anything-tool/segment-anything-tool.component.tsx
jpggvilaca: Done

application/ui/src-tauri/capabilities/default.json
jpggvilaca: you mean including them on .prettierignore? This change was caught by running `prettier` so it's included, not ignored

application/ui/src/features/annotator/tools/segment-anything-tool/segment-anything-tool.component.tsx
```diff
@@ -1,87 +1,101 @@
...
+
+    return (
+        <MaskAnnotations isEnabled annotations={previewAnnotations} width={image.width} height={image.height}>
+            {previewAnnotations.map((annotation) => (
+                <g
+                    key={annotation.id}
+                    aria-label='Segment anything preview'
+                    style={
+                        {
+                            '--energy-blue-shade': '#0095ca',
```
camiloHimura: we can add a new CSS var, wdyt?

camiloHimura (dismissed):

application/ui/src/features/annotator/tools/segment-anything-tool/segment-anything-tool.component.tsx
jpggvilaca: Agree. Added to index.css!

camiloHimura (dismissed):

camiloHimura (approved):

PR #5099, opened 2025-12-29
Add undo redo logic

jpggvilaca:
### Summary

Add undo redo to annotator

You can ignore the first commit as it's just copy paste of undo/redo files.

### Checklist

### DIFF
--- a/application/ui/src/features/dataset/media-preview/media-preview.component.tsx
+++ b/application/ui/src/features/dataset/media-preview/media-preview.component.tsx
@@ -79,7 +79,6 @@ export const MediaPreview = ({ mediaItem, close, onSelectedMediaItem }: MediaPre
                 >
                     <AnnotationActionsProvider
                         mediaItem={mediaItem}
-                        key={mediaItem.id}
                         initialAnnotationsDTO={annotationsData?.annotations ?? []}
                         isUserReviewed={annotationsData?.user_reviewed ?? false}
                     >
--- a/application/ui/src/features/dataset/media-preview/media-preview.component.tsx
+++ b/application/ui/src/features/dataset/media-preview/media-preview.component.tsx
@@ -80,8 +80,8 @@ export const MediaPreview = ({ mediaItem, close, onSelectedMediaItem }: MediaPre
                     <AnnotationActionsProvider
                         mediaItem={mediaItem}
                         key={mediaItem.id}
-                        initialAnnotationsDTO={annotationsData?.annotations}
-                        isUserReviewed={annotationsData?.user_reviewed}
+                        initialAnnotationsDTO={annotationsData?.annotations ?? []}
+                        isUserReviewed={annotationsData?.user_reviewed ?? false}
                     >
                         <ZoomProvider>
                             <Suspense fallback={<CanvasAreaLoading />}>
--- a/application/ui/src/features/dataset/media-preview/media-preview.component.tsx
+++ b/application/ui/src/features/dataset/media-preview/media-preview.component.tsx
@@ -4,7 +4,10 @@
 import { Suspense } from 'react';
 
 import { Content, Dialog, dimensionValue, Divider, Flex, Grid, Heading, Loading, View } from '@geti/ui';
+import { useProjectIdentifier } from 'hooks/use-project-identifier.hook';
+import { isObject } from 'lodash-es';
 
+import { $api } from '../../../api/client';
 import { ZoomProvider } from '../../../components/zoom/zoom.provider';
 import type { DatasetItem } from '../../../constants/shared-types';
 import { AnnotationActionsProvider } from '../../../shared/annotator/annotation-actions-provider.component';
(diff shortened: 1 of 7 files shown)


application/ui/src/shared/annotator/annotation-actions-provider.component.tsx
```diff
@@ -55,104 +56,78 @@ const AnnotationsContext = createContext<AnnotationsContextValue | null>(null);
...
+        return mapServerAnnotationsToLocal(initialAnnotationsDTO, projectLabels);
+    }, [initialAnnotationsDTO, project?.task?.labels]);
+
+    const [annotations, setAnnotations, undoRedoActions] = useUndoRedoState<Annotation[]>([]);
+
+    useEffect(() => {
+        if (localFormattedAnnotations.length > 0) {
+            undoRedoActions.reset(localFormattedAnnotations);
+        }
+    }, [localFormattedAnnotations, undoRedoActions]);
```
MarkRedeman: Do we really need this use effect? I.e. wouldn't the state be reset automatically due to the `key={mediaItem.id}` that is set on the `AnnotationActionsProvider`?

Alternatively I'd move the memo stuff inside of the useEffect. Another thing that might help is to extract this into a dedicated hook (I found `localFormattedAnnotatiosn` to be a rather weirdly named variable which made me confused :D )

Maybe something like this,

```typescript
const useResetAnnotations(initialAnnotations: AnnotationsDTO, projectLabels, undoRedoActions) = () => {
    useEffect(() => {
        const annotations = mapServerAnnotationsToLocal(initialAnnotations, projectLabels ?? []);

        if (annotations.length > 0) {
            undoRedoActions.reset(annotations);
        }
    }, [initialAnnotationsDTO, project?.task?.labels, undoRedoActions]);
}
```

and use like

```suggestion
    const [annotations, setAnnotations, undoRedoActions] = useUndoRedoState<Annotation[]>([]);
    useResetAnnotations(initialAnnotationsDTO, project?.task?.labels, undoRedoActions)
```

MarkRedeman (dismissed):

application/ui/src/shared/annotator/annotation-actions-provider.component.tsx
jpggvilaca: i always try to avoid using useEffect but in this case it's necessary or else even with the media preview changing, once the server annotations are loaded, they are not displayed (render doesnt happen). So here we do need the useEffect, i suspect because on how the useUndoRedo works. In any case, this can be solved in another way, by adding a Suspense and using suspenseQuery instead of Query., then we dont need to react/refresh or force anything. But i want to stay away from refactors to stay within he content of this PR.

Updated: Moved var from useMemo inside useffect

application/ui/src/features/dataset/media-preview/primary-toolbar/undo-redo/use-undo-redo-state.test.tsx
```diff
@@ -0,0 +1,353 @@
...
+
+    it("Won't fail when calling undo multiple times", (): void => {
+        render(<MockAppMultipleCalls method={'undo'} />);
+
+        const undoButton = screen.getByRole('button', { name: 'undo' });
+        fireEvent.click(undoButton);
+        expect(screen.getByText('0')).toBeInTheDocument();
+    });
+
+    it("Won't fail when calling redo multiple times", (): void => {
```
ActiveChooN, line 257: Should the test call redo twice?

ActiveChooN (approved):
LGTM. Just small question

application/ui/src/features/dataset/media-preview/media-preview.component.tsx
```diff
@@ -55,7 +77,12 @@ export const MediaPreview = ({ mediaItem, close, onSelectedMediaItem }: MediaPre
                     }}
                     areas={['toolbar header aside', 'toolbar canvas aside', 'toolbar footer aside']}
                 >
-                    <AnnotationActionsProvider mediaItem={mediaItem}>
+                    <AnnotationActionsProvider
+                        mediaItem={mediaItem}
+                        key={mediaItem.id}
```
camiloHimura: Can you clarify why this approach is being used for this component?

jpggvilaca: It was from the previous implementation. We had a useEffect that relied on the mediaItem id, and instead of using a useEffect, we can use a key to get the component to rerender. But in this case, the mediaItem is already taking care of that so we actually dont need this anymore. Thank you, removed!

application/ui/src/features/dataset/media-preview/primary-toolbar/undo-redo/use-undo-redo-state.test.tsx
jpggvilaca: It actually calls it thrice. You can check line 49. Pressing the button on "MockAppMultipleCalls" will trigger the undo (or redo) 3 times.

camiloHimura (approved):

PR #5109, opened 2025-12-30
Reliable WebRTC connection in restrictive networks

### LINKED ISSUE #5080: Reliable WebRTC connection in restrictive networks
The goal of this ticket is to make the WebRTC stream (to display predictions) work reliably under different network topologies, even with strict firewall settings.

In line with the work done in Geti Prompt in https://github.com/open-edge-platform/geti-prompt/pull/483 and https://github.com/open-edge-platform/geti-prompt/pull/515, Geti Tune should also provide support for ICE server configuration and the (optional) setup of a TURN server for restrictive networks.

itallix:
### Summary

This PR integrate `Geti Prompt` app improvements to support different network topologies:
- [x] Add new profile with coturn server for traffic relay
- [x] Introduce `ICE_SERVERS` and `WEB_ADVERTISE_IP` env variables to support `STUN` and `TURN` protocols

### How to test

The links below point to GCE VMs running the Geti Tune Docker container with different ICE configurations.

**Debugging:**
- Open `chrome://webrtc-internals` (for chromium-based browsers only) in a separate tab to inspect ICE candidate gathering and connectivity checks.
- Look for the **selected candidate pair** under the active `RTCPeerConnection`.

**Expected ICE candidate types:**
- **STUN-only**: Selected pair should use `srflx` (server-reflexive) local candidate when NAT traversal succeeds. Connection may fail on restrictive networks.
- **TURN (relay-only)**: Selected pair should use `relay` local candidate. All media flows through the TURN server; works even on restrictive networks but adds latency.

**Test deployments:**
- [STUN-only](http://34.13.219.73) – [ICE config](http://34.13.219.73/api/webrtc/config)
- [TURN (relay-only)](http://34.12.70.108) – [ICE config](http://34.12.70.108/api/webrtc/config) ← verify `relay` candidate is selected in `webrtc-internals`

### Checklist


application/backend/app/api/routers/webrtc.py
```diff
@@ -42,3 +44,21 @@ async def create_webrtc_offer(offer: Offer, webrtc_manager: Annotated[WebRTCMana
...
     """Update webrtc input with user data"""
     webrtc_manager.set_input(data)
+
+
+@router.get(
+    path="/config",
+    response_model=WebRTCConfigResponse,
+    responses={
+        status.HTTP_200_OK: {"description": "WebRTC configuration"},
+        status.HTTP_500_INTERNAL_SERVER_ERROR: {"description": "Internal Server Error"},
```
leoll2: We don't have to document 5xx responses (server-side errors)

application/backend/app/api/routers/webrtc.py
```diff
@@ -42,3 +44,21 @@ async def create_webrtc_offer(offer: Offer, webrtc_manager: Annotated[WebRTCMana
...
+    },
+)
+async def get_webrtc_config(ice_servers: Annotated[list[dict], Depends(get_ice_servers)]) -> WebRTCConfigResponse:
+    """Get WebRTC configuration including ICE servers"""
+    try:
+        servers = [WebRTCIceServer(**server) for server in ice_servers]
+    except TypeError as e:
+        logger.exception("Error processing WebRTC configuration")
+        raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail=str(e))
```
leoll2: This `try..catch` block seems redundant. If L60 throws any exception, the server would return error 500 anyway, and I believe the logs would contain the traceback too.

application/backend/app/webrtc/sdp_handler.py
```diff
@@ -0,0 +1,82 @@
...
+    """Handler for SDP manipulation and processing."""
+
+    async def resolve_hostname(self, hostname: str) -> str:
+        """Resolve hostname to IP address.
+
+        Args:
+            hostname: The hostname to resolve.
+
+        Returns:
+            The resolved IP address or the original hostname if resolution fails.
```
leoll2, line 23: Why return the original hostname in case of failure, instead of raising an exception?

itallix: I guess the idea here is to avoid failing the whole operation when resolving the advertised IP. Only a subset of ICE candidates may contain a hostname. If hostname resolution fails, the ICE agent likely won’t be able to validate those candidates anyway, but other candidates may still be usable. So we shouldn’t fail the `offer/answer` cycle because of that.

cc @mpryahin

application/backend/app/api/routers/webrtc.py
itallix: Indeed

leoll2 (approved):

jpggvilaca (approved):
UI side LGTM!

PR #5122, opened 2026-01-05
Windows Installer

### LINKED ISSUE #4534: Bundle server and its dependencies into a single executable
Use [pyinstaller](https://pyinstaller.org/en/stable/) (or similar tool) to bundle the FastAPI server and its dependencies into a single, standalone executable package.

### LINKED ISSUE #4535: Bundle UI and server as a desktop application
This task is about bundling the UI and the server binary as a native desktop application.

If the chosen solution is based on Tauri, the following links may help:

- https://v2.tauri.app/develop/sidecar/
- https://github.com/dieharders/example-tauri-v2-python-server-sidecar
- https://github.com/maxim-saplin/flutter_python_starter/tree/main

### LINKED ISSUE #4536: Setup CI to bundle the cross-platform native app
Create a GHA workflow to build the desktop application for the target OS's (Windows, Ubuntu, ...).

warrkan:
This PR introduces a Github Action to build MSIX Windows package with Intel Geti Tune.


application/ui/src-tauri/Cargo.toml
```diff
@@ -1,8 +1,8 @@
 [package]
-name = "app"
+name = "geti_tune_ui"
 version = "0.1.0"
-description = "A Tauri App"
-authors = ["you"]
+description = "Geti Tune UI"
```
leoll2, line 4: Does it only include the UI component or the whole bundled application? If the latter, I suggest to rename to `geti_tune_desktop` (`Geti Tune Desktop Application`) to avoid ambiguity.

application/ui/src-tauri/msix/Assets/logo.png
leoll2, line 1: What's this file used for? The image looks completely black.

application/backend/app/main.py
```diff
@@ -1,15 +1,21 @@
 # Copyright (C) 2025 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 
+import sys
+
+if getattr(sys, "frozen", False) and __name__ == "__main__":
+    print("Calling multiprocessing.freeze_support()")
+    import multiprocessing
+
+    multiprocessing.freeze_support()
```
leoll2, line 20: This block of code should be after L16. Actually, the commented instructions in L12-L16 can be removed.

leoll2, line 20: Could you a comment to explain the purpose of this block code?

application/backend/app/main.py
```diff
@@ -50,11 +69,18 @@
 )
 
 # Include all API routers from the routers package
-for router_info in pkgutil.iter_modules(routers.__path__):
-    router_name = router_info.name
-    module = importlib.import_module(f"app.api.routers.{router_name}")
-    if hasattr(module, "router"):
-        app.include_router(module.router)
```
leoll2, line 57: What is the problem with the old approach?

application/backend/app/settings.py
```diff
@@ -87,6 +89,14 @@ def ensure_dirs_exist(self) -> None:
             if d:
                 d.mkdir(parents=True, exist_ok=True)
 
+    @field_validator("static_files_dir", "alembic_config_path", "alembic_script_location", mode="after")
+    def prefix_paths(cls, v: str | Path | None) -> str | Path | None:
+        if v and getattr(sys, "frozen", False) and hasattr(sys, "_MEIPASS"):
+            # If application is running in pyinstaller bundle, adjust the path accordingly.
+            prefixed_path = os.path.join(getattr(sys, "_MEIPASS", ""), v)
+            return Path(prefixed_path) if isinstance(v, Path) else prefixed_path
+        return v
```
leoll2, line 101: Can you add a more detailed comment for this block of code? Why does pyinstaller need a different path and how do you determine it?

application/backend/app/main.py
warrkan: Pyinstaller has issues with detecting modules dynamically, so more predictable way will be to import and declare routers explicitly

application/backend/app/settings.py
warrkan: I believe this page can explain it better 😄 https://pyinstaller.org/en/stable/runtime-information.html but I'll add a comment

application/ui/src-tauri/Cargo.toml
warrkan: This is a Cargo file to build a Rust UI application from web sources. So it's UI only

application/ui/src-tauri/msix/Assets/logo.png
warrkan: That's done on purpose until we get proper assets to include - logos, icons, etc

leoll2 (approved):

jpggvilaca (approved):

AlexanderBarabanov (approved):

PR #5126, opened 2026-01-06
Design for dataset I/E

### LINKED ISSUE #4988: Create new project reusing existing dataset
Users often want to experiment with different tasks and labels, while also reusing previously created datasets. To facilitate this process, we want to enable the creation of new projects from existing datasets.

In Geti Classic, a similar usecase was addressed through dataset import/export: users had to export their datasets to Datumaro format, and later reimport them to create a new project. In Geti Tune, we want to avoid the intermediate I/E steps, all data should stay inside the application during this operation.

## Requirements

- The user should have an option to create a new project initialized with the dataset of another project
  - The dataset should be a _copy_, so that operations on one project don't affect the other (cross-project isolation)
  - The new project may have a different task type - see below for the list of combinations to support. Annotations must be converted to the target type, as described below.
  - The new project may have different labels than the original one. The user should have the option to map each source label to a target label, or none (in this case, annotations with that label are discarded). It is allowed to assign two source labels to the same target label.

### Supported task conversions

| From | To | How to convert shapes |
| - | - | - |
| Instance segmentation | Object detection | Create the smallest bounding box enclosing the polygon |
| Instance segmentation | Multilabel classification | Apply labels of all polygons in the image |
| Object detection | Instance segmentation | Reinterpret bbox as polygon |
| Object detection | Multilabel classification | Apply labels of all polygons in the image |

### LINKED ISSUE #5113: Write design for dataset I/E

leoll2:
### Summary

This PR proposes a design for the import/export features (#5113), as well as for the creation of new projects reusing existing datasets (#4988)

Resolves #5113
Resolves #4988

### Checklist


application/docs/dataset-ie.md
```diff
@@ -0,0 +1,443 @@
...
+To accommodate such variety of workflows, the design seeks to break down the aforementioned use cases into a set of
+complementary and orthogonal operations. This approach also has the advantage of reducing code duplication, with obvious
+benefits in terms of maintainability and reliability.
+
+### Assumptions and Constraints
+
+- Datasets can be arbitrarily large, and most of the operations scale linearly with dataset size.
+  - Example of operations: uploading, downloading, archiving, extracting, parsing, transforming.
+- Metadata about the dataset (e.g., type of annotations, labels, number of items) can be obtained without fully loading
+the dataset into memory, provided that the dataset is stored in Datumaro format.
```
jpggvilaca: I dont have all the knowledge but i believe this is not the only benefit of dealing with Datumaro format. Should we, across the app, push for this? Visually speaking, adding a "Recommended" tag everywhere?

application/docs/dataset-ie.md
```diff
@@ -0,0 +1,443 @@
...
+- Datasets may be imported to projects of different task types and labels. If so, annotations need to be adapted
+accordingly (see [Task compatibility](task-compatibility.md) for more details) and the labels must explicitly re-mapped.
```
jpggvilaca: This was definitely a hard scenario to tackle on the previous Geti. What if, on the first iteration, we make it easy for everyone and just tell the user that either A) The annotations are all gone or B) The chosen task must be X, where X is the appropriate one for that dataset that he/she imported. WDYT? We can iterate later and make the system smarter. But to start, i think it would help tremendously if we don't have to tackle this.

application/docs/api.md
```diff
@@ -115,9 +115,21 @@
...
+| `GET`  | `/api/jobs/<id>`        | -                   | job info     | Get info about a specific job                      |
+| `POST` | `/api/jobs/<id>:cancel` | -                   | -            | Cancel a job                                       |
+| `GET`  | `/api/jobs/<id>/status` | -                   | job status   | Stream real-time status updates for a specific job |
+| `GET`  | `/api/jobs/<id>/logs`   | -                   | job logs     | Stream real-time log output for a specific job     |
+
+Job types:
+
+- `train`
+- `prepare_dataset_for_import`
+- `import_dataset_to_existing_project`
```
jpggvilaca, line 131: Would it be possible to have only 1 endpoint for this and move the action choice to the backend?  This would make it simpler for at least the client (UI or user using the API). Can I not, as a user, just call the same endpoint, and then the backend figures out if the project is a new or an existing one? (we could even pass a flag if necessary, but that defeats the purpose a little bit)

application/docs/dataset-ie.md
```diff
@@ -0,0 +1,443 @@
...
+The first interesting thing to note is that all use cases involve moving a dataset from a source to a destination.
+The source could be an external zip archive, a project's dataset within Geti Tune, or a dataset revision, while
+the destination could be a zip file or the dataset of a new or existing project.
+Between the source and destination, several intermediate steps may need to take place, depending on the specific use
+case: format conversion, compression / decompression, filtering, label remapping.
+Through careful interface modeling, it is possible to get all sources to work with all destinations, then realize all
+use cases by composing the necessary intermediate steps in a set of endpoints and jobs.
+
+Two important design choices follow from these observations:
+1. All datasets that are uploaded or downloaded by the user, regardless of how, must be zip archives in a standard
```
jpggvilaca:
```suggestion
1. All datasets that are uploaded or downloaded by the user, regardless of the format, must be zip archives in a standard
```
?

application/docs/dataset-ie.md
```diff
@@ -0,0 +1,443 @@
...
+involved in dataset import and export.
+
+- **prepare_dataset_for_import**: extracts a compressed dataset archive, saving it to Datumaro format while determining
+    its characteristics (task type, labels, number of media and annotations).
+- **import_dataset_to_existing_project**: loads a Datumaro dataset and inserts its items into the dataset of a project,
+    applying optional filtering, annotation conversion and label remapping.
+    If the target is a new project, this job also creates the project.
+- **import_dataset_as_new_project**: creates a new project, then loads a Datumaro dataset and inserts its items into
+    the dataset of the newly created project, applying optional filtering and annotation conversion.
+- **stage_dataset**: loads a project's dataset (or dataset revision) as a Datumaro dataset, applying optional filtering,
```
jpggvilaca, line 211: Why is staging, an intermediary internal step, be exposed? What is the use case for the user to do this instead of just importing/exporting?

application/docs/dataset-ie.md
AlbertvanHouten: This will be much easier with the new datumaro implementation. It will take care of e.g., converting from polygons to bounding boxes. Ofcourse, not all scenarios make sense to convert as there is missing information if you were to go from classification to a segmentation task. I guess this is defined in the task-compatibility file (@leoll2 the link to that file isn't working, should it still be added?)

application/docs/dataset-ie.md
```diff
@@ -0,0 +1,443 @@
...
+In the above example, the staging area contains two datasets being processed, each uniquely identified by a UUID:
+- The first one is a zip archive in COCO format. A dataset in this form (compressed) is normally the result of an upload
+  operation (for later import) or an export operation (dataset ready for download).
+- The second one the other is a Datumaro dataset in uncompressed form. This form is ideal for internal processing, as
+  Datumaro datasets can be efficiently filtered and transformed thanks to its internal polars-based representation.
+
+Notably, the staging area exists purely in the file system; there is no corresponding representation in the database.
+This is essential to allow users to access the staging area directly, without going through the application API.
+In other words, the users can put datasets directly in the staging area for import operations, or take them directly
+from there after an export operation, without needing to go through the API (→ simple and efficient upload / download).
```
AlbertvanHouten, line 87: How long will files remain in the staging area for? I guess they would need to be cleaned up at some point.

application/docs/dataset-ie.md
```diff
@@ -0,0 +1,443 @@
...
+
+#### Dataset Export
+
+<img src="media/dataset-export-data-flow.jpg" alt="Dataset export data flow" width="900" />
+
+<img src="media/dataset-export-detailed-flow.jpg" alt="Dataset export detailed flow with filtering" width="900" />
+
+1. Upon user request, the dataset (or dataset revision) to be exported is loaded from the database as a `dm.Dataset` 
+   object and filtered according to request (e.g., only include images with selected labels). These operations are
+   executed in a background job, since they may take a long time depending on dataset size.
```
AlbertvanHouten: I think you should write a section explaining how to deal with concurrent operations on the dataset when it is being exported. What will happen if an image is deleted while a background export is running for example?

application/docs/dataset-ie.md
```diff
@@ -0,0 +1,443 @@
...
+
+> [!WARNING]
+> The API described below is only for design purposes and it may not reflect the actual implementation.
+> Please refer to the OpenAPI specification for the definitive reference.
+
+### Staging Area Management
+
+#### Upload archive to staging area
+
+_Endpoint:_ `POST /api/staged_datasets`
```
AlbertvanHouten, line 238: Ideally this endpoint should have resumable upload support if compressed=true.

application/docs/dataset-ie.md
leoll2: Rephrased a little, by "how" I mean "the approach used to upload the dataset (api vs direct copy)"

application/docs/dataset-ie.md
leoll2: Correct, Datumaro format is very convenient for several reasons. In Geti Tune, for example, it was the only format that could seamlessly support all task types and videos natively. We can definitely add a "recommended" tag in the UI of dataset export.

application/docs/dataset-ie.md
leoll2: I added a note about TUS. In theory, resumable uploads would be the ideal approach to upload large files; in practice, such feature adds some complexity to the API which I'd rather avoid unless the users request it.

application/docs/dataset-ie.md
leoll2: It is used in the "dataset copy/fork" usecase. Users can basically take the dataset of an existing project and copy it into a new/existing project, optionally filtering and remapping labels.

application/docs/api.md
leoll2: If you mean combining `prepare_dataset_for_import` and `import_dataset_to_existing_project`, it is not possible because we need the user input after uploading and scanning the dataset, namely to decide which labels to include and how to map them.

If you mean combining `import_dataset_to_existing_project` and `import_dataset_as_new_project`, that's technically possible but the payload is quite different (see the API section) so two distinct endpoints are preferrable imho.

application/docs/dataset-ie.md
leoll2: As Albert said, this feature comes out of the box with the new Datumaro. It won't require much effort other than testing the various combinations (we only support 4, actually).

I fixed the link, it points to section later in this document.

application/docs/dataset-ie.md
leoll2: Added a section about it. My proposal is to use a simple file-based locking mechanism.

application/docs/dataset-ie.md
leoll2: Added a section about it. Staged datasets can be deleted manually, via API or automatically by a daemon thread after 24h.

application/docs/api.md
jpggvilaca: I meant the latter. If we can somehow find a union of all data we need that fits both, then we could merge both. But not a big deal in any case.

AlbertvanHouten (approved):

application/docs/api.md
leoll2: Yeah that was my initial idea before I realized that the payload would differ a lot:
- Import as new project requires project name, task type, labels to create, ...
- Import to existing project requires labels mapping, ...

while the job, internally, would need to perform significantly different validation checks and steps, depending on the destination. For these reasons, I think it's more maintainable to have separate jobs.

jpggvilaca (approved):

PR #5147, opened 2026-01-07
Refactoring: Remaining changes

### LINKED ISSUE #4980: Introduce service layer entities and extract business logic to services
In Geti Tune backend, refactor the code to achieve a proper separation of the following three tiers and entities:

- DB layer: the classes are stored in `app/db/`, they define the database schema
- Service layer: the models are stored in `app/models`, they are used in services which implement the business logic
- Presentation layer: the models are stored in `app/api/schemas`, they define the schema of the REST API

warrkan:
Moving remaining entities from `schemas` to `models`.
Resolves #4980


application/backend/app/api/routers/pipelines.py
```diff
@@ -12,9 +12,9 @@
 
 from app.api.dependencies import get_pipeline_metrics_service, get_pipeline_service, get_system_service
 from app.api.schemas import PipelineView
+from app.api.schemas.metrics import PipelineMetricsView
```
leoll2: nitpicky: this import can be merged with the previous (L14) if you export the classes in `schemas/__init__.py`

application/backend/pyproject.toml
```diff
@@ -155,7 +155,6 @@ name = "Schema-Model"
 type = "layers"
 id = "schema_model"
 layers = [
-    "app.schemas",
```
leoll2, line 158: If you remove this line, then this contract can be removed entirely. But I think it would be better to replace this line with `app.api.schemas`.

application/backend/app/api/schemas/metrics.py
```diff
@@ -0,0 +1,29 @@
+# Copyright (C) 2025 Intel Corporation
```
leoll2: Year 2026. Same in other newly added files.

leoll2 (commented):
Minor comments, the changes LGTM, thank you

application/backend/pyproject.toml
warrkan: `app.api.schemas` is already in another group, so I've removed this one

itallix (approved):

leoll2 (approved):

dwesolow (approved):

PR #5151, opened 2026-01-08
Use selected device for OTX training

### LINKED ISSUE #5136: Select the training device according to configuration
After #5115, the training job has an input parameter `device` to specify the training device.

To make this parameter actually work, the `OTXEngine` `device` parameter must be initialized accordingly:

https://github.com/open-edge-platform/training_extensions/blob/20ee77902e338c037e8a914f696fce483ba5a0fb/application/backend/app/services/training/otx_trainer.py#L324

Within this ticket, make sure the training runs on the selected device.

A-Artemis:
### Summary

OTX training uses the requested device for training

### Checklist

### DIFF
--- a/application/backend/app/services/training/otx_trainer.py
+++ b/application/backend/app/services/training/otx_trainer.py
@@ -321,7 +321,7 @@ class OTXTrainer(Trainer):
         otx_engine = OTXEngine(
             model=otx_model,
             data=otx_datamodule,
-            device=device.type,
+            device=f"{device.type}:{device.index if device.index else 0}",
             work_dir=f"./otx-workspace-{model_id}",
         )
 
--- a/application/backend/app/services/training/otx_trainer.py
+++ b/application/backend/app/services/training/otx_trainer.py
@@ -26,7 +26,6 @@ from otx.data.dataset.base import OTXDataset
 from otx.data.module import OTXDataModule
 from otx.data.transform_libs.torchvision import TorchVisionTransformLib
 from otx.tools.converter import GetiConfigConverter
-from otx.types.device import DeviceType
 from otx.types.export import OTXExportFormatType
 from otx.types.precision import OTXPrecisionType
 from otx.types.task import OTXTaskType
@@ -36,6 +35,7 @@ from app.core.jobs.models import TrainingJobParams
 from app.core.run import ExecutionContext
 from app.models import DatasetItemAnnotationStatus, Task, TaskType, TrainingStatus
 from app.models.training_configuration.configuration import TrainingConfiguration
+from app.schemas.system import DeviceInfo
 from app.services import (
     BaseWeightsService,
     DatasetRevisionService,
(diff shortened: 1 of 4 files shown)


application/backend/tests/unit/services/training/test_otx_trainer.py
```diff
@@ -200,8 +200,9 @@ def test_prepare_weights_with_parent_model_no_file_raises_error(
         otx_trainer = fxt_otx_trainer()
 
         # Act
-        with pytest.raises(FileNotFoundError, match=f"Parent model weights not found at {expected_weights_path}"):
+        with pytest.raises(FileNotFoundError) as excinfo:
             otx_trainer.prepare_weights(training_params)
+        assert excinfo.value.args[0] == f"Parent model weights not found at {expected_weights_path}"
```
A-Artemis, line 205: Required so that the test passes on Windows. 

The `path` it is checking has `C:\Users\...` which is evaluated as `C:\\Users\\...` as a `str` when on windows. This causes `\\U` to be interpreted as an escape character in the regex for the `match=...`.

application/backend/app/services/training/otx_trainer.py
```diff
@@ -321,7 +321,7 @@ def train_model(
         otx_engine = OTXEngine(
             model=otx_model,
             data=otx_datamodule,
-            device=DeviceType.cpu,  # TODO make device configurable (#5040)
+            device=f"{device.type}:{device.index if device.index else 0}",
```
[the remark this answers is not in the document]
A-Artemis: `cpu:0` is valid.

application/backend/tests/unit/services/training/test_otx_trainer.py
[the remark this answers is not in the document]
A-Artemis: Check above comment.

application/backend/app/services/training/otx_trainer.py
leoll2, line 324: The OTX parameter is annotated as a `DeviceType` enum, which doesn't seem to support indices. Is it just a matter of bad type-annotations in OTX? Can you provide a reference to the OTX code where the index is parsed and processed?

https://github.com/open-edge-platform/training_extensions/blob/094b05066a703b754254615e0729d6a5e4ec406e/library/src/otx/backend/native/engine.py#L92

kprokofi: Yes, with the current typing, we can’t configure a device by its ID.
Aurelien’s approach will work, but it requires adjusting the typing.
The simplest solution is to change the type to DeviceType | str, and make the same change in DeviceConfig. Additionally, we need to add validation to ensure that if a str is passed, it corresponds to a valid device.

leoll2: Good to know, thanks Kirill! If the code works as it is, let's proceed merging this PR then we can improve the type annotations later.

leoll2 (approved):

PR #5167, opened 2026-01-12
Export model to ONNX format

### LINKED ISSUE #5142: Export ONNX model during training
Extend the training job `export_model` function to also export as ONNX (`model.onnx`). Also update `store_model_artifacts` accordingly.

_Note: in the future, we want to create the ONNX model on-demand rather than by default_

A-Artemis:
### Summary

Adds logic to export the trained model to ONNX format with FP16 precision.

### Checklist

### DIFF
--- a/application/backend/app/services/training/otx_trainer.py
+++ b/application/backend/app/services/training/otx_trainer.py
@@ -370,11 +370,11 @@ class OTXTrainer(Trainer):
         )
         logger.info("Model exported to OpenVINO format: {}", exported_ov_model_path)
 
-        logger.info("Exporting the model to ONNX format (FP32 precision)...")
+        logger.info("Exporting the model to ONNX format (FP16 precision)...")
         exported_onnx_model_path = otx_engine.export(
             checkpoint=model_checkpoint_path,
             export_format=OTXExportFormatType.ONNX,
-            export_precision=OTXPrecisionType.FP32,
+            export_precision=OTXPrecisionType.FP16,
         )
         logger.info("Model exported to ONNX format: {}", exported_onnx_model_path)
         return ExportedModels(openvino_model_path=exported_ov_model_path, onnx_model_path=exported_onnx_model_path)
--- a/application/backend/app/services/training/otx_trainer.py
+++ b/application/backend/app/services/training/otx_trainer.py
@@ -358,17 +358,17 @@ class OTXTrainer(Trainer):
         self,
         otx_engine: OTXEngine,
         model_checkpoint_path: Path,
-        format: OTXExportFormatType,
+        export_format: OTXExportFormatType,
         precision: OTXPrecisionType,
     ) -> Path:
         """Export the trained model to desired format and precision"""
-        logger.info("Exporting the model to {} format with {} precision...", format, precision)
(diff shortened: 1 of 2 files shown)


application/backend/app/services/training/otx_trainer.py
```diff
@@ -429,13 +435,19 @@ def run(self, ctx: ExecutionContext) -> None:
...
         )
         self.evaluate_model(otx_engine=otx_engine, model_checkpoint_path=trained_model_path)
-        exported_model_path = self.export_model(otx_engine=otx_engine, model_checkpoint_path=trained_model_path)
-        self.store_model_artifacts(
-            otx_work_dir=Path(otx_engine.work_dir),
-            training_params=training_params,
-            trained_model_path=trained_model_path,
-            exported_model_path=exported_model_path,
-        )
+        for format in OTXExportFormatType:
```
A-Artemis: This `for` logic can be updated in the future to handle on-request export of ONNX format by using the available `ctx`.

application/backend/app/services/training/otx_trainer.py
```diff
@@ -429,13 +435,19 @@ def run(self, ctx: ExecutionContext) -> None:
...
+            exported_model_path = self.export_model(
```
leoll2: @itallix Is it okay to invoke the same job step twice, with different arguments? Does it have any side-effect on the progress reporting?

leoll2 (dismissed):

application/backend/app/services/training/otx_trainer.py
itallix: It should be reported twice in this case (duplicated logs). Maybe we should track the state (record triggered steps) to avoid that effect?

leoll2: I see. Then it's probably easier to have a single "Export model" step to export both ONNX and OpenVINO. The step can return an `ExportedModels` namedtuple/dataclass containing the paths of the ONNX and OpenVINO exported models.

application/backend/app/services/training/otx_trainer.py
```diff
@@ -353,25 +359,33 @@ def evaluate_model(self, otx_engine: OTXEngine, model_checkpoint_path: Path) ->
...
         )
-        logger.info(f"Model exported to: {exported_model_path}")
-        return exported_model_path
+        logger.info("Model exported to OpenVINO format: {}", exported_ov_model_path)
+
+        logger.info("Exporting the model to ONNX format (FP32 precision)...")
+        exported_onnx_model_path = otx_engine.export(
+            checkpoint=model_checkpoint_path,
+            export_format=OTXExportFormatType.ONNX,
+            export_precision=OTXPrecisionType.FP32,
```
leoll2: Does it not work with FP16?

A-Artemis: It does work, but for ONNX it defaults to FP32 so I decided to explicitly place it here to show the difference between OpenVINO and ONNX.

leoll2: If it works, then it is better to export everything at FP16. All modern hardware basically supports FP16, there is no reason to use FP32 (larger file, slower at inference).

leoll2 (approved):

PR #5175, opened 2026-01-13
Add training parameter settings

jpggvilaca:
### Summary

* Add training parameters settings and some layout fixes

Screenshot 2026-01-13 at 10 26 58

### Checklist

### DIFF
--- a/application/ui/src/features/models/model-listing/model-metrics/model-metrics.component.tsx
+++ b/application/ui/src/features/models/model-listing/model-metrics/model-metrics.component.tsx
@@ -47,8 +47,8 @@ export const ModelMetrics = () => {
     const jobDuration = '00:04:35';
 
     return (
-        <Flex direction='column'>
-            <Flex alignItems={'center'} marginBottom={'size-300'}>
+        <Flex direction='column' gap={'size-300'}>
+            <Flex alignItems={'center'}>
                 <Text UNSAFE_style={{ color: 'var(--spectrum-global-color-gray-900)' }}>
                     {`Model training time: ${trainingTime}`}
                 </Text>
@@ -60,7 +60,7 @@ export const ModelMetrics = () => {
 
             <ModelMetricsGraphs lossData={mockLossData} accuracyData={mockAccuracyData} />
 
-            <Link marginTop={'size-300'}>See more</Link>
+            <Link UNSAFE_style={{ textDecoration: 'none' }}>See more</Link>
         </Flex>
     );
 };
--- a/application/ui/src/features/models/model-listing/model-metrics/model-metrics.component.tsx
+++ b/application/ui/src/features/models/model-listing/model-metrics/model-metrics.component.tsx
@@ -48,7 +48,7 @@ export const ModelMetrics = () => {
 
     return (
         <Flex direction='column'>
-            <Flex alignItems={'center'} marginY={'size-300'}>
+            <Flex alignItems={'center'} marginBottom={'size-300'}>
                 <Text UNSAFE_style={{ color: 'var(--spectrum-global-color-gray-900)' }}>
                     {`Model training time: ${trainingTime}`}
                 </Text>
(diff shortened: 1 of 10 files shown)


application/ui/src/features/models/model-listing/model-metrics/model-metrics.component.tsx
```diff
@@ -48,7 +48,7 @@ export const ModelMetrics = () => {
 
     return (
         <Flex direction='column'>
-            <Flex alignItems={'center'} marginY={'size-300'}>
+            <Flex alignItems={'center'} marginBottom={'size-300'}>
```
dwesolow: Like in a previous PR, u can use `gap`.

application/ui/src/features/models/model-listing/model-training-parameters/model-training-parameters.component.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+import { ReactNode } from 'react';
+
+import { Content, Flex, Grid, Heading, Text } from '@geti/ui';
+
+import styles from './model-training-parameters.module.scss';
+
+const Box = ({ title, content }: { title: string; content: ReactNode }) => {
+    return (
+        <Flex direction={'column'} height={'100%'}>
```
dwesolow, line 12: I think we don't need `Flex` for that, `View` should be enough.

application/ui/src/features/models/model-listing/model-training-parameters/model-training-parameters.component.tsx
```diff
@@ -0,0 +1,94 @@
...
+                {title}
+            </Heading>
+            <Content UNSAFE_className={styles.boxContent}>{content}</Content>
+        </Flex>
+    );
+};
+
+export const ModelTrainingParameters = () => {
+    return (
+        <Grid columns={['1fr', '1fr', '1fr', '1fr']} gap={'size-200'}>
```
dwesolow, line 27: Probably we should not force to always have 4 columns, we should use `repeat` and `minmax` to make it more responsive.

application/ui/src/features/models/model-listing/model-training-parameters/model-training-parameters.component.tsx
```diff
@@ -0,0 +1,94 @@
...
+            <Box
+                title={'LEARNING PARAMETERS'}
+                content={
+                    <Grid columns={['1fr', '1fr']} rows={['auto', 'auto', 'auto', 'auto', 'auto']} gap={'size-100'}>
```
dwesolow: Same like before:
1. That should be separate component.
2. I think implicit rows would work here as well (without defining `rows`).

application/ui/src/index.css
```diff
@@ -104,3 +104,7 @@ svg {
 [data-testid='underlay'] {
     z-index: 2 !important;
 }
+
+[role='link'] {
```
dwesolow: I'm always sceptical about global styles, IMO it's better to change the style per instance (it prevents from having a case to revert the change to have the original behavior).

application/ui/src/features/models/model-listing/model-training-parameters/model-training-parameters.component.tsx
jpggvilaca: Works out of the box with Flex, the growth. With View you need to set more heights (children)

application/ui/src/features/models/model-listing/model-metrics/model-metrics.component.tsx
jpggvilaca: Updated

application/ui/src/features/models/model-listing/model-training-parameters/model-training-parameters.component.tsx
jpggvilaca: Agree, but for now i have no idea how many columns will be.

application/ui/src/features/models/model-listing/model-training-parameters/model-training-parameters.component.tsx
jpggvilaca: Updated

application/ui/src/index.css
jpggvilaca: Updated

dwesolow (approved):

PR #5176, opened 2026-01-13
Fix device selection logic in training job

leoll2:
### Summary

Fix a bug in the training job where OTX would not recognize the selected device.

### How to test

Submit a training job with different `"device"` values, e.g. `"cpu"`, `"xpu"`, `"xpu-0"`, ...

### Checklist

### DIFF
--- a/application/backend/app/services/training/otx_trainer.py
+++ b/application/backend/app/services/training/otx_trainer.py
@@ -26,6 +26,7 @@ from otx.data.dataset.base import OTXDataset
 from otx.data.module import OTXDataModule
 from otx.data.transform_libs.torchvision import TorchVisionTransformLib
 from otx.tools.converter import GetiConfigConverter
+from otx.types.device import DeviceType as OTXDeviceType
 from otx.types.export import OTXExportFormatType
 from otx.types.precision import OTXPrecisionType
 from otx.types.task import OTXTaskType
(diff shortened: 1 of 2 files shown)


application/backend/app/services/training/otx_trainer.py
```diff
@@ -337,10 +340,12 @@ def train_model(
 
         # Start training
         logger.info("Starting the training loop (model_id={})", model_id)
+        train_kwargs = {"devices": [device.index]} if device.type is not DeviceType.CPU and device.index else {}
```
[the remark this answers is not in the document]
leoll2: This is fine: if index is 0, the engine still selects the correct device (the first available one) even if we don't provide "devices".

A-Artemis (approved):

application/backend/app/services/training/otx_trainer.py
kprokofi, line 343: If we use "train_kwargs", maybe it is better to add there other train parameters that are added to the train function?
Doesn't it work if we pass devices = [0] with device type CPU?

kprokofi (approved):

application/backend/app/services/training/otx_trainer.py
leoll2: `devices = [0]` doesn't work with CPU:

> TypeError: `devices` selected with `CPUAccelerator` should be an int > 0.

PR #5200, opened 2026-01-14
Fix augmentation switcher for detection, fix img_info resizing when using torchvision.v2

kprokofi:
### Summary

- [x] Fix img_info resizing using native torchvision.v2 library (we define kernels to update img_info)
- [x] Fix datumaro commit to avoid incompatibility with ongoing development
- [x] Fix dynamic augmentation switch
- [x] Fix for deimV2 to resize bbox with no_aug policy

### Checklist


library/pyproject.toml
```diff
@@ -27,7 +27,7 @@ classifiers = [
     "Programming Language :: Python :: 3.12",
 ]
 dependencies = [
-    "datumaro[experimental] @ git+https://github.com/open-edge-platform/datumaro.git@develop",
+    "datumaro[experimental] @ git+https://github.com/open-edge-platform/datumaro.git@1a2653a006c550406845408219228fd55850cab2",
```
leoll2: Can you elaborate on the need to pin the Datumaro version? `1a2653a` is more than 1 month old, it doesn't include recent commits that are necessary for Geti Tune. If there is a stopper issue in Datumaro, please file a bug ticket and link it here in the code.

kprokofi: Yes, update to the latest develop leads to an issue with data typing in label fields. It is easy to fix. However, I found that even with this required change -> an error occurs accessing not existing parameter on the Datumaro side. 
I pin the Datumaro version for now.

I will provide filled issue

kprokofi: OK, the second issue is gone. I updated the typing, now we can use latest Datumaro

library/src/otx/data/entity/sample.py
```diff
@@ -175,7 +175,7 @@ class DetectionSample(OTXSample):
     image: tv_tensors.Image | np.ndarray | torch.Tensor = image_field(
         dtype=pl.UInt8(), format="BGR", channels_first=True
     )
-    label: torch.Tensor = label_field(pl.Int32(), is_list=True)
+    label: torch.Tensor = label_field(pl.UInt8(), is_list=True)
```
leoll2, line 178: Can you also apply `dtype=pl.UInt8()` to `InstanceSegmentationSample.label`, for consistency?

AlbertvanHouten (approved):

leoll2 (approved):

PR #5211, opened 2026-01-15
Extend model deletion endpoint for files only

### LINKED ISSUE #5189: Extend model deletion endpoint with option 'files_only'
**Endpoint:** `DELETE /api/projects/{project_id}/models/{model_id}`

**New query parameter:** `files_only` (optional bool, default False)

**Response**: 204 No Content (success)

When `files_only=True`, the endpoint should delete all files from the model data folder, then set `files_deleted` to True. If the files are already deleted, then the endpoint shoudn't do anything but simply return.

The purpose of this query parameter is to allow the user to free disk space, while keep a reference of the model revision and basic metadata (stored in the DB instead of in the model data folder).

Note that `ModelService.get_model_files_path` already has the right logic to deal with models with deleted files.

A-Artemis:
### Summary

Extend the delete model endpoint with a new query `files_only=False` which when set to `True` will delete only the model files from disk to save space, but persist the db document with `files_deleted=True`.

### Checklist

### DIFF
--- a/application/backend/app/services/model_service.py
+++ b/application/backend/app/services/model_service.py
@@ -116,6 +116,33 @@ class ModelService(BaseSessionManagedService):
         except IntegrityError:
             raise ResourceInUseError(ResourceType.MODEL, str(model_id))
 
+    @parent_process_only
+    def delete_model_files(self, project_id: UUID, model_id: UUID) -> None:
+        """
+        Delete only the model files from disk, keeping the model revision record in the database and setting its
+        files_deleted flag to True.
+
+        Args:
+            project_id (UUID): The unique identifier of the project.
+            model_id (UUID): The unique identifier of the model.
+        """
+        try:
+            model_files_path = self.get_model_files_path(project_id=project_id, model_id=model_id)
+        except FileNotFoundError:
+            return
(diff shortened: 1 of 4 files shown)


application/backend/app/services/model_service.py
```diff
@@ -116,6 +116,33 @@ def delete_model(self, project_id: UUID, model_id: UUID) -> None:
...
+
+        # Delete model files from disk
+        shutil.rmtree(model_files_path)
+        logger.info("Deleted model files at '{}'", model_files_path)
+
+        # Mark as deleted in the database
+        model_rev_repo = ModelRevisionRepository(project_id=str(project_id), db=self.db_session)
+        model_rev_db = model_rev_repo.get_by_id(str(model_id))  # model_revision is present as model_files_path exists
+        model_rev_db.files_deleted = True  # type: ignore[union-attr]
+        model_rev_repo.update(model_rev_db)  # type: ignore[arg-type]
```
leoll2: You can remove the `type: ignore` if you `cast(ModelRevisionDB, model_rev_repo.get_by_id(str(model_id)))`

leoll2 (approved):

PR #5217, opened 2026-01-16
Update camera enumeration

maxxgx:
### Summary

Fix from https://github.com/open-edge-platform/anomalib/pull/3264
- Update `cv2-enumrate-cameras` to latest version `1.3.3` (contains a fix for MacOS).
- Use default capture backend to avoid duplicated cameras listings using different backends. For example, the library supports two backends in Linux and both results are returned if backend is not specified:
image

### Checklist

### DIFF
--- a/application/backend/app/services/system_service.py
+++ b/application/backend/app/services/system_service.py
@@ -1,8 +1,10 @@
 # Copyright (C) 2025 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
-
+import platform
 import re
+from collections import defaultdict
 
+import cv2
 import psutil
 import torch
 from cv2_enumerate_cameras import enumerate_cameras
@@ -11,6 +13,11 @@ from app.models.system import CameraInfo, DeviceInfo, DeviceType
 
 DEVICE_PATTERN = re.compile(r"^(cpu|xpu|cuda)(-(\d+))?$")
 DEFAULT_DEVICE = "cpu"
+CV2_BACKENDS = {
+    "Windows": cv2.CAP_MSMF,
+    "Linux": cv2.CAP_V4L2,
+    "Darwin": cv2.CAP_AVFOUNDATION,
+}
 
 
 class SystemService:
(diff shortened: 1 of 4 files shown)


application/backend/app/services/system_service.py
```diff
@@ -162,8 +169,22 @@ def _parse_device(device_str: str) -> tuple[DeviceType, int]:
...
         """
         Get available camera devices.
+        If duplicate names are present, append a suffix to make them unique.
+        Example: ["camera", "camera"] -> ["camera", "camera (1)"]
 
         Returns:
             list[CameraInfo]: List of available camera devices
         """
-        return [CameraInfo(index=camera.index, name=camera.name) for camera in enumerate_cameras()]
+        if (backend := CV2_BACKENDS.get(platform.system())) is None:
```
leoll2, line 176: Would it be more reliable to use `platform.uname` instead of `platform.system`. From the [docs](cv2-enumerate-cameras):

> Returns the system/OS name, such as 'Linux', 'Darwin', 'Java', 'Windows'. An empty string is returned if the value cannot be determined.
> On iOS and Android, this returns the user-facing OS name (i.e, 'iOS, 'iPadOS' or 'Android'). To obtain the kernel name ('Darwin' or 'Linux'), use [os.uname()](https://docs.python.org/3/library/os.html#os.uname).

ashwinvaidya17 (dismissed):

application/backend/app/services/system_service.py
```diff
@@ -162,8 +169,22 @@ def _parse_device(device_str: str) -> tuple[DeviceType, int]:
     def get_camera_devices() -> list[CameraInfo]:
         """
         Get available camera devices.
+        If duplicate names are present, append a suffix to make them unique.
+        Example: ["camera", "camera"] -> ["camera", "camera (1)"]
```
leoll2: The Windows-style handling of duplicates doesn't help the user to recognize the cameras imho, because the suffix doesn't represent anything. I suggest to construct the name as `"{cam.name} [{cam.index}]"` (e.g. `"Logitech WebCam [3]"`) which is unique and easy to implement.

leoll2 (commented):
Thanks for the fix in `cv2-enumerate-cameras` and backporting!

application/backend/app/services/system_service.py
leoll2: Nvm, `uname` is not available on Windows

application/backend/app/services/system_service.py
maxxgx: The suffix just serves to distinguish between cameras with the same name. 
Are you suggesting adding the index to all items, regardless of whether it's a duplicate or not?
For example, if the system has `Logitech Webcam` x 3, then enumerate as follow:
```
Laptop Camera [1]
Logitech Webcam [2]
Logitech Webcam [3]
Logitech Webcam [4]
```

leoll2: Correct. In my opinion, adding the index doesn't affect the readability and it helps to disambiguate between devices with similar or identical names.

maxxgx: Fixed. I agree it's a simply and effective solution

leoll2: Thanks! Could you also update the docstring accordingly?

maxxgx: Done

leoll2 (approved):

Issue #4946, opened 2025-10-22
Video support


leoll2:
## Objective

Allowing users to work with videos in Geti Tune, supporting video file uploads, frame annotation, and the use of annotated frames to fine-tune models.

### Planned for MVP

- [Upload](#upload)
- [Navigation](#navigation)
- [Annotation](#annotation)
- [Training](#training)
- [Dataset export with videos](#dataset-export-with-videos)

### Not planned for MVP

Additional features, NOT in scope of the current requirement, to be planned later:

- [Inference during playback](#inference-during-playback)
- [Bulk annotation](#bulk-annotation)

---

### Upload

- Users should be able to upload video files in various common formats (MP4, AVI, MKV, ...).
  - Basic validation can be performed, e.g. on the extension of the uploaded file
  - It is not necessary to validate the video content at this stage. If the data is corrupted/unreadable, an error will be generated later when the user attempts to browse the video.
  - It is also not necessary to fully resolve the video metadata; attributes like "width" or "height" can be initialized lazily, when the video needs to be opened for the first time.
- After the video is fully uploaded, it must be shown in the project dataset.
  - The video should have a thumbnail, which may be generated on-the-fly on the first request.
- Users should be able to download videos.
- Users should be able to delete videos.

### Navigation
 
- Users should be able to open the video at any frame timestamp and view the corresponding frame.
- Users should be able to playback the video from any timestamp, and pause/resume the stream when they desire.

### Annotation

- Users should be able to annotate individual video frames, at the specified timestamp.
- The user should be able to retrieve and view the annotations of a video frame that was previously annotated
  - Annotations may apply to an _exact_ timestamp, as opposed to a small time interval corresponding to a frame
    - Example 1: if the client annotates a frame at t=20.0 then fetches the annotation info for t=20.0, the server must return the annotation.
    - Example 2: if the client annotates a frame at t=20.0 then fetches the annotation info for t=20.001, the server may not return the annotation even if, in practice, both timestamps would resolve to the same frame.
      - The actual behavior will be decided at design phase.
- The user should be able to edit a previously created annotation.
- During video playback, the user should have an option to view/hide the annotations of the displayed frames.

### Training

- For training-related workflows, annotated video frames are functionally equivalent to annotated images; each frame can be represented as a dataset item and assigned to a subset.
- A dataset item should be created for a specific frame when it is annotated for the first time.

### Dataset export with videos

[...]

PR #5224, opened 2026-01-16
Video upload design

### LINKED ISSUE #5181: Design for video support in Geti Tune (core functionalities)
Write a design proposal to provide support for videos in Geti Tune, with the basic functionalities described in #4946 (phase 1): upload, annotation, navigation, training.

The design should cover the following items:
- API changes
- DB and storage changes
- Data flow (i.e., describe what happens from when a video is uploaded to when it's used for training)
- Validation

For the video features categorized as phase 2 and 3, it is not necessary to fully architect them, but it's worth considering them to avoid design choices that could hinder subsequent iterations.

warrkan:
### Checklist


application/docs/dataset.md
```diff
@@ -7,20 +7,42 @@ New items can be added to the dataset from:
...
 The dataset, uniquely associated with a project, serves as input for various analytical workflows such as model
 fine-tuning, evaluation, and data exploration. The user has full control over the dataset, being able to add and remove
 elements, edit annotations and other attributes.
 
+### Media vs Dataset item
+
+In Geti Tune, a _media_ refers to the actual image or video file that contains visual content. Media files are the raw
+data that the system processes, analyzes, and displays. They can be in various formats such as JPEG, PNG, MP4, etc.
+Video frame is also considered a media and is stored as an individual image file and cannot be annotated.
+Video itself cannot be annotated, only its frames can.
```
leoll2: L20 says "Video frame ... cannot be annotated"
L21 says "Video ... frames can"

Probably there is a typo somewhere because the two sentences are contradicting.

application/docs/dataset.md
```diff
@@ -7,20 +7,42 @@ New items can be added to the dataset from:
...
+
+A _dataset item_, on the other hand, is a more comprehensive entity that includes not only the media itself but also
+additional metadata and annotations associated with that media. Dataset items are the units of data that are used
+within the context of a dataset for machine learning tasks. Each dataset item typically contains:
+- The media file (image or video frame). Video CANNOT have a dataset item associated with it.
+- Annotations: Labels, bounding boxes, segmentation masks, or other forms of annotations that describe the content of the media.
+- Metadata: Additional information such as timestamps, source information, tags, and other relevant attributes.
+- Source reference: Information about where the media originated from (optionally).
```
leoll2: 'Source' should be an attribute of the media instead of the dataset item

application/docs/api.md
```diff
@@ -55,24 +55,31 @@
...
+| Method   | Path                                              | Payload | Return                | Description                                     |
+| -------- |---------------------------------------------------| ------- |-----------------------|-------------------------------------------------|
+| `GET`    | `/api/projects/<id>/dataset/media`                | -       | list of dataset media | List the dataset media (images and videos)      |
+| `GET`    | `/api/projects/<id>/dataset/media/<id>`           | -       | dataset media info    | Get info about a dataset media                 |
+| `GET`    | `/api/projects/<id>/dataset/media/<id>/binary`    | -       | binary                | Get the image data of a dataset item (full res) |
+| `GET`    | `/api/projects/<id>/dataset/media/<id>/thumbnail` | -       | binary                | Get the thumbnail of a dataset item             |
+| `POST`   | `/api/projects/<id>/dataset/media`                | binary  | media info            | Upload an image or a video to the dataset       |
```
leoll2: Can you share an example of response for images and videos?

application/docs/api.md
```diff
@@ -55,24 +55,31 @@
...
 
 ### Annotations
 
 | Method   | Path                                                | Payload         | Return          | Description                               |
 | -------- | --------------------------------------------------- | --------------- | --------------- | ----------------------------------------- |
-| `GET`    | `/api/projects/<id>/dataset/items/<id>/annotations` | -               | annotation info | Get the annotation/prediction for a media |
-| `POST`   | `/api/projects/<id>/dataset/items/<id>/annotations` | annotation info | annotation info | Annotate a media                          |
-| `DELETE` | `/api/projects/<id>/dataset/items/<id>/annotations` | -               | -               | Delete the annotation for a media         |
+| `GET`    | `/api/projects/<id>/dataset/media/<id>/annotations` | -               | annotation info | Get the annotation/prediction for a media |
+| `POST`   | `/api/projects/<id>/dataset/media/<id>/annotations` | annotation info | annotation info | Annotate a media                          |
```
leoll2, line 75: In the video case, am I right that `<media_id>` refers to the video frame id, i.e. a UUID that uniquely identifies the frame? If so, how does the client get to know the id of a specific frame? Do you return all frame ids in the `GET /media/<video_id>` response (cons: very large response size), or do you have an endpoint to resolve video_id + timestamp to a video_frame_id, or something else?

application/docs/dataset.md
```diff
@@ -85,6 +107,13 @@ items), but the actual items are removed and can no longer be accessed.
 
 The database saves the relevant information in the following tables:
 
+- `media`: contains all media records, for images, videos & video frames. Each record is identified by a unique id
+  and contains metadata about the media (filename, format, size, shape, etc.).
+- `videos`: contains one record for each video, identified by a media id. Video share the id with corresponding `media` record. 
+  Each record contains metadata about the video (fps, frame count) in addition to base media metadata.
```
leoll2: The `videos` table can be probably merged with `media` to avoid frequent JOINs. Image- or video-specific fields can be made nullable, possibly with some column-level constraints. Example:
```sql
CREATE TABLE media (
    id     INTEGER PRIMARY KEY,
    type   TEXT NOT NULL CHECK (type IN ('image', 'video')),
    fps    INTEGER,

    CHECK (
        (type = 'image' AND fps IS NULL) OR
        (type = 'video' AND fps IS NOT NULL)
    )
);
```

application/docs/dataset.md
```diff
@@ -85,6 +107,13 @@ items), but the actual items are removed and can no longer be accessed.
...
+- `video_frames`: contains one record for each video frame, identified by a media id. Video frame share the id with corresponding `media` record. 
+  Each record contains metadata about the video frame (timestamp) in addition to base media metadata.
```
leoll2: If the base media metadata is already stored in `media`, why replicate it here for each video frame?

application/docs/diagrams/storage/full.puml
```diff
@@ -60,9 +60,38 @@ table(labels) {
...
+  fk(media_id) : UUID NOT NULL
+  --
+  'Video related details
+  fps : INTEGER NOT NULL
+  frames_count : INTEGER NOT NULL
+}
+
+table(video_frames) {
+  fk(media_id) : UUID NOT NULL
+  fk(video_id) : UUID NOT NULL
```
leoll2, line 87: Please clarify the difference between `media_id` and `video_id` with a comment.

Based on my understanding, `media_id` is the video frame id, i.e. it points to the `media` record with `type: video_frame`, whereas `video_id` is the id of the parent video (`media` record with `type: video`).

application/docs/diagrams/storage/full.puml
```diff
@@ -60,9 +60,38 @@ table(labels) {
...
+
+table(videos) {
+  fk(media_id) : UUID NOT NULL
+  --
+  'Video related details
+  fps : INTEGER NOT NULL
+  frames_count : INTEGER NOT NULL
+}
+
+table(video_frames) {
```
leoll2, line 83: Are the `video_frames` records created synchronously on media upload, or when the frame is later extracted (at annotation time)?

application/docs/dataset.md
warrkan: Indeed, fixed!

application/docs/dataset.md
warrkan: Correct, moved it to the `media`

application/docs/dataset.md
warrkan: Agree, here we can optimize schema a bit to reduce number of joins. Merged two tables

application/docs/diagrams/storage/full.puml
warrkan: They are created synchronously later when certain frame is being annotated

application/docs/diagrams/storage/full.puml
warrkan: Added comments

application/docs/dataset.md
warrkan: Indeed, removed this extra sentence part. We don't have additional metadata in `video_frames`

application/docs/api.md
warrkan: For adding a media (both image and video), we'll use existing schema https://github.com/open-edge-platform/training_extensions/blob/e8e3d4233438d3f7832250a5a00d9afaf181cc23/application/backend/app/api/schemas/dataset_item.py#L52, just rename it to `MediaView`.

application/docs/api.md
warrkan: Not really, when we add the annotation to a video frame, it's not extracted yet. Therefore there is no media_id yet.
In this case client passes video's `media_id` and frame timestamp. Backend extracts video frame binary, stores it to file system and then adds record to both `media` and `video_frame` tables.

application/docs/api.md
leoll2: Some fields are not applicable though, and some fields may be image- or video- specific. For example:
- `subset` should be removed, or more precisely moved to the dataset item API
- only for videos, the response should also include fps, number of frames, ...

application/docs/api.md
leoll2: Okay, so if I understand correctly, only the following entities have a `media_id` (UUID):
- images
- videos
- annotated (extracted) video frames

while unannotated video frames don't have an id. I'm not against this idea, but we have to consider possible inconsistencies/non-orthogonal behaviors. For example, to annotate a video frame, you propose something like `POST /media/<video_id>/annotations?timestamp=<ts>`, which means that:
- The annotation endpoint must generate an id (media_id) for the video frame, but only if it's a newly annotated frame. It doesn't for images, or previously annotated frames. In other words, the responsibility of creating media entities (records with media_id) is spread across the media upload and the annotation endpoints.
- The annotation endpoint must return the media id; not a big deal, but it's one more API change.
- There are potentially two ways to get/update the annotation of an already annotated frame:
  -  `GET/POST /media/<video_id>/annotations?timestamp=<ts>`
  - `GET/POST /media/<video_frame_id>/annotations`
  We can disallow the latter, but then I wonder if there is any value at all in assigning a media_id to video frames. 

Potentially, the API could still work without ever assigning a UUID to video frames, by relying on the image/video UUID as `media_id` plus an optional query parameter `timestamp` that discriminate between videos and video frames. What do you think?

application/docs/api.md
warrkan: Indeed, for video we need to return fps and number of frames as well, so the entity will be extended. We may keep subset in the output if needed, for video it will has an array with all distinct frames subsets.

leoll2: >  We may keep subset in the output if needed, for video it will has an array with all distinct frames subsets.

Not needed, the subset only matters for ML tasks, i.e. at the dataset item level. Media don't need it.

application/docs/api.md
warrkan: Yes right, images, videos and vide frames will have `media_id`.

I think we need to store additionally information about the frame, like resolution, format. We might want for example to resize frames used in dataset, then the resolution will differ from video.
Then frame concept fits perfectly into media table, and then we need to generate a media_id for it.
We can support two was to identify video frame:
1. Using media_id of the video frame
2. Using media_id of the parent video + timestamp.

Then all the endpoints which operate with media will support both options.

application/docs/api.md
```diff
@@ -55,24 +55,32 @@
...
+## Media
+
+| Method   | Path                                              | Payload | Return                         | Description                                     |
+| -------- |---------------------------------------------------| ------- |--------------------------------|-------------------------------------------------|
+| `GET`    | `/api/projects/<id>/dataset/media`                | -       | list of dataset media          | List the dataset media (images and videos)      |
+| `GET`    | `/api/projects/<id>/dataset/media/<id>`           | -       | dataset media info             | Get info about a dataset media                  |
+| `GET`    | `/api/projects/<id>/dataset/media/<id>/frames`    | -       | list of annotated video frames | List the annotated video frames                  |
+| `GET`    | `/api/projects/<id>/dataset/media/<id>/binary`    | -       | binary                         | Get the image data of a dataset item (full res) |
+| `GET`    | `/api/projects/<id>/dataset/media/<id>/thumbnail` | -       | binary                         | Get the thumbnail of a dataset item             |
```
leoll2:
```suggestion
| `GET`    | `/api/projects/<id>/dataset/media/<id>/binary`    | -       | binary                         | Get the image data of a media (full res) |
| `GET`    | `/api/projects/<id>/dataset/media/<id>/thumbnail` | -       | binary                         | Get the thumbnail of a media             |
```

application/docs/dataset.md
```diff
@@ -7,20 +7,41 @@ New items can be added to the dataset from:
...
 The dataset, uniquely associated with a project, serves as input for various analytical workflows such as model
 fine-tuning, evaluation, and data exploration. The user has full control over the dataset, being able to add and remove
 elements, edit annotations and other attributes.
 
+### Media vs Dataset item
+
+In Geti Tune, a _media_ refers to the actual image or video file that contains visual content. Media files are the raw
+data that the system processes, analyzes, and displays. They can be in various formats such as JPEG, PNG, MP4, etc.
+Video frame is also considered a media and is stored as an individual image file and can be annotated.
+Video itself cannot be annotated, only its frames can. Media can have a reference to its source, such as a camera for example.
```
leoll2:
```suggestion
Images and video frames can be annotated. When a frame is annotated, Geti Tune extracts it 
from the video, assigns it an id, and saves it as an image; the extracted frame can also be
considered a media, supporting all the operations that are normally possible for images.
A media may have a reference to a source, e.g. the camera from which it was acquired.
```
slightly rephrased for clarity

application/docs/dataset.md
```diff
@@ -7,20 +7,41 @@ New items can be added to the dataset from:
...
+
+A _dataset item_, on the other hand, is a more comprehensive entity that includes not only the media itself but also
+additional metadata and annotations associated with that media. Dataset items are the units of data that are used
+within the context of a dataset for machine learning tasks. Each dataset item typically contains:
+- The media file (image or video frame). Video CANNOT have a dataset item associated with it.
```
leoll2:
```suggestion
A _dataset item_, on the other hand, is the fundamental unit of a machine learning dataset.
Each dataset item corresponds to an annotated image or video frame, enriched with metadata
that is relevant for ML tasks. Each dataset item typically contains:
- A reference to the media file (image or video frame). Video CANNOT have a dataset item associated with it.
```
slightly rephrased for better clarity

application/docs/dataset.md
```diff
@@ -7,20 +7,41 @@ New items can be added to the dataset from:
...
+- Annotations: Labels, bounding boxes, segmentation masks, or other forms of annotations that describe the content of the media.
+- Metadata: Additional information such as timestamps, source information, tags, and other relevant attributes.
+- Subset assignment: Information about whether the item belongs to the training, validation, or testing subset of the dataset.
+- Other relevant properties that may be needed for dataset management and processing.
+
+Dataset item and media are closely related, but they serve different purposes within the Geti Tune ecosystem.
+Dataset item and media share the same unique identifier (UUID), which allows for easy correlation between them.
```
leoll2:
```suggestion
A dataset item shares the same unique identifier (UUID) with the media it refers to, which allows easy correlation.
In other words, the id of a dataset item always corresponds to the id of an image or an extracted video frame.
```

leoll2 (commented):
Minor comments, LGTM

application/docs/dataset.md
```diff
@@ -7,20 +7,44 @@ New items can be added to the dataset from:
...
 
 The dataset, uniquely associated with a project, serves as input for various analytical workflows such as model
 fine-tuning, evaluation, and data exploration. The user has full control over the dataset, being able to add and remove
 elements, edit annotations and other attributes.
 
+### Media vs Dataset item
+
+In Geti Tune, a _media_ refers to the actual image or video file that contains visual content. Media files are the raw
+data that the system processes, analyzes, and displays. They can be in various formats such as JPEG, PNG, MP4, etc.
+Images and video frames can be annotated. When a frame is annotated, Geti Tune extracts it
```
AlbertvanHouten, line 20: How will this work for predictions on frames? Will each predicted frame also be extracted?

application/docs/diagrams/er/full.puml
```diff
@@ -68,6 +68,27 @@ entity(DatasetRevision) {
...
+  format : TEXT
+  width : INTEGER
+  height : INTEGER
+  size : INTEGER
+}
+
+entity(Video) {
+  pk(media_id) : UUID
+  fps : INTEGER
+  frames_count : INTEGER
```
AlbertvanHouten: Either frame_count or total_frames would be better IMO

application/docs/diagrams/storage/datasets.puml
```diff
@@ -75,6 +94,15 @@ class projects << (T,#CCCCCC) >> {
 projects ||---o{ dataset_items
 projects ||---o{ dataset_views
 projects ||---o{ dataset_revisions
+projects ||---o{ media
+
+dataset_items ||---|| media
+
+
+media ||---o| video_frames
```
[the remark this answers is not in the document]
warrkan: media and video_frames have two relations - one via id and another via video_id

application/docs/dataset.md
warrkan: No, we will extracted frames only for fine-tuning, not for predictions

leoll2: Prediction on videos is a feature planned for a later date (see req. #4946). The UX isn't ready yet, but I imagine that the user should have an option to save predictions and accept/edit them. If so, we likely need to extract the frames while we run inference on them.

application/docs/entities.md
```diff
@@ -2,6 +2,8 @@
 
 ![Full ER diagram with all entities](media/er/full.svg)
 
+TODO: update
```
leoll2: Open TODO

application/docs/diagrams/storage/full.puml
```diff
@@ -60,9 +60,39 @@ table(labels) {
...
+  ' Reference to a media record with type 'video_frame' containing all the attributes and metadata of the video frame itself
+  fk(media_id) : UUID NOT NULL
+  ' Reference to the video media record this frame belongs to with certain timestamp
+  fk(video_id) : UUID NOT NULL
+  --
+  'Video frame related details
+  timestamp : INTEGER NOT NULL
+}
+
+table(dataset_items) {
```
leoll2, line 93: The media-related details (name, format, width, height, size) can be removed from `dataset_items`, since they're stored in `media` and easily accessible through `media_id`.

application/docs/diagrams/er/full.puml
```diff
@@ -68,6 +68,27 @@ entity(DatasetRevision) {
...
+entity(Video) {
+  pk(media_id) : UUID
+  fps : INTEGER
+  frame_count : INTEGER
+}
+
+entity(VideoFrame) {
+  pk(media_id) : UUID
+  timestamp: INTEGER
+}
```
leoll2: For simplicity, you can omit `Video` and `VideoFrame` in this diagram, the important relations are captured by `Media`. Else, if you want to keep `Video` and `VideoFrame`, you should also add `Image` for consistency, but it makes the diagram overly complicated imho.

Video-specific and frame-specific attributes can be moved to `Media` with a comment.

application/docs/dataset.md
warrkan: Correct, but not all of them. It will work similar to data collection

leoll2 (approved):

warrkan:
Resolves #5181

PR #5227, opened 2026-01-16
Delete & rename model

jpggvilaca:
### Summary

Screenshot 2026-01-16 at 16 12 48
Screenshot 2026-01-16 at 16 12 55

### Checklist


application/ui/src/features/models/hooks/api/use-get-model.hook.ts
```diff
@@ -12,7 +12,7 @@ export const useGetModel = (modelId: string | null | undefined) => {
         return undefined;
     }
 
-    return $api.useSuspenseQuery('get', '/api/projects/{project_id}/models/{model_id}', {
+    return $api.useQuery('get', '/api/projects/{project_id}/models/{model_id}', {
```
jpggvilaca, line 15: this is because `useSuspense` throws and `useQuery` doesnt, which is what we want or else if we get a 404 (which i did because i was using mocked data), the app crashes ungracefully

application/ui/src/features/models/model-listing/components/expandable-search/expandable-search.component.tsx
```diff
@@ -38,7 +37,7 @@ export const ExpandableSearch = ({ value, onChange }: ExpandableSearchProps) =>
     };
 
     return (
-        <Flex UNSAFE_className={clsx(classes.container, { [classes.expanded]: isExpanded })}>
+        <Flex>
```
jpggvilaca, line 40: previous comment from Dawid

application/ui/src/features/models/model-listing/components/model-row/model-row.container.tsx
```diff
@@ -0,0 +1,90 @@
+// Copyright (C) 2025 Intel Corporation
```
jpggvilaca, line 1: made the ModelRow dumb and easily testable, and created the `container` to handle all heavy logic

application/ui/src/features/models/model-listing/components/model-row/model-row.container.tsx
```diff
@@ -0,0 +1,90 @@
...
+    const [isRenameDialogOpen, setIsRenameDialogOpen] = useState(false);
+    const [isDeleteDialogOpen, setIsDeleteDialogOpen] = useState(false);
+
+    const handleAction = (key: Key) => {
+        if (key === MODEL_ACTIONS.DELETE) {
+            setIsDeleteDialogOpen(true);
+        } else if (key === MODEL_ACTIONS.RENAME) {
+            setIsRenameDialogOpen(true);
+        } else if (key === MODEL_ACTIONS.EXPORT) {
+            // TODO: Implement export functionality
```
jpggvilaca, line 42: no endpoint yet.

application/ui/src/features/models/model-listing/components/model-row/rename-model-dialog.component.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+        onClose();
+    };
+
+    return (
+        <Dialog>
+            <Heading>Rename Model</Heading>
+            <Divider />
+            <Content>
+                <TextField label='Model name' value={newName} onChange={setNewName} width='100%' />
```
camiloHimura: How do we indicate an empty value with a red border warning?

jpggvilaca: At the moment we dont, we just disable the button. Do we need the red border?

application/ui/src/features/models/hooks/api/use-delete-model.hook.ts
```diff
@@ -0,0 +1,16 @@
+// Copyright (C) 2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+import { $api } from '../../../../api/client';
+
+export const useDeleteModel = (modelId: string | null | undefined) => {
+    if (!modelId) {
```
dwesolow: We don't need `modelId` in this hook when we don't call mutation inside.

application/ui/src/features/models/hooks/api/use-rename-model.hook.ts
```diff
@@ -0,0 +1,16 @@
+// Copyright (C) 2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+import { $api } from '../../../../api/client';
+
+export const useRenameModel = (modelId: string | null | undefined) => {
+    if (!modelId) {
```
dwesolow: Are we sure we're not going to support updating more properties than name of the model?
Maybe it'd be better to name it `useUpdateModel`?

Same thing like with `useDeleteModel`.

application/ui/src/features/models/model-listing/components/model-row/model-row.component.tsx
```diff
@@ -1,27 +1,32 @@
...
+type ModelRowProps = {
     model: SchemaModelView;
-}
-
-export const ModelRow = ({ model }: ModelRowProps) => {
-    const { activeModelId, onExpandModel } = useModelListing();
+    activeModelId?: string;
+    parentRevisionModel?: SchemaModelView;
+    onExpandModel: (modelId: string) => void;
+    handleModelAction: (key: Key) => void;
```
dwesolow: np. `onModelAction`

application/ui/src/features/models/model-listing/components/model-row/model-row.component.tsx
```diff
@@ -1,27 +1,32 @@
...
 
-import { SchemaModelView } from '../../../../api/openapi-spec';
-import { useGetModel } from '../../hooks/api/use-get-model.hook';
-import { GRID_COLUMNS } from '../constants';
-import { AccuracyIndicator } from '../model-variants/accuracy-indicator.component';
-import { useModelListing } from '../provider/model-listing-provider';
-import { formatTrainingDateTime } from '../utils/date-formatting';
-import { ActiveModelTag } from './active-model-tag.component';
-import { ParentRevisionModel } from './parent-revision-model.component';
+import { SchemaModelView } from '../../../../../api/openapi-spec';
```
dwesolow, line 7: Sometimes we import from `shared-types.ts`, sometimes from `openapi-spec`. I'm sure that even if the type is not shared yet (only used inside one file), it will be used by more than one component. That's why I'd suggest using types directly from `@/api/types` for example.

application/ui/src/features/models/hooks/api/use-delete-model.hook.ts
jpggvilaca: True! Removed from both hooks

application/ui/src/features/models/hooks/api/use-rename-model.hook.ts
jpggvilaca: I dont see which other fields we would update. Even the name is very overkill imho, so for now i'd rather leave it as it is. If we extend it we can update the name

application/ui/src/features/models/model-listing/components/model-row/model-row.component.tsx
jpggvilaca: We can definitely improve on this later, agree. We dont necessarily need the alias but least have the shared-types contain all of them, and have all files importing from it

application/ui/src/features/models/model-listing/components/model-row/model-row.component.tsx
jpggvilaca: Updated

application/ui/src/features/models/model-listing/components/model-row/model-row.container.tsx
```diff
@@ -0,0 +1,90 @@
...
+        }
+    };
+
+    const handleRename = (newName: string) => {
+        renameModelMutation?.mutate({
```
dwesolow: This caller or hook should invalidate the cache so we the model details entity is refetched as well.

application/ui/src/features/models/model-listing/components/model-row/rename-model-dialog.component.tsx
```diff
@@ -0,0 +1,44 @@
...
+    const [newName, setNewName] = useState(currentName);
+
+    const isValid = newName.trim().length > 0;
+
+    const handleRename = () => {
+        if (isValid && newName !== currentName) {
+            onRename(newName.trim());
+        }
+
+        onClose();
```
dwesolow: We should call `onClose` only when action was completed successfully.

application/ui/src/features/models/model-listing/components/model-row/model-row.container.tsx
jpggvilaca: I did invalidate the query on the hook, for all models. I chose to do this to make the caller much simpler. I could be more granular and do it always on the caller, but then the "boilerplate" is larger.

application/ui/src/features/models/model-listing/components/model-row/rename-model-dialog.component.tsx
```diff
@@ -0,0 +1,44 @@
...
+    onClose: () => void;
+}
+
+export const RenameModelDialog = ({ currentName, onRename, onClose }: RenameModelDialogProps) => {
+    const [newName, setNewName] = useState(currentName);
+
+    const isValid = newName.trim().length > 0;
+
+    const handleRename = () => {
+        if (isValid && newName !== currentName) {
```
dwesolow: What about disabling button when `newName === currentName`? Then we don't need to do this `if` here, only call `onRename(newName.trim())`.

application/ui/src/features/models/model-listing/components/model-row/rename-model-dialog.component.tsx
dwesolow: In prompt I refactored a code so we always/almost always use `Form validationBehavior="native" onSubmit={handleRename}` with `isRequired` on `TextField`. This way we'd also support updating model with "enter" key.

application/ui/src/features/models/model-listing/components/model-row/rename-model-dialog.component.tsx
```diff
@@ -0,0 +1,44 @@
...
+            <Heading>Rename Model</Heading>
+            <Divider />
+            <Content>
+                <TextField label='Model name' value={newName} onChange={setNewName} width='100%' />
+            </Content>
+            <ButtonGroup>
+                <Button variant='secondary' onPress={onClose}>
+                    Cancel
+                </Button>
+                <Button variant='accent' onPress={handleRename} isDisabled={!isValid}>
```
dwesolow: "If the action takes some time, we should also disable the rename button (as we usually do with isPending={mutation.isPending}) to avoid sending multiple identical requests."

application/ui/src/features/models/model-listing/components/model-row/rename-model-dialog.component.tsx
jpggvilaca: I think it's fine if we do it either way. The notification will show regardless

application/ui/src/features/models/model-listing/components/model-row/rename-model-dialog.component.tsx
jpggvilaca: Not a big difference, but updated

application/ui/src/features/models/model-listing/components/model-row/rename-model-dialog.component.tsx
jpggvilaca: I like it . updated

application/ui/src/features/models/model-listing/components/model-row/rename-model-dialog.component.tsx
jpggvilaca: Agree, updated, even tho the isPending only triggers after 1 second, which still lets me click a few times

application/ui/src/features/models/hooks/api/use-rename-model.hook.ts
dwesolow: Ok, just asking because in prompt we only supported model activation, hook was called `useActivateModel` and then we added support for configuration so I had to update the hook name as well (together with file).

application/ui/src/features/models/model-listing/components/model-row/rename-model-dialog.component.tsx
dwesolow: yeah but I think we should not close when the error happened (which might happen because of wrong name that was not detected by ui validation).
At least i'd be nervous when form action would not be successful and dialog disappears, maybe I had special name invented? :D

application/ui/src/features/models/model-listing/components/model-row/rename-model-dialog.component.tsx
dwesolow: Difference is that we indicate to the user that the name must be changed to submit the action.

application/ui/src/features/models/model-listing/components/model-row/rename-model-dialog.component.tsx
dwesolow: Yeah that's why I usually do `isButtonDisabled = mutation.isPending | !isValid` etc.

application/ui/src/features/models/model-listing/components/model-row/model-row.container.tsx
```diff
@@ -0,0 +1,98 @@
...
+            {
+                onSuccess: () => {
+                    setIsRenameDialogOpen(false);
+                },
+            }
+        );
+    };
+
+    const handleDelete = () => {
+        deleteModelMutation.mutate({ params: { path: { project_id: projectId, model_id: model.id } } });
```
dwesolow, line 61: do we need to close the alert dialog when the deletion is completed or it happens out of the box?

jpggvilaca: it happens out of the box

dwesolow (approved):

application/ui/src/features/models/model-listing/components/model-row/model-row.container.tsx
dwesolow: Discussed offline, don't needed for now. To revisit later.

PR #5240, opened 2026-01-19
Correct model training info time

### LINKED ISSUE #5190: Model revision endpoint returns null for training start_time / end_time
The endpoint `GET /api/projects/{project_id}/models/{model_id}` is expected to return the start/end time of the training job as follows:
```json
{
  ...
  "training_info": {
    "start_time": "2021-06-29T16:24:30.928000+00:00",
    "end_time": "2021-06-29T16:24:30.928000+00:00",
    ...
  },
}
```
In the DB, this information is stored in the `"model_revisions"` table as `training_started_at` and `training_finished_at`; however, the value don't seem to be correctly mapped to the business/presentation model `TrainingInfo` ([ref](https://github.com/open-edge-platform/training_extensions/blob/9d27aadfbfea87c7c9186486a1e3331d4d79ed4b/application/backend/app/models/model_revision.py#L72)), as a result the endpoint returns `null` for `start_time` and `end_time`.

A-Artemis:
### Summary

ModelRevisionDB and ModelView uses different naming scheme for time stamps, the mapping has been fixed.

### How to test

`GET /api/projects/{project_id}/models/{model_id}`

### Checklist


application/backend/app/models/model_revision.py
```diff
@@ -33,6 +33,19 @@ class TrainingInfo(BaseEntity):
     end_time: datetime | None = None
     dataset_revision_id: UUID | None = None
 
+    @model_validator(mode="before")
+    @classmethod
+    def populate_training_info(cls, data: object) -> object:
```
leoll2, line 44: This approach based on `model_validator` is structurally fragile, we'll keep encountering issues like the one in this ticket as long as we perform implicit conversions between the entities of different layers.
In this case, I suggest to initialize `TrainingInfo` explicitly, without relying on before validators. At L79 or somewhere else:
```python
return TrainingInfo(
    start_time=model_rev_db.start_time,
    ...
)
```

@itallix what do you think?

leoll2 (commented):
Manually tested, the training time fields work but dataset_revision_id is missing, as pointed out by Copilot

image

leoll2 (approved):

PR #5250, opened 2026-01-20
Cleanup unnecessary OTX entities

### LINKED ISSUE #5201: Cleanup unnecessary OTX entities
After Experimental Datumaro implementation (https://github.com/open-edge-platform/training_extensions/pull/4920) some refactoring is needed.

Recommended changes in scope of this ticket:

- Remove `OTXDataItem` since we are using `OTXSample` from Datumaro
  - Remove torch entity
  - Move / Combine `OTXDataBatch` / Predict with new `OTXSample`
  - Rename all docstrings and typing
- Remove OTX `ImageInfo`, since we can use now Datumaro Img Info
- Remove Polygons from the whole pipeline, since we don't use them anymore
- Fix unit tests accordingly (migrate from OTXDataItem to OTXSample, remove polygons, add new tests for OTXSample, etc.)

---

Other related refactoring tickets:
- #5208 
- #5209

AlbertvanHouten:
### Summary

Removed OTXDataItem/OTXPrediction/OTXBatch classes and replaced with datumaro samples 
Removed torch entity
Cleaned up ImageInfo class by removing unused attributes
Removed Polygons from the whole pipeline.

resolves #5201

### Checklist


library/tests/unit/data/dataset/test_keypoint_detection.py
```diff
@@ -44,7 +44,6 @@ def test_get_item_impl_subset(
         assert isinstance(entity.img_info, ImageInfo)
         assert hasattr(entity, "label")
         assert isinstance(entity.label, Tensor)
-        assert hasattr(entity, "bboxes")
         assert hasattr(entity, "keypoints")
```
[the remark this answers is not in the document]
AlbertvanHouten: This check previously worked due to inheritance from base sample class which defined many properties including bbox with None value. I have now removed those properties as they should be defined in the subclasses that need them.

library/src/otx/data/entity/sample.py
```diff
@@ -57,7 +58,16 @@ def flatten_fn(obj: object) -> tuple[list[Any], list[str]]:
         return (list(obj_dict.values()), list(obj_dict.keys()))
 
     def unflatten_fn(values: list[Any], context: list[str]) -> object:
-        return cls(**dict(zip(context, values)))
+        kwargs = dict(zip(context, values))
+        # Extract _img_info to set after construction (since __post_init__ would overwrite it)
+        img_info = kwargs.pop("_img_info", None)
```
leoll2, line 73: Where is the attribute `_img_info` first declared and initialized? I see that the `image_info` property tries to read it, but I can't find where it is initialized.

library/src/otx/data/entity/sample.py
```diff
@@ -57,7 +58,16 @@ def flatten_fn(obj: object) -> tuple[list[Any], list[str]]:
...
+        # Remove _transforms as it's not a constructor argument
+        kwargs.pop("_transforms", None)
```
leoll2: Same for `_transforms`, where is the declaration?

library/src/otx/data/entity/sample.py
```diff
@@ -264,3 +249,349 @@ def __post_init__(self) -> None:
...
+    """Collate OTXSamples into a batch.
+
+    Args:
+        samples: List of OTXSamples to batch
+
+    Returns:
+        Batched OTXSampleBatch with stacked tensors
+    """
+    # Check if all images have the same size. TODO(kprokofi): remove this check once OV IR models are moved.
+    if all(sample.image.shape == samples[0].image.shape for sample in samples):
```
leoll2, line 272: Is this check/TODO still relevant?

library/src/otx/data/entity/sample.py
```diff
@@ -264,3 +249,349 @@ def __post_init__(self) -> None:
...
+        batch_size: The number of samples in the batch.
+        images: The batch of images as a tensor or list of tensors.
+        labels: List of label tensors, optional.
+        masks: List of masks, optional.
+        bboxes: List of bounding boxes, optional.
+        keypoints: List of keypoint tensors, optional.
+        imgs_info: Sequence of image information, optional.
+    """
+
+    batch_size: int  # TODO(ashwinvaidya17): Remove this
```
leoll2: I guess this attribute can be removed if you add a `__len__` method that returns the length of `images`.

library/src/otx/data/entity/sample.py
```diff
@@ -264,3 +249,349 @@ def __post_init__(self) -> None:
...
+        images = [sample.image for sample in samples]
+
+    return OTXSampleBatch(
+        batch_size=len(samples),
+        images=images,
+        labels=[sample.label for sample in samples],
+        bboxes=[sample.bboxes for sample in samples],
+        keypoints=[sample.keypoints for sample in samples],
+        masks=[sample.masks for sample in samples],
+        imgs_info=[sample.img_info for sample in samples],
```
leoll2, line 284: Is `imgs_info` actually used anywhere? I see it was used in `library/src/otx/backend/openvino/models/instance_segmentation.py`, but that's now removed in this PR.

library/src/otx/data/entity/sample.py
```diff
@@ -264,3 +249,349 @@ def __post_init__(self) -> None:
...
+            msg = "Feature vector must have 2 dimensions"
+            raise ValueError(msg)
+
+    @property
+    def has_xai_outputs(self) -> bool:
+        """Check if the batch has XAI outputs.
+
+        Necessary for compatibility with tests.
+        """
+        # TODO(ashwinvaidya17): the tests should directly refer to saliency map.
```
leoll2: Can we remove this property, as suggested by the TODO?

library/src/otx/data/entity/sample.py
AlbertvanHouten: Its set in the post_init of the sample classes (there is a setter in the base sample class that sets `_img_info` specifically). The problem is that it is set from the value from datumaro's image info, which does not change, but throughout the augmentations OTX is updating the image info to reflect the current state of the image. I think it would be best to move this logic away from the sample classes but its out of scope for this PR.

library/src/otx/data/entity/sample.py
AlbertvanHouten: @kprokofi  is it needed?

library/src/otx/data/entity/sample.py
AlbertvanHouten: This can indeed be removed

library/src/otx/data/entity/sample.py
AlbertvanHouten: It is still used widely 

image

kprokofi: We use it to derive information about initial image size in post-processing stage for OD, IS and KP

library/src/otx/data/entity/sample.py
kprokofi: Please keep it for now, I will address it later

library/src/otx/data/entity/sample.py
```diff
@@ -264,3 +247,336 @@ def __post_init__(self) -> None:
...
+
+        # Generic handler for all other fields
+        for field in ["labels", "bboxes", "keypoints", "masks"]:
+            value = getattr(self, field)
+            if value is not None:
+                kwargs[field] = [maybe_wrap_tv(v) if v is not None else None for v in value]
+
+        return self.wrap(**kwargs)
+
+    def wrap(self, **kwargs) -> OTXSampleBatch:
```
kprokofi, line 364: wrap will trigger validation once again. We can rewrite it in a more smarter way:

```
def wrap(self, **kwargs) -> OTXSampleBatch:
    updated = object.__new__(self.__class__)
    updated.__dict__.update(asdict(self))
    updated.__dict__.update(kwargs)
    return updated
```

library/src/otx/data/entity/sample.py
```diff
@@ -264,3 +247,336 @@ def __post_init__(self) -> None:
...
+        """Get the number of samples in the batch."""
+        if isinstance(self.images, torch.Tensor):
+            return self.images.shape[0]
+        return len(self.images)
+
+    def __post_init__(self) -> None:
+        """Validate the batch after initialization."""
+        self._validate()
+
+    def _validate(self) -> None:
```
kprokofi, line 319: We incorporate the validation logic into batch entity. I feel this is not clean way. I would rather implemented it in the same way as we did for torch entities. In a separate Mixin class.

library/src/otx/data/entity/sample.py
```diff
@@ -264,3 +247,336 @@ def __post_init__(self) -> None:
...
+    """
+
+    scores: list[torch.Tensor] | None = None
+    feature_vector: list[torch.Tensor] | None = None
+    saliency_map: list[torch.Tensor] | None = None
+
+    def __post_init__(self) -> None:
+        """Validate the prediction batch after initialization."""
+        super().__post_init__()
+        if self.scores is not None:
```
kprokofi, line 398: Same here, I propose to separate the logic for validation and entities. Entities should be clean and well readable

library/src/otx/data/entity/base.py
```diff
@@ -64,10 +64,6 @@ class ImageInfo(tv_tensors.TVTensor):
         scale_factor: Scale factor (height, width) if the image is resized during preprocessing.
```
kprokofi, line 64: Can we use ImageInfo from datumaro? I know that we change them during augmentations. However, maybe it is possible. Basically we need to change img_shape and keep ori_shape. These fields can be updated to fit datumaro entities.

AlbertvanHouten: I don't think we should be mixing Datumaro and OTXs ImageInfo classes. OTX ImageInfo is inheriting from tv_tensors.TVTensor and has functionality to work with the transformations pipeline. If anything, it probably would make more sense to remove datumaro's imageinfo from the sample classes and use the OTX implementation completely.

library/src/otx/data/entity/sample.py
AlbertvanHouten: Moved it to a separate validation file.

library/src/otx/data/entity/sample.py
AlbertvanHouten: fixed

kprokofi (approved):

library/src/otx/data/entity/sample.py
kprokofi: I meant about Mixin Class instead of functions to not spawn imports. Do you think functions are better?

AlbertvanHouten: All these functions are standalone so adding a class would mostly be unnecessary boilerplate. We could import the entire module and call the functions from there if reducing the imports is important for this repo.

leoll2 (approved):

PR #5251, opened 2026-01-20
Integrate model variants API

jpggvilaca:
### Summary

And address previous PR's comments

Screenshot 2026-01-20 at 13 40 09
Screenshot 2026-01-20 at 13 40 14
Screenshot 2026-01-20 at 13 40 17

### Checklist

### DIFF
--- a/application/ui/src/features/models/hooks/api/use-get-model.hook.ts
+++ b/application/ui/src/features/models/hooks/api/use-get-model.hook.ts
@@ -11,7 +11,7 @@ export const useGetModel = (modelId: string | null | undefined) => {
     return $api.useQuery(
         'get',
         '/api/projects/{project_id}/models/{model_id}',
-        { params: { path: { project_id: projectId, model_id: modelId! } } },
-        { enabled: !!modelId }
+        { params: { path: { project_id: projectId, model_id: String(modelId) } } },
+        { enabled: Boolean(modelId) }
     );
 };
--- a/application/ui/src/features/models/model-listing/model-variants/model-variant-table.component.tsx
+++ b/application/ui/src/features/models/model-listing/model-variants/model-variant-table.component.tsx
@@ -3,7 +3,6 @@
 
 import {
     ActionButton,
-    Button,
     Cell,
     Column,
     Flex,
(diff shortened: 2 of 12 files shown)


application/ui/src/features/models/hooks/api/use-get-model.hook.ts
```diff
@@ -11,7 +11,7 @@ export const useGetModel = (modelId: string | null | undefined) => {
     return $api.useQuery(
```
dwesolow, line 11: Could we either not call `useGetModel` if there is no `modelId` or do early return in this hook, so whenever query gets called it has `modelId`?

application/ui/src/features/models/model-listing/model-variants/model-variant-table.component.tsx
```diff
@@ -17,61 +16,58 @@ import {
...
 import { DownloadIcon, MoreMenu } from '@geti/ui/icons';
 
-import type { SchemaModelView } from '../../../../api/openapi-spec';
+import { Model } from '../../../../constants/shared-types';
+import { formatModelSize } from '../utils/format-model-size';
 
 interface ModelVariantTableProps {
-    model: SchemaModelView;
+    model: Model;
+    format: 'openvino' | 'pytorch' | 'onnx';
```
dwesolow: Could we use `format` defined by the api?

application/ui/src/features/models/model-listing/utils/format-model-size.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright (C) 2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+import prettyBytes from 'pretty-bytes';
+
+export const formatModelSize = (bytes: number): string => prettyBytes(bytes);
```
dwesolow, line 6: What format we want use to display bytes, `kb` or `KB`? The difference is that for 2048 bytes, it will be `2 KB` (https://en.wikipedia.org/wiki/JEDEC_memory_standards) and `2,048 kB` (https://en.wikipedia.org/wiki/Metric_prefix). `prettyBytes` uses `si` format.

application/ui/src/features/models/hooks/api/use-get-model.hook.ts
jpggvilaca: I remember those rules but in this case no

application/ui/src/features/models/model-listing/utils/format-model-size.ts
jpggvilaca: i didnt think too much of it. But we can ask the backend team

application/ui/src/features/models/model-listing/model-variants/model-variant-table.component.tsx
jpggvilaca: Good point ,updated

camiloHimura (dismissed):

application/ui/src/features/models/hooks/api/use-get-model.hook.ts
dwesolow: Why not?

application/ui/src/features/models/model-listing/utils/format-model-size.ts
dwesolow: It's better to ask, I remember having that issue in geti years ago

application/ui/src/features/models/model-listing/model-variants/model-variant-table.component.tsx
```diff
@@ -17,61 +16,58 @@ import {
...
 interface ModelVariantTableProps {
-    model: SchemaModelView;
+    model: Model;
+    format: ModelFormat;
 }
 
-export const ModelVariantTable = ({ model }: ModelVariantTableProps) => {
-    // TODO: Replace with dynamic data
+export const ModelVariantTable = ({ model, format }: ModelVariantTableProps) => {
+    const variants = (model.variants ?? []).filter((variant) => variant.format.toLowerCase() === format);
```
dwesolow: Do we still need to lowercase format? Probably not since we use format supported by the api.

jpggvilaca: Nope, not anymore ^^ updated

application/ui/src/features/models/model-listing/utils/format-model-size.ts
jpggvilaca: I asked, SI is the format, so all good

application/ui/src/features/models/hooks/api/use-get-model.hook.ts
jpggvilaca: We could if we moved the hook inside the dumb component, but i'd rather keep it like that, since this container is responsible for this. So, it's not that we can't, i just think we shouldnt

dwesolow (approved):

Issue #5043, opened 2025-11-27
Add UI selector for the inference device


leoll2:
Add the UI element to select the inference device.

To get a list of available devices, refer to #5041.

To change the device, refer to #5042 / #5061 

<img width="609" height="449" alt="Image" src="https://github.com/user-attachments/assets/3895b024-7218-4e5c-aaa2-64c3d295be08" />

PR #5252, opened 2026-01-20
Inference device selector

camiloHimura:
#5043

### Summary

https://github.com/user-attachments/assets/e714f9cf-fd70-4bf8-8c3b-537750bbf8e3

### Checklist

### DIFF
--- a/application/ui/src/features/inference/header/active-model.component.tsx
+++ b/application/ui/src/features/inference/header/active-model.component.tsx
@@ -2,17 +2,12 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { Flex, Text } from '@geti/ui';
-import { useProjectIdentifier } from 'hooks/use-project-identifier.hook';
 import { isEmpty } from 'lodash-es';
 
-import { $api } from '../../../api/client';
+import { useGetActiveModelArchitectureId } from '../../models/hooks/api/use-get-active-model-architecture-id.hook';
 
 export const ActiveModel = () => {
-    const projectId = useProjectIdentifier();
-
-    const modelsQuery = $api.useSuspenseQuery('get', '/api/projects/{project_id}/models', {
-        params: { path: { project_id: projectId } },
-    });
+    const activeModelArchitectureId = useGetActiveModelArchitectureId();
 
     return (
         <Flex gap='size-50' alignItems='center'>
@@ -28,7 +23,7 @@ export const ActiveModel = () => {
                     color: 'var(--spectrum-global-color-gray-700)',
                 }}
             >
-                {!isEmpty(modelsQuery.data) ? modelsQuery.data[0].architecture : 'Unknown'}
+                {isEmpty(activeModelArchitectureId) ? 'Unknown' : activeModelArchitectureId}
             </Text>
         </Flex>
     );
--- a/application/ui/src/features/inference/header/active-model.component.tsx
+++ b/application/ui/src/features/inference/header/active-model.component.tsx
@@ -1,0 +1,35 @@
+// Copyright (C) 2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+import { Flex, Text } from '@geti/ui';
+import { useProjectIdentifier } from 'hooks/use-project-identifier.hook';
+import { isEmpty } from 'lodash-es';
+
+import { $api } from '../../../api/client';
+
+export const ActiveModel = () => {
+    const projectId = useProjectIdentifier();
+
+    const modelsQuery = $api.useSuspenseQuery('get', '/api/projects/{project_id}/models', {
+        params: { path: { project_id: projectId } },
+    });
+
+    return (
+        <Flex gap='size-50' alignItems='center'>
+            <Text
+                UNSAFE_style={{
+                    color: 'var(--spectrum-global-color-gray-900)',
+                }}
+            >
+                Model:
+            </Text>
+            <Text
+                UNSAFE_style={{
+                    color: 'var(--spectrum-global-color-gray-700)',
+                }}
+            >
+                {!isEmpty(modelsQuery.data) ? modelsQuery.data[0].architecture : 'Unknown'}
+            </Text>
+        </Flex>
+    );
+};
--- a/application/ui/src/features/inference/header/inference-devices.component.tsx
+++ b/application/ui/src/features/inference/header/inference-devices.component.tsx
@@ -13,7 +13,7 @@ export const InferenceDevices = () => {
     const { data: devices } = $api.useSuspenseQuery('get', '/api/system/devices/inference');
     const { data: pipeline } = usePipeline();
     const projectId = useProjectIdentifier();
-    const [selectedKey, setSelectedKey] = useState<Key | null>(pipeline.device.toLowerCase() ?? null);
+    const [selectedKey, setSelectedKey] = useState<Key | null>(pipeline.device);
 
     const updatePipeline = $api.useMutation('patch', '/api/projects/{project_id}/pipeline', {
         meta: {
@@ -22,7 +22,7 @@ export const InferenceDevices = () => {
         onError: (error) => {
             if (error) {
                 toast({ type: 'error', message: String(error.detail) });
-                setSelectedKey(pipeline.device.toLowerCase() ?? null);
+                setSelectedKey(pipeline.device);
             }
         },
     });
@@ -43,7 +43,7 @@ export const InferenceDevices = () => {
 
     return (
         <Picker
-            maxWidth={'size-3000'}
+            maxWidth='size-3000'
             label='Inference Compute: '
             aria-label='inference compute'
             labelAlign='end'
--- a/application/ui/src/features/inference/header/inference-devices.component.tsx
+++ b/application/ui/src/features/inference/header/inference-devices.component.tsx
@@ -1,0 +1,58 @@
+// Copyright (C) 2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+import { useState } from 'react';
+
+import { Item, Key, Picker, toast } from '@geti/ui';
+
+import { $api } from '../../../api/client';
+import { usePipeline } from '../../../hooks/api/pipeline.hook';
+import { useProjectIdentifier } from '../../../hooks/use-project-identifier.hook';
+
+export const InferenceDevices = () => {
+    const { data: devices } = $api.useSuspenseQuery('get', '/api/system/devices/inference');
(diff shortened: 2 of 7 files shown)


application/ui/src/features/inference/header/inference-devices.component.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+import { $api } from '../../../api/client';
+import { usePipeline } from '../../../hooks/api/pipeline.hook';
+import { useProjectIdentifier } from '../../../hooks/use-project-identifier.hook';
+
+export const InferenceDevices = () => {
+    const { data: devices } = $api.useSuspenseQuery('get', '/api/system/devices/inference');
+    const { data: pipeline } = usePipeline();
+    const projectId = useProjectIdentifier();
+    const [selectedKey, setSelectedKey] = useState<Key | null>(pipeline.device.toLowerCase() ?? null);
```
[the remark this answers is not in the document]
camiloHimura: updated

application/ui/src/features/inference/header/inference-devices.component.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+    const handleChange = (key: Key | null) => {
+        if (key === null) {
+            return;
+        }
+
+        setSelectedKey(key);
+        updatePipeline.mutate({
+            params: { path: { project_id: projectId } },
+            body: { device: key },
```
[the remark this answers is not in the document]
camiloHimura: updated

application/ui/src/features/inference/header/inference-devices.component.tsx
```diff
@@ -0,0 +1,58 @@
...
+        });
+    };
+
+    return (
+        <Picker
+            maxWidth={'size-3000'}
```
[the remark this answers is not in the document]
camiloHimura: updated

application/ui/src/features/inference/header/active-model.component.tsx
```diff
@@ -0,0 +1,35 @@
...
+                }}
+            >
+                Model:
+            </Text>
+            <Text
+                UNSAFE_style={{
+                    color: 'var(--spectrum-global-color-gray-700)',
+                }}
+            >
+                {!isEmpty(modelsQuery.data) ? modelsQuery.data[0].architecture : 'Unknown'}
```
jpggvilaca: the first one might not be the active one. Let's use the hook that we have to get the active model instead

application/ui/src/features/inference/header/inference-devices.component.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { usePipeline } from '../../../hooks/api/pipeline.hook';
+import { useProjectIdentifier } from '../../../hooks/use-project-identifier.hook';
+
+export const InferenceDevices = () => {
+    const { data: devices } = $api.useSuspenseQuery('get', '/api/system/devices/inference');
+    const { data: pipeline } = usePipeline();
+    const projectId = useProjectIdentifier();
+    const [selectedKey, setSelectedKey] = useState<Key | null>(pipeline.device);
+
+    const updatePipeline = $api.useMutation('patch', '/api/projects/{project_id}/pipeline', {
```
jpggvilaca: np we can use `usePatchPipeline` from pipeline hooks

jpggvilaca (dismissed):

application/ui/src/features/inference/header/active-model.component.tsx
camiloHimura: updated

application/ui/src/features/inference/header/inference-devices.component.tsx
camiloHimura: updated

application/ui/src/features/inference/header/active-model.component.tsx
```diff
@@ -0,0 +1,30 @@
...
+                }}
+            >
+                Model:
+            </Text>
+            <Text
+                UNSAFE_style={{
+                    color: 'var(--spectrum-global-color-gray-700)',
+                }}
+            >
+                {isEmpty(activeModelArchitectureId) ? 'Unknown' : activeModelArchitectureId}
```
dwesolow, line 26: When is it possible to have unknown model?

application/ui/src/features/inference/header/inference-devices.component.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { useState } from 'react';
+
+import { Item, Key, Picker, toast } from '@geti/ui';
+
+import { $api } from '../../../api/client';
+import { usePatchPipeline, usePipeline } from '../../../hooks/api/pipeline.hook';
+import { useProjectIdentifier } from '../../../hooks/use-project-identifier.hook';
+
+export const InferenceDevices = () => {
+    const { data: devices } = $api.useSuspenseQuery('get', '/api/system/devices/inference');
```
dwesolow, line 13: Could we extract api hooks to `api` dir? So we don't mix responsibilities of api hooks with component

application/ui/src/features/inference/header/inference-devices.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+            http.get('/api/system/devices/inference', () => HttpResponse.json(devicesResponse)),
+            http.get('/api/projects/{project_id}/pipeline', () => HttpResponse.json(pipelineResponse)),
+            http.patch('/api/projects/{project_id}/pipeline', () => {
+                pipelinePatchSpy();
+                return HttpResponse.json({}, { status });
+            })
+        );
+
+        render(
+            <TestProviders client={new QueryClient()}>
```
dwesolow, line 40: Why new client needs to be passed?

application/ui/src/features/inference/header/inference-devices.component.tsx
dwesolow, line 13: We use suspense query but we don't have a suspense boundary around this component which will remount this and probably more components.

application/ui/src/features/inference/header/inference-devices.component.tsx
```diff
@@ -0,0 +1,57 @@
...
+    const { data: pipeline } = usePipeline();
+    const projectId = useProjectIdentifier();
+    const [selectedKey, setSelectedKey] = useState<Key | null>(pipeline.device);
```
dwesolow, line 16: Is it possible that `pipeline.device` is not part of `devices`? (it might be the edge case but it'd be better to double check)

application/ui/src/features/inference/header/inference-devices.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+import userEvent from '@testing-library/user-event';
+import { getMockedPipeline } from 'mocks/mock-pipeline';
+import { HttpResponse } from 'msw';
+import { TestProviders } from 'test-utils/render';
+
+import { http } from '../../../api/utils';
+import { server } from '../../../msw-node-setup';
+import { InferenceDevices } from './inference-devices.component';
+
+vi.mock('hooks/use-project-identifier.hook', () => ({ useProjectIdentifier: () => '123' }));
```
dwesolow, line 15: For later: we can use test renderer from prompt where we specify the url (together with project id) so we would not need to mock `useProjectIdentifier`.

application/ui/src/features/inference/header/inference-devices.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+    it('updates device on selection change', async () => {
+        const pipelinePatchSpy = renderApp();
+
+        await screen.findByLabelText('inference compute');
+
+        await userEvent.click(screen.getByRole('button', { name: /inference compute/i }));
+        await userEvent.click(screen.getByRole('option', { name: /GPU/i }));
+
+        await waitFor(() => {
+            expect(pipelinePatchSpy).toHaveBeenCalled();
```
dwesolow, line 67: Instead of checking if the spy is called it'd be more appropriate to check what is the boy of the `patch` request.

application/ui/src/features/inference/header/active-model.component.tsx
camiloHimura: I’m not sure, I just moved the component to its own file, and I don’t know why it was implemented that way originally.

application/ui/src/features/inference/header/inference-devices.component.tsx
camiloHimura: why?

application/ui/src/features/inference/header/inference-devices.test.tsx
camiloHimura: It was reusing the same query configuration and returning responses from the preview test configurations.

application/ui/src/features/inference/header/inference-devices.test.tsx
jpggvilaca: Context: https://github.com/open-edge-platform/geti-prompt/blob/main/application/ui/src/test-utils/render.tsx

jpggvilaca (approved):

application/ui/src/features/inference/header/inference-devices.test.tsx
camiloHimura: Perhaps, but not long ago I spent some time trying to make that assertion work, and it turned out to be quite problematic. Because of that, I went with the easiest approach.

application/ui/src/features/inference/header/inference-devices.component.tsx
camiloHimura: No, that’s not possible. And even if it were, I believe this would be a backend issue rather than something on our side.

application/ui/src/features/inference/header/inference-devices.component.tsx
camiloHimura: I'll handle it on my next PR

application/ui/src/features/inference/header/inference-devices.test.tsx
dwesolow: I did that approach a lot in prompt and I haven't noticed any issue. What was the issue that you got?

application/ui/src/features/inference/header/inference-devices.component.tsx
dwesolow: To not mix responsibilities.

application/ui/src/features/inference/header/inference-devices.test.tsx
dwesolow: If that's the case we should not fix the problem locally, but for each case (always pass new query client inside the test providers)

PR #5178, opened 2026-01-13
Enable Pyrefly in Geti Tune

leoll2:
### Summary

This PR enables Pyrefly as a static type checker for the Python code in Geti Tune (application), replacing mypy. 
CI workflows and pre-commit checks are updated accordingly.

Resolves #5169

### Checklist

- [x] The PR title and description are clear and descriptive
- [x] I have manually tested the changes
- [x] All changes are covered by automated tests
- [x] All related issues are linked to this PR (if applicable)
- [ ] Documentation has been updated (if applicable)

PR #5253, opened 2026-01-20
Enable Pyrefly in OTX

### LINKED ISSUE #5006: Enable static type-checker in OTX library
mypy detects >1k typing violations In `library/otx`. These errors were previously undetected due to a misconfiguration of pre-commit, that would cause mypy not to follow some imports.

To show all the mypy errors, run:
```shell
cd library
uv run mypy .
```

The goal of this ticket is to re-enable a static type-checker (mypy, pyrefly or equivalent) in CI as soon as possible, then gradually resolve the errors.

leoll2:
### Summary

Enable Pyrefly in the `library/` code as a replacement for mypy. This PR is complementary to #5178, which did the same for `application/backend/`. Since OTX has hundreds of pre-existing type warnings, we exploit a [baseline file](https://pyrefly.org/en/docs/error-suppressions/#baseline-files-experimental) to enable the tool while suppressing such errors; hopefully, as the code gets refactored, we can remove most of the typing violations.

Resolves #5006 

### How to test

- Type-checking without baseline: `uv run pyrefly check`
- Type-checking with baseline: `uv run pyrefly check --baseline=pyrefly-baseline.json`
- Update baseline `uv run pyrefly check --baseline=pyrefly-baseline.json --update-baseline`

### Checklist


AlbertvanHouten:
I cannot see the content of the baseline file in this PR without checking out the code but does it take exact line numbers or does it use a smarter approach? I am wondering what happens if we add one new line at the top of an existing file, would that cause the baseline to still see the original errors or do all errors in that file now need to be fixed?

leoll2:
> I cannot see the content of the baseline file in this PR without checking out the code but does it take exact line numbers or does it use a smarter approach? I am wondering what happens if we add one new line at the top of an existing file, would that cause the baseline to still see the original errors or do all errors in that file now need to be fixed?

According to the [docs](https://pyrefly.org/en/docs/error-suppressions/#baseline-files-experimental):
> Errors are matched with the baseline by looking at file, error code, and column number.

But I tried to modify one of the files with pre-existing issues, and the errors were still suppressed even though the problematic code was shifted down by a few lines. Which makes me think that Pyrefly implements a smarter mechanism than exact matching of row/column.

AlexanderBarabanov (dismissed):

kprokofi (dismissed):

kprokofi (approved):

AlexanderBarabanov (approved):

itallix (approved):

PR #5256, opened 2026-01-21
Fix run.sh for macOS

leoll2:
### Summary

Fix an issue with `run.sh` that occurs on macOS platforms:

> ./run.sh: line 58: data: unbound variable

The issue is caused by the `set -euo pipefail` at the beginning of the script. The -u flag treats unset variables as errors, and on macOS, the associative array expansion "${!VIDEO_URLS[@]}" may behave differently when the array is empty or not yet initialized.

### How to test

`DOWNLOAD_FILES=true ./run.sh`

### Checklist


leoll2:
Still fails on bash 3.x, the default for macOS. Since that bash 3.2 is almost 20 years old, I recommend macOS users to manually install a newer bash.

PR #5290, opened 2026-01-22
API integration pt3

jpggvilaca:
### Summary

* Prepare for model details integration
* Integrate model/model-variant download
* Fix sortBy default value
* Remove model variant deletion (not supported)
* Address comment from previous PR

### Checklist

### DIFF
--- a/application/ui/package-lock.json
+++ b/application/ui/package-lock.json
@@ -7470,9 +7470,9 @@
             }
         },
         "node_modules/@remix-run/router": {
-            "version": "1.23.0",
-            "resolved": "https://registry.npmjs.org/@remix-run/router/-/router-1.23.0.tgz",
-            "integrity": "sha512-O3rHJzAQKamUz1fvE0Qaw0xSFqsA/yafi2iqeE0pvdFtCO1viYx8QL6f3Ln/aCCTLxs68SLf0KPM9eSeM8yBnA==",
+            "version": "1.23.2",
+            "resolved": "https://registry.npmjs.org/@remix-run/router/-/router-1.23.2.tgz",
+            "integrity": "sha512-Ic6m2U/rMjTkhERIa/0ZtXJP17QUi2CbWE7cqx4J58M8aA3QTfW+2UlQ4psvTX9IO1RfNVhK3pcpdjej7L+t2w==",
             "license": "MIT",
             "engines": {
                 "node": ">=14.0.0"
@@ -14110,9 +14110,9 @@
             }
         },
         "node_modules/lodash": {
-            "version": "4.17.21",
-            "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.21.tgz",
-            "integrity": "sha512-v2kDEe57lecTulaDIuNTPy3Ry4gLGJ6Z1O3vE1krgXZNrsQ+LFTGHVxVjcXPs17LhbZVGedAJv8XZ1tvj5FvSg==",
+            "version": "4.17.23",
+            "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.23.tgz",
+            "integrity": "sha512-LgVTMpQtIopCi79SJeDiP0TfWi5CNEc/L/aRdTh3yIvmZXTnheWpKjSZhnvMl8iXbC1tFg9gdHHDMLoV7CnG+w==",
             "license": "MIT"
         },
         "node_modules/lodash-es": {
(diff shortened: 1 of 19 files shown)


application/ui/src/features/models/hooks/api/use-download-model.hook.ts
```diff
@@ -0,0 +1,52 @@
...
+            const blob = data as Blob;
+            const url = URL.createObjectURL(blob);
+            const format = variables.params.query?.format;
+
+            const link = document.createElement('a');
+
+            link.href = url;
+            link.download = format ? `model-${modelId}-${format}.zip` : `model-${modelId}.zip`;
+
+            document.body.appendChild(link);
```
dwesolow: I think we don't need to append it to body.

jpggvilaca: true. Thanks! removed

application/ui/src/features/models/model-listing/components/group-models-container/group-models-container.component.tsx
```diff
@@ -26,23 +26,29 @@ export const GroupModelsContainer = ({ group, models }: GroupModelsContainerProp
...
-                >
-                    <DisclosureTitle UNSAFE_className={classes.disclosureItem}>
-                        <ModelRowContainer model={model} />
-                    </DisclosureTitle>
-                    <DisclosurePanel>
-                        <ModelDetailsTabs model={model} />
-                    </DisclosurePanel>
-                </Disclosure>
-            ))}
+            {models.map((model) => {
```
dwesolow, line 29: np. we could just filter models that has ids (I still don't understand why model might not have the id 😢 )

jpggvilaca: I asked the backend to make it required (tune sync channel)

application/ui/src/features/models/model-listing/model-variants/model-variant-table.component.tsx
```diff
@@ -51,19 +42,14 @@ export const ModelVariantTable = ({ model, format }: ModelVariantTableProps) =>
                         <Cell>{formatModelSize(variant.weights_size)}</Cell>
                         <Cell>
                             <Flex gap={'size-100'} justifyContent='end' alignItems='center'>
-                                <ActionButton isQuiet aria-label={`Download ${variant.format} model`}>
+                                <ActionButton
+                                    isQuiet
+                                    aria-label={`Download ${variant.format} model`}
+                                    isDisabled={isDownloading}
+                                    onPress={() => downloadModel(variant.format as ModelFormat)}
```
dwesolow, line 48: We casting is needed? I'd expect this to have the same format type.

jpggvilaca: Likewise, but TS says no :D

application/ui/src/features/models/model-listing/components/group-models-container/group-models-container.component.tsx
dwesolow: Could we update it in the UI then? If we're sure we always have it, let's update UI type and once they fix it we can remove our typing. It's sad that we need to do `model.id ?? ''` in lot's of places only to make ts happy.
Thoughts?

jpggvilaca: i didnt want to sidetrack too much from the source of truth. I'd rather wait for them to do it, and then i delete all of this in bulk : /

dwesolow: It would be easier to update one place instead of many places where we need `model`, but it's up to you.
I already had to define types on my own for advanced settings because they don't support them yet.

application/ui/package-lock.json
dwesolow, line 1: why this change?

jpggvilaca: Oh, it was the bonus commit, `npm audit` to fix these 4 issues. Also an excuse to trigger the build. But GHA seem to be not working rn

application/ui/src/features/models/model-listing/components/group-models-container/group-models-container.component.tsx
jpggvilaca: Updated

application/ui/src/features/models/hooks/api/use-get-model.hook.ts
```diff
@@ -1,17 +1,19 @@
...
 // SPDX-License-Identifier: Apache-2.0
 
+import { UseQueryResult } from '@tanstack/react-query';
 import { useProjectIdentifier } from 'hooks/use-project-identifier.hook';
 
 import { $api } from '../../../../api/client';
+import type { ExtendedModel } from '../../../../constants/shared-types';
 
-export const useGetModel = (modelId: string | null | undefined) => {
+export const useGetModel = (modelId: string | null | undefined): UseQueryResult<ExtendedModel> => {
```
dwesolow, line 10: After you type change, can we remove `null | undefined` from here?

application/ui/src/features/models/hooks/api/use-get-models.hook.ts
```diff
@@ -1,14 +1,16 @@
...
 // SPDX-License-Identifier: Apache-2.0
 
+import { UseQueryResult } from '@tanstack/react-query';
 import { useProjectIdentifier } from 'hooks/use-project-identifier.hook';
 
 import { $api } from '../../../../api/client';
+import type { Model } from '../../../../constants/shared-types';
 
-export const useGetModels = () => {
+export const useGetModels = (): UseQueryResult<Model[]> => {
```
dwesolow, line 10: could you add a TODO comment to remember to remove that once API is updated?

dwesolow (approved):

application/ui/src/features/models/hooks/api/use-get-models.hook.ts
jpggvilaca: I will, added locally on my next PR

application/ui/src/features/models/hooks/api/use-get-model.hook.ts
jpggvilaca: Agree, will do on my next PR

PR #5299, opened 2026-01-23
Bug fixing after triage

jpggvilaca:
### Summary

* Fix project creation when picking a color from the picker
* Show error notification when trying to delete self project
* Fix layout issue with sidebar open; Update pipeline icon
* Fix annotations persisting when changing media items
* Increase train model dialog width
* Current model training -> add polling
* Add TODO
* Update MutationCache to trigger toast by default onError

### Checklist


application/ui/src/features/annotator/label-utils.ts
```diff
@@ -0,0 +1,30 @@
+// Copyright (C) 2025 Intel Corporation
```
jpggvilaca, line 1: copied file from Classic without the extra utils

application/ui/src/shared/annotator/annotation-actions-provider.component.tsx
```diff
@@ -82,11 +82,9 @@ export const AnnotationActionsProvider = ({
 
         const localAnnotations = mapServerAnnotationsToLocal(initialAnnotationsDTO, projectLabels);
```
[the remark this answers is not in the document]
jpggvilaca: The condition if (localAnnotations.length > 0) means that when switching to a media item with NO annotations (empty array), the reset is never called, so the previous annotations persist

application/ui/src/utils.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright (C) 2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+import { isObject, isString } from 'lodash-es';
+
+export const isErrorWithDetail = (error: unknown): error is { detail: string } => {
```
camiloHimura: :)

camiloHimura (dismissed):

application/ui/src/components/project-panel/projects-list.component.tsx
```diff
@@ -58,6 +59,15 @@ export const ProjectsList = ({ projects, setProjectInEdition, projectIdInEdition
                         navigate(paths.project.index({}));
                     }
                 },
+                onError: (error) => {
+                    let message = 'Failed to delete project';
```
dwesolow: I think we should always use server error that actually indicates why operation failed.
What about introducing https://github.com/open-edge-platform/geti-prompt/blob/main/application/ui/src/query-client/query-client.ts#L93? IMO that's the way we should follow.

application/ui/src/features/annotator/label-utils.ts
dwesolow, line 1: np. `label-colors.ts`

application/ui/src/hooks/api/jobs.hook.ts
```diff
@@ -19,10 +19,18 @@ export const useListJobs = () => {
 
 export const useGetCurrentTrainingJob = () => {
     const projectId = useProjectIdentifier();
-    const activeJobs = useListJobs();
+    const activeJobs = $api.useQuery('get', '/api/jobs', undefined, {
+        refetchInterval: (query) => {
+            const hasActiveJob = query.state.data?.some(
+                (job) => job.status === 'RUNNING' || job.status === 'PENDING' || job.status === 'CANCELLING'
```
dwesolow: I think we should check also the type of the job, right (it must be `job_type === 'train'`)?
Why do we care about cancelled job?

application/ui/src/shared/annotator/annotation-actions-provider.component.tsx
```diff
@@ -82,11 +82,9 @@ export const AnnotationActionsProvider = ({
...
 
-        if (localAnnotations.length > 0) {
-            undoRedoActions.reset(localAnnotations);
-        }
+        undoRedoActions.reset(localAnnotations);
         // eslint-disable-next-line react-hooks/exhaustive-deps
-    }, [initialAnnotationsDTO, project?.task?.labels]);
+    }, [initialAnnotationsDTO, project?.task?.labels, mediaItem.id]);
```
dwesolow: TBH I don't like this hack, no code inside that effect uses `mediaItem.id`, why we should put that to the `useEffect` (I know the answer right now, but on Monday I won't remember that). I think we should follow the approach with the `key` that we discussed yesterday.

application/ui/src/hooks/api/jobs.hook.ts
jpggvilaca: Because cancelling is not immediate

application/ui/src/features/annotator/label-utils.ts
jpggvilaca: we could have more utils inside, not just color related

application/ui/src/components/project-panel/projects-list.component.tsx
jpggvilaca: The reason for this is that the backend doesnt always send a meaningful message. Remember on Geti classic and those long id's and cryptic messages? I wanted to avoid that. But indeed, let's add this anyway, ill notify the BE to update their error messages

application/ui/src/hooks/api/jobs.hook.ts
dwesolow: Yeah understand but do we want to show cancelled job in the models page?

application/ui/src/components/project-panel/projects-list.component.tsx
dwesolow: Yeah, we should force BE to always send well defined error messages, then we don't need to care about them :)

application/ui/src/hooks/api/jobs.hook.ts
jpggvilaca: We want them to disappear from the list. It will only happen if you refresh the page or if the current job is undefined. Which will be after a while

jpggvilaca: Actually i was already refresh /jobs onCancel success. So all good, removed

application/ui/src/shared/annotator/annotation-actions-provider.component.tsx
jpggvilaca: the `key` also wont work. But now that i removed the `if` condition, this is not necessary. removed

dwesolow: `key` always works 🏄

application/ui/src/features/inference/sinks/sink-list/sink-menu/sink-menu.component.tsx
```diff
@@ -21,12 +21,14 @@ export const SinkMenu = ({ id, name, isConnected, onEdit }: SinkMenuProps) => {
     const removeSink = $api.useMutation('delete', '/api/sinks/{sink_id}', {
         meta: {
             invalidateQueries: [['get', '/api/sinks']],
+            errorMessage: `Failed to remove "${name}".`,
```
dwesolow: I thought that we agreed to use server errors, no?

application/ui/src/shared/annotator/annotation-actions-provider.component.tsx
jpggvilaca: I mean the refresh caused by the key change wont work ^^ , the removal of the `if` did the rtick

application/ui/src/features/inference/sinks/sink-list/sink-menu/sink-menu.component.tsx
dwesolow: why `mutateAsync` and not `mutate`? We don't need `async` here, we can just use `onSuccess`

application/ui/src/utils.ts
dwesolow: No longer user, right?

jpggvilaca: used on error-page.tsx

application/ui/src/features/inference/sinks/sink-list/sink-menu/sink-menu.component.tsx
jpggvilaca: true, let me update tests then.

application/ui/src/features/inference/sinks/sink-list/sink-menu/sink-menu.component.tsx
jpggvilaca: I did not create this, but sure, also works. Updated

application/ui/src/utils.ts
dwesolow: if it's used only in one place then it's ok to remove that file?

application/ui/src/features/inference/sinks/sink-list/sink-menu/sink-menu.component.tsx
```diff
@@ -46,37 +46,37 @@ export const SinkMenu = ({ id, name, isConnected, onEdit }: SinkMenuProps) => {
 
     const handleConnect = async () => {
         try {
```
dwesolow, line 48: try/catch is no longer needed

application/ui/src/features/inference/sources/source-menu/source-menu.component.tsx
dwesolow, line 45: For later, whenever we have time

application/ui/src/features/inference/sinks/sink-list/sink-menu/sink-menu.component.tsx
jpggvilaca: next PR ^^

dwesolow (approved):

PR #5318, opened 2026-01-26
Fix error 500 when loading dataset revision items

leoll2:
### Summary

Changes:
- Initialize dataset revision item `id` from the filename, since it matches the id by construction
- Remove unnecessary property `name` from the dataset revision items

This PR fixes a 500 Internal Server Error when loading the item:
```
  File "/home/leoll2/training_extensions/application/backend/app/api/routers/dataset_revisions.py", line 49, in list_dataset_revision_items
    dataset_revision_items, total_count = dataset_revision_service.list_dataset_revision_items(
                                          │                        └ <function DatasetRevisionService.list_dataset_revision_items at 0x781a77297060>
                                          └ <app.services.dataset_revision_service.DatasetRevisionService object at 0x781a4a0126c0>

  File "/home/leoll2/training_extensions/application/backend/app/services/dataset_revision_service.py", line 182, in list_dataset_revision_items
    "id": item["id"],
          └ {'image': 'data/projects/9d6af8e8-6017-4ebe-9126-33aae739c5fa/dataset/3f8665c1-de48-4fa6-9daf-245a2ff4f173.jpg', 'image_info'...

KeyError: 'id'
```

### How to test

Train a model, then view the training dataset.

### Checklist

### DIFF
--- a/application/backend/app/services/dataset_revision_service.py
+++ b/application/backend/app/services/dataset_revision_service.py
@@ -179,12 +179,11 @@ class DatasetRevisionService(BaseSessionManagedService):
             dataset_revision_items.append(
                 DatasetRevisionItem.model_validate(
                     {
-                        "id": item["id"],
-                        "name": Path(item["image"]).stem,
+                        "id": UUID(Path(item["image"]).stem),
                         "format": Path(item["image"]).suffix.lstrip("."),
                         "width": item["image_info"]["width"],
                         "height": item["image_info"]["height"],
-                        "subset": item["subset"],
+                        "subset": DatasetItemSubset[item["subset"]],
                     }
                 )
             )
--- a/application/backend/app/api/schemas/dataset_item.py
+++ b/application/backend/app/api/schemas/dataset_item.py
@@ -6,11 +6,11 @@ from uuid import UUID
 
 from pydantic import BaseModel
 
-from app.core.models import BaseRequiredIDModel, BaseRequiredIDNameModel, Pagination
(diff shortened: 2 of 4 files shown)


application/backend/app/services/dataset_revision_service.py
```diff
@@ -179,12 +179,11 @@ def list_dataset_revision_items(
             dataset_revision_items.append(
                 DatasetRevisionItem.model_validate(
                     {
-                        "id": item["id"],
-                        "name": Path(item["image"]).stem,
+                        "id": UUID(Path(item["image"]).stem),
```
JortBergfeld, line 182: Would it not make sense to add the id to the dataframe? I find it rather hard to properly code based on implicit assumptions on how the parquet file is created, which can only be understood by reading the actual code. I ran into a similar problem testing the new dataset revision endpoint, which required me to understand these implicit assumptions.

JortBergfeld (approved):
Looks good to me. However, I think we should have a more robust and explicit way of generating these filenames if we depend on them for id generation.

application/backend/app/services/dataset_revision_service.py
leoll2: I agree, there are a couple of things that I don't like about the current format: one is the lack of an explicit id column, as you pointed out, and the other is the absolute image path which introduces portability problems. I want to merge this PR first to fix the bug, then I'll review the dataframe and propose some improvements.

Issue #5245, opened 2026-01-20
Annotator for classification task


jpggvilaca:
Since we did not have a classification project on the seeding list, it was completely forgotten. We need to make sure the annotator works properly for that type of project, namely assignment of labels (single & multi)

PR #5329, opened 2026-01-27
Classification project

camiloHimura:
### Summary
#5245

image

### Checklist


application/ui/src/components/zoom/use-sync-zoom.hook.tsx
```diff
@@ -8,7 +8,7 @@ import { useSetZoom } from './zoom.provider';
 
 export type Size = { width: number; height: number };
 
-const DEFAULT_SCREEN_ZOOM = 1;
+const DEFAULT_SCREEN_ZOOM = 0.9;
```
jpggvilaca, line 11: why?

application/ui/src/components/zoom/zoom.test.tsx
```diff
@@ -15,7 +15,7 @@ describe('Zoom', () => {
     it('Scales and translates content so that it fits the screen', () => {
         const screenSize = { width: 500, height: 500 };
         const contentSize = { width: 500, height: 500 };
-        const expectedZoom = { translate: { x: 0, y: 0 }, scale: 1 };
+        const expectedZoom = { translate: { x: 25, y: 25 }, scale: 0.9 };
```
jpggvilaca, line 19: same here

application/ui/src/components/zoom/use-sync-zoom.hook.tsx
camiloHimura: so the labels at the top are visible

application/ui/src/components/zoom/zoom.test.tsx
camiloHimura: Since this is no longer the initial zoom, the values also need to be updated, in this case, to `0.9` and `25`

application/ui/src/features/annotator/annotations/annotation-shape-with-labels.component.tsx
```diff
@@ -39,8 +55,7 @@ export const AnnotationShapeWithLabels = ({ annotation }: AnnotationShapeProps)
         );
     }
```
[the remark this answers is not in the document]
camiloHimura: The shapes are validated on lines 40 and 49.

application/ui/src/features/dataset/media-preview/secondary-toolbar/util.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright (C) 2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+import { Label } from '../../../../constants/shared-types';
+
+export const toggleLabel = (newLabel: Label, labels: Label[]): Label[] => {
+    const hasNewLabel = labels.some(({ id }) => id === newLabel.id);
```
dwesolow: I think u meant `hasExistingLabel`. When there is a new label we append that (line 13)

application/ui/src/hooks/use-selected-project.hook.tsx
```diff
@@ -0,0 +1,17 @@
+// Copyright (C) 2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+import { useProjects } from './api/project.hook';
+import { useProjectIdentifier } from './use-project-identifier.hook';
+
+export const useSelectedProject = () => {
+    const { data } = useProjects();
```
dwesolow: Why not `useProject`? It already returns current project, no need to fetch all projects and then find the current one.

application/ui/src/components/zoom/use-sync-zoom.hook.tsx
jpggvilaca: That means if i zoom in then i wont see them again, so this just hides the issue i think. We need to think of a trick to display them at all times. Maybe move them outside of the canvas, since we wont need zoomProvider at all. Thopughts?

camiloHimura: This is the same behavior we have in Geti Classic. When users zoom in on a classification project, they still don’t see the labels. I don’t see this as hiding an issue since, in this type of project, labels are only shown at the top, zooming in or out doesn’t really help the annotation process.

With the current approach, it’s perfectly usable, and we don’t need to add any additional code

application/ui/src/features/dataset/media-preview/secondary-toolbar/util.ts
camiloHimura: it's actually `isExistingLabel`

application/ui/src/hooks/use-selected-project.hook.tsx
camiloHimura: I wasn’t aware of that hook, the name is too generic. I’ll replace it in my next commit

application/ui/src/components/zoom/use-sync-zoom.hook.tsx
jpggvilaca: Fair enough :)

jpggvilaca (dismissed):

application/ui/src/features/annotator/annotations/full-image-shape/full-image-shape.component.tsx
```diff
@@ -0,0 +1,25 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import type { Annotation } from '../../../../shared/types';
+import { useSelectedData } from '../../../dataset/selected-data-provider.component';
+
+type FullImageShapeProps = {
+    annotation: Annotation;
+};
+
+export const FullImageShape = ({ annotation }: FullImageShapeProps) => {
```
dwesolow: Why do we need separate component for this? It's `rect` like in the `rectangle` type.
Moreover, this component misses styles for the prediction (dashed stroke).
If you want to keep it as separate component, please add it to the `AnnotationShape` component and pass styles of that component as a prop (like we do for polygon and rect).

application/ui/src/features/annotator/annotations/annotation-shape-with-labels.component.tsx
```diff
@@ -3,33 +3,49 @@
...
 
+    if (shape.type === 'full_image') {
+        return (
+            <g>
+                <FullImageShape annotation={annotation} />
+                <AnnotationLabels labels={labels} onRemove={removeLabels} />
+            </g>
+        );
+    }
+
```
dwesolow, line 47: The app breaks when we want to use canvas settings and hide labels:

Image

application/ui/src/features/dataset/media-preview/secondary-toolbar/secondary-toolbar.component.tsx
```diff
@@ -49,13 +51,25 @@ export const SecondaryToolbar = ({
...
+        addAnnotations,
+        updateAnnotations,
+        deleteAnnotations,
+        submitAnnotations,
+        submitPredictions,
+    } = useAnnotationActions();
 
     const hasAnnotations = !isEmpty(annotations);
+    const isMultiLabel = selectedProject.task.exclusive_labels === false;
+    const isClassification = selectedProject.task.task_type === 'classification';
```
dwesolow: We do have `isClassificationTask` helper that checks that.

application/ui/src/features/dataset/media-preview/secondary-toolbar/secondary-toolbar.component.tsx
```diff
@@ -99,10 +113,40 @@ export const SecondaryToolbar = ({
...
+    const handleClassificationSelect = (value: Key | null) => {
+        const label = projectLabels.find(({ id }) => id === value);
+        const labels = label ? [label] : [];
+
+        if (isEmpty(annotations)) {
+            addAnnotations([{ type: 'full_image' }], labels);
+        } else if (isMultiLabel) {
+            updateClassificationAnnotations(labels[0]);
+        } else {
+            updateAnnotations(annotations.map((annotation) => ({ ...annotation, labels })));
```
dwesolow, line 125: Shouldn't we `toggleLabel` here as well?

application/ui/src/shared/annotator/annotator-provider.component.tsx
```diff
@@ -31,8 +31,10 @@ export const AnnotatorProviderContext = createContext<AnnotatorContext | null>(n
 
 const useSelectedLabel = () => {
     const labels = useProjectLabels();
-
-    const [selectedLabelId, setSelectedLabelId] = useState<string | null>(labels.length === 0 ? null : labels[0].id);
+    const { data: project } = useProject();
+    const hasDefaultLabel = project.task.task_type !== 'classification';
```
dwesolow: np. `!isClassificationTask(...)`

dwesolow (changes requested):
Minor comments and one bug that breaks the app.
Could we also add an indicator to the projects list that the project is either single classification or multi label classification (can be in a separate PR if we want to do that, I think that might be useful).
The rest LGTM :).

application/ui/src/features/dataset/media-preview/secondary-toolbar/secondary-toolbar.component.tsx
camiloHimura: No, in the case of single selection, we should replace all labels with the newly selected one, not toggle them.

application/ui/src/features/annotator/annotations/full-image-shape/full-image-shape.component.tsx
camiloHimura: Since we need this configuration in AnnotationShape, the extra component is no longer needed, so it’s been removed.

application/ui/src/features/annotator/annotations/annotation-shape-with-labels.component.tsx
camiloHimura: fixed

application/ui/src/features/dataset/media-preview/secondary-toolbar/secondary-toolbar.component.tsx
camiloHimura: updated

application/ui/src/shared/annotator/annotator-provider.component.tsx
camiloHimura: updated

camiloHimura:
> Minor comments and one bug that breaks the app. Could we also add an indicator to the projects list that the project is either single classification or multi label classification (can be in a separate PR if we want to do that, I think that might be useful). The rest LGTM :).

I’ll ping Marius about the label selection. Once he gets back to us, I’ll send a new PR.

application/ui/src/features/dataset/media-preview/secondary-toolbar/secondary-toolbar.component.tsx
dwesolow: Asking because in geti when we select a label that is already assigned it unassigns that label (e.g. we have annotation with label spade, we click on a label spade, it removes the label from annotation). Do we want to have the same behavior?

application/ui/src/features/annotator/annotations/annotation-shape/annotation-shape.component.tsx
```diff
@@ -1,20 +1,35 @@
...
+import { useSelectedData } from '../../../dataset/selected-data-provider.component';
+import { getFormattedPoints, isPrediction } from '../utils';
 
 type AnnotationShapeProps = {
     annotation: Annotation;
 };
 
 export const AnnotationShape = ({ annotation }: AnnotationShapeProps) => {
     const { shape, labels } = annotation;
+    const { selectedMediaItem } = useSelectedData();
```
dwesolow, line 14: To consider: that hook is needed only in `full_image` but is called for each annotation type, does it make sense/impacts performance? (IMO we could keep full image as a separate component but accept styles like stroke color, stroke dash).
Bonus: if that component was extracted, we would not need to update unit test that uses `AnnotationShape` but with annotations as rectangle or polygon.

dwesolow (dismissed):

application/ui/src/components/zoom/use-sync-zoom.hook.tsx
ActiveChooN: Should this default zoom be applied only to classification projects? It looks a bit off for other projects.

jpggvilaca (approved):

application/ui/src/features/annotator/annotations/annotation-shape/annotation-shape.component.tsx
camiloHimura: I'll revisit this on my next PR

application/ui/src/features/dataset/media-preview/secondary-toolbar/secondary-toolbar.component.tsx
camiloHimura: Yes, but let’s not address this in this PR, since the label selector will change with Dima’s updates

PR #5335, opened 2026-01-28
Unit test updates

jpggvilaca:
### Summary

* Use render/renderHook wrapper from geti-prompt to inject url params
* Update imports/mocks everywhere (We should now use our custom `render`)
* Update 1 test and add another

### Checklist

### DIFF
--- a/application/ui/src/features/inference/header/inference-devices.test.tsx
+++ b/application/ui/src/features/inference/header/inference-devices.test.tsx
@@ -41,7 +41,7 @@ describe('InferenceDevices', () => {
 
     beforeEach(() => {
         vi.resetAllMocks();
-        queryClient.clear();
+        queryClient.removeQueries();
     });
 
     it('displays current device selection', async () => {
--- a/application/ui/src/features/inference/header/inference-devices.test.tsx
+++ b/application/ui/src/features/inference/header/inference-devices.test.tsx
@@ -1,18 +1,15 @@
 // Copyright (C) 2025 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 
-import { QueryClient } from '@tanstack/react-query';
-import { render, screen, waitFor } from '@testing-library/react';
+import { screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { getMockedPipeline } from 'mocks/mock-pipeline';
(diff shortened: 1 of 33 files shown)


application/ui/src/features/inference/header/inference-devices.test.tsx
```diff
@@ -36,17 +34,14 @@ describe('InferenceDevices', () => {
...
-            </TestProviders>
-        );
+        render(<InferenceDevices />);
 
         return pipelinePatchSpy;
     };
 
     beforeEach(() => {
         vi.resetAllMocks();
+        queryClient.clear();
```
dwesolow: Why do we need this?
If it's because query client is shared then we need to fix the root case, i.e. each render should create its own client instance.

application/ui/src/features/models/model-listing/current-model-training/training-model-row.component.test.tsx
```diff
@@ -0,0 +1,121 @@
...
+                },
+            },
+            status: 'RUNNING',
+        });
+
+        render(<TrainingModelRow job={job} onCancel={mockCancel} />);
+
+        const cancelButton = await screen.findByRole('button', { name: /cancel training job/i });
+        expect(cancelButton).toBeVisible();
+        expect(cancelButton).not.toBeDisabled();
```
dwesolow: np. `toBeEnabled`

ActiveChooN (dismissed):

application/ui/src/features/inference/header/inference-devices.test.tsx
jpggvilaca: Update to `invalidateQueries` but we need it because this test calls the component multiple times with different responses AND it's wrapped in suspense, so we need to invalidate the client for each one

application/ui/src/features/models/model-listing/current-model-training/training-model-row.component.test.tsx
jpggvilaca: Updated

application/ui/src/features/inference/header/inference-devices.test.tsx
dwesolow: I'll double check it later. Thanks!

dwesolow (approved):

PR #5346, opened 2026-01-28
Unit tests & cleanup

jpggvilaca:
### Summary

* Add unit tests for model row and rename model dialog

### Checklist


application/ui/src/hooks/api/models.hook.ts
```diff
@@ -0,0 +1,156 @@
+// Copyright (C) 2025 Intel Corporation
```
dwesolow: Why this change? IMO hooks should be close part of the feature "slice".

jpggvilaca: Because that's how all other hooks are done, instead of having 50 files i consolidated in one. What i can do is perhaps keep the api ones here and move the useWhatever closer to models? but then we wont know where they are..

ActiveChooN: Can we have general API hooks in one file and special hooks in separate files in the same dir?

dwesolow: I'm not against consolidating hooks into one file. I'm against moving models hooks away from models. Even if the hook is used in two places, it's still part of the models feature (not a reason to move it to shared or whatever other name)

dwesolow (approved):

Issue #5350, opened 2026-01-29
Store 'id' explicitly in dataset revision dataframe


leoll2:
Add an `id` column in the dataset revision Polars dataframe, so that we don't need to reconstruct it from the 'image' field in a hacky way.

PR #5351, opened 2026-01-29
Fix server errors in dataset revision endpoints

### LINKED ISSUE #5258: Revise endpoint to fetch thumbnail for dataset revision item
The endpoint [`get_dataset_revision_item_thumbnail`](https://github.com/open-edge-platform/training_extensions/blob/5f48337ade6734dd6b03f7494d3136dc32e12a0e/application/backend/app/api/routers/dataset_revisions.py#L123) is only partially implemented, as it currently returns the full image instead of the thumbnail.

The goal of this ticket is to provide a better implementation and return an actual thumbnail. The thumbnail can be generated on-the-fly (e.g. with [`Image.thumbnail`](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.thumbnail)) and it doesn't have to be stored on disk. On-the-fly computation shouldn't be a major problem, since the UI loads the dataset items in pages of bounded size, and it can potentially cache the response.

Thumbnail size can be small, e.g. 64x64 pixel.

### LINKED ISSUE #5348: Server error raised by dataset revision item detail and binary endpoints
The following endpoints return a 500 Internal Server Error:
- `GET /api/projects/{project_id}/dataset_revisions/{revision_id}/items/{item_id}`
- `GET /api/projects/{project_id}/dataset_revisions/{revision_id}/items/{item_id}/binary`

leoll2:
### Summary

This PR fixes several bugs in the dataset revision endpoints that would trigger '500 Internal Server Errors'.

_Note:_ the internal structure of dataset revision is going to change after #5350 and https://github.com/open-edge-platform/datumaro/issues/2019; this PR anticipates those changes providing forward-compatibility.

Fixes #5348 
Resolves #5258 

### How to test

Call the dataset revision endpoints, also with invalid parameters, and verify that the API behaves correctly.

### Checklist

### DIFF
--- a/application/backend/app/api/routers/dataset_revisions.py
+++ b/application/backend/app/api/routers/dataset_revisions.py
@@ -1,10 +1,10 @@
 # Copyright (C) 2025 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
-
+from io import BytesIO
 from typing import Annotated
 
 from fastapi import APIRouter, Depends, HTTPException, Query, status
-from starlette.responses import FileResponse
+from starlette.responses import FileResponse, StreamingResponse
 
 from app.api.dependencies import get_dataset_revision, get_dataset_revision_service, get_project
 from app.api.schemas.dataset_item import (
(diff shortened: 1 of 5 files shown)


application/backend/app/api/routers/dataset_revisions.py
```diff
@@ -159,36 +161,51 @@ def get_dataset_revision_item_binary(
...
-    return FileResponse(binary_path, media_type="image/jpeg")
+    try:
+        thumbnail = dataset_revision_service.get_dataset_revision_item_thumbnail(
+            project_id=project.id,
+            dataset_revision=dataset_revision,
+            item_id=str(dataset_item_id),
+        )
+        buffer = BytesIO()
+        thumbnail.save(buffer, format="JPEG")
+        buffer.seek(0)
```
itallix, line 198: Any reason why it can't be placed on the service layer?

leoll2: It can, I placed it here to mirror the implementation of the media thumbnail endpoint.

https://github.com/open-edge-platform/training_extensions/blob/f72b2e0384fa77b1e0ff9354b4cf06a4d612e79d/application/backend/app/api/routers/media.py#L168

We can refactor both in a separate PR.

A-Artemis (approved):

application/backend/app/api/routers/dataset_revisions.py
itallix: Yeah, it makes more sense to move it to the service and side effects will be tested in integration tests.

PR #5362, opened 2026-01-29
feat: auto-select new shape and improve outline styles of labels

ActiveChooN:
### Summary

1. Automatically select any new shape created.
2. Fixed transitions for the outline of the label picker. 
3. Fixed width added to the zoom selector to prevent jumping.

### Checklist

### DIFF
--- a/application/ui/src/features/annotator/labels/labels.module.scss
+++ b/application/ui/src/features/annotator/labels/labels.module.scss
@@ -22,7 +22,11 @@
     border: none;
     flex-shrink: 0;
     height: var(--spectrum-global-dimension-size-350);
-    transition: outline 0.15s ease;
+    outline: var(--spectrum-global-dimension-size-25) solid transparent;
+    outline-offset: 0;
+    transition:
+        outline-color 0.15s ease,
+        outline-offset 0.15s ease;
 }
 
 .badge:hover {
@@ -32,7 +36,7 @@
 .selected,
 .selected:focus {
     outline-offset: var(--spectrum-global-dimension-size-25);
-    outline: var(--spectrum-global-dimension-size-25) solid var(--labelBgColor);
+    outline-color: var(--labelBgColor);
 }
 
 .disabled {
--- a/application/ui/src/features/annotator/tools/bounding-box-tool/bounding-box-tool.component.tsx
+++ b/application/ui/src/features/annotator/tools/bounding-box-tool/bounding-box-tool.component.tsx
@@ -3,24 +3,18 @@
 
 import { useZoom } from '../../../../components/zoom/zoom.provider';
 import type { Label } from '../../../../constants/shared-types';
-import { useAnnotationActions } from '../../../../shared/annotator/annotation-actions-provider.component';
 import { useAnnotator } from '../../../../shared/annotator/annotator-provider.component';
-import { useSelectedAnnotations } from '../../../../shared/annotator/select-annotation-provider.component';
 import type { Rect } from '../../../../shared/types';
 import { DrawingBox } from '../drawing-box-tool/drawing-box.component';
+import { useAddAndSelectAnnotations } from '../use-add-and-select-annotations.hook';
 
 export const BoundingBoxTool = () => {
     const { scale: zoom } = useZoom();
-    const { addAnnotations } = useAnnotationActions();
-    const { setSelectedAnnotations } = useSelectedAnnotations();
+    const { addAndSelectAnnotations } = useAddAndSelectAnnotations();
     const { mediaItem, image, selectedLabel } = useAnnotator();
 
     const handleComplete = (shapes: Rect[], labels: Label[]): string[] => {
-        const newIds = addAnnotations(shapes, labels);
-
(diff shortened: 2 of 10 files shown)


application/ui/src/shared/annotator/annotation-actions-provider.component.tsx
```diff
@@ -104,15 +104,16 @@ export const AnnotationActionsProvider = ({
...
+    const addAnnotations = (shapes: Shape[], labels: Label[]): string[] => {
+        const newAnnotations = shapes.map((shape) => ({
+            shape,
+            id: uuid(),
+            labels,
+        }));
+
+        setAnnotations((prevAnnotations) => [...prevAnnotations, ...newAnnotations]);
+
+        return newAnnotations.map((annotation) => annotation.id);
```
dwesolow, line 116: I'm wondering if we should add a new method `addAndSelectAnnotations` what would combine `addAnnotations` and `setSelectedAnnotations` since we want that behavior in each tool.
WDYT?

application/ui/src/features/annotator/labels/labels.module.scss
```diff
@@ -22,7 +22,11 @@
     border: none;
     flex-shrink: 0;
     height: var(--spectrum-global-dimension-size-350);
-    transition: outline 0.15s ease;
+    outline: var(--spectrum-global-dimension-size-25) solid transparent;
+    outline-offset: 0;
```
jpggvilaca, line 26: do we need the offset?

application/ui/src/shared/annotator/annotation-actions-provider.component.tsx
ActiveChooN: I thought about it, but it will couple two providers together, I don't think it's a very good idea

application/ui/src/features/annotator/labels/labels.module.scss
ActiveChooN: Yes, it's starting position of transition. So we start from outline-offset: 0 and transparent color, and finish at outline-offset: size-25 and labelBgColor

application/ui/src/shared/annotator/annotation-actions-provider.component.tsx
dwesolow: I didn't mean coupling those two providers but we might have extra one just for adding annotations.

application/ui/src/features/annotator/tools/bounding-box-tool/bounding-box-tool.component.tsx
```diff
@@ -2,22 +2,34 @@
...
+import type { Rect } from '../../../../shared/types';
 import { DrawingBox } from '../drawing-box-tool/drawing-box.component';
 
 export const BoundingBoxTool = () => {
     const { scale: zoom } = useZoom();
     const { addAnnotations } = useAnnotationActions();
+    const { setSelectedAnnotations } = useSelectedAnnotations();
     const { mediaItem, image, selectedLabel } = useAnnotator();
 
+    const handleComplete = (shapes: Rect[], labels: Label[]): string[] => {
```
jpggvilaca, line 16: why dont we move the "auto-select" behavior to the annotation itself? i.e., onMount, select oneself. WDYT?

application/ui/src/shared/annotator/annotation-actions-provider.component.tsx
camiloHimura: I like the concept, but I’m not convinced about the new provider path. Would a hook be a better option?

dwesolow: Yeah, hooks is a good option as well. Like `useAddAndSelectAnnotation` which combines both methods.

application/ui/src/features/annotator/tools/bounding-box-tool/bounding-box-tool.component.tsx
dwesolow: I think better option is not to react but to implicitly do actions, less side effects.

application/ui/src/shared/annotator/annotation-actions-provider.component.tsx
ActiveChooN: I think it's good idea, I've added separate hook

dwesolow (approved):
Great work!

application/ui/src/features/annotator/tools/bounding-box-tool/bounding-box-tool.component.tsx
ActiveChooN: I think it would introduce hidden behavior to the annotation, causing more problems, for example, in cases when we don't want to auto-select (undo, redo, or paste). In the case of creating multiple shapes like SAM, it can cause flickering and other problems.

jpggvilaca: Agree

jpggvilaca (approved):

PR #5364, opened 2026-01-29
feat: labels overflow compaction

ActiveChooN:
### Summary

image
A "Show More" button has been added to the labels list. For now, it just expands the list. In the future, it will open the label management popover.

### Checklist


application/ui/src/features/annotator/labels/labels.module.scss
```diff
@@ -29,6 +35,12 @@
     opacity: 0.9;
 }
 
+.hidden {
+    position: absolute;
```
dwesolow, line 47: why do we need this?

ActiveChooN: We need to use `visibility: hidden;` to calculate the width of an element without showing it. However, in that case, the element will still occupy its place. Therefore, we need to use `position: absolute;` to remove it from the current layout.

application/ui/src/features/dataset/media-preview/secondary-toolbar/use-visible-labels-count.hook.ts
```diff
@@ -0,0 +1,84 @@
...
+const BADGE_GAP = 8; // size-100 gap
+const OTHER_COMPONENTS_RESERVATION = 104;
+
+interface UseVisibleLabelsCountProps {
+    toolbarRef: RefObject<HTMLDivElement | null>;
+    labelsContainerRef: RefObject<HTMLDivElement | null>;
+    totalLabels: number;
+}
+
+const calculateVisibleLabels = (labelsContainer: HTMLElement, availableWidth: number): number => {
```
jpggvilaca, line 16: Like we discussed yesterday, let's keep it simple first. This is very error prone and overly complicated imho. What do you think of:
1) Fixed width (a % of the dialog) for the labels container
2) Overflow-x scroll

What we need is to also set a MAX FIXED WIDTH for the labels themselves, to have text-overflow ellipsis for long names.

ActiveChooN: Imo, fixed width with overflow would look less slick and would provide worse UX. But I'll add overflow for the labels themself

jpggvilaca: Absolutely. I just want to make it simple. I'd rather have obvious simple code then all these calculations before we need them. Also it's very hard to test

PR #5367, opened 2026-01-30
Integrate dataset revisions

jpggvilaca:
### Summary

* Next PR: Add rename/delete dataset revision

Screenshot 2026-01-30 at 11 45 50
Screenshot 2026-01-30 at 11 45 53

### Checklist


application/ui/src/hooks/use-get-dataset-revision.hook.ts
```diff
@@ -0,0 +1,22 @@
...
+    return $api.useQuery(
+        'get',
+        '/api/projects/{project_id}/dataset_revisions/{dataset_revision_id}',
+        {
+            params: {
+                path: { project_id, dataset_revision_id: datasetRevisionId },
+            },
+        },
+        {
+            enabled: !!datasetRevisionId && datasetRevisionId !== 'unknown',
```
dwesolow: Can we remove `unknown` option? If there is not id we can just display `Unknown` but not assign to `datasetRevisionId`. It can be easily forgotten while development.
Why can it be empty? Type suggest that it can be only a `string`.

application/ui/src/features/models/train-model/train-model-provider.component.tsx
dwesolow, line 71: IRRC we assume here that dataset revisions are present (not empty) and we select `selectedDatasetRevision` by default. Question if it's a big deal to have it automatically selected (taking into account that we need to support training from a given dataset on a models screen.

application/ui/src/hooks/use-get-dataset-revision.hook.ts
jpggvilaca: We dont even need this hook, leftover. Removed.

application/ui/src/features/models/train-model/train-model-provider.component.tsx
jpggvilaca: Good point, but let me address this on a subsequent PR

dwesolow (approved):

application/ui/src/features/models/train-model/train-model-provider.component.tsx
dwesolow: maybe replacing to `useSuspenseQuery` will work if we have proper suspense boundaries.

jpggvilaca: Will also take that into consideration

PR #5376, opened 2026-01-30
fix error 404 when listing dataset revisions after deletion

### LINKED ISSUE #5374: Error 404 after soft-deleting dataset revision
**Steps to reproduce**

1. Train a model
2. Soft-delete a dataset revision with `DELETE /dataset_revisions/{id}`
3. List dataset revisions with `GET /dataset_revisions`

**Expected**

The revision is listed with `files_deleted: true`.

**Actual result**

Error 404 - Project not found

leoll2:
### Summary

Changes:
- `item_counts` is now only returned if the revision files are not deleted
  - Fixes #5374 : error 404 when listing a dataset revision with deleted files
- Improved docstrings and API descriptions

### How to test

Follow the steps described in #5374, and run some operations involving the deleted revision.

image

### Checklist


dwesolow (dismissed):
UI approved

application/backend/app/api/routers/dataset_revisions.py
```diff
@@ -59,16 +59,17 @@ def list_dataset_revisions(
...
-        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Project not found")
+            dataset_revision_view_data = dataset_revision.model_dump()
+            if not dataset_revision.files_deleted:
+                item_counts = dataset_revision_service.count_items_by_subset(
+                    project_id=project.id, dataset_revision_id=dataset_revision.id
+                )
+                dataset_revision_view_data |= {"item_counts": item_counts}
+            dataset_revision_view = DatasetRevisionView.model_validate(dataset_revision_view_data, from_attributes=True)
+            dataset_revision_views.append(dataset_revision_view)
+        return dataset_revision_views
```
itallix, line 70: Do we have unit tests that verify `files_deleted` scenario?

application/backend/app/api/routers/dataset_revisions.py
```diff
@@ -59,16 +59,17 @@ def list_dataset_revisions(
...
-            )
-            dataset_revision_views.append(dataset_revision.model_dump() | {"item_counts": item_counts})
-        return [
-            DatasetRevisionView.model_validate(dataset_revision_view, from_attributes=True)
-            for dataset_revision_view in dataset_revision_views
-        ]
-    except ResourceNotFoundError:
-        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Project not found")
+            dataset_revision_view_data = dataset_revision.model_dump()
+            if not dataset_revision.files_deleted:
```
JortBergfeld: I don't think this check belongs in the endpoint. This should be done in the service.

I also noted that the files_deleted is set to True before actually deleting the files, not sure if that is the best way.

application/backend/app/api/routers/dataset_revisions.py
```diff
@@ -118,16 +121,26 @@ def rename_dataset_revision(
     dataset_revision_service: Annotated[DatasetRevisionService, Depends(get_dataset_revision_service)],
 ) -> DatasetRevisionView:
     """Rename a dataset revision."""
+    try:
+        new_name = dataset_revision_metadata["name"]
+    except KeyError:
```
JortBergfeld, line 126: I think it is better to properly annotate the input data, so that we use the default pydantic validation mechanism for this.

camiloHimura (dismissed):

application/backend/app/api/routers/dataset_revisions.py
leoll2: > I don't think this check belongs in the endpoint. This should be done in the service.

Updated, thanks for the suggestion

> I also noted that the files_deleted is set to True before actually deleting the files, not sure if that is the best way.

This is intentional, to avoid race conditions: if you reverse the order, i.e. if you first delete the files and later set `files_deleted`, then a GET /{dataset_revision_id} request submitted concurrently could lead to a server error if the flag is False but the files are already gone.

application/backend/app/api/routers/dataset_revisions.py
leoll2: Added!

application/backend/app/api/routers/dataset_revisions.py
leoll2: Good idea, I'll create a ticket to do it in another PR. The PATCH model revision should be updated as well, while the PATCH project already has a clear schema (ProjectUpdateName).

dwesolow (dismissed):

itallix (dismissed):

itallix (approved):

dwesolow (approved):

PR #5379, opened 2026-02-02
chore: Use type-safe invalidateQueries

dwesolow:
### Summary

This PR adds strong typings to query keys and fixes existing issues related to that.
(type copied from geti prompt)

### Checklist

### DIFF
--- a/application/ui/src/query-client/query-client.interface.ts
+++ b/application/ui/src/query-client/query-client.interface.ts
@@ -1,0 +1,40 @@
+// Copyright (C) 2025-2026 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+import type { HttpMethod } from 'openapi-typescript-helpers';
+
+import { type paths } from '../api/openapi-spec';
+
+type OperationFor<Paths extends paths, P extends keyof Paths, Method extends HttpMethod> = Method extends keyof Paths[P]
+    ? Paths[P][Method]
+    : never;
+
+type PathParamsFor<Paths extends paths, P extends keyof Paths, Method extends HttpMethod> =
+    OperationFor<Paths, P, Method> extends { parameters: { path: infer PP } } ? PP : never;
+
+type MethodsForPath<Paths extends paths, P extends keyof Paths> = Extract<keyof Paths[P], HttpMethod>;
+
+export type QueryKey<Paths extends paths> = {
+    [P in keyof Paths]: {
+        [M in MethodsForPath<Paths, P>]: PathParamsFor<Paths, P, M> extends never
+            ? [M, P]
+            : [
+                  M,
+                  P,
+                  {
+                      params: {
+                          path: PathParamsFor<Paths, P, M>;
+                      };
+                  },
+              ];
+    }[MethodsForPath<Paths, P>];
+}[keyof Paths];
+
+export type Meta = {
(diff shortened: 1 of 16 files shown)


application/ui/src/query-client/query-client.ts
```diff
@@ -0,0 +1,77 @@
...
+        mutations: {
+            retry: false,
+        },
+    },
+    mutationCache: new MutationCache({
+        onSuccess: (_data, _variables, _context, mutation) => {
+            const meta = mutation.meta;
+            const invalidateQueries = meta?.invalidateQueries;
+
+            if (invalidateQueries) {
```
jpggvilaca, line 50: not sure if i agree here. Either we invalidate by default for ALL, or we do it manually case-by-base (I prefer this one). So we can get rid of this one here

jpggvilaca: I know we already had this here but perhaps we can get rid of it?

jpggvilaca: Disregard. Misunderstood

jpggvilaca (dismissed):

application/ui/src/query-client/query-client.interface.ts
```diff
@@ -0,0 +1,40 @@
...
+type OperationFor<Paths extends paths, P extends keyof Paths, Method extends HttpMethod> = Method extends keyof Paths[P]
+    ? Paths[P][Method]
+    : never;
+
+type PathParamsFor<Paths extends paths, P extends keyof Paths, Method extends HttpMethod> =
+    OperationFor<Paths, P, Method> extends { parameters: { path: infer PP } } ? PP : never;
+
+type MethodsForPath<Paths extends paths, P extends keyof Paths> = Extract<keyof Paths[P], HttpMethod>;
+
+export type QueryKey<Paths extends paths> = {
```
camiloHimura, line 17: It would be helpful to add some comments, as this isn’t immediately clear to read.

application/ui/src/query-client/query-client.ts
```diff
@@ -0,0 +1,77 @@
...
+declare module '@tanstack/react-query' {
+    interface Register {
+        mutationMeta: Meta;
+        queryMeta: Meta;
+    }
+}
+
+const TOAST_DURATION = 5000;
+
+const getErrorMessage = (error: unknown): string => {
```
camiloHimura, line 19: It’s best to avoid more than two levels of nesting, since deeper nesting increases cognitive load. This is another way to structure the code.

```suggestion
const getErrorMessage = (error: unknown): string => {
       if (!error || typeof error !== 'object') {
            return 'An unexpected error occurred. Please try again.';
        }

        if ('detail' in error && typeof error.detail === 'string') {
            return error.detail;
        }

        if ('message' in error && typeof error.message === 'string') {
            const message = error.message;

            if (error instanceof TypeError && message.includes('Failed to fetch')) {
                return 'Network error. Please check your connection and try again.';
            }

            return message;
        }

        return 'An unexpected error occurred. Please try again.';
        }
```

application/ui/src/query-client/query-client.ts
camiloHimura: I also prefer the first option. Since it’s the pattern used in other applications, let’s aim to stay consistent.

jpggvilaca (approved):

PR #5385, opened 2026-02-02
Video upload

### LINKED ISSUE #5280: Implement video upload endpoints
Add support for video files in the media upload endpoints and implement the related logic to store them.

warrkan:
Resolves #5280

### Checklist


application/backend/app/services/video/video_service.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class VideoMetadata(BaseModel):
+    width: int
+    height: int
+    frame_count: int
+    fps: float
+
+
+class FfprobeOutput(BaseModel):
```
itallix:
```suggestion
class FFProbeOutput(BaseModel):
```

application/backend/app/services/media_service.py
```diff
@@ -17,13 +18,17 @@
...
+from app.models.media import ImageFormat, VideoFormat
 from app.repositories import MediaRepository
+from app.services.video import extract_video_frame, get_video_metadata
 from app.utils.images import crop_to_thumbnail
 
 from .base import BaseSessionManagedService, ResourceNotFoundError, ResourceType
 
 DEFAULT_THUMBNAIL_SIZE = 256
 
+VIDEO_WRITE_CHUNK_SIZE = 1024 * 1024 * 1024
```
itallix: Shall it be 1MB instead of 1GB?

application/backend/app/services/video/video_service.py
warrkan: That's redundant code, removed

application/backend/app/services/media_service.py
warrkan: Made it 100 Mb 😄

application/backend/app/models/media.py
```diff
@@ -37,6 +48,8 @@ class Media(BaseEntity):
         width: Width of the media in pixels.
         height: Height of the media in pixels.
```
leoll2, line 49: nitpicky: document when/why it can be None

application/backend/app/api/routers/media.py
```diff
@@ -40,15 +54,24 @@ def add_media(
...
+                format=format,
+            )
+        else:
+            media = media_service.create_video(
+                project=project,
+                data=file.file,
+                name=name,
+                format=format,
+            )
         dataset_service.create_dataset_item(
```
leoll2: On video upload, no dataset item should be created. Instead, dataset items should be created for the individual frames when annotating them.

application/backend/app/services/video/video_service.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+class VideoMetadata(BaseModel):
+    width: int
+    height: int
+    frame_count: int
+    fps: float
+
+
+def get_video_metadata(video_path: Path) -> VideoMetadata:
```
leoll2, line 17: How long does this method take for a relatively large high-res video? If it's not more than a few seconds, we may consider calling it on video upload, so that all metadata attributes are always available and never None.

application/backend/app/services/media_service.py
```diff
@@ -180,19 +224,69 @@ def get_media_thumbnail_path_by_id(self, project: Project, media_id: UUID) -> Pa
         media = self.get_media_by_id(project_id=project.id, media_id=media_id)
         return self.projects_dir / f"{project.id}/dataset/{media.id}-thumb.jpg"
 
+    def _update_video_metadata(self, project: Project, media_id: UUID) -> Media:
```
leoll2: If I'm not mistaken, this method is only called on `generate_media_thumbnail`. If a client interacts through the REST API, he may never call the thumbnail endpoint, and the video metadata will never be initialized.

If possible, let's initialize the metadata on video upload (see my other comment). If not, you should call this method everywhere the metadata is needed and may be missing (e.g. annotation endpoint).

application/backend/app/utils/images.py
```diff
@@ -26,7 +26,9 @@ def crop_to_thumbnail(image: Image.Image, target_height: int, target_width: int)
     scale_width = target_width / image.width
     scale_height = target_height / image.height
     scaling_factor = max(scale_width, scale_height)
-    resized_image = image.resize((int(image.width * scaling_factor), int(image.height * scaling_factor)))
+    resized_image = image.resize(
+        (int(image.width * scaling_factor), int(image.height * scaling_factor)), resample=Image.Resampling.BILINEAR
```
leoll2: Any specific reason for choosing BILINEAR interpolation?

application/backend/app/services/datumaro_converter.py
```diff
@@ -218,7 +218,7 @@ def convert_classification_dataset(
     def _convert_sample(
         dataset_item: DatasetItem, media: Media, image_path: str, project_labels_ids: list[UUID]
     ) -> ClassificationSample | None:
-        if dataset_item.annotation_data is None:
+        if dataset_item.annotation_data is None or media.width is None or media.height is None:
```
leoll2: I think the condition `media.width is None or media.height is None` can never happen: if a media is annotated (`annotation_data is not None`), then the media attributes must be already set otherwise the annotation couldn't happen in the first place.

application/backend/app/services/dataset_service.py
```diff
@@ -222,6 +222,8 @@ def _validate_annotations(annotations: list[DatasetItemAnnotation], project: Pro
 
     @staticmethod
     def _validate_annotations_coordinates(annotations: list[DatasetItemAnnotation], media: Media | MediaDB) -> None:
+        if media.width is None or media.height is None:
+            raise AnnotationValidationError(f"Media {media.id} doesn't have width or height defined.")
```
leoll2: `AnnotationValidationError` corresponds to 400 Bad Request and it indicates that the annotation submitted by the client is invalid. This is not the case: if the media width/height attributes are not initialized, it's either a server error or you have to initialize them on the fly (see my prev comment).

application/backend/app/models/media.py
warrkan: Made these fields mandatory

application/backend/app/api/routers/media.py
warrkan: Thanks for spotting! Removed it

application/backend/app/services/video/video_service.py
warrkan: Moved video metadata reading into video upload method

application/backend/app/services/media_service.py
warrkan: Moved video metadata reading into video upload method

application/backend/app/services/datumaro_converter.py
warrkan: Width and height are mandatory now, removed this check

application/backend/app/utils/images.py
warrkan: Ah, thanks! I think it remained from some of the experiments

application/backend/app/services/dataset_service.py
warrkan: Removed this check

application/backend/app/services/datumaro_converter.py
```diff
@@ -234,7 +234,7 @@ def _convert_sample(
             return ClassificationSample(
                 id=str(dataset_item.id),
                 image=image_path,
-                image_info=ImageInfo(width=media.width, height=media.height),
+                image_info=ImageInfo(width=media.width, height=media.height),  # pyrefly: ignore[bad-argument-type]
```
leoll2: What causes the `bad-argument-type` error here?

application/backend/app/api/routers/media.py
```diff
@@ -40,20 +54,29 @@ def add_media(
...
+                media=media,
+                user_reviewed=False,
+            )
+        else:
+            media = media_service.create_video(
+                project=project,
+                data=file.file,
+                name=name,
+                format=format,
+            )
```
leoll2, line 80: I suggest to add a comment here, explaining that for videos we don't create dataset items on upload, but lazily on individual frames when they get annotated.

application/backend/app/services/video/video_service.py
```diff
@@ -0,0 +1,72 @@
...
+    height: int = Field(..., description="Video height", ge=0)
+    frame_count: int = Field(..., description="Video frames number", ge=0)
+    fps: float = Field(..., description="Video frames per second", ge=0)
+
+
+def get_video_metadata(video_path: Path) -> VideoMetadata:
+    """
+    Extracts a video metadata
+
+    :param video_path: Video binary file path
```
leoll2: Please use google style docstrings (Args instead of :param:)

application/backend/tests/integration/services/test_media_service.py
```diff
@@ -414,18 +429,58 @@ def fxt_project_with_subset_items(fxt_project_with_pipeline, db_session) -> tupl
     return project, db_dataset_items
 
 
+def _ensure_ffmpeg():
+    """Ensure ffmpeg is available on PATH; skip test if not."""
+    exe = shutil.which("ffmpeg")
+    if exe is None:
+        pytest.skip("ffmpeg not found in PATH; skipping video related tests")
+    return exe
```
leoll2: Tests shouldn't depend on ffmpeg if the application doesn't. You can use OpenCV to generate a dummy video for testing purposes (fxt_video_data).

application/backend/app/services/datumaro_converter.py
leoll2: I see you already removed it at L237, but there are more at L286, L358 and L438, respectively. I think they can be removed as well.

leoll2 (approved):

jpggvilaca (approved):

Issue #5247, opened 2026-01-20
Annotator context menu


camiloHimura:
Adds a reusable context menu component for the annotator. The initial use case is the polygon tool: when a user right-clicks on a polygon point, the menu is displayed with an option to delete that point.

PR #5389, opened 2026-02-02
feat: Context menu

camiloHimura:
### Summary
#5247
image

### Checklist

### DIFF
--- a/application/ui/src/components/cursor-context-menu/cursor-context-menu.component.tsx
+++ b/application/ui/src/components/cursor-context-menu/cursor-context-menu.component.tsx
@@ -14,7 +14,7 @@ export interface CursorContextMenuProps {
     state: OverlayTriggerState;
     children: ReactNode;
     onOpen: () => void;
-    isValidTrigger: (element: Element) => boolean;
+
     triggerRef: RefObject<Element | null>;
 }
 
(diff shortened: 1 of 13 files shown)


jpggvilaca:
Screenshot 2026-02-03 at 09 21 51

Shouldn't we only have 1 instance of the menu at any times? It feels a bit weird to have N context menus

application/ui/src/components/cursor-context-menu/cursor-context-menu.component.tsx
```diff
@@ -0,0 +1,75 @@
...
+            isDismissable: true,
+            shouldCloseOnBlur: false,
+            onClose: state.close,
+        },
+        ref
+    );
+
+    if (!state.isOpen) return null;
+
+    return createPortal(
```
dwesolow: Have you considered using spectrum's popover?
I spent some time checking how to implement something like context menu in geti prompt and I managed to implement that using `Popover` from react aria components. The main benefit of that solution is that it supports "on outside click" out of the box, it has nice animation and more.
Feel free to take a look https://github.com/open-edge-platform/geti-prompt/blob/main/application/ui/src/features/annotator/tools/segment-anything-tool/create-label.component.tsx.

camiloHimura:
> Shouldn't we only have 1 instance of the menu at any times? It feels a bit weird to have N context menus
Updated in my last commit with Spectrum Popover

jpggvilaca (approved):

application/ui/src/components/cursor-context-menu/cursor-context-menu.component.tsx
camiloHimura: Great suggestion. I switched to the `Spectrum Popover`, which uses `react-aria-components/Popover` under the hood, so we don’t need to add the `ThemeProvider` manually.

PR #5400, opened 2026-02-03
feat: Handle empty label

### LINKED ISSUE #5182: Handle empty label in annotator
Unlike in Geti Classic, Geti Tune server will not return the empty label in the list of project labels (it currently does, but it's a bug).
At API level, the UI is expected to submit an empty list of annotations ([]) when the empty label is applied; likewise, if the server returns []  annotations, the UI should interpret it as empty label.

Some API examples in this PR: https://github.com/open-edge-platform/training_extensions/pull/5179

dwesolow:
### Summary

We display:
- "No label" for multilabel classification;
- "No objects" for detection and segmentation.
- If image has no annotations, server returns 404.
- If image has no annotations but was saved with empty label before, server returns annotations: [].
- If we want to submit annotations with empty label, payload should send empty annotations.

image

Close: #5182 

### Checklist

- [X] All related issues are linked to this PR (if applicable)

### DIFF
--- a/application/ui/src/features/annotator/labels/labels.component.tsx
+++ b/application/ui/src/features/annotator/labels/labels.component.tsx
@@ -149,10 +149,12 @@ export const Labels = ({ isClassification = false, isMultiLabel = false, isReadO
     };
 
     const isLabelSelected = (label: Label): boolean => {
+        if (label.id === EMPTY_LABEL_ID) {
+            return false;
+        }
+
         if (isClassification) {
-            return annotations.some((annotation) =>
-                annotation.labels.some((l) => l.id === label.id && l.id !== EMPTY_LABEL_ID)
-            );
+            return annotations.some((annotation) => annotation.labels.some((l) => l.id === label.id));
         }
         return selectedLabelId === label.id;
     };
--- a/application/ui/src/features/annotator/labels/labels.component.tsx
+++ b/application/ui/src/features/annotator/labels/labels.component.tsx
@@ -146,7 +146,9 @@ export const Labels = ({ isClassification = false, isMultiLabel = false, isReadO
 
     const isLabelSelected = (label: Label): boolean => {
         if (isClassification) {
-            return annotations.some((annotation) => annotation.labels.some((l) => l.id === label.id));
+            return annotations.some((annotation) =>
+                annotation.labels.some((l) => l.id === label.id && l.id !== EMPTY_LABEL_ID)
+            );
         }
         return selectedLabelId === label.id;
     };
--- a/application/ui/src/features/annotator/labels/labels.component.tsx
+++ b/application/ui/src/features/annotator/labels/labels.component.tsx
@@ -10,6 +10,7 @@ import { isEmpty } from 'lodash-es';
 import type { Label } from '../../../constants/shared-types';
 import { useAnnotationActions } from '../../../shared/annotator/annotation-actions-provider.component';
 import { useAnnotator } from '../../../shared/annotator/annotator-provider.component';
+import { EMPTY_LABEL_ID } from '../../../shared/annotator/labels';
 import { useSelectedAnnotations } from '../../../shared/annotator/select-annotation-provider.component';
 import type { Annotation } from '../../../shared/types';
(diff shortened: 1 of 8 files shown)


application/ui/src/features/annotator/labels/labels.component.tsx
```diff
@@ -57,8 +60,19 @@ export const Labels = ({ isClassification = false, isMultiLabel = false, isReadO
             return;
         }
 
+        if (label.id === EMPTY_LABEL_ID && annotations.length !== 0) {
```
jpggvilaca, line 64: Suggestion: Add a `addEmptyLabel` to the provider and let that be responsible for deleting/adding and filtering out empty labels. Basically likes 63-75

dwesolow: Provider is responsible for adding annotations, so maybe `addEmptyLabelledAnnotation`? (64-65)

jpggvilaca: Yeah, much better name! (but wait on the rest of the review pls)

application/ui/src/features/annotator/labels/labels.component.tsx
```diff
@@ -133,13 +156,16 @@ export const Labels = ({ isClassification = false, isMultiLabel = false, isReadO
...
             >
                 {labels.map((label) => (
-                    <LabelBadge
-                        key={label.id}
-                        label={label}
-                        isSelected={isLabelSelected(label)}
-                        isDisabled={isReadOnly}
-                        onClick={() => handleLabelClick(label)}
-                    />
+                    <>
```
jpggvilaca, line 170: I think adding a comment here will make more sense:

My idea was to not have the "empty label" be part of the labels, because in theory (or in practice) it isnt. We want a way of saying the MEDIA has no ANNOTATIONS. Labels are not part of the equation imho.

So, in terms of UI, this would be:

Labels | NoObjectButton

Pressing the button will make the media have no annotations
Label selection or tool usage will do the opposite

There should be method for this, that removes all annotations every time is pressed (either to make it NO OBJECT, or to "Reset" before adding new ones)

This will make it easier because we wont have to filter 50 times (based on project and stuff) the labels.
Now what's left is to fix the visualisation. For that we can

1) Delete all annotations
2) Add dummy annotation with "No objects" label

Which you partially already are doing.

TLDR: Do not make EMPTY_LABEL an actual label

Thoughts?

dwesolow: Understood, but how would u like to render annotation without empty label being treated as a label but with labelled as empty?

dwesolow: We need to remember that when we get `annotations: []` from the server we want to show global annotations with empty label as well. So it's not only about creating a global annotation with empty label but also receiving the annotation with empty label.
In `mapServerAnnotationsToLocal` we actually filter out labels that are not part of the annotation and project. It means that if we really want to remove empty label from labels, we'd need to add if statement there to not remove that annotation which is empty.

application/ui/src/features/annotator/labels/labels.component.tsx
```diff
@@ -57,8 +61,20 @@ export const Labels = ({ isClassification = false, isMultiLabel = false, isReadO
...
+            return;
+        }
+
         if (isMultiLabel) {
-            const updatedAnnotations = annotations.map((annotation) => ({
+            // If there is an annotation with the empty label, and we are trying to add assign new label to it, we need
+            // to remove the empty label from the annotation first.
+            const annotationWithoutEmptyLabel = annotations.map((annotation) => ({
+                ...annotation,
+                labels: filterOutEmptyLabels(annotation.labels),
```
jpggvilaca: if the annotations dont have empty label, do we still need this filter?

dwesolow: No, just thought that it's just one operation less but I can add that check to make it more clear.

jpggvilaca (dismissed):

jpggvilaca (approved):

PR #5416, opened 2026-02-04
Fix annotation validation to properly handle the empty label case

leoll2:
### Summary

The following tasks should allow the empty label (represented as `annotations: []`):
- multilabel classification
- detection
- instance segmentation

Whereas the following don't:
- multiclass classification

### How to test

Create a project of each task, upload an image and submit an annotation with the empty label.

### Checklist

### DIFF
--- a/application/backend/app/services/dataset_service.py
+++ b/application/backend/app/services/dataset_service.py
@@ -94,7 +94,7 @@ class DatasetService(BaseSessionManagedService):
         if annotations is not None:
             labels = self._label_service.list_all(project_id=project.id)
             DatasetService._validate_annotations_labels(annotations=annotations, labels=labels)
-            DatasetService._validate_annotations(annotations=annotations, project=project)
+            DatasetService._validate_annotations(annotations=annotations, task=project.task)
             DatasetService._validate_annotations_coordinates(annotations=annotations, media=media)
 
             dataset_item.annotation_data = [annotation.model_dump(mode="json") for annotation in annotations]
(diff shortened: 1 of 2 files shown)


application/backend/tests/unit/services/test_dataset_service.py
```diff
@@ -166,7 +166,10 @@ def test_validate_annotations_multilabel_classification(self, fxt_multilabel_cla
                 shape=FullImage(type="full_image"),
             )
         ]
-        DatasetService._validate_annotations(annotations=annotations, project=fxt_multilabel_classification_project)
+        DatasetService._validate_annotations(annotations=annotations, task=fxt_multilabel_classification_project.task)
+
+    def test_validate_annotations_multilabel_classification_empty(self, fxt_multilabel_classification_project) -> None:
+        DatasetService._validate_annotations(annotations=[], task=fxt_multilabel_classification_project.task)
```
JortBergfeld, line 174: I think it is nice to do an explicit assertion and also for the ones below
```suggestion
        assert DatasetService._validate_annotations(annotations=[], task=fxt_multilabel_classification_project.task) is None
```

application/backend/app/services/dataset_service.py
```diff
@@ -191,15 +191,20 @@ def _validate_annotations_labels(annotations: list[DatasetItemAnnotation], label
...
                 if len(annotations) > 1:
                     raise AnnotationValidationError("Classification project doesn't allow more than one annotation.")
                 annotation = annotations[0]
                 if not isinstance(annotation.shape, FullImage):
                     raise AnnotationValidationError("Classification project supports only full_image shapes.")
-                if project.task.exclusive_labels and len(annotation.labels) > 1:
+                if task.exclusive_labels and len(annotation.labels) > 1:
                     raise AnnotationValidationError(
                         "Multiclass classification project doesn't allow more than one label per annotation."
                     )
```
itallix, line 210: I suggest we combine this check with L197-201

leoll2: This condition (L207) depends on `annotation`, which is extracted at L204. Logically, the checks follow this order:
1. Validate that the number of annotations is compatible with the task
2. Validate each annotation individually

application/backend/tests/unit/services/test_dataset_service.py
leoll2: I don't see the point of asserting the return value of a function that isn't meant to return anything. The validation functions do nothing but raise an exception if and only if the input data is invalid; equivalently, if the function doesn't raise an exception, then the input data is considered valid.

JortBergfeld: It does not change the test, but it makes it more visible that the test is checking something. Now the test simply calls a function and that's it. It takes a bit of mental power to understand what the test is doing. You can also consider adding a comment, simply stating `Check that validation passes successfully`, to make it easier to understand the purpose of the test.

JortBergfeld (dismissed):

application/backend/tests/unit/services/test_dataset_service.py
leoll2: I've added some comments in the tests to explain when the method should or shouldn't raise an exception, and why.

JortBergfeld (approved):

PR #5418, opened 2026-02-04
feat: Popover for label management

ActiveChooN:
### Summary

image

Label management popover was added to the labels picker component. Implemented:
- annotation label selection
- label color change
- label name change
- delete label
- create new label 

TBD:
- label shortcuts
- label pins

### Checklist


application/ui/src/features/annotator/labels/labels.component.tsx
```diff
@@ -159,26 +181,110 @@ export const Labels = ({ isClassification = false, isMultiLabel = false, isReadO
         return selectedLabelId === label.id;
     };
 
+    const handleRequestDeleteLabel = (label: Label) => {
+        setLabelToDelete(label);
+        popoverState.close();
```
ActiveChooN: It looks not so good, but we have already dialog over popover over modal

application/ui/src/components/label-fields/label-color-picker.component.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { DISTINCT_COLORS } from '../../features/annotator/label-utils';
+
+interface LabelColorPickerProps {
+    color: string;
+    onChange: (color: string) => void;
+}
+
+export const LabelColorPicker = ({ color, onChange }: LabelColorPickerProps) => {
+    return (
+        <SpectrumColorPicker value={color} onChange={(c) => onChange(c.toString('hex'))} rounding={'none'}>
```
jpggvilaca:
```suggestion
        <SpectrumColorPicker value={color} onChange={(newColor) => onChange(newColor.toString('hex'))} rounding={'none'}>
```

application/ui/src/components/label-fields/silent-checkbox.component.tsx
```diff
@@ -0,0 +1,27 @@
+// Copyright (C) 2026 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+import { Checkmark } from '@geti/ui/icons';
+import { clsx } from 'clsx';
+
+import classes from './silent-checkbox.module.scss';
+
+interface SilentCheckboxProps {
```
jpggvilaca:
```suggestion
type SilentCheckboxProps = {
```

application/ui/src/components/label-fields/silent-checkbox.component.tsx
```diff
@@ -0,0 +1,27 @@
...
+    isSelected: boolean;
+    onChange: () => void;
+    'aria-label': string;
+}
+
+export const SilentCheckbox = ({ isSelected, onChange, 'aria-label': ariaLabel }: SilentCheckboxProps) => {
+    return (
+        <button
```
jpggvilaca, line 17: why not use spectrum's button?

application/ui/src/components/label-fields/silent-checkbox.component.tsx
```diff
@@ -0,0 +1,27 @@
...
+            className={clsx(classes.silentCheckbox, { [classes.selected]: isSelected })}
+            onClick={onChange}
+            aria-label={ariaLabel}
+            aria-pressed={isSelected}
+            type='button'
```
jpggvilaca, line 22: np: redundant

application/ui/src/components/label-fields/silent-checkbox.module.scss
```diff
@@ -0,0 +1,23 @@
+.silentCheckbox {
```
jpggvilaca, line 1: wondering if we need a checkbox to achieve this behavior. Perhaps a native html `label for=...` would do?

application/ui/src/features/annotator/labels/edit-label-row/edit-label-row.component.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+import { LabelColorPicker } from '../../../../components/label-fields/label-color-picker.component';
+import { validateLabelName } from '../../../../components/label-fields/label-validation';
+import { SilentCheckbox } from '../../../../components/label-fields/silent-checkbox.component';
+import type { Label } from '../../../../constants/shared-types';
+import { useUpdateLabel } from '../api/use-update-label.hook';
+
+import classes from './edit-label-row.module.scss';
+
+interface EditLabelRowProps {
```
jpggvilaca:
```suggestion
type EditLabelRowProps = {
```

application/ui/src/features/annotator/labels/edit-labels-popover/edit-labels-popover.component.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { Add } from '@geti/ui/icons';
+
+import type { Label } from '../../../../constants/shared-types';
+import { EMPTY_LABEL_ID } from '../../../../shared/annotator/labels';
+import { EditLabelRow } from '../edit-label-row/edit-label-row.component';
+import { NewLabelRow } from '../new-label-row/new-label-row.component';
+
+import classes from './edit-labels-popover.module.scss';
+
+interface EditLabelsPopoverProps {
```
jpggvilaca: type

application/ui/src/features/annotator/labels/labels.component.test.tsx
```diff
@@ -44,6 +44,16 @@ vi.mock('../../../shared/annotator/annotation-actions-provider.component', () =>
     }),
 }));
 
+vi.mock('../../../hooks/api/project.hook', () => ({
```
jpggvilaca: why do we need this one?

application/ui/src/features/annotator/labels/labels.component.tsx
jpggvilaca: Yeah, it's a clear smell. Design flaw.

jpggvilaca (changes requested):
Visually looks and works great. I am just worried about maintaining this. A few notes, going from simple to complex:

1) X-label-row should be X-label-field
2) "edit-labels-popover" should be "labels-popover" since you can do more than just edit
3) all label logic (shared between these new components) could like in its own Class or Hook, like `useLabels` with all the shenanigans associated like validation and stuff. Easily tested and label agnostic. We should keep the UIs as dumb as possible (you almost nailed this part) but at the same time we should not handle a lot of logic everywhere. Centralising sounds better to me. 
4) Popover logic should also be its own thing, unrelated with labels or annotators, just a logic to open/close based on your conditions and with your handlers and refs and whatnot.

application/ui/src/features/annotator/labels/edit-label-row/edit-label-row.component.tsx
ActiveChooN: I can see that we have both options in the project. Do we have any convention in that regard?

jpggvilaca: it was a discussion some time ago. We prefer `type`

application/ui/src/components/label-fields/silent-checkbox.component.tsx
ActiveChooN: Since we are using a generic button, it's better to leave it as it is, since it's neither "submit" nor "reset."

application/ui/src/components/label-fields/silent-checkbox.component.tsx
ActiveChooN: I think it would be too much of a hassle to deal with the button or actionButton styles. Semantically, it's not used for what Spectrum buttons are used for.

application/ui/src/components/label-fields/silent-checkbox.module.scss
ActiveChooN: Could you elaborate on this? Are you suggesting a hidden native checkbox with a "label for"?
IMO it's a toggle button with checkbox styling, so it should fit here.

application/ui/src/features/annotator/labels/labels.component.test.tsx
ActiveChooN: leftovers

application/ui/src/components/label-fields/silent-checkbox.module.scss
jpggvilaca: From the designs i dont see any checkbox so i guess the whole item could be clickable to select it. But looking a second time at it, it has some many elements that it's better to have a single place to click it, like a checkbox indeed. But could we make it more visible on hover? It blends a lot with the bg

ActiveChooN:
> X-label-row should be X-label-field

Why do it should be a "label-field" when we also have a few buttons and a color picker here?

> labels-popover

Good point, renamed

> `useLabels`

Sounds legitimate, extracted to a separate hook

> Popover logic should also be its own thing

But it's basically two `overlayTriggerState` with a couple of calls. Why do we need to extract them?

jpggvilaca:
> > X-label-row should be X-label-field
> 
> Why do it should be a "label-field" when we also have a few buttons and a color picker here?
> 
> > labels-popover
> 
> Good point, renamed
> 
> > `useLabels`
> 
> Sounds legitimate, extracted to a separate hook
> 
> > Popover logic should also be its own thing
> 
> But it's basically two `overlayTriggerState` with a couple of calls. Why do we need to extract them?

1) Because it's mainly a TextField. Rows usually imply tables or columns, which isnt the case. Speaking about that, it would be a good idea to use Grid here instead of Flex actually
2) The point is to, if we need to move your label stuff to another screen, we dont need to worry about popovers and multiple refs. Labels and popover should not live together, unless we're at the container level, i.e.

MyComponent
    > Popover
          > Labels inside
          
Right now labels and popover are like songoku and vegeta, merged together

application/ui/src/components/label-fields/silent-checkbox.module.scss
ActiveChooN: Added actionButton-like border

ActiveChooN:
> Because it's mainly a TextField. Rows usually imply tables or columns

Mainly, but not limited by. And it's indeed a column element

> Labels and popover should not live together

Extracted

jpggvilaca (approved):

Issue #5407, opened 2026-02-03
Make drawing box the default tool for non-classification projects


jpggvilaca:
Right now it's selection, we want drawing box

PR #5419, opened 2026-02-04
feat: Annotator default tool

camiloHimura:
### Summary
#5407 Select the default tool based on the project type

### Checklist


application/ui/src/features/project/task-type-guards.ts
```diff
@@ -6,3 +6,7 @@ import { TaskType } from '../../constants/shared-types';
 export const isClassificationTask = (taskType: TaskType | null): boolean => {
     return taskType === 'classification';
 };
+
+export const isSegmentationTask = (taskType: TaskType | null): boolean => {
+    return taskType === 'instance_segmentation';
+};
```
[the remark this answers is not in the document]
jpggvilaca: We only have 1 type of segmentation now, right?

application/ui/src/shared/annotator/annotator-provider.component.tsx
```diff
@@ -47,8 +47,21 @@ const useSelectedLabel = () => {
     };
 };
 
+const getDefaultTool = (taskType: TaskType | null): ToolType => {
+    if (isClassificationTask(taskType)) {
+        return 'selection';
+    }
+
+    if (isSegmentationTask(taskType)) {
+        return 'polygon';
```
jpggvilaca, line 56: we also have bounding box now for segmentation, so i think that one should be the default one. wdyt?

application/ui/src/features/project/task-type-guards.ts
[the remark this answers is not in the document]
camiloHimura: yes, only one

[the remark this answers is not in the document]
jpggvilaca: ok so we dont need 2 functions for this?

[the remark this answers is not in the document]
camiloHimura: not at all, only one

application/ui/src/shared/annotator/annotator-provider.component.tsx
dwesolow: IMO we can stay with polygon or even with SAM.
TBH I'd pick SAM as default for every task.

application/ui/src/shared/annotator/annotator-provider.component.tsx
```diff
@@ -47,8 +47,21 @@ const useSelectedLabel = () => {
     };
 };
 
+const getDefaultTool = (taskType: TaskType | null): ToolType => {
+    if (isClassificationTask(taskType)) {
+        return 'selection';
```
dwesolow: For classification we don't display any tool, no?

application/ui/src/shared/annotator/annotator-provider.component.tsx
```diff
@@ -47,8 +47,21 @@ const useSelectedLabel = () => {
...
+        return 'polygon';
+    }
+
+    return 'bounding-box';
+};
+
 export const AnnotatorProvider = ({ mediaItem, children }: { mediaItem: Media; children: ReactNode }) => {
-    const [activeTool, setActiveTool] = useState<ToolType>('selection');
+    const { data: selectedProject } = useProject();
+    const [activeTool, setActiveTool] = useState<ToolType>(getDefaultTool(selectedProject?.task.task_type));
```
dwesolow: When we call a function inside the `useState`, that function is called on every rerender. If we do `useState(() => getDefaultTool(...))` then that function is called only once.

application/ui/src/shared/annotator/annotator-provider.component.tsx
jpggvilaca: We dont but i guess we still need a valid tool for the hook

application/ui/src/shared/annotator/annotator-provider.component.tsx
jpggvilaca: BBox is the simplest and requires no webworkers, that way we also stay consistent. SAM is considered "Advanced" or "Auto" mode so it should not be default imho

dwesolow: Okok. I don't see bbox tool  in segmentation tho.

application/ui/src/shared/annotator/annotator-provider.component.tsx
dwesolow: It's a bit misleading, it indicates that we support that tool. We should update the hook to accept `null/undefined`.

application/ui/src/shared/annotator/annotator-provider.component.tsx
jpggvilaca: Yeah we need to add it. Sorry @camiloHimura  i thought i added on my previous PR but i didnt

application/ui/src/shared/annotator/annotator-provider.component.tsx
jpggvilaca: that can also be a solution

application/ui/src/shared/annotator/annotator-provider.component.tsx
camiloHimura: Discussed offline, since segmentation projects don’t support rectangle shapes, we won’t add the bounding-box tool for now.

application/ui/src/shared/annotator/annotator-provider.component.tsx
camiloHimura: updated

application/ui/src/shared/annotator/annotator-provider.component.tsx
camiloHimura: updated

jpggvilaca (dismissed):

dwesolow (dismissed):

jpggvilaca (approved):

dwesolow (approved):

PR #5424, opened 2026-02-04
Read-only or "simplified" annotator

jpggvilaca:
### Summary

* This annotator is not MUUUUCH simpler but it's some progress. It will be used on the main annotator while on prediction mode and on training dataset gallery media items. This last part will come after I manage to train a model and to actually test it. For now I put the first part for review.
* Also included a bunch of other small fixes. Checking commit by commit is easier

### Checklist

### DIFF
--- a/application/ui/src/features/dataset/gallery/gallery.component.tsx
+++ b/application/ui/src/features/dataset/gallery/gallery.component.tsx
@@ -69,6 +69,7 @@ export const Gallery = ({ items, hasNextPage, isFetchingNextPage, fetchNextPage 
                                 justifyContent={'center'}
                             >
                                 <Checkbox
+                                    aria-label={'Select media item ' + item.name}
                                     onChange={() => toggleSelectedKeys([String(item.id)])}
                                     isSelected={isSetSelectedKeys && selectedKeys.has(String(item.id))}
                                 />
--- a/application/ui/src/features/dataset/gallery/gallery.component.tsx
+++ b/application/ui/src/features/dataset/gallery/gallery.component.tsx
@@ -69,7 +69,7 @@ export const Gallery = ({ items, hasNextPage, isFetchingNextPage, fetchNextPage 
                                 justifyContent={'center'}
                             >
                                 <Checkbox
-                                    aria-label={'Select media item ' + item.name}
+                                    aria-label={`Select media item ${item.name}`}
                                     onChange={() => toggleSelectedKeys([String(item.id)])}
                                     isSelected={isSetSelectedKeys && selectedKeys.has(String(item.id))}
                                 />
--- a/application/ui/src/features/dataset/media-preview/bottom-toolbar/bottom-toolbar.component.tsx
+++ b/application/ui/src/features/dataset/media-preview/bottom-toolbar/bottom-toolbar.component.tsx
@@ -20,6 +20,7 @@ import classes from './bottom-toolbar.module.scss';
 type BottomToolbarProps = {
     isUserReviewed: boolean;
     mediaItem: Media;
+    hideHotkeys?: boolean;
 };
 
 const useSubsets = (mediaItemId: string) => {
@@ -58,7 +59,7 @@ const useSubsets = (mediaItemId: string) => {
     return { currentSubset: data?.subset ?? null, handleSubsetChange };
 };
 
-export const BottomToolbar = ({ isUserReviewed, mediaItem }: BottomToolbarProps) => {
+export const BottomToolbar = ({ isUserReviewed, mediaItem, hideHotkeys }: BottomToolbarProps) => {
     const fileName = `${mediaItem.name}.${mediaItem.format} (${mediaItem.width} x ${mediaItem.height} px)`;
 
     const { currentSubset, handleSubsetChange } = useSubsets(mediaItem.id);
@@ -67,9 +68,11 @@ export const BottomToolbar = ({ isUserReviewed, mediaItem }: BottomToolbarProps)
         <Flex justifyContent={'end'}>
             <Toolbar.Container>
                 <Grid autoFlow={'column'} autoColumns={'max-content'} gap={'size-50'}>
-                    <Toolbar.Section>
-                        <Hotkeys />
-                    </Toolbar.Section>
+                    {!hideHotkeys && (
+                        <Toolbar.Section>
+                            <Hotkeys />
+                        </Toolbar.Section>
+                    )}
 
                     <Toolbar.Section>
                         <Flex gap={'size-100'} alignItems={'center'}>
--- a/application/ui/src/features/dataset/media-preview/media-preview.component.tsx
+++ b/application/ui/src/features/dataset/media-preview/media-preview.component.tsx
@@ -49,18 +49,6 @@ const MediaPreviewContent = ({ items, mediaItem, onSelectedMediaItem, onClose }:
     const isUserReviewed = annotationsData?.user_reviewed ?? false;
     const annotationsDTO = annotationsData?.annotations ?? [];
 
-    if (mode === 'prediction') {
-        return (
-            <ReadOnlyAnnotator
-                mediaItem={mediaItem}
-                annotationsDTO={getInitialPredictions(mode, isUserReviewed, annotationsDTO)}
-                isUserReviewed={isUserReviewed}
-                onModeChange={setMode}
-                onClose={onClose}
-            />
-        );
-    }
-
     return (
         <AnnotatorProviders
             key={mediaItem.id}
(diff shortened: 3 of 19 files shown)


application/ui/src/features/dataset/gallery/gallery.component.tsx
```diff
@@ -69,6 +69,7 @@ export const Gallery = ({ items, hasNextPage, isFetchingNextPage, fetchNextPage
                                 justifyContent={'center'}
                             >
                                 <Checkbox
+                                    aria-label={'Select media item ' + item.name}
```
camiloHimura:
```suggestion
                                    aria-label={`Select media item ${item.name}`}
```

camiloHimura:
It looks fine overall; however, zooming via mouse scroll doesn’t work in prediction mode

application/ui/src/features/dataset/gallery/gallery.component.tsx
jpggvilaca: Updated

jpggvilaca:
> It looks fine overall; however, zooming via mouse scroll doesn’t work in prediction mode

That is because of the pointer-events being none. You can still manipulate zoom via the zoom component on the bottom bar. I didnt find a clean way to get it work so im willing to live with it. Unless there are better suggestions

dwesolow:
I haven't check the code yet, only saw the screenshot: in prediction mode there needs to be a `Confirm prediction` button.

application/ui/src/features/dataset/media-preview/media-preview.module.scss
```diff
@@ -1,7 +1,5 @@
-.readOnlyCanvas {
-    & * {
-        pointer-events: none !important;
-    }
+.primaryToolbar {
```
dwesolow: I think that's not needed anymore

dwesolow:
> > It looks fine overall; however, zooming via mouse scroll doesn’t work in prediction mode
> 
> That is because of the pointer-events being none. You can still manipulate zoom via the zoom component on the bottom bar. I didnt find a clean way to get it work so im willing to live with it. Unless there are better suggestions

I had to disable pointer events because there was the issue that user was allowed to select and edit annotation.
If we really want to have separated read only mode, we can go further and separate providers, which ones are only needed for the read only view. Moreover, to handle the issue that I mentioned with annotation selecting and editing, we can remove that part from the `Annotation` (that was the goal of doing composable providers).
Those changes will require more work and we will need to have really good tests to have easier life while maintaining it.

jpggvilaca:
> > > It looks fine overall; however, zooming via mouse scroll doesn’t work in prediction mode
> > 
> > 
> > That is because of the pointer-events being none. You can still manipulate zoom via the zoom component on the bottom bar. I didnt find a clean way to get it work so im willing to live with it. Unless there are better suggestions
> 
> I had to disable pointer events because there was the issue that user was allowed to select and edit annotation. If we really want to have separated read only mode, we can go further and separate providers, which ones are only needed for the read only view. Moreover, to handle the issue that I mentioned with annotation selecting and editing, we can remove that part from the `Annotation` (that was the goal of doing composable providers). Those changes will require more work and we will need to have really good tests to have easier life while maintaining it.

I do think the pointer events should be there and is a simple and good solution.
I looked into separating but the bottom toolbar does need quite a few (nearly the same ones) providers so it's not as straight forward. I started by creating simpler duplicates but realised the gains were not as much, and it would be very hard to maintain. 

What we can do is to create a ReadOnlyCanvas and ReadOnlyAnnotations without the EditableAnnotation and SelectableAnnotation (as you suggested) but that will be a bit hard to maintain. But if we all agree, we can go towards that direction. WDYT?

application/ui/src/features/dataset/media-preview/media-preview.module.scss
jpggvilaca: Indeed! Deleted the file

jpggvilaca:
> I haven't check the code yet, only saw the screenshot: in prediction mode there needs to be a `Confirm prediction` button.

lol true. I was handling the SubsetGallery edge case and forgot. Will fix. ty!

application/ui/src/features/models/model-listing/model-listing.container.tsx
```diff
@@ -25,7 +25,6 @@ const ModelListingContent = () => {
                 direction={'column'}
                 height={'100%'}
                 alignItems={'center'}
-                justifyContent={'center'}
```
jpggvilaca, line 28: bonus fix: When the model is training on an empty project, the "CurrentTraining" was vertically in the middle of the page.

application/ui/src/features/dataset/media-preview/bottom-toolbar/bottom-toolbar.component.tsx
```diff
@@ -58,7 +59,7 @@ const useSubsets = (mediaItemId: string) => {
     return { currentSubset: data?.subset ?? null, handleSubsetChange };
 };
 
-export const BottomToolbar = ({ isUserReviewed, mediaItem }: BottomToolbarProps) => {
+export const BottomToolbar = ({ isUserReviewed, mediaItem, hideHotkeys }: BottomToolbarProps) => {
```
camiloHimura, line 62: np. `!hideHotkeys` feels like a double negative and is a bit confusing at a quick glance. What about renaming it to `hasHotkeys` and only rendering when it’s true?

application/ui/src/features/dataset/media-preview/secondary-toolbar/secondary-toolbar.component.tsx
```diff
@@ -1,80 +1,47 @@
...
-    mode,
     onModeChange,
+    onAcceptPrediction,
 }: SecondaryToolbarProps) => {
-    const queryClient = useQueryClient();
-    const { setMediaState } = useSelectedData();
     const { data: selectedProject } = useProject();
 
-    const { annotations, isSaving, submitAnnotations } = useAnnotationActions();
+    const { canSubmit, isSaving, submit } = useSubmitPredictions({ onSuccess: onAcceptPrediction });
```
camiloHimura, line 41: I think naming it `onSubmit` instead of `submit` makes it clearer that an action will be triggered. Just an observation, no need to address it.

application/ui/src/features/dataset/media-preview/bottom-toolbar/bottom-toolbar.component.tsx
jpggvilaca: The reason is that the positive is the default. So hiding the hotkeys is opt out.

jpggvilaca:
> It looks fine overall; however, zooming via mouse scroll doesn’t work in prediction mode

should work now

application/ui/src/features/dataset/media-preview/media-preview.component.tsx
```diff
@@ -53,60 +55,74 @@ const MediaPreviewContent = ({ items, mediaItem, onSelectedMediaItem, onClose }:
...
-                                <AnnotatorCanvasSettings>
-                                    <AnnotatorCanvas mediaItem={mediaItem} />
-                                </AnnotatorCanvasSettings>
-                            </View>
-                        </CanvasSettingsProvider>
-                    </AnnotatorProvider>
-                </AnnotationVisibilityProvider>
-            </SelectAnnotationProvider>
-        </AnnotationActionsProvider>
+            {mode === 'prediction' ? (
```
camiloHimura, line 88: At some point, would we want to persist the prediction mode in the URL?

camiloHimura (dismissed):

ActiveChooN (commented):
Bug: I was able to draw an annotation in prediction mode. Steps to reproduce:
1. Select Drawing BBS.
2. Go to prediction mode.
3. Crosshairs will be displayed, and you will be able to draw.
4.  Go back to the annotator; the annotation will be there.

https://github.com/user-attachments/assets/39931419-195a-4c12-84d0-8cf8d0d9bd96

jpggvilaca:
> Bug: I was able to draw an annotation in prediction mode. Steps to reproduce:
> 
> 1. Select Drawing BBS.
> 2. Go to prediction mode.
> 3. Crosshairs will be displayed, and you will be able to draw.
> 4. Go back to the annotator; the annotation will be there.
> 
>  Screen.Recording.2026-02-05.at.19.04.58.mov

ah, might be because of the DEFAULT tool that we merged. Good catch thank you!

application/ui/src/features/dataset/media-preview/media-preview.component.tsx
jpggvilaca: I think so, yes

jpggvilaca:
> Bug: I was able to draw an annotation in prediction mode. Steps to reproduce:
> 
> 1. Select Drawing BBS.
> 2. Go to prediction mode.
> 3. Crosshairs will be displayed, and you will be able to draw.
> 4. Go back to the annotator; the annotation will be there.
> 
>  Screen.Recording.2026-02-05.at.19.04.58.mov

actually unrelated with this PR but it's fixed now on my last commit

ActiveChooN (approved):

PR #5453, opened 2026-02-06
Subset gallery view

jpggvilaca:
### Summary

* Add media dialog view (read only annotator) for subset items
* Also fixed a few routing issues pointed by Leonardo

Tests for model details will follow (will create a new spec to avoid having a million lines on models.spec)

Screenshot 2026-02-06 at 17 08 09
Screenshot 2026-02-06 at 17 08 20

rec

### Checklist


application/ui/src/shared/dataset-revision-utils.ts
```diff
@@ -0,0 +1,30 @@
...
+ * They share the same ID as the original media item, allowing us to fetch
+ * annotations from the main dataset endpoint.
+ *
+ * This conversion adds the required Media fields (name, type, size, source_id)
+ * that are needed by the annotator components.
+ *
+ * NOTE: In the future the whole media/datasetItem/datasetRevisionItem will hopefully be
+ * merged and we wont need multiple endpoints to get the full dataset of the item.
+ */
+export const datasetRevisionItemToMedia = (item: Partial<DatasetRevisionItem> & { id: string }): Media => {
```
dwesolow: Why `item.format,width,height` are optional?
What about video file, will we need to update the `type` to `frame/video` and `format` to XYZ? If yes, how we would know about that?

jpggvilaca: The types are a bit messy but in this case the open api spec is actually wrong. I also suspected but didnt care too much, but now that you commented i went to look deeper and the backend returns the correct type but the spec is wrong. So i reverted all the Partial's and typed the hook itself `useGetDatasetRevisionItems`. Thanks! As for frame/video, that is a conversation for another day ^^

jpggvilaca: Fix here: https://github.com/open-edge-platform/training_extensions/pull/5455#discussion_r2781561914

application/ui/src/hooks/use-get-dataset-revision-items.hook.ts
```diff
@@ -47,7 +47,8 @@ export const useGetDatasetRevisionItems = ({ datasetRevisionId, subset }: UseGet
         }
     );
 
-    const items = data?.pages.flatMap((page) => page.items) ?? [];
+    // Note: OpenAPI spec incorrectly types this as DatasetItem[], but the API actually returns DatasetRevisionItem[]
+    const items: DatasetRevisionItem[] = (data?.pages.flatMap((page) => page.items) ?? []) as DatasetRevisionItem[];
```
dwesolow, line 52: Will it be fixed in https://github.com/open-edge-platform/training_extensions/pull/5455#discussion_r2781561914?

application/ui/src/shared/dataset-revision-utils.ts
```diff
@@ -0,0 +1,30 @@
+// Copyright (C) 2025 Intel Corporation
```
dwesolow: why shared?

application/ui/src/hooks/use-get-dataset-revision-items.hook.ts
```diff
@@ -4,7 +4,7 @@
 import { useProjectIdentifier } from 'hooks/use-project-identifier.hook';
 
 import { $api } from '../api/client';
-import { DatasetSubset } from '../constants/shared-types';
+import { DatasetRevisionItem, DatasetSubset } from '../constants/shared-types';
```
dwesolow: `type`

application/ui/src/hooks/use-get-dataset-revision-items.hook.ts
jpggvilaca: Yeah

dwesolow (dismissed):

application/ui/src/hooks/use-get-dataset-revision-items.hook.ts
jpggvilaca: Updated

application/ui/src/shared/dataset-revision-utils.ts
jpggvilaca: Moved to training-datasets folder

application/ui/src/hooks/use-get-dataset-revision-items.hook.ts
jpggvilaca: Added TODO

dwesolow (approved):

PR #5462, opened 2026-02-09
feat: Display performance metrics

dwesolow:
### Summary

This PR handles the name and value of the metric that we want to show in the models page.
Algorithm is simple:
1. Get testing set.
2. Find metric that has `primary: true`.
3. Use that metric's name and value. Name is the column name, value is column value.

Moreover I fixed sorting by model size and score.

image

### Checklist

### DIFF
--- a/application/ui/src/features/models/model-listing/components/model-row/utils.ts
+++ b/application/ui/src/features/models/model-listing/components/model-row/utils.ts
@@ -1,0 +1,30 @@
+// Copyright (C) 2025-2026 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+import { type Model } from '../../../../../constants/shared-types';
+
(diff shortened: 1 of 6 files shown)


application/ui/src/features/models/model-listing/components/model-row/utils.ts
```diff
@@ -0,0 +1,32 @@
+// Copyright (C) 2025-2026 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+import { type Model } from '../../../../../constants/shared-types';
+
+export const getTestingMetric = (model: Model): { name: string; value: number } | undefined => {
```
camiloHimura, line 6: Relying on the default function return is fine, but for clarity it’s better to be explicit.

dwesolow: Updated

jpggvilaca (approved):

PR #5513, opened 2026-02-12
Revamp project seeding

### LINKED ISSUE #5457: Revamp DB seeding
The current [DB seeding script](https://github.com/open-edge-platform/training_extensions/blob/develop/application/backend/app/db_seeder.py), which initializes the application with sample data for development purposes, require each table to be populated with hard-coded data. As the application grows, it is increasingly costly to maintain these scripts, and the seeded projects are often incomplete and buggy.

The goal of this ticket is to refactor the seeding logic. Ideally, projects should be stored as zip archives in a remote archive, and the seeding script should simply download the projects and import them.

leoll2:
### Summary

This PR overhauls our mechanism to populate the application with demo projects, for development and testing purposes.
In the previous approach, the DB would be filled with manually constructed data, and a bunch of individual binary files would be downloaded and put in the data folder. As the application and data models grow in complexity, this approach is no longer sustainable because it requires a lot of entities to be initialized and their relationships aren't always straightforward.
The new approach is more agnostic to the internal data model, and the whole demo project is distributed as a single zip file for convenience. It consists of the following elements:
- An export script can be used to export any existing project as a zip archive, containing a dump of the project-related DB data and all its binaries (models, dataset revisions, ...). The DB records are dumped to JSON files, one per table.
- An import script can be used to extract the zip file and reconstruct the full project state in the application. The imported project is a perfectly identical copy of the original one.
- The projects archives are stored in S3, at a publicly accessible URL.
- The `run.sh` downloads the demo projects and imports them.

Tasks:
- [x] Implement import/export scripts
- [x] Update `run.sh` logic
- [x] Create and upload one demo project to S3 for each task type

_Note: the code does not have automated tests because the project import/export functionality is not production-ready, it's exclusively for development purposes. In the future, it may be further refined and exposed to users._

Resolves https://github.com/open-edge-platform/training_extensions/issues/5457

demo-projects

### How to test

Try yourself with `./run.sh --clean --setup-demo`.

If you want to export a project: `uv run app/cli.py export-project --project-id <id> --output <zip-file>`

To import a project: `uv run app/cli.py import-project --input <zip-file>`

### Checklist


application/backend/app/db/import_export/import_project.py
```diff
@@ -0,0 +1,303 @@
...
+    parser.add_argument(
+        "-f",
+        "--force-import",
+        action="store_true",
+        help="Bypass database schema version check and attempt import anyway (use at your own risk)",
+    )
+
+    args = parser.parse_args()
+
+    import_project(input_archive=args.input, allow_mismatching_db_schema=args.force_import)
```
itallix: Does it make more sense to include it into `cli.py` and define as `click` commands?

itallix (commented):
Nice work! What about adding sources/sinks to the existing demos so we can immediately test inference via WebRTC? Including 3 different source types would showcase various integration patterns.

application/backend/app/db/import_export/import_project.py
leoll2: Good idea, I will try to move it there

leoll2:
> Nice work! What about adding sources/sinks to the existing demos so we can immediately test inference via WebRTC? Including 3 different source types would showcase various integration patterns.

Thanks. Regarding the sources, I'm pondering about a few things:
- Which sources to add? I think only videos make sense, because camera is HW dependent and folders isn't very usable (we can borrow UX/backend improvements from Prompt for it).
- Where to initialize the sources, since they are not part of the projects? Maybe reintroduce a small seeding command in the CLI, but only for sources (e.g. `cli seed-sources`)?

itallix:
> * Which sources to add? I think only videos make sense, because camera is HW dependent and folders isn't very usable (we can borrow UX/backend improvements from Prompt for it).

Indeed, videos are the most relevant. Then, we should probably have 3 sources with videos tailored for each project.

> * Where to initialize the sources, since they are not part of the projects? Maybe reintroduce a small seeding command in the CLI, but only for sources (e.g. `cli seed-sources`)?

A small command makes sense. We can probably keep it agnostic to specific resource type, and use it as an entry point for various non-project resources - sinks can be added there as well (e.g. `pre-seed`).

application/backend/app/db/import_export/import_project.py
leoll2: Done

leoll2:
> Then, we should probably have 3 sources with videos tailored for each project.

Done, I added a short video for each project.

> A small command makes sense. We can probably keep it agnostic to specific resource type, and use it as an entry point for various non-project resources - sinks can be added there as well (e.g. pre-seed).

I added a command `setup_demo_sources`; if later we need to initialize more non-project resources, we can rename/encapsulate it in a `pre-seed` command.

itallix:
> I added a command `setup_demo_sources`; if later we need to initialize more non-project resources, we can rename/encapsulate it in a `pre-seed` command.

There are still few steps before we can enable inference and visualize the stream:

  1. Make a model active in the project
  2. Connect a relevant source
  3. Create and connect a new sink

Shall we preconfigure these steps to speed up demo onboarding?

leoll2:
Thanks for the suggestions! I re-uploaded the archives to S3 after making all models 'active'. I also extended the initialization script to create a folder sink.
As for the source/sink selection and pipeline enablement, I prefer to keep these as manual steps.

itallix:
> after making all models 'active'. 

Model is active only in one project:
Screenshot 2026-02-16 at 09 36 50

leoll2:
> > after making all models 'active'.
> 
> Model is active only in one project:

Please delete your `data/.demo_cache/` folder, if you haven't already. It could be that your locally cached projects are stale.

itallix (approved):

PR #5518, opened 2026-02-12
Added media duration field

### LINKED ISSUE #5516: Add video duration to the video response
Please include video duration in the video response (together with `frame_count` etc).

warrkan:
Resolves #5516


application/backend/app/api/schemas/media.py
```diff
@@ -22,6 +22,12 @@ class MediaView(BaseRequiredIDNameModel):
     frame_count: int | None
     source_id: UUID | None = None
 
+    @computed_field
+    @property
+    def duration(self) -> int | None:
+        """Return duration in seconds"""
+        return int(self.frame_count // self.fps) if self.frame_count is not None and self.fps is not None else None
```
leoll2: Why do you return an int instead of a float?

warrkan: I don't think that UI needs a fractional duration, does it?

leoll2: I don't know how the UI will use this attribute, but in general it's better to return accurate information from the API and let the UI do the rounding, if needed.

CC @dwesolow

jpggvilaca (dismissed):

leoll2 (approved):

ActiveChooN (approved):

PR #5526, opened 2026-02-13
Monitor stale jobs

### LINKED ISSUE #5392: Training in Tune fails but job still running
1. Start by running `compose up`
2. Import several images. Annotate them.
3. Start the training. As the option was disabled in the UI, I ran the curl: `curl -X POST -H "Content-Type: application/json" localhost:8080/api/jobs -d '{"job_type": "train", "project_id": "178293b8-a0ed-4de8-b688-655dee4d31c0", "parameters": { "model_architecture_id": "Custom_Object_Detection_Gen3_ATSS", "device": "cpu" }}' | jq`

The training crashes (see #5391) but the job is still running, which can be seen in both the UI and the jobs endpoint:
```
user@localhost:~$ curl localhost:8080/api/jobs | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   440  100   440    0     0   119k      0 --:--:-- --:--:-- --:--:--  143k
[
  {
    "job_id": "03e02d38-5133-40c4-bc34-77f62f6c2f79",
    "job_type": "train",
    "metadata": {
      "project": {
        "id": "178293b8-a0ed-4de8-b688-655dee4d31c0"
      },
      "model": {
        "id": "89483028-709a-4faf-b57f-cdc9e66f66a3",
        "architecture": "Custom_Object_Detection_Gen3_ATSS",
        "parent_revision_id": null,
        "dataset_revision_id": null
      }
    },
    "status": "RUNNING",
    "progress": 10.0,
    "message": "Started: Train Model",
    "error": null,
    "started_at": "2026-02-03T08:08:17.312874Z",
    "finished_at": null
  }
]
```
The status should be FAILED (or similar).

itallix:
### Summary

This PR adds a new task in `JobController` that monitors stale jobs and schedules cancellation if detected.

### Checklist

### DIFF
--- a/application/backend/app/core/jobs/control_plane/queue.py
+++ b/application/backend/app/core/jobs/control_plane/queue.py
@@ -69,7 +69,7 @@ class JobQueue:
 
     def list_stale_jobs(self, threshold_seconds: int = 3600) -> list[Job]:
         """
-        List jobs that are in a stale state (e.g., RUNNING for too long).
+        List jobs that are in a stale state (determined by being in RUNNING status for longer than the threshold).
 
         Args:
             threshold_seconds: Time in seconds to consider a job stale
--- a/application/backend/app/core/jobs/control_plane/queue.py
+++ b/application/backend/app/core/jobs/control_plane/queue.py
@@ -7,7 +7,7 @@ from uuid import UUID
 
 from loguru import logger
 
-from app.core.jobs.models import Job, JobStatus
+from app.core.jobs.models import Job, JobStatus, now_utc_ts
 
 
 class CancellationResult(StrEnum):
@@ -66,6 +66,23 @@ class JobQueue:
     def list_non_completed(self) -> list[Job]:
         """List all non-completed jobs."""
         return [job for job in self._by_id.values() if job.status < JobStatus.DONE]
+
+    def list_stale_jobs(self, threshold_seconds: int = 3600) -> list[Job]:
+        """
+        List jobs that are in a stale state (e.g., RUNNING for too long).
+
+        Args:
+            threshold_seconds: Time in seconds to consider a job stale
+
+        Returns:
+            List of stale jobs that are currently in a running state
+        """
+        current_time = now_utc_ts()
+        return [
+            job
+            for job in self._by_id.values()
+            if job.status == JobStatus.RUNNING and (current_time - job.updated_at) > threshold_seconds
+        ]
 
     def cancel(self, job_id: UUID) -> tuple[Job | None, CancellationResult]:
         """Attempt to cancel a job by its ID."""
--- a/application/backend/app/core/jobs/control_plane/controller.py
+++ b/application/backend/app/core/jobs/control_plane/controller.py
@@ -24,14 +24,16 @@ class JobController:
 
     Architecture:
         - Supervisor loop: Continuously polls the job queue for runnable jobs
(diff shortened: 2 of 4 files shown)


application/backend/app/core/jobs/control_plane/queue.py
```diff
@@ -67,6 +67,23 @@ def list_non_completed(self) -> list[Job]:
         """List all non-completed jobs."""
         return [job for job in self._by_id.values() if job.status < JobStatus.DONE]
 
+    def list_stale_jobs(self, threshold_seconds: int = 3600) -> list[Job]:
+        """
+        List jobs that are in a stale state (e.g., RUNNING for too long).
```
leoll2:
```suggestion
        List jobs that are in a stale state (e.g., RUNNING and not updated for a relatively long time).
```

application/backend/app/core/jobs/control_plane/queue.py
```diff
@@ -67,6 +67,23 @@ def list_non_completed(self) -> list[Job]:
...
+            threshold_seconds: Time in seconds to consider a job stale
+
+        Returns:
+            List of stale jobs that are currently in a running state
+        """
+        current_time = now_utc_ts()
+        return [
+            job
+            for job in self._by_id.values()
+            if job.status == JobStatus.RUNNING and (current_time - job.updated_at) > threshold_seconds
```
leoll2, line 84: Does the value of `updated_at` also get updated when a job reports a progress percentage that didn't increase? Example:
- at 10:00 am, the job reports progress=80%
- at 11:00 am, the job reports progress=80% again (same value)

This job can be considered stuck, since it remained at 80% for an hour or more. My question is: can your mechanism, based on `updated_at`, detect this situation?

application/backend/app/core/jobs/control_plane/queue.py
itallix: mentioned the threshold

application/backend/app/core/jobs/control_plane/queue.py
itallix: The current mechanism assumes that if a job process is alive and communicating, it will trigger an update to the job's status, which includes updating the `updated_at` timestamp. This happens regardless of whether the progress percentage has increased.

Therefore, in your scenario:
1. At 10:00 am, when the job reports 80% progress, `updated_at` is set to 10:00 am.
2. At 11:00 am, when the job reports 80% progress again, `updated_at` is updated to 11:00 am.

This means the `list_stale_jobs method`, as it is currently written, would not detect this job as stale, because the `updated_at` field is being refreshed. The mechanism is designed to detect jobs that have stopped communicating entirely (i.e., are truly stuck or have crashed), not jobs that are communicating but failing to make progress.

leoll2: Alright, thanks for clarifying. I was asking because, in theory, a job could be implemented with separate worker and progress reporting threads; if the worker gets stuck, the progress reporter stays alive but it outputs the same percentage over and over. In practice, I don't remember if our training job is implemented that way: if not, this scenario can be ignored.

itallix: This is not the case at the moment.

leoll2 (approved):

PR #5533, opened 2026-02-13
Logging utilities and hooks for model logs

ActiveChooN:
### Summary

Add training logs data layer: 
- SSE streaming hook for live logs, 
- React Query hook for completed model logs, 
- Loguru-typed log format,  
- Parsing/filtering utilities

These will be used in the upcoming PR with components and integration

### Checklist


application/ui/src/features/models/training-logs/hooks/use-stream-job-logs.hook.ts
```diff
@@ -0,0 +1,39 @@
...
+import { useSSE } from '../../../../hooks/use-sse.hook';
+import { type LogEntry } from '../log-types';
+
+type UseStreamJobLogsReturn = {
+    logs: LogEntry[];
+    isConnected: boolean;
+    error: Error | null;
+};
+
+export const useStreamJobLogs = (jobId: string | undefined): UseStreamJobLogsReturn => {
```
jpggvilaca, line 15: where are we going to use this? I only see it being used in tests

application/ui/src/features/models/training-logs/hooks/use-stream-job-logs.hook.ts
```diff
@@ -0,0 +1,39 @@
+// Copyright (C) 2026 Intel Corporation
```
jpggvilaca, line 1: why do we need "stream" in the name? I think it would be more readable to state that these are job logs. Same way we have model-logs. Lastly i think LOGS is a feature in itself so we should move it to features/logs instead of being inside models. We might want more/different logs for different screens but even if we dont, let's keep entities separated

jpggvilaca: Aside from that, everything looks good but im reviewing a bit in the dark because i dont know how the integration works yet. But i guess we'll see it on next PRs

application/ui/src/features/models/training-logs/hooks/use-stream-job-logs.hook.ts
ActiveChooN: In LogViewer for ongoing training jobs

application/ui/src/features/models/training-logs/hooks/use-stream-job-logs.hook.ts
ActiveChooN: > why do we need "stream" in the name?

Because I want to make it clear that it's SSE integration, not a query.

> i think LOGS is a feature in itself so we should move it to features/logs

We only have 5 top-level features: annotator, dataset, inference, models, and project. I don't think this feature would be equally complex. Besides that, it's only used for trained models or training jobs, so I think it fits well under the "models" feature.

> i dont know how the integration works yet.

You can check the browser console on the Models page. There will be logs for the trained model, as well as for any training jobs you have running.

application/ui/src/features/models/training-logs/hooks/use-model-logs.hook.ts
```diff
@@ -0,0 +1,40 @@
...
+        .split('\n')
+        .filter((line) => line.trim())
+        .map((line) => parseLogLine(line))
+        .filter((entry): entry is LogEntry => entry !== null);
+};
+
+export const useModelLogs = (modelId: string | undefined) => {
+    const projectId = useProjectIdentifier();
+
+    return useQuery({
```
dwesolow, line 34: why cannot we use query from `openapi-react-query`?

application/ui/src/features/models/training-logs/hooks/use-stream-job-logs.hook.ts
dwesolow: I think that feature might be used for dataset import/export as well.

application/ui/src/features/models/training-logs/hooks/use-stream-job-logs.hook.ts
```diff
@@ -0,0 +1,39 @@
...
+    logs: LogEntry[];
+    isConnected: boolean;
+    error: Error | null;
+};
+
+export const useStreamJobLogs = (jobId: string | undefined): UseStreamJobLogsReturn => {
+    const [logs, setLogs] = useState<LogEntry[]>([]);
+    const [isConnected, setIsConnected] = useState(false);
+    const [error, setError] = useState<Error | null>(null);
+    const logsRef = useRef<LogEntry[]>([]);
```
dwesolow: Why do we need this?

application/ui/src/features/models/training-logs/hooks/use-model-logs.hook.ts
ActiveChooN: There is no reason to do that.
1. The server responds with plain text containing JSON from Loguru that we need to parse on the client side.
2. Because of this, the response is not typed and will return "unknown."

application/ui/src/features/models/training-logs/hooks/use-stream-job-logs.hook.ts
ActiveChooN: Yep, I think it should work just with functional updater. Changed

application/ui/src/features/models/training-logs/hooks/use-stream-job-logs.hook.ts
ActiveChooN: I think we can reconsider its placement when we implement the new log integration.

application/ui/src/features/models/training-logs/hooks/use-model-logs.hook.ts
```diff
@@ -0,0 +1,40 @@
...
+
+import { useQuery } from '@tanstack/react-query';
+import { useProjectIdentifier } from 'hooks/use-project-identifier.hook';
+
+import { API_BASE_URL } from '../../../../api/client';
+import { type LogEntry } from '../log-types';
+import { parseLogLine } from '../log-utils';
+
+const fetchModelLogs = async (projectId: string, modelId: string): Promise<LogEntry[]> => {
+    const response = await fetch(`${API_BASE_URL}/api/projects/${projectId}/models/${modelId}/logs`);
```
dwesolow: we might use `fetchClient` so we don't need to provide the URL, it will get it automatically (might prevent from regressions)

application/ui/src/features/models/training-logs/hooks/use-stream-job-logs.hook.ts
```diff
@@ -0,0 +1,37 @@
...
+    useSSE<LogEntry>(jobId ? `/api/jobs/${jobId}/logs` : undefined, {
+        onMessage: (entry) => {
+            setLogs((prev) => [...prev, entry]);
+        },
+        onOpen: () => {
+            setIsConnected(true);
+        },
+        onError: () => {
+            setIsConnected(false);
+            setError(new Error('SSE connection lost'));
```
dwesolow, line 30: Is it possible that we will get the error for the first time (we set the error) and then we connect successfully? In that case we miss clearing the error (`setError(null)`)

camiloHimura: +1

application/ui/src/features/models/training-logs/hooks/use-model-logs.hook.ts
ActiveChooN: Good point, changed to fetchClient

application/ui/src/features/models/training-logs/hooks/use-stream-job-logs.hook.ts
ActiveChooN: Added error cleanup for onOpen

application/ui/src/features/models/training-logs/hooks/use-model-logs.hook.ts
```diff
@@ -0,0 +1,43 @@
...
+        .filter((line) => line.trim())
+        .map((line) => parseLogLine(line))
+        .filter((entry): entry is LogEntry => entry !== null);
+};
+
+export const useModelLogs = (modelId: string | undefined) => {
+    const projectId = useProjectIdentifier();
+
+    return useQuery({
+        queryKey: [
```
dwesolow: Forgot about this part, please use `getQueryKey`. That's identity function that helps us keeping the type safety for query key.

dwesolow (dismissed):

application/ui/src/features/models/training-logs/hooks/use-model-logs.hook.ts
ActiveChooN: Fixed

dwesolow (approved):

PR #5545, opened 2026-02-16
Annotating a video frame

### LINKED ISSUE #5281: Implement video annotation endpoints
Add support for videos in the annotation endpoints, and implement the necessary logic to extract frames and get/store the annotation.

warrkan:
Resolves #5281


ActiveChooN (dismissed):

dwesolow (dismissed):
UI approved, thanks!

application/backend/app/models/video_frame.py
leoll2, line 1: This file only contains a small class definition (`VideoFrame`), I suggest to move it to the existing `media.py`. After all, video frames are media.

application/backend/app/api/routers/media.py
```diff
@@ -221,3 +286,140 @@ def delete_media(
...
+        status.HTTP_404_NOT_FOUND: {"description": "Media, dataset item or project not found"},
+    },
+)
+def set_media_annotations(
+    project: Annotated[Project, Depends(get_project)],
+    media_id: MediaID,
+    media_annotations: Annotated[SetMediaAnnotations, Body(openapi_examples=SET_MEDIA_ANNOTATIONS_BODY_EXAMPLES)],
+    media_service: Annotated[MediaService, Depends(get_media_service)],
+    dataset_service: Annotated[DatasetService, Depends(get_dataset_service)],
+    ts: Annotated[float | None, Query(description="Video frame timestamp in seconds")] = None,
```
leoll2: Preferably, let's avoid abbreviations in the API. I suggest to rename this parameter to `timestamp`.

application/backend/app/api/routers/media.py
```diff
@@ -221,3 +286,140 @@ def delete_media(
...
+                )
+            video_frame = media_service.get_frame_by_video_id_and_timestamp(video_id=media_id, timestamp=ts)
+            if video_frame is None:
+                video_frame_media, video_frame = media_service.extract_video_frame(
+                    project=project, video_id=media_id, timestamp=ts
+                )
+                dataset_service.create_dataset_item(
+                    project=project,
+                    media=video_frame_media,
+                    user_reviewed=False,
```
leoll2:
```suggestion
                dataset_service.create_dataset_item(
                    project=project,
                    media=video_frame_media,
                    user_reviewed=True,
```
Annotations submitted by the client are considered user-reviewed, unlike predictions.

application/backend/app/services/video_frame_service.py
```diff
@@ -0,0 +1,38 @@
...
+# SPDX-License-Identifier: Apache-2.0
+from uuid import UUID
+
+from app.db.schema import VideoFrameDB
+from app.models import Media, VideoFrame
+from app.repositories.video_frame_repo import VideoFrameRepository
+from app.services import BaseSessionManagedService
+
+
+class VideoFrameService(BaseSessionManagedService):
```
leoll2, line 11: Not sure if we need a `VideoFrameService`, imho it can be merged into `MediaService` to simplify the initialization of the latter. Do you see any benefit to keep a separate `VideoFrameService`?

application/backend/app/services/video_frame_service.py
```diff
@@ -0,0 +1,38 @@
...
+            id=str(video_frame_media.id),
+            video_id=str(video.id),
+            timestamp=timestamp,
+        )
+
+        repo = VideoFrameRepository(db=self.db_session)
+        db_video_frame = repo.save(db_video_frame)
+        return VideoFrame.model_validate(db_video_frame)
+
+    def get_frame_by_video_id_and_timestamp(self, video_id: UUID, timestamp: float) -> VideoFrame | None:
```
leoll2, line 30: Please add docstrings. In what case does the method return `None`?

application/backend/app/services/video_frame_service.py
```diff
@@ -0,0 +1,38 @@
...
+        repo = VideoFrameRepository(db=self.db_session)
+        db_video_frame = repo.get_by_video_id_and_timestamp(video_id=str(video_id), timestamp=timestamp)
+        return VideoFrame.model_validate(db_video_frame) if db_video_frame else None
+
+    def get_frames_by_video_id(self, video_id: UUID) -> list[VideoFrame]:
```
leoll2: This method has linear complexity w.r.t. video size and it returns a list of unbounded size. Where is it used?

application/backend/app/services/media_service.py
```diff
@@ -291,3 +300,46 @@ def delete_media(self, project: Project, media_id: UUID) -> None:
             logger.warning("Media {} thumbnail was not found during deletion", media_id)
 
         repo.delete(obj_id=str(media.id))
+
+    def get_frame_by_video_id_and_timestamp(self, video_id: UUID, timestamp: float) -> VideoFrame | None:
+        return self.video_frame_service.get_frame_by_video_id_and_timestamp(video_id=video_id, timestamp=timestamp)
+
+    def extract_video_frame(self, project: Project, video_id: UUID, timestamp: float) -> tuple[Media, VideoFrame]:
+        """Extract video frame by video ID and timestamp"""
```
leoll2, line 308: What happens if a video frame already exists for the given `(video_id, timestamp)`?

application/backend/app/services/media_service.py
```diff
@@ -291,3 +300,46 @@ def delete_media(self, project: Project, media_id: UUID) -> None:
             logger.warning("Media {} thumbnail was not found during deletion", media_id)
 
         repo.delete(obj_id=str(media.id))
+
+    def get_frame_by_video_id_and_timestamp(self, video_id: UUID, timestamp: float) -> VideoFrame | None:
+        return self.video_frame_service.get_frame_by_video_id_and_timestamp(video_id=video_id, timestamp=timestamp)
```
leoll2, line 305: What happens if a `VideoFrame` does not exist for the given `timestamp`, but one exists for a very close timestamp (e.g. 1ms difference) that would resolve to the same physical frame?

application/backend/app/api/routers/media.py
```diff
@@ -221,3 +286,140 @@ def delete_media(
 ) -> None:
     """Delete media from the dataset"""
     media_service.delete_media(project=project, media_id=media_id)
+
+
+@router.post(
+    "/{media_id}/annotations",
+    status_code=status.HTTP_201_CREATED,
+    responses={
+        status.HTTP_201_CREATED: {"description": "Media/dataset item annotation created", "model": MediaAnnotations},
```
leoll2:
```suggestion
        status.HTTP_201_CREATED: {"description": "Annotation created or updated", "model": MediaAnnotations},
```

application/backend/app/api/routers/media.py
```diff
@@ -221,3 +286,140 @@ def delete_media(
...
+        )
+    except AnnotationValidationError as e:
+        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(e))
+
+
+@router.get(
+    "/{media_id}/annotations",
+    status_code=status.HTTP_200_OK,
+    responses={
+        status.HTTP_200_OK: {"description": "Media/dataset item found", "model": MediaAnnotations},
```
leoll2:
```suggestion
        status.HTTP_200_OK: {"description": "Annotation found", "model": MediaAnnotations},
```

application/backend/app/api/routers/media.py
```diff
@@ -221,3 +286,140 @@ def delete_media(
...
+    # Dataset item has the same ID as media
+    dataset_item_id = media_id
+    try:
+        media = media_service.get_media_by_id(project_id=project.id, media_id=media_id)
+        if media.type == MediaType.VIDEO:
+            # Video frames can be identified by video ID and frame timestamp
+            if ts is None:
+                raise HTTPException(
+                    status_code=status.HTTP_400_BAD_REQUEST, detail="Video frame timestamp is not provided."
+                )
```
leoll2, line 318: Do you validate that the timestamp is within boundaries, i.e. between 0 and the video duration? If it's not, the endpoint should return 400 Bad Request.

warrkan: Added validation of timestamp boundaries

application/backend/app/services/video_frame_service.py
warrkan: Added parameters to filter frames by timestamp. That method will be used by UI

application/backend/app/services/video_frame_service.py
warrkan: Let's keep it for now. We have one service per entity, later on it will be extended further

application/backend/app/api/routers/media.py
warrkan: Right! Thanks for spotting

application/backend/app/api/routers/media.py
warrkan: Renamed it to `timestamp`

application/backend/app/services/media_service.py
warrkan: That's a caller responsibility to check the precondition

application/backend/app/services/media_service.py
warrkan: We agreed, that we do not perform any closes frame lookup, the only major thing is that I've added float value rounding to ms precision

application/backend/app/services/video_frame_service.py
warrkan: Added

application/backend/app/models/video_frame.py
warrkan: For entity - agree. Done

application/backend/app/models/media.py
```diff
@@ -64,3 +66,15 @@ class Media(BaseEntity):
     fps: float | None
     frame_count: int | None
     source_id: UUID | None
+
+    @computed_field
+    @property
+    def duration(self) -> float | None:
+        """Return duration in seconds"""
+        return self.frame_count / self.fps if self.frame_count is not None and self.fps is not None else None
```
leoll2, line 74: The `Media` model could be refined as follows:
```python
class Media(BaseEntity):
    id: UUID
    project_id: UUID
    name: str
    format: MediaFormat
    width: int
    height: int
    size: int
    source_id: UUID | None

class Image(Media):
    type: Literal[MediaType.IMAGE] = MediaType.IMAGE

class Video(Media):
    type: Literal[MediaType.VIDEO] = MediaType.VIDEO
    fps: float
    frame_count: int

    @computed_field
    @property
    def duration(self) -> float:
```
This excludes `None` from several fields and enables better type validation with Pyrefly.

application/backend/app/api/routers/media.py
```diff
@@ -221,3 +286,155 @@ def delete_media(
...
+    dataset_item_id = media_id
+    try:
+        media = media_service.get_media_by_id(project_id=project.id, media_id=media_id)
+        if media.type == MediaType.VIDEO:
+            # Video frames can be identified by video ID and frame timestamp
+            if timestamp is None:
+                raise HTTPException(
+                    status_code=status.HTTP_400_BAD_REQUEST, detail="Video frame timestamp is not provided."
+                )
+            if timestamp > media.duration:  # pyrefly: ignore[unsupported-operation]
```
leoll2, line 319: This `pyrefly: ignore` could be removed by implementing the suggestion in my other comment.

leoll2 (approved):

dwesolow (approved):

application/backend/app/models/media.py
warrkan: That's not a simple change. We also need to change checks of `type` with `isinstance`. Additionally we need to split API schemas as well, it requires lots of changes in tests.. I've created a separate ticket to address this comment: https://github.com/open-edge-platform/training_extensions/issues/5563

PR #5547, opened 2026-02-16
feat: Logs components

### LINKED ISSUE #5409: Implement Logs visualisation similar to Geti Inspect

ActiveChooN:
### Summary

Added log viewer for trained models and training jobs
image

Depends on #5533
Closes #5409 

### Checklist


application/ui/package.json
```diff
@@ -12,7 +12,7 @@
...
         "update-spec": "npm run build:api:download && npm run build:api",
-        "clone-geti-ui-packages": "npx tiged --force --mode=git https://github.com/open-edge-platform/geti.git/web_ui/packages/config#60a7884983f199f2ca208c55bf071d54eb1729d2 packages/config && npx tiged --force --mode=git https://github.com/open-edge-platform/geti.git/web_ui/packages/ui#60a7884983f199f2ca208c55bf071d54eb1729d2 packages/ui && npx tiged --force --mode=git https://github.com/open-edge-platform/geti.git/web_ui/packages/smart-tools#60a7884983f199f2ca208c55bf071d54eb1729d2 packages/smart-tools",
+        "clone-geti-ui-packages": "npx tiged --force --mode=git https://github.com/open-edge-platform/geti.git/web_ui/packages/config#5f2032c6b2ef28a82721443e3725b11f70c7a89b packages/config && npx tiged --force --mode=git https://github.com/open-edge-platform/geti.git/web_ui/packages/ui#5f2032c6b2ef28a82721443e3725b11f70c7a89b packages/ui && npx tiged --force --mode=git https://github.com/open-edge-platform/geti.git/web_ui/packages/smart-tools#5f2032c6b2ef28a82721443e3725b11f70c7a89b packages/smart-tools",
```
camiloHimura, line 15: what is this change for?

camiloHimura:
what option allows to see them all?
image

camiloHimura:
Overall, this looks great. Should we consider implementing this for other types of jobs at some point?

ActiveChooN:
> what option allows to see them all? image

Debug, as the braodest filter. However, I don't think the backend publishes any debug logs.


> Overall, this looks great. Should we consider implementing this for other types of jobs at some point?

Yes, I think it would be beneficial. But it's out of the scope of this PR

application/ui/package.json
ActiveChooN: Expose useDialogContainer hook
https://github.com/open-edge-platform/geti/pull/1707

dwesolow:
>> what option allows to see them all? image
>
>Debug, as the braodest filter. However, I don't think the backend publishes any debug logs.
>
> > Overall, this looks great. Should we consider implementing this for other types of jobs at some point?
> 
> Yes, I think it would be beneficial. But it's out of the scope of this PR

I think `DEBUG` is not intuitive that its purpose is to see all the logs. WDYT about `ALL` option?

application/ui/src/api/fetch-sse.ts
```diff
@@ -49,9 +50,48 @@ export const connectSSE = <T = unknown>(path: string, options: SSEOptions<T>): S
...
+                        createEventSource();
+                    }
+                }, retryDelay);
+
+                retryDelay = retryDelay * 2;
+                retryCount += 1;
+            } else {
+                options.onError?.(event);
+                close();
+            }
```
dwesolow, line 70: Why this change?

application/ui/src/features/models/model-listing/current-model-training/training-model-row.component.tsx
```diff
@@ -144,8 +146,17 @@ export const TrainingModelRow = ({
 
                 <Text UNSAFE_className={classes.smallText}>...</Text>
 
-                {onCancel ? <CancelTraining onCancel={onCancel} job={job} /> : <div />}
+                <Flex gap={'size-100'} direction={'column'} alignItems={'center'}>
+                    <Button variant={'secondary'} onPress={() => setIsLogsDialogOpen(true)} aria-label={'View logs'}>
+                        Logs
+                    </Button>
```
dwesolow: What about wrapping button `Logs` and `TrainingLogsDialog` into one component that would be responsible for triggering the dialog etc? Doing so we won't rerender the whole model row when we open that logs, only that small component would rerender. WDYT?

application/ui/src/features/models/model-listing/utils/utils.ts
```diff
@@ -6,10 +6,15 @@ import { Model } from '../../../../constants/shared-types';
...
+    Successful: 'successful',
 } as const;
 
 export const isFailedModel = (model: Pick<Model, 'training_info'>): boolean =>
     model.training_info?.status === TRAINING_STATUS.Failed;
 
 export const isTrainingModel = (model: Pick<Model, 'training_info'>): boolean =>
     model.training_info?.status === TRAINING_STATUS.InProgress;
+
+export const hasTrainingLogs = (model: Pick<Model, 'training_info'>): boolean =>
```
dwesolow: why in progress training has no logs?

application/ui/src/api/fetch-sse.ts
ActiveChooN: The server responds with a 404 error when a log request is made for a while after job creation. Therefore, we need to reconnect.

application/ui/src/features/models/model-listing/utils/utils.ts
ActiveChooN: It has, but we're using SSE to fetch them

application/ui/src/features/models/model-listing/current-model-training/training-model-row.component.tsx
ActiveChooN: Extracted

ActiveChooN:
@dwesolow 
> I think DEBUG is not intuitive that its purpose is to see all the logs. WDYT about ALL option?

I prefer to stick with the common approach of setting a minimal logging level for filtering. It's commonly used across the industry. We set the default level to "info," which most users should not need to change.

camiloHimura:
> @dwesolow
> 
> > I think DEBUG is not intuitive that its purpose is to see all the logs. WDYT about ALL option?
> 
> I prefer to stick with the common approach of setting a minimal logging level for filtering. It's commonly used across the industry. We set the default level to "info," which most users should not need to change.

Good point, but we should still inform the user just in case. Perhaps a contextual info button with a brief description would help

application/ui/package.json
camiloHimura: got it 👍🏼

application/ui/src/api/fetch-sse.ts
dwesolow: What about starting connection only when job is created? It'd allow us to simplify the code.

application/ui/src/features/models/training-logs/hooks/use-auto-scroll.hook.ts
```diff
@@ -0,0 +1,54 @@
...
+                setAutoScroll(true);
+            }
+        } else {
+            isUserScrolling.current = true;
+            setAutoScroll(false);
+        }
+    }, [isDisabled]);
+
+    return { scrollRef, autoScroll, setAutoScroll, handleScroll };
+};
```
dwesolow: Could we use https://css-tricks.com/almanac/properties/o/overflow-anchor/#aa-demo to achieve the same behavior but using css?

application/ui/src/features/models/training-logs/log-entry.component.tsx
```diff
@@ -0,0 +1,52 @@
...
+};
+
+const formatTimestamp = (timestamp: number): string => {
+    if (!timestamp) {
+        return '';
+    }
+
+    const date = new Date(timestamp * 1000);
+
+    return date.toLocaleTimeString('en-US', { hour12: false, hour: '2-digit', minute: '2-digit', second: '2-digit' });
```
dwesolow: What about using `dayjs`? (we use that for other formatters)

application/ui/src/features/models/training-logs/log-viewer.component.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+const CONNECTION_STATUS_LABEL: Record<ConnectionStatus, string> = {
+    connecting: 'Connecting...',
+    connected: 'Live',
+    disconnected: 'Disconnected',
+    error: 'Connection error',
+};
+
+export const LogViewer = ({ logs, isStreaming = false, connectionStatus }: LogViewerProps) => {
+    const [minLevel, setMinLevel] = useState<LogLevel>('INFO');
```
dwesolow, line 34: np. Why `minLevel`? What about `logLevel`?

application/ui/src/features/models/training-logs/training-logs-dialog.component.tsx
```diff
@@ -0,0 +1,87 @@
...
+                </Text>
+            </Flex>
+        );
+    }
+
+    return <LogViewer logs={logs ?? []} />;
+};
+
+export const TrainingLogsDialog = ({ jobId, modelId }: TrainingLogsDialogProps) => {
+    const dialogContainer = useDialogContainer();
```
dwesolow, line 60: I've never used that hook. Do I understand correctly that it will get the parent dialog container and will allow us to dismiss that dialog without getting `onClose` from the parent component?

ActiveChooN: Yep, you can get type and function to close it 
https://react-spectrum.adobe.com/v3/DialogContainer.html#usedialogcontainer

application/ui/src/features/models/model-listing/utils/utils.ts
dwesolow: It's a bit confusing then.
We use that helper to decide wether `view logs` is enabled or disabled. Do I understand correctly that we should enable that only when training status is in progress?

application/ui/src/features/models/training-logs/hooks/use-auto-scroll.hook.ts
ActiveChooN: afaik this can help when content is inserted above the viewport. However, in our case, we need to chase new items down from the previous ones. I'm not sure how we can use it here.

dwesolow: image
I thought this is what we want to have, no? (keeping the scrollbar at the bottom when new content is there)

application/ui/src/features/models/training-logs/training-logs-dialog.component.tsx
dwesolow: Yeah but what is a benefit of using that vs passing `onClose` from the parent (the parent already has a state that toggles the dialog)?

application/ui/src/features/models/training-logs/log-viewer.component.tsx
ActiveChooN: wdym?

application/ui/src/features/models/training-logs/log-entry.component.tsx
ActiveChooN: It makes sense, changed

application/ui/src/features/models/training-logs/log-viewer.component.tsx
dwesolow: I meant the state name

application/ui/src/features/models/training-logs/training-logs-dialog.component.tsx
ActiveChooN: No special reason, but I think, it's little a bit cleaner in that case

application/ui/src/features/models/training-logs/log-viewer.component.tsx
ActiveChooN: It's the minimum log level, so minLevel was a good fit

application/ui/src/features/models/training-logs/hooks/use-auto-scroll.hook.ts
ActiveChooN: I missed the second demo. I experimented with it a bit and found it quite useful. We still need a ref for scrollToBottom, but it eliminates the need for manual scrolling with each new item. Thanks for your finding!

application/ui/src/api/fetch-sse.ts
ActiveChooN: We only show the current training job item with a "View Logs" button when there is a training job. However, there is a delay between job creation and log availability, so we need to account for it.

application/ui/src/features/models/model-listing/utils/utils.ts
ActiveChooN: This helper was used in the model-actions component and we show here only logs for trained or failed models. But reversed isTrainingModel helper can be used here

dwesolow (approved):

PR #5548, opened 2026-02-17
fix: SAM tool, dataset revision created at, models page

dwesolow:
### Summary

This PR fixes:
1. Training dialog, option "Show more" - The default "Sort models by" is Z-A. Should be A-Z. -> just changed the default value.
2. Dataset revisions - Should display the actual creation date instead of "-". -> we didn't use dataset revision creation date, updated it.
3. Project creation, label color picker - the right column (transparency?) doesn't work. I suggest to remove it. -> added `hideAlphaChannel`
4. Segmentation smart tool - when I move the cursor outside of the image, no area should be selected. -> the issue was that we updated state in `onPointerLeave` properly but there was pending decoding function that updated the state again (e.g. we updated state to be empty array, decoding update the state to have the shape again). We missed the proper function cancellation, I updated it.
5. Project creation - for classification, single-label should be the default. -> just changed the default value.

from https://github.com/open-edge-platform/training_extensions/issues/5532.

5 fixes, 5 commits, one commit per fix.

### Checklist

### DIFF
--- a/application/ui/src/features/annotator/tools/segment-anything-tool/segment-anything-tool.component.tsx
+++ b/application/ui/src/features/annotator/tools/segment-anything-tool/segment-anything-tool.component.tsx
@@ -1,7 +1,7 @@
 // Copyright (C) 2025 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 
-import { PointerEvent, useRef, useState } from 'react';
+import { PointerEvent, useCallback, useRef, useState } from 'react';
 
 import { clampPointBetweenImage } from '@geti/smart-tools/utils';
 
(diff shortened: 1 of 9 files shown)


application/ui/src/features/annotator/tools/segment-anything-tool/segment-anything-tool.component.tsx
```diff
@@ -50,6 +51,44 @@ const PreviewAnnotations = ({ previewAnnotations, image }: PreviewAnnotationsPro
     );
 };
 
+const useWithCancel = (fn: (points: InteractiveAnnotationPoint[]) => Promise<Shape[]>) => {
```
ActiveChooN: Can we move it to the common hooks? It looks quite generic

ActiveChooN (dismissed):

application/ui/src/features/annotator/tools/segment-anything-tool/segment-anything-tool.component.tsx
dwesolow: I thought about this but since it's only used here, in one file I decided to keep it there (scoped). Tho I can extract that hook to the separate file, will remember about this. Thanks!

dwesolow: I had to resolve conflicts so I updated this part.

ActiveChooN (approved):

PR #5575, opened 2026-02-20
Split Media entity into subentities

### LINKED ISSUE #5563: Split Media entity into subentities: Image and Video
Proposed way to refine the media models:

```python
class Media(BaseEntity):
    id: UUID
    project_id: UUID
    name: str
    format: MediaFormat
    width: int
    height: int
    size: int
    source_id: UUID | None

class Image(Media):
    type: Literal[MediaType.IMAGE] = MediaType.IMAGE

class Video(Media):
    type: Literal[MediaType.VIDEO] = MediaType.VIDEO
    fps: float
    frame_count: int

    @computed_field
    @property
    def duration(self) -> float:
```
See https://github.com/open-edge-platform/training_extensions/pull/5545#discussion_r2823229753.

warrkan:
The `Media` is split into three subentities - Image, Video and VideoFrame to manage specific fields more effectively

Resolves #5563


application/backend/app/models/media.py
```diff
@@ -50,31 +51,45 @@ class Media(BaseEntity):
...
-    frame_count: int | None
     source_id: UUID | None
 
+
+class Image(BaseMedia):
+    type: Literal[MediaType.IMAGE]
+
+
+class VideoFrame(BaseMedia):
+    type: Literal[MediaType.VIDEO_FRAME]
```
leoll2, line 78: `VideoFrame` and `VideoFrameView` should also have an `index` field. I believe it is interesting information for the client.

application/backend/app/repositories/media_repo.py
```diff
@@ -116,3 +116,25 @@ def delete(self, obj_id: str) -> bool:
...
+    def get_video_frames_by_video_id(
+        self, video_id: str, frame_index_from: int = 0, frame_index_to: int = 10
+    ) -> list[MediaDB]:
+        stmt = (
+            self._base_select()
+            .join(VideoFrameDB, VideoFrameDB.id == MediaDB.id)
+            .where(
+                VideoFrameDB.video_id == video_id,
+                VideoFrameDB.frame_index >= frame_index_from,
+                VideoFrameDB.frame_index < frame_index_to,
```
leoll2:
```suggestion
                VideoFrameDB.frame_index <= frame_index_to,
```

application/backend/app/services/media_service.py
```diff
@@ -301,10 +295,7 @@ def delete_media(self, project: Project, media_id: UUID) -> None:
 
         repo.delete(obj_id=str(media.id))
 
-    def get_frame_by_video_id_and_index(self, video_id: UUID, frame_index: int) -> VideoFrame | None:
-        return self.video_frame_service.get_frame_by_video_id_and_index(video_id=video_id, frame_index=frame_index)
-
-    def extract_video_frame(self, project: Project, video_id: UUID, frame_index: int) -> tuple[Media, VideoFrame]:
+    def extract_video_frame(self, project: Project, video_id: UUID, frame_index: int) -> Media:
```
leoll2, line 298: Shouldn't this method be annotated to return `VideoFrame` instead of `Media`? Same question for `get_video_frame_by_video_id_and_index` and `get_video_frames_by_video_id`.

warrkan: I've decided to remove `VideoFrame` completely and merged it to `Media`.

application/backend/app/models/media.py
warrkan: Added it to the entity

ActiveChooN (approved):

application/backend/app/services/media_service.py
leoll2: What do you mean? `VideoFrame` is defined in `application/backend/app/models/media.py`, L72

warrkan: Sorry I meant `video_frame` table and pervious `VideoFrame` entity. Now it's just a subclass of `Media`

leoll2: That's my point. Since `VideoFrame` is a subclass of `Media`, this method could be type-hinted to return `VideoFrame` instead of `Media`, so that the code which consumes this function can make stronger assumptions on the returned object, e.g. he can access attributes like `frame_index` and `video_id` which are only defined for `VideoFrame` but not for a generic `Media`.

leoll2 (approved):

application/backend/app/services/media_service.py
warrkan: Ah, I see your point. Will do this in the next PR

PR #5581, opened 2026-02-20
Fix paths that trigger the execution of backend and build GHA workflows

### LINKED ISSUE #5580: GHA server checks skipped when library changes
Example: https://github.com/open-edge-platform/training_extensions/actions/runs/22190071812

leoll2:
### Summary

Backend and build GHA workflows should be also triggered when there are changes in the library, since they both depend on otx.

Fixes #5580 

### Checklist

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -40,9 +40,8 @@ jobs:
         with:
           files_yaml: |
             test:
-              - application/backend/**
-              - application/docker/**
-              - application/ui/**
+              - application/**
+              - '!application/docs/**'
               - library/**
               - .github/**
 
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -40,7 +40,10 @@ jobs:
         with:
           files_yaml: |
             test:
-              - application/**
+              - application/backend/**
+              - application/docker/**
+              - application/ui/**
+              - library/**
               - .github/**
 
   build:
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -40,7 +40,10 @@ jobs:
         with:
           files_yaml: |
             test:
-              - application/**
+              - application/backend/**
+              - application/docker/**
+              - application/ui/**
```
mramotowski: Consider:
```
- application/**
- '!application/docs/**'
```
As in https://github.com/tj-actions/changed-files?tab=readme-ov-file#on-pull_request- 3 Example.

mramotowski (dismissed):

.github/workflows/build.yaml
leoll2: Thanks for the suggestion, applied!

mramotowski (approved):

.github/workflows/build.yaml
```diff
@@ -41,6 +41,8 @@ jobs:
           files_yaml: |
             test:
               - application/**
+              - '!application/docs/**'
+              - library/**
```
itallix, line 45: Shall we trigger it only when dependencies updated (i.e. `library/pyproject.toml` changed)?

leoll2: No, it would be too optimistic. There several scenarios that can cause build problems other than changes to `pyproject.toml`, `uv.lock` and `Dockerfile`, for example:
1. New source files that are not whitelisted in `.dockerignore`
2. Refactor of modules/folders directly referenced in the Dockerfile (e.g. `application/backend/app`)
3. Changes in `main.py`
4. Changes to env variables and config files

In all above situations, the unit/integration tests may remain green, while the Docker container would throw errors.

PR #5585, opened 2026-02-20
refactor: Improve annotator UX by moving/creating providers and hooks (responsibilities)

dwesolow:
### Summary

This PR is fixes user experience of the annotator (and potential bugs for the video player toolbar). Currently in the develop, when we change media/submit annotations, we unmount primary toolbar, labels, submit annotation button, canvas and bottom toolbar. This caused the issue that when we select label A and tool B on the image X, we change to image Y, label and tool are reseted. This PR fixes that issue, that significantly improves UX of the annotator (it will improve playing video later as well).

More details:
`AnnotatorProvider` is replaced by `MediaItemImageProvider` and `AnnotatorLabelsProvider`. Each of these providers has its own responsibility. `SelectedMediaItemProvider` is going to be used more in the upcoming PRs.

### Checklist


application/ui/src/features/annotator/selected-media-item-provider.component.tsx
```diff
@@ -0,0 +1,88 @@
...
+    children: ReactNode;
+};
+
+export const SelectedMediaItemProvider = ({
+    mediaItem: initialMediaItem,
+    children,
+}: SelectedMediaItemProviderProps) => {
+    const [mediaItem, setMediaItem] = useState<Media>(initialMediaItem);
+
+    /* TODO: Check if we need this */
```
ActiveChooN, line 29: Will this be checked in upcoming PRs?

application/ui/src/shared/annotator/annotation-actions-provider.component.tsx
```diff
@@ -99,6 +100,13 @@ export const AnnotationActionsProvider = ({
         return mapServerAnnotationsToLocal(initialAnnotationsDTO, projectLabels);
     });
 
+    const prevInitialAnnotationsDTORef = useRef(initialAnnotationsDTO);
+
+    if (!isEqual(prevInitialAnnotationsDTORef.current, initialAnnotationsDTO)) {
```
ActiveChooN, line 105: Should this be an effect?

application/ui/src/features/annotator/selected-media-item-provider.component.tsx
dwesolow: Yeah, once I integrate this with API

application/ui/src/shared/annotator/annotation-actions-provider.component.tsx
dwesolow: It's made on purpose to not use effect (doing in effect it would trigger whenever initial annotations, project labels or setAnnotations change. Moreover this way is recommended by react team if we need to sync state).

application/ui/src/features/annotator/selected-media-item-provider.component.tsx
```diff
@@ -0,0 +1,88 @@
...
+    useEffect(() => {
+        setMediaItem(initialMediaItem);
+    }, [initialMediaItem]);
+
+    const roi: RegionOfInterest = { x: 0, y: 0, width: mediaItem.width, height: mediaItem.height };
```
camiloHimura, line 34: I’m not sure this is actually necessary. We don’t support task chains, and for the components that require it, the source of truth should be the actual media item size.

dwesolow: We didn't change that code from geti, we use basically the same concept of roi (this PR didn't change it, only moved to separate files/providers).

application/ui/src/features/dataset/media-preview/api/use-annotations-query.ts
```diff
@@ -16,12 +17,12 @@ const isUnannotatedError = (error: unknown): boolean => {
 export const useAnnotationsQuery = (mediaId: string) => {
     const projectId = useProjectIdentifier();
 
-    return useSuspenseQuery({
```
camiloHimura, line 19: why this change?

dwesolow: Whenever we get new annotation it caused the whole media preview (content of the annotator dialog apart from sidebar) to be unmounted. In that case we lost the state of the selected tool, selected label etc.

application/ui/src/features/annotator/selected-media-item-provider.component.tsx
```diff
@@ -0,0 +1,88 @@
...
+export const MediaItemImageLoader = ({ children }: { children: ReactNode }) => {
+    const { mediaItem } = useSelectedMediaItem();
+
+    // TODO: Use getMediaItem when API supports video frames
+    const { data: image } = useLoadImageQuery(mediaItem.id);
+
+    return <MediaItemImageContext value={{ image }}>{children}</MediaItemImageContext>;
+};
+
+export const useMediaItemImage = (): MediaItemImageContextType => {
```
camiloHimura, line 80: It needs a more descriptive name that clearly indicates it’s responsible for loading the image. Initially, I was confused about why this was necessary since we already have `useSelectedMediaItem`

application/ui/src/features/dataset/media-preview/api/use-annotations-query.ts
camiloHimura: Since we tend to use useSuspenseQuery in most cases, I think we should add a comment explaining why we’re not using it here, so that someone doesn’t try to switch it back later.

camiloHimura (approved):

PR #5597, opened 2026-02-23
Support content negotiation for model logs endpoint

### LINKED ISSUE #5583: Support plain/text response for model logs endpoint
Currently, the GET `/api/projects/{project_id}/models/{model_id}/logs` endpoint returns `NDJSON` formatted log files as persisted by `Loguru`. This ticket aims to extend the endpoint to support content negotiation based on the `Accept` header, allowing clients to request either the raw `NDJSON` format or plain text logs.

Implement content negotiation based on the Accept header:
- `Accept: application/x-ndjson` or missing header → Return the original NDJSON log file
- `Accept: text/plain` → Parse the NDJSON file and return only the formatted log messages (extract the text field from each JSON object)

**Acceptance Criteria**
- [ ] Endpoint returns plain text logs when `Accept: text/plain` is specified
- [ ] Endpoint returns NDJSON logs when `Accept: application/x-ndjson` is specified
- [ ] Default behavior when no `Accept` header is provided
- [ ] Appropriate `Content-Type` header is set in the response
- [ ] Error handling for malformed JSON lines is implemented
- [ ] OpenAPI documentation is updated to reflect supported content types

itallix:
### Summary

Enhanced the `GET /api/projects/{project_id}/models/{model_id}/logs` endpoint to support content negotiation via the `Accept` header:
- `Accept: text/plain` - Returns plain text format (parsed from `NDJSON`)
- `Accept: application/json` or `Accept: application/x-ndjson` - Returns `NDJSON` format (default)

### How to test

- **text:** `curl -H "Accept: text/plain" http://localhost:7860/api/projects/2d3280e3-2cb2-4b13-8f4a-dcaaed969678/models/61e82b16-b782-4e5d-8fda-e17d2e06352f/logs -o training.txt`
- **json:** `curl http://localhost:7860/api/projects/2d3280e3-2cb2-4b13-8f4a-dcaaed969678/models/61e82b16-b782-4e5d-8fda-e17d2e06352f/logs -o training.log`

### Checklist

### DIFF
--- a/application/backend/app/api/routers/models.py
+++ b/application/backend/app/api/routers/models.py
@@ -240,7 +240,7 @@ def get_training_logs(
     Supports content negotiation via Accept header (text/plain, application/x-ndjson).
     """
     try:
-        as_text: bool = accept and "text/plain" in accept.lower()  # pyrefly: ignore
+        as_text = accept is not None and "text/plain" in accept.lower()
 
         training_log = model_service.get_logs(project_id=project.id, model_id=model_id, as_text=as_text)
         if training_log is None:
--- a/application/backend/app/api/routers/models.py
+++ b/application/backend/app/api/routers/models.py
@@ -240,7 +240,7 @@ def get_training_logs(
     Supports content negotiation via Accept header (text/plain, application/x-ndjson).
     """
     try:
-        as_text = (accept and "text/plain" in accept.lower()) or False
+        as_text: bool = accept and "text/plain" in accept.lower()  # pyrefly: ignore
 
         training_log = model_service.get_logs(project_id=project.id, model_id=model_id, as_text=as_text)
         if training_log is None:
--- a/application/backend/app/api/routers/models.py
+++ b/application/backend/app/api/routers/models.py
@@ -4,11 +4,13 @@
 import io
 import os
 import zipfile
-from typing import Annotated
+from collections.abc import Iterable
+from pathlib import Path
+from typing import Annotated, cast
 
 from fastapi import APIRouter, Body, Depends, Header, HTTPException, Query, status
 from fastapi.openapi.models import Example
-from fastapi.responses import Response, StreamingResponse
+from fastapi.responses import StreamingResponse
 from starlette.responses import FileResponse
 
 from app.api.dependencies import get_model_service, get_project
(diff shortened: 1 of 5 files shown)


application/backend/app/api/routers/models.py
```diff
@@ -222,18 +229,31 @@ def get_training_metrics(
...
+    accept: Annotated[str | None, Header(description="Accept header to specify the desired response format")] = None,
+) -> FileResponse | StreamingResponse:
     """
     Download the training log file for a given model.
+    Supports content negotiation via Accept header (text/plain, application/x-ndjson).
     """
     try:
-        log_file = model_service.get_logs(project_id=project.id, model_id=model_id)
-        if not log_file:
+        as_text = (accept and "text/plain" in accept.lower()) or False
```
leoll2:
```suggestion
        as_text = (accept and "text/plain" in accept.lower())
```
The `or False` part has no effect; the left part already returns a boolean value, and the logical expression `X or False` is equivalent to `X` itself.

application/backend/app/services/model_service.py
```diff
@@ -542,4 +546,19 @@ def get_logs(self, project_id: UUID, model_id: UUID) -> Path | None:
...
+            with open(log_file, encoding="utf-8") as f:
+                for line in f:
+                    try:
+                        entry = json.loads(line)
+                        text = entry.get("text", "")
+                        if text:
+                            yield text
+                    except json.JSONDecodeError as e:
+                        logger.warning("Failed to parse log line: {}", e)
+                        continue
```
leoll2: In case of error, would it be maybe better to return a placeholder string instead of `continue`, e.g. `[MISSING LINE, COULD NOT DECODE]`. Since the primary goal of logs is to troubleshoot issue, I find it important to inform that some information that was supposed to be logged is actually missing from the log due to an unrelated deserialization problem.

application/backend/app/api/routers/models.py
itallix: This was added in response to pyrefly `bad-assignment`. Let me put `pyrefly: ignore` for this line instead.

```
❯ uv run pyrefly check .
ERROR `Literal[''] | bool | None` is not assignable to `bool` [bad-assignment]
   --> app/api/routers/models.py:243:25
    |
243 |         as_text: bool = accept and "text/plain" in accept.lower()
    |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
 INFO 1 error (114 suppressed)
```

application/backend/app/services/model_service.py
itallix: Indeed, while this is not expected behaviour, it still could serve as an indicator for looking into original `ndjson` source.

application/backend/app/api/routers/models.py
leoll2: Actually there is an issue: `accept` is type-hinted as `str | None`, and a null value causes the logical expression to evaluate as `None` instead of `bool`. That's what Pyrefly complains about.

Your code should be:
```python
as_text: bool = accept is not None and "text/plain" in accept.lower()
```

itallix: Thanks! Fixed.

leoll2 (approved):

PR #5599, opened 2026-02-23
Added stride field to video entity

### LINKED ISSUE #5565: Add stride field to the video entity
To make current approach compatible with Geti Classic, a stride field have to be added with `fps` value.

warrkan:
For Geti Classic compatibility `stride` field is added to `Video` entity.
Resolves #5565

### DIFF
--- a/application/backend/app/models/media.py
+++ b/application/backend/app/models/media.py
@@ -97,6 +97,11 @@ class Video(BaseMedia):
         """Return duration in seconds"""
         return self.frame_count / self.fps
 
+    @computed_field
+    @property
+    def stride(self) -> int:
+        return int(self.fps)
+
 
 Media = Annotated[Image | Video | VideoFrame, Field(discriminator="type")]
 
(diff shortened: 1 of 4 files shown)


application/backend/app/models/media.py
```diff
@@ -97,6 +97,11 @@ def duration(self) -> float:
         """Return duration in seconds"""
         return self.frame_count / self.fps
 
+    @computed_field
+    @property
+    def stride(self) -> int:
+        return int(self.fps)
```
leoll2, line 103: What does this `stride` property represent? Since it's computed as `int(self.fps)`, I don't think it represents any physical property of the video.

warrkan: UI uses it as a step for annotator, in Geti Classic there was a separate value for it, but it was never set, so fps has been used all the time. Here we introduce it for a backward compatibility.

leoll2: Hmm, I disagree with "dirtying" the API with UI-specific implementation details. This `stride` property is not meaningful for the server nor any client consuming the API; I would suggest that the UI calculate this value on their end, rather than expecting it to be returned by the server.

warrkan:
Closing this PR, UI will use int'ed FPS value as a stride

PR #5600, opened 2026-02-23
Restructure training configuration API

### LINKED ISSUE #5492: Restructure training configuration API
The current training configuration API is not optimal and needs some refinement.

Things to fix:
- The "Get Training Configuration" endpoint also allows to fetch a generic task-level configuration, although it's not usable in practice under the current UX
- The response doesn't include user-friendly names for the categories.
- ExtendedModelView returns the raw training config dict, which is unstructured therefore hard to process for the client.
- API docs do not contain examples

leoll2:
### Summary

This PR revisits the API, data structures and logic that power the 'advanced training configuration' feature.

Changes:
- Add a new endpoint `GET /{model_id}/training_configuration` to retrieve the (structured) configuration used to train a specific model revision
- Simplify `GET /{model_id}` endpoint response by removing the (now redundant) training configuration
- Update the `GET/PATCH /training_configuration` endpoints
  - Remove the `model_revision_id` parameter
  - Update a more structured model for the response (`TrainingConfigurationView`)
  - Handle errors for unrecognized manifest ids
- Adjust descriptions, types and validators of several configuration-related Pydantic models
- Refactor helpers to parse and validate model manifests to a new service class `ModelManifestService`
- Remove unnecessary `supported_gpus` field from manifests

Resolves #5492 

### How to test

Get the current configuration:
```sh
curl -s -X GET localhost:7860/api/projects/${PROJECT_ID}/training_configuration?model_architecture_id=${ARCHITECTURE_ID} | jq
```

Apply some changes to the configuration:
```sh
url -s -X PATCH localhost:7860/api/projects/${PROJECT_ID}/training_configuration?model_architecture_id=${ARCHITECTURE_ID} -H "Content-Type: application/json" --data '{"dataset_preparation.augmentation.random_affine.max_rotate_degree": 20.0}' | jq

curl -s -X PATCH localhost:7860/api/projects/${PROJECT_ID}/training_configuration?model_architecture_id=${ARCHITECTURE_ID} -H "Content-Type: application/json" --data '{"training.max_epochs": 5}' | jq
```

Reload the configuration to verify that the changes have been applied:
```sh
curl -s -X GET localhost:7860/api/projects/${PROJECT_ID}/training_configuration?model_architecture_id=${ARCHITECTURE_ID} | jq
```

Train a model and check its configuration:
```sh
curl -s -X GET localhost:7860/api/projects/${PROJECT_ID}/models/${MODEL_ID}/training_configuration | jq
```

### Checklist


application/backend/app/models/training_configuration/augmentation.py
```diff
@@ -1,20 +1,19 @@
...
-from pydantic import Field, model_validator
+from pydantic import BaseModel, Field, model_validator
 
-from app.models.base import BaseModelNoExtra
 
-
-class RandomResizeCrop(BaseModelNoExtra):
+class BaseAugmentationParameter(BaseModel):
     enable: bool = Field(
         default=False,
```
A-Artemis, line 9: Making all the classes now inherit form `BaseAugmentationParameter` suggest that all parameters are `enable=False` by default. Are there any parameters which are enabled by default?

Additionally, it can make knowing which setting is enabled or not be default harder to spot, and some parameters might require others to be enabled as well. 

e.g. looking at the `GaussianBlur` parameter, the `kernal_size` can be individually disabled, yet the `sigma` is individually enabled. Is that scenario possible?

application/backend/app/models/training_configuration/augmentation.py
```diff
@@ -74,8 +62,7 @@ class RandomAffine(BaseModelNoExtra):
             "For example, max_rotate_degree=10 allows up to ±10 degrees rotation."
         ),
     )
-    max_translate_ratio: float | None = Field(
-        default=None,
```
A-Artemis, line 78: How is the default value now derived? Is it taken from the `.yaml`?

application/backend/app/services/training_configuration_service.py
```diff
@@ -1,171 +1,139 @@
...
         )
 
-        validated_updated_config = PartialTrainingConfiguration(**updated_config)  # type: ignore[arg-type]
+    def update(
+        self, project_id: UUID, model_architecture_id: str, training_configuration: TrainingConfiguration
+    ) -> None:
+        """
+        Update the training configuration.
+
+        This method can handle both updates to task-level parameters (which apply to all model architectures within
```
A-Artemis, line 110: Is this clear to the user that changing a detection parameter here will change it for all detection projects?

A-Artemis (commented):
Looks good and cleaner than the previous implementation.

application/backend/app/models/training_configuration/augmentation.py
leoll2: > Making all the classes now inherit form BaseAugmentationParameter suggest that all parameters are enable=False by default. Are there any parameters which are enabled by default?

This is indeed the expected behavior. The default `enable=False` of the Pydantic models can be overridden by the corresponding value in the manifest, if provided. For example, random horizontal flip is enabled by default for all detection models:
https://github.com/open-edge-platform/training_extensions/blob/137d9257d7585e9db66d91772984d9f3195fe7e8/application/backend/app/supported_models/manifests/detection/base.yaml#L24

> Additionally, it can make knowing which setting is enabled or not be default harder to spot, and some parameters might require others to be enabled as well.

The defaults are primarily controlled through the manifests. To determine if a certain parameter would be enabled by default for a certain architecture, you should start from the manifest for that architecture; if the value is not there, check the base manifests; if it's also not there, the default is taken from the Pydantic model.

> e.g. looking at the GaussianBlur parameter, the kernal_size can be individually disabled, yet the sigma is individually enabled. Is that scenario possible?

`gaussian_blur.kernel_size` and `gaussian_blur.sigma` values can be configured independently from `gaussian_blur.enable`, however they take effect if and only if `gaussian_blur.enable=true`.

application/backend/app/models/training_configuration/augmentation.py
leoll2: It must be provided by the manifest. In this case, the Pydantic model doesn't provide a default because there is no good default value.

application/backend/app/services/training_configuration_service.py
leoll2: I split the parameters in two categories - algo-level vs task-level - depending if it makes sense to share the value of the parameter across different architectures or not.
For example, the learning rate depends heavily on the model architecture, it wouldn't make sense to share it. Conversely, subset splitting only makes sense at task-level, each dataset item belongs to a subset and it can't be changed later otherwise all metrics become useless.

leoll2: > Is this clear to the user that changing a detection parameter here will change it for all detection projects?

To answer your original question, probably not: we should improve the UX to better communicate the scope of each parameter.

application/backend/app/api/schemas/training_configuration.py
```diff
@@ -0,0 +1,318 @@
...
+        if hasattr(field_info, "metadata"):
+            for constraint in field_info.metadata:
+                if hasattr(constraint, "ge"):
+                    min_value = constraint.ge
+                elif hasattr(constraint, "gt"):
+                    min_value = constraint.gt
+                if hasattr(constraint, "le"):
+                    max_value = constraint.le
+                elif hasattr(constraint, "lt"):
+                    max_value = constraint.lt
```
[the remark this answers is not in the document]
leoll2: I will improve it in a later PR if needed

application/backend/app/models/training_configuration/augmentation.py
A-Artemis: > gaussian_blur.kernel_size and gaussian_blur.sigma values can be configured independently from gaussian_blur.enable, however they take effect if and only if gaussian_blur.enable=true.

If `gaussian_blur.enable=true` and then the "sub" parameter `kernel_size` is enabled, but `sigma` is disabled, can the Guassian blur still take place? It feels like all sub parameters must be enabled all once or not at all, being dependent of the parent parameter `gaussian_blur.enable=true`.

leoll2: `sigma` can't be disabled: it is defined as `sigma: tuple[float, float]`, not nullable. The `enable` applies to the `GaussianBlur` object, its fields (`kernel_size`, `sigma`, ...) don't have their own `enable` option.

A-Artemis (dismissed):

A-Artemis (approved):

PR #5601, opened 2026-02-24
feat: Implement video timeline navigation

### LINKED ISSUE #5598: Implement video player timeline navigation logic

dwesolow:
### Summary

This PR includes:
1. Adds logic to video controllers to be able to navigate the video.
2. Improves `query-client.interface.ts` so whenever API exposes `query` parameter, that `query` can be used as part of the strongly typed `queryKey`.
3. https://github.com/open-edge-platform/training_extensions/pull/5575 introduced discriminated types, also for `MediaVideoFrameDTO` but that type won't be used (endpoint does not return frames, only video. We can retrieve only thumbnail and binary of the frame). Because of that we create our local `MediaVideoFrame` type that we use across the components.

https://github.com/user-attachments/assets/cc98c303-62cf-407e-af9a-f3f4e50b06a2

Closes #5598 

### Checklist

### DIFF
--- a/application/ui/src/features/annotator/video-player/use-video-controls.ts
+++ b/application/ui/src/features/annotator/video-player/use-video-controls.ts
@@ -19,15 +19,15 @@ export type VideoControls = {
 
 export const useVideoControls = (
     videoRef: RefObject<HTMLVideoElement | null>,
-    mediaItem: MediaVideoFrame | undefined,
+    videoFrame: MediaVideoFrame | undefined,
     selectVideoFrame: (media: MediaVideoFrame) => void,
     changeCurrentFrameIndex: (index: number) => void
 ): VideoControls => {
     const [isPlaying, setIsPlaying] = useState<boolean>(false);
 
-    const totalFrames = mediaItem?.frame_count ?? 1;
-    const step = mediaItem?.frame_stride;
-    const currentFrameNumber = mediaItem.frame_number;
+    const totalFrames = videoFrame?.frame_count ?? 1;
+    const step = videoFrame?.frame_stride ?? 1;
+    const currentFrameNumber = videoFrame?.frame_number ?? 0;
 
     const round = (x: number) => Math.round(x / step) * step;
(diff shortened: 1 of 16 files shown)


application/ui/src/features/annotator/video-player/video-frame.component.tsx
```diff
@@ -58,17 +61,25 @@ const useRequestVideoFrameCallback = (
...
+    const { videoRef, videoControls } = useVideoPlayer();
 
     useRequestVideoFrameCallback(videoRef, canvasRef);
 
+    const handleEnded = () => {
+        if (videoRef.current === null) {
+            return;
+        }
+        videoControls.pause();
+        videoRef.current.currentTime = 0;
```
camiloHimura, line 80: when the video ends the the thumb-slider got back to the initial frame?

dwesolow: No, it stays at the end. After that when `play` button is pressed, we play from the beginning.

application/ui/src/features/annotator/video-player/use-video-controls.ts
```diff
@@ -39,40 +62,63 @@ export const useVideoControls = (
...
+            changeCurrentFrameIndex(previousVideoFrameNumber);
+            videoRef.current.currentTime -= 1;
+        }
     };
 
-    const goto = (_frameNumber: number) => {
-        //
+    const goto = (frameNumber: number) => {
+        if (videoRef.current === null || videoFrame === undefined) {
+            return;
```
camiloHimura, line 106: Should we notify the user that the video isn't ready/available yet?

camiloHimura (approved):

application/ui/src/features/annotator/video-player/use-video-controls.ts
dwesolow: It should never happen, `use-video-controls.ts` methods are not used if we don't have the video. It's to make ts happy.

PR #5606, opened 2026-02-24
Media Endpoints - Get frame binary, thumbnail. Get list of annotated frames

### LINKED ISSUE #5282: Implement video loading endpoints
Implement the endpoint to get the video details, list of frames and to fetch the video content.

### LINKED ISSUE #5531: Media thumbnail endpoint should return pre-created thumbnail
The [`get_media_thumbnail`](https://github.com/open-edge-platform/training_extensions/blob/480f91ff6c8173cbad2988fb90e50107908f9bc6/application/backend/app/api/routers/media.py#L194) endpoint incorrectly returns a thumbnail generated on the fly, instead of serving the thumbnail that was previously created at upload time ([source](https://github.com/open-edge-platform/training_extensions/blob/480f91ff6c8173cbad2988fb90e50107908f9bc6/application/backend/app/services/media_service.py#L105)).

EDIT: for images, we already generate the thumbnail on upload. For videos, we currently don't but we should.

Note that we have several thumbnail endpoints, each with different expected behavior:
- [`get_media_thumbnail`](https://github.com/open-edge-platform/training_extensions/blob/480f91ff6c8173cbad2988fb90e50107908f9bc6/application/backend/app/api/routers/media.py#L194) (scope of this ticket): should return a pre-created thumbnail, ~at least for images~ for both images and videos. The path can be obtained through `MediaService.get_media_thumbnail_path_by_id`.
- [`get_dataset_revision_item_thumbnail`](https://github.com/open-edge-platform/training_extensions/blob/develop/application/backend/app/api/routers/dataset_revisions.py#L218) -> generates a thumbnail on the fly, because it can't be saved in the revision (persisted Datumaro dataset).
- [`get_project_tuhumbnail`](https://github.com/open-edge-platform/training_extensions/blob/develop/application/backend/app/api/routers/projects.py#L255) -> returns the thumbnail of the earliest uploaded media in the project.

warrkan:
Resolves #5282

Following endpoints were modified

```
GET /api/projects/{project_id}/dataset/media/{media_id}/binary
```
This endpoint now can return annotated frames pre-extracted image content, as well it can extract image frames on the fly in case if it's not annotated yet.
It can work with both frame ID and combination of video ID + frame index.

```
GET /api/projects/{project_id}/dataset/media/{media_id}/thumbnail
```
This endpoint now can return annotated frames pre-extracted image thumbnail, as well it can extract image frames and crop them on the fly in case if it's not annotated yet.
It can work with both frame ID and combination of video ID + frame index.

New endpoint has been added:
```
GET /api/projects/{project_id}/dataset/media/{media_id}/frames
```
It returns the list of annotated video frames.

### DIFF
--- a/application/backend/app/api/routers/media.py
+++ b/application/backend/app/api/routers/media.py
@@ -8,10 +8,12 @@ from uuid import UUID
 
 from fastapi import APIRouter, Body, Depends, File, HTTPException, Query, UploadFile, status
 from fastapi.openapi.models import Example
+from PIL.Image import Image
 from starlette.responses import FileResponse, StreamingResponse
 
 from app.api.dependencies import get_dataset_service, get_file_name_and_extension, get_media_service, get_project
 from app.api.schemas.media import (
+    AnnotatedVideoFrame,
     MediaAnnotations,
     MediaView,
     MediaViewAdapter,
@@ -30,6 +32,9 @@ router = APIRouter(prefix="/api/projects/{project_id}/dataset/media", tags=["Med
 
 DEFAULT_MEDIA_NUMBER_RETURNED = 10
 MAX_MEDIA_NUMBER_RETURNED = 100
+
+DEFAULT_FRAME_INDEX_FROM = 0
+DEFAULT_FRAME_INDEX_TO = 50
 
 SET_MEDIA_ANNOTATIONS_BODY_EXAMPLES = {
     "single_label": Example(
@@ -104,6 +109,16 @@ def _parse_media_format(extension: str) -> ImageFormat | VideoFormat:
                 status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
                 detail=f"Unsupported media extension: {extension}",
             )
+
+
+def _write_image_to_response(image: Image, filename: str, cache_control: str | None = None) -> StreamingResponse:
+    buffer = BytesIO()
+    image.save(buffer, format="JPEG")
+    buffer.seek(0)
+    headers = {"Content-Disposition": f"inline; filename={filename}"}
+    if cache_control:
+        headers["Cache-Control"] = cache_control
+    return StreamingResponse(buffer, media_type="image/jpeg", headers=headers)
 
 
 @router.post(
@@ -229,7 +244,46 @@ def get_media(
 
 
 @router.get(
+    "/{media_id}/frames",
+    responses={
+        status.HTTP_200_OK: {"description": "List of annotated video frames", "model": list[AnnotatedVideoFrame]},
+        status.HTTP_404_NOT_FOUND: {"description": "Project or video not found"},
+    },
+)
+def list_video_frames(
+    project: Annotated[Project, Depends(get_project)],
+    media_id: MediaID,
+    media_service: Annotated[MediaService, Depends(get_media_service)],
+    frame_index_from: Annotated[int, Query(ge=0)] = DEFAULT_FRAME_INDEX_FROM,
+    frame_index_to: Annotated[int, Query(ge=0)] = DEFAULT_FRAME_INDEX_TO,
+) -> list[AnnotatedVideoFrame]:
+    """Lists annotated video frames with frame index range"""
+    media = media_service.get_media_by_id(project_id=project.id, media_id=media_id)
+    if media.type != MediaType.VIDEO:
+        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail="Requested media is not video.")
+    annotated_video_frames = media_service.list_annotated_video_frames_by_video_id(
+        project=project,
+        video_id=media_id,
+        frame_index_from=frame_index_from,
+        frame_index_to=frame_index_to,
+    )
+    return [
+        AnnotatedVideoFrame(
+            media_id=video_frame.id,
+            frame_index=video_frame.frame_index,
+            dataset=MediaAnnotations(
+                annotations=dataset_item.annotation_data,  # type: ignore[arg-type]
+                prediction_model_id=dataset_item.prediction_model_id,
+                user_reviewed=dataset_item.user_reviewed,
+            ),
+        )
+        for (dataset_item, video_frame) in annotated_video_frames
+    ]
+
+
+@router.get(
     "/{media_id}/binary",
+    response_model=None,
     responses={
         status.HTTP_200_OK: {"description": "Media binary found"},
         status.HTTP_400_BAD_REQUEST: {"description": "Invalid media ID or project ID"},
@@ -240,10 +294,27 @@ def get_media_binary(
     project: Annotated[Project, Depends(get_project)],
     media_id: MediaID,
     media_service: Annotated[MediaService, Depends(get_media_service)],
-) -> FileResponse:
+    frame_index: Annotated[int | None, Query(description="Video frame index", ge=0)] = None,
+) -> FileResponse | StreamingResponse:
     """Get media binary content"""
-    binary_path = media_service.get_media_binary_path_by_id(project_id=project.id, media_id=media_id)
     media = media_service.get_media_by_id(project_id=project.id, media_id=media_id)
+    print(media)
+    if media.type == MediaType.VIDEO and frame_index is not None:
+        # Video frames can be identified by video ID and frame index
+        if frame_index >= media.frame_count:  # pyrefly: ignore[unsupported-operation]
+            raise HTTPException(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                detail=f"Video frame index {frame_index} exceeds video frames count {media.frame_count}.",
+            )
+        video_frame = media_service.get_video_frame_by_video_id_and_index(
+            project=project, video_id=media_id, frame_index=frame_index
+        )
+        if video_frame is None:
+            frame_binary = media_service.get_frame_binary(project=project, video=media, frame_index=frame_index)
+            return _write_image_to_response(image=frame_binary, filename=f"{media.name}_frame_{frame_index}.jpeg")
+        media = video_frame
+
+    binary_path = media_service.get_media_binary_path_by_id(project_id=project.id, media_id=media.id)
     filename = f"{media.name}.{media.format.value.lower()}"
     return FileResponse(path=binary_path, filename=filename)
 
@@ -260,19 +331,28 @@ def get_media_thumbnail(
     project: Annotated[Project, Depends(get_project)],
     media_id: MediaID,
     media_service: Annotated[MediaService, Depends(get_media_service)],
+    frame_index: Annotated[int | None, Query(description="Video frame index", ge=0)] = None,
 ) -> StreamingResponse:
     """Get media thumbnail binary content"""
-    thumbnail = media_service.generate_media_thumbnail(project=project, media_id=media_id)
-    buffer = BytesIO()
-    thumbnail.save(buffer, format="JPEG")
-    buffer.seek(0)
-    return StreamingResponse(
-        buffer,
-        media_type="image/jpeg",
-        headers={
-            "Content-Disposition": f"inline; filename={media_id}.jpeg",
-            "Cache-Control": "public, max-age=31536000",
-        },
+    media = media_service.get_media_by_id(project_id=project.id, media_id=media_id)
+    if media.type == MediaType.VIDEO and frame_index is not None:
+        # Video frames can be identified by video ID and frame index
+        if frame_index >= media.frame_count:  # pyrefly: ignore[unsupported-operation]
+            raise HTTPException(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                detail=f"Video frame index {frame_index} exceeds video frames count {media.frame_count}.",
+            )
+        video_frame = media_service.get_video_frame_by_video_id_and_index(
+            project=project, video_id=media_id, frame_index=frame_index
+        )
+        if video_frame is None:
+            frame_binary = media_service.get_frame_thumbnail(project=project, video=media, frame_index=frame_index)
+            return _write_image_to_response(image=frame_binary, filename=f"{media.name}_frame_{frame_index}_thumb.jpeg")
+        media = video_frame
+
+    thumbnail = media_service.generate_media_thumbnail(project=project, media=media)
+    return _write_image_to_response(
+        image=thumbnail, filename=f"{media_id}_thumb.jpeg", cache_control="public, max-age=31536000"
     )
 
 
@@ -331,9 +411,7 @@ def set_media_annotations(
                 project=project, video_id=media_id, frame_index=frame_index
             )
             if video_frame is None:
-                video_frame = media_service.extract_video_frame(
-                    project=project, video_id=media_id, frame_index=frame_index
-                )
+                video_frame = media_service.extract_video_frame(project=project, video=media, frame_index=frame_index)
                 dataset_service.create_dataset_item(
                     project=project,
                     media=video_frame,
--- a/application/backend/app/api/routers/media.py
+++ b/application/backend/app/api/routers/media.py
@@ -1,6 +1,6 @@
 # Copyright (C) 2026 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
-
+import os.path
 from datetime import datetime
 from typing import Annotated
 from uuid import UUID
@@ -348,6 +348,8 @@ def get_media_thumbnail(
         )
 
     thumbnail_path = media_service.get_media_thumbnail_path(project=project, media=media)
+    if not os.path.exists(thumbnail_path):
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Media thumbnail file is not found.")
     return write_file_to_response(
         path=thumbnail_path, filename=f"{media.id}-thumb.jpeg", cache_control="public, max-age=31536000"
     )
--- a/application/backend/app/api/routers/media.py
+++ b/application/backend/app/api/routers/media.py
@@ -260,20 +260,10 @@ def get_media_thumbnail(
     project: Annotated[Project, Depends(get_project)],
     media_id: MediaID,
     media_service: Annotated[MediaService, Depends(get_media_service)],
-) -> StreamingResponse:
+) -> FileResponse:
     """Get media thumbnail binary content"""
-    thumbnail = media_service.generate_media_thumbnail(project=project, media_id=media_id)
-    buffer = BytesIO()
-    thumbnail.save(buffer, format="JPEG")
-    buffer.seek(0)
-    return StreamingResponse(
-        buffer,
-        media_type="image/jpeg",
-        headers={
-            "Content-Disposition": f"inline; filename={media_id}.jpeg",
-            "Cache-Control": "public, max-age=31536000",
-        },
-    )
+    thumbnail_path = media_service.get_media_thumbnail_path_by_id(project=project, media_id=media_id)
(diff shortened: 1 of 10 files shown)


application/backend/app/services/video/video_service.py
```diff
@@ -47,15 +47,13 @@ def get_video_metadata(video_path: Path) -> VideoMetadata:
...
     frame_index: int,
 ) -> np.ndarray:
     """
-    Extracts a video frame and saves it to a local FS to specified file.
+    Extracts a video frame.
 
     Args:
         video_path: Video binary file path
-        video_frame_path: Path of the file to write generated video frame to
         frame_index: Frame index
```
leoll2, line 57:
```suggestion
        frame_index: Frame index
    
    Returns:
        Extracted video frame as numpy array (RGB format)
```

application/backend/app/services/media_service.py
```diff
@@ -232,51 +231,32 @@ def get_media_thumbnail_path_by_id(self, project: Project, media_id: UUID) -> Pa
         media = self.get_media_by_id(project_id=project.id, media_id=media_id)
         return self.projects_dir / f"{project.id}/dataset/{media.id}-thumb.jpg"
 
-    def generate_media_thumbnail(self, project: Project, media_id: UUID) -> Image.Image:
+    def generate_media_thumbnail(self, project: Project, media: Media) -> Image.Image:
```
leoll2: Is this method intended to work with video frames? If not, it should validate that `media.type != MediaType.VIDEO_FRAME`.

application/backend/app/services/media_service.py
```diff
@@ -232,51 +231,32 @@ def get_media_thumbnail_path_by_id(self, project: Project, media_id: UUID) -> Pa
...
         """Regenerate a media thumbnail by its ID"""
-        media = self.get_media_by_id(project_id=project.id, media_id=media_id)
         dataset_dir = self.projects_dir / f"{project.id}/dataset"
         binary_path = dataset_dir / f"{media.id}.{media.format}"
```
leoll2: It's better to obtain the path through `get_media_binary_path` instead of reimplementing the same logic.

application/backend/app/api/routers/media.py
```diff
@@ -240,10 +294,27 @@ def get_media_binary(
     project: Annotated[Project, Depends(get_project)],
     media_id: MediaID,
     media_service: Annotated[MediaService, Depends(get_media_service)],
-) -> FileResponse:
+    frame_index: Annotated[int | None, Query(description="Video frame index", ge=0)] = None,
+) -> FileResponse | StreamingResponse:
```
leoll2: Ideally, an endpoint should not return union types: it makes life more difficult for the client because it has to handle both response types. In this case, I suggest to always return a `StreamingResponse`. 

The FastAPI docs provide an example of how to return a `StreamingResponse` from file objects: https://fastapi.tiangolo.com/advanced/custom-response/#using-streamingresponse-with-file-like-objects

There is also an example in this repo: https://github.com/open-edge-platform/training_extensions/blob/da04a37b8f20b39ab65357cec01461095bb28117/application/backend/app/api/routers/dataset_ie.py#L108

application/backend/app/api/routers/media.py
```diff
@@ -240,10 +294,27 @@ def get_media_binary(
...
     """Get media binary content"""
-    binary_path = media_service.get_media_binary_path_by_id(project_id=project.id, media_id=media_id)
     media = media_service.get_media_by_id(project_id=project.id, media_id=media_id)
+    print(media)
```
leoll2: Leftover print statement

application/backend/app/api/routers/media.py
```diff
@@ -106,6 +111,16 @@ def _parse_media_format(extension: str) -> ImageFormat | VideoFormat:
             )
 
 
+def _write_image_to_response(image: Image, filename: str, cache_control: str | None = None) -> StreamingResponse:
```
leoll2: This function could be moved to `api/routers/io_utils.py`

application/backend/app/api/routers/media.py
```diff
@@ -228,8 +243,47 @@ def get_media(
...
+        project=project,
+        video_id=media_id,
+        frame_index_from=frame_index_from,
+        frame_index_to=frame_index_to,
+    )
+    return [
+        AnnotatedVideoFrame(
+            media_id=video_frame.id,
+            frame_index=video_frame.frame_index,
+            dataset=MediaAnnotations(
```
JortBergfeld: Why is this called dataset, shouldn't it be annotation_data or dataset_item?

application/backend/app/api/routers/media.py
```diff
@@ -240,10 +294,27 @@ def get_media_binary(
...
     media_id: MediaID,
     media_service: Annotated[MediaService, Depends(get_media_service)],
-) -> FileResponse:
+    frame_index: Annotated[int | None, Query(description="Video frame index", ge=0)] = None,
+) -> FileResponse | StreamingResponse:
     """Get media binary content"""
-    binary_path = media_service.get_media_binary_path_by_id(project_id=project.id, media_id=media_id)
     media = media_service.get_media_by_id(project_id=project.id, media_id=media_id)
+    print(media)
+    if media.type == MediaType.VIDEO and frame_index is not None:
```
JortBergfeld: I feel there is too much back and forth between the router and the service, and there is too much logic in the router now. Can you create a method in the service that handles this? Similar for getting the thumbnail.

application/backend/app/api/routers/media.py
```diff
@@ -260,19 +331,28 @@ def get_media_thumbnail(
...
-    buffer.seek(0)
-    return StreamingResponse(
-        buffer,
-        media_type="image/jpeg",
-        headers={
-            "Content-Disposition": f"inline; filename={media_id}.jpeg",
-            "Cache-Control": "public, max-age=31536000",
-        },
+    media = media_service.get_media_by_id(project_id=project.id, media_id=media_id)
+    if media.type == MediaType.VIDEO and frame_index is not None:
```
JortBergfeld: I am working on a ticket #5531 to change this endpoint to return pre-generated and saved thumbnails for videos and images. Do we want to save thumbnails for (annotated) frames as well? @leoll2

application/backend/app/api/routers/media.py
```diff
@@ -240,10 +294,27 @@ def get_media_binary(
...
-) -> FileResponse:
+    frame_index: Annotated[int | None, Query(description="Video frame index", ge=0)] = None,
+) -> FileResponse | StreamingResponse:
     """Get media binary content"""
-    binary_path = media_service.get_media_binary_path_by_id(project_id=project.id, media_id=media_id)
     media = media_service.get_media_by_id(project_id=project.id, media_id=media_id)
+    print(media)
+    if media.type == MediaType.VIDEO and frame_index is not None:
+        # Video frames can be identified by video ID and frame index
+        if frame_index >= media.frame_count:  # pyrefly: ignore[unsupported-operation]
```
JortBergfeld: If you move most code to the media service, you could consider introducing a new dependency like `get_project`, called `get_frame_index` that validates the frame index directly against the media data within the DB and throws a 400:
`frame_index: Annotated[int | None, Depends(get_frame_index)]`

application/backend/app/api/routers/media.py
```diff
@@ -240,10 +294,27 @@ def get_media_binary(
...
+            raise HTTPException(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                detail=f"Video frame index {frame_index} exceeds video frames count {media.frame_count}.",
+            )
+        video_frame = media_service.get_video_frame_by_video_id_and_index(
+            project=project, video_id=media_id, frame_index=frame_index
+        )
+        if video_frame is None:
+            frame_binary = media_service.get_frame_binary(project=project, video=media, frame_index=frame_index)
+            return _write_image_to_response(image=frame_binary, filename=f"{media.name}_frame_{frame_index}.jpeg")
```
JortBergfeld: If you change the response to always be a StreamingResponse, can you also rewrite the code to move this return to the bottom. I find this if - return statement a bit difficult to read as it is quite a complex path, I prefer all return statements at the bottom, with some exceptions directly at the top.

application/backend/app/api/routers/media.py
```diff
@@ -260,19 +331,28 @@ def get_media_thumbnail(
...
+            raise HTTPException(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                detail=f"Video frame index {frame_index} exceeds video frames count {media.frame_count}.",
+            )
+        video_frame = media_service.get_video_frame_by_video_id_and_index(
+            project=project, video_id=media_id, frame_index=frame_index
+        )
+        if video_frame is None:
+            frame_binary = media_service.get_frame_thumbnail(project=project, video=media, frame_index=frame_index)
+            return _write_image_to_response(image=frame_binary, filename=f"{media.name}_frame_{frame_index}_thumb.jpeg")
```
JortBergfeld: Similar here, can you rewrite the code to have only one return statement at the bottom, I find it easier to read in this case

application/backend/app/api/schemas/media.py
```diff
@@ -159,3 +159,9 @@ class MediaAnnotations(BaseModel):
             }
         }
     }
+
+
+class AnnotatedVideoFrame(BaseModel):
+    media_id: UUID
+    frame_index: int
+    dataset: MediaAnnotations
```
JortBergfeld: Again, I think we should rename dataset to something else, like annotation data.

application/backend/app/services/video/video_service.py
```diff
@@ -66,9 +64,9 @@ def extract_video_frame(
         cap.set(cv2.CAP_PROP_POS_FRAMES, frame_index)
         # Read the frame at the requested position
         read_success, frame = cap.read()
+        cv2.cvtColor(frame, cv2.COLOR_BGR2RGB)
```
JortBergfeld: I think this statement should be below the `if not read_success` check.

application/backend/app/services/media_service.py
```diff
@@ -232,51 +231,32 @@ def get_media_thumbnail_path_by_id(self, project: Project, media_id: UUID) -> Pa
         media = self.get_media_by_id(project_id=project.id, media_id=media_id)
         return self.projects_dir / f"{project.id}/dataset/{media.id}-thumb.jpg"
 
-    def generate_media_thumbnail(self, project: Project, media_id: UUID) -> Image.Image:
+    def generate_media_thumbnail(self, project: Project, media: Media) -> Image.Image:
         """Regenerate a media thumbnail by its ID"""
```
JortBergfeld: I don't think we should regenerate any media thumbnail, so probably this method should only support video frames? Even for loading the saved thumbnails from disk the current implementation is pretty slow (due to loading the Project and Media objects from the DB to validate their existence), so in particular for videos I don't see a need to regenerate.

Perhaps this method should first check if a thumbnail exists on disk and only generate a new one if it doesn't?

application/backend/app/services/media_service.py
```diff
@@ -232,51 +231,32 @@ def get_media_thumbnail_path_by_id(self, project: Project, media_id: UUID) -> Pa
...
         binary_path = dataset_dir / f"{media.id}.{media.format}"
 
-        if media.type == MediaType.IMAGE:
-            return MediaService.generate_image_thumbnail(binary_path)
         if media.type == MediaType.VIDEO:
-            return MediaService.generate_video_thumbnail(
-                binary_path=binary_path,
-                frame_index=media.frame_count // 2,  # pyrefly: ignore
+            video_frame = MediaService._get_frame_binary_from_video_file(
+                video_path=binary_path, frame_index=media.frame_count // 2
```
JortBergfeld: If I remember correctly, generating a video thumbnail using the middle frame leads to very long generation times for long videos, so perhaps we should introduce a max frame here? But perhaps I'm wrong, so please check the code of Geti Classic on this.

application/backend/app/services/media_service.py
```diff
@@ -232,51 +231,32 @@ def get_media_thumbnail_path_by_id(self, project: Project, media_id: UUID) -> Pa
...
-        """Regenerate an image thumbnail by its path"""
+            return MediaService._crop_image_to_thumbnail(video_frame)
         try:
             with Image.open(binary_path) as image:
-                thumbnail = crop_to_thumbnail(
-                    image=image, target_width=DEFAULT_THUMBNAIL_SIZE, target_height=DEFAULT_THUMBNAIL_SIZE
-                )
-                if thumbnail.mode not in ("RGB", "L", "CMYK"):
-                    thumbnail = thumbnail.convert("RGB")
+                return MediaService._crop_image_to_thumbnail(image)
```
JortBergfeld: Although readable here, it is still weird to have two return statements, when both are doing the exact same thing. Are you sure the video_frame shouldn't be opened/generated within this try-except block as well?

application/backend/app/services/media_service.py
warrkan: Indeed, thanks!

application/backend/app/api/routers/media.py
warrkan: Thanks! Switched it to the StreamResponse

application/backend/app/api/routers/media.py
warrkan: Moved IO related functions to this module. Thanks for heads up

application/backend/app/api/routers/media.py
warrkan: Renamed it to `annotation_data`

application/backend/app/api/routers/media.py
warrkan: I've extracted similar code to handle `frame_index` parameter to a separate method, now it should be much simpler. Thanks for the suggestion

application/backend/app/api/routers/media.py
warrkan: Good point. Done

application/backend/app/api/routers/media.py
warrkan: Now it looks a bit different. Can you please check once again?

application/backend/app/api/routers/media.py
warrkan: The same here, please check once again

application/backend/app/services/video/video_service.py
warrkan: Oh, good spot. Thanks!

application/backend/app/api/routers/media.py
```diff
@@ -106,6 +111,33 @@ def _parse_media_format(extension: str) -> ImageFormat | VideoFormat:
             )
 
 
+class NotAnnotatedFrame(BaseModel):
+    video: Video
+    frame_index: int
```
leoll2: Can be moved to `application/backend/app/api/schemas/media.py` next to `AnnotatedVideoFrame`

application/backend/app/api/routers/media.py
```diff
@@ -228,8 +260,47 @@ def get_media(
...
+    },
+)
+def list_video_frames(
+    project: Annotated[Project, Depends(get_project)],
+    media_id: MediaID,
+    media_service: Annotated[MediaService, Depends(get_media_service)],
+    frame_index_from: Annotated[int, Query(ge=0)] = DEFAULT_FRAME_INDEX_FROM,
+    frame_index_to: Annotated[int, Query(ge=0)] = DEFAULT_FRAME_INDEX_TO,
+) -> list[AnnotatedVideoFrame]:
+    """Lists annotated video frames with frame index range"""
```
leoll2:
```suggestion
    """List annotated video frames with frame index range"""
```

application/backend/app/services/video/video_service.py
```diff
@@ -68,7 +69,7 @@ def extract_video_frame(
         read_success, frame = cap.read()
         if not read_success:
             raise RuntimeError(f"Cannot read frame at {frame_index} index from video: {video_path}")
-        cv2.imwrite(str(video_frame_path), frame)
+        cv2.cvtColor(frame, cv2.COLOR_BGR2RGB)
```
AlbertvanHouten: cvtColor returns the converted frame 

```suggestion
        frame = cv2.cvtColor(frame, cv2.COLOR_BGR2RGB)
```

application/backend/app/services/media_service.py
warrkan: Removed this method completely

application/backend/app/services/media_service.py
warrkan: We've moved thumbnail generating to media upload, so certain latency increase is allowed there

application/backend/app/api/routers/media.py
warrkan: Added pre-generating of video & video frame thumbnails and saving them to the disk

application/backend/app/services/media_service.py
warrkan: This method is removed

application/backend/app/services/video/video_service.py
warrkan: Good point! Fixed

application/backend/tests/integration/services/test_media_service.py
```diff
@@ -610,7 +610,7 @@ def test_create_video(
 
         # Do not generate thumbnail on video upload
         thumbnail_file_path = tmp_path / f"projects/{project.id}/dataset/{created_media.id}-thumb.jpg"
-        assert not os.path.exists(thumbnail_file_path)
+        assert os.path.exists(thumbnail_file_path)
```
AlbertvanHouten: This comment doesn't align with the test behavior anymore

warrkan: Fixed

leoll2 (approved):

application/backend/app/api/routers/media.py
```diff
@@ -1,27 +1,29 @@
...
 
 from app.api.dependencies import get_dataset_service, get_file_name_and_extension, get_media_service, get_project
+from app.api.io_utils import write_file_to_response, write_image_to_response
 from app.api.schemas.media import (
+    AnnotatedVideoFrame,
     MediaAnnotations,
     MediaView,
     MediaViewAdapter,
     MediaWithPagination,
+    NotAnnotatedFrame,
```
JortBergfeld, line 20: `NotAnnotatedVideoFrame` would be more consistent naming

application/backend/app/api/routers/media.py
```diff
@@ -106,6 +111,28 @@ def _parse_media_format(extension: str) -> ImageFormat | VideoFormat:
             )
 
 
+def _get_request_media(
```
JortBergfeld, line 114: The other dependencies are in a file called dependencies.py, perhaps this one belongs there as well? Although, I have to say the return type is a bit tricky, perhaps we should always return a media and frame index?

JortBergfeld (commented):
Looks good, some minor comments.

application/backend/app/api/routers/media.py
leoll2: Initially I also wanted to suggest moving this feature to `dependencies.py`, then I realized that it contains relatively complex and specialized logic for media, so it makes sense to stay in `api/routers/media.py`.

warrkan: This dependency is very specific to the media, that's why I've decided to leave it here for now.

application/backend/app/api/routers/media.py
warrkan: I'll update it in the next PR, thanks

PR #5635, opened 2026-02-27
Add endpoint to retrieve dataset statistics

### LINKED ISSUE #5560: Implement dataset statistics endpoint
**Endpoint**: `GET /api/projects/{project_id}/dataset/statistics`

**Response**:
```json
{
  "media_counts": {
    "images": 10,
    "videos": 3,
    "video_frames": 312,
  }
  "annotations_counts": {
    "annotated_images": 8,
    "annotated_videos": 2,
    "annotated_video_frames": 29,
    "instances": 56,
    "instances_per_label": [
      {
        "label_id": "5fffd195-7766-4171-8efe-4064a6eb0e95",
        "instances": 24
      },
      {
        "label_id": "20f1defc-8d40-47ff-9f9d-8e82f0ef224d",
        "instances": 32
      },
    ]
  }
}
```
Definitions:
- **Media** -> an image, a video or a video frame
- **Annotated video** -> A video with one or more annotated frames
- **Annotation instance** -> A shape with labels; for classification, there is always 1 instance per image (full-image rectangle); for detection and segmentation, there are as many instances as bboxes/polygons.

JortBergfeld:
Add endpoint for retrieving dataset statistics.

The main logic is in `dataset_item_repo` where the statistics are obtained as a whole. The dataset service and router are responsible for transforming the dictionary of statistics to a `DatasetStatisticsView`.

Resolves #5560 

### How to test

Unit test added for router and integration test added for dataset service, which calls and checks the dataset item repository.

### Checklist

### DIFF
--- a/application/backend/app/api/schemas/dataset.py
+++ b/application/backend/app/api/schemas/dataset.py
@@ -77,31 +77,33 @@ class StagedDatasetView(BaseRequiredIDModel):
         return data
 
 
-class MediaCounts(BaseModel):
+class MediaCountsView(BaseModel):
     """Media counts"""
 
-    images: int
-    videos: int
-    video_frames: int
+    images: int = Field(0, description="Number of images in dataset")
+    videos: int = Field(0, description="Number of videos in dataset")
+    video_frames: int = Field(0, description="Number of video frames in dataset")
 
     model_config = {"json_schema_extra": {"example": {"images": 10, "videos": 3, "video_frames": 312}}}
 
 
-class InstancesPerLabel(BaseModel):
-    label_id: UUID
-    instances: int
+class InstancesPerLabelView(BaseModel):
+    label_id: UUID = Field(..., description="Unique identifier of label")
+    instances: int = Field(0, description="Number of video frames in dataset")
 
     model_config = {
         "json_schema_extra": {"example": {"label_id": "5fffd195-7766-4171-8efe-4064a6eb0e95", "instances": 24}}
     }
 
 
-class AnnotationCounts(BaseModel):
-    annotated_images: int
-    annotated_videos: int
-    annotated_video_frames: int
-    instances: int
-    instances_per_label: list[InstancesPerLabel]
+class AnnotationCountsView(BaseModel):
+    annotated_images: int = Field(0, description="Number of annotated images in dataset")
+    annotated_videos: int = Field(0, description="Number of annotated videos in dataset")
+    annotated_video_frames: int = Field(0, description="Number of annotated video frames in dataset")
+    instances: int = Field(0, description="Number of annotation shapes in dataset")
+    instances_per_label: list[InstancesPerLabelView] = Field(
+        [], description="List of number of annotation shapes per label"
+    )
 
     model_config = {
         "json_schema_extra": {
@@ -124,8 +126,10 @@ class DatasetStatisticsView(BaseModel):
     Dataset statistics
     """
 
-    media_counts: MediaCounts
-    annotations_counts: AnnotationCounts
+    media_counts: MediaCountsView = Field(..., description="Number of media per media type in dataset")
+    annotations_counts: AnnotationCountsView = Field(
+        ..., description="Number of annotated media per media type en labels"
+    )
 
     model_config = {
         "json_schema_extra": {
--- a/application/backend/app/api/schemas/dataset.py
+++ b/application/backend/app/api/schemas/dataset.py
@@ -89,7 +89,7 @@ class MediaCountsView(BaseModel):
 
 class InstancesPerLabelView(BaseModel):
     label_id: UUID = Field(..., description="Unique identifier of label")
-    instances: int = Field(0, description="Number of video frames in dataset")
+    instances: int = Field(0, description="Number of annotation instances with this label in dataset")
 
     model_config = {
         "json_schema_extra": {"example": {"label_id": "5fffd195-7766-4171-8efe-4064a6eb0e95", "instances": 24}}
@@ -128,7 +128,7 @@ class DatasetStatisticsView(BaseModel):
 
     media_counts: MediaCountsView = Field(..., description="Number of media per media type in dataset")
     annotations_counts: AnnotationCountsView = Field(
-        ..., description="Number of annotated media per media type en labels"
+        ..., description="Number of annotated media per media type and labels"
     )
 
     model_config = {
--- a/application/backend/app/api/schemas/dataset.py
+++ b/application/backend/app/api/schemas/dataset.py
@@ -1,6 +1,8 @@
 # Copyright (C) 2026 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 
+from uuid import UUID
+
 from pydantic import BaseModel, Field, model_validator
 
 from app.core.models import BaseRequiredIDModel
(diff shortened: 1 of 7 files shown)


application/backend/app/api/schemas/dataset.py
```diff
@@ -72,3 +74,76 @@ def populate_metadata(cls, data: object) -> object:
...
+    images: int = Field(0, description="Number of images in dataset")
+    videos: int = Field(0, description="Number of videos in dataset")
+    video_frames: int = Field(0, description="Number of video frames in dataset")
+
+    model_config = {"json_schema_extra": {"example": {"images": 10, "videos": 3, "video_frames": 312}}}
+
+
+class InstancesPerLabelView(BaseModel):
+    label_id: UUID = Field(..., description="Unique identifier of label")
+    instances: int = Field(0, description="Number of video frames in dataset")
```
leoll2: This description is wrong, probably a copy-paste.

application/backend/app/api/schemas/dataset.py
```diff
@@ -72,3 +74,76 @@ def populate_metadata(cls, data: object) -> object:
...
+
+
+class DatasetStatisticsView(BaseModel):
+    """
+    Dataset statistics
+    """
+
+    media_counts: MediaCountsView = Field(..., description="Number of media per media type in dataset")
+    annotations_counts: AnnotationCountsView = Field(
+        ..., description="Number of annotated media per media type en labels"
```
leoll2:
```suggestion
        ..., description="Counters of annotated media, aggregated by media type or label"
```

application/backend/app/models/dataset.py
```diff
@@ -21,3 +25,73 @@ class StagedDataset(BaseRequiredIDModel):
...
+    @model_validator(mode="before")
+    @classmethod
+    def populate_media_counts(cls, data: object) -> object:
+        if isinstance(data, dict):
+            return {
+                "images": data.get("images", 0),
+                "videos": data.get("videos", 0),
+                "video_frames": data.get("video_frames", 0),
+            }
+        return data
```
leoll2: Is this validator actually needed? The attributes names and types match exactly the dictionary keys, I think `model_validate` could work fine without it. Same for the other classes except `InstancesPerLabel`, which needs the validator because of the UUID type casting.

application/backend/app/repositories/dataset_item_repo.py
```diff
@@ -219,3 +220,44 @@ def get_subset_distribution(self) -> dict[str, int]:
...
+
+        # Annotation Counts (annotated images, videos, video frames):
+        stmt = (
+            select(MediaDB.type, func.count(DatasetItemDB.id).label("count"))
+            .join(DatasetItemDB, DatasetItemDB.id == MediaDB.id)
+            .where(DatasetItemDB.project_id == self.project_id, DatasetItemDB.annotation_data.isnot(None))
+            .group_by(MediaDB.type)
+        )
+        result = self.db.execute(stmt)
+        annotated_counts: dict[str, Any] = {f"annotated_{row.type}s": row.count for row in result}
```
leoll2: Are should sure this works for `video` type? Afaik _videos_ don't have a record in the `DatasetItemDB` table (_video frames_ do) so they will be filtered out during the join. I think the number of annotated videos has to be computed separately, by querying the number of distinct `MediaDB.video_id` values across all annotated media of `type=video_frame`.

application/backend/app/repositories/dataset_item_repo.py
```diff
@@ -219,3 +220,44 @@ def get_subset_distribution(self) -> dict[str, int]:
...
+
+        # Total instances:
+        stmt = select(DatasetItemDB.annotation_data).where(
+            DatasetItemDB.project_id == self.project_id, DatasetItemDB.annotation_data.isnot(None)
+        )
+        result = list(self.db.execute(stmt))
+        annotated_counts["instances"] = sum(len(item.annotation_data) for item in result)
```
leoll2: This query may return a large number of annotation objects to be loaded in-memory; since you only need their `len()`, you can exploit the [`json_array_length`](https://sqlite.org/json1.html#the_json_array_length_function) method provided by SQLite.

application/backend/app/repositories/dataset_item_repo.py
```diff
@@ -219,3 +220,44 @@ def get_subset_distribution(self) -> dict[str, int]:
...
+        labels_counts = Counter(
+            label["id"] for item in result for annotation in item.annotation_data for label in annotation["labels"]
+        )
+        annotated_counts["instances_per_label"] = [
+            {"label_id": label_id, "instances": count} for label_id, count in labels_counts.items()
+        ]
```
leoll2, line 289: These counters can be extracted more efficiently (from a memory footprint perspective) from the `DatasetItemLabelDB` table.

JortBergfeld: I did that at first, but the dataset item label only contains a reference to the whole dataset item, not to the annotations (for example: detecting 3 planes, 2 Boeing, 1 Airbus in one picture). So if we want to count the instances per label we have to inspect the annotation data itself, which is stored as json and can't be properly inspected by the DB itself, as far as I know.

application/backend/app/repositories/dataset_item_repo.py
JortBergfeld: I would then need to run the same query twice, for the instances per label part. Probably to be safe that is better.

application/backend/app/repositories/dataset_item_repo.py
JortBergfeld: You are correct. Sorry, video support is not yet implemented making it a bit harder to test. I am working on a better integration test to catch these errors.

application/backend/app/models/dataset.py
JortBergfeld: I will test this.

application/backend/app/repositories/dataset_item_repo.py
leoll2: You are right, my suggestion to use `DatasetItemLabelDB` is not a viable approach.

> we have to inspect the annotation data itself, which is stored as json and can't be properly inspected by the DB itself

You can take a look at the JSON functions offered by SQLite, especially `json_extract` (exposed as `funct.json_extract` in SQLAlchemy): you may be able to extract the labels without the rest of the unnecessary data (shapes, ...)

https://sqlite.org/json1.html#jex

application/backend/app/repositories/dataset_item_repo.py
```diff
@@ -219,3 +220,74 @@ def get_subset_distribution(self) -> dict[str, int]:
...
+                DatasetItemDB.user_reviewed,
+            )
+            .group_by(MediaDB.type)
+        )
+        result = self.db.execute(annotated_images_frames_stmt)
+        annotated_counts: dict[str, Any] = {f"annotated_{row.type}s": row.count for row in result}
+
+        # Annotation Counts (annotated videos)
+        annotated_video_stmt = (
+            select(func.count(func.distinct(MediaDB.id)).label("count"))
```
leoll2: Shouldn't it be `distinct(MediaDB.video_id)`? Otherwise you're counting the distinct video frames instead of the videos.

application/backend/tests/integration/services/test_dataset_service.py
```diff
@@ -476,6 +476,138 @@ def fxt_project_with_subset_items(fxt_project_with_pipeline, db_session) -> tupl
...
+    db_session.flush()
+    item_img = DatasetItemDB(
+        id=str(media_img.id),
+        project_id=str(project.id),
+        subset="unassigned",
+        created_at=datetime.fromisoformat("2025-02-01T00:00:00Z"),
+    )
+    db_session.add(item_img)
+    db_dataset_items.append(item_img)
+    db_session.flush()
```
leoll2: This fixture implementation is long and repetitive, if flushes for every object while I think a single flush at the end would be sufficient. Maybe refactor like this?
```python
db_objects = [
    MediaDB(...),
    MediaDB(...),
    ...
    DatasetItemDB(...),
    ...
]
db_session.add_all(db_objects)
db_session.flush()
```

application/backend/tests/integration/services/test_dataset_service.py
```diff
@@ -476,6 +476,138 @@ def fxt_project_with_subset_items(fxt_project_with_pipeline, db_session) -> tupl
...
+        created_at=datetime.fromisoformat("2025-02-01T00:00:00Z"),
+    )
+    db_session.add(media_vid)
+    db_session.flush()
+    item_vid = DatasetItemDB(
+        id=str(media_vid.id),
+        project_id=str(project.id),
+        subset="unassigned",
+        created_at=datetime.fromisoformat("2025-02-01T00:00:00Z"),
+    )
```
leoll2: This is wrong, videos don't have a dataset item record in the DB. Only annotated video frames do.

application/backend/tests/integration/services/test_dataset_service.py
```diff
@@ -476,6 +476,138 @@ def fxt_project_with_subset_items(fxt_project_with_pipeline, db_session) -> tupl
...
+    db_session.add(item_vid)
+    db_dataset_items.append(item_vid)
+    db_session.flush()
+
+    # Video frame (annotated with label_1)
```
leoll2, line 556: For the sake of testing, it would be more reliable to have 2 or more annotated frames.

application/backend/tests/integration/services/test_dataset_service.py
JortBergfeld: Unfortunately, there is no guarantee that the statements are run in the correct order, leading to foreign key errors. Therefore I had to flush in between the statements. I will look into streamlining the process.

leoll2 (approved):

PR #5640, opened 2026-02-27
Serve files as FastAPI FileResponse

warrkan:
To fix issues with video playback, FileResponse is used to serve static content to the client.
Additionally, byte range requests should be handled more efficiently in this case and we don't need to process data into intermediate buffer (even with chunks) in case of large files.

### DIFF
--- a/application/backend/app/api/routers/media.py
+++ b/application/backend/app/api/routers/media.py
@@ -7,7 +7,7 @@ from uuid import UUID
 
 from fastapi import APIRouter, Body, Depends, File, HTTPException, Query, UploadFile, status
 from fastapi.openapi.models import Example
-from starlette.responses import Response
+from starlette.responses import FileResponse, StreamingResponse
 
 from app.api.dependencies import get_dataset_service, get_file_name_and_extension, get_media_service, get_project
 from app.api.io_utils import write_file_to_response, write_image_to_response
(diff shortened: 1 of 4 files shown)


application/backend/app/api/routers/media.py
```diff
@@ -306,7 +306,7 @@ def get_media_binary(
     project: Annotated[Project, Depends(get_project)],
     media: Annotated[Media | NotAnnotatedFrame, Depends(_get_request_media)],
     media_service: Annotated[MediaService, Depends(get_media_service)],
-) -> StreamingResponse:
+) -> Response:
```
leoll2: Should be `FileResponse`. `Response` is too generic, it includes anything like `HTMLResponse`, `JSONResponse`, ... which are clearly not what this endpoint returns.

application/backend/app/api/routers/media.py
```diff
@@ -332,7 +332,7 @@ def get_media_thumbnail(
...
```
leoll2: Also here, should be `FileResponse`

application/backend/app/api/routers/media.py
warrkan: There are two execution branches, one returns FileResponse, while second - StreamingResponse. That's why a generic response is set here

leoll2: I see. What about `FileResponse | StreamingResponse` to narrow down a bit?

leoll2 (approved):

PR #5651, opened 2026-03-02
Publish inference result in order

### LINKED ISSUE #5437: Inference on video goes back several frames and continues
1. Create a project, import images, annotate and train a model.
2. Configure a pipeline that takes a video file as source (I am using a 1024x720 30 s long video).
3. Start the pipeline.
The played video is shaking, which is expected if the inference takes too long, but it is "shaking backwards", which is not expected. The result can be seen in the attachment.

https://github.com/user-attachments/assets/7931e35e-09fd-45de-aa18-dfb8c9fa3a07

JortBergfeld:
### Summary

This PR ensures the predicitons generated async from the source, are added to the prediction queue for the sink and the webstream in order. This fixes a bug where the webstream (and also the sink) is jumping back and forth in a video, like a stuttering effect.

Adds a `PredictionBuffer` to the `InferenceWorker` that has a list of timestamps (inference start time) in the expected order, and a dictionary of timestamps - predictions (`StreamData`). When a prediction is finished:
-  it is first added to the buffer, 
- then `get_ready_predictions` outputs the predictions in the order it was requested, until a prediction is not found in the buffer,
- if too many (10 by default) predictions are expected, without the first expected being generated, the oldest expected is dropped
- if a prediction is generated, but the timestamp is not expected, the prediction is dropped. A warning will be logged

When a new model is loaded, the `PredictionBuffer` is cleared. However, the prediction queue is not cleared, or the webstream queue is not cleared, so one will still see frames from previous streams, before the new stream starts.
 
Resolves #5437

### How to test

To test, run the server with demos, go to the project with airplanes, or horses, setup a pipeline and enable it. When the webstream is started, the video runs smoothly, that is, the video never jumps backwards a couple of frames.

### Checklist

### DIFF
--- a/application/backend/app/workers/inference.py
+++ b/application/backend/app/workers/inference.py
@@ -103,9 +103,6 @@ class InferenceWorker(BaseProcessWorker):
 
         # To ensure broadcasting frames in order
         self._prediction_buffer = PredictionBuffer()
-        self._frame_index = 0
-        self._next_to_broadcast = 0
-        self._result_buffer = {}
 
     def setup(self) -> None:
         super().setup()
--- a/application/backend/app/workers/inference.py
+++ b/application/backend/app/workers/inference.py
@@ -3,7 +3,7 @@
 
 import multiprocessing as mp
 import queue
-from collections.abc import Generator
+import threading
 from dataclasses import dataclass
 from multiprocessing.synchronize import Event as EventClass
 from multiprocessing.synchronize import Lock
@@ -34,16 +34,16 @@ class InferenceWorkerConfig:
     logger_: LoguruLogger
 
 
-class PredictionBuffer:
+class PredictionReorderBuffer:
     """Buffers predictions to ensure async generated predictions are fed to the prediction queue in order"""
 
     def __init__(self, max_size: int = 10):
         self._buffer: dict[float, StreamData] = {}
         self._expected_timestamps: list[float] = []
         self._max_size = max_size
-        self._lock = mp.Lock()
+        self._lock = threading.Lock()
 
-    def append_expected_timestamp(self, timestamp: float) -> None:
+    def register_expected_timestamp(self, timestamp: float) -> None:
         with self._lock:
             self._expected_timestamps.append(timestamp)
             if len(self._expected_timestamps) > self._max_size:
@@ -51,7 +51,7 @@ class PredictionBuffer:
                 if oldest_timestamp in self._buffer:
                     del self._buffer[oldest_timestamp]
 
-    def add_to_buffer(self, timestamp: float, stream_data: StreamData) -> None:
+    def add_prediction_for_timestamp(self, timestamp: float, stream_data: StreamData) -> None:
         with self._lock:
             if timestamp in self._expected_timestamps:
                 self._buffer[timestamp] = stream_data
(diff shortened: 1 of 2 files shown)


application/backend/app/workers/inference.py
```diff
@@ -33,6 +34,51 @@ class InferenceWorkerConfig:
...
 
 
+class PredictionBuffer:
+    """Buffers predictions to ensure async generated predictions are fed to the prediction queue in order"""
+
+    def __init__(self, max_size: int = 10):
+        self._buffer: dict[float, StreamData] = {}
+        self._expected_timestamps: list[float] = []
+        self._max_size = max_size
+        self._lock = mp.Lock()
```
leoll2: Since this lock is used to protect `self._buffer`, which is accessed by _threads_ spawned by the same process, you can use a `threading.Lock` instead of a `mp.Lock`, it is lighter and more efficient.

application/backend/app/workers/inference.py
```diff
@@ -33,6 +34,51 @@ class InferenceWorkerConfig:
...
+                    timestamp,
+                )
+
+    def get_ready_predictions(self) -> Generator[StreamData]:
+        with self._lock:
+            if not self._expected_timestamps:
+                return
+            while self._expected_timestamps[0] in self._buffer:
+                timestamp = self._expected_timestamps.pop(0)
+                yield self._buffer.pop(timestamp)
```
leoll2: Any reason for yielding the predictions one by one instead of returning all of them in a tuple? Yielding inside a lock is a bit of an anti-pattern because the lock is never released between yields, delaying the other threads that are waiting for the lock. Instead, returning all ready predictions together would minimize the critical section.

application/backend/app/workers/inference.py
```diff
@@ -33,6 +34,51 @@ class InferenceWorkerConfig:
...
+
+    def append_expected_timestamp(self, timestamp: float) -> None:
+        with self._lock:
+            self._expected_timestamps.append(timestamp)
+            if len(self._expected_timestamps) > self._max_size:
+                oldest_timestamp = self._expected_timestamps.pop(0)
+                if oldest_timestamp in self._buffer:
+                    del self._buffer[oldest_timestamp]
+
+    def add_to_buffer(self, timestamp: float, stream_data: StreamData) -> None:
```
leoll2: Rename suggestions:

- `append_expected_timestamp` -> `register_expected_timestamp`
- `add_to_buffer` -> `add_prediction_for_timestamp`

The proposed names are agnostic to the internal implementation of PredictionBuffer.

application/backend/app/workers/inference.py
```diff
@@ -33,6 +34,51 @@ class InferenceWorkerConfig:
     logger_: LoguruLogger
 
 
+class PredictionBuffer:
```
leoll2: Rename suggestion: `PredictionReorderBuffer`.

application/backend/app/workers/inference.py
JortBergfeld: Changed to list. I implemented this before the lock, so I considered potential issues with new predictions coming in while processing. But that was unsafe in its own way, so I had to introduce a lock.

leoll2 (dismissed):

leoll2 (commented):
LGTM, please add some tests

leoll2 (approved):

PR #5661, opened 2026-03-02
Improve sam ux

### LINKED ISSUE #5646: Improve SAM performance
In Geti Classic the Auto segmentation tool, once initialized, was instantly active for subsequent images during annotation.
For Geti tune, the model is being initialized every time a new image is displayed.

We need to improve the UX

jpggvilaca:
### Summary

* Improve readability of some queries (on top of what Dawid already had improved)
* Replace SAM useEffect (to load the worker/instantiate the class) with a RQ approach
* Add hook to fetch/prefetch data/annotations (images or video frames)
* Leverage encodingQuery to fetch/prefetch SAM encoding

On the video you can see the awesome speed if the user selected the NEXT media item, but if the user selects stuff at random the speed will be more or less the same.

Closes https://github.com/open-edge-platform/training_extensions/issues/5646

Video:
rec

Image:
rec

I thought about tests but if i added unit tests, then we would be testing react-query (useless), and if i added component maybe we could check speed or if the loading didnt even show? I think that would be veeeery flaky so if you have any suggestions lmk

### Checklist

### DIFF
--- a/application/ui/src/features/annotator/hooks/use-load-image-query.hook.ts
+++ b/application/ui/src/features/annotator/hooks/use-load-image-query.hook.ts
@@ -9,24 +9,28 @@ import { isVideoFrame } from '../../../shared/media-item-utils';
 import { getMediaBinaryUrl, getVideoFrameBinaryUrl } from '../../../shared/media-url.utils';
 import { getImageData, loadImage } from '../tools/utils';
 
+export const getLoadImageQueryKey = (projectId: string, media: Media) => {
+    return isVideoFrame(media)
+        ? ['mediaItem', projectId, media.id, media.frame_number]
+        : ['mediaItem', projectId, media.id];
+};
+
+export const loadImageQueryFn = async (projectId: string, media: Media) => {
+    const url = isVideoFrame(media)
+        ? getVideoFrameBinaryUrl(projectId, media.id, media.frame_number)
+        : getMediaBinaryUrl(projectId, media.id);
+
+    const image = await loadImage(url);
+
+    return getImageData(image);
+};
+
 export const useLoadImageQuery = (media: Media) => {
     const projectId = useProjectIdentifier();
 
-    const queryKey = isVideoFrame(media)
-        ? ['mediaItem', projectId, media.id, media.frame_number]
-        : ['mediaItem', projectId, media.id];
-
     return useQuery({
-        queryKey,
-        queryFn: async () => {
-            const url = isVideoFrame(media)
-                ? getVideoFrameBinaryUrl(projectId, media.id, media.frame_number)
-                : getMediaBinaryUrl(projectId, media.id);
-
-            const image = await loadImage(url);
-
-            return getImageData(image);
-        },
+        queryKey: getLoadImageQueryKey(projectId, media),
+        queryFn: () => loadImageQueryFn(projectId, media),
         placeholderData: keepPreviousData,
         // The image of a media item never changes so we don't want to refetch stale data
         staleTime: Infinity,
--- a/application/ui/src/features/annotator/hooks/use-load-image-query.hook.ts
+++ b/application/ui/src/features/annotator/hooks/use-load-image-query.hook.ts
@@ -1,7 +1,7 @@
 // Copyright (C) 2025 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 
-import { keepPreviousData, useQuery } from '@tanstack/react-query';
+import { keepPreviousData, queryOptions, useQuery } from '@tanstack/react-query';
 import { useProjectIdentifier } from 'hooks/use-project-identifier.hook';
 
 import { Media } from '../../../constants/shared-types';
@@ -9,31 +9,29 @@ import { isVideoFrame } from '../../../shared/media-item-utils';
 import { getMediaBinaryUrl, getVideoFrameBinaryUrl } from '../../../shared/media-url.utils';
 import { getImageData, loadImage } from '../tools/utils';
 
-export const getLoadImageQueryKey = (projectId: string, media: Media) => {
-    return isVideoFrame(media)
-        ? ['mediaItem', projectId, media.id, media.frame_number]
-        : ['mediaItem', projectId, media.id];
-};
+export const loadImageQueryOptions = (projectId: string, media: Media) =>
+    queryOptions({
+        queryKey: isVideoFrame(media)
+            ? ['mediaItem', projectId, media.id, media.frame_number]
+            : ['mediaItem', projectId, media.id],
+        queryFn: async () => {
+            const url = isVideoFrame(media)
+                ? getVideoFrameBinaryUrl(projectId, media.id, media.frame_number)
+                : getMediaBinaryUrl(projectId, media.id);
 
-export const loadImageQueryFn = async (projectId: string, media: Media) => {
-    const url = isVideoFrame(media)
-        ? getVideoFrameBinaryUrl(projectId, media.id, media.frame_number)
-        : getMediaBinaryUrl(projectId, media.id);
+            const image = await loadImage(url);
 
-    const image = await loadImage(url);
-
-    return getImageData(image);
-};
+            return getImageData(image);
+        },
+        staleTime: Infinity,
+        retry: 0,
+    });
 
 export const useLoadImageQuery = (media: Media) => {
     const projectId = useProjectIdentifier();
 
     return useQuery({
-        queryKey: getLoadImageQueryKey(projectId, media),
-        queryFn: () => loadImageQueryFn(projectId, media),
+        ...loadImageQueryOptions(projectId, media),
         placeholderData: keepPreviousData,
-        // The image of a media item never changes so we don't want to refetch stale data
-        staleTime: Infinity,
-        retry: 0,
     });
 };
--- a/application/ui/src/features/annotator/tools/segment-anything-tool/segment-anything-tool.component.tsx
+++ b/application/ui/src/features/annotator/tools/segment-anything-tool/segment-anything-tool.component.tsx
@@ -10,7 +10,7 @@ import selectionCursor from '../../../../assets/icons/selection.svg?url';
 import { useZoom } from '../../../../components/zoom/zoom.provider';
 import type { Label } from '../../../../constants/shared-types';
 import type { Annotation, RegionOfInterest, Shape } from '../../../../shared/types';
-import { useNextMediaPrefetch } from '../../../dataset/media-preview/utils';
+import { useNextMediaItem } from '../../../dataset/media-preview/utils';
 import { AnnotationShape } from '../../annotations/annotation-shape/annotation-shape.component';
 import { MaskAnnotations } from '../../annotations/mask-annotations.component';
 import { useAnnotatorLabels } from '../../annotator-labels-provider.component';
@@ -62,10 +62,10 @@ export const SegmentAnythingTool = () => {
     const zoom = useZoom();
     const { roi, image, mediaItem } = useSelectedMediaItem();
     const { items } = useGetDatasetMediaItems();
-    const { nextMediaItem, nextImage, isNextImageReady } = useNextMediaPrefetch(mediaItem, items);
+    const nextMediaItem = useNextMediaItem(mediaItem, items);
     const { selectedLabel } = useAnnotatorLabels();
     const { addAndSelectAnnotations } = useAddAndSelectAnnotations();
-    const { isLoading, decodingQueryFn } = useSegmentAnythingModel({ nextMediaItem, nextImage, isNextImageReady });
+    const { isLoading, decodingQueryFn } = useSegmentAnythingModel({ nextMediaItem });
     const throttledDecodingQueryFn = useSingleStackFn(decodingQueryFn);
     const cancellableThrottledDecodingQueryFn = useWithCancel(throttledDecodingQueryFn);
 
--- a/application/ui/src/features/annotator/tools/segment-anything-tool/segment-anything-tool.component.tsx
+++ b/application/ui/src/features/annotator/tools/segment-anything-tool/segment-anything-tool.component.tsx
@@ -4,11 +4,13 @@
 import { PointerEvent, useRef, useState } from 'react';
 
 import { clampPointBetweenImage } from '@geti/smart-tools/utils';
+import { useGetDatasetMediaItems } from 'hooks/use-get-dataset-media-items.hook';
 
 import selectionCursor from '../../../../assets/icons/selection.svg?url';
 import { useZoom } from '../../../../components/zoom/zoom.provider';
 import type { Label } from '../../../../constants/shared-types';
 import type { Annotation, RegionOfInterest, Shape } from '../../../../shared/types';
+import { useNextMediaPrefetch } from '../../../dataset/media-preview/utils';
 import { AnnotationShape } from '../../annotations/annotation-shape/annotation-shape.component';
 import { MaskAnnotations } from '../../annotations/mask-annotations.component';
 import { useAnnotatorLabels } from '../../annotator-labels-provider.component';
@@ -58,10 +60,12 @@ export const SegmentAnythingTool = () => {
     const ref = useRef<SVGSVGElement>(null);
 
     const zoom = useZoom();
-    const { roi, image } = useSelectedMediaItem();
+    const { roi, image, mediaItem } = useSelectedMediaItem();
+    const { items } = useGetDatasetMediaItems();
+    const { nextMediaItem, nextImage, isNextImageReady } = useNextMediaPrefetch(mediaItem, items);
     const { selectedLabel } = useAnnotatorLabels();
     const { addAndSelectAnnotations } = useAddAndSelectAnnotations();
-    const { isLoading, decodingQueryFn } = useSegmentAnythingModel();
+    const { isLoading, decodingQueryFn } = useSegmentAnythingModel({ nextMediaItem, nextImage, isNextImageReady });
     const throttledDecodingQueryFn = useSingleStackFn(decodingQueryFn);
     const cancellableThrottledDecodingQueryFn = useWithCancel(throttledDecodingQueryFn);
 
--- a/application/ui/src/features/annotator/tools/segment-anything-tool/use-segment-anything.hook.ts
+++ b/application/ui/src/features/annotator/tools/segment-anything-tool/use-segment-anything.hook.ts
@@ -2,13 +2,14 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { EncodingOutput } from '@geti/smart-tools/segment-anything';
-import { queryOptions, useQuery } from '@tanstack/react-query';
+import { queryOptions, useQuery, useQueryClient } from '@tanstack/react-query';
 import { Remote, wrap } from 'comlink';
 import { useProject } from 'hooks/api/project.hook';
 
 import type { Media } from '../../../../constants/shared-types';
 import { isVideoFrame } from '../../../../shared/media-item-utils';
 import { isDetectionTask } from '../../../project/task-type-guards';
+import { loadImageQueryOptions } from '../../hooks/use-load-image-query.hook';
 import { useSelectedMediaItem } from '../../selected-media-item-provider.component';
 import type {
     SegmentAnythingWorkerApi,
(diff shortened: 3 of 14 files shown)


application/ui/src/features/dataset/media-preview/media-preview.component.tsx
```diff
@@ -32,9 +43,66 @@ type MediaPreviewContentProps = {
...
+            queryFn: segmentAnythingWorkerQueryFn('SEGMENT_ANYTHING_ENCODER'),
+            staleTime: Infinity,
+        });
+
+        queryClient.prefetchQuery({
+            queryKey: getSegmentAnythingEncodingQueryKey(nextItem),
+            queryFn: () => segmentAnythingEncodingQueryFn(encoderModel, nextImage),
+            staleTime: Infinity,
+            gcTime: 3600 * 15,
+        });
```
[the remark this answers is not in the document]
jpggvilaca: It was defined before. Also i dont want to abuse the browser's memory.

application/ui/src/features/dataset/media-preview/media-preview.component.tsx
```diff
@@ -32,9 +43,66 @@ type MediaPreviewContentProps = {
...
+    items,
+    selectedItem,
+}: {
+    queryClient: ReturnType<typeof useQueryClient>;
+    projectId: string;
+    items: Media[];
+    selectedItem: Media;
+}) => {
+    const prefetch = async () => {
+        const selectedIndex = items.findIndex((item) => item.id === selectedItem.id);
```
dwesolow: it will work only for images, for video frames it won't work.

jpggvilaca: Good point! Should i leverage the util you added `getNextMediaItem` ?

dwesolow: yeah, should work ;)

jpggvilaca: Updated

camiloHimura:
Not to address it right now, but I think we could use TanStack Query’s [queryOptions](https://tanstack.com/query/v5/docs/framework/react/reference/queryOptions) to centralize the query keys and query functions. That way we’d have a single, reusable, type-safe source of truth.

camiloHimura (dismissed):

application/ui/src/features/annotator/hooks/use-load-image-query.hook.ts
```diff
@@ -9,24 +9,28 @@ import { isVideoFrame } from '../../../shared/media-item-utils';
 import { getMediaBinaryUrl, getVideoFrameBinaryUrl } from '../../../shared/media-url.utils';
 import { getImageData, loadImage } from '../tools/utils';
 
-export const useLoadImageQuery = (media: Media) => {
-    const projectId = useProjectIdentifier();
-
-    const queryKey = isVideoFrame(media)
+export const getLoadImageQueryKey = (projectId: string, media: Media) => {
```
dwesolow: IMO we should use `queryOptions` to create `loadImageQueryOptions` and reuse that instead of reusing single functions.

application/ui/src/features/dataset/media-preview/media-preview.component.tsx
```diff
@@ -32,9 +44,74 @@ type MediaPreviewContentProps = {
     onSelectedMediaItem: (item: Media) => void;
 };
 
+const getNextMediaItemForPrefetch = (selectedItem: Media, items: Media[]): Media | undefined => {
+    const step = selectedItem.type === 'video_frame' ? selectedItem.frame_stride : 1;
```
dwesolow: we have util `isVideoFrame`

dwesolow: Step can be changed in the video player toolbar as well to be either 1 or `frame_stride` so this condition is not true. We should use `step` from the video player (source of truth for that).

application/ui/src/features/dataset/media-preview/media-preview.component.tsx
```diff
@@ -32,9 +44,74 @@ type MediaPreviewContentProps = {
...
+    const nextItem = getNextMediaItem(selectedItem, items, step);
+
+    // getNextMediaItem never returns undefined. When there is no next item, it returns the same item.
```
dwesolow: If we need we can return `undefined` in that case.

jpggvilaca: Ok then, updated

application/ui/src/features/dataset/media-preview/media-preview.component.tsx
jpggvilaca: updated to use annotator's already existing `useNextMedia`

jpggvilaca: Updated

application/ui/src/features/annotator/hooks/use-load-image-query.hook.ts
jpggvilaca: Updated

jpggvilaca:
> Not to address it right now, but I think we could use TanStack Query’s [queryOptions](https://tanstack.com/query/v5/docs/framework/react/reference/queryOptions) to centralize the query keys and query functions. That way we’d have a single, reusable, type-safe source of truth.

Updated

application/ui/src/features/annotator/tools/segment-anything-tool/use-segment-anything.hook.ts
```diff
@@ -1,82 +1,103 @@
...
+        ? ['segment-anything-model', 'encoding', mediaItem.id, mediaItem.frame_number]
+        : ['segment-anything-model', 'encoding', mediaItem.id];
+};
 
-    return modelRef.current;
+export const segmentAnythingEncodingQueryFn = (
+    model: SegmentAnythingRemoteModel,
+    image: ImageData
+): Promise<EncodingOutput> => {
+    return model.processEncoder(image);
```
dwesolow: imo overkill a bit.

application/ui/src/features/annotator/tools/tool-manager.component.tsx
```diff
@@ -1,14 +1,26 @@
...
 import { MagneticLasso } from './magnetic-lasso/magnetic-lasso.component';
 import { PolygonTool } from './polygon-tool/polygon-tool.component';
 import { SegmentAnythingTool } from './segment-anything-tool/segment-anything-tool.component';
+import { usePreloadSAMWorkers } from './segment-anything-tool/use-segment-anything.hook';
 
 export const ToolManager = () => {
     const { activeTool } = useTool();
+    const { data: project } = useProject();
+
+    const isPreloadEnabled = project !== undefined && isPrefetchEnabledForTask(project.task.task_type);
```
dwesolow: why do we need this?

application/ui/src/features/dataset/media-preview/api/use-annotations-query.ts
```diff
@@ -15,64 +15,79 @@ const isUnannotatedError = (error: unknown): boolean => {
...
-            return data;
-        },
+    return data;
+};
+
+export const useAnnotationsQuery = (media: Media) => {
+    const projectId = useProjectIdentifier();
+
+    return useQuery({
+        queryKey: getAnnotationsQueryKey(projectId, media),
```
dwesolow: np. we can use `annotationsQueryOptions`

application/ui/src/features/dataset/media-preview/annotator.component.tsx
```diff
@@ -37,10 +44,31 @@ const Annotator = ({
     onClose,
 }: AnnotatorProps) => {
     const getNextMediaItem = useNextMedia(mediaItem, items);
```
dwesolow: IIRC this function is not memoized so will cause unnecessary rerenders in the useEffect in line 53

application/ui/src/features/project/task-type-guards.ts
```diff
@@ -14,3 +14,7 @@ export const isDetectionTask = (taskType: TaskType | null): boolean => {
 export const isSegmentationTask = (taskType: TaskType | null): boolean => {
     return taskType === 'instance_segmentation';
 };
+
+export const isPrefetchEnabledForTask = (taskType: TaskType | null): boolean => {
```
dwesolow, line 18: why don't we want to do that for classification?

jpggvilaca: because we dont have tools on classification projects.

application/ui/src/features/annotator/tools/tool-manager.component.tsx
jpggvilaca: Because we dont want to do anything with webworkers if we're on classification tasks

application/ui/src/features/dataset/media-preview/annotator.component.tsx
jpggvilaca: Good catch. Wrapped in useCallback!

application/ui/src/features/dataset/media-preview/api/use-annotations-query.ts
jpggvilaca: updated

application/ui/src/features/annotator/tools/segment-anything-tool/use-segment-anything.hook.ts
jpggvilaca: Indeed, agree. removed

application/ui/src/features/dataset/media-preview/api/use-annotations-query.ts
```diff
@@ -87,7 +87,6 @@ export const useAnnotationsQuery = (media: Media) => {
     const projectId = useProjectIdentifier();
 
     return useQuery({
-        queryKey: getAnnotationsQueryKey(projectId, media),
-        queryFn: () => annotationsQueryFn(projectId, media),
+        ...annotationsQueryOptions(projectId, media),
```
dwesolow: np. we don't need to spread, just pass `annotationsQueryOptions(projectId, media)`

application/ui/src/features/annotator/tools/segment-anything-tool/use-segment-anything.hook.ts
```diff
@@ -1,82 +1,124 @@
...
-    return modelRef.current;
+        const nextImage = await queryClient.ensureQueryData(loadImageQueryOptions(projectId, nextItem));
+        const encoderModel = await queryClient.ensureQueryData(
+            segmentAnythingWorkerQueryOptions('SEGMENT_ANYTHING_ENCODER')
+        );
+
+        queryClient.prefetchQuery(segmentAnythingEncodingQueryOptions(nextItem, encoderModel, nextImage));
+    };
+
+    prefetch();
```
camiloHimura: Could the code just be executed directly without this additional function?

application/ui/src/features/annotator/tools/tool-manager.component.tsx
```diff
@@ -1,14 +1,26 @@
 // Copyright (C) 2025 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 
+import { useProject } from 'hooks/api/project.hook';
```
camiloHimura, line 4: use relative paths

application/ui/src/features/annotator/tools/tool-manager.component.tsx
```diff
@@ -1,14 +1,26 @@
...
 export const ToolManager = () => {
     const { activeTool } = useTool();
+    const { data: project } = useProject();
+
+    const isPreloadEnabled = project !== undefined && isPrefetchEnabledForTask(project.task.task_type);
+
+    // Preload SAM workers when the tool manager is mounted, so that the tool is ready
+    // to use as soon as the user selects it. Not a huge performance gain but
+    // it helps a bit.
+    usePreloadSAMWorkers(isPreloadEnabled);
```
camiloHimura, line 23: Let’s wrap this into a hook with a more descriptive name, like `usePreloadSamForClassification` (just as an example).

application/ui/src/features/dataset/media-preview/api/use-annotations-query.ts
```diff
@@ -15,64 +15,76 @@ const isUnannotatedError = (error: unknown): boolean => {
...
         : undefined;
+};
+
+export const annotationsQueryOptions = (projectId: string, media: Media) =>
+    queryOptions({
+        queryKey: getAnnotationsQueryKey(projectId, media),
+        queryFn: () => annotationsQueryFn(projectId, media),
+    });
+
+export const getAnnotationsQueryKey = (projectId: string, media: Media): QueryKey => {
```
camiloHimura: I don’t think we need to export this anymore. In any case, if we need access to the query key, we can just call `annotationsQueryOptions().queryKey`.

application/ui/src/features/dataset/media-preview/secondary-toolbar/util.ts
```diff
@@ -49,13 +51,26 @@ export const getNextMediaItem = (currentMediaItem: Media, allMediaItems: Media[]
...
+    if (currentIndex >= allMediaItems.length - 1) {
+        return undefined;
+    }
+
+    return allMediaItems[currentIndex + 1];
+};
+
+export const isSameMediaItem = (firstItem: Media, secondItem: Media): boolean => {
+    if (isVideoFrame(firstItem) && isVideoFrame(secondItem)) {
+        return firstItem.id === secondItem.id && firstItem.frame_number === secondItem.frame_number;
```
camiloHimura: what about to use `isEqual(firstItem, secondItem)`?

application/ui/src/features/dataset/media-preview/annotator.component.tsx
```diff
@@ -1,12 +1,18 @@
-// Copyright (C) 2025-2026 Intel Corporation
+// Copyright (C) 2025 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 
+import { useEffect } from 'react';
+
 import { View } from '@geti/ui';
+import { useQueryClient } from '@tanstack/react-query';
+import { useProjectIdentifier } from 'hooks/use-project-identifier.hook';
```
camiloHimura: relative paths

application/ui/src/features/dataset/media-preview/annotator.component.tsx
```diff
@@ -36,11 +43,35 @@ const Annotator = ({
...
+
+    useEffect(() => {
+        prefetchNextMediaAndAnnotationsData({
+            queryClient,
+            projectId,
+            getNextMediaItem,
+        });
+
+        if (isSAMContext) {
+            prefetchNextSAMEncodingData({
```
dwesolow: I think this should be inside `useSegmentAnythingModel`, it has the same responsibility as `useEncodingQuery`, could we just call `useEncodingQuery(nextMediaItem, nextImage, ...)`? The logic is the same and it would simplify a lot.

application/ui/src/features/annotator/tools/tool-manager.component.tsx
dwesolow: `useProject` is a suspense query, i.e.`project` should we always defined

application/ui/src/features/dataset/media-preview/secondary-toolbar/util.ts
```diff
@@ -49,13 +51,26 @@ export const getNextMediaItem = (currentMediaItem: Media, allMediaItems: Media[]
...
 
-    const nextIndex = Math.min(currentIndex + 1, allMediaItems.length - 1);
+    if (currentIndex >= allMediaItems.length - 1) {
+        return undefined;
+    }
+
+    return allMediaItems[currentIndex + 1];
+};
+
+export const isSameMediaItem = (firstItem: Media, secondItem: Media): boolean => {
```
dwesolow: Is it used anywhere?

application/ui/src/features/dataset/media-preview/utils.ts
```diff
@@ -10,3 +14,34 @@ export const getInitialAnnotations = (isUserReviewed: boolean, annotationsDTO: A
...
 };
+
+// When the user navigates to next media, image data and annotations will be already in React Query cache,
+// so the UI will feel smoother
+// whenever the user switches image. Unless he/she changes to a random or item. We could also consider
+// those cases but I feel like it's overkill. Let's see how this improvement performs and then we can iterate on it.
+//
+// ensureQueryData will get the data from cache if it's there, or call the queryFn and cache the result if it's not.
+// prefetchQuery will fetch the data and cache it
+export const prefetchNextMediaAndAnnotationsData = ({
```
dwesolow: what about converting this to hook `usePrefetchNextMediaAndAnnotationsData` that contains everything that is needed to handle that logic (useEffect etc)?

application/ui/src/features/dataset/media-preview/api/use-annotations-query.ts
jpggvilaca: Updated

application/ui/src/features/annotator/tools/tool-manager.component.tsx
jpggvilaca: why? I see that we are not consistent but we added an alias so i thought we wanted to use it
Screenshot 2026-03-04 at 10 42 11
Screenshot 2026-03-04 at 10 42 03

application/ui/src/features/annotator/tools/segment-anything-tool/use-segment-anything.hook.ts
jpggvilaca: This is because the function is async and is executed inside an useEffect.

application/ui/src/features/annotator/tools/tool-manager.component.tsx
jpggvilaca: The idea was to make this versatile. This controls the preloading, which knows nothing about tasks, but then the consumer decides when to enable this. Maybe based on task, or by something else entirely.

application/ui/src/features/dataset/media-preview/api/use-annotations-query.ts
jpggvilaca: Removed `exported`

application/ui/src/features/dataset/media-preview/secondary-toolbar/util.ts
jpggvilaca: sounds good, updated : )

application/ui/src/features/dataset/media-preview/annotator.component.tsx
jpggvilaca: same as above

application/ui/src/features/dataset/media-preview/secondary-toolbar/util.ts
jpggvilaca: Not anymore after the 50 million commits, updated ^^

application/ui/src/features/annotator/tools/tool-manager.component.tsx
jpggvilaca: Updated

application/ui/src/features/dataset/media-preview/utils.ts
jpggvilaca: Agree. Updated (last commit)

application/ui/src/features/dataset/media-preview/annotator.component.tsx
jpggvilaca: Agree. Updated (last commit)

application/ui/src/features/annotator/tools/tool-manager.component.tsx
camiloHimura: Because the rule in this repository was not to use it. If we’re allowing it now, I’m fine with that as well

camiloHimura (dismissed):

application/ui/src/features/annotator/tools/tool-manager.component.tsx
dwesolow: I always vote for absolute paths.

ActiveChooN (dismissed):

application/ui/src/features/annotator/tools/tool-manager.component.tsx
jpggvilaca: Discussed offline

application/ui/src/features/annotator/tools/segment-anything-tool/segment-anything-tool.component.tsx
```diff
@@ -58,10 +60,12 @@ export const SegmentAnythingTool = () => {
     const ref = useRef<SVGSVGElement>(null);
 
     const zoom = useZoom();
-    const { roi, image } = useSelectedMediaItem();
+    const { roi, image, mediaItem } = useSelectedMediaItem();
+    const { items } = useGetDatasetMediaItems();
+    const { nextMediaItem, nextImage, isNextImageReady } = useNextMediaPrefetch(mediaItem, items);
```
dwesolow: I thought that we want to prefetch next media always, no matter the tool. I see that we need this to be inside the sam, but IMO we should place that in the annotator as well (e.g. when we use polygon tool and we change media, it'd be nice if we have img straight away)

jpggvilaca: You're right. I had it the other way around. Moved prefetch to annotator from SAM

ActiveChooN (approved):

dwesolow (approved):

PR #5663, opened 2026-03-02
Fix include_unannotated handling in export filters

### LINKED ISSUE #5564: Fix include_unannotated handling in export filters
Current implementation of `include_unannotated` includes items that do not have any labels assigned.

This filtering setting has nothing to do with `DatasetItemAnnotationStatus.UNANNOTATED` status, but should be processed in combination with labels (`DatasetItemAnnotationStatus.REVIEWED`).

**Example:**
Project with 3 labels: "cat", "dog", "horse".
Image A annotated with labels "cat", "horse".
Image B annotated with label "dog"
Image C annotated with label "horse"
 
_Scenario 1:_ I'm not interested in horses, I want to create a dataset of only cats and dogs: I will export with `labels=["cat", "dog"]` and `include_unannotated=false`. Expected output:
Image A with label "cat"
Image B with label "dog"
 
_Scenario 2:_ I want to focus exclusively on canine species; some of the images also contain wolves, which aren't labeled yet. I will export with `labels=["dog"]` and `include_unannotated=true`. Expected output:
Image A without labels
Image B with label "dog"
Image C without labels

**Acceptance Criterias:**
- Reimplement the processing using datumaro API so it's unified for project datasets and training snapshots.
- Update BDD tests

itallix:
### Summary

- [x] Use new datumaro `Dataset.filter_by_labels` API in export filtering
- [x] Fix unit tests
- [x] Refactor bdd tests to use tables to setup image distribution
- [x] Support `include_unannotated` in import

### Checklist

### DIFF
--- a/application/backend/app/execution/dataset_export/export.py
+++ b/application/backend/app/execution/dataset_export/export.py
@@ -52,7 +52,7 @@ class ExportDataset(Execution[ExportDatasetJobParams]):
         self._db_session_factory = db_session_factory
 
     @step("Prepare dataset for export", 20)
-    def prepare_dataset(self, export_params: ExportDatasetJobParams) -> tuple[UUID, Dataset | None]:
+    def prepare_dataset(self, export_params: ExportDatasetJobParams) -> tuple[UUID, Dataset]:
         with self._db_session_factory() as session:
             if export_params.dataset_id is None:
                 self._dataset_service.set_db_session(session)
@@ -103,12 +103,8 @@ class ExportDataset(Execution[ExportDatasetJobParams]):
 
     def execute(self, params: ExportDatasetJobParams) -> None:
         dataset_id, dataset = self.prepare_dataset(params)
-        if not dataset:
-            logger.warning(
-                "Dataset {} for project {} is empty after applying filters. Nothing to export.",
-                dataset_id,
-                params.project_id,
-            )
+        if len(dataset) == 0:
+            self.pin_message("Dataset is empty after applying filters. Nothing to export.")
             return
         self.update_metadata({"dataset_id": dataset_id})
         self.export_dataset(dataset_id, dataset, params.export_format)
--- a/application/backend/app/execution/dataset_export/export.py
+++ b/application/backend/app/execution/dataset_export/export.py
@@ -56,24 +56,22 @@ class ExportDataset(Execution[ExportDatasetJobParams]):
         with self._db_session_factory() as session:
             if export_params.dataset_id is None:
                 self._dataset_service.set_db_session(session)
-                annotation_status = (
-                    DatasetItemAnnotationStatus.REVIEWED_OR_UNANNOTATED
-                    if export_params.include_unannotated
-                    else DatasetItemAnnotationStatus.REVIEWED
-                )
                 dataset = self._dataset_service.get_dm_dataset(
                     project_id=export_params.project_id,
                     task=export_params.task,
-                    annotation_status=annotation_status,
-                    label_names=export_params.labels,
+                    annotation_status=DatasetItemAnnotationStatus.REVIEWED,
                 )
             else:
                 self._dataset_revision_service.set_db_session(session)
                 dataset = self._dataset_revision_service.load_revision(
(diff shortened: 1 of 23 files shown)


jpggvilaca (dismissed):
UI lgtm

dwesolow (dismissed):

application/backend/app/execution/dataset_export/export.py
leoll2: Is `if not dataset` the proper way to check if the dataset (a `Dataset` object) is empty? Shouldn't it be `if len(dataset) == 0`?

application/backend/app/execution/dataset_import/import_to_project.py
```diff
@@ -105,21 +105,25 @@ def create_items(self, dataset: Dataset, params: ImportDatasetToProjectJobParams
                 label_categories=cast(LabelCategories, label_attr.categories),
                 label_mapping=params.labels_mapping,
             )
-            logger.info("Found {} labels for project {}", (label.name for label in labels), params.project_id)
+            logger.info("Found {} labels for project {}", [label.name for label in labels], params.project_id)
             size, min_p, max_p = len(dataset), 10, 100
             progress_interval = max(1, size // self.BATCH_PROGRESS_INTERVAL)
+            total = 0
```
leoll2: I suggest to rename the following variables:
- `size` -> `unfiltered_dataset_size`
- `total` -> `num_imported_media`

application/backend/app/execution/dataset_import/import_to_project.py
```diff
@@ -128,8 +132,17 @@ def create_items(self, dataset: Dataset, params: ImportDatasetToProjectJobParams
                     annotations=annotations,
                     subset=DatasetItemSubset(item.subset.name.lower()),
                 )
+                total += 1
                 if (idx > 0 and idx % progress_interval == 0) or idx == size - 1:
                     self.update_progress(min_p + ((idx + 1) / size) * (max_p - min_p))
+            if total == 0:
+                logger.warning(
+                    "No items were imported from the dataset to project {}. "
+                    "This may be due to filtering options that excluded all items.",
```
leoll2: In addition to the warning, can you also update the job message so that this important information gets also displayed in the UI?

application/backend/app/execution/dataset_export/export.py
itallix: It will trigger magic method `__len__`. But let's be more explicit.

application/backend/app/execution/dataset_import/import_to_project.py
itallix: Currently, there is no mechanism to keep the message if the step completed successfully. When a step finishes, the message is overridden with: `Completed: {step}`. Some options for how we can keep the message:

1. Fail the step.
2. Introduce a new state to indicate that the step has not failed, but completed with some warnings.
3. Use a different job attribute, e.g.: `result: dict | None = None  # result of the job, if completed.`

I like the 3rd approach - we can surface import statistics for both use cases. 
Let me know what you think.

itallix: Option 4 could be to introduce a flag at the decorator level, e.g.: `self._pinned_message: bool = False`. A combination of 3 and 4 seems the most convenient - we won't override the message with `Completed: {step}` in case of a warning, and will still provide import statistics in the `result` field when something was imported.

leoll2: Option 1 could work in this case, but it doesn't generalize well for other situations where a job can be considered successful but with some remark for the user, e.g. "A model was trained but its accuracy is below 10%".

Option 2 adds a new state to our model (state machine), adding extra complexity in the `JobController`; I would avoid this option.

Option 3 seems the most flexible. Can you implement it without API changes, e.g. exposing `Job.result.get("message")` through `JobView.message` for completed jobs?

itallix: I will proceed with pinning (Option 4).

leoll2 (dismissed):

jpggvilaca (approved):

leoll2 (approved):

PR #5681, opened 2026-03-04
chore: Display video timeline in collapsed view

dwesolow:
### Summary

Collapsed:
image

Expanded:
image

### Checklist

### DIFF
--- a/application/ui/src/features/annotator/video-player/video-toolbar/video-annotator/video-timeline/video-player-slider/video-player-slider.component.tsx
+++ b/application/ui/src/features/annotator/video-player/video-toolbar/video-annotator/video-timeline/video-player-slider/video-player-slider.component.tsx
@@ -18,7 +18,7 @@ type VideoPlayerSliderProps = {
     videoFrame: MediaVideoFrame;
     step: number;
     frameNumber: number;
-    sizePerSquare: number;
+    sizePerSquare?: number;
     frameOffset?: number;
     ref?: RefObject<HTMLDivElement | null>;
     selectFrame: (frameNumber: number) => void;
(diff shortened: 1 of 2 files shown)


jpggvilaca:
the "Frames" Text should be removed from the collapsed view. We can leave the space (empty div if we're using grid) but get rid of the word. Secondly we dont support sound nor does it matter so we could get rid of audio stuff entirely. As for the rest of the code, LGTM!

dwesolow:
> the "Frames" Text should be removed from the collapsed view. We can leave the space (empty div if we're using grid) but get rid of the word. Secondly we dont support sound nor does it matter so we could get rid of audio stuff entirely. As for the rest of the code, LGTM!

"Frames" are also present in the geti classic:
image
Should we keep or remove "Frames" then?

I think it makes sense having the sound. Imagine a video that user would like to annotate based on the weird sound, e.g. of the alarm on the street (video street recording) and they would like to annotate anything related to that.

jpggvilaca:
> > the "Frames" Text should be removed from the collapsed view. We can leave the space (empty div if we're using grid) but get rid of the word. Secondly we dont support sound nor does it matter so we could get rid of audio stuff entirely. As for the rest of the code, LGTM!
> 
> "Frames" are also present in the geti classic: image Should we keep or remove "Frames" then?
> 
> I think it makes sense having the sound. Imagine a video that user would like to annotate based on the weird sound, e.g. of the alarm on the street (video street recording) and they would like to annotate anything related to that.

I think we can get rid of it. I means nothing on collapsed mode. And it wouldnt cause layout shift if we keep the columns width. As for sound, not blocking but for me makes no sense. It really means nothing in the context of annotating VISUAL input. But we can ask on the UX channel

dwesolow:
> > > the "Frames" Text should be removed from the collapsed view. We can leave the space (empty div if we're using grid) but get rid of the word. Secondly we dont support sound nor does it matter so we could get rid of audio stuff entirely. As for the rest of the code, LGTM!
> > 
> > 
> > "Frames" are also present in the geti classic: image Should we keep or remove "Frames" then?
> > I think it makes sense having the sound. Imagine a video that user would like to annotate based on the weird sound, e.g. of the alarm on the street (video street recording) and they would like to annotate anything related to that.
> 
> I think we can get rid of it. I means nothing on collapsed mode. And it wouldnt cause layout shift if we keep the columns width. As for sound, not blocking but for me makes no sense. It really means nothing in the context of annotating VISUAL input. But we can ask on the UX channel

Removed "Frames", layout shift is not noticable.

jpggvilaca (approved):

PR #5682, opened 2026-03-04
Update sink rate limit fields

jpggvilaca:
### Summary

* Fix last bug of https://github.com/open-edge-platform/training_extensions/issues/5532
* Add rate limit fields as requested

Screenshot 2026-03-04 at 15 23 19
Screenshot 2026-03-04 at 15 41 03
Screenshot 2026-03-04 at 15 41 08
Screenshot 2026-03-04 at 15 41 11
Screenshot 2026-03-04 at 15 42 45

### Checklist

### DIFF
--- a/application/ui/src/features/inference/sinks/rate-limit/rate-limit-fields.component.tsx
+++ b/application/ui/src/features/inference/sinks/rate-limit/rate-limit-fields.component.tsx
@@ -1,0 +1,23 @@
+// Copyright (C) 2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+import { Flex, NumberField, Text } from '@geti/ui';
+
+import { positiveNumberOrUndefined } from '../utils';
+
+type RateLimitFieldsProps = {
+    rateLimit: number | null | undefined;
+};
+
+export const RateLimitFields = ({ rateLimit }: RateLimitFieldsProps) => {
+    const samples = positiveNumberOrUndefined(rateLimit);
+    const seconds = samples !== undefined ? 1 : undefined;
+
+    return (
+        <Flex direction='row' gap='size-100' alignItems={'end'}>
+            <NumberField label='Samples' name='rate_limit_samples' minValue={1} step={1} defaultValue={samples} />
+            <Text>every</Text>
+            <NumberField label='Seconds' name='rate_limit_seconds' minValue={0.1} step={0.1} defaultValue={seconds} />
+        </Flex>
+    );
+};
--- a/application/ui/src/features/inference/sinks/rate-limit/rate-limit-fields.component.tsx
+++ b/application/ui/src/features/inference/sinks/rate-limit/rate-limit-fields.component.tsx
@@ -15,7 +15,7 @@ export const RateLimitFields = ({ rateLimit }: RateLimitFieldsProps) => {
 
     return (
         <Flex direction='row' gap='size-100' alignItems={'end'}>
-            <NumberField label='Samples' name='rate_limit_samples' minValue={1} step={1} defaultValue={samples} />
+            <NumberField label='Samples' name='rate_limit_samples' minValue={0.1} step={0.1} defaultValue={samples} />
             <Text>every</Text>
             <NumberField label='Seconds' name='rate_limit_seconds' minValue={0.1} step={0.1} defaultValue={seconds} />
         </Flex>
--- a/application/ui/src/features/inference/sinks/utils.test.ts
+++ b/application/ui/src/features/inference/sinks/utils.test.ts
@@ -1,7 +1,7 @@
 // Copyright (C) 2025 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 
-import { getObjectFromFormData } from './utils';
+import { formatRateLimit, getObjectFromFormData, rateLimitFromFormData } from './utils';
 
 describe('getObjectFromFormData', () => {
     it('return an object mapping keys to values', () => {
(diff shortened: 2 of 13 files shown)


application/ui/src/features/inference/sinks/utils.test.ts
```diff
@@ -46,3 +46,83 @@ describe('getObjectFromFormData', () => {
...
+        expect(formatRateLimit(null)).toBe('Not set');
+        expect(formatRateLimit(0)).toBe('Not set');
+    });
+
+    it('formats singular sample and second labels', () => {
+        expect(formatRateLimit(1)).toBe('1 sample every 1 second');
+    });
+
+    it('formats plural samples for rates above one', () => {
+        expect(formatRateLimit(2)).toBe('2 samples every 1 second');
```
dwesolow, line 135: what about `1 sample every 0.5 second`? it will also rate limit 2.

application/ui/src/features/inference/sinks/utils.ts
```diff
@@ -48,3 +74,45 @@ export const getObjectFromFormData = (keys: FormDataEntryValue[], values: FormDa
...
+
+    if (samplesValue === null || secondsValue === null) {
+        return null;
+    }
+
+    const samplesRaw = toStringAndTrim(samplesValue);
+    const secondsRaw = toStringAndTrim(secondsValue);
+
+    const samples = parseFiniteNumber(samplesRaw);
+    const seconds = parseFiniteNumber(secondsRaw);
```
dwesolow: why do we need these utils? I thought that the value is a string that includes a number so only `parseFloat` would suffice, no?

application/ui/src/features/inference/sinks/utils.ts
```diff
@@ -48,3 +74,45 @@ export const getObjectFromFormData = (keys: FormDataEntryValue[], values: FormDa
...
+        const normalizedSeconds = toDisplayNumber(seconds);
+        const secondsLabel = normalizedSeconds === '1' ? 'second' : 'seconds';
+
+        return `1 sample every ${normalizedSeconds} ${secondsLabel}`;
+    }
+
+    const normalizedSamples = toDisplayNumber(normalizedRateLimit);
+    const sampleLabel = normalizedSamples === '1' ? 'sample' : 'samples';
+
+    return `${normalizedSamples} ${sampleLabel} every 1 second`;
```
dwesolow, line 92: I think it won't show correct initial values, e.g `0.5` can be  1/2 or 5/10 etc. Could we store both values on the server side and then server can do `rate_limit = rate_sample / rate_seconds`?

application/ui/src/features/inference/sinks/utils.test.ts
jpggvilaca: Correct, makes sense?

application/ui/src/features/inference/sinks/utils.ts
jpggvilaca: But the string can be very weird like `1.2,3` or use just commas or just spaces, or contain special characters.

application/ui/src/features/inference/sinks/utils.ts
jpggvilaca: I think asking to change the API will take longer. What is important here is the rate limit, not which values you used to get there. Makes sense?

dwesolow: maybe, but we will deceive the user, if they set 16 samples / 4 seconds and we will show 4 samples / 1 second, it might not be true?

application/ui/src/features/inference/sinks/utils.ts
dwesolow: Why it can be like this? It's number field so it should not happen?

application/ui/src/features/inference/sinks/utils.test.ts
dwesolow: I meant https://github.com/open-edge-platform/training_extensions/pull/5682#discussion_r2889338958, maybe we should ask BE team if we should care about this.

application/ui/src/features/inference/sinks/utils.ts
```diff
@@ -39,6 +39,32 @@ export enum WebhookHttpMethod {
...
+    }
+
+    const normalizedValue = compactValue.replace(',', '.');
+    const parsedValue = Number(normalizedValue);
+
+    return Number.isFinite(parsedValue) ? parsedValue : null;
+};
+
+const toDisplayNumber = (value: number) => {
+    const roundedValue = Number(value.toFixed(2));
```
camiloHimura: `Math.round` is more descriptive than using toFixed + Casting

application/ui/src/features/inference/sinks/utils.test.ts
```diff
@@ -46,3 +46,83 @@ describe('getObjectFromFormData', () => {
...
+    });
+
+    it('returns null for invalid mixed separators', () => {
+        const formData = new FormData();
+        formData.set('rate_limit_samples', '1,2.3');
+        formData.set('rate_limit_seconds', '1');
+
+        expect(rateLimitFromFormData(formData)).toBeNull();
+    });
+});
```
camiloHimura, line 120: It might need some test cases with empty strings and `NaN`. wdyt?

application/ui/src/features/inference/sinks/rate-limit/rate-limit-fields.component.tsx
```diff
@@ -0,0 +1,23 @@
...
+    rateLimit: number | null | undefined;
+};
+
+export const RateLimitFields = ({ rateLimit }: RateLimitFieldsProps) => {
+    const samples = positiveNumberOrUndefined(rateLimit);
+    const seconds = samples !== undefined ? 1 : undefined;
+
+    return (
+        <Flex direction='row' gap='size-100' alignItems={'end'}>
+            <NumberField label='Samples' name='rate_limit_samples' minValue={0.1} step={0.1} defaultValue={samples} />
```
camiloHimura, line 18: Should they be required?

application/ui/src/features/inference/sinks/utils.ts
jpggvilaca: We want the 2 decimals points, with round you wont get that. But i did remove the unnecessary `Number`

application/ui/src/features/inference/sinks/utils.ts
jpggvilaca: I see your point, but i would also argue that the user wouldnt do that. If you want half, why would you do higher numbers and combinations? you would always use the smallest combo possible. Anyway this is more an UX issue, ill ask on the next meeting

application/ui/src/features/inference/sinks/utils.ts
jpggvilaca: "should" yeah :D https://react-spectrum.adobe.com/v3/NumberField.html

application/ui/src/features/inference/sinks/rate-limit/rate-limit-fields.component.tsx
jpggvilaca: @leoll2 ^

application/ui/src/features/inference/sinks/utils.test.ts
jpggvilaca: Sounds good! added

application/ui/src/features/inference/sinks/utils.test.ts
jpggvilaca: @leoll2  ^

application/ui/src/features/inference/sinks/utils.ts
jpggvilaca: Actually scratch that. We dont need the decimals at all. Updated :)

application/ui/src/features/inference/sinks/utils.ts
leoll2: I agree it's pointless to change the API for this. 16/4 and 4/1 are equivalent, just normalized differently. An alternative approach is to use a single input field, fixing the other value to 1, as in `"1 frame every [ ] seconds"`. If so, however, we should add an explicit toggle button to enable/disable the rate limiter, because it's not obvious that leaving the field blank disables the limiter.

application/ui/src/features/inference/sinks/rate-limit/rate-limit-fields.component.tsx
jpggvilaca: `No, it is an optional setting. If rate_limit=null, the server doesn't enforce a limit`

application/ui/src/features/inference/sinks/utils.ts
jpggvilaca: Checked props and removed `parseFiniteNumber`, check last commit

dwesolow: ah, it's American notation, imo we should not care about this

application/ui/src/features/inference/sinks/utils.test.ts
leoll2: "2 samples every 1 second" and "1 sample every 0.5 second" are both acceptable for me

application/ui/tests/inference/inference.spec.ts
```diff
@@ -314,7 +314,8 @@ test('Inference', async ({ streamPage, page, network }) => {
         await page.getByRole('button', { name: 'Add new sink' }).click();
         await page.getByRole('button', { name: 'Folder' }).click();
         await page.locator('input[name="name"]').fill('New Folder');
-        await page.locator('input[aria-roledescription="Number field"]').fill('5');
+        await page.locator('input[aria-roledescription="Number field"]').first().fill('5');
```
dwesolow, line 317: np. we usually try to use `getByX` whenever we can, e.g. `page.getByRole('textbox', { name: 'Number field'} )` (didn't check if this work, just pseudo code)

application/ui/src/features/inference/sinks/utils.ts
dwesolow: @jpggvilaca do you want to add toggle button in this pr?

dwesolow (approved):

application/ui/src/features/inference/sinks/utils.ts
jpggvilaca: Discussed offline. No need for this

PR #5685, opened 2026-03-05
Training parameters integration

### LINKED ISSUE #5289: [API] Update model training parameters

jpggvilaca:
### Summary

* Integrate API on the final tab training parameters

Closes https://github.com/open-edge-platform/training_extensions/issues/5289

Screenshot 2026-03-05 at 09 22 45

### Checklist


application/ui/src/features/models/hooks/api/use-get-model-training-configuration.hook.ts
```diff
@@ -0,0 +1,19 @@
...
+export const useGetModelTrainingConfiguration = (modelId: string) => {
+    const projectId = useProjectIdentifier();
+
+    return $api.useQuery(
+        'get',
+        '/api/projects/{project_id}/models/{model_id}/training_configuration',
+        {
+            params: { path: { project_id: projectId, model_id: modelId } },
+        },
+        { enabled: Boolean(modelId) }
```
camiloHimura:
```suggestion
        { enabled: isNonEmptyString(modelId) }
```

camiloHimura (dismissed):

dwesolow:
IMO only the section whose content is too long should have scrollbar, not the whole container. WDYT?
image

application/ui/mocks/mock-training-configuration.ts
```diff
@@ -0,0 +1,126 @@
...
+                            {
+                                type: 'parameter',
+                                key: 'enable',
+                                name: 'Enable',
+                                description: '',
+                                value: true,
+                                default_value: true,
+                                value_type: 'bool',
+                                min_value: null,
+                                max_value: null,
```
dwesolow: that's weird, `min_value` and `max_value` should be only for ints/floats, not for `bool`.

application/ui/mocks/mock-training-configuration.ts
```diff
@@ -0,0 +1,126 @@
...
+                                allowed_values: null,
```
dwesolow: that one should be only for `enum`.

application/ui/src/features/models/hooks/api/use-get-model-training-configuration.hook.ts
dwesolow: I see `modelId` is always defined (query parameter), I believe we can remove this.

application/ui/src/features/models/model-listing/model-training-parameters/model-training-parameters.component.tsx
```diff
@@ -1,80 +1,64 @@
...
 
-                        <Text>Maximum epochs:</Text>
-                        <Text>200</Text>
+type ModelTrainingParametersProps = {
+    modelId: string;
+};
 
-                        <Text>Row 3:</Text>
-                        <Text>Value 3</Text>
+const buildColumnContent = (parameters: TrainingConfigurationParameter[] | undefined) => {
```
dwesolow: It's just component, we can name it `TrainingConfigurationParametersList` and do `<TrainingConfigurationParametersList parameters={...} />` in line 51 etc.

application/ui/src/features/models/model-listing/model-training-parameters/utils.ts
```diff
@@ -0,0 +1,101 @@
...
+
+export const isParameterGroup = (
+    parameter: TrainingConfigurationParameter
+): parameter is ConfigurableParameterGroup => {
+    return parameter.type === 'parameter_group';
+};
+
+export const findGroupByKey = (
+    parameters: TrainingConfigurationParameter[] | undefined,
+    key: string
```
dwesolow, line 18: do we want to type it to parameters' keys?

application/ui/src/features/models/model-listing/model-training-parameters/utils.ts
```diff
@@ -0,0 +1,101 @@
...
+ * Example output rows from this function:
+ * [
+ *   { name: "Data augmentation / Random affine / Rotation degrees", value: "10" },
+ *   { name: "Data augmentation / Random affine / Scaling ratio range", value: "0.5 - 1.5" }
+ * ]
+ *
+ * {ParentLabel} / {ParentLabel} / {ParameterName} {ParameterValue}
+ *
+ */
+export const flattenParameters = (
```
dwesolow, line 82: I don't think we should flatten, maybe it makes sense to display the parameters according to the data structure? We followed that approach in the geti classic:

Image
(nested parameter has indentation)

I'd do one improvement to geti classic implementation: if the parameter is not enabled (`off`) we don't list its children parameters (simply it does not make sense to show them).

application/ui/mocks/mock-training-configuration.ts
```diff
@@ -0,0 +1,126 @@
+// Copyright (C) 2025-2026 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+import { ConfigurableParameterGroup, TrainingConfigurationParameter } from '../src/constants/shared-types';
+
+export const getMockParameterTree = (): TrainingConfigurationParameter[] => {
```
dwesolow: np. `getMockedTrainingConfiguration`

application/ui/src/features/models/model-listing/model-training-parameters/model-training-parameters.component.tsx
```diff
@@ -1,80 +1,64 @@
...
+    const filteringGroup = findGroupByKey(datasetPreparationGroup?.parameters, 'filtering');
+    const augmentationGroup = findGroupByKey(datasetPreparationGroup?.parameters, 'augmentation');
 
-                        <Text>Param 4:</Text>
-                        <Text>Value 4</Text>
-                    </Grid>
-                }
+    return (
+        <Grid columns={['1fr', '1fr', '1fr']} gap={'size-200'}>
+            <Box
```
dwesolow, line 52: TBH I don't like the design we have for parameters, i.e. that for each parameter group we display a box. If we have more parameters that design won't scale well (ofc not a blocker, but thing to revisit)

application/ui/tests/models/model-details.spec.ts
```diff
@@ -299,4 +309,23 @@ test.describe('Model Details', () => {
...
+
+            const learningParametersHeading = page.getByRole('heading', { name: 'LEARNING PARAMETERS' });
+            const filtersHeading = page.getByRole('heading', { name: 'FILTERS' });
+            const augmentationsHeading = page.getByRole('heading', { name: 'AUGMENTATIONS' });
+
+            await expect(learningParametersHeading).toBeVisible();
+            await expect(filtersHeading).toBeVisible();
+            await expect(augmentationsHeading).toBeVisible();
+
+            await expect(learningParametersHeading.locator('..').getByText('Maximum epochs')).toBeVisible();
```
dwesolow: could we have a better selector?

application/ui/src/features/models/hooks/api/use-get-model-training-configuration.hook.ts
jpggvilaca: Removed altogether

jpggvilaca: We can yes, ty!

application/ui/src/features/models/model-listing/model-training-parameters/model-training-parameters.component.tsx
jpggvilaca: I like it! updated

application/ui/src/features/models/model-listing/model-training-parameters/utils.ts
jpggvilaca: Mixed feelings, indentation might not be obvious (we could increase padding but yeah) and not obvious to the average user. Some people (engineers from geti) even asked if the layout was broken. So I actually prefer this way, but maybe there's a 3rd and better option?

application/ui/src/features/models/model-listing/model-training-parameters/model-training-parameters.component.tsx
jpggvilaca: We only want these 3. If we want more then Mariusz needs to design something new

application/ui/tests/models/model-details.spec.ts
jpggvilaca: i checked and ('..') is the standard to get the parent selector. Here i checked the title of the container, then its parent, to access the value, but i can always (try to) add an aria-label

application/ui/src/features/models/model-listing/model-training-parameters/utils.ts
jpggvilaca: hmmm maybe overkill?

application/ui/mocks/mock-training-configuration.ts
jpggvilaca: ?

application/ui/mocks/mock-training-configuration.ts
jpggvilaca: It's in the spec. @leoll2  ^

leoll2: As Dawid correctly says, `min_value` and `max_value` only make sense if `value_type="int"` or `value_type="float"`. For other types, we do return the min/max because it's simpler to implement, but the value will be always `null` in this case. If that's a problem for the UI, I can refine the response to omit `min_value` / `max_value` when they are null.

jpggvilaca: Tbh I'd rather have values with null that not having them altogether.

application/ui/mocks/mock-training-configuration.ts
dwesolow: in geti classic `value_type` could be one of `bool | int | float | enum | array` , `allowed_values` was only for `enum` type.

application/ui/tests/models/model-details.spec.ts
dwesolow: yeah IMO `aria-label` or anything else would be better. Current approach might break if we add a wrapper div around the heading

jpggvilaca: Fair enough, added test-id

application/ui/mocks/mock-training-configuration.ts
jpggvilaca: ah i see

application/ui/src/features/models/model-listing/model-training-parameters/utils.ts
dwesolow: What about this: keep the flatten structure while displaying but don't display parameters "enable ..." parameters and don't display the rest of parameters from that group if the "enable ..." is off.?
Based on the screenshot above, we would not display anything ("enable ..." parameters are off).

application/ui/mocks/mock-training-configuration.ts
dwesolow: We should use discriminated union for each parameter type to provide strong type safety.

dwesolow (dismissed):

dwesolow (approved):

PR #5689, opened 2026-03-05
chore: Add video player to read only, display annotations only in annotation, display only predictions in prediction mode

dwesolow:
### Summary

This PR includes the following changes:
1. Add video player to read only mode.
2. Render video annotations while playing video only in annotation mode. Support for predictions will be added once backend supports that.
3. Render annotation segments for video only in annotation mode.
4. Render prediction segments for video only in prediction mode.
5. Move `AnnotatorMode` type to `shared/annotator`.

Each change has its own separate commit to make it easier to review.

image

image

### Checklist


application/ui/src/features/dataset/media-preview/read-only-annotator.component.tsx
```diff
@@ -34,6 +38,7 @@ type ReadOnlyAnnotatorProps = {
  * It uses the same gridArea structure as the normal annotator but with fewer elements.
  */
 export const ReadOnlyAnnotator = ({
+    mode,
```
ActiveChooN, line 41: qq: If I'm not mistaken, ReadOnlyAnnotator is only used in prediction mode. Why do we need to pass mode to it?

application/ui/src/features/dataset/media-preview/read-only-annotator.component.tsx
```diff
@@ -83,6 +85,12 @@ export const ReadOnlyAnnotator = ({
                 </AnnotatorCanvasSettings>
             </View>
 
+            {(isVideo(mediaItem) || isVideoFrame(mediaItem)) && (
+                <View gridArea={'video-toolbar'}>
+                    <VideoToolbar />
```
jpggvilaca, line 90: Subset gallery will never use this, so im wondering if we should couple this with the canvas, or make a new component with canvas + videotoolbar. It's not a big deal but it's a smell to me

application/ui/src/features/annotator/annotator-canvas/annotator-canvas.tsx
```diff
@@ -90,9 +95,10 @@ type AnnotatorCanvasProps = {
     mediaItem: Media;
     image: ImageData;
     isReadOnly?: boolean;
+    mode: AnnotatorMode;
 };
 
-export const AnnotatorCanvas = ({ mediaItem, image, isReadOnly = false }: AnnotatorCanvasProps) => {
+export const AnnotatorCanvas = ({ mode, mediaItem, image, isReadOnly = false }: AnnotatorCanvasProps) => {
```
jpggvilaca, line 101: looking at the project, we access `mode` on so many components that im wondering if we should move this to a provider and expose it there.

application/ui/src/features/annotator/video-player/video-toolbar/video-annotator/video-timeline/video-frame-segment/video-frame-segment.component.tsx
```diff
@@ -88,12 +88,101 @@ const useVideoTimelineQueries = ({ frameNumber }: { frameNumber: number }) => {
...
+
+const PredictionsLabelsSegments = ({ labels, colIndex, frameNumber }: PredictionsLabelsSegmentsProps) => {
+    const { predictedLabels, isPredictionLoading } = useVideoTimelinePredictions({ frameNumber });
+
+    return (
+        <LabelsSegments
+            labels={labels}
+            colIndex={colIndex}
+            frameNumber={frameNumber}
+            renderLabelSegment={(label) => (
```
jpggvilaca, line 147: why do we need this renderProp? The only thing i see different is the stripe, which can be controlled through a prop `isPrediction` and updating the styles, no?

application/ui/src/features/dataset/media-preview/read-only-annotator.component.tsx
dwesolow: It's used in prediction mode and in the training subsets (to display annotations in the img of the trained dataset).
@jpggvilaca this is the confusion I shared with you today.

application/ui/src/features/dataset/media-preview/read-only-annotator.component.tsx
dwesolow: Of we do `Annotator`, `PredictionAnnotator`, `SubsetAnnotator` (or other name) annotators. WDYT?

application/ui/src/features/annotator/annotator-canvas/annotator-canvas.tsx
dwesolow: In the subset datasets we won't have provider but we reuse the same components, that's why I decided to prop drill.

application/ui/src/features/annotator/video-player/video-toolbar/video-annotator/video-timeline/video-frame-segment/video-frame-segment.component.tsx
dwesolow: I just wanted to allow consumer to pass a label without `isPrediction`, `striped` and whatever else props might be there.

jpggvilaca: do we need this flexibility? I think for now we can just show stripe or no stripe and call it a day. If this gets reused in the future we improve it

application/ui/src/features/annotator/annotator-canvas/annotator-canvas.tsx
jpggvilaca: Discussed offline: annotator mode should prob live in a provider, tbd which one, and at which level

application/ui/src/features/dataset/media-preview/read-only-annotator.component.tsx
jpggvilaca: nah, it's all good

application/ui/src/features/annotator/video-player/video-toolbar/video-annotator/video-timeline/video-frame-segment/video-frame-segment.component.tsx
dwesolow: It was not because of the flexibility, IMPOV it was just a cleaner solution. Without render props pattern, we would need to pass add to the interface `isPrediction?: boolean; isLoading: boolean; label: Label`, with render prop we don't need. I can add those if you're leaning toward that solution, np.

jpggvilaca (approved):

PR #5698, opened 2026-03-05
Improve presentation model for configurable parameters

leoll2:
### Summary

Changes:
- Use a discriminator for the parameters based on their type (int, str, bool, ...) 
- Omit `min_value`/`max_value` attributes when not applicable - keep it only for int and float
- Omit `allowed_values` for bool parameters

### Checklist

### DIFF
--- a/application/backend/app/api/schemas/training_configuration.py
+++ b/application/backend/app/api/schemas/training_configuration.py
@@ -22,10 +22,6 @@ class _BaseConfigurableParameterView(BaseModel):
     key: str = Field(title="Key to identify the parameter")
     name: str = Field(title="User-friendly name of the parameter")
     description: str = Field(title="Extended description of the parameter", default="")
-    allowed_values: list | None = Field(
-        default=None,
-        title="List of allowed values for the parameter. None if it doesn't have a predefined set of valid values.",
-    )
 
 
 class BoolParameterView(_BaseConfigurableParameterView):
@@ -42,6 +38,10 @@ class StringParameterView(_BaseConfigurableParameterView):
     value_type: Literal["str"] = "str"
     value: str | None = Field(title="Actual value of the parameter")
     default_value: str | None = Field(title="Default value of the parameter")
+    allowed_values: list[str] | None = Field(
+        default=None,
+        title="List of allowed values for the parameter. None if it doesn't have a predefined set of valid values.",
+    )
 
 
 class IntParameterView(_BaseConfigurableParameterView):
@@ -52,6 +52,10 @@ class IntParameterView(_BaseConfigurableParameterView):
     default_value: int | None = Field(title="Default value of the parameter")
     min_value: int | float | None = Field(default=None, title="Minimum value for numeric parameters. None if unbounded")
     max_value: int | float | None = Field(default=None, title="Maximum value for numeric parameters. None if unbounded")
+    allowed_values: list[int] | None = Field(
+        default=None,
+        title="List of allowed values for the parameter. None if it doesn't have a predefined set of valid values.",
+    )
 
 
 class FloatParameterView(_BaseConfigurableParameterView):
@@ -62,6 +66,10 @@ class FloatParameterView(_BaseConfigurableParameterView):
     default_value: float | None = Field(title="Default value of the parameter")
     min_value: int | float | None = Field(default=None, title="Minimum value for numeric parameters. None if unbounded")
     max_value: int | float | None = Field(default=None, title="Maximum value for numeric parameters. None if unbounded")
+    allowed_values: list[float] | None = Field(
+        default=None,
+        title="List of allowed values for the parameter. None if it doesn't have a predefined set of valid values.",
+    )
 
 
 class FloatRangeParameterView(_BaseConfigurableParameterView):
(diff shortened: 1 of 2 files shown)


application/backend/app/api/schemas/training_configuration.py
```diff
@@ -15,34 +15,79 @@ class ConfigurableParameterViewElementType(StrEnum):
...
-    default_value: ParameterT | None = Field(title="Default value of the parameter")
-    value_type: Literal["bool", "str", "int", "float", "float_range"] = Field(title="Type of the parameter value")
-    min_value: int | float | None = Field(
-        default=None, title="Minimum value for numeric parameters. None if unbounded or not applicable"
-    )
-    max_value: int | float | None = Field(
-        default=None, title="Maximum value for numeric parameters. None if unbounded or not applicable"
-    )
-    allowed_values: list[ParameterT] | None = Field(
+    allowed_values: list | None = Field(
```
dwesolow: in geti classic it was for `enum` type of parameter, do we support `enum` here?

leoll2: Now enums are represented with `type: str` and a non-null `allowed_values`.

Example:
```json
{
    "key": "evaluation",
    "name": "Evaluation parameters",
    "description": "Configurable parameters related to the model evaluation.",
    "parameters": [
        {
            "type": "parameter",
            "key": "validation_metric",
            "name": "Validation metric",
            "description": "Metric used to evaluate model performance during validation.",
            "value": "default",
            "default_value": "default",
            "value_type": "str",
            "allowed_values": ["default", "accuracy", "map"],
        },
    ],
    "type": "parameter_group",
},
```

application/backend/app/api/schemas/training_configuration.py
```diff
@@ -15,34 +15,91 @@ class ConfigurableParameterViewElementType(StrEnum):
...
+
+class IntParameterView(_BaseConfigurableParameterView):
+    """Configurable integer parameter with optional min/max bounds."""
+
+    value_type: Literal["int"] = "int"
+    value: int | None = Field(title="Actual value of the parameter")
+    default_value: int | None = Field(title="Default value of the parameter")
+    min_value: int | float | None = Field(default=None, title="Minimum value for numeric parameters. None if unbounded")
+    max_value: int | float | None = Field(default=None, title="Maximum value for numeric parameters. None if unbounded")
+    allowed_values: list[int] | None = Field(
```
dwesolow, line 55: Could u share any parameter with `allowed_values` of type int or float?
My concern is that about presenting parameter. We used slider when `min_value` and `max_value` were defined, otherwise we used number field. For `allowed_values` we would need to provide 3rd way of presenting parameter that does not care about `min_value` and `max_value`. For me those parameters are conflicting.

leoll2: `input_size_width` is an int parameter with a set of allowed values. Note that `allowed_values` always takes precedence over `min_value` and `max_value` (not sure if there's a way to enforce that in the model). For parameter like `input_size_width`, a good visualization could be a dropdown list to select one value among the allowed values, as in Geti Classic.

```json
{
    "type": "parameter",
    "key": "input_size_width",
    "name": "Input size width",
    "description": (
        "Width size in pixels for model input images. "
        "Determines the horizontal resolution at which images are processed."
    ),
    "value": 992,
    "default_value": 992,
    "value_type": "int",
    "min_value": 0,
    "max_value": None,
    "allowed_values": [128, 256, 384, 512, 640, 800, 992, 1024],
},
```

image

dwesolow: I remember for this parameter we added custom presentation.
In geti classic it was just of `enum` type.
image

Could u share why did we decide not to use `enum` but to have `str` instead? I just don't see any benefit of the parameter which has `allowed_values` and `min/max_value`.

leoll2: > Could u share why did we decide not to use enum but to have str instead?

The new schema is more precise than the old one, as it carries information about the type of the enum in addition to the possible values. An enum-like field can be now identified by the presence of a non-null `allowed_values` field, while the `type` field indicates the type of the enum value.
Examples:
- Before:
  - `type: enum`, `allowed_values: [10, 20, 30]`
  - `type: enum`, `allowed_values: ["foo", "bar"]`
- Now:
  - `type: int`, `allowed_values: [10, 20, 30]`
  - `type: str`, `allowed_values: ["foo", "bar"]`

> I just don't see any benefit of the parameter which has allowed_values and min/max_value

As as I said before, `allowed_values` takes precedence over `min_value` and `max_value`. In other words, if `allowed_values` is not null, you can disregard the `min_value` and `max_value` attributes.

application/backend/app/api/schemas/training_configuration.py
```diff
@@ -15,34 +15,91 @@ class ConfigurableParameterViewElementType(StrEnum):
...
+        title="List of allowed values for the parameter. None if it doesn't have a predefined set of valid values.",
+    )
+
+
+class FloatParameterView(_BaseConfigurableParameterView):
+    """Configurable float parameter with optional min/max bounds."""
+
+    value_type: Literal["float"] = "float"
+    value: float | None = Field(title="Actual value of the parameter")
+    default_value: float | None = Field(title="Default value of the parameter")
```
A-Artemis, line 66: What was the choice for separating the numbers into int and float views? They are the same except for the `value`. 

Can the `IntParameterView` really have a float min and max?

A-Artemis (approved):

application/backend/app/api/schemas/training_configuration.py
leoll2: Distinguish between `int` and `float` can help the UI to properly display the parameter. For example, if an int parameter is rendered as a slider from 0 to 10, the steps should be discrete (you can move the slider from 3 to 4, but not to 3.5).

PR #5716, opened 2026-03-09
Prerequisite methods for inference service & endpoint

warrkan:
In order not to publish huge PR's, separating prerequisites for inference server as a separate PR.
Adding a few methods for video frames & refactoring frame extraction


application/backend/app/services/media_service.py
```diff
@@ -226,6 +226,12 @@ def get_media_by_id(self, project_id: UUID, media_id: UUID) -> Media:
             raise ResourceNotFoundError(ResourceType.MEDIA, str(media_id))
         return MediaAdapter.validate_python(db_media)
 
+    def get_media_by_ids(self, project_id: UUID, media_ids: list[UUID]) -> list[Media]:
+        """Get a media list by its IDs"""
+        repo = MediaRepository(project_id=str(project_id), db=self.db_session)
+        db_media_list = repo.get_by_ids([str(media_id) for media_id in media_ids])
+        return [MediaAdapter.validate_python(db_media) for db_media in db_media_list]
```
leoll2: What happens if one of the ids is not found? Method `get_media_by_id` handles this case by raising `ResourceNotFoundError`, while I don't see an equivalent error handling here in `get_media_by_ids`.

application/backend/app/services/media_service.py
```diff
@@ -329,7 +338,27 @@ def get_video_frame_by_video_id_and_index(self, project: Project, video_id: UUID
         """
         repo = MediaRepository(project_id=str(project.id), db=self.db_session)
         db_media = repo.get_video_frame_by_video_id_and_index(video_id=str(video_id), frame_index=frame_index)
-        return MediaAdapter.validate_python(db_media) if db_media else None
+        return VideoFrame.model_validate(db_media, from_attributes=True) if db_media else None
```
leoll2, line 347: Like in the other comment, probably you should handle the case of missing video id. Same in `get_video_frames_by_video_id_and_indexes`

application/backend/app/services/media_service.py
warrkan: Good point. Adjusted the logic

application/backend/app/services/media_service.py
warrkan: Here I've renamed method to `search_video_frames_by_video_id_and_indexes`, in my use case we need to test certain list of indexes, not necessarily all of them are annotated

leoll2 (approved):

PR #5723, opened 2026-03-09
Improve media upload part 1

jpggvilaca:
### Summary

* Move upload logic to a dedicated hook (it was coupled on the toolbar)
* Add tests to hook and component

Next PRs im gonna implement concurrency, cancellation and other nice things

### Checklist

### DIFF
--- a/application/ui/src/features/dataset/api/use-media-upload.ts
+++ b/application/ui/src/features/dataset/api/use-media-upload.ts
@@ -1,0 +1,58 @@
+// Copyright (C) 2026 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+import { toast } from '@geti/ui';
+import { useQueryClient } from '@tanstack/react-query';
+import { useProjectIdentifier } from 'hooks/use-project-identifier.hook';
+
+import { $api } from '../../../api/client';
+import { getQueryKey } from '../../../query-client/query-client';
+
+export const useMediaUpload = () => {
+    const projectId = useProjectIdentifier();
+    const queryClient = useQueryClient();
+
+    const addItemMutation = $api.useMutation('post', '/api/projects/{project_id}/dataset/media');
+
+    const uploadMedia = async (files: File[]) => {
+        const uploadPromises = files.map((file) => {
+            const formData = new FormData();
+            formData.append('file', file);
+
+            return addItemMutation.mutateAsync({
+                params: { path: { project_id: projectId } },
+                // eslint-disable-next-line @typescript-eslint/no-explicit-any
+                body: formData as any,
+            });
+        });
+
+        const promises = await Promise.allSettled(uploadPromises);
+
+        const succeeded = promises.filter((result) => result.status === 'fulfilled').length;
+        const failed = promises.filter((result) => result.status === 'rejected').length;
+
+        await queryClient.invalidateQueries({
+            queryKey: getQueryKey([
+                'get',
+                '/api/projects/{project_id}/dataset/media',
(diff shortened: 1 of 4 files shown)


application/ui/src/features/dataset/api/use-media-upload.ts
```diff
@@ -0,0 +1,58 @@
...
+                type: 'warning',
+                message: `Uploaded ${succeeded} item(s), ${failed} failed`,
+            });
+        }
+    };
+
+    return {
+        uploadMedia,
+        isUploading: addItemMutation.isPending,
+    };
```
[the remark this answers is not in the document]
jpggvilaca: This will be a non-issue on the next PR ((concurrency)

application/ui/src/features/dataset/api/use-media-upload.ts
```diff
@@ -0,0 +1,58 @@
...
+                '/api/projects/{project_id}/dataset/media',
+                { params: { path: { project_id: projectId } } },
+            ]),
+        });
+
+        if (failed === 0) {
+            toast({ type: 'success', message: `Uploaded ${succeeded} item(s)` });
+        } else if (succeeded === 0) {
+            toast({ type: 'error', message: `Failed to upload ${failed} item(s)` });
+        } else {
```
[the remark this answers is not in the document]
jpggvilaca: Will be addressed on the next PR

application/ui/src/features/dataset/gallery/toolbar/toolbar.component.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+}));
+
+vi.mock('../hooks/use-select-dataset-item.hook', () => ({
+    useSelectDatasetItem: () => ({
+        onSelectedMediaItemChange: onSelectedMediaItemChangeMock,
+        selectedMediaItem: null,
+    }),
+}));
+
+vi.mock('../../../models/train-model/train-model.component', () => ({
```
dwesolow, line 30: We should not do it, I think it hits that the code structure is not good for testing. What about moving `AddMediaButton` to separate component and only test that newly created component? I think it will simplify writing tests a lot.

jpggvilaca: The point is to not have to mock every single provider from these components. We should only be concerned with the toolbar. Testing the TrainModel should be (and it is) done in a different test

camiloHimura: That’s a tough one, I kind of agree with both sides

jpggvilaca: I dont like mocking to start with but imho, if we have Component A composed of B, C and D. We should test component A. The rest will have their own tests. Or else it's not a real unit tests, it's more of an integration test. Ofc we could maybe refactor and separate more making it easier to test or having less mocks, but we cannot afford to refactor everyday ^^

dwesolow: Yeah, that's exactly my point not to mock every single provider. Testing media upload is not related to any logic in toolbar at all, that's why I don't see a reason to mock train model and other components just to test media upload.

jpggvilaca: We are not testing upload here, we are testing the toolbar. I just added tests here because toolbar had no tests. The real upload tests are on the hook use-media-upload

dwesolow (approved):

PR #5724, opened 2026-03-09
Inference server

### LINKED ISSUE #5619: Implement inference server
Implement an `InferenceServer` component which acts as a lifecycle manager for the model used for on-demand inference, and as bridge between the inference API (prediction endpoint) and the model itself.

The class should have, more or less, the following methods:
- `get_status()` -> returns details about the server status, e.g. whether it is idle or it has a loaded model, info about the loaded model (if any), model TTL, ...
- `set_inference_model(model)` -> shuts down the currently loaded model (if any, and if different) and it loads the new one through Model API ([Model.create_model](https://github.com/open-edge-platform/model_api/blob/master/src/model_api/models/model.py#L171)).
- `infer_batch(media)` -> Loads the input data, executes batch inference through Model API ([`Model.infer_batch`](https://github.com/open-edge-platform/model_api/blob/fc24e51d68a92d1382791300f20369091d25fe32/src/model_api/models/model.py#L464)), and finally converts the results to our internal data structures
- `stop()` -> cancels any pending inference request and unloads the model, if any

In the first iteration, we only allow one model to be loaded at any time. The loaded model has a TTL (Time To Live) of configurable duration, e.g. 3 minutes by default: if the server doesn't receive any inference request for that model before the TTL expires, the model should be unloaded; an inference request resets the TTL of the model.

_Note:_ Model API also offers an [`AsyncPipeline`](src/model_api/pipelines/async_pipeline.py) class that could be potentially used instead of `Model`, but the latter seems for flexible and easy to adapt to our usecases.

warrkan:
Inference server component.

This component is responsible for:

1. Loading a model for inference 
2. Performing inference by delegating calls to a loaded model
3. Returning info about loaded model
4. Shutting down the loaded model

Additionally a monitor thread is introduced which monitors server calls and performs loaded model TTL count down.

Resolves #5619


application/backend/app/models/media.py
```diff
@@ -123,3 +123,23 @@ def video_id(self) -> UUID:
...
 
 MediaAdapter: TypeAdapter[Media] = TypeAdapter(Media)
+
+
+class VideoRange(BaseModel):
+    start_frame: Annotated[int, Field(description="Index of the first frame to predict, inclusive", ge=0)]
+    end_frame: Annotated[int, Field(description="Index of the last frame to predict, inclusive", ge=0)]
+    stride: Annotated[
+        int, Field(description="Predict every nth frame in the range, e.g. stride=1 means predict every frame", ge=1)
+    ]
```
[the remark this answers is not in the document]
warrkan: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/open-edge-platform/training_extensions/pull/5724#discussion_r2912097532)

application/backend/app/models/media.py
```diff
@@ -123,3 +123,29 @@ def video_id(self) -> UUID:
...
+
+    @model_validator(mode="after")
+    def validate_frame_range(self) -> "VideoRange":
+        if self.end_frame < self.start_frame:
+            raise ValueError(f"end_frame ({self.end_frame}) must be >= start_frame ({self.start_frame})")
+        return self
+
+
+class MediaPredictionRequest(BaseModel):
+    id: UUID
```
leoll2: I would rename this field to `media_id` for clarity, else it's not clear if it's an identifier for the media or for the request.

application/backend/app/models/media.py
```diff
@@ -123,3 +123,29 @@ def video_id(self) -> UUID:
...
+    range: VideoRange
```
leoll2: Shouldn't `range` be optional for images?

application/backend/app/services/inference/inference_server.py
```diff
@@ -0,0 +1,116 @@
...
+
+class InferenceStatus(StrEnum):
+    IDLE = "IDLE"
+    ACTIVE = "ACTIVE"
+
+
+@dataclass(frozen=True)
+class InferenceState:
+    status: InferenceStatus
+    model: InferenceModel | None = None
```
leoll2: These classes would better fit in the `app/models` imho. If some classes are only meant to be private for `InferenceServer`, you can prefix their name with `_`.

application/backend/app/services/inference/inference_server.py
```diff
@@ -0,0 +1,116 @@
...
+class InferenceServer:
+    def __init__(self, data_dir: Path) -> None:
+        self._projects_dir = data_dir / "projects"
+        self._loaded_model: LoadedModel | None = None
+
+    def _get_model_file_path(self, project_id: UUID, model_id: UUID, extension: str = "xml") -> Path:
+        file_path = self._projects_dir / f"{project_id}/models/{model_id}/model.{extension}"
+        if not file_path.is_file():
+            raise FileNotFoundError(f"Model file not found: {file_path}")
+        return file_path
```
leoll2: The `InferenceServer` should be ideally agnostic to the storage layout: it should rely on `ModelService` to get the model and its path. In this case, you can call [`ModelService.get_model_binary_files`](https://github.com/open-edge-platform/training_extensions/blob/7beb755fc55cb8355f38b3a829a4478ddcf9e0b3/application/backend/app/services/model_service.py#L373).

Note that model variants (OV, ONNX, Pytorch, ...) are now identified by ID; here you need to load the variant with `format=ModelFormat.OPENVINO` and `precision == ModelPrecision.FP16` (guaranteed to exist and unique, assuming the model files are not deleted).

application/backend/app/services/inference/inference_server.py
```diff
@@ -0,0 +1,116 @@
...
+class LoadedModel:
+    id: UUID
+    model: Model
+    device: str
+    ttl: int
+    load_timestamp: datetime
+
+
+@dataclass(frozen=True)
+class InferenceModel:
```
leoll2: Can you add docstrings to these classes? `InferenceModel` and `LoadedModel` largely overlap, I don't fully understand their difference.

application/backend/app/workers/inference_server_monitor.py
```diff
@@ -0,0 +1,89 @@
...
+from functools import wraps
+from multiprocessing.synchronize import Event
+
+from loguru import logger
+
+from app.services.inference import InferenceServer
+from app.workers.base import BaseThreadWorker
+
+
+class InferenceServerMonitorThread(BaseThreadWorker):
```
leoll2, line 16: Please add a docstring to clarify the responsibilities of `InferenceServerMonitorThread` with respect to `InferenceServer`.

If I understand correctly:
- `InferenceServer` is responsible for managing one ModelAPI model and performing batch inference with it. Notably, `InferenceServer` should be agnostic to TTL and any decision logic about when to start/stop/switch the model.
- `InferenceServerMonitorThread` orchestrates the `InferenceServer`, proxying the inference requests and managing the lifecycle of the inference model, including decisions on when to shutdown/replace it.

application/backend/app/services/inference/inference_server.py
```diff
@@ -0,0 +1,116 @@
...
+                model_id=self._loaded_model.id,
+                device=self._loaded_model.device,
+                ttl=self._loaded_model.ttl,
+                load_timestamp=self._loaded_model.load_timestamp,
+                remaining_seconds=self._loaded_model.ttl
+                - (datetime.now() - self._loaded_model.load_timestamp).total_seconds(),
+            ),
+        )
+
+    def infer_batch(self, labels: list[Label], inputs: list[BatchInferenceInput]) -> BatchInferenceResult:
```
leoll2, line 124: I think that `infer_batch`, `set_inference_model`, ... should acquire a lock to prevent race conditions where e.g. the model is replaced/shutdown while an inference request is in progress.

application/backend/app/workers/inference_server_monitor.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def __init__(self, server: InferenceServer, stop_event: Event) -> None:
+        super().__init__(stop_event=stop_event)
+
+        self._server = server
+        self._model_loaded = False
+        self._model_stopped = False
+        self._inference_requested = False
+        self._ttl = 0
+        self._ttl_start_time = -1.0
```
leoll2, line 35: This class maintains a complex state, partially duplicating `InferenceState`, `InferenceServer._loaded_model`, etc... I'm worried that such complexity/redundancy would cause race conditions and state inconsistencies.

For example, can't `_model_loaded` and `_model_stopped` be retrieved from the `_server` state?

application/backend/app/workers/inference_server_monitor.py
```diff
@@ -0,0 +1,89 @@
...
+    def setup(self) -> None:
+        logger.debug("Setting up inference server")
+
+        orig_set_inference_model = self._server.set_inference_model
+
+        @wraps(orig_set_inference_model)
+        def wrapped_set_inference_model(*args, **kwargs):
+            model_loaded = orig_set_inference_model(*args, **kwargs)
+            if model_loaded:
+                self._ttl = kwargs.get("ttl")  # pyrefly: ignore[bad-assignment]
```
leoll2: This pattern of wrapping with `*args` and `**kwargs` may create problems for type-checking, because I assume that Pyrefly can't verify that `set_inference_model` is called with the expected arguments. For example, code like `kwargs.get("ttl")` is brittle, it can easily break on refactoring of the `InferenceServer` interface. I'd suggest to update the wrapper signature to explicitly mirror the one original method.

application/backend/app/workers/inference_server_monitor.py
```diff
@@ -0,0 +1,89 @@
...
+
+            if self._model_stopped:
+                logger.debug("Model stopped, stopping TTL countdown")
+                self._model_stopped = False
+                self._ttl_start_time = -1.0
+
+            if self._inference_requested:
+                logger.debug("Batch inference requested, resetting TTL countdown")
+                self._inference_requested = False
+                self._ttl_start_time = time.perf_counter()
```
leoll2: Can't the TTL be refreshed synchronously wrt the request in `wrapped_infer_batch`?

application/backend/app/services/inference/inference_server.py
warrkan: Refactored, thanks for heads up!

application/backend/app/models/media.py
warrkan: Indeed, fixed

application/backend/app/workers/inference_server_monitor.py
warrkan: It does not duplicate. The only part which is duplicated is TTL. In monitor it's used for countdown, in server - for inference status, but we need it in both.

application/backend/app/workers/inference_server_monitor.py
warrkan: Good point, fixed

application/backend/app/workers/inference_server_monitor.py
warrkan: Didn't get this point, can you please elaborate a bit?

leoll2: In the current implementation, the TTL start time is refreshed here in `run_loop` following an inference request. Based on my understanding, this `run_loop` runs in the background and _asynchronously_ w.r.t. to inference requests (calls to `infer_batch`). My proposal is to do it _synchronously_ in `wrapped_infer_batch`, i.e. let that method update `_ttl_start_time` directly instead of `_inference_requested`. The advantage is that you don't need the `_inference_requested` variable state anymore, and it avoids race conditions like the following:
1. Client calls `inference_server.infer_batch` -> sets `_inference_requested`
2. Client calls `inference_server.stop` -> sets `_model_stopped`
3. The `run_loop` executes: at L87 it sets `_ttl_start_time` to -1.0, then it (incorrectly) updates the value again at L92.

Race conditions like this exist because most of the state updates are handled asynchronously. My suggestion is to use a synchronous approach when possible (model loaded/stopped, inference request); the only procedure that must be necessarily asynchronous is the TTL expiration check (L94:L99).

application/backend/app/workers/inference_server_monitor.py
leoll2: I'd argue that the `InferenceServer` status shouldn't include the TTL, since that is fully managed by the `InferenceServerMonitorThread`. I don't think there is a use case that requires exposing the TTL in the server status, but even if it were, it should retrieve this piece of information from `InferenceServerMonitorThread` instead.

application/backend/app/services/inference/inference_server.py
```diff
@@ -0,0 +1,122 @@
...
+
+MODELAPI_NSTREAMS = os.getenv("MODELAPI_NSTREAMS", "2")
+
+
+@dataclass(frozen=True)
+class _LoadedModel:
+    id: UUID
+    model: Model
+    device: str
+    ttl: int
```
leoll2: I think this can be removed, see my other comment https://github.com/open-edge-platform/training_extensions/pull/5724#discussion_r2923698147

application/backend/app/services/inference/inference_server.py
```diff
@@ -0,0 +1,122 @@
...
+    The server maintains the currently loaded model and its associated metadata, such as device and TTL,
+    to ensure efficient inference operations.
+    """
+
+    def __init__(self, model_service: ModelService) -> None:
+        self._model_service = model_service
+        self._loaded_model: _LoadedModel | None = None
+        self._lock = threading.Lock()
+
+    def set_inference_model(self, project_id: UUID, model_id: UUID, device: str, ttl: int) -> bool:
```
leoll2, line 44: Please add docstrings with 'Args' and 'Returns' sections

application/backend/app/services/inference/inference_server.py
```diff
@@ -0,0 +1,122 @@
...
+                nstreams=MODELAPI_NSTREAMS,
+            )
+
+            self._loaded_model = _LoadedModel(
+                id=model_id, model=model, device=device, ttl=ttl, load_timestamp=datetime.now()
+            )
+            return True
+
+    def get_status(self) -> InferenceState:
+        with self._lock:
```
leoll2: Since this method is read-only - it doesn't _write_ to the state - it can safely run without the lock. Just make sure to assign `self._loaded_model` to a local variable at the beginning, to guarantee that all properties like `id`, `device`, ... are read from the same Python object (i.e. the same model); without a lock, it could happen that `self._loaded_model` is updated concurrently by another method, but if you only read it once and cache the object in a variable you won't have race conditions.

application/backend/app/services/inference/inference_server.py
```diff
@@ -0,0 +1,122 @@
...
+                    load_timestamp=self._loaded_model.load_timestamp,
+                    remaining_seconds=self._loaded_model.ttl
+                    - (datetime.now() - self._loaded_model.load_timestamp).total_seconds(),
+                ),
+            )
+
+    def infer_batch(self, labels: list[Label], inputs: list[BatchInferenceInput]) -> BatchInferenceResult:
+        with self._lock:
+            if self._loaded_model is None:
+                raise RuntimeError("No model loaded for inference")
```
leoll2:
```suggestion
        if self._loaded_model is None:
            raise RuntimeError("No model loaded for inference")
        with self._lock:
            if self._loaded_model is None:
                raise RuntimeError("No model loaded for inference")
```
Double-checked locking, so that the method can quickly return (without waiting for the lock) when there's no loaded model. The 2nd check is not redundant, it's there to prevent a (rare) race condition where the model is unloaded while waiting for the lock.

application/backend/app/services/inference/inference_server.py
```diff
@@ -0,0 +1,122 @@
...
+                    device=self._loaded_model.device,
+                    ttl=self._loaded_model.ttl,
+                    load_timestamp=self._loaded_model.load_timestamp,
+                    remaining_seconds=self._loaded_model.ttl
+                    - (datetime.now() - self._loaded_model.load_timestamp).total_seconds(),
+                ),
+            )
+
+    def infer_batch(self, labels: list[Label], inputs: list[BatchInferenceInput]) -> BatchInferenceResult:
+        with self._lock:
```
leoll2: I suggest to add a timeout to this lock (e.g. 10s); if `set_inference_model` takes unexpectedly long to load the model, one or more `infer_batch` requests could remain stuck on this lock indefinitely. With the timeout, instead, the pending request can fail after 10s and the client can try it again. The 2nd request has a higher chance to succeed because the model may have finished loading in the meantime.

application/backend/app/services/inference/inference_server.py
```diff
@@ -0,0 +1,122 @@
...
+                    resource_type=ResourceType.MODEL,
+                    resource_id=f"{model_id} with format {format.value}",
+                )
+            model_xml_path, _ = paths
+
+            model = Model.create_model(
+                model=str(model_xml_path),
+                device=device,
+                nstreams=MODELAPI_NSTREAMS,
+            )
```
leoll2, line 96: This operation could take many seconds, especially for large models. To avoid holding the lock for a such long time, I suggest to add a new state `LOADING` to the inference server. The semantic of each state would be:
- `IDLE` -> no model is loaded or being loaded in the server; client may request to load a new model.
- `LOADING` -> the server is busy loading a model; requests to load a new model will be rejected (since `Model.create_model` can't be cancelled afaik) while inference requests may experience a delay or even timeout
- `ACTIVE` (maybe rename to `READY`?) -> a model is fully loaded in the server, ready to serve inference requests. Client may request to switch to another model, bringing the status back to `LOADING`.

A secondary but important advantage of this extended state model is that we can potentially show it in the UI, to inform the user about possible delays/timeouts in the inference while the server is loading the model.

warrkan: Added one more state for Loading

application/backend/app/services/inference/inference_server.py
warrkan: Good point, added a timeout

application/backend/app/workers/inference_server_monitor.py
warrkan: Ah, got it. Reworked this part a bit

application/backend/app/workers/inference_server_monitor.py
warrkan: Removed TTL from inference server itself

application/backend/app/services/inference/inference_server.py
```diff
@@ -0,0 +1,146 @@
...
+    load_timestamp: datetime
+
+
+class InferenceServer:
+    """
+    Inference Server manages the lifecycle of a loaded model for inference, including loading models,
+    tracking their status, and performing batch inference.
+    It interacts with the ModelService to retrieve model files and uses the Model API to run inference on input data.
+    The server maintains the currently loaded model and its associated metadata, such as device and TTL,
+    to ensure efficient inference operations.
```
leoll2, line 35:
```suggestion
    Inference Server manages the lifecycle of a loaded model for inference, including loading models,
    tracking their status, and performing batch inference.
    It interacts with the ModelService to retrieve model files and uses the Model API to run inference on input data.
```
Leftover reference to TTL in the last sentence; I think it can be removed altogether, the previous sentences already provide enough details.

application/backend/app/services/inference/inference_server.py
leoll2: Where is the timeout? As far as I can see, this lock is still acquired without timeout (L121, `with self._lock`)

application/backend/app/services/inference/inference_server.py
```diff
@@ -0,0 +1,146 @@
...
+            for idx, input in enumerate(inputs):
+                result.predictions.append(
+                    BatchInferencePrediction(
+                        media=BatchInferenceMedia(id=input.media_id, frame_index=input.frame_index),
+                        prediction=convert_prediction(
+                            labels=labels, frame_data=input_data[idx], prediction=inference_result[idx]
+                        ),
+                    )
+                )
+            return result
```
leoll2: This section could be moved outside of the lock scope

application/backend/app/services/inference/inference_server.py
```diff
@@ -0,0 +1,146 @@
...
+                project_id=project_id, model_id=model_id, model_variant_id=model_variant.id
+            )
+            if not files_exist:
+                raise ResourceNotFoundError(
+                    resource_type=ResourceType.MODEL,
+                    resource_id=f"{model_id} with format {format.value}",
+                )
+            model_xml_path, _ = paths
+
+            self._loading_model = True
```
leoll2: I'd move this to L46, i.e. right after acquiring the lock

application/backend/app/services/inference/inference_server.py
```diff
@@ -0,0 +1,146 @@
...
+            model = Model.create_model(
+                model=str(model_xml_path),
+                device=device,
+                nstreams=MODELAPI_NSTREAMS,
+            )
+            self._loading_model = False
```
leoll2: This should be moved to L83 in the `finally` clause, to make sure it's properly reset even if `Model.create_model` throws an exception.

application/backend/app/services/inference/inference_server.py
leoll2: This comment was marked as resolved but it isn't

application/backend/app/services/inference/inference_server.py
```diff
@@ -0,0 +1,146 @@
...
+        self._model_service = model_service
+        self._loaded_model: _LoadedModel | None = None
+        self._loading_model: bool = False
+        self._lock = threading.Lock()
+
+    def set_inference_model(self, project_id: UUID, model_id: UUID, device: str, ttl: int) -> bool:
+        self._lock.acquire(timeout=10)
+        try:
+            if self._loaded_model and self._loaded_model.id == model_id and self._loaded_model.device == device:
+                return False  # same model and device, no need to reload
```
leoll2, line 67:
```suggestion
    def set_inference_model(self, project_id: UUID, model_id: UUID, device: str, ttl: int) -> bool:
        # Optimization: if the given model is already loaded on the given device, return immediately
        m = self._loaded_model   # use a variable to avoid multiple unprotected reads of self._loaded_model
        if m and m.id == model_id and m.device == device:
            return False
        self._lock.acquire(timeout=10)
        try:
            if self._loaded_model and self._loaded_model.id == model_id and self._loaded_model.device == device:
                return False  # same model and device, no need to reload
```
Another optimization with double-checked locking

application/backend/app/services/inference/inference_server.py
warrkan: Indeed, done

application/backend/app/services/inference/inference_server.py
warrkan: Ah, sorry, added it to the wrong method. Fixed

application/backend/app/services/inference/inference_server.py
warrkan: Right. Fixed

leoll2 (approved):

PR #5726, opened 2026-03-09
Make submit logic less permissive

### LINKED ISSUE #5704: Submit annotation on last media removes annotations
Steps to reproduce:
1. Create project.
2. Upload dataset.
3. Annotate.
4. When last media is annotated and submitted, press submit again.

Expected result:
Ideally submit button should be disabled because nothing has changed (annotations are the same).

Current result:
Annotations are removed.

jpggvilaca:
### Summary

* Submit only if
1) Annotation mode: The canvas is "dirty", i.e. if the current annotations are different from the initial ones
2) Prediction mode: we have predictions and the user did not review them
Closes https://github.com/open-edge-platform/training_extensions/issues/5704

BONUS: Fix sporadic failure on train-model unit test (missing handlers)

### Checklist

### DIFF
--- a/application/ui/src/features/annotator/video-player/video-toolbar/video-controls.component.tsx
+++ b/application/ui/src/features/annotator/video-player/video-toolbar/video-controls.component.tsx
@@ -11,10 +11,6 @@ export const VideoControls = () => {
     const { isPlaying, play, pause, previousFrame, nextFrame, canSelectPreviousFrame, canSelectNextFrame } =
         videoControls;
 
-    const handlePlay = async () => {
-        await play();
-    };
-
     return (
         <Flex alignItems={'center'} gap={'size-100'}>
             <Flex alignItems={'center'}>
@@ -31,7 +27,7 @@ export const VideoControls = () => {
                         <Pause />
                     </ActionButton>
                 ) : (
-                    <ActionButton isQuiet aria-label={'Play video'} onPress={handlePlay}>
+                    <ActionButton isQuiet aria-label={'Play video'} onPress={play}>
                         <Play />
                     </ActionButton>
                 )}
--- a/application/ui/src/features/annotator/video-player/video-toolbar/video-controls.component.tsx
+++ b/application/ui/src/features/annotator/video-player/video-toolbar/video-controls.component.tsx
@@ -4,18 +4,15 @@
 import { ActionButton, Flex } from '@geti/ui';
 import { Pause, Play, SoundOff, SoundOn, StepBackward, StepForward } from '@geti/ui/icons';
 
-import { useAnnotationActions } from '../../../../shared/annotator/annotation-actions-provider.component';
 import { useVideoPlayer } from '../video-player-provider.component';
 
 export const VideoControls = () => {
     const { isMuted, toggleMute, videoControls } = useVideoPlayer();
     const { isPlaying, play, pause, previousFrame, nextFrame, canSelectPreviousFrame, canSelectNextFrame } =
         videoControls;
-    const { resetAnnotations } = useAnnotationActions();
 
     const handlePlay = async () => {
         await play();
-        resetAnnotations();
     };
 
     return (
--- a/application/ui/src/shared/annotator/annotation-actions-provider.component.tsx
+++ b/application/ui/src/shared/annotator/annotation-actions-provider.component.tsx
@@ -56,6 +56,7 @@ export const AnnotationActionsProvider = ({
     mode,
     isReadOnly = false,
 }: AnnotationActionsProviderProps) => {
+    const mediaItemKey = isVideoFrame(mediaItem) ? `${mediaItem.id}-${mediaItem.frame_number}` : mediaItem.id;
     const projectId = useProjectIdentifier();
     const saveMutation = $api.useMutation('post', '/api/projects/{project_id}/dataset/media/{media_id}/annotations', {
         meta: {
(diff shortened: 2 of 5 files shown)


application/ui/src/shared/annotator/annotation-actions-provider.component.tsx
```diff
@@ -142,7 +145,7 @@ export const AnnotationActionsProvider = ({
     };
 
     const resetAnnotations = () => {
-        undoRedoActions.reset([]);
+        undoRedoActions.reset(mapServerAnnotationsToLocal(initialAnnotationsDTO, projectLabels));
```
dwesolow: the intention of this was to clear canvas, not to get the initial annotations.

jpggvilaca: Agree, let's not change resetAnnotations and move this change to the mutation. Updated

application/ui/src/shared/annotator/annotation-actions-provider.component.tsx
```diff
@@ -157,7 +160,7 @@ export const AnnotationActionsProvider = ({
             body: { annotations: annotationsDTO },
         });
 
-        resetAnnotations();
+        undoRedoActions.reset(mapServerAnnotationsToLocal(annotationsDTO, projectLabels));
```
camiloHimura, line 173: After this change, do we still need `resetAnnotations`?

jpggvilaca: Not in this case. Because this is after the mutation so we update immediately with the annotations we're submitting

application/ui/src/shared/annotator/annotation-actions-provider.component.tsx
dwesolow: We need to call `reset([])`

application/ui/src/shared/annotator/annotation-actions-provider.component.tsx
dwesolow: Could you please the video case? If the annotation is drawn on frame N and we open frame N+1, we should not see the previous annotations.
I believe it should happen for the img as well, annotation from img N should not be visible for image N+1 (I guess in this case it's not a problem because we remount annotations actions based on the media id).

application/ui/src/shared/annotator/annotation-actions-provider.component.tsx
```diff
@@ -179,6 +182,15 @@ export const AnnotationActionsProvider = ({
...
     };
 
+    const hasChangedAnnotations = useMemo(() => {
+        const filteredAnnotations = filterOutAnnotationWithEmptyLabel(annotations);
+        const currentServerAnnotations = mapLocalAnnotationsToServer(filteredAnnotations);
+
+        return !isEqual(currentServerAnnotations, initialAnnotationsDTO);
+    }, [annotations, initialAnnotationsDTO]);
+
+    const canSubmit = mode === 'prediction' ? !isUserReviewed && predictions.length > 0 : hasChangedAnnotations;
```
dwesolow: I believe `predictions` are only with `!isUserReviewed`, we don't need to check this here.

jpggvilaca: Updated

application/ui/src/shared/annotator/annotation-actions-provider.component.tsx
jpggvilaca: Good point. I did check and it was not working for video. I tried changing the id on the provider but it result in the flicker we discussed offline. The final solution (first to last commit) was to store the key in a ref and then make the reset based on that

camiloHimura (dismissed):

application/ui/src/shared/annotator/annotation-actions-provider.component.tsx
jpggvilaca: Updated

application/ui/src/shared/annotator/annotation-actions-provider.component.tsx
```diff
@@ -87,27 +89,42 @@ export const AnnotationActionsProvider = ({
...
+    const initialAnnotations = useMemo(() => {
         return mapServerAnnotationsToLocal(initialAnnotationsDTO, projectLabels);
-    });
+    }, [initialAnnotationsDTO, projectLabels]);
 
-    const prevInitialAnnotationsDTORef = useRef(initialAnnotationsDTO);
+    const [annotations, setAnnotations, undoRedoActions] = useUndoRedoState<Annotation[]>(initialAnnotations);
+
+    const resetAnnotations = () => {
+        undoRedoActions.reset(initialAnnotations);
```
dwesolow, line 99: I think we should skip history so new media it's a new, fresh state?

application/ui/src/features/annotator/video-player/video-toolbar/video-controls.component.tsx
```diff
@@ -4,18 +4,15 @@
...
 import { useVideoPlayer } from '../video-player-provider.component';
 
 export const VideoControls = () => {
     const { isMuted, toggleMute, videoControls } = useVideoPlayer();
     const { isPlaying, play, pause, previousFrame, nextFrame, canSelectPreviousFrame, canSelectNextFrame } =
         videoControls;
-    const { resetAnnotations } = useAnnotationActions();
 
     const handlePlay = async () => {
         await play();
```
dwesolow: we no longer need to wrap `play` with `handlePlay`, we can call `play` on a video.

application/ui/src/shared/annotator/annotation-actions-provider.component.tsx
dwesolow: After your other changes, what is the reason of having this reset? If we draw an annotation and submit, we already have those annotation, we don't need reset to have them, correct?

jpggvilaca: Is it a good observation but we still need to, to make sure the undoRedo stack is fresh and prevents the user to undo to a "pre-submit" state. This (probably) wont happen on a normal connection but on a slow one this can definitely happen. You can try this yourself by throttling the connection and submitting an annotation (while commenting this line). We basically force the undoRedo reset instead of waiting for the new media refetch

application/ui/src/shared/annotator/annotation-actions-provider.component.tsx
jpggvilaca: I believe `reset` already does this. If you pass a value this will be new fresh state

application/ui/src/features/annotator/video-player/video-toolbar/video-controls.component.tsx
jpggvilaca: True. Removed

application/ui/src/shared/annotator/annotation-actions-provider.component.tsx
dwesolow: (we need to fix the issue with flickering img from top left corner to the middle, it'd allow us to remount this provider on video frame change and will handle everything out of the box for us)

dwesolow (approved):

PR #5729, opened 2026-03-10
Improve media upload part 2

jpggvilaca:
### Summary

* Allow uploading folders
* Add concurrency for media uploading (i think we should do the same for DELETING media, but maybe ill request an endpoint to bulk delete). I also invalidate the query for each chunk uploaded, if we dont want that it will be easy to remove
* On the next PR i will make the UI changes (show "in progress" etc)

Example uploading 12 images:
Screenshot 2026-03-09 at 13 38 57
Screenshot 2026-03-09 at 13 39 05
Screenshot 2026-03-09 at 13 39 15

### Checklist

### DIFF
--- a/application/ui/src/components/add-media-button/add-media-button.component.tsx
+++ b/application/ui/src/components/add-media-button/add-media-button.component.tsx
@@ -3,11 +3,10 @@
 
 import { ChangeEvent, useRef } from 'react';
 
-import { Button } from '@geti/ui';
+import { Button, Item, Key, Menu, MenuTrigger } from '@geti/ui';
 
 type AddMediaButtonProps = {
     onFilesSelected: (files: File[]) => void;
-    multiple?: boolean;
 };
 
 const VALID_VIDEO_EXT = ['mp4', 'avi', 'mkv', 'mov', 'webm', 'm4v'];
@@ -16,8 +15,20 @@ const VALID_EXT = [...VALID_VIDEO_EXT, ...VALID_IMAGE_EXT];
 
 export const acceptedExtensions = VALID_EXT.map((ext) => `.${ext}`).join(',');
 
-export const AddMediaButton = ({ onFilesSelected, multiple = true }: AddMediaButtonProps) => {
+export const AddMediaButton = ({ onFilesSelected }: AddMediaButtonProps) => {
     const fileInputRef = useRef<HTMLInputElement>(null);
+    const folderInputRef = useRef<HTMLInputElement>(null);
+
+    const setFolderInputRef = (input: HTMLInputElement | null) => {
+        folderInputRef.current = input;
+
+        if (input === null) {
+            return;
+        }
+
+        input.setAttribute('webkitdirectory', '');
+        input.setAttribute('directory', '');
+    };
 
     const handleFileChange = (event: ChangeEvent<HTMLInputElement>) => {
         const files = event.target.files;
@@ -33,8 +44,12 @@ export const AddMediaButton = ({ onFilesSelected, multiple = true }: AddMediaBut
         }
     };
 
-    const handleClick = () => {
-        fileInputRef.current?.click();
+    const handleMenuAction = (action: Key) => {
+        if (action === 'files') {
+            fileInputRef.current?.click();
+        } else if (action === 'folder') {
+            folderInputRef.current?.click();
+        }
     };
 
     return (
@@ -42,15 +57,28 @@ export const AddMediaButton = ({ onFilesSelected, multiple = true }: AddMediaBut
             <input
                 ref={fileInputRef}
                 type='file'
-                multiple={multiple}
+                multiple
                 onChange={handleFileChange}
                 style={{ display: 'none' }}
                 aria-label={'Upload media files'}
                 accept={acceptedExtensions}
             />
-            <Button variant={'secondary'} onPress={handleClick}>
-                Upload Files
-            </Button>
+            <input
+                ref={setFolderInputRef}
+                type='file'
+                multiple
+                onChange={handleFileChange}
+                style={{ display: 'none' }}
+                aria-label={'Upload media folder'}
+                accept={acceptedExtensions}
+            />
+            <MenuTrigger>
+                <Button variant={'secondary'}>Upload media</Button>
+                <Menu onAction={handleMenuAction} aria-label={'Upload menu'}>
+                    <Item key={'files'}>Files</Item>
+                    <Item key={'folder'}>Folder</Item>
+                </Menu>
+            </MenuTrigger>
         </>
     );
 };
--- a/application/ui/src/features/dataset/api/use-media-upload.test.tsx
+++ b/application/ui/src/features/dataset/api/use-media-upload.test.tsx
@@ -5,10 +5,11 @@ import { act, screen } from '@testing-library/react';
 import { getMockedMediaImage } from 'mocks/mock-media';
 import { HttpResponse } from 'msw';
 import { renderHook } from 'test-utils/render';
+import { v4 as uuid } from 'uuid';
 
 import { http } from '../../../api/utils';
 import { server } from '../../../msw-node-setup';
-import { useMediaUpload } from './use-media-upload';
+import { MEDIA_UPLOAD_CONCURRENCY, useMediaUpload } from './use-media-upload';
 
 describe('useMediaUpload', () => {
     it('uploads all selected files and shows a success toast', async () => {
@@ -42,18 +43,17 @@ describe('useMediaUpload', () => {
     });
 
     it('shows a warning toast when some uploads fail', async () => {
-        server.use(
-            http.post('/api/projects/{project_id}/dataset/media', async ({ request }) => {
-                const formData = await request.formData();
-                const file = formData.get('file') as File;
+        let requestCount = 0;
 
-                if (file.name === 'broken.jpg') {
-                    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-                    // @ts-expect-error
-                    return HttpResponse.json({ detail: 'Upload failed' }, { status: 500 });
+        server.use(
+            http.post('/api/projects/{project_id}/dataset/media', async () => {
+                requestCount += 1;
+
+                if (requestCount === 2) {
+                    return HttpResponse.error();
                 }
 
-                return HttpResponse.json(getMockedMediaImage({ id: crypto.randomUUID() }), { status: 201 });
+                return HttpResponse.json(getMockedMediaImage({ id: uuid() }), { status: 201 });
             })
         );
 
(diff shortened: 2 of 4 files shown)


application/ui/src/components/add-media-button/add-media-button.component.tsx
```diff
@@ -33,24 +44,41 @@ export const AddMediaButton = ({ onFilesSelected, multiple = true }: AddMediaBut
...
+        }
     };
 
     return (
         <>
             <input
                 ref={fileInputRef}
                 type='file'
-                multiple={multiple}
+                multiple
```
jpggvilaca, line 60: we always want `multiple` so i removed the optional prop (which was never passed anywhere)

application/ui/src/features/dataset/api/use-media-upload.test.tsx
```diff
@@ -42,18 +43,17 @@ describe('useMediaUpload', () => {
...
-                const formData = await request.formData();
-                const file = formData.get('file') as File;
+            http.post('/api/projects/{project_id}/dataset/media', async () => {
+                requestCount += 1;
 
-                if (file.name === 'broken.jpg') {
-                    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-                    // @ts-expect-error
-                    return HttpResponse.json({ detail: 'Upload failed' }, { status: 500 });
+                if (requestCount === 2) {
```
jpggvilaca, line 52: make the second request fail

application/ui/src/features/dataset/api/use-media-upload.ts
```diff
@@ -8,37 +8,64 @@ import { useProjectIdentifier } from 'hooks/use-project-identifier.hook';
...
             queryKey: getQueryKey([
                 'get',
                 '/api/projects/{project_id}/dataset/media',
                 { params: { path: { project_id: projectId } } },
             ]),
         });
 
+    const uploadMedia = async (files: File[]) => {
+        const uploadPromises = files.map((file) => {
+            return () => {
```
jpggvilaca, line 69: this is the main change. Instead of creating an array of promises, we created an array of functions that call these promises, so we can control when/how they are called (done on the new `executeInBatches` func).

application/ui/src/features/dataset/api/use-media-upload.ts
```diff
@@ -8,37 +8,64 @@ import { useProjectIdentifier } from 'hooks/use-project-identifier.hook';
...
-                // eslint-disable-next-line @typescript-eslint/no-explicit-any
-                body: formData as any,
-            });
-        });
+        settledResults.push(...batchResults);
 
-        const promises = await Promise.allSettled(uploadPromises);
+        if (onBatchCompleted !== undefined) {
+            await onBatchCompleted();
+        }
```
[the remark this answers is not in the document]
jpggvilaca: Yeah, mixed feelings about this. For me I would wait for ALL uploads, to make everything simpler and more predicable. WDYT guys? This was my initial thought but then i remember on classic we were showing everything as it was uploaded, that is why i added `onBatchCompleted`. TBH it's best to wait for ALL of it and not use `onBatchCompleted` at all

application/ui/src/features/dataset/api/use-media-upload.ts
```diff
@@ -8,37 +8,64 @@ import { useProjectIdentifier } from 'hooks/use-project-identifier.hook';
...
+        const uploadPromises = files.map((file) => {
+            return () => {
+                const formData = new FormData();
+                formData.append('file', file);
+
+                return addItemMutation.mutateAsync({
+                    params: { path: { project_id: projectId } },
+                    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+                    body: formData as any,
+                });
```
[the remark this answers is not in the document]
jpggvilaca: Nice, i did not know about `useIsMutating`. Updated

application/ui/src/features/dataset/api/use-media-upload.test.tsx
```diff
@@ -5,10 +5,11 @@ import { act, screen } from '@testing-library/react';
 import { getMockedMediaImage } from 'mocks/mock-media';
 import { HttpResponse } from 'msw';
 import { renderHook } from 'test-utils/render';
+import { v4 as uuid } from 'uuid';
```
camiloHimura, line 8: let's consider to use https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID

application/ui/src/features/dataset/api/use-media-upload.ts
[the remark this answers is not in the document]
camiloHimura: I tend to lean toward waiting until everything is fully loaded.

application/ui/src/features/dataset/api/use-media-upload.ts
```diff
@@ -2,43 +2,73 @@
...
-            return addItemMutation.mutateAsync({
-                params: { path: { project_id: projectId } },
-                // eslint-disable-next-line @typescript-eslint/no-explicit-any
-                body: formData as any,
-            });
-        });
+        settledResults.push(...batchResults);
 
-        const promises = await Promise.allSettled(uploadPromises);
+        if (onBatchCompleted !== undefined) {
```
camiloHimura:
```suggestion
        if (isFunction(onBatchCompleted)) {
```

application/ui/src/components/add-media-button/add-media-button.component.tsx
```diff
@@ -33,24 +44,41 @@ export const AddMediaButton = ({ onFilesSelected, multiple = true }: AddMediaBut
         }
     };
 
-    const handleClick = () => {
-        fileInputRef.current?.click();
+    const handleMenuAction = (action: Key) => {
+        if (action === 'files') {
```
camiloHimura, line 48: Could it have another kind of action besides files and folders?

application/ui/src/features/dataset/api/use-media-upload.ts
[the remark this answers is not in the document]
jpggvilaca: Same, but i can also see the benefits of the other side. Let's  see what @dwesolow  thinks as well

application/ui/src/features/dataset/api/use-media-upload.test.tsx
jpggvilaca: It was there before, i replaced it. What is the main difference?

application/ui/src/components/add-media-button/add-media-button.component.tsx
jpggvilaca: No : )

application/ui/src/features/dataset/api/use-media-upload.ts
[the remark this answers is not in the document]
ActiveChooN: I agree with waiting for a complete upload, especially since the most common use case is local and there shouldn't be any throughput problems

application/ui/src/components/add-media-button/add-media-button.component.tsx
dwesolow: Camera 😄

jpggvilaca: Definitely wont add that right now XD

jpggvilaca: but yes, good point!

application/ui/src/features/dataset/api/use-media-upload.ts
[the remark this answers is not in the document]
dwesolow: IMO it depends on the use case. If user uploads 1000 files and upload takes 1 minute or so, it'd be better to show files progressively instead of after 1 minute. TBH We can call `onBatchCompleted` but not `await`, just fire and forget, in the background files will be refetched. WDYT?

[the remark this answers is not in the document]
jpggvilaca: I thought about that too but the "issue" is the amount of requests, not the blocking part. We will refetch the media query for each batch completed

application/ui/src/features/dataset/api/use-media-upload.ts
```diff
@@ -2,43 +2,73 @@
...
-        const succeeded = promises.filter((result) => result.status === 'fulfilled').length;
-        const failed = promises.filter((result) => result.status === 'rejected').length;
+export const useMediaUpload = () => {
+    const projectId = useProjectIdentifier();
+    const queryClient = useQueryClient();
 
-        await queryClient.invalidateQueries({
+    const addItemMutation = $api.useMutation('post', '/api/projects/{project_id}/dataset/media');
+    const currentlyUploading = useIsMutating({
+        mutationKey: ['post', '/api/projects/{project_id}/dataset/media'],
```
dwesolow, line 51: is `mutationKey` set by default on a `addItemMutation`?
BTW I think we miss `{ params: { path: { project_id: projectId } } }` here.

application/ui/src/features/dataset/api/use-media-upload.test.tsx
camiloHimura: I’d say they’re the same, but with `crypto.randomUUID()` we wouldn’t need the third-party library anymore.

jpggvilaca: For now since we're using the package i'd like to leave it. But we can open a PR to get rid of it in one go

application/ui/src/features/dataset/api/use-media-upload.ts
jpggvilaca: Good catch! Used query predicate for this

```
/**
     * Include mutations matching this predicate function
     */
    predicate?: (mutation: Mutation<TData, TError, TVariables, TContext>) => boolean;
```

application/ui/src/features/dataset/api/use-media-upload.ts
jpggvilaca: Updated

application/ui/src/features/dataset/api/use-media-upload.test.tsx
camiloHimura: Sure, I just wanted to put it on the table as a valid option to consider

camiloHimura (approved):

application/ui/src/features/dataset/api/use-media-upload.test.tsx
jpggvilaca: And that is a good point : ) . I suggest getting rid of it entirely on a separate PR

application/ui/src/features/dataset/api/use-media-upload.ts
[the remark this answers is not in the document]
jpggvilaca: Ok for now i will leave it as is, with `await`. On the next PR we can improve on this because i will add the logic to DISPLAY STATUS, so we might still need `onBatchCompleted` here.

dwesolow (approved):

ActiveChooN (approved):

PR #5730, opened 2026-03-10
Add benchmark metrics to model manifest (and remove performance ratings)

### LINKED ISSUE #5676: Update manifests and API to return benchmark info for model architectures
In the model manifests, replace the `performance_ratings` section (subjective) with a new `benchmark_metrics` object that contains objective metrics computed on standard datasets.

What we have:
```yaml
stats:
  ...
  performance_ratings:  # scores range from 1 (below average) to 3 (excellent)
    accuracy: 3
    training_time: 2
    inference_speed: 2
```

What we want (classification):
```yaml
stats:
  ...
  benchmark_metrics:
    imagenet_top1_accuracy: 77.1  # percentage
    imagenet_top5_accuracy: 93.3  # percentage (optional)
```

What we want (detection and segmentation):
```yaml
stats:
  ...
  benchmark_metrics:
    coco_map_50_95: 51.2  # mAP at IoU=0.50:0.95
    coco_map_50: 69.6     # mAP at IoU=0.50 (optional)
```

Then update the `ModelManifest` class and the parser accordingly. Also update the response of the `/model_architectures` endpoint to mirror the new structure.

JortBergfeld:
### Summary

Add benchmark metrics to model manifests: top 1 and top 5 accuracy on imagenet for classificaiton and mAP@0.5 and mAP@0.5;0.95 for detection and segmentation.
Remove subjective performance ratings from the model manifest 

Resolves #5676 

### How to test

Adjusted tests to new layout

### Checklist

### DIFF
--- a/application/backend/app/models/model_manifest.py
+++ b/application/backend/app/models/model_manifest.py
@@ -21,36 +21,41 @@ class ModelManifestDeprecationStatus(str, Enum):
         return str(self.name)
 
 
-class PerformanceRatings(BaseModel):
-    """Ratings for different performance aspects of a model."""
+class BenchmarkMetrics(BaseModel):
+    """Benchmark metrics for different model tasks."""
 
     model_config = ConfigDict(extra="forbid")
-    accuracy: int = Field(
-        ge=1,
-        le=3,
-        default=1,
-        title="Accuracy rating",
-        description="Rating of the model accuracy. "
-        "The value should be interpreted relatively to the other available models, "
-        "and it ranges from 1 (below average) to 3 (above average).",
+
+    # Classification metrics
+    imagenet_top1_accuracy: float | None = Field(
+        default=None,
+        title="ImageNet Top-1 Accuracy",
+        description="Top-1 accuracy on ImageNet (percentage)",
+        ge=0,
+        le=100,
     )
-    training_time: int = Field(
-        ge=1,
-        le=3,
-        default=1,
-        title="Training time rating",
-        description="Rating of the model training time. "
-        "The value should be interpreted relatively to the other available models, "
-        "and it ranges from 1 (below average/slower) to 3 (above average/faster).",
+    imagenet_top5_accuracy: float | None = Field(
+        default=None,
+        title="ImageNet Top-5 Accuracy",
+        description="Top-5 accuracy on ImageNet (percentage, optional)",
+        ge=0,
+        le=100,
     )
-    inference_speed: int = Field(
-        ge=1,
-        le=3,
-        default=1,
-        title="Inference speed rating",
-        description="Rating of the model inference speed. "
-        "The value should be interpreted relatively to the other available models, "
-        "and it ranges from 1 (below average/slower) to 3 (above average/faster).",
+
+    # Detection/Segmentation metrics
+    coco_map_50_95: float | None = Field(
+        default=None,
+        title="COCO mAP 50-95",
+        description="COCO mean Average Precision at IoU=0.50:0.95 (percentage)",
+        ge=0,
+        le=100,
+    )
+    coco_map_50: float | None = Field(
+        default=None,
+        title="COCO mAP 50",
+        description="COCO mean Average Precision at IoU=0.50 (percentage, optional)",
+        ge=0,
+        le=100,
     )
 
 
@@ -67,8 +72,8 @@ class ModelStats(BaseModel):
         title="Trainable parameters (millions)",
         description="Number of trainable parameters in the model, expressed in millions",
     )
-    performance_ratings: PerformanceRatings = Field(
-        title="Performance ratings", description="Standardized ratings for model performance metrics"
+    benchmark_metrics: BenchmarkMetrics = Field(
+        title="Benchmark metrics", description="Standardized benchmark metrics for model performance"
     )
 
 
--- a/application/backend/app/models/model_manifest.py
+++ b/application/backend/app/models/model_manifest.py
@@ -58,23 +58,6 @@ class BenchmarkMetrics(BaseModel):
         le=100,
     )
 
-    @model_validator(mode="after")
-    def validate_metrics(self) -> "BenchmarkMetrics":
-        if self.imagenet_top1_accuracy is None and self.coco_map_50_95 is None:
-            raise ValueError(
-                f"For classification 'imagenet_top1_accuracy' is required, for detection and segmentation "
-                f"'coco_map_50_95' is required. However, both or 'None': {self}"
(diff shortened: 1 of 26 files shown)


application/backend/app/supported_models/manifests/classification/deit_tiny.yaml
```diff
@@ -9,10 +9,9 @@ description: "DeiT (Data-efficient Image Transformers) is a vision transformer m
 stats:
   gigaflops: 1.26
   trainable_parameters: 5.2
-  performance_ratings: # scores range from 1 (below average) to 3 (excellent)
-    accuracy: 2
-    training_time: 3
-    inference_speed: 3
+  benchmark_metrics:
+    imagenet_top1_accuracy: 72.2 # percentage
+    imagenet_top5_accuracy: 91.1 # percentage (optional)
```
leoll2: Let's remove these inline comments from the YAML file; the meaning of every field is already defined in the Pydantic model.

application/backend/app/models/model_manifest.py
```diff
@@ -21,36 +21,41 @@ def __str__(self) -> str:
...
+        default=None,
+        title="ImageNet Top-1 Accuracy",
+        description="Top-1 accuracy on ImageNet (percentage)",
+        ge=0,
+        le=100,
+    )
+    imagenet_top5_accuracy: float | None = Field(
+        default=None,
+        title="ImageNet Top-5 Accuracy",
+        description="Top-5 accuracy on ImageNet (percentage, optional)",
```
leoll2:
```suggestion
        description="Top-5 accuracy on ImageNet (percentage, null if N/A)",
```

application/backend/app/models/model_manifest.py
```diff
@@ -21,36 +21,41 @@ def __str__(self) -> str:
...
-        le=3,
-        default=1,
-        title="Inference speed rating",
-        description="Rating of the model inference speed. "
-        "The value should be interpreted relatively to the other available models, "
-        "and it ranges from 1 (below average/slower) to 3 (above average/faster).",
+    coco_map_50: float | None = Field(
+        default=None,
+        title="COCO mAP 50",
+        description="COCO mean Average Precision at IoU=0.50 (percentage, optional)",
```
leoll2:
```suggestion
        description="COCO mean Average Precision at IoU=0.50 (percentage, null if N/A)",
```

application/backend/app/models/model_manifest.py
```diff
@@ -21,36 +21,41 @@ def __str__(self) -> str:
         return str(self.name)
 
 
-class PerformanceRatings(BaseModel):
-    """Ratings for different performance aspects of a model."""
+class BenchmarkMetrics(BaseModel):
```
leoll2, line 24: Can you add a `@model_validator` to validate that:
- At least one of `imagenet_top1_accuracy` and `coco_map_50_95` is not null
- If `imagenet_top1_accuracy` is not null, then the mAP ones are null
- If `coco_map_50_95` is not null, then the top-K accuracy ones are null

leoll2 (commented):
Backend changes LGTM

application/ui/src/features/models/train-model/sort-model-architectures/utils.ts
```diff
@@ -20,15 +20,17 @@ type SortingHandler = (
     modelArchitectures: ModelArchitectureWithPerformanceCategory[]
 ) => ModelArchitectureWithPerformanceCategory[];
 
+const getAccuracyMetricBasedOnTask = ({
+    stats: { benchmark_metrics: benchmarkMetrics },
+}: ModelArchitectureWithPerformanceCategory) => {
+    return benchmarkMetrics.imagenet_top1_accuracy ?? benchmarkMetrics.coco_map_50_95 ?? benchmarkMetrics.coco_map_50;
```
dwesolow, line 26: What do those metrics mean? And why is that order of selecting them?

jpggvilaca: i have no idea what they mean, but Leo said we should use these. As for the order, iirc, you will have certain fields for classification, and certain for detection/segmentation. We basuically pick the ones available. cc @leoll2

leoll2: These metrics provide an indication of the accuracy of a given model architecture. For classification models we expose `imagenet_top1_accuracy`, namely "Top-1 Accuracy on ImageNet" dataset, while for detection we return `coco_map_50_95`, namely "mAP@0.5:0.95 on COCO" dataset.

UI should use these values to sort the model architectures by performance: the higher the score, the more accurate the model. The metric value can be displayed either in the respective model architecture card, or within the tooltip with the extended description.

dwesolow: Thanks for explanation! So it means that in the UI we should select the metric based on the task, no? @jpggvilaca

jpggvilaca: If i understood correctly the backend will send the right parameters based on the task type. @JortBergfeld  @leoll2  ?

leoll2: Yes, the server response contains a non-null value for `imagenet_top1_accuracy` or `coco_map_50_95` depending on the task type. It's a exclusive OR: one or the other, but never both null or both set.

leoll2 (dismissed):

application/backend/app/supported_models/manifests/detection/deim_x.yaml
```diff
@@ -9,10 +9,9 @@ description: "DEIM is an advanced training framework designed to enhance the mat
 stats:
   gigaflops: 202
   trainable_parameters: 62
-  performance_ratings: # scores range from 1 (below average) to 3 (excellent)
-    accuracy: 3
-    training_time: 1
-    inference_speed: 1
+  benchmark_metrics:
+    coco_map_50_95: 55.8
```
kprokofi: I think it is a mistake. The numbers are the same comparing with dfine_x

kprokofi (commented):
While Imagenet and COCO can provide some hints which model to choose, from my experience, models can perform very differently depending on the given task. @leoll2 I would suggest to expose metrics on Fine-grained datasets in the future along with metrics from the papers.

dwesolow (dismissed):

kprokofi (approved):

dwesolow (approved):

leoll2 (approved):

PR #5737, opened 2026-03-10
Add more training parameters

### LINKED ISSUE #5711: Expose more training parameters
New keys:
- `training.batch_size` (int)
- `training.weight_decay` (float 0-1)
- `training.scheduler.type` (str enum, ["reduce_lr_on_plateau", "cosine_annealing"])
- `training.scheduler.warmup.enable` (bool)
- `training.scheduler.warmup.epochs` (int)
- `training.scheduler.factor` (float 0-1) (only for reduce_lr_on_plateau)
- `training.scheduler.patience` (int) (only for reduce_lr_on_plateau)
- `training.scheduler.min_lr` (float 0-1) (only for cosine_annealing)
- `training.gradient_accumulation.enable` (bool)
- `training.gradient_accumulation.batches` (int)
- `training.gradient_clip.enable` (bool)
- `training.gradient_clip.max_grad_norm` (float)

leoll2:
### Summary

This PR exposes a bunch of new parameters in the training configuration:
- `training.batch_size` (int)
- `training.weight_decay` (float 0-1)
- `training.scheduler.type` (str enum, ["reduce_lr_on_plateau", "cosine_annealing"])
- `training.scheduler.warmup.enable` (bool)
- `training.scheduler.warmup.epochs` (int)
- `training.scheduler.factor` (float 0-1) (only for reduce_lr_on_plateau)
- `training.scheduler.patience` (int) (only for reduce_lr_on_plateau)
- `training.scheduler.min_lr` (float 0-1) (only for cosine_annealing)
- `training.gradient_accumulation.enable` (bool)
- `training.gradient_accumulation.batches` (int)
- `training.gradient_clip.enable` (bool)
- `training.gradient_clip.max_grad_norm` (float)

Model manifests are updated accordingly, however the OTX configuration converter will be updated in another PR to avoid conflicts with #5021. This means that these parameters are not functional yet, but they will be very soon.

Besides, the parameters in the `TrainingConfigurationView` now have an additional property `depends_on`, which should be used by the frontend to conditionally display/hide certain parameters based on the value of other params.

Resolves #5711

### Checklist

### DIFF
--- a/application/backend/app/models/training_configuration/training.py
+++ b/application/backend/app/models/training_configuration/training.py
@@ -27,17 +27,17 @@ class SchedulerType(StrEnum):
     COSINE_ANNEALING = "cosine_annealing"
 
 
-class LearningRateWarmupParameters(BaseModel):
+class LrLinearWarmupParameters(BaseModel):
     enable: bool = Field(
         default=False,
         title="Enable",
-        description="Toggle to enable or disable a warmup phase at the beginning of training.",
+        description="Toggle to enable or disable the LR linear warmup phase at the beginning of training.",
     )
     epochs: int = Field(
         default=5,
         ge=1,
         title="Warmup epochs",
-        description="Number of epochs for the warmup phase.",
+        description="Number of epochs for the LR linear warmup phase.",
     )
 
 
@@ -52,9 +52,9 @@ class SchedulerParameters(BaseModel):
             "gradually decreasing over the course of training."
         ),
     )
-    warmup: LearningRateWarmupParameters = Field(
-        default_factory=LearningRateWarmupParameters,
-        title="Learning rate warmup",
+    warmup: LrLinearWarmupParameters = Field(
+        default_factory=LrLinearWarmupParameters,
+        title="Learning rate linear warmup",
         description=(
             "Learning rate warmup is a technique where the learning rate starts at a lower value and gradually "
             "increases to the initial learning rate over a specified number of epochs at the beginning of training. "
@@ -94,8 +94,11 @@ class GradientAccumulationParameters(BaseModel):
     batches: int = Field(
         ge=1,
         default=1,
-        title="Accumulation batches",
-        description="Number of batches to accumulate gradients before performing a weight update.",
+        title="Gradient accumulation batches",
+        description=(
+            "Number of steps (batches) to accumulate gradients before performing gradient descent step. "
+            "Effective batch size during training: batch_size * accumulate_grad_batches."
+        ),
     )
 
 
@@ -108,8 +111,8 @@ class GradientClipParameters(BaseModel):
     max_grad_norm: float = Field(
         gt=0,
         default=1.0,
-        title="Maximum gradient norm",
-        description="Maximum norm of the gradients. Gradients with norm larger than this value will be clipped.",
+        title="Maximum gradient L2 norm",
+        description="Maximum L2 norm of the gradients. Gradients with norm larger than this value will be clipped.",
     )
 
 
@@ -131,7 +134,8 @@ class AlgoLevelTrainingParameters(BaseModel):
         description=(
             "Number of training samples processed before the model's internal parameters are updated. "
             "A larger batch size can speed up training but may require more memory, while a smaller batch size "
-            "can lead to more stable convergence but may take longer to train."
+            "can help avoid OOM (Out of Memory) errors at the cost of longer training times and potentially noisier "
+            "gradient estimates."
         ),
     )
     early_stopping: EarlyStopping = Field(
@@ -159,7 +163,8 @@ class AlgoLevelTrainingParameters(BaseModel):
         title="Weight decay",
         description=(
             "Weight decay is a regularization technique that adds a penalty to the loss function based on the "
-            "magnitude of the model weights. It helps prevent overfitting by discouraging large weight values."
+            "squared magnitude of the model weights (L2 regularization). "
+            "It helps prevent overfitting by discouraging large weight values."
         ),
     )
     scheduler: SchedulerParameters = Field(
--- a/application/backend/app/models/training_configuration/training.py
+++ b/application/backend/app/models/training_configuration/training.py
@@ -94,8 +94,8 @@ class GradientAccumulationParameters(BaseModel):
     batches: int = Field(
         ge=1,
         default=1,
-        title="Accumulation steps",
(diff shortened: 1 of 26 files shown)


application/backend/app/models/training_configuration/training.py
```diff
@@ -1,34 +1,139 @@
...
     enable: bool = Field(
         default=False,
-        title="Toggle early stopping",
-        description="Whether to stop training early when performance stops improving",
+        title="Enable",
+        description="Toggle to enable or disable early stopping during training.",
     )
     patience: int = Field(
-        gt=0,
+        ge=1,
```
kprokofi, line 16: is it minimum?

application/backend/app/models/training_configuration/training.py
```diff
@@ -1,34 +1,139 @@
...
+    enable: bool = Field(
+        default=False,
+        title="Enable",
+        description="Toggle to enable or disable a warmup phase at the beginning of training.",
+    )
+    epochs: int = Field(
+        default=5,
+        ge=1,
+        title="Warmup epochs",
+        description="Number of epochs for the warmup phase.",
```
kprokofi: Could you mention here that it is linear warmup? We don't expose other options for now

application/backend/app/models/training_configuration/training.py
```diff
@@ -1,34 +1,139 @@
...
+        description="Minimum learning rate after annealing.",
+        json_schema_extra={"depends_on": {"type": "cosine_annealing"}},
+    )
+
+
+class GradientAccumulationParameters(BaseModel):
+    enable: bool = Field(
+        default=False, title="Enable", description="Toggle to enable or disable gradient accumulation during training."
+    )
+    batches: int = Field(
```
kprokofi, line 94: This parameter is unclear. Let's stick to the PyTorch Lightning naming https://lightning.ai/docs/pytorch/stable/advanced/training_tricks.html#accumulate-gradients

accumulate_grad_batches: int 
Description: Number of steps (batches) to accumulate gradients before performing gradient descent step.  Effective batch size during training: batch_size * accumulate_grad_batches

application/backend/app/models/training_configuration/training.py
```diff
@@ -1,34 +1,139 @@
...
+    enable: bool = Field(
+        default=False,
+        title="Enable",
+        description="Toggle to enable or disable gradient clipping during training.",
+    )
+    max_grad_norm: float = Field(
+        gt=0,
+        default=1.0,
+        title="Maximum gradient norm",
+        description="Maximum norm of the gradients. Gradients with norm larger than this value will be clipped.",
```
kprokofi: Here we can emphasize that the norm here is Euclidean norm or L2

application/backend/app/models/training_configuration/training.py
```diff
@@ -1,34 +1,139 @@
...
         ),
     )
+    batch_size: int = Field(
+        ge=1,
+        default=4,
+        title="Batch size",
+        description=(
+            "Number of training samples processed before the model's internal parameters are updated. "
+            "A larger batch size can speed up training but may require more memory, while a smaller batch size "
+            "can lead to more stable convergence but may take longer to train."
```
kprokofi: This statement generally is not correct. Smaller batch sizes introduce noise to the training and on the contrary could lead to instability. I would write here, that smaller batch size helps to mitigate the OOM problem, but in cost of numerical stability and training time. Sometimes smaller batch size could play a role of regularization for smaller datasets, but it is another one topic. 
We could suggest here to use gradient accumulation parameter to synthetically increase effective batch size and better convergence.

application/backend/app/models/training_configuration/training.py
```diff
@@ -47,6 +152,39 @@ class AlgoLevelTrainingParameters(BaseModel):
             "training but risk overshooting minima in the loss landscape."
         ),
     )
+    weight_decay: float = Field(
+        ge=0,
+        lt=1,
+        default=1e-4,
+        title="Weight decay",
+        description=(
+            "Weight decay is a regularization technique that adds a penalty to the loss function based on the "
```
kprokofi, line 165: Please, add a note that we use L2 norm for regularization (it applies for all optimizers we use)

application/backend/app/supported_models/manifests/instance_segmentation/rtmdet_tiny.yaml
```diff
@@ -70,5 +70,15 @@ hyperparameters:
...
+    weight_decay: 0.0001
+    scheduler:
+      factor: 0.1
+      patience: 9
+      warmup:
+        enable: true
+        epochs: 20
+    gradient_clip:
+      enable: true
+      max_grad_norm: 0.1
```
kprokofi: we set this parameter to 35.0 for this model. I think this parameter should be reconsidered since we use L2 norm. Anyway, let's set everything as in recipes

leoll2: Good catch, updated

application/backend/app/models/training_configuration/training.py
leoll2: Update description

application/backend/app/models/training_configuration/training.py
leoll2: Updated description

application/backend/app/models/training_configuration/training.py
leoll2: Updated title and description

application/backend/app/models/training_configuration/training.py
leoll2: Updated parameter title and description

application/backend/app/models/training_configuration/training.py
leoll2: Yes, it's a validation constraint on the minimum `patience` value. Note that this parameter only takes effect when `warmup.enable=true`.

application/backend/app/models/training_configuration/training.py
leoll2: Updated title. The full parameter key is `training.gradient_accumulation.batches` and will be displayed as:
```
Gradient accumulation
    Enable
    Gradient accumulation batches
```
I hope this is clear enough

kprokofi (approved):

A-Artemis (approved):

PR #5744, opened 2026-03-11
Improve media upload part 3 (last part)

### LINKED ISSUE #5747: Enhance media upload UX
* Allow folder upload (MVP)
* Stagger uploads to avoid server overload
* Add status report when the upload starts, updates, and finishes through the toast medium

jpggvilaca:
### Summary

* Show notification during upload
* Add/edit tests
* Revert upload change (do not allow folder upload for MVP)
* Unskip test
* For now i got rid of any queues and complex logic and just disable the upload button IF THERE IS AN UPLOAD ONGOING

rec

### Checklist


application/ui/src/features/dataset/hooks/use-display-upload-progress.ts
```diff
@@ -0,0 +1,117 @@
+// Copyright (C) 2026 Intel Corporation
```
dwesolow, line 1: it's not related to the API hooks, shouldn't be here.

application/ui/src/features/dataset/hooks/use-display-upload-progress.ts
```diff
@@ -0,0 +1,117 @@
...
+            const nextSucceeded = previousProgress.succeeded + succeeded;
+            const nextFailed = previousProgress.failed + failed;
+
+            const msg = `Uploading ${total} item(s)... (${nextSucceeded} succeeded, ${nextFailed} failed)`;
+
+            toast({
+                id: UPLOAD_TOAST_ID,
+                type: 'info',
+                message: msg,
+            });
```
dwesolow, line 67: I don't think this should be part of the state update.

application/ui/src/features/dataset/api/use-media-upload.test.tsx
```diff
@@ -34,95 +34,93 @@ describe('useMediaUpload', () => {
...
+            result.current.uploadMedia(files);
         });
 
-        expect(maxRunningUploads).toBeLessThanOrEqual(MEDIA_UPLOAD_CONCURRENCY);
-        expect(await screen.findByText('Uploaded 12 item(s)')).toBeVisible();
+        await waitFor(() => {
+            expect(result.current.uploadProgress.isUploading).toBe(false);
+        });
+
+        expect(result.current.uploadProgress).toEqual({
```
dwesolow, line 117: Could we test the UI instead of the implementation details? I.e. to check if the proper toast is visible instead of current approach.

dwesolow:
TBH I don't see any value of uploading media in batches, for me this is only a limitation. If server is able to handle thousands of files, why do we want to restrict that.
On the other hand, if we really want to have only 5 ongoing requests, I'd not force all 5 media items to be uploaded before starting the next batch. If 1 of the requests finished, we upload a new media while the rest requests (4) are still ongoing).

application/ui/src/features/dataset/api/use-media-upload.test.tsx
jpggvilaca: I think we should check both. This one is valuable for testing the hook. The other can be a component test actually uploading and checking the toasts. I will add on my next commit

jpggvilaca:
> TBH I don't see any value of uploading media in batches, for me this is only a limitation. If server is able to handle thousands of files, why do we want to restrict that. On the other hand, if we really want to have only 5 ongoing requests, I'd not force all 5 media items to be uploaded before starting the next batch. If 1 of the requests finished, we upload a new media while the rest requests (4) are still ongoing).

Fair point. It was requested to us to be more server-friendly with the argument that the user will upload hundreds (if not thousands) of media items at a time. On a normal network this hardly caps the upload speed. For now let's no do any more complex logic and stick with the basics. If worse case scenario we want to revert all we need to do is to get rid of that BATCH_LIMIT number variable. For now i increased the batch limit from 5 to 10

application/ui/src/features/dataset/hooks/use-display-upload-progress.ts
jpggvilaca: Moved to `dataset/hooks`

application/ui/src/features/dataset/hooks/use-display-upload-progress.ts
jpggvilaca: I also dont like it but it's either that or add a `useEffect`. If not then you'll end up with buggy out-of-sync values

application/ui/src/features/dataset/api/use-media-upload.test.tsx
jpggvilaca: Added component test

application/ui/tests/datasets/datasets.spec.ts
```diff
@@ -94,4 +95,37 @@ test.describe('Dataset', () => {
...
     });
+
+    test('upload shows start, progress and finish toasts', async ({ page, network }) => {
+        let shouldResumeUpload = false;
+
+        network.use(
+            http.post('/api/projects/{project_id}/dataset/media', async () => {
+                while (!shouldResumeUpload) {
+                    await new Promise((resolve) => setTimeout(resolve, 10));
+                }
```
dwesolow: Why do we need this?

jpggvilaca: It's basically a small delay so we are able to see the "in progress" toasts, or else all the requests finish immediately and we only see start/finish. But I replaced the `while` for simplicity and added a comment

application/ui/tests/datasets/datasets.spec.ts
```diff
@@ -97,13 +98,17 @@ test.describe('Dataset', () => {
...
         network.use(
             http.post('/api/projects/{project_id}/dataset/media', async () => {
-                while (!shouldResumeUpload) {
-                    await new Promise((resolve) => setTimeout(resolve, 10));
-                }
+                uploadRequestCount += 1;
+                const isLastUploadRequest = uploadRequestCount === totalFiles;
+
+                // Small delay to test the "in progress" toast or else we would only see start and finish toasts
+                await new Promise((resolve) => setTimeout(resolve, isLastUploadRequest ? 250 : 30));
```
jpggvilaca, line 110: basically the ${firstBatchCount} will finish quickly (30ms) and the last will take a bit longer, so we can still the "in progress" toast

dwesolow (dismissed):

dwesolow (approved):

PR #5775, opened 2026-03-13
General fixes

jpggvilaca:
### Summary

Closes a few issues from https://github.com/open-edge-platform/training_extensions/issues/5674

* Show evaluations separately (screenshot)
* X-axis for all charts now starts at 0 (model metrics)
* X and Y labels now have the same color (model metrics)
* For classification multi-label, require at least 2 labels
* Add component tests for the point above
* Hide labels from annotations if they're being edited (preventing the anchors to be on top of the label, or the label being on top of anchors). For now only applied to polygon tool (Bugfix) but we should disable for all of them imho
* If we remove a label with EMPTY_LABEL, we need to remove the selected label, or else if we annotate after that, then we will use the EMPTY_LABEL as pre-selected label
* Add general.spec to host random tests that dont fit in any feature spec

Screenshot 2026-03-13 at 13 16 26
Screenshot 2026-03-13 at 13 17 33
Screenshot 2026-03-13 at 13 46 58
Screenshot 2026-03-13 at 15 22 45

### Checklist

### DIFF
--- a/application/ui/src/features/annotator/annotations/annotation-shape-with-labels.component.tsx
+++ b/application/ui/src/features/annotator/annotations/annotation-shape-with-labels.component.tsx
@@ -9,8 +9,10 @@ import polylabel from 'polylabel';
 import type { Label } from '../../../constants/shared-types';
 import { useAnnotationActions } from '../../../shared/annotator/annotation-actions-provider.component';
 import { useAnnotationVisibility } from '../../../shared/annotator/annotation-visibility-provider.component';
+import { EMPTY_LABEL_ID } from '../../../shared/annotator/labels';
 import type { Annotation } from '../../../shared/types';
 import { isClassificationTask } from '../../project/task-type-guards';
+import { useAnnotatorLabels } from '../annotator-labels-provider.component';
 import { AnnotationLabels } from './annotation-labels/annotation-labels.component';
 import { AnnotationShape } from './annotation-shape/annotation-shape.component';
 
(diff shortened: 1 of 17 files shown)


leoll2:
Nice work with all the fixes! I only have 2 comments about this one:
image

1. It's **multiclass** classification that needs 2 labels, not **multilabel**
2. The warning looks intimidating, it pops up in the middle of the label creation process as if the user did something wrong. I suggest to display that warning if and only if the user clicks "Create button" while there is only 1 label for this project type, and also add a hint like "minimum 2 labels" under the "What objects should the model learn to classify" text.

application/ui/src/features/annotator/annotations/annotation-shape-with-labels.component.tsx
```diff
@@ -30,6 +33,10 @@ export const AnnotationShapeWithLabels = ({ annotation }: AnnotationShapeProps)
             return;
         }
 
+        if (labelId === EMPTY_LABEL_ID && selectedLabelId === EMPTY_LABEL_ID) {
```
camiloHimura, line 36: This updates a state outside the component’s context, should it be inside a `useEffect`?

application/ui/src/features/project/create/create-project-form.tsx
```diff
@@ -42,8 +42,16 @@ export const CreateProjectForm = ({ projects }: CreateProjectFormProps) => {
         projects.map((project) => project.name)
     );
 
+    const isMultiLabelClassification = isClassificationTask(selectedTask) && classificationTaskType === 'multi-label';
+
+    const needsMinimumNumberOfLabels = isMultiLabelClassification && labels.length < 2;
+    const multiLabelClassificationErrorMessage = 'At least 2 labels are required for multi-label classification';
```
camiloHimura: np, this variable is only used on line 139, so the string can be used directly.

application/ui/src/features/annotator/annotations/annotation-shape-with-labels.component.tsx
jpggvilaca: what do you mean?

application/ui/src/features/project/create/create-project-form.tsx
jpggvilaca: Not applicable anymore

jpggvilaca:
> Nice work with all the fixes! I only have 2 comments about this one: image
> 
> 1. It's **multiclass** classification that needs 2 labels, not **multilabel**
> 2. The warning looks intimidating, it pops up in the middle of the label creation process as if the user did something wrong. I suggest to display that warning if and only if the user clicks "Create button" while there is only 1 label for this project type, and also add a hint like "minimum 2 labels" under the "What objects should the model learn to classify" text.

Updated

camiloHimura (approved):

PR #5787, opened 2026-03-16
Lighthouse audit

jpggvilaca:
### Summary

We could improve further by lazy loading all routes, but i'd rather leave that for after MVP

Screenshot 2026-03-16 at 16 11 54

### Checklist


application/docker/Dockerfile
```diff
@@ -85,6 +85,7 @@ RUN npm ci --audit=false --ignore-scripts
 COPY --link application/ui/tsconfig.json ./
 COPY --link application/ui/rsbuild.config.ts ./
 COPY --link application/ui/src/ src/
+COPY --link application/ui/public/index.html ./public/index.html
```
leoll2: I see that the `application/ui/public` folder already contains other files, namely the icons of the annotation tools. Those files are not already copied to the Docker image, but they should be, right? Actually, wouldn't it be better to move those icons into `src/api/assets/icons`? As for the `index.html`, would it make sense to move it to `ui/src/`? If not, just ignore my suggestion; I'm not that familiar with how the files are organized in ui folder, here I'm just wondering how to keep the Dockerfile as simple as possible.

jpggvilaca: Good points! I copied the whole `public` folder for simplicity and clarity. For the icons, i like it ,updated. As for the index, that is standard for the rspack/rsbuild (what we use to build and run the app) so i'd like to keep it there. Updated

leoll2 (approved):

camiloHimura (approved):

PR #5788, opened 2026-03-16
Media batch inference REST endpoint

### LINKED ISSUE #5618: Implement prediction endpoint
Refer to #5491 for the endpoint specifications.

warrkan:
This PR introduces intermedia MediaPredictionService which loads Geti data and passes it to InferenceServer, after inference it converts data, optionally stores results in database & filesystem.
Additionally it introduces REST endpoint for clients to send inference requests

Resolves #5618


application/backend/app/services/inference/inference_server.py
leoll2, line 46: `ttl` parameter is unused and can be removed from the function signature (and docstring)

application/backend/app/api/routers/media.py
```diff
@@ -511,3 +520,42 @@ def delete_media_annotation(
...
+            1
+            if media_request.range is None
+            else len(range(media_request.range.start_frame, media_request.range.end_frame, media_request.range.stride))
+            for media_request in request.media
+        ]
+    )
+    if items_count > inference_media_limit:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="Too many media items to predict. Please reduce the number of media or frame range size.",
```
leoll2: In the error message, I suggest reporting:
- the maximum number of media items to predict
- the fact that the limit can be increased through the INFERENCE_MEDIA_LIMIT env variable

application/backend/app/services/media_prediction_service.py
```diff
@@ -0,0 +1,208 @@
...
+
+    def _load_media(self, project: Project, request: MediaListPredictionRequest) -> LoadedMedia:
+        single_media: list[Image | VideoFrame] = []
+        video_frames: list[VideoFrame | NotAnnotatedVideoFrame] = []
+
+        media_ids: list[UUID] = [media_request.media_id for media_request in request.media]
+        media_list = self._media_service.get_media_by_ids(project_id=project.id, media_ids=media_ids)
+
+        for media_request in request.media:
+            media = next((media for media in media_list if media.id == media_request.media_id), None)
```
leoll2: The complexity here is quadratic, although it is probably not an issue since both lists have at most INFERENCE_MEDIA_LIMIT elements. Anyway, the complexity can be reduced to linear if you create an index dictionary out of `media_list`, and use it to retrieve `media` by it instead of `next(...)`.

application/backend/app/services/media_prediction_service.py
leoll2, line 1: I suggest to add some `logger.debug` statements and comments throughout this file. The prediction logic has many conditional branches, some logging would definitely help to debug potential issues.

application/backend/app/services/media_prediction_service.py
```diff
@@ -0,0 +1,208 @@
...
+                    annotations=prediction.prediction,
+                    user_reviewed=False,
+                    prediction_model_id=model_id,
+                )
+
+    def predict_media(
+        self,
+        project: Project,
+        request: MediaListPredictionRequest,
+    ) -> BatchInferenceResult:
```
leoll2, line 205: Missing docstring for public method

application/backend/app/services/media_prediction_service.py
```diff
@@ -0,0 +1,208 @@
...
+            if annotated_frame is not None:
+                video_frames.append(annotated_frame)
+            else:
+                video_frame = NotAnnotatedVideoFrame(video=video, frame_index=frame_index)
+                video_frames.append(video_frame)
+        return video_frames
+
+    def _load_media(self, project: Project, request: MediaListPredictionRequest) -> LoadedMedia:
+        single_media: list[Image | VideoFrame] = []
+        video_frames: list[VideoFrame | NotAnnotatedVideoFrame] = []
```
leoll2, line 82: Why is it necessary to have two lists?

application/backend/app/services/media_prediction_service.py
warrkan: Thanks! Fixed

application/backend/app/services/media_prediction_service.py
warrkan: Added more logging, indeed that would be helpful

application/backend/app/services/media_prediction_service.py
warrkan: Good one! Done

application/backend/app/api/routers/media.py
warrkan: Changed the message to be more useful

application/backend/app/services/media_prediction_service.py
warrkan: Because we treat them differently. Items from the second list should be always returned as video_id + frame_index combination event if they are already annotated and have their own ID's. Because they were requested as frame ranges.
While items from the first list are always returned using their own media ID's

application/backend/app/services/inference/inference_server.py
warrkan: Unfortunately we cannot omit it here, the inference monitor and inference server methods should have the same signature. Otherwise we can change it to `kwargs` and extract `ttl` from it in monitor

leoll2: True, I didn't consider that. Let's keep the code as it is.

leoll2 (dismissed):

leoll2 (approved):

PR #5769, opened 2026-03-12
Implement video support in import & export jobs

itallix:
### Summary

- [x] Added video export support in `GETI` format
- [x] Added video import support in `GETI` format
- [x] Separated `Sample` hierarchies for training and import/export use cases
- [x] Extended staged dataset metadata with `num_videos`, `num_annotated_frames`, `num_annotated_images`
- [x] Tests

### How to test

1. In the demo setup, upload video and annotate any frame. 
2. Try export/import project with annotated video.

**Expected:** videos with their annotations should be successfully exported/imported.

### Checklist

- [x] The PR title and description are clear and descriptive
- [x] I have manually tested the changes
- [x] All changes are covered by automated tests
- [x] All related issues are linked to this PR (if applicable)
- [ ] Documentation has been updated (if applicable)

PR #5791, opened 2026-03-17
Update MultiLabelClassificationSample to contain list of labels

AlbertvanHouten:
MultiLabelClassificationSample should hold a list of multi labels per annotation scene, not just one group of multi labels per annotation scene. 

### Checklist

### DIFF
--- a/application/backend/app/datumaro_converter/domain/samples.py
+++ b/application/backend/app/datumaro_converter/domain/samples.py
@@ -72,7 +72,7 @@ class MultilabelClassificationSample(BaseSample):
         confidence: Array of confidence scores for each label. Only for model predictions.
     """
 
-    label: NDArrayInt = label_field(dtype=pl.UInt8(), multi_label=True)
+    label: NDArrayInt = label_field(dtype=pl.UInt8(), multi_label=True, is_list=True)
     confidence: NDArrayFloat32 | None = numeric_field(dtype=pl.Float32(), is_list=True, semantic="confidence")
 
     @staticmethod


application/backend/app/datumaro_converter/domain/samples.py
```diff
@@ -72,7 +72,7 @@ class MultilabelClassificationSample(BaseSample):
         confidence: Array of confidence scores for each label. Only for model predictions.
     """
 
-    label: NDArrayInt = label_field(dtype=pl.UInt8(), multi_label=True)
+    label: NDArrayInt = label_field(dtype=pl.UInt8(), multi_label=True, is_list=True)
```
leoll2, line 75: After this change, I don't think I fully understand the difference between `multi_label` and `is_list` anymore. I thought that `multi_label=True` allows each annotation to have multiple labels, while `is_list=True` allows each image to have multiple annotations. What's the correct way to interpret these parameters?

leoll2:
FYI this PR needs to be rebased since #5769 was merged, which split the samples into separate definitions for training and dataset I/E.

application/backend/app/datumaro_converter/domain/samples.py
AlbertvanHouten: Its a bit confusing to understand but is_list should basically always be True, it indicates that there can be multiple shapes per annotation. If it is False, there would only be a single shape per annotation(like a global annotation from geti classic). With multi_label=True, each shape in the annotation can then have multiple labels.

leoll2: > is_list ... indicates that there can be multiple shapes per annotation

But classification doesn't allow multiple shapes per image, even in the multilabel case. In multilabel, we have 1 shape (the full image) with one or more labels assigned to it.

AlbertvanHouten: Sorry, you are right. I somehow confused myself with the issue you reported which has a list of labels imported in a classification task. MultiLabelClassification sample indeed should only only have multi_label is True.

PR #5813, opened 2026-03-19
Fix include_unannotated handling

### LINKED ISSUE #5806: Only annotated images exported though "all" speficifed
1. Create a detection project.
2. Import several images, annotate only two of them.
3. Export dataset. Select "Include media without annotations". Leave GETI as the selection.
4. Check the contents of the exported zip.

Two problems:
- (main) only two images in the exported zip,
- We shouldn't use all the capital letter for Geti.

1. The same applies to datasets exported to COCO and YOLO.
2. Regardless of the choice (you can deselect everything, for instance) in the export window, the annotated images are exported.

itallix:
### Summary

- [x] Remove annotation status filter when fetching items before export

### Checklist

### DIFF
--- a/application/backend/app/execution/dataset_export/export.py
+++ b/application/backend/app/execution/dataset_export/export.py
@@ -14,7 +14,7 @@ from sqlalchemy.orm import Session
 
 from app.datumaro_converter import SampleMode
 from app.execution.base import Execution, step
-from app.models import DatasetFormat, DatasetItemAnnotationStatus, ExportDatasetJobParams
+from app.models import DatasetFormat, ExportDatasetJobParams
 from app.services import DatasetRevisionService, DatasetService
 
 
@@ -61,7 +61,7 @@ class ExportDataset(Execution[ExportDatasetJobParams]):
                 dataset = self._dataset_service.get_dm_dataset(
                     project_id=export_params.project_id,
                     task=export_params.task,
-                    annotation_status=DatasetItemAnnotationStatus.REVIEWED,
+                    annotation_status=None,
                     sample_mode=SampleMode.IMPORT_EXPORT,
                 )
             else:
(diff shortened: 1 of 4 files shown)


application/backend/app/execution/dataset_export/export.py
```diff
@@ -61,7 +61,7 @@ def prepare_dataset(self, export_params: ExportDatasetJobParams) -> tuple[UUID,
                 dataset = self._dataset_service.get_dm_dataset(
                     project_id=export_params.project_id,
                     task=export_params.task,
-                    annotation_status=DatasetItemAnnotationStatus.REVIEWED,
+                    annotation_status=None,
```
leoll2: Removing the filtering here is correct, but you should also filter out all predictions while exporting.

Example: the dataset has an image with some bounding boxes and labels predicted by the inference pipeline; `annotation_data=[...]` (not empty), `user_reviewed=false`.  
- Current behavior: the image is not exported
- New behavior: the image is exported along with the prediction
- Expected behavior: the image is exported but without the prediction, i.e. unannotated

itallix: @AlbertvanHouten would it be possible to remove predictions for `keep_empty_samples`?

leoll2: The difference between annotations and predictions is a Geti concept, Datumaro shouldn't be aware of that.
What you need to do is build a Dataset where the annotation content (bboxes, polygons, labels, ...) is initialized conditionally to the `user_reviewed` status of the respective item: if `true` use the annotation stored in the database, if `false` ignore the prediction and initialize the annotation fields with empty arrays.

itallix: > The difference between annotations and predictions is a Geti concept, Datumaro shouldn't be aware of that. What you need to do is build a Dataset where the annotation content (bboxes, polygons, labels, ...) is initialized conditionally to the `user_reviewed` status of the respective item: if `true` use the annotation stored in the database, if `false` ignore the prediction and initialize the annotation fields with empty arrays.

Indeed, that makes sense.

leoll2 (approved):

PR #5836, opened 2026-03-20
Optimize Docker builds with shared caching and buildx support

AlbertvanHouten:
Overhauls the Docker build pipeline to dramatically improve build speed through better layer caching, a smaller build context, and remote cache backed by GHCR. The CI workflow is simplified by extracting the Docker build logic into a reusable composite action.

### Key changes

- **Two-step dependency install in Dockerfile** — third-party deps are installed in a separate layer keyed only on lock files, so source-code changes skip the expensive `uv sync` step. App code and UI assets are copied last.
- **Cache mounts** — `apt` and `uv` caches use `--mount=type=cache` instead of download-then-delete, speeding up both dependency resolution and system package installation.
- **New reusable `docker-build` composite action** — encapsulates Buildx setup, registry cache (zstd-compressed on GHCR), and smart driver selection (`docker` for local builds, `docker-container` when pushing).
- **CI workflow rewrite** — images are pushed to GHCR instead of uploaded as multi-GB artifacts; smoke tests (health check + UI check) and image-size PR comments run inline per matrix job, eliminating a separate `comment-sizes` job.
- **Hatchling + `VERSION` file** — app version is now dynamic, read from a single `application/VERSION` file.

[Before](https://github.com/open-edge-platform/training_extensions/actions/runs/23424590664?pr=5850):
image

[After](https://github.com/open-edge-platform/training_extensions/actions/runs/23550374358?pr=5836):
image

### Checklist


application/Justfile
```diff
@@ -19,14 +19,42 @@ logs-volume := application-name + "-logs"
 # Build Geti Docker image for a specific training backend
 [arg("accelerator", long="accelerator", short="a", pattern="cpu|xpu|cuda")]
 [arg("tag", long="tag", short="t")]
-build-image accelerator="cpu" tag="latest": install-uv
-    @echo "Building docker image for accelerator: {{ accelerator }}"
-    @docker build \
+[arg("extra-tags", long="extra-tags", help="Space-separated list of additional tags to apply to the image")]
```
leoll2: Instead of `tag` and `extra-tags`, it is cleaner to have a single parameter `tags` that accepts one or more space-separated tags, similarly to how we do for `run-image --volumes`.

AlbertvanHouten:
> ## Docker Image Sizes
> ### CPU
> Image	Size
> ghcr.io/open-edge-platform/training_extensions/geti-cpu:pr-5836	1.10G
> ### CUDA
> Image	Size
> ghcr.io/open-edge-platform/training_extensions/geti-cuda:pr-5836	5.62G
> ### XPU
> Image	Size
> ghcr.io/open-edge-platform/training_extensions/geti-xpu:pr-5836	3.43G

The slight increase in docker image size is due to setting `UV_COMPILE_BYTECODE=1` in the dockerfile. This flag tells uv to compile every .py file into .pyc bytecode at install time. `UV_COMPILE_BYTECODE=1` makes container startup faster (Python doesn't need to compile bytecode on first import), but makes the image larger. Without it, Python compiles bytecode lazily on first import and caches it at runtime (if writable), or just compiles in-memory each time so either way the running image will inflate to this size.

.github/workflows/build.yaml
```diff
@@ -46,226 +43,48 @@ jobs:
...
-          uv-version-file: application/backend/pyproject.toml # Install the uv version specified in pyproject.toml
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Extract metadata (tags, labels) for Docker
         id: meta
         uses: docker/metadata-action@c299e40c65443455700f0fdfc63efafe5b349051 # v5.10.0
         with:
+          images: ghcr.io/${{ github.repository }}/geti-${{ matrix.device }}
```
AlexanderBarabanov: I propose to use `ghcr.io/open-edge-platform/geti-${{ matrix.device }}` - repo name in container image name is redundant.

To map container image to the repo - we can use `LABEL` in Dockerfile - see this example https://github.com/open-edge-platform/physical-ai-studio/blob/main/application/docker/Dockerfile#L151

application/docker/Dockerfile
```diff
@@ -151,100 +190,39 @@ FROM runtime-base AS geti-cpu
...
-    && apt-get clean
+      nvidia-open=590.48.01-1
 
 USER ${USER_NAME}
 
 COPY --from=cuda-base --chown=${UID}:${UID} /application/backend/.venv /application/backend/.venv
 
-USER root
-
-# CUDA EULA Compliance Cleanup
```
AlexanderBarabanov, line 180: Yes, we may remove this logic as it was implemented incorrectly.
It shall be implemented in a **separate** stage, not in a final stage as in this case non-distributable components still live in distributed container image (but not in final layer) ,see e.g. Anomalib implementation https://github.com/open-edge-platform/anomalib/blob/main/application/docker/Dockerfile#L122

For motivation, please see https://docs.nvidia.com/cuda/eula/index.html#attachment-a

.github/actions/docker-build/action.yaml
```diff
@@ -0,0 +1,80 @@
...
+        # Disable all caching (local and remote) when use-cache=false to
+        # ensure reproducible builds for RC/release without any cached layers.
+        no-cache: ${{ inputs.use-cache != 'true' }}
+        # The caller must log in to ghcr.io before invoking this action.
+        # Cache reads always work (public packages); writes require
+        # packages:write which fork PRs do not have.
+        # Cache is disabled for RC/release builds (use-cache=false)
+        # to prevent cache poisoning and ensure reproducible builds.
+        cache-from: ${{ inputs.use-cache == 'true' && format('type=registry,ref=ghcr.io/{0}:cache-{1}', github.repository, inputs.device) || '' }}
+        cache-to: ${{ inputs.cache-write == 'true' && inputs.use-cache == 'true' && format('type=registry,ref=ghcr.io/{0}:cache-{1},mode=max,compression=zstd,image-manifest=true,ignore-error=true', github.repository, inputs.device) || '' }}
```
AlexanderBarabanov: Can we cache under each container image, via

```
cache-from: ${{ inputs.use-cache == 'true' && format('type=registry,ref=ghcr.io/open-edge-platform/geti-{0}:buildcache', inputs.device) || '' }}
cache-to: ${{ inputs.cache-write == 'true' && inputs.use-cache == 'true' && format('type=registry,ref=ghcr.io/open-edge-platform/geti-{0}:buildcache,mode=max', inputs.device) || '' }}
```

In this case, there will be a container image with a `buildcache` tag (e.g. `ghcr.io/open-edge-platform/geti-xpu:buildcache`) per device instead of separate container image that may potentially confuse consumers (https://github.com/open-edge-platform/training_extensions/pkgs/container/training_extensions)


Example from `physical-ai-studio` repo - https://github.com/open-edge-platform/physical-ai-studio/pkgs/container/physical-ai-studio-xpu

Image

.github/workflows/build.yaml
```diff
@@ -46,226 +43,49 @@ jobs:
               - .github/**
 
   build:
-    name: Build and upload docker image (${{ matrix.device }})
+    name: Build and push docker image (${{ matrix.device }})
```
AlexanderBarabanov, line 45: Just to double-check - do we really need to upload container images for every PR?
Or we can e.g. build container for every PR but push daily from the main branch?

.github/workflows/build.yaml
```diff
@@ -46,226 +43,49 @@ jobs:
...
       - name: Extract metadata (tags, labels) for Docker
         id: meta
         uses: docker/metadata-action@c299e40c65443455700f0fdfc63efafe5b349051 # v5.10.0
         with:
+          images: ghcr.io/${{ github.repository }}/geti-${{ matrix.device }}
           tags: |
             type=sha
             type=ref,event=branch
             type=ref,event=pr
             type=semver,pattern={{version}}
```
AlexanderBarabanov, line 86: How should we retrieve the version value?

I propose reusing the `physical-ai-studio` repository approach - create a `VERSION` file in the application root as a single source of truth as currently, versions are scattered across `application/ui/src-tauri/Cargo.toml`, `application/backend/pyproject.toml` and `application/ui/package.json`.

For container images built on pull requests, we may generate version as `VERSION-dev-<short-sha>`, e.g. `0.1.0-dev-0d3d7664`. 
Here is an example https://github.com/open-edge-platform/physical-ai-studio/blob/main/.github/workflows/docker-publish.yml#L34

Alternatively, for pull requests, we could tag images with just the PR number and short SHA instead.

AlexanderBarabanov (commented):
Thanks for the effort!

.github/workflows/build.yaml
AlbertvanHouten: Fixed to only push when merging

.dockerignore
leoll2, line 1: We already have `.dockerignore` files under `application/` and `library/`. I don't mind having a single unified `.dockerignore` at the root level, but then you should remove the nested ones.

.dockerignore
```diff
@@ -0,0 +1,37 @@
...
+!application/ui/src
+!application/ui/public
+!application/ui/packages
+!application/ui/package.json
+!application/ui/package-lock.json
+!application/ui/tsconfig.json
+!application/ui/rsbuild.config.ts
+
+# Nginx config
+!application/docker/nginx.conf
```
leoll2: This file doesn't exist

.dockerignore
```diff
@@ -0,0 +1,37 @@
...
+# Library – only what uv needs to resolve deps and build the otx wheel
+!library/pyproject.toml
+!library/MANIFEST.in
+!library/src/otx
+
+# Exclude non-essential files that sneak in via the above whitelists
+**/__pycache__
+**/.pytest_cache
+**/.mypy_cache
+**/.ruff_cache
```
leoll2, line 30: You can also exclude `.import_linter_cache` (created by a pre-commit hook) and `otx-workspace*` (created by training jobs when running the app without Docker)

application/VERSION
leoll2, line 1: If you store the version in a file, please update `pyproject.toml` accordingly.

Something like:
```toml
[project]
name = "geti"
dynamic = ["version"]

[tool.hatch.version]
path = "../VERSION"
```

application/docker/Dockerfile
```diff
@@ -109,31 +152,31 @@ ENV DATA_DIR="/application/data"
...
+RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
+    --mount=type=cache,target=/var/lib/apt/lists,sharing=locked \
+    apt-get update \
     && apt-get install -y --no-install-recommends \
         git=1:2.47.3-0* \
         libgl1=1.7.0-1+b2 \
         libglib2.0-0t64=2.84.4-3~deb13u2 \
-        libglx-mesa0=25.0.7-2 \
-    && rm -rf /var/lib/apt/lists/* \
-    && apt-get clean
```
leoll2, line 128: Is it not worth it to cleanup the `apt/lists`? Docker documentation mentions it among the best practices:
https://docs.docker.com/build/building/best-practices/#apt-get

> when you clean up the apt cache by removing /var/lib/apt/lists it reduces the image size, since the apt cache isn't stored in a layer.

.github/workflows/build.yaml
```diff
@@ -46,226 +43,56 @@ jobs:
...
-        working-directory: application
-        run: |
-          for TAG in $TAGS ; do
-              just build-image --accelerator $DEVICE --tag $TAG
-          done
-        env:
-          TAGS: ${{ steps.meta.outputs.tags }}
-          DEVICE: ${{ matrix.device }}
-
-      - name: Run simple checks on the built image
```
leoll2, line 101: These checks are very important to ensure that our PRs don't break the Docker deployment, we shouldn't remove them. I see that the Docker image sanity checks usually take 2-2.5 minutes to complete, which is more than acceptable in my opinion.

.github/workflows/build.yaml
```diff
@@ -46,226 +43,56 @@ jobs:
...
-        uses: ishworkh/container-image-artifact-upload@5d71a2417f0576fa11fe770fb04ece58c4587714
-        with:
-          image: "geti-${{ matrix.device }}:${{ steps.first-tag.outputs.tag }}"
-          retention_days: "3"
-        continue-on-error: true
-        env:
-          FIRST_TAG: ${{ steps.first-tag.outputs.tag }}
-
-  comment-sizes:
-    name: Comment PR with all image sizes
```
leoll2, line 207: This step is also important and shouldn't be removed.

.github/workflows/build.yaml
AlbertvanHouten: The checks themselves aren't the problem but they need to run the full image which isn't build completely locally anymore. Loading the image locally adds a lot of overhead, especially for cuda image (~8 minutes). I'll see if we can run health checks on just the CPU image.

.github/workflows/build.yaml
AlbertvanHouten: This has the same problem with loading the full image. I don't think we should monitor it on every PR. Maybe we can just run it if the dockerfile changed or make it part of a release check.

.github/workflows/build.yaml
leoll2: > the full image isn't build completely locally anymore

Can you elaborate on this part? Is it because `push: true` so the image is sent directly to GHCR without being saved locally? If so, can't you push the image in a separate step, _after_ running the health checks?

AlbertvanHouten: nevermind, I've found a workaround. The slowness will still be there in the merge queue but that shouldn't be as much of a problem.

.github/workflows/build.yaml
AlbertvanHouten: added again.

application/docker/Dockerfile
AlbertvanHouten: Its not necessary anymore with the cache mount as they are not part of the image layer. Buildkit stores them separately.

.dockerignore
AlbertvanHouten: Will remove them. Those other docker ignores never worked since the build context started from the repo root

.github/workflows/build.yaml
```diff
@@ -46,226 +42,142 @@ jobs:
...
-              echo "❌ Health check failed after 60 seconds"
-              exit 1
-          fi
-
-          EXPECTED_TITLE="Geti"
-          echo "Checking HTML title..."
-
-          HTML_CONTENT=$(curl -sf $TEST_URL)
-          if echo "$HTML_CONTENT" | grep -q "<title>.*$EXPECTED_TITLE.*</title>"; then
-              echo "✅ HTML title check passed - found: $EXPECTED_TITLE"
```
leoll2, line 144: Can you also restore this part of the health check? It validates that the UI landing page is accessible at `localhost:7860` and loads as expected.

AlbertvanHouten: added again

leoll2 (dismissed):

leoll2 (approved):

AlexanderBarabanov (approved):

PR #5850, opened 2026-03-23
Delete deleted labels from annotations

### LINKED ISSUE #5773: Update annotations after label removal
When the user removes a label, the existing annotations should be updated accordingly by removing that label where present. If a shape remains without a label, that shape should be removed as well. Likewise, if an image remains without shapes, it becomes empty-labeled (`annotation_data=[]`). For tasks that don't support empty label, namely multiclass classification, set it as unannotated (`annotation_data=None` with `user_reviewed=false`).

The `dataset_items_labels` table can be used to efficiently find the annotated media that should be updated.

For simplicity, the annotation update can be done synchronously with the project labels update request, even if this may increase the endpoint latency when the dataset is large. We can optimize this at a later time.

JortBergfeld:
### Summary

After deleting a label, we now also delete this label from any annotation that contains this label. If no labels remain, the annotation is removed as well. If no annotation remains, the annotation data is set to None if a project does not support the empty label, otherwise it is left as an empty list.

Because the `DatasetService` depends on `LabelService` and `MediaService`, the code is implemented in the `LabelService` completely to avoid weird circular dependencies.

Resolves #5773 

### How to test

New integration tests have been added

### Checklist

### DIFF
--- a/application/backend/app/services/label_service.py
+++ b/application/backend/app/services/label_service.py
@@ -179,8 +179,6 @@ class LabelService(BaseSessionManagedService):
         # Find all dataset items that reference this label
         affected_items = dataset_item_repo.find_items_by_label_id(label_id_str)
 
-        is_multiclass = task.task_type is TaskType.CLASSIFICATION and task.exclusive_labels
-
         for item in affected_items:
             if item.annotation_data is None:
                 continue
@@ -200,7 +198,7 @@ class LabelService(BaseSessionManagedService):
                     user_reviewed=item.user_reviewed,
                     prediction_model_id=item.prediction_model_id,
                 )
-            elif is_multiclass:
+            elif task.is_multiclass:
                 dataset_item_repo.delete_annotation_data(obj_id=item.id)
             else:
                 dataset_item_repo.set_annotation_data(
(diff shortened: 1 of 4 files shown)


application/backend/app/services/label_service.py
```diff
@@ -163,7 +163,57 @@ def _update_label(
...
+        - For multi-class classification (which doesn't support empty labels),
+          set annotation_data=None and user_reviewed=False.
+        """
+        dataset_item_repo = DatasetItemRepository(project_id=str(project_id), db=self.db_session)
+        label_id_str = str(label_id)
+
+        # Find all dataset items that reference this label
+        affected_items = dataset_item_repo.find_items_by_label_id(label_id_str)
+
+        is_multiclass = task.task_type is TaskType.CLASSIFICATION and task.exclusive_labels
```
itallix: I suggest we move it to the `Task` as a new property:

```python
@property
def is_multiclass(self) -> bool:
    return self.task_type is TaskType.CLASSIFICATION and self.exclusive_labels
```

This logic can potentially be used in multiple places.

itallix (approved):

PR #5873, opened 2026-03-23
Webrtc pause resume behavior

### LINKED ISSUE #5794: Pause / resume WebRTC stream
In the Inference page, the user can monitor the predictions in real time through a WebRTC video stream. However, there are currently no controls to stop/resume the stream; this would be a useful feature because the stream consumes resources (cpu, memory, network bandwidth) so the user may want to close the connection.

jpggvilaca:
### Summary

* Add play/pause behavior youtube-like
* Update test and locator

Closes https://github.com/open-edge-platform/training_extensions/issues/5794

Screenshot 2026-03-23 at 14 10 29
Screenshot 2026-03-23 at 14 53 43

### Checklist

### DIFF
--- a/application/ui/src/features/inference/stream/stream-container.tsx
+++ b/application/ui/src/features/inference/stream/stream-container.tsx
@@ -3,7 +3,7 @@
 
 import { useEffect, useState } from 'react';
 
-import { Button, Flex, Loading, toast, View } from '@geti/ui';
+import { Flex, Loading, toast, View } from '@geti/ui';
 import { Pause, Play } from '@geti/ui/icons';
 
 import { Stream } from './stream';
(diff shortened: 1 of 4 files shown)


application/ui/src/features/inference/stream/stream-container.tsx
```diff
@@ -23,33 +27,52 @@ export const StreamContainer = () => {
 
     return (
         <View gridArea={'canvas'} overflow={'hidden'} maxHeight={'100%'}>
-            {status === 'idle' && (
-                <div className={classes.canvasContainer}>
-                    <View backgroundColor={'gray-200'} width='90%' height='90%'>
+            <div className={classes.canvasContainer} onClick={isConnected ? stop : undefined}>
```
dwesolow: why we need this handler? I don't like that we mix wrapper that only adds `isConnected` handler and below granular handlers.

jpggvilaca: We do need it because we want the whole canvas to be clickable. Or else the user can only click on the button.

dwesolow: The whole canvas should be clickable only to stop? why?
If we want to have the whole canvas to be treated as a button we don't need nested buttons and `event.stopPropagation` etc.

jpggvilaca: We do want. We want the user to click on the whole canvas either to stop or resume. I think it's a better experience, just like Youtube. No?

dwesolow: ok, so we don't need nested button, correct?

jpggvilaca: True, now we dont anymore :) , removed!

dwesolow: There is still button for play

dwesolow (dismissed):

dwesolow (dismissed):

dwesolow (approved):

application/ui/src/features/inference/stream/stream-container.tsx
jpggvilaca: Removed

PR #5889, opened 2026-03-24
Model screen updates

jpggvilaca:
### Summary

This PR fixes a bunch of issues with model screen:
* Opening the model metrics tab was slow (UI hanged for like 3 seconds) because we render a lot of graphs, we will now render them lazily
* Improve Box component style reusability across all 4 tabs. Also moved the scrollbar to the content and not the box itself
* Pressing 'delete' on model row will now delete its files, as it should (bugfix), so i also updated the cases where the files are deleted (variants and metrics are gone)
* Add UX improvement to let the user know the model download started

Screenshot 2026-03-23 at 16 30 50
Screenshot 2026-03-24 at 09 20 47
Screenshot 2026-03-24 at 10 09 19
Screenshot 2026-03-24 at 10 11 29
Screenshot 2026-03-24 at 10 10 31

### Checklist


application/ui/src/features/models/model-listing/model-metrics/model-metrics-graphs.component.tsx
```diff
@@ -18,18 +20,64 @@ type GraphData = {
     data: MetricGraphPoint[];
 };
 
+const useProgressiveList = <T,>(items: T[]): T[] => {
```
jpggvilaca: this is a bit complex imho but i didnt figure out a better way. This is basically a fancier version of using setTimeout to delay rendering. I tried `useDeferredValue` and `useTransition`, none worked.

application/ui/src/features/models/model-listing/model-metrics/model-metrics-graphs.component.tsx
```diff
@@ -18,18 +20,64 @@ type GraphData = {
...
+            })),
+        [trainingMetrics]
+    );
+
+    const visibleGraphs = useProgressiveList(graphs);
 
     return (
         <Flex width={'100%'} direction={'row'} gap={'size-300'} wrap>
-            {graphs.map((graph) => (
+            {visibleGraphs.map((graph) => (
```
dwesolow: We should use virtualizer for this.

application/ui/src/features/models/model-listing/model-variants/model-variant-tabs.component.tsx
```diff
@@ -17,10 +17,10 @@ type ModelVariantsTabsProps = {
 };
 
 export const ModelVariantsTabs = ({ model }: ModelVariantsTabsProps) => {
-    if (isEmpty(model.variants)) {
+    if (isEmpty(model.variants) || model.files_deleted) {
```
dwesolow, line 20: Is there any indicator on the model row that says that model files are removed?
Would it be a case that user removed model files but wanted to check model details? If no then I'd disable the model row so user cannot open its details.

jpggvilaca: The user can still check the training parameters and training datasets

application/ui/src/features/models/model-listing/model-metrics/model-metrics-graphs.component.tsx
jpggvilaca: I think virtualiser is better for large lists no? In this case the amount of graphs is not high, but its information is

dwesolow: imo that's the same case

jpggvilaca: It's different because here you have less elements, but heavier ones. But I can at least try. I'll let you know

jpggvilaca: It did not solve the performance issue

application/ui/src/features/models/model-listing/model-metrics/model-metrics-graphs.component.tsx
```diff
@@ -18,18 +20,64 @@ type GraphData = {
...
+        return () => {
+            if (frameRef.current !== null) {
+                cancelAnimationFrame(frameRef.current);
+                frameRef.current = null;
+            }
+        };
+    }, [items]);
+
+    return visible;
+};
```
dwesolow: This solution smells. In geti classic we had more charts and the performance was good enough that we didn't have virtualization there. Maybe we should check more deeply what causes that hang?

jpggvilaca: It's really the chart that is heavy. Even if we set the animations to false. No other solution fixed the performance issue

dwesolow: What about using `useIsVisible` or `LazyLoadSection` that I used for data management parameters in the training dialog?
In latest react there is `Activity` component that might be useful in such cases.

jpggvilaca: I forgot about `useIsVisible`, it works as well! Thanks, updated

application/ui/src/features/models/training-logs/training-logs-dialog.component.tsx
```diff
@@ -58,7 +58,7 @@ const HistoricalModelLogs = ({ modelId }: { modelId: string }) => {
 
 export const TrainingLogsDialog = ({ jobId, modelId }: TrainingLogsDialogProps) => {
     const dialogContainer = useDialogContainer();
-    const { downloadModelLogs, isDownloading } = useDownloadModelLogs(modelId);
+    const { downloadModelLogs, isDownloading } = useDownloadModelLogs(String(modelId));
```
dwesolow, line 61: `modelId` should be passed to `downloadModelLogs` in line 71 :).

dwesolow (approved):

application/ui/src/features/models/training-logs/training-logs-dialog.component.tsx
jpggvilaca: Will update on my next one :)

PR #5902, opened 2026-03-24
Add a work-around for OV inference

### LINKED ISSUE #5877: OpenVINO exported models produce 0 accuracy
**Describe the bug**
Component: otx.backend.native / model export pipeline
Severity: Critical
Affected models: All models across all supported tasks (detection, instance segmentation, classification)

Description:
After training any OTX model and exporting it to OpenVINO IR format, the exported model produces 0 accuracy on evaluation. The PyTorch model evaluates correctly with expected metrics, but the OpenVINO IR model outputs incorrect predictions, resulting in 0.0 across all evaluation metrics.

from otx.backend.native.engine import OTXEngine

# Works with any recipe / task combination
engine = OTXEngine(
    recipe="src/otx/recipe/<task>/<model>.yaml",
    work_dir="work_dir",
)

# 1. Train — works fine
engine.train(data_root="path/to/dataset")

# 2. Test PyTorch model — produces valid metrics
engine.test()  # e.g., mAP ~0.45 or accuracy ~0.85

# 3. Export to OpenVINO
engine.export()

# 4. Test OpenVINO model — produces 0 accuracy
engine.test(checkpoint="work_dir/exported_model.xml")  # All metrics = 0.0

Expected behavior:
The OpenVINO exported model should produce accuracy comparable to the PyTorch model (within a small tolerance, typically <1-2% mAP drop).

Actual behavior:
The OpenVINO model produces 0.0 accuracy on all metrics (mAP@50, mAP@50:95, etc.).

Possible root causes:

Incorrect pre/post-processing embedding during export — RF-DETR uses non-standard input sizes (multiples of 24, e.g., 624×624) and custom normalization. The export pipeline may not correctly embed these preprocessing parameters into the IR model.

**Environment:**

- OS: 
- Framework version: 
- Python version:
- OpenVINO version:
- CUDA/cuDNN version:
- GPU model and memory:

kprokofi:
### Summary

OTX now exports OpenVINO IR models with mean/std values in 0-1 scale (e.g. mean=0.485, std=0.229) instead of the old uint8 scale (e.g. mean=123.675, std=58.395).

The application inference server (InferenceServer, ActiveModelService, InferenceWorker) was unconditionally passing raw uint8 [0–255] images to ModelAPI with a default dtype=int (u8) IR input tensor, causing wildly incorrect normalization

This resulted in broken predictions for all models exported with the new OTX normalization format.

Solution
Added automatic IR format detection at model load time. The system inspects the embedded mean_values/scale_values from the IR <rt_info> and decides once — zero runtime overhead per frame.

resolves https://github.com/open-edge-platform/training_extensions/issues/5877

### Checklist

### DIFF
--- a/application/backend/app/services/active_model_service.py
+++ b/application/backend/app/services/active_model_service.py
@@ -7,7 +7,7 @@ from pathlib import Path
 from uuid import UUID
 
 from loguru import logger
-from model_api.adapters import OpenvinoAdapter, create_core
+from model_api.adapters import create_core
 from model_api.models import Model
 
 from app.db.engine import get_db_session
@@ -15,7 +15,7 @@ from app.models.model_activation import ModelActivationState
 from app.models.model_revision import ModelFormat, ModelPrecision
 from app.repositories import ModelRevisionRepository, ModelVariantRepository
 from app.repositories.active_model_repo import ActiveModelRepo
-from app.utils.ir_format import FP32OpenvinoAdapter, needs_float32_input
+from app.utils.ir_format import FP32OpenvinoAdapter
 
 MODELAPI_NSTREAMS = os.getenv("MODELAPI_NSTREAMS", "2")
 
@@ -26,7 +26,6 @@ class LoadedModel:
     model_variant_id: UUID
     model: Model
     device: str
-    float32_input: bool  # True → InferenceWorker must scale images to [0,1] float32
 
 
 @dataclass(frozen=True)
@@ -161,15 +160,8 @@ class ActiveModelService:
                     variant_id=active_variant_id,
                     extension="bin",
                 )
-                use_float32 = needs_float32_input(model_xml_path)
                 ie = create_core()
-                adapter_cls = FP32OpenvinoAdapter if use_float32 else OpenvinoAdapter
-                logger.info(
-                    "IR format detected: {} (float32_input={})",
-                    model_xml_path.name,
-                    use_float32,
-                )
-                adapter = adapter_cls(
+                adapter = FP32OpenvinoAdapter(
                     ie,
                     str(model_xml_path),
                     device=device,
@@ -185,6 +177,5 @@ class ActiveModelService:
                 model=mapi_model,
                 model_variant_id=active_variant_id,
                 device=device,
-                float32_input=use_float32,
             )
         return self._loaded_model
--- a/application/backend/app/services/active_model_service.py
+++ b/application/backend/app/services/active_model_service.py
@@ -15,21 +15,9 @@ from app.models.model_activation import ModelActivationState
 from app.models.model_revision import ModelFormat, ModelPrecision
 from app.repositories import ModelRevisionRepository, ModelVariantRepository
 from app.repositories.active_model_repo import ActiveModelRepo
-from app.utils.ir_format import needs_float32_input
+from app.utils.ir_format import FP32OpenvinoAdapter, needs_float32_input
 
 MODELAPI_NSTREAMS = os.getenv("MODELAPI_NSTREAMS", "2")
-
-
-class _FP32OpenvinoAdapter(OpenvinoAdapter):
-    """OpenvinoAdapter that forces float32 input tensors.
-
-    Used when the IR embeds mean/std in the 0-1 scale (new OTX format).
-    Overrides ``embed_preprocessing`` so ModelAPI sets the input tensor to f32.
-    """
-
-    def embed_preprocessing(self, *args, **kwargs) -> None:
-        kwargs["dtype"] = float
-        super().embed_preprocessing(*args, **kwargs)
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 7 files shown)


kprokofi:
This solution doesn't resolve problem with 16bit images support in Geti and ModelAPI
It can be as temporary workaround to support inference with old pipeline

application/backend/app/utils/ir_format.py
```diff
@@ -0,0 +1,69 @@
...
+        ``True``  → new 0-1 scale → use ``_FP32OpenvinoAdapter`` + scale to float32.
+        ``False`` → old uint8 scale (or no normalisation) → use default adapter + uint8 input.
+    """
+    tree = ET.parse(str(model_xml_path))
+    root = tree.getroot()
+    if root is None:
+        return False
+    rt_info = root.find("rt_info")
+    if rt_info is None:
+        return False
```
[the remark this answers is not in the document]
kprokofi: Done

application/backend/app/services/active_model_service.py
```diff
@@ -7,23 +7,38 @@
...
+    Overrides ``embed_preprocessing`` so ModelAPI sets the input tensor to f32.
+    """
+
+    def embed_preprocessing(self, *args, **kwargs) -> None:
+        kwargs["dtype"] = float
+        super().embed_preprocessing(*args, **kwargs)
+
+
 @dataclass(frozen=True)
 class LoadedModel:
```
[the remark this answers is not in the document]
kprokofi: Done

library/src/otx/backend/openvino/models/base.py
```diff
@@ -42,6 +43,69 @@
...
+    the 0-1 normalisation scale used by new OTX exports.  Raises
+    ``ValueError`` if the IR still stores mean/scale in the 0-255 range.
+    """
+
+    # Values above this threshold indicate uint8 (0-255) scale rather than 0-1 scale.
+    _UINT8_SCALE_THRESHOLD = 1.0
+
+    def embed_preprocessing(self, *args, **kwargs) -> None:
+        mean = kwargs.get("mean") or (args[5] if len(args) > 5 else None)
+        scale = kwargs.get("scale") or (args[6] if len(args) > 6 else None)
```
leoll2:
```suggestion
        mean = kwargs.get("mean")
        scale = kwargs.get("scale")
```
I would remove the fallback path, it's very fragile. In fact, [mean](https://github.com/open-edge-platform/model_api/blob/517a889bb2e46124e2123fbd557bfc2816429a7c/src/model_api/adapters/openvino_adapter.py#L406) doesn't seem to be the 6th argument of `embed_preprocessing`.

If `mean` or `scale` are missing, I think it's safer to just raise an exception.

application/backend/app/services/inference/inference_server.py
```diff
@@ -151,7 +169,12 @@ def infer_batch(self, labels: list[Label], inputs: list[BatchInferenceInput]) ->
                 raise RuntimeError("No model loaded for inference")
             logger.debug("Running inference on batch of {} inputs", len(inputs))
 
-            input_data = [input.data for input in inputs]
+            if self._loaded_model.float32_input:
+                # New OTX format: IR mean/std in 0-1 scale → pre-scale images to float32 [0, 1].
+                input_data = [inp.data.astype(np.float32) / 255.0 for inp in inputs]
+            else:
+                # Old OTX format: IR mean/std in uint8 (0-255) scale → pass raw uint8 images.
```
leoll2: Do we actually need this? Since we haven't released 3.0 yet, can't we safely assume that all models created in Geti 3.0+ use the new [0-1] scale?

kprokofi: We can...
But it means no models from Geti 2.0 will be supported. Is it OK? If so -> I will remove it

leoll2: Yes it's okay, we are not providing backward compatibility with 2.x models.

application/backend/app/utils/ir_format.py
```diff
@@ -0,0 +1,101 @@
...
+        return _orig_pad(arg, pads_begin, pads_end, pad_mode, arg_pad_value, name)
+
+    _opset.pad = _pad_with_type_cast  # pyrefly: ignore[bad-assignment]
+    try:
+        embed_fn(*args, **kwargs)
+    finally:
+        _opset.pad = _orig_pad
+
+
+def needs_float32_input(model_xml_path: str | Path) -> bool:
```
leoll2: This function can be removed too, under the same assumption that all models now use [0-1]

application/backend/app/services/inference/inference_server.py
kprokofi: Done

library/src/otx/backend/openvino/models/base.py
kprokofi: Done

application/backend/app/utils/ir_format.py
kprokofi: Done

leoll2 (approved):

PR #5904, opened 2026-03-25
Show picker only if the subset is unassigned

jpggvilaca:
### Summary

* Previously, the picker would also show, and if the user tried to assign to a different subset, the server will throw an error. Now, we just show a static subset, to not let the user change it. Not sure if it's the best solution but it's better than what we had, since it would throw an error 100% of the time (by design).

Screenshot 2026-03-25 at 09 02 15
Screenshot 2026-03-25 at 09 02 20

### Checklist

### DIFF
--- a/application/ui/src/features/dataset/media-preview/bottom-toolbar/bottom-toolbar.component.tsx
+++ b/application/ui/src/features/dataset/media-preview/bottom-toolbar/bottom-toolbar.component.tsx
@@ -5,6 +5,7 @@ import { Flex, Grid, Item, Key, Picker, Tag, Text } from '@geti/ui';
 import { Accept, Search } from '@geti/ui/icons';
 import { clsx } from 'clsx';
 import { useProjectIdentifier } from 'hooks/use-project-identifier.hook';
+import { capitalize } from 'lodash-es';
 
 import { $api } from '../../../../api/client';
 import { DatasetSubset, Media } from '../../../../constants/shared-types';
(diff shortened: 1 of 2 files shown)


application/ui/src/features/dataset/media-preview/bottom-toolbar/bottom-toolbar.component.tsx
```diff
@@ -92,16 +95,20 @@ export const BottomToolbar = ({ mediaItem, hideHotkeys }: BottomToolbarProps) =>
...
+                                    placeholder={'Select subset'}
+                                    aria-label={'Select subset'}
+                                    onSelectionChange={handleSubsetChange}
+                                >
+                                    <Item key={'validation'}>Validation</Item>
+                                    <Item key={'testing'}>Testing</Item>
+                                    <Item key={'training'}>Training</Item>
+                                </Picker>
+                            ) : (
+                                <Tag withDot={false} text={capitalize(String(currentSubset))} />
```
dwesolow, line 110: I thought about displaying Picker with disabled state but I think your current solution is better.

dwesolow (dismissed):

application/ui/src/features/dataset/media-preview/bottom-toolbar/bottom-toolbar.component.tsx
jpggvilaca: Yeah i thought about that too first, but then i remember we wanted to improve our overall UX by not displaying disabled buttons

dwesolow (approved):

PR #5908, opened 2026-03-25
Combine DEIM recipes with the base models

kprokofi:
### Summary

The DEIMDFine and DEIMV2 model classes are architecturally identical to DFine/DINOv3-DETR with two additions: the DEIMCriterion loss (which adds a Matchability-Aware Loss on top of D-FINE's existing losses) and the AugmentationSchedulerCallback (which drives a 3-phase strong→light→no-aug schedule during training). Previously these were exposed as separate model entries (deim-d-fine-m/l/x), duplicating the UX surface for what is essentially the same model with an optional training enhancement.

### Checklist


application/backend/app/api/schemas/model_architecture.py
```diff
@@ -48,11 +48,13 @@ class ModelArchitectures(BaseModel):
...
                         "task": "detection",
-                        "name": "DEIM-D-FINE-M",
-                        "description": "DEIM is an advanced training framework designed to enhance the matching"
-                        " mechanism in DETRs, enabling faster convergence and improved accuracy.",
+                        "name": "D-FINE-M",
+                        "description": "D-FINE is a powerful real-time object detector that redefines the bounding box"
+                        " regression task in DETRs as Fine-grained Distribution Refinement (FDR)."
+                        " Combined with the DEIM adaptive augmentation scheduling framework"
+                        " (enabled by default), it achieves outstanding performance with faster convergence.",
                         "capabilities": {"xai": True, "tiling": True},
```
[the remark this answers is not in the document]
kprokofi: Done

library/src/otx/tools/converter.py
```diff
@@ -673,6 +677,23 @@ def get_callback_idx(callbacks: list, name: str) -> int:
                 return idx
         return -1
 
+    @staticmethod
+    def _disable_deim_framework(config: dict) -> None:
+        """Disable the DEIM adaptive augmentation scheduling framework.
+
+        Removes the AugmentationSchedulerCallback from the callbacks list,
+        falling back to the static augmentation pipeline defined in
+        overrides.data.train_subset.
```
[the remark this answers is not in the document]
kprokofi: Done

application/backend/app/models/training_configuration/augmentation.py
```diff
@@ -616,6 +650,16 @@ class AugmentationParameters(BaseModel):
...
+    )
+    deim_framework: DEIMFramework | None = Field(
+        default=None,
+        title="DEIM framework",
+        description=(
+            "Controls the DEIM adaptive augmentation scheduling framework. "
+            "When enabled, training uses multi-phase augmentation scheduling for faster convergence. "
+            "When disabled, a static augmentation pipeline is used."
+        ),
     )
```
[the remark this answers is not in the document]
kprokofi: It is intentional due to UI contract

application/backend/app/models/training_configuration/augmentation.py
```diff
@@ -616,6 +650,16 @@ class AugmentationParameters(BaseModel):
             "allowing images to become sharper, improving robustness to varying image quality. "
             "Applied after resize."
         ),
+        json_schema_extra={"depends_on": {"deim_framework.enable": False}},
+    )
+    deim_framework: DEIMFramework | None = Field(
```
leoll2: One issue of `{"depends_on": {"deim_framework.enable": False}}` is that  `deim_framework` is nullable, so the condition should also evaluate true for `deim_framework: null`. In other words, the full condition should be (in pseudo-code) `deim_framework is None OR deim_framework.enable is False` but the current `depends_on` syntax is not flexible enough to model it. In theory, we could extend `depends_on` to support arbitrary AND/OR logical expression, but it would add more complexity to an already complex UI. Moreover, UI currently doesn't support the dot-notation in the `depends_on` keys.

Under the assumption that `DEIMFramework` doesn't have child parameters other than `enable`, I'd propose an alternative approach:
- represent `deim_framework` as a `bool | None` attribute
- in the other parameters, set `{"depends_on": {"deim_framework": [False, None]}}`, where the list represents a set of values that can be matched to satisfy the condition

```python
class AugmentationParameters:
    random_zoom_out: RandomZoomOut | None = Field(
        ...,
        json_schema_extra={"depends_on": {"deim_framework": [False, None]}}
    )
    deim_framework: bool | None = Field(...)
```
This way, the UI only needs the following two extensions:
- support `depends_on` for parameters group (same depth-level assumption on the key still holds)
- support a list of values in `depends_on`, to be interpreted with _anyOf_ semantics

CC @dwesolow

application/backend/app/models/training_configuration/augmentation.py
```diff
@@ -616,6 +629,16 @@ class AugmentationParameters(BaseModel):
             "allowing images to become sharper, improving robustness to varying image quality. "
             "Applied after resize."
         ),
+        json_schema_extra={"depends_on": {"deim_framework": [False, None]}},
+    )
+    deim_framework: bool | None = Field(
```
leoll2: Let's move `deim_framework` to the top of the augmentations list (L521), because I think it affects the order in which parameters are rendered in the UI.

application/backend/app/models/training_configuration/augmentation.py
kprokofi: Done.
What changes in UI should be made?
@dwesolow can you help me with that?

application/backend/app/supported_models/manifests/detection/dfine_l.yaml
```diff
@@ -32,13 +32,6 @@ hyperparameters:
...
-      random_vertical_flip: null
-      gaussian_blur: null
-      color_jitter: null
-      gaussian_noise: null
-      random_erasing: null
-      random_grayscale: null
-      random_sharpness: null
+      deim_framework: true
+      random_zoom_out:
+        enable: true
```
leoll2, line 37: `random_zoom_out` has `json_schema_extra={"depends_on": {"deim_framework": [False, None]}}`: this means that under the default settings, `random_zoom_out` won't have effect but it will be enabled by default if the user switches off `deim_framework`. Is this the intended behavior?

kprokofi: Yes, that is what I wanted to achieve. If Deim is off, we switch to augmentation pipeline similar to RFDETR or DFINE previously

application/backend/app/models/training_configuration/augmentation.py
kprokofi: OK

application/backend/app/models/training_configuration/augmentation.py
dwesolow: That will require more changes on the UI side, we don't have resources to implement that in the current release. Please create a task, we will address that after releasing current version.

leoll2: Created a task: https://github.com/open-edge-platform/training_extensions/issues/5968

I think we can still merge this PR even without the UI changes: the `deim_framework` parameter will be exposed for the relevant models, the only side-effect is at the UX level - the other augmentation parameters won't be hidden when `deim_framework=true`.

kprokofi: Won't it mislead users? If other augmentations will be still visible and configurable, but will not change the actual pre-processing? We can add a warning at least

leoll2: Adding a warning seems a good compromise: could you add one in the description of the `deim_framework` parameter?

kprokofi: Added

leoll2 (approved):

PR #5917, opened 2026-03-25
Enhance device management for inference and training

### LINKED ISSUE #5886: Validate device in predict endpoint
In `media:predict` endpoint, improve how the `device` parameter is validated and handled:
- devices should be specified in the same format as for the training job (e.g., 'cpu', 'xpu-0', 'cuda-1')
- valid devices are those returned by the `/devices/inference` endpoint
- value `auto` is also valid, and should be used as default
- the inference server should properly convert these names to the format expected by OpenVINO API (for reference, check how it's done in the inference worker in the pipeline)

A-Artemis:
### Summary

Validates the inference device when calling the media predict endpoint.
Improves validation for inference and training device.

### Checklist


application/backend/app/services/system_service.py
```diff
@@ -144,19 +148,52 @@ def get_device_info(self, device_str: str) -> DeviceInfo:
             raise ValueError(f"Device '{device_str}' is not available on the system.")
 
         device_type, device_index = self._parse_device(device_str)
-        if device_type == DeviceType.CPU:
+        if device_type in [DeviceType.AUTO, DeviceType.CPU]:
             return DeviceInfo(type=DeviceType.CPU, name="CPU", memory=None, index=None)
```
leoll2: > 'auto' defaults to CPU.

This default behavior doesn't make sense to me. For device AUTO, we should return `DeviceInfo(type=DeviceType.AUTO, name="AUTO", memory=None, index=None)`.

application/backend/app/services/inference/inference_server.py
```diff
@@ -98,7 +99,7 @@ def set_inference_model(self, project_id: UUID, model_id: UUID, device: str, ttl
 
                 model = Model.create_model(
                     model=str(model_xml_path),
-                    device=device,
+                    device=device.type.name,
```
leoll2: This won't work for Intel GPU because `DeviceInfo` calls it `XPU` whereas Model API expects GPU. Moreover, you are disregarding the device index.

You need a function that converts your `DeviceInfo` to a device string in OpenVINO format: we already implemented that logic for the inference pipeline: 
https://github.com/open-edge-platform/training_extensions/blob/28243ede66ee4c852b6b54342c2882d9c3a06496/application/backend/app/services/active_model_service.py#L40

leoll2 (approved):

PR #5927, opened 2026-03-26
fix: Round frame number properly

dwesolow:
### Summary

The issue was that `fps` is not the integer but `float`. In this case it was `29.992017030335283` instead of `30` which caused not rounded frame indexes, which caused the issue while trying to create annotations (frame number must be integer).

### Checklist

### DIFF
--- a/application/ui/src/features/annotator/video-player/use-video-controls.ts
+++ b/application/ui/src/features/annotator/video-player/use-video-controls.ts
@@ -76,7 +76,10 @@ export const useVideoControls = ({
         setIsPlaying(false);
         videoRef.current.pause();
 
-        goto(currentFrameNumber);
+        const maxNearestFrame = Math.floor((videoFrame.frame_count - 1) / step) * step;
+        const nearestFrame = Math.min(maxNearestFrame, roundFrameNumber(currentFrameNumber, step));
+
+        goto(nearestFrame);
     };
 
     const nextFrame = () => {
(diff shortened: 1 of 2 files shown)


application/ui/src/features/annotator/video-player/use-video-controls.ts
```diff
@@ -17,6 +17,8 @@ export type VideoControls = {
     canPlay?: boolean;
 };
 
+const roundFrameNumber = (frameNumber: number, step: number) => Math.round(Math.round(frameNumber / step) * step);
```
jpggvilaca: do we need to round twice?

dwesolow: yes because step is a float but the result should be integer (first rounding is needed because we want to get the closest multiple of step).

application/ui/src/features/annotator/video-player/use-video-controls.ts
```diff
@@ -122,7 +120,7 @@ export const useVideoControls = ({
         }
 
         videoRef.current.currentTime = (frameNumber + 1) / videoFrame.fps;
```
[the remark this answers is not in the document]
dwesolow: removed

application/ui/src/features/annotator/video-player/use-video-controls.ts
dwesolow: Changed this. `step` that might be float was used in many places and it would be really hard to track correct implementation. In the new approach we round `step` so whenever we calculate, the calculation is based on integer step.

jpggvilaca (dismissed):

jpggvilaca (approved):

PR #5958, opened 2026-03-27
Increase num_workers to resolve a problem with starved data loader

kprokofi:
### PR Summary

**Problem:**  
GPU training throughput is much lower than expected. Profiling shows the GPU is idle **70–92% of the time**, waiting for data rather than performing computations.

**Root cause:**  
The `num_workers` for the DataLoader was set to 2 (classification), which worked fine for CPU training where a forward+backward pass takes ~8 s. On GPU (RTX 3090), each step takes ~100 ms, making the GPU **3–10× faster than the current data pipeline** can supply batches.

**Measured numbers:**

| Metric | CPU training | GPU training |
|--------|--------------|--------------|
| data_time per batch | ~0.002 s | ~0.217 s |

Before: 3 it /s, data_time 0.217
Now: 13 it / s data_time 0.003-0.005


library/src/otx/recipe/_base_/data/detection_tile.yaml
```diff
@@ -10,7 +10,7 @@ tile_config:
 train_subset:
   subset_name: train
   batch_size: 1
-  num_workers: 2
+  num_workers: 4
```
leoll2, line 13: Why 4 and not 8 here?

library/src/otx/data/module.py
```diff
@@ -421,7 +421,7 @@ def train_dataloader(self) -> DataLoader:
             "persistent_workers": config.num_workers > 0,
             "sampler": sampler,
             "shuffle": sampler is None,
-            "prefetch_factor": 2 if config.num_workers > 0 else None,
+            "prefetch_factor": 4 if config.num_workers > 0 else None,
```
AlbertvanHouten: Fetching 2 batches ahead if we have fast enough workers should be good enough so I think we can leave it at 2. It will use double the memory when set to 4 and the increase in the workers is also already x2 or x4 the previously used memory.

leoll2 (dismissed):

leoll2:
> Now: 13 it / s data_time 0.003-0.005

Is GPU slower than CPU?

library/src/otx/recipe/_base_/data/classification.yaml
```diff
@@ -6,7 +6,7 @@ unannotated_items_ratio: 0.0
 train_subset:
   subset_name: train
   batch_size: 64
-  num_workers: 2
+  num_workers: 8
```
AlbertvanHouten, line 9: Maybe we should dynamically set this based on the system its being run on and remove it from the recipe. What do you think of just using something like this in the dataloader? 

`num_workers = min(cpu_cores - 2, 8)`

library/src/otx/data/module.py
kprokofi: Agree

library/src/otx/recipe/_base_/data/classification.yaml
kprokofi: we could introduce smart num_workers in the future, indeed
However, we should test it and think about corner cases instead of setting just "min(cpu_cores - 2, 8)"
IMO, setting it to 8 should be safe for now

library/src/otx/recipe/_base_/data/detection_tile.yaml
kprokofi: For regular detection, each __getitem__ call costs roughly the same for every worker: read one image, apply transforms.
For tile, get_tiles is called per item and does:

* Read one full-res image from disk
* Slice it into N tiles 
* Apply transforms to each of the N tiles
So one worker call does O(N_tiles) work and holds N tile tensors in memory simultaneously. With 8 workers you'd have 8 × N tile tensors live at once. The bottleneck shifts from CPU throughput to memory pressure. 
4 is more safe for tiling, IMO

kprokofi:
> > Now: 13 it / s data_time 0.003-0.005
> 
> Is GPU slower than CPU?

it was true only for batch preparation phase. Inference of course was much faster

library/src/otx/recipe/_base_/data/detection_tile.yaml
kprokofi: I also found that create_tiling_transform is done every time in get_item
I moved it to the init

library/src/otx/data/module.py
kprokofi: Returned

library/src/otx/recipe/_base_/data/classification.yaml
kprokofi: I will create an issue to track this functionality

leoll2 (approved):

library/src/otx/recipe/_base_/data/classification.yaml
kprokofi: https://github.com/open-edge-platform/training_extensions/issues/5978

AlbertvanHouten (approved):

PR #5959, opened 2026-03-27
Bugfix friday 27th

jpggvilaca:
### Summary

* Fix a few issues from [[Umbrella] UI medium/low prio bugs](https://github.com/open-edge-platform/training_extensions/issues/5903)
* Data collection field update
* Disable ssim worker loading since we dont have the enabled tool yet
* Subset split % fix
* Remove image folder option for mvp

### Checklist

### DIFF
--- a/application/ui/src/features/inference/aside/data-collection.component.test.tsx
+++ b/application/ui/src/features/inference/aside/data-collection.component.test.tsx
@@ -1,0 +1,90 @@
+// Copyright (C) 2026 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+import { screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { getMockedPipeline } from 'mocks/mock-pipeline';
+import { HttpResponse } from 'msw';
+import { render } from 'test-utils/render';
+
+import { http } from '../../../api/utils';
+import { server } from '../../../msw-node-setup';
+import { DataCollection } from './data-collection.component';
+
+describe('DataCollection capture rate fields', () => {
+    const renderApp = () => {
+        const pipelinePatchSpy = vi.fn();
+
+        server.use(
+            http.get('/api/projects/{project_id}/pipeline', () => {
+                return HttpResponse.json(
+                    getMockedPipeline({
+                        data_collection: {
+                            max_dataset_size: 500,
+                            policies: [
+                                {
+                                    type: 'fixed_rate',
+                                    rate: 12,
+                                    enabled: true,
+                                },
+                                {
+                                    type: 'confidence_threshold',
+                                    confidence_threshold: 0.5,
+                                    min_sampling_interval: 2.5,
+                                    enabled: false,
+                                },
+                            ],
+                        },
+                    })
+                );
+            }),
+            http.patch('/api/projects/{project_id}/pipeline', async ({ request }) => {
+                pipelinePatchSpy(await request.json());
+
+                return HttpResponse.json({});
+            })
+        );
+
+        render(<DataCollection />);
+
+        return pipelinePatchSpy;
+    };
+
+    beforeEach(() => {
+        vi.resetAllMocks();
+    });
+
+    it('renders Frames and Seconds fields for capture rate', async () => {
+        renderApp();
+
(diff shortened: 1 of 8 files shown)


application/ui/src/features/inference/aside/data-collection.component.tsx
```diff
@@ -103,16 +116,31 @@ export const DataCollection = () => {
...
+                    <NumberField
+                        label='Frames'
+                        minValue={MIN_FRAME_SAMPLING_VALUE}
+                        step={0.1}
+                        value={localRateFrames}
+                        onChange={(nextFrames) => {
+                            setLocalRateFrames(nextFrames);
+                            updateRatePolicy(nextFrames, localRateSeconds);
+                        }}
+                        isDisabled={!ratePolicy?.enabled || isUpdating}
```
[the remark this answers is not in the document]
jpggvilaca: Spectrum's numberfield only commits onBlur/Enter key, so this doesnt apply

application/ui/src/features/inference/aside/data-collection.component.tsx
```diff
@@ -103,16 +116,31 @@ export const DataCollection = () => {
...
-                />
+                <Flex direction='row' gap='size-100' alignItems={'end'} marginBottom={'size-200'}>
+                    <NumberField
+                        label='Frames'
+                        minValue={MIN_FRAME_SAMPLING_VALUE}
+                        step={0.1}
+                        value={localRateFrames}
+                        onChange={(nextFrames) => {
+                            setLocalRateFrames(nextFrames);
+                            updateRatePolicy(nextFrames, localRateSeconds);
```
dwesolow, line 135: each time when value changes we update `updateRatePolicy` and `setLocalRateFrames`. Could we not use `ratePolicy` from the server side and not store it locally (i.e. not to have duplicated state and not synced).

application/ui/src/features/models/model-listing/components/three-section-range/three-section-range.component.tsx
```diff
@@ -12,7 +12,10 @@ type ThreeSectionRangeProps = {
     testingValue: number;
 };
 
+const formatter = new Intl.NumberFormat('en-US', { style: 'percent', maximumFractionDigits: 0 });
```
dwesolow: np. we could use `useNumberFormatter` from aria.

application/ui/src/features/inference/aside/data-collection.component.tsx
jpggvilaca: We need the frame & second values to immediately display to the user. The rate `updateRatePolicy` is just the calculation plus patch. So they are 2 different things

application/ui/src/features/models/model-listing/components/three-section-range/three-section-range.component.tsx
jpggvilaca: Updated

application/ui/src/features/inference/aside/data-collection.component.tsx
dwesolow: Ok but if `updateRatePolicy` fails, our local state no longer indicates the current server state, will that be the issue?

jpggvilaca: Good point. Added onError scenario

application/ui/src/features/inference/aside/data-collection.component.test.tsx
```diff
@@ -0,0 +1,109 @@
...
+                })
+            );
+        });
+
+        await userEvent.clear(secondsInput);
+        await userEvent.type(secondsInput, '3');
+        secondsInput.blur();
+
+        await waitFor(() => {
+            expect(pipelinePatchSpy).toHaveBeenLastCalledWith(
```
camiloHimura, line 95: Since it’s called multiple times, shouldn’t we also assert the number of calls?

dwesolow (approved):

application/ui/src/features/inference/aside/data-collection.component.test.tsx
jpggvilaca: I'll add that on my next PR , ty!

PR #5983, opened 2026-03-30
Performance model variants part 2

### LINKED ISSUE #5977: Show performance of model variants
Since #5632, the model endpoint returns evaluation scores for each model variant that is evaluated. This change was made in the context of the model quantization feature, since we now must re-asses the performance of the model after quantization. For the user, it is very important to compare the quantized and non-quantized models performance-wise, because this information determines the accuracy-speed tradeoff between the two variants and ultimately affects the user's decision of which model to deploy.

According to the UX design, this information can be effectively exposed to the user like this:
Image

The goal of this ticket is to align the UI with the UX design:
- Restore the "Accuracy"/"mAP" column next to each model variant
- Populate the column with the value found in the "evaluations" object of each variant
- If "evaluations" is empty, copy the value from the FP32 Pytorch variant (guaranteed to be non-empty)

jpggvilaca:
### Summary

 * Closes https://github.com/open-edge-platform/training_extensions/issues/5977
 * Also addresses previous PR's comments

Screenshot 2026-03-30 at 16 07 46

### Checklist

### DIFF
--- a/application/ui/src/features/models/model-listing/model-variants/model-variant-table.component.test.tsx
+++ b/application/ui/src/features/models/model-listing/model-variants/model-variant-table.component.test.tsx
@@ -47,8 +47,8 @@ describe('ModelVariantTable', () => {
 
         render(<ModelVariantTable model={model} format='openvino' />);
 
-        expect(screen.getByText('Accuracy')).toBeInTheDocument();
-        expect(screen.getByText('92%')).toBeInTheDocument();
+        expect(screen.getByTestId('model-variant-performance-column')).toHaveTextContent('Accuracy');
+        expect(screen.getByTestId('model-variant-value-accuracy-fp16')).toHaveTextContent('92%');
     });
 
     it('falls back to fp32 pytorch primary metric when evaluations are empty', () => {
@@ -77,8 +77,8 @@ describe('ModelVariantTable', () => {
 
         render(<ModelVariantTable model={model} format='openvino' />);
 
-        expect(screen.getByText('mAP')).toBeInTheDocument();
-        expect(screen.getByText('87%')).toBeInTheDocument();
+        expect(screen.getByTestId('model-variant-performance-column')).toHaveTextContent('mAP');
+        expect(screen.getByTestId('model-variant-value-accuracy-fp16')).toHaveTextContent('87%');
     });
 
     it('shows size and performance deltas for the quantized variant', () => {
@@ -151,8 +151,8 @@ describe('ModelVariantTable', () => {
 
         render(<ModelVariantTable model={model} format='openvino' />);
 
-        expect(screen.getByText('mAP')).toBeInTheDocument();
-        expect(screen.getByTestId('model-variant-value-accuracy-fp16')).toBeInTheDocument();
+        expect(screen.getByTestId('model-variant-performance-column')).toHaveTextContent('mAP');
+        expect(screen.getByTestId('model-variant-value-accuracy-fp16')).toHaveTextContent('-');
         expect(screen.queryByText('87%')).not.toBeInTheDocument();
     });
 
@@ -170,8 +170,8 @@ describe('ModelVariantTable', () => {
 
         render(<ModelVariantTable model={model} format='openvino' />);
 
-        expect(screen.getByText('Score')).toBeInTheDocument();
-        expect(screen.getByTestId('model-variant-value-accuracy-fp16')).toBeInTheDocument();
+        expect(screen.getByTestId('model-variant-performance-column')).toHaveTextContent('Score');
+        expect(screen.getByTestId('model-variant-value-accuracy-fp16')).toHaveTextContent('-');
     });
 
     it('triggers download for selected variant', async () => {
--- a/application/ui/src/features/models/model-listing/model-variants/model-variant-table.component.test.tsx
+++ b/application/ui/src/features/models/model-listing/model-variants/model-variant-table.component.test.tsx
@@ -51,7 +51,7 @@ describe('ModelVariantTable', () => {
         expect(screen.getByText('92%')).toBeInTheDocument();
     });
 
-    it('falls back to fp32 pytorch primary metric when current variant evaluations are empty', () => {
+    it('falls back to fp32 pytorch primary metric when evaluations are empty', () => {
         const model = getMockedModel({
             variants: [
                 getMockedVariant({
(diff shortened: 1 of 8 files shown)


application/ui/src/features/models/model-listing/model-variants/model-variant-table.component.tsx
```diff
@@ -4,59 +4,33 @@
...
+};
 export const ModelVariantTable = ({ model, format }: ModelVariantTableProps) => {
     const { downloadModel, isDownloading } = useDownloadModel(model.id);
     const allVariants = model.variants ?? [];
-    const variants = (model.variants ?? []).filter((variant) => variant.format === format);
-    const fp32PytorchVariant = allVariants.find(
-        (variant) => variant.format === 'pytorch' && variant.precision === 'fp32'
-    );
+    const variants = allVariants.filter((variant) => variant.format === format);
+    const baselineVariant = variants[0];
```
dwesolow: That's not safe assumption that baseline model is the first item in the array, easy to break. Is there any better alternative?

application/ui/tests/models/model-details.spec.ts
```diff
@@ -282,6 +294,18 @@ test.describe('Model Details', () => {
...
+            await expect(modelsPage.getQuantizationDialog()).toBeHidden();
+
+            await expect(page.getByText('INT8', { exact: true })).toBeVisible();
+            await expect(page.getByText('52.4 MB')).toBeVisible();
+
+            // Size delta: INT8 is -75% smaller than the FP32 baseline
+            await expect(page.getByText('-75%')).toBeVisible();
+
+            // Accuracy delta: INT8 is 79% vs FP32 baseline 82% -> -4% drop
+            await expect(page.getByText('79%')).toBeVisible();
```
dwesolow: could we narrow the selector, i.e. sth like `getByRole('row', { name: 'fp32' } ).getByTestId('accuracy-delta'))`?

application/ui/src/features/models/model-listing/model-variants/model-variant-table.component.tsx
jpggvilaca: Agree. FP16 should always be there, but if not, fp32 will (by design), so i just did : "pick fp16 if exist, if not, pick fp32"

application/ui/tests/models/model-details.spec.ts
jpggvilaca: Agree, updated

application/ui/src/features/models/model-listing/model-variants/model-variant-table.component.tsx
```diff
@@ -4,59 +4,34 @@
...
 export const ModelVariantTable = ({ model, format }: ModelVariantTableProps) => {
     const { downloadModel, isDownloading } = useDownloadModel(model.id);
     const allVariants = model.variants ?? [];
-    const variants = (model.variants ?? []).filter((variant) => variant.format === format);
-    const fp32PytorchVariant = allVariants.find(
-        (variant) => variant.format === 'pytorch' && variant.precision === 'fp32'
-    );
+    const variants = allVariants.filter((variant) => variant.format === format);
+    const baselineVariant = getBaselineVariant(variants);
+    const fp32PytorchVariant = getFp32PytorchVariant(allVariants);
```
dwesolow, line 28: should it not use `variants` like `getBaselineVariant`?

application/ui/src/features/models/model-listing/model-variants/model-variant-table.component.test.tsx
```diff
@@ -81,6 +81,44 @@ describe('ModelVariantTable', () => {
...
+                            metrics: [{ name: 'Accuracy', value: 0.89, primary: true }],
+                        },
+                    ],
+                }),
+            ],
+        });
+
+        render(<ModelVariantTable model={model} format='openvino' />);
+
+        expect(screen.getByText('-50%')).toHaveStyle({ color: 'var(--moss-tint-1)' });
```
dwesolow, line 118: could we improve this selector like for other cases? I.e. to be sure that `-50%` is exactly where we expect it to be.

application/ui/src/features/models/model-listing/model-variants/model-variant-table.component.test.tsx
```diff
@@ -114,7 +152,7 @@ describe('ModelVariantTable', () => {
         render(<ModelVariantTable model={model} format='openvino' />);
 
         expect(screen.getByText('mAP')).toBeInTheDocument();
-        expect(screen.getByText('-')).toBeInTheDocument();
+        expect(screen.getByTestId('model-variant-value-accuracy-fp16')).toBeInTheDocument();
```
dwesolow: what about checking the value instead of checking only the visibility?

application/ui/src/features/models/model-listing/utils/variant-metrics.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    if (primaryMetric === undefined) {
+        return undefined;
+    }
+
+    return { name: primaryMetric.name, value: Math.round(primaryMetric.value * 100) };
+};
+
+export const getFp32PytorchVariant = (variants: ModelVariant[]): ModelVariant | undefined => {
+    return variants.find((variant) => variant.format === 'pytorch' && variant.precision === 'fp32');
```
dwesolow, line 29: Why do we compare it to `pytorch`? We use that metric as a fallback in the tab where we display `openvino` models.

jpggvilaca: because of the REQ: ```If "evaluations" is empty, copy the value from the FP32 Pytorch variant (guaranteed to be non-empty)```

application/ui/src/features/models/model-listing/model-variants/model-variant-table.component.tsx
jpggvilaca: No because we need to fetch across all variants, in this case, to find the pytorch one

application/ui/src/features/models/model-listing/utils/variant-metrics.ts
jpggvilaca: cc @leoll2

application/ui/src/features/models/model-listing/model-variants/model-variant-table.component.test.tsx
jpggvilaca: Agree, updated

application/ui/src/features/models/model-listing/model-variants/model-variant-table.component.test.tsx
jpggvilaca: Agree, updated

dwesolow: That part is not updated

application/ui/src/features/models/model-listing/utils/variant-metrics.test.ts
```diff
@@ -0,0 +1,96 @@
...
+                {
+                    dataset_revision_id: 'dataset-1',
+                    subset: 'testing',
+                    metrics: [{ name: 'Accuracy', value: 0.9, primary: false }],
+                },
+            ],
+        });
+
+        expect(getVariantPerformanceValue(withMetric, fp32Metric)).toBe(89);
+        expect(getVariantPerformanceValue(emptyEvaluations, fp32Metric)).toBe(87);
```
camiloHimura:
```suggestion
        expect(getVariantPerformanceValue(emptyEvaluations, fp32Metric)).toBe(fp32Metric.value);
```

camiloHimura (dismissed):

application/ui/src/features/models/model-listing/utils/variant-metrics.test.ts
jpggvilaca: Updated

dwesolow (approved):

PR #6035, opened 2026-04-02
Fix Augmentation Pipeline (Detection)

### LINKED ISSUE #6034: Zero Detection Accuracy After Augmentation Pipeline Upgrade
**Affected models:** All detection models (ATSS, YOLOX, RF-DETR)
**Introduced by:** #5021 — *Upgrade Augmentations Pipeline in OTX + 16bit images support*

---

Need to check and validate the whole pipeline and to find a root cause. 
Probably due to:

GPUAugmentationPipeline._sanitize_annotations()` is unconditionally called after every GPU
pipeline forward pass, including **validation and test**. It clips bounding boxes to the current
image tensor dimensions `(H, W)`:

```python
batch_size, _, h, w = images.shape   # e.g. 800 × 992 (network input size)
clipped[:, 0::2].clamp_(0, w)        # x-coords clamped to 992
clipped[:, 1::2].clamp_(0, h)        # y-coords clamped to 800

### PR DESCRIPTION
**Pull request by kprokofi**
### Summary

Resolves https://github.com/open-edge-platform/training_extensions/issues/6034

## Problem

After the data pipeline refactor, several augmentation transforms were broken, producing near-zero detection accuracy. The root causes were in CachedMosaic, CachedMixUp, and incorrect/missing recipe configurations.

### Core transforms (`library/src/otx/data/augmentation/transforms.py`)
- **CachedMosaic** stripped back to a straightforward 2×-canvas mosaic — the old version was corrupting tile placement
- **CachedMixUp** now samples the blending ratio from `Beta(alpha, alpha)` instead of using a fixed `mix_ratio`. Default `alpha=1.5` gives a gentle bell around 50/50 blending, matching standard practice

### YOLOX recipes (s, l, x, tiny)
- Rebuilt the full training pipeline: Mosaic → RandomResizedCrop → SanitizeBBox → MixUp (CPU), then ColorJiggle → HFlip → RandomAffine on GPU
- `yolox_x.yaml` was completely empty (just Resize + Flip) — now has the full pipeline like the other variants
- RandomAffine set to `p=0.7` with no scaling (`scale=[1.0, 1.0]`) — mild geometric augmentation without always deforming the image
- Val/test pipelines: aspect-ratio-preserving resize, no GPU augmentations

### DEIM / DEIMv2 recipes
Verified correct against baseline, no changes needed.

### RF-DETR
- Fixed `rfdetr_large` default input size: was `(560, 560)`, corrected to `(704, 704)` per official RF-DETR config

### Removed 46 identity Normalize blocks
Removed across 15 recipe files — these were `mean=0, std=1` (no-op) left over from the refactor.

### Backend application alignment
- `Mixup` model: replaced `mix_ratio` (fixed 0–1 float) with `alpha` (Beta distribution param, 0.1–10.0, default 1.5)
- All 4 YOLOX manifests updated: added `alpha: 1.5`, fixed `random_affine` defaults to match recipes (`scale=[1.0,1.0]`, `p=0.7`)
- Relaxed `scaling_ratio_range` validator to allow equal values (needed for `[1.0, 1.0]` = no scaling)
- Updated all related backend and library tests

### CODE REVIEW

File: library/src/otx/data/augmentation/pipeline.py [hunk 1]
```diff
@@ -453,6 +453,10 @@ class GPUAugmentationPipeline(nn.Module):
         augmentations: List of Kornia augmentation modules.
         data_keys: List of data keys to transform (e.g., ["input", "bbox", "mask"]).
             Defaults to ["input"] for image-only augmentation.
+        sanitize_annotations: Whether to clip and filter bboxes/keypoints after
+            geometric transforms.  Set to ``False`` for validation/test pipelines
+            where ground-truth coordinates may be in original image space to prevent
+            clipping them to the smaller network input dimensions.
```
leoll2, line 459: I don't fully understand the part about validation/test: if the augmentation alters the geometry of the image, why wouldn't you adjust the annotation coordinates accordingly?

kprokofi: We don't use any augmentations for val/test and basically we should not do that. 
For fair validation we need to utilize all provided annotations
For training we can alter it whatever we want

leoll2 (approved):

library/src/otx/data/augmentation/pipeline.py
kprokofi: I did the second layer of protection because we don't resize annotations for Detection task in validation. That means even if we sanitize_annotations only after geometric augmentations, they will completely disappear

PR #6049, opened 2026-04-07
fix: Clear LS after the app closes

### LINKED ISSUE #6029: Dataset import: An error occurred during import. Job not found (after dataset import and app restart)
Install Geti -> Create a project from Dataset -> wait till completion and project created successfully
Close Geti and Open again -> back to project list page -> look at the error

Image

camiloHimura:
### Summary
Fixes https://github.com/open-edge-platform/training_extensions/issues/6029.

This PR adds a new Tauri handler that clears the local storage import/export information when the app closes, so no invalid IDs are refetched the next time it opens.

### Checklist


application/ui/src/hooks/storage/use-dataset-import-storage.hook.ts
```diff
@@ -1,14 +1,14 @@
...
 
-import { useLocalStorage } from 'usehooks-ts';
+import { useSessionStorage } from 'usehooks-ts';
 
-import { DatasetImportState, getParsedLocalStorage } from './utils';
+import { DatasetImportState, getParsedSessionStorage } from './utils';
 
 export const useDatasetImportStorage = <TStep, TEntry extends DatasetImportState<TStep>>(storageKey: string) => {
-    const [importEntries, setImportEntries] = useLocalStorage<TEntry[] | null>(
+    const [importEntries, setImportEntries] = useSessionStorage<TEntry[] | null>(
```
dwesolow: Why this change? IMO that was a nice feature that user could open different tabs with the app and still see the progress of the import.

camiloHimura: Local Storage doesn’t work well for the desktop version. When the user closes and reopens the app, all jobs are cleared, which leads to this error: https://github.com/open-edge-platform/training_extensions/issues/6029

Using session storage instead ensures that all job-related information is cleared automatically.

dwesolow: IMO with this change we hide the problem instead of solving that properly. If I understand correctly, the issue right now is that when user opened the desktop app again the local storage might contain the outdated job that was removed automatically when closing the desktop app previously. What about clearing job state in the local storage when closing the app?

camiloHimura: I disagree. We need the storage to be cleared when the app closes, and session storage already gives us that behavior. Your approach adds more code to achieve the same result, so to me it doesn’t feel like the right solution.

application/ui/src/hooks/storage/use-dataset-import-storage.hook.ts
```diff
@@ -1,14 +1,14 @@
...
         storageKey,
-        () => getParsedLocalStorage<TEntry[]>(storageKey) ?? null
+        () => getParsedSessionStorage<TEntry[]>(storageKey) ?? null
     );
```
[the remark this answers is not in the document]
camiloHimura: updated

application/ui/src/hooks/storage/use-dataset-import-storage.hook.ts
itallix: If we still need to keep the progress between tabs when using a browser, then the approach suggested by @dwesolow would be more preferable. It comes with some extra code, indeed, but the feature is nice :-)

application/ui/src-tauri/capabilities/default.json
```diff
@@ -3,5 +3,5 @@
     "identifier": "default",
     "description": "enables the default permissions",
     "windows": ["main"],
-    "permissions": ["core:default"]
+    "permissions": ["core:default", "core:window:allow-on-close-requested"]
```
jpggvilaca, line 6: what does this do?

application/ui/src/features/dataset/import-export/import-jobs-list/import-jobs-list.component.tsx
```diff
@@ -36,6 +36,13 @@ export const ImportJobsList = () => {
                 { params: { path: { project_id: projectId } } },
             ]),
         });
+        queryClient.invalidateQueries({
```
jpggvilaca: why didnt we need this before?

application/ui/src/tauri/utils/storage-cleanup-registry/storage-cleanup-registry.ts
```diff
@@ -0,0 +1,30 @@
...
+const isDatasetStorageKey = (key: string): boolean =>
+    DATASET_STORAGE_KEYS.includes(key) || DATASET_STORAGE_PREFIXES.some((prefix) => key.startsWith(prefix));
+
+const cleanDatasetStorage = (): void => {
+    Object.keys(localStorage)
+        .filter(isDatasetStorageKey)
+        .forEach((key) => localStorage.removeItem(key));
+};
+
+const isTauri = (): boolean => '__TAURI_INTERNALS__' in window;
```
jpggvilaca, line 20: Can we extract Tauri-related code to another folder? We are mixing storage and tauri here

jpggvilaca (commented):
PR title says we're replacing local with session storage but i dont see sessionStorage being used anywhere?

camiloHimura:
> PR title says we're replacing local with session storage but i dont see sessionStorage being used anywhere?

it was the original intention but we move to a different approach, so I'll update the title

application/ui/src/features/dataset/import-export/import-jobs-list/import-jobs-list.component.tsx
camiloHimura: It was actually missing, this ticket was created because of it: https://github.com/open-edge-platform/training_extensions/issues/6078. However, it has its own dedicated PR, so I’ll remove that code from this PR.

application/ui/src-tauri/capabilities/default.json
camiloHimura: it grants permission to handle window close events, so we can clear the LS

jpggvilaca (approved):

PR #6052, opened 2026-04-07
Add a parameter for branch selection for windows installer workflow

### LINKED ISSUE #6014: Add parameter to specify branch in windows-installer.yml
In `windows-installer.yml` workflow, add a parameter to choose the branch to checkout before starting the build, so that the app is built based on the given branch workflow. Note that this is different from the existing "Use workflow from" option, which only specifies where to pull the GH action itself but not the branch of the application source code.
Default should be `develop`.

A-Artemis:
### Summary

Added a parameter to allow branch selection for the windows installer workflow
image

### Checklist

### DIFF
--- a/.github/workflows/windows-installer.yml
+++ b/.github/workflows/windows-installer.yml
@@ -16,11 +16,6 @@ on:
         type: boolean
         required: false
         default: false
-      checkout_branch:
-        description: "Branch to checkout for building the application source code"
-        type: string
-        required: false
-        default: "develop"
 
   # allow to manually trigger this workflow
   workflow_dispatch:
@@ -42,11 +37,10 @@ on:
         required: false
         type: boolean
         default: false
-      checkout_branch:
-        description: "Branch to checkout for building the application source code"
+      checkout_sha:
+        description: "Full Commit SHA to checkout for building the application source code"
+        required: true
         type: string
-        required: false
-        default: "develop"
 
 env:
   GIT_SH_PATH: C:\Program Files\Git\bin\sh.exe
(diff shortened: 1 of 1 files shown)


.github/workflows/windows-installer.yml
```diff
@@ -16,6 +16,11 @@ on:
         type: boolean
         required: false
         default: false
+      checkout_branch:
```
AlexanderBarabanov: Could you please clarify why do we need this additional `checkout_branch` input for `workflow_call` trigger case? 
I propose to remove `checkout_branch` from `workflow_call` trigger and use caller ref in checkout (that is available via `github.ref`) - this will be more secure option as potentially prevent untrusted code execution:


```
  - name: Checkout code
    uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
    with:
      ref: ${{ inputs.checkout_sha || github.ref }}
      persist-credentials: false
```

.github/workflows/windows-installer.yml
```diff
@@ -37,6 +42,11 @@ on:
         required: false
         type: boolean
         default: false
+      checkout_branch:
```
AlexanderBarabanov: We should use immutable references to check out code here to prevent potential race conditions and GH token/secrets exfiltration (risk is low, but it's still better to harden the approach in case it is copy-pasted).

Here is a real-world use case for reference: https://adnanthekhan.com/posts/dependabot-core-toctou-writeup/ - please check the "How To Fix It (in most cases)" section for mitigation.

I propose to replace `checkout_branch` with a full commit SHA.

.github/workflows/windows-installer.yml
A-Artemis: I thought it would be useful for automation. Thanks for pointing out your option, I will make the changes.

AlexanderBarabanov (approved):
Thank you!

PR #6056, opened 2026-04-08
Automated GetiTune benchmarking pipeline

### LINKED ISSUE #5046: Automated workflow for model performance evaluation
## Objective

Setup an automated workflow to evaluate the performance of OTX models on a curated set of testing datasets, to facilitate the comparison between different models or different implementations of the training pipeline.

## Motivation

OTX is a framework for training vision models. To keep pace with recent developments in deep learning, engineers are constantly required to implement new features and therefore introduce changes to the training pipeline.
At the same time, it is crucial that such changes do not introduce bugs and regressions in the already supported features and architectures: all supported models must keep working and achieve the expected results in terms of accuracy, training time, and inference speed. To this end, models must be periodically tested and benchmarked: however, the model evaluation is currently carried out _manually_ by the engineers themselves. This is not only very time-consuming, but also error-prone, not transparent and not easy to reproduce. We would like to automate this step, so that engineers can fully focus on the development of new features while ensuring that model performance remains optimal.

### Datasets

- Define a minimal but representative set of datasets for each task, to use for testing purposes and benchmarking.
  - All datasets must have meaningful and accurate annotations
  - Datasets should be classified by size (tiny, small, medium, large) - this allows evaluating the model performance over different dataset sizes, which is very important for applications like Geti where the training dataset is built incrementally.

### Workflow

- The workflow must automatically pull the required input data (dataset, initial weights) from a publicly available archive.
- The workflow must be compatible with all models, and allow the developers to select a subset of models to test (single model, all models or a list).
- The workflow must run one or more full experiments with a specific model architecture, covering different scenarios
  - Examples: train with default parameters, train with tiling, etc...
- The workflow must create a report of the experiments, including information about the final model performance (e.g. accuracy) and total training time
  - Recommended: the workflow should compare the results with the reports generated from the latest release and the previous run on develop.
    - Consider integrating mlflow or similar tools to track metrics, run comparisons and and visualize the results
- The workflow should be parametrized to run on different runners with different HW setups, ideally covering all classes of supported accelerators (GPU, XPU, ...)
- The workflow must tolerate failures that occur during the training, report them and continue with the other experiments.
- The workflow should run in the CI on a regular basis (e.g. daily/weekly) or on-demand.

AlbertvanHouten:
### Summary
This PR introduces a full benchmarking workflow for `getitune` so model quality/performance can be measured consistently across tasks, datasets, scenarios, and seeds. The goal is to make benchmark runs reproducible and comparable over time, while controlling resource usage for long benchmark batches.

### Changes

- Added a CLI-driven benchmark package with subcommands for dataset provisioning, execution, report generation, and retention cleanup in `library/src/getitune/benchmark/cli.py`.
- Implemented dataset catalog parsing and script-based provisioning in `library/src/getitune/benchmark/catalog.py`, including streamed prep-script logs and .ready sentinel handling to avoid treating partial datasets as complete. Download scripts are under `library/scripts/benchmark_datasets`
- Implemented manifest parsing and experiment enumeration in `library/src/getitune/benchmark/manifest.py`, including task/model/dataset/scenario filters, per-task thresholds, scenario-level overrides, and size-tier filtering.
- Added phase-based experiment execution in `library/src/getitune/benchmark/experiment.py` (train/test/export/optimize), metric scraping from Lightning CSV logs, wall-time/latency capture, resume-point detection, and structured failure results.
- Added orchestration logic in `library/src/getitune/benchmark/runner.py` with:
  - experiment/seed looping,
  - deterministic fallback retries (True -> "warn" -> False) for unsupported deterministic ops,
  - subprocess isolation per seed to contain long-run memory growth,
  - periodic resource cleanup,
  - optional model rotation-group filtering,
  - optional checkpoint/model artifact cleanup after metric extraction.
- Added MLflow tracking integration in `library/src/getitune/benchmark/tracking.py`, including normalized metric keys, per-seed + rollup run structure, baseline resolution across benchmark experiments, structured failure tagging, and old-run purge support.
- Added regression/report pipeline in `library/src/getitune/benchmark/report.py` to generate:
  - Markdown summary (report.md),
  - aggregated CSV (aggregated.csv),
  - structured failure log (failed_experiments.json).
- Added shared helpers for benchmark dataset prep scripts in `library/src/getitune/benchmark/dataset_helpers.py`.
- Added benchmark configuration files `library/benchmark_catalog.yaml` and `library/benchmark_manifest.yaml` plus MLflow deployment helper `library/src/getitune/benchmark/docker-compose.yaml`.
- Added/maintained unit coverage for benchmark modules under library/tests/unit/benchmark

Resolves https://github.com/open-edge-platform/training_extensions/issues/5046

### Checklist


library/docs/design/automated-model-benchmarking.md
```diff
@@ -0,0 +1,1711 @@
...
+#### Size Tier Definitions
+
+| Tier       | Item Count (images / samples) | Typical Use                                          |
+| ---------- | ----------------------------- | ---------------------------------------------------- |
+| **tiny**   | **< 50**                      | Unit-level smoke tests, PR checks, fast iteration    |
+| **small**  | **50 – 200**                  | Nightly quick benchmarks, scenario sweeps            |
+| **medium** | **200 – 1 000**               | Weekly full benchmarks, release validation           |
+| **large**  | **> 1 000**                   | On-demand deep evaluation, publication-grade results |
+
+> **Guideline:** Every task should have at least one `tiny` and one `small`
```
kprokofi, line 195: Does it mean  that you propose to run performance test every PR with tiny dataset? Small every day and medium every week?
+ large before release?

library/docs/design/automated-model-benchmarking.md
```diff
@@ -0,0 +1,1711 @@
...
+> `large` datasets are reserved for weekly and on-demand runs where statistical
+> significance and real-world representativeness matter more than speed.
+
+### 5.2 Schema
+
+```yaml
+# benchmark_catalog.yaml
+version: 1
+
+datasets:
```
kprokofi, line 219: I like this yaml approach

library/docs/design/automated-model-benchmarking.md
```diff
@@ -0,0 +1,1711 @@
...
+    # train → train only
+    # export → train + export
+    # optimize → train + export + optimize
+  deterministic: true # see note below
+
+experiments:
+  detection:
+    models:
+      - name: atss_mobilenetv2
+        category: default
```
kprokofi: category is redundant for performance tests. They are only useful for Geti

library/docs/design/automated-model-benchmarking.md
```diff
@@ -0,0 +1,1711 @@
...
+        priority: core
```
kprokofi, line 369: Instead of strange naming like "core", "extend" .. I propose to define here what models to run on the weekly basis and during release

library/docs/design/automated-model-benchmarking.md
```diff
@@ -0,0 +1,1711 @@
...
+        recipe: detection/atss_mobilenetv2.yaml
```
kprokofi, line 370: recipe path is also redundant. We have task info and model name. Our recipes follow by "task/name.yaml"

As sum up: here I would define only task, name and frequency/status when to run

library/docs/design/automated-model-benchmarking.md
```diff
@@ -0,0 +1,1711 @@
...
+      # ...full list auto-discoverable from recipe directory
+
+    datasets:
+      # References entries in benchmark_catalog.yaml by name
+      - pothole_tiny
+      - wgisd_small
+      - diopsis_medium
+      - visdrone_large
+
+    scenarios:
```
kprokofi, line 395: I don't understand why we need this. We design our performance tests to check that everything working correctly, not as experiments framework.
Regarding tiling: I would propose to include them as ordinary models, we define models with tiling capabilities with separated recipe.
So, "dfine_tile.yaml".

Also, I don't think we should really test every tiling version. We need to check tiling algorithm in general and test only base model, IMO

library/docs/design/automated-model-benchmarking.md
```diff
@@ -0,0 +1,1711 @@
...
+    criteria:
+      # The primary accuracy metric for this task. The `{metric}` placeholder
+      # in threshold keys below is replaced with this value at manifest load
+      # time. For detection this produces "training:val/f1-score", etc.
+      # Each task section declares its own accuracy_metric (e.g. "accuracy"
+      # for classification, "Dice" for segmentation, "PCK" for keypoints).
+      # This mirrors the existing TASK_METRIC_MAP in tests/perf_v2/summary.py.
+      accuracy_metric: f1-score
+      thresholds:
+        "training:val/{metric}": { compare: ">=", margin: 0.10 }
```
kprokofi, line 464: As we discussed it previously, before defining criteria, we need to be sure that experiments are reproducible and reliable

library/docs/design/automated-model-benchmarking.md
```diff
@@ -0,0 +1,1711 @@
...
+          overrides.data.train_subset.augmentations_gpu: []
+
+    criteria:
+      # The primary accuracy metric for this task. The `{metric}` placeholder
+      # in threshold keys below is replaced with this value at manifest load
+      # time. For detection this produces "training:val/f1-score", etc.
+      # Each task section declares its own accuracy_metric (e.g. "accuracy"
+      # for classification, "Dice" for segmentation, "PCK" for keypoints).
+      # This mirrors the existing TASK_METRIC_MAP in tests/perf_v2/summary.py.
+      accuracy_metric: f1-score
```
kprokofi: Minor: we switched to mAP for accuracy

library/docs/design/automated-model-benchmarking.md
```diff
@@ -0,0 +1,1711 @@
...
+      - pneumonia_tiny
+      # ...
+    criteria:
+      accuracy_metric: accuracy
+      # ...
+
+  # ...one section per task
+```
+
+> **`deterministic: true` trade-off:** Enabling deterministic mode disables
```
kprokofi, line 485: We need to set this parameter carefully.
1) not all Ops suports deterministic -> some models will fail 
2) XPU previously couldn't work normally with deterministic parameter. -> need to recheck

library/docs/design/automated-model-benchmarking.md
```diff
@@ -0,0 +1,1711 @@
...
+```
+
+All commands above run each model with its **default recipe configuration**
+unless `--scenario` is explicitly provided.
+
+This directly satisfies the requirement:
+
+> _"Allow the developers to select a subset of models to test (single model, all models, or a list)."_
+
+### 6.3 Configurable Parameter Scenarios (Optional)
```
kprokofi, line 529: Again, let's not complicate things and focus only on regression tests first.

library/docs/design/automated-model-benchmarking.md
```diff
@@ -0,0 +1,1711 @@
...
+unless a user explicitly requests it:
+
+| Trigger                         | Models                                     | Scenarios Run                                                                |
+| ------------------------------- | ------------------------------------------ | ---------------------------------------------------------------------------- |
+| **Weekly full benchmark**       | `core` + `extended`                        | `default` only                                                               |
+| **Nightly quick run**           | `core` only                                | `default` only (tiny+small datasets)                                         |
+| **PR benchmark**                | Affected `core` models                     | `default` only (tiny datasets, single seed)                                  |
+| **Manual `workflow_dispatch`**  | User-selected via `--model` / `--priority` | User-selected via `--scenario` (defaults to `default` if omitted)            |
+| **PR label `benchmark:params`** | Affected `core` models                     | `default` + all `configurable`-tagged scenarios (tiny datasets, single seed) |
```
kprokofi, line 699: What does it mean? We should manually label PR somehow? 
Models trigger should be set for all models, not only selected as "core"

Running on tiny will produce garbage result, since they are not reliable...
However, we should experiment with that, but from my experience metric can change drastically from run to run on tiny datasets

library/docs/design/automated-model-benchmarking.md
```diff
@@ -0,0 +1,1711 @@
...
+priority tiers**, **rotating schedules**, and **smart filtering** so that CI
+budgets remain bounded while still guaranteeing coverage over time.
+
+#### 6.4.1 Model Priority Tiers
+
+Every model in the manifest declares a `priority` field:
+
+| Priority          | Description                                                                                                                      | Expected Count                | Benchmark Cadence                    |
+| ----------------- | -------------------------------------------------------------------------------------------------------------------------------- | ----------------------------- | ------------------------------------ |
+| **`core`**        | Flagship models, the most-used or strategically important model per task/category. These are the models that _must not_ regress. | ~1–3 per task (~15–25 total)  | Every nightly + Weekly               |
```
kprokofi: I propose the following schema:
1. **default**: default models for Geti. 1 model per task (every night)
2. **recomended**: all recommended by Geti models. 3 models per task. Run every week (weekends run)
2. **extended**: 1 model per family: run every release after FF
3. **on_demand**: all others, run by demand / after architecture change

library/docs/design/automated-model-benchmarking.md
```diff
@@ -0,0 +1,1711 @@
...
+Assigning priorities is a **human decision** made when adding a model to the
+manifest. The guideline is:
+
+- If a regression in this model would block a release → `core`.
+- If a regression matters but can be fixed in the next sprint → `extended`.
+- If this model is experimental or rarely used in production → `exploratory`.
+
+#### 6.4.2 Rotating Schedule for Extended Models
+
+With 50+ `extended` models, running them all every week may still exceed the CI
```
kprokofi, line 742: Please, see above comment. I think simplified workflow could work as well

library/docs/design/automated-model-benchmarking.md
```diff
@@ -0,0 +1,1711 @@
...
+| Export latency       | Metric `export:test/latency`   | `0.012`                 |
+| Recipe YAML          | Artifact                       | `atss_mobilenetv2.yaml` |
+| Raw metrics CSV      | Artifact                       | `metrics.csv`           |
+
+### 8.3 Deployment Options
+
+| Option                           | When to use                                                                                                                                    |
+| -------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------- |
+| **Local file store** (`mlruns/`) | Local developer runs, quick experiments. Artifacts stored as CI job output.                                                                    |
+| **Remote tracking server**       | Team-wide dashboard. A long-running MLflow server (e.g. on an internal VM or managed service). CI jobs post results via `MLFLOW_TRACKING_URI`. |
```
kprokofi, line 1076: what kind of remote server we could use for Geti?

library/docs/design/automated-model-benchmarking.md
```diff
@@ -0,0 +1,1711 @@
...
+    )
+    if not runs:
+        return None
+    return runs[0].data.metrics
+```
+
+The query filters ensure that:
+
+- **Accelerator-scoped**: A GPU run is only compared against a previous GPU
+  baseline, never against XPU or CPU.
```
kprokofi, line 1143: That is definitely unwanted behavior, we should be able to compare between NVIDIA and XPU

library/docs/design/automated-model-benchmarking.md
```diff
@@ -0,0 +1,1711 @@
...
+## 🏋️ OTX Benchmark Report — 2026-03-26
+
+**Branch:** `feature/new-augmentation` | **Commit:** `a1b2c3d` | **Accelerator:** NVIDIA A100
+
+### Detection
+
+| Model            | Dataset        | Val F1 ↑     | Test F1 ↑ | Export F1 ↑ | Train Time ↓ | GPU Mem ↓ | Test Latency ↓ | Status                     |
+| ---------------- | -------------- | ------------ | --------- | ----------- | ------------ | --------- | -------------- | -------------------------- |
+| yolox_s          | pothole_tiny   | 0.87 (±0.01) | 0.85      | 0.84        | 5m 12s       | 4.8 GB    | 12ms           | ✅                         |
+| atss_mobilenetv2 | diopsis_medium | 0.72 (±0.02) | 0.70      | 0.68        | 12m 05s      | 3.2 GB    | 8ms            | ⚠️ val F1 -13% vs baseline |
```
kprokofi: baseline should also be reported here

library/docs/design/automated-model-benchmarking.md
```diff
@@ -0,0 +1,1711 @@
...
+| dfine_x | visdrone_large | 2    | OOM — CUDA out of memory |
+
+📊 [Full results on MLflow](https://mlflow.example.com/experiments/42)
+```
+
+---
+
+## 10. PR Change Detection & Targeted Benchmarks
+
+> **Note:** This section describes a future enhancement. The detailed design
```
kprokofi, line 1239: I think this should be implemented as the second iteration. Generally, it is a good idea to trigger benchmark based on modified files with manual skip possibility

library/docs/design/automated-model-benchmarking.md
```diff
@@ -0,0 +1,1711 @@
...
+The same workflow file is parameterized by `accelerator`. Self-hosted runners
+are labeled by hardware class:
+
+| Label | Hardware                    | Use case                         |
+| ----- | --------------------------- | -------------------------------- |
+| `gpu` | NVIDIA A100 / A10           | Primary GPU benchmark            |
+| `xpu` | Intel Arc / Data Center GPU | XPU support validation           |
+| `cpu` | Any x86_64                  | CPU-only export/optimize testing |
+
+The weekly schedule runs on `gpu` and `xpu`. The `workflow_dispatch` trigger lets
```
kprokofi, line 1468: I would add additional field in configuration for the benchmark: device support: GPU / XPU
Generally all models should support XPU, but sometimes slowness and degradation can happen and we should exclude this specific model to run on XPU

library/docs/design/automated-model-benchmarking.md
```diff
@@ -0,0 +1,1711 @@
...
+
+### 14.1 Source Code
+
+```
+library/
+├── src/otx/benchmark/
+│   ├── __init__.py
+│   ├── __main__.py
+│   ├── cli.py
+│   ├── catalog.py
```
kprokofi, line 1633: Do we really need 2 separate python file to parse yaml?

library/docs/design/automated-model-benchmarking.md
```diff
@@ -0,0 +1,1711 @@
...
+| 5   | Should we support multi-GPU training benchmarks?    | Not in v1                                                 | The existing `benchmark.py` has `num_devices` support. Add as a scenario later if needed (`num_devices > 1`).                                                                                                                                                                                                                           |
+| 6   | Maximum acceptable CI runtime for weekly run?       | ~20 hours on a single A100                                | All benchmarks run sequentially on one self-hosted GPU runner. The runner is given a 48h timeout. Use `--priority`, `--size-tier`, and model rotation (§6.4.2) to keep runtime within budget. Nightly quick runs (~5 hours) use `core` models + `tiny`+`small` datasets only.                                                           |
```
kprokofi: I would opt for the whole weekends (Friday night till Monday morning)

kprokofi (commented):
Thank you Albert for this proposal.

First of all, I like the following points:

1) The existing approach in tests/perf_v2/ hardcodes MODEL_REGISTRY, DATASET_REGISTRY, and CRITERIA_REGISTRY as Python dicts. Moving to declarative YAML is a clean and good improvement.

2) Phase-level resume (§13.3).
This is a material improvement over the current perf_v2 resume which only checks for checkpoint existence. The marker-file approach with corruption detection is practical and well-thought-out. It was my pain with current benchmark.

3) Catch-and-continue over assert-and-crash.
The current Criterion class in utils.py raises on first violation. The proposed approach of collecting all results and reporting regressions without aborting is clearly the right choice for a continuous monitoring system.

4) MLflow-based baseline resolution.
I like the solution. As we discussed previously, no separate baseline files to maintain, no version-stamped CSVs — just query MLflow for the latest successful develop run. LGTM

____
However, I have few comments / suggestions:

1. The proposed priority system (`core` / `extended` / `exploratory`) with hash-based rotation groups adds unnecessary complexity. The rotation logic (`hash(model_name) % N`, ISO week numbers, staleness detection) is clever but over-engineered for our needs — it makes the system harder to understand and debug.

Proposed alternative — align with Geti's model tiers:

| Run Cadence          | What to Run                                                                                             | Scope                          |
| -------------------- | ------------------------------------------------------------------------------------------------------- | ------------------------------ |
| **Nightly**          | The **default model per task** (as defined in Geti)                                                     | `tiny` + `small` datasets      |
| **Weekly (weekend)** | The **recommended models per task** (~3 per task, matching Geti's recommendations)                      | All dataset tiers              |
| **Pre-release**      | **One model per architecture family** (e.g., one YOLOX variant, one ATSS variant, one D-FINE variant)   | All datasets                   |
| **On-demand**        | Any other model                                                                                         | By explicit `workflow_dispatch` |

This mirrors the model tiers that already exist in Geti (default, recommended), avoids inventing a new categorization, and is immediately understandable to anyone familiar with the product. The pre-release scope is narrow enough to be feasible yet broad enough to catch family-level regressions.

2. Remove Configurable Parameter Scenarios — This Is a Regression Framework, Not an Experimentation Platform

Sections 6.3.1–6.3.6 describe an elaborate system for running benchmark variants with different learning rates, batch sizes, optimizers, precision modes, and augmentation settings. This turns the benchmark framework into a hyperparameter experimentation tool, which is not its purpose.

**The benchmark framework should answer one question:** _"Did the latest code change break any model's expected performance?"_ It should not answer _"What happens if I change the learning rate to 0.002?"_ — that is an ad-hoc experiment that belongs outside this framework.

I would recommend to remove the entire configurable-parameter scenario system (§6.3). Keep only the `default` scenario (unmodified recipe) and the `tiling` (which is a distinct recipe variant, not a parameter tweak). This eliminates:

- The `--override` / `--train-kwarg` CLI flags
- The `_resolve_overrides()` function and JSON serialization logic
- Per-scenario baselines in MLflow
- The `benchmark:params` PR label
- The `--scenario-tag configurable` machinery

Engineers who need to experiment with hyperparameters can do so locally using `OTXEngine` directly or via the existing `otx train` CLI with recipe overrides. The benchmark framework should stay focused on regression detection.

3. Schema Validation for YAML Manifests

The proposed `benchmark_catalog.yaml` and `benchmark_manifest.yaml` files become the single source of truth for the entire benchmark system. If someone introduces a typo in a model name, references a nonexistent recipe, omits a required field like `sha256`, or uses an invalid `size_tier` value — the error would only surface at runtime, potentially hours into a long benchmark run.

I would recommend to define JSON Schema (or Pydantic models) for both YAML files and add a `python -m otx.benchmark validate` command that:

- Validates all required fields are present and correctly typed.
- Verifies that every model's `recipe` path points to a file that actually exists.
- Checks that dataset names referenced in the manifest exist in the catalog.
- Runs as a lightweight CI step on every PR that modifies either YAML file.

This is low effort and prevents a class of frustrating, late-discovered errors.

4.  Address Reproducibility Limitations Before Setting Regression Thresholds

The proposal assumes `deterministic: true` will produce reproducible results across seeds (§6.1). In practice, this assumption does not hold:

- **Some operations are non-deterministic even with the flag enabled** — certain CUDA kernels (e.g., `atomicAdd`-based scatter/gather, some convolution algorithms) have no deterministic implementation and will silently fall back to non-deterministic behavior.
- **Some models are incompatible with deterministic mode** — they use operations that throw errors when `torch.use_deterministic_algorithms(True)` is set.
- **XPU does not support deterministic mode at all**, making cross-accelerator reproducibility impossible. (at least it did not in the past)
- **Tiny datasets are inherently noisy** — with fewer than 50 samples, a single mislabeled or ambiguous image has outsized impact on metrics. Variance across seeds on tiny datasets can easily reach 5–10% even under identical conditions.

If we set regression thresholds (e.g., `margin: 0.10`) without understanding the natural variance, we will either get flooded with false-positive regressions (margins too tight) or miss real regressions entirely (margins too loose).

I would recommend to conduct a variance study during Phase 2:

1. Select 3–5 representative models across tasks.
2. Run each model 5–10 times under identical conditions (same code, same dataset, same hardware).
3. Measure the standard deviation of each metric (accuracy, training time, GPU memory).
4. Set regression margins at **1.5× the observed standard deviation** per metric per dataset tier.

5. Public, Standardized Datasets

The proposal does not specify criteria for dataset selection beyond size tiers. For the benchmark results to be meaningful and reportable externally, the datasets must be:

- **Public** — freely available, not proprietary or access-restricted, so that results can be independently reproduced and shared.
- **Standardized** — ideally dataset should be widely used benchmarks with established baselines in the literature (e.g., COCO, VOC, Cityscapes subsets) so that our numbers are directly comparable to published results from other frameworks.
- **Properly licensed** — compatible with our use case (benchmarking, CI, potential publication).

For each task, we should identify at least one recognized public benchmark dataset. Document the rationale for each dataset choice (why this dataset, what it represents, what published baselines exist). This should happen before Phase 1 implementation begins

6. Move Disk Space Management to Phase 1

The proposal defers checkpoint cleanup and disk space management to Phase 3 (§16). However, training checkpoints for even small models are hundreds of megabytes. A single weekly run with ~100 experiments × 3 seeds could generate **100+ GB of artifacts**. On a self-hosted runner with limited disk, this will cause failures within the first few runs.

I would recommend to implement automatic cleanup in Phase 1 as part of the core runner loop:

- Delete training checkpoints and exported models immediately after metrics are extracted.
- Keep only `metrics.csv` and `result.json` marker files.
- Provide a `--keep-checkpoints` flag for debugging.
- Add disk space checks before each experiment and warn/abort if free space is below a threshold.

This is a small implementation effort and prevents a predictable production incident.

 7. Baseline Pinning Mechanism

The proposal's MLflow-based baseline system (§9.1.1) always compares against the most recent successful `develop` run. This has no mechanism for manual intervention, which creates problems in practice:

- **Legitimate improvement followed by revert:** If a model's accuracy improves by 15% due to a new feature, that becomes the new baseline. If that feature is later reverted, every subsequent run reports a "regression" — even though performance simply returned to the original level.
- **Known baseline corruption:** If a `develop` run succeeds but produces atypical results (e.g., due to a transient hardware issue), that bad run becomes the baseline for all future comparisons.
- **Intentional model changes:** When a model is deliberately restructured with an expected accuracy trade-off, there is no way to acknowledge this and reset the baseline.

I would recommend to add a `--baseline-run-id <mlflow_run_id>` CLI flag that allows manually specifying which MLflow run to use as the baseline. Additionally, support tagging specific MLflow runs as `is_baseline: true` so that the baseline query can prefer tagged runs over simply the most recent one. This gives the team an escape hatch for the scenarios described above.

library/docs/design/automated-model-benchmarking.md
AlbertvanHouten: I don't think we should run tests on every PR but with a change detection mechanism we can use the tiny or small dataset to quickly verify the changes. 

For weekly tests we either use small or medium and large is optional for release but ideally we can cover any regression with the smaller datasets.

library/docs/design/automated-model-benchmarking.md
AlbertvanHouten: This is all optional. I did include it here because it will be much easier to track experiments ran by different engineers as the results will all be aggregated in MLflow. The main usecase will still be to check for regressions but if you want to do tests with different configurable parameters it can be done while storing the results. 

For tiling, I agree. Just testing one task type should be good enough.

library/docs/design/automated-model-benchmarking.md
AlbertvanHouten: Yes, I am hoping this will be fixed by deterministic mode but this will need to be carefully checked.

library/docs/design/automated-model-benchmarking.md
AlbertvanHouten: I'll remove this for now. The idea was that you could assign a label to your PR which starts the workflow running several scenarios but lets not overcomplicate it at this point.

library/docs/design/automated-model-benchmarking.md
AlbertvanHouten: I don't think we should split default and recommended models but this will depend on how long testing takes. 

As for extended/on_demand, Radwan wanted to potentially supports 100s of models so I think we should have some sort of model testing rotation where we try to validate each model at least once a month. This is of course hugely dependant on how long it takes to test everything but we can easily change this logic later on.

library/docs/design/automated-model-benchmarking.md
AlbertvanHouten: Its not directly compared because not all metrics will be comparable. For example, training time will differ. I do agree that accuracy should match between GPU/XPU.

library/docs/design/automated-model-benchmarking.md
AlbertvanHouten: Yes, lets keep clear separation based on what its parsing. Each python file will have dataclasses and logic specific to the yaml file.

library/docs/design/automated-model-benchmarking.md
```diff
@@ -0,0 +1,1780 @@
...
+> dataset so that PR-level and nightly benchmarks can run quickly. `medium` and
+> `large` datasets are reserved for weekly and on-demand runs where statistical
+> significance and real-world representativeness matter more than speed.
+
+### 5.2 Schema
+
+Datasets are declared as a **flat list** — each dataset is listed once and
+referenced by name from the benchmark manifest. Since the same dataset can be
+used by multiple tasks (e.g. a multi-annotation dataset used for detection,
+segmentation, and classification), the catalog is **task-independent**. The
```
leoll2, line 205: I agree with the idea of task-independent dataset. On the other hand, I would suggest to document the type of annotation for each catalog entry, e.g.:
```yaml
  - name: pothole_tiny
    script: "scripts/benchmark_datasets/prepare_pothole.py"
    size_tier: tiny
    annotation_type: bounding_box
```
This makes it easier to determine which datasets are suitable for a certain task and which not.

library/docs/design/automated-model-benchmarking.md
```diff
@@ -0,0 +1,1780 @@
...
+for `input_size`, or a full optimizer dict for swapping `SGD` → `AdamW`).
+
+**`train_kwargs`** are passed as-is to `engine.train()`. This covers
+engine-level parameters that live outside the recipe (e.g., `precision`,
+`max_epochs`, `deterministic`).
+
+#### 6.3.2 Supported Parameter Categories
+
+The following categories of configurable parameters are first-class targets
+for scenario-based benchmarking:
```
leoll2, line 564: I don't understand this part, what does it mean in practice that a certain parameter is "first-class"?

library/pyproject.toml
```diff
@@ -57,6 +57,7 @@ dependencies = [
     "rfdetr @ git+https://github.com/kprokofi/rf-detr.git@bf69fe2d11ee6b7e0698e4a27af2081134f9b483",
     "certifi>=2026",
     "roboflow~=1.2",
+    "defusedxml>=0.7.1",
```
leoll2: Where is this dependency used?

library/pyproject.toml
```diff
@@ -138,7 +139,7 @@ Repository = "https://github.com/open-edge-platform/training_extensions/"
 
 [tool.uv]
 default-groups = ["dev", "lint"]
-required-version = "~=0.10.4"
+required-version = ">=0.10.4,<0.12"
```
leoll2: Upgrading uv is out of scope of this task, should be done in a separate PR.

library/benchmark_catalog.yaml
```diff
@@ -0,0 +1,69 @@
...
+  - name: wgisd
+    script: "scripts/benchmark_datasets/prepare_wgisd.py"
+    size_tier: small
+    description: "Wine Grape Instance Segmentation Dataset — 137 images of grape clusters with COCO-style polygon annotations (5 varieties)."
+    compatible_tasks:
+      - instance_segmentation
+      - detection # polygons → bounding boxes
+      - rotated_detection # polygons → oriented bounding boxes
+      - semantic_segmentation # polygons → pixel masks
+      - classification/multi_class_cls # crop-level or image-level class from category
```
leoll2, line 30: This relates to my other comment, where I suggest to add `annotation_type`. This property `compatible_task` solves the same problem, but I think it's less maintainable because it needs to be updated when we add a new task.

library/benchmark_manifest.yaml
```diff
@@ -0,0 +1,283 @@
...
+# regression criteria for each task.  The runner iterates over the
+# cross-product (model × dataset × scenario × seed) after applying
+# any runtime filters.
+version: 1
+defaults:
+  num_seeds: 3
+  eval_upto: optimize
+  deterministic: false # Setting to true is extremely slow for certain models
+  rotation:
+    extended_groups: 2
```
leoll2, line 13: Is the value `2` provisional or have you already tuned it based on the duration of the test suite?

library/docs/design/automated-model-benchmarking.md
```diff
@@ -0,0 +1,1780 @@
...
+---
+
+## 9. Reporting & Regression Detection
+
+### 9.1 Baseline System
+
+Every benchmark run logs all metrics to MLflow with structured tags (`branch`,
+`git_sha`, `accelerator`, `scenario`, etc.). To determine the baseline for
+comparison, the reporter simply queries MLflow for the most recent successful
+`develop` run matching the same (model, dataset, scenario, accelerator)
```
leoll2, line 1091: Comparing against the most recent successful `develop` run may not be enough, because it allows metrics to slowly degrade over the course of multiple commits without ever raising a warning. I think we should _also_ compare against the last `release` run, so that we can detect metrics that significantly dropped since the last stable version of the product.

library/src/getitune/backend/lightning/callbacks/epoch_summary.py
```diff
@@ -0,0 +1,111 @@
...
+    except (TypeError, ValueError):
+        return None
+
+
+class EpochSummary(Callback):
+    """Log a one-line summary of training/validation metrics per epoch.
+
+    The summary is emitted via the standard ``logging`` module at INFO level,
+    so it is captured by any handler the engine has configured (file logging,
+    stdout, etc.) and timestamps line up with the rest of the training log.
```
leoll2, line 52: Could you cherry-pick this change to develop in a separate PR? It's a very nice improvement that I would like to include  in the coming release.

library/src/getitune/data/dataset/keypoint_detection.py
```diff
@@ -57,7 +57,7 @@ def __init__(
             transforms=transforms,
             max_refetch=max_refetch,
         )
-        labels = dm_subset.schema.attributes["keypoints"].categories.labels
+        labels = dm_subset.label_categories.labels  # type: ignore[missing-attribute]
```
leoll2: Just double-checking: all the other tasks have have their label categories nested inside `attributes["label"]`, while keypoint detection has them inside `attributes["keypoints"]`. Is it correct?

library/src/getitune/recipe/instance_segmentation/rtmdet_inst_tiny.yaml
```diff
@@ -6,11 +6,11 @@ model:
     model_name: rtmdet_inst_tiny
 
     optimizer:
-      class_path: torch.optim.SGD
+      class_path: torch.optim.AdamW
```
leoll2, line 9: Changing the optimizer is quite a big change, and unrelated to this PR main focus. If you found that RTMDet trains better with AdamW than SGD, please open a separate PR to make this change and share your benchmarks there.

library/src/getitune/benchmark/docker-compose.yaml
leoll2, line 1: I would rename this file to `mlflow-server-docker-compose.yaml` to make the scope clearer.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,292 @@
+# OTX Model Benchmark Workflow
```
leoll2: Please replace "OTX" references with "getitune", also in other parts of this file.

library/src/getitune/benchmark/docker-compose.yaml
```diff
@@ -0,0 +1,81 @@
+# MLflow Tracking Server for OTX Benchmark
```
leoll2, line 1: Please replace "OTX" references with "getitune", also in other parts of this file.

leoll2 (commented):
The architecture looks solid, great work. I especially like the reporting part and the tiering+rotation system.

library/src/getitune/data/dataset/keypoint_detection.py
AlbertvanHouten: That was the case but it doesn't matter anymore which field has the label categories. The label_categories property of the datumaro dataset will find which field has a LabelCategories attached and return them. Previous logic was brittle as a minor field name difference could break this line.

library/src/getitune/recipe/instance_segmentation/rtmdet_inst_tiny.yaml
AlbertvanHouten: I was working on this here https://github.com/open-edge-platform/training_extensions/pull/6418

library/src/getitune/benchmark/docker-compose.yaml
AlbertvanHouten: I've moved the compose file to the benchmark folder instead

library/benchmark_catalog.yaml
AlbertvanHouten: Chances of that are pretty slim IMO. Even if we add a new task there might also be a new annotation type so it is not a guaranteed fix.

library/benchmark_manifest.yaml
AlbertvanHouten: It is still provisional

library/docs/design/automated-model-benchmarking.md
AlbertvanHouten: Good point. I think it would be best to implement some sort of drift detection that checks the last X runs as well as comparing against a baseline commit which can be from the release branch. Lets add that in a follow up PR though as this one is already very large.

.github/workflows/lib-lint-and-test.yaml
```diff
@@ -178,13 +190,54 @@ jobs:
...
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      # TODO remove after switching to an official Datumaro release with pre-built wheels
+      - name: Installing Rust toolchain
+        uses: dtolnay/rust-toolchain@3c5f7ea28cd621ae0bf5283f0e981fb97b8a7af9
+        with:
+          toolchain: stable
```
leoll2, line 215: Can be removed now, right?

leoll2 (dismissed):

library/src/getitune/backend/lightning/models/keypoint_detection/base.py
```diff
@@ -68,6 +68,16 @@ def __init__(
             torch_compile=torch_compile,
         )
 
+    def _reset_prediction_layer(self, num_classes: int) -> None:
```
kprokofi: What is this for?

library/src/getitune/backend/lightning/models/keypoint_detection/base.py
```diff
@@ -107,19 +117,14 @@ def _customize_outputs(
             if inputs.imgs_info[i] is None:
                 msg = f"The image information for the image {i} is not provided."
                 raise ValueError(msg)
-            # scale to the original image size
```
kprokofi, line 110: Did you validate that it can be safely removed? We decode key points to original input size for proper validation

library/src/getitune/backend/openvino/engine.py
```diff
@@ -186,6 +186,30 @@ def _derive_task_from_onnx(self, onnx_path: PathLike) -> TaskType:
...
+
+        Args:
+            model: The OpenVINO model.
+            datamodule: The data module.
+            pipeline: Name of the pipeline (e.g. "test" or "predict") for the error message.
+
+        Raises:
+            ValueError: If label info differs and the task is not keypoint detection.
+        """
+        if model.label_info != datamodule.label_info and model.task != TaskType.KEYPOINT_DETECTION:
```
kprokofi: Why this "model.task != TaskType.KEYPOINT_DETECTION" exception? I would rather to check on the dataset level that label_info are the same between all subsets and then check only once with model.label_info during model creation

library/src/getitune/benchmark/catalog.py
```diff
@@ -0,0 +1,223 @@
...
+    behind by a crashed prep run is therefore treated as "not ready" and
+    the script is re-executed.
+
+    Returns the path to the prepared dataset directory.
+    """
+    dataset_dir = data_root / entry.relative_path
+    ready_marker = dataset_dir / ".ready"
+
+    if ready_marker.exists():
+        logger.info("Dataset '%s' already exists, skipping.", entry.name)
```
kprokofi, line 181: minor: I would use f-strings instead of old string formatting

library/benchmark_catalog.yaml
```diff
@@ -0,0 +1,69 @@
...
+    description: "Oxford Flowers 102 — 8,189 images across 102 flower categories with official train/val/test splits."
+    compatible_tasks:
+      - classification/multi_class_cls
+
+  - name: oxford_pets_hlabel
+    script: "scripts/benchmark_datasets/prepare_oxford_pets_hlabel.py"
+    size_tier: large
+    description: "Oxford-IIIT Pet with hierarchical labels — 7,349 images with natural 2-level hierarchy: species (cat/dog) → 37 breeds."
+    compatible_tasks:
+      - classification/h_label_cls
```
kprokofi, line 46: we don't need to run benchmarks for  h_label_cls and rotated_detection. These tasks are obsolete

library/benchmark_catalog.yaml
```diff
@@ -0,0 +1,69 @@
...
+# All datasets are stored in the experimental Datumaro format.
+#
+# Fields:
+#   name        – unique identifier (referenced from benchmark_manifest.yaml)
+#   script      – path to preparation script (relative to library/ root)
+#   size_tier   – tiny | small | medium | large
+
+version: 1
+
+datasets:
```
kprokofi, line 19: Previously we agreed to collect several tiny and small datasets for a weekly quick run. However, here I see all datasets except wgisd are relatively large. Also, VOC is not a great choice to benchmark models pretrained on COCO

kprokofi: So, what I wanted to say,  that we need to reconsider datasets choice

library/scripts/benchmark_datasets/prepare_voc2007.py
```diff
@@ -0,0 +1,197 @@
...
+                "name": name,
+                "bbox": [xmin, ymin, xmax - xmin, ymax - ymin],
+                "difficult": difficult,
+            }
+        )
+
+    return {"filename": filename, "width": width, "height": height, "objects": objects}
+
+
+def _build_dataset(voc_root: Path) -> Dataset:
```
kprokofi: Why datumaro doesn't support conversions between data formats? I thought it is the most useful feature of the dataumaro. However, here I see a lot of custom code to convert from VOC to COCO / Parquet  dataset

kprokofi (commented):
First bunch of comments

library/benchmark_manifest.yaml
```diff
@@ -0,0 +1,280 @@
...
+  detection:
+    models:
+      - name: atss_mobilenetv2
+        priority: core
+        recipe: detection/atss_mobilenetv2.yaml
+      - name: yolox_s
+        priority: core
+        recipe: detection/yolox_s.yaml
+      - name: dfine_x
+        priority: core
```
kprokofi, line 83: please, align with current default models. We don't use atss and fine_x as core models

library/scripts/benchmark_datasets/prepare_voc2007.py
AlbertvanHouten: Datumaro can do that but I didn't pay too much attention to these dataset preparation scripts other than that they created good datasets that can be used for training. They are generated with AI.

I can convert it to load using VOC but it is working fine like this.

library/benchmark_catalog.yaml
AlbertvanHouten: Yes, lets add more in the future. I didn't get too great results with smaller datasets, even WGISD for segmentation isn't working that well. Using small datasets which result in lower accuracy scores also mean its harder to track drops in performance. For now, these datasets are able to run weekly with 2 rotation groups.

library/src/getitune/benchmark/catalog.py
AlbertvanHouten: It is best practice to not use f-strings in loggers https://docs.astral.sh/ruff/rules/logging-f-string/

.github/workflows/lib-lint-and-test.yaml
AlbertvanHouten: Yes, I created an issue for this so it can be resolved for all workflows at once. https://github.com/open-edge-platform/training_extensions/issues/6740

library/benchmark_manifest.yaml
AlbertvanHouten: done

library/benchmark_catalog.yaml
AlbertvanHouten: removed

library/src/getitune/backend/openvino/engine.py
AlbertvanHouten: reverted as keypoint detection now is no longer benchmarked

library/src/getitune/benchmark/cli.py
```diff
@@ -0,0 +1,481 @@
...
+        help="MLflow tracking URI (local path or remote server URL).",
+    )
+    parser.add_argument(
+        "--branch",
+        type=str,
+        default="",
+        help="Git branch name (auto-detected if empty).",
+    )
+    parser.add_argument(
+        "--trigger",
```
kprokofi, line 72: Can you explain, why should we add trigger command here instead of relying on GitHub actions orchestration? By the design, how should this script run?

library/benchmark_manifest.yaml
```diff
@@ -0,0 +1,214 @@
...
+        priority: extended
+        recipe: detection/yolox_tiny.yaml
+      - name: yolox_l
+        priority: extended
+        recipe: detection/yolox_l.yaml
+      - name: yolox_x
+        priority: extended
+        recipe: detection/yolox_x.yaml
+      - name: rtdetr_50
+        priority: extended
```
kprokofi, line 110: We have two groups here, core: default models and extended. I propose to add a family property to run tests based on family. For example, it is better to run 3 models from different family to test instead of 3 from one like core for Instance Seg

library/src/getitune/benchmark/README.md
```diff
@@ -0,0 +1,200 @@
...
+2. See all benchmark commands:
+
+   ```bash
+   python -m getitune.benchmark --help
+   ```
+
+3. Provision datasets referenced by the catalog/filters:
+
+   ```bash
+   python -m getitune.benchmark provision
```
kprokofi, line 39: provision here means  .yaml file?

library/src/getitune/benchmark/README.md
```diff
@@ -0,0 +1,200 @@
...
+```bash
+python -m getitune.benchmark run --task detection --model yolox_s --dataset wgisd --override model.init_args.optimizer.init_args.lr=0.01 --max-epochs 10 --train-kwarg precision=32 --no-tracking
+```
+
+## Report and cleanup commands
+
+Regenerate a report from tracked MLflow runs:
+
+```bash
+python -m getitune.benchmark report --mlflow-uri http://localhost:5000 --output-root results --accelerator <cpu|cuda|xpu>
```
kprokofi, line 101: What artifacts will be used to generate a report? current folder?

library/src/getitune/benchmark/README.md
```diff
@@ -0,0 +1,200 @@
...
+## Add a new dataset or model
+
+Use this flow when extending benchmark coverage.
+
+### Add a new dataset
+
+1. Add a dataset provisioning script under `scripts/benchmark_datasets/` (for example, `prepare_my_dataset.py`).
+2. Add an entry in `benchmark_catalog.yaml` that points `script` to that provisioning script.
+3. Reference the dataset name from the relevant task in `benchmark_manifest.yaml` under `experiments.<task>.datasets`.
+4. Provision the dataset.
```
kprokofi, line 136: Can we omit this step and automatically prepare a dataset during run if it is not prepared?

library/src/getitune/benchmark/README.md
```diff
@@ -0,0 +1,200 @@
+# GetiTune Benchmarking
```
kprokofi, line 1: Please, add to this README a schedule how we will run models on CI, which should include: dataset size, models group / categories, timeline. It is needed to everyone understand how and when something tested.

library/src/getitune/benchmark/README.md
```diff
@@ -0,0 +1,200 @@
...
+
+- `data/<dataset>/.ready`
+
+## Add a new dataset or model
+
+Use this flow when extending benchmark coverage.
+
+### Add a new dataset
+
+1. Add a dataset provisioning script under `scripts/benchmark_datasets/` (for example, `prepare_my_dataset.py`).
```
kprokofi, line 133: Can I run my local dataset? In Parquet format, for example. We should be able to run experiments without scripts how to prepare them.  Ideally in academic formats, but at least in datumaro

kprokofi (commented):
Thank you for your work. Generally, looks very good according to our plan! Unfortunately, I didn't have time to properly test every functionality, but I hope you tested it properly.

Few points:
1. we need to establish and fix datasets list before real validation on CI. Right now, I think datasets are not appropriate. Pets and Flowers are too generic and easy datasets, VOC is not a good choice for Detection models trained on COCO, also we lack of small - tiny datasets. Let's discuss separately regarding datasets choice. 
2. I would like to see a detailed test schedule in documentation (README). It will be helpful for every developer
3. It would be nice if local datasets are supported avoiding scripts step

library/src/getitune/benchmark/README.md
AlbertvanHouten: Yes, it will do that. Every run checks if the needed datasets are available and otherwise prepares them.

library/src/getitune/benchmark/cli.py
AlbertvanHouten: Normal users won't need to bother with this trigger argument. The CI will pass it so we can separate manual runs from weekly runs for report generation. The trigger is basically the last part of an experiments name like manual in the below picture.

Image

library/src/getitune/benchmark/README.md
AlbertvanHouten: It means creating the datasets defined in the dataset catalog

library/src/getitune/benchmark/README.md
AlbertvanHouten: The results produced by the experiments. By default, they will be placed in `results` directory. Additionally it also queries mlflow to compare against previous results.

library/src/getitune/benchmark/README.md
AlbertvanHouten: I will create an issue as a future improvement for this.

library/benchmark_manifest.yaml
AlbertvanHouten: Will address this in a future improvement PR https://github.com/open-edge-platform/training_extensions/issues/6750

library/src/getitune/benchmark/README.md
AlbertvanHouten: The automation will live in another repo. Shall I add it there? https://github.com/intel-innersource/applications.ai.geti.geti-release/pull/349/changes#diff-960d2e68401a7f938a95be8a307bd155d0edab62da7b0f8106f55ff998d3fc08

AlbertvanHouten: I added a small section here too https://github.com/open-edge-platform/training_extensions/pull/6056/commits/11cf453eca0c7f5099fecce4f5ab07afd4c7af33

kprokofi (dismissed):

AlexanderBarabanov (dismissed):

leoll2 (dismissed):

kprokofi (approved):

AlexanderBarabanov (approved):

PR #6058, opened 2026-04-08
Include number of annotated frames for a video in media listing

### LINKED ISSUE #6050: Number of annotated frames for a video
`/dataset/media` returns both the total number of frames and the number of annotated frames.

Image

itallix:
### Summary

- [x] Add `annotated_frame_count` to `VideoView`

### Checklist

### DIFF
--- a/application/backend/app/services/media_service.py
+++ b/application/backend/app/services/media_service.py
@@ -252,7 +252,13 @@ class MediaService(BaseSessionManagedService):
         db_media = repo.get_by_id(str(media_id))
         if not db_media:
             raise ResourceNotFoundError(ResourceType.MEDIA, str(media_id))
-        return MediaAdapter.validate_python(db_media)
+        return (
+            Video.model_validate(db_media).model_copy(
+                update={"annotated_frame_count": repo.count_annotated_video_frames_by_video_id(db_media.id)}
+            )
+            if db_media.type == MediaType.VIDEO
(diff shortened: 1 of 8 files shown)


application/backend/app/services/media_service.py
```diff
@@ -247,7 +252,13 @@ def get_media_by_id(self, project_id: UUID, media_id: UUID) -> Media:
         db_media = repo.get_by_id(str(media_id))
         if not db_media:
             raise ResourceNotFoundError(ResourceType.MEDIA, str(media_id))
-        return MediaAdapter.validate_python(db_media)
+        return (
+            Video.model_validate(db_media).model_copy(
```
A-Artemis, line 256: Having two different models (video and MediaAdapter) feels strange for just a media return endpoint.

I think this should be revisited later to have 1 Media return model.

A-Artemis (approved):
LGTM

application/backend/app/services/media_service.py
itallix: Yeah, it looks a bit weird.
But we are dealing with 3 different models here, `MediaAdapter` performs conversion based on `type` field:

```
Media = Annotated[Image | Video | VideoFrame, Field(discriminator="type")]
```

In case of videos, we still need to enrich the db model with additional attribute. That's why we end up with this construct.

camiloHimura (approved):

PR #6080, opened 2026-04-10
fix: remove outdated import/export jobs

### LINKED ISSUE #5919: Dataset import messages surprise on the projects page
I could observe it several times. I cannot say what are the exact steps to reproduce. 
One that I have gathered listed below:
1. Start the app `just run-image`
2. Import a dataset. 
3. When the import finishes (succeefully), close the message the appears on top in your work area.
4. Do other stuff.
5. Shut down the app. (This step is not required, it's just the step which I've done recently.)
6. Start the app `just run-image`

The project page shows this several hours old, already closed, message.
Image

camiloHimura:
### Summary

closes #5919

This PR removes import/export LS job entries when the ID is no longer valid.

### Checklist

### DIFF
--- a/application/ui/src/components/loading-import-dataset/loading-import-dataset.component.tsx
+++ b/application/ui/src/components/loading-import-dataset/loading-import-dataset.component.tsx
@@ -4,7 +4,7 @@
 import { toast } from '@geti/ui';
 import { useImportJobStatus } from 'hooks/api/jobs/use-import-job-status.hook';
 import { useDeleteStagedDataset } from 'hooks/api/staged-dataset.hook';
-import { isJobFailed, isJobPending, isJobRunning } from 'hooks/api/util';
+import { isInvalidJob, isJobFailed, isJobPending, isJobRunning } from 'hooks/api/util';
 
 import { formatBytes } from '../../shared/util';
 import { ImportActiveJob } from '../import-card-status/import-active-job/import-active-job.component';
@@ -27,23 +27,25 @@ export const LoadingImportDataset = ({
     onSuccess,
     deleteEntry,
 }: LoadingImportDatasetProps) => {
-    const deleteFileMutation = useDeleteStagedDataset({ stagedDatasetId });
+    const deleteStagedFileMutation = useDeleteStagedDataset({ stagedDatasetId });
 
-    const {
-        error,
-        isError,
-        data: job,
-    } = useImportJobStatus({
+    const { data: job, ...response } = useImportJobStatus({
         jobId,
         onSuccess: () => {
-            deleteEntry();
-            deleteFileMutation.mutate();
             onSuccess();
+            deleteEntry();
+            deleteStagedFileMutation.mutate();
 
             toast({
                 message: `Dataset ${fileName} ${formatBytes(size)} imported successfully.`,
                 type: 'success',
             });
+        },
+        onError: (error) => {
+            if (isInvalidJob(error)) {
+                deleteEntry();
+                deleteStagedFileMutation.mutate();
+            }
         },
     });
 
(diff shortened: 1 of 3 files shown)


application/ui/src/components/loading-import-dataset/loading-import-dataset.test.tsx
```diff
@@ -47,7 +47,7 @@ describe('LoadingImportDataset', () => {
         );
     };
 
-    it('renders failed job state and displays job message and error', async () => {
+    it('displays job message, error, and does not delete entry when job fails', async () => {
```
dwesolow, line 51: Why we don't remove the entry (staged-file) when import job fails? When that entry is removed?

camiloHimura: The requirement is to remove it when the job is invalid (line 95). For any other errors, we should display the error information so the user knows what happened. When they manually close the card, the staged file is removed.

application/ui/src/hooks/api/jobs/use-import-job-status.hook.tsx
```diff
@@ -31,17 +31,17 @@ export const useImportJobStatus = ({ jobId, onError, onSuccess }: UseImportJobSt
...
     }, [onError, response.error, response.isError]);
 
     useEffect(() => {
         if (isJobFailed(response.data)) {
-            isFunction(onError) && onError();
+            isFunction(onError) && onError(response.error);
             toast({ type: 'error', message: `An error occurred during import. ${response.data?.message}` });
         }
-    }, [onError, response.data]);
+    }, [onError, response.data, response.error]);
```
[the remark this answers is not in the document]
camiloHimura: updated

application/ui/src/components/loading-import-dataset/loading-import-dataset.component.tsx
```diff
@@ -27,24 +27,22 @@ export const LoadingImportDataset = ({
...
-        error,
-        isError,
-        data: job,
-    } = useImportJobStatus({
+    const { data: job, ...response } = useImportJobStatus({
         jobId,
         onSuccess: () => {
             deleteEntry();
             deleteFileMutation.mutate();
-            onSuccess();
```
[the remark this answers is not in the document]
camiloHimura: updated

application/ui/src/components/loading-import-dataset/loading-import-dataset.test.tsx
dwesolow: I understand that but the question is why we don't remove the entry (staged-file) when import job fails?

camiloHimura: To avoid having the local storage entry point to a staged file that no longer exists, it could be removed when an error occurs. However, I think it should remain until all of its references have been cleared.

application/ui/src/components/loading-import-dataset/loading-import-dataset.component.tsx
```diff
@@ -27,24 +27,26 @@ export const LoadingImportDataset = ({
...
+            deleteEntry();
+            deleteStagedFileMutation.mutate();
 
             toast({
                 message: `Dataset ${fileName} ${formatBytes(size)} imported successfully.`,
                 type: 'success',
             });
         },
+        onError: (error) => {
+            if (isInvalidJob(error)) {
```
jpggvilaca, line 45: if we're doing it onError and onsuccess, can we not do it on onSettled?

camiloHimura: It’s possible, but I’d prefer not to. Under the hood, we’re using `useEffect` to trigger those events, and I want to avoid adding more than strictly necessary.

jpggvilaca: i meant using onSettled: () => {
   ...the 2 lines that are repeated
}

wont it work?

camiloHimura: Yes, but we’d also need to add an extra event and additional validation, since `deleteEntry` and `deleteStagedFileMutation.mutate` are not called for every error

jpggvilaca (approved):

PR #6103, opened 2026-04-14
Empty prediction

### LINKED ISSUE #6097: Empty prediction is not shown
**Steps to reproduce**

1. Train a detection / segmentation model
2. Upload an unannotated image, preferably unrelated to the training dataset (to increase the likelihood of empty prediction)
3. Open the image in the annotator in 'Prediction' mode

**Expected**

Server returns an empty prediction:

Image

UI interprets such server response as "No object", hence displays the empty label.

**Actual**

The UI doesn't doesn't correctly interpret the `predictions: []` response, and no label is displayed. From user perspective, it is confusing because it looks like the prediction was not requested at all.

Image

jpggvilaca:
### Summary

* Display "empty" prediction when media's `prediction` array is empty (same behavior as with annotations)

Screenshot 2026-04-14 at 10 18 52

* Closes https://github.com/open-edge-platform/training_extensions/issues/6097

### Checklist

### DIFF
--- a/application/ui/src/features/annotator/api/use-media-predictions.ts
+++ b/application/ui/src/features/annotator/api/use-media-predictions.ts
@@ -6,6 +6,7 @@ import { useProjectIdentifier } from 'hooks/use-project-identifier.hook';
 
 import { fetchClient } from '../../../api/client';
 import { PredictionVideoRangePayload } from '../../../constants/shared-types';
+import { EMPTY_LABEL_ID } from '../../../shared/annotator/labels';
 
 export const mediaPredictionsQueryOptions = ({
     projectId,
@@ -34,7 +35,25 @@ export const mediaPredictionsQueryOptions = ({
             });
 
             if (response.error) return [];
-            return response.data?.predictions ?? [];
+
+            const predictions = response.data?.predictions ?? [];
+
+            return predictions.map((predictionItem) => {
+                if (predictionItem.prediction.length === 0) {
+                    return {
+                        ...predictionItem,
+                        prediction: [
+                            {
+                                shape: { type: 'full_image' },
+                                labels: [{ id: EMPTY_LABEL_ID }],
+                                confidences: null,
+                            },
+                        ],
+                    };
+                }
+
+                return predictionItem;
+            });
         },
         enabled: modelId !== undefined,
     });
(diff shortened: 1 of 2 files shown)


application/ui/src/features/annotator/api/use-media-predictions.ts
```diff
@@ -34,7 +35,25 @@ export const mediaPredictionsQueryOptions = ({
...
+
+            return predictions.map((predictionItem) => {
+                if (predictionItem.prediction.length === 0) {
+                    return {
+                        ...predictionItem,
+                        prediction: [
+                            {
+                                shape: { type: 'full_image' },
+                                labels: [{ id: EMPTY_LABEL_ID }],
+                                confidences: null,
```
dwesolow: I'm not sure if that should be `null`. If server does not return any prediction does it mean that they are 100% sure that there is no object or they didn't detect anything?
If the first one then we cannot return `null` as confidences. When `confidences` is `null` then we don't display styles and indicator for prediction (then that prediction is displayed as annotation)

leoll2: > If server does not return any prediction does it mean that they are 100% sure that there is no object or they didn't detect anything?

At inference time, models don't distinguish between these two situations. For example, a detection model that doesn't detect any object in the image simply returns an empty response: you can't tell whether it is a false negative (i.e. there is actually an object in the image, but the model failed to find it) or there's actually no object in the image.

From user perspective, what is important is to tell the user that the inference request was successfully processed but the response is empty (nothing detected). An idea could be to call the empty label "No object detected" instead of "No object" in Prediction mode, what do you think?

jpggvilaca: I think for consistency i'd leave the label name as is. I think Dawid's concern was the annotation styles. I will adjust the `confidences` to make sure we have the correct styles. Thanks both

jpggvilaca: Updated

application/ui/src/features/annotator/api/use-media-predictions.ts
```diff
@@ -34,7 +35,25 @@ export const mediaPredictionsQueryOptions = ({
...
+                            },
+                        ],
+                    };
+                }
```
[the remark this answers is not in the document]
jpggvilaca: Updated

dwesolow (approved):

PR #6123, opened 2026-04-15
Stream updates

jpggvilaca:
### Summary

* Improve layout of stream & icons
* Improve video dimension calculation (get rid of util hook)
* Get rid of overlay and add animation for pause action

Screenshot 2026-04-16 at 10 56 30

### Checklist

### DIFF
--- a/application/ui/src/features/inference/stream/stream-container.tsx
+++ b/application/ui/src/features/inference/stream/stream-container.tsx
@@ -1,6 +1,8 @@
 // Copyright (C) 2025 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 
+import { useState } from 'react';
+
 import { dimensionValue, Flex, Loading, Text, toast, View } from '@geti/ui';
 import { Pause, Play } from '@geti/ui/icons';
 import { clsx } from 'clsx';
@@ -20,12 +22,16 @@ export const StreamContainer = () => {
     const isConnecting = status === 'connecting';
     const isConnected = status === 'connected';
 
+    const [isPaused, setIsPaused] = useState(false);
+
     const canStart = isStopped && isPipelineRunning;
 
     const handleClick = async () => {
         if (isConnected) {
+            setIsPaused(true);
             await stop();
         } else if (canStart) {
+            setIsPaused(false);
             await start();
 
             if (webRTCConnectionRef.current?.getStatus() === 'failed') {
@@ -37,7 +43,7 @@ export const StreamContainer = () => {
     return (
         <View gridArea={'canvas'} overflow={'hidden'} maxHeight={'100%'}>
             <div className={classes.canvasContainer} onClick={handleClick}>
-                <View backgroundColor={'gray-200'} height={'100%'}>
+                <View height={'100%'} position='relative'>
                     {isStopped && (
                         <Flex alignItems={'center'} justifyContent={'center'} height='100%'>
                             <Flex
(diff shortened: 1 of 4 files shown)


application/ui/src/features/inference/stream/stream.tsx
```diff
@@ -84,37 +55,16 @@ const useStreamToVideo = () => {
...
-                        height={size.height}
-                        controls={false}
-                        style={{
-                            background: 'var(--spectrum-global-color-gray-200)',
-                        }}
-                    />
-                )}
-            </View>
-        </ZoomTransform>
+        <View gridArea={'innercanvas'} width={'100%'} height={'100%'}>
```
dwesolow: Why did u remove the zoom?

dwesolow:
I think that we didn't solve the root issue. The root issue was that when user hovers over the canvas, pause icon is displayed (no matter if there is a opacity etc). I thought that we agreed that we will show the button for some time (like for 3s) and the button disappears.

application/ui/src/features/inference/stream/stream.tsx
jpggvilaca: We can still have it even without the hook calculation, but i dont see why we would need zoom on a video. Youtube doesnt support this for instance

jpggvilaca:
> I think that we didn't solve the root issue. The root issue was that when user hovers over the canvas, pause icon is displayed (no matter if there is a opacity etc). I thought that we agreed that we will show the button for some time (like for 3s) and the button disappears.

I understand the flash argument as we discussed, but after playing with it, it requires adding several lines of code, useEffects and whatnot, to have such a tiny feature. Lowering the opacity helps tremendously. You can test a bit. I will counter-argue about the 3s flashy behavior

application/ui/src/features/inference/stream/stream.tsx
dwesolow: Yeah but we don't want full you tube behavior, I'd ask leaders/management before removing this.
@leoll2 wdyt?

yvolokitin (dismissed):

application/ui/src/features/inference/stream/stream.tsx
leoll2: Zoom could be useful to magnify objects and annotations that look small in the input stream. Not a critical feature, but a nice to have if it doesn't have side effects.

jpggvilaca: Reverted. readded zoom

application/ui/src/features/inference/stream/stream-container.tsx
```diff
@@ -14,42 +14,53 @@ import { useWebRTCConnection } from './web-rtc-connection-provider';
...
+
     const canStart = isStopped && isPipelineRunning;
-    const handleClick = isConnected ? stop : canStart ? start : undefined;
 
-    useEffect(() => {
-        if (status === 'failed') {
-            toast({ type: 'error', message: 'Failed to connect to the stream' });
+    const handleClick = async () => {
+        if (isConnected) {
+            setIsPaused(true);
```
dwesolow, line 31: Why do we need state for `isPaused`? Could we use `status`, maybe `isStopped` for this? IIRC we don't have a `pause` support, we just start or stop connection, no?

jpggvilaca: For the animation

application/ui/src/features/inference/stream/stream-container.tsx
```diff
@@ -14,42 +14,53 @@ import { useWebRTCConnection } from './web-rtc-connection-provider';
...
                                 <Play
                                     color={'currentColor'}
-                                    width={dimensionValue('size-800')}
-                                    height={dimensionValue('size-800')}
+                                    width={dimensionValue('size-400')}
+                                    height={dimensionValue('size-400')}
                                     aria-label={isPipelineRunning ? 'Start stream' : 'Enable pipeline to start stream'}
                                     aria-disabled={!isPipelineRunning}
                                 />
+                                <Text>Start stream</Text>
```
dwesolow: I think that button is self explanatory, no?

dwesolow (dismissed):

application/ui/src/features/inference/stream/stream-container.tsx
jpggvilaca: Imho looks much better. The icon is small. Plus it gives more real state to click. You can try locally the difference. We could make the icon bigger but i prefer it this way. Let's try and see what feels best. I also asked Yura to try

yvolokitin (dismissed):

dwesolow (approved):

PR #6127, opened 2026-04-15
Min/Max values for float range sliders

### LINKED ISSUE #6094: Return min_value and max_value for float_range params, when needed
For some parameters of type `float_range`, the server should return a `min_value` and/or `max_value` property to set the lower/upper constraint of the range, if it must be constrained.

For example, `random_zoom_out.side_range` should have `min_value=1.0` as per description. For this specific parameter, there is technically no upper bound, but we may still set a reasonably high maximum (e.g. 16) for the sake of UX - if and only if we provide both lower and upper bounds, UI can render the parameter as a slider.

Image

Review the other float range parameters and adjust their min/max accordingly. To determine their values, check the parameter description (e.g. if it explicitly states that min value is 1) or set a reasonable value based on the parameter meaning and the defaults set in the model recipes.

Note: there is a known UI bug concerning the visualization of `float_range` parameters #6093

A-Artemis:
### Summary

Adds the range bounds for float range sliders.

### Checklist

### DIFF
--- a/application/backend/app/api/schemas/training_configuration.py
+++ b/application/backend/app/api/schemas/training_configuration.py
@@ -88,6 +88,8 @@ class FloatRangeParameterView(_BaseConfigurableParameterView):
     value_type: Literal["float_range"] = "float_range"
     value: tuple[float, float] = Field(title="Actual value of the parameter")
     default_value: tuple[float, float] = Field(title="Default value of the parameter")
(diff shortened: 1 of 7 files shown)


leoll2 (commented):
LGTM, thanks @A-Artemis. Does it work out of the box in the UI, or need some changes by @dwesolow ?

jpggvilaca (dismissed):

camiloHimura (dismissed):

leoll2 (approved):

PR #6135, opened 2026-04-15
Rename OTX classes and remaining OTX references 

### LINKED ISSUE #6118: Rename OTX Entities to match new Geti name
Rename all OTX classes and remanning references

kprokofi:
All OTX* prefixes were replaced with names that reflect either the library used (Lightning* for PyTorch Lightning-backed classes) or the domain (Detection*, Segmentation*, etc.). 
It is done according to the backends concept.
Generic infrastructure classes (OTXDataModule → DataModule, OTXTaskType → TaskType, …) drop the prefix entirely since the package namespace already provides the context.

backend/native → backend/lightning
The sub-package was named native for historical reasons. Renaming it to lightning makes the dependency on PyTorch Lightning explicit and consistent with the class names above.

backend/lightning/lightning/ flattened
accelerators/ and strategies/ were one level too deep after the directory rename, producing the redundant import path backend.lightning.lightning.*. They were moved up one level.

Checkpoint keys & DataModule methods
Internal string keys (otx_version, otx_detection_model, …) and DataModule helper methods (from_otx_datasets, _setup_otx_dataset) were updated to match the new naming convention.

Resolves https://github.com/open-edge-platform/training_extensions/issues/6118


application/backend/app/api/schemas/evaluation.py
```diff
@@ -23,7 +23,7 @@ class EvaluationMetricName(StrEnum):
     MAR_100 = "mAR@100"
 
 
-# Mapping from raw metric names (returned by OTX) to user-friendly metric names (returned by the API).
+# Mapping from raw metric names (returned by Geti Tune) to user-friendly metric names (returned by the API).
```
leoll2:
```suggestion
# Mapping from raw metric names (returned by getitune) to user-friendly metric names (returned by the API).
```
It's better to refer to the library as 'getitune'

application/backend/app/execution/common/geti_config_converter.py
```diff
@@ -187,7 +187,7 @@ class ModelStatus(str, Enum):
 class TransformsUpdater:
     """Handles augmentation updates for the new CPU/GPU augmentation pipeline.
 
-    Maps Geti augmentation names to OTX/kornia/torchvision class paths and the
+    Maps Geti augmentation names to Geti Tune/kornia/torchvision class paths and the
```
leoll2: Same here, let's replace all usages of "Geti Tune" with "getitune"

kprokofi: Yes, I will replace

application/backend/app/api/schemas/evaluation.py
kprokofi: Done

application/backend/app/execution/common/getitune_converters.py
```diff
@@ -12,17 +12,17 @@
...
-    """Map internal Task to OTXTaskType."""
+def get_getitune_task_type_by_task(task: Task) -> GetiTuneTaskType:
+    """Map internal Task to GetiTuneTaskType."""
     match task.task_type:
         case TaskType.CLASSIFICATION:
             if task.exclusive_labels:
-                return OTXTaskType.MULTI_CLASS_CLS
-            return OTXTaskType.MULTI_LABEL_CLS
+                return GetiTuneTaskType.MULTI_CLASS_CLS
+            return GetiTuneTaskType.MULTI_LABEL_CLS
```
AlbertvanHouten, line 21: Some of these names are pretty verbose. Just task type would be enough IMO

kprokofi: You are right and previously I did it. However, it turned out that in Geti we already have this name and it conflicts with this one. So, I decided to make it GetiTune...

kprokofi: @leoll2 what do you think about the name choice in application?

AlbertvanHouten: I don't think we should care about what Geti already has, they can rename the import if needed.

kprokofi: In Library I use just TaskType, I import here in application file as GetiTuneTaskType

kprokofi: I follow this:

All OTX* prefixes were replaced with names that reflect either the library used (Lightning* for PyTorch Lightning-backed classes) or the domain (Detection*, Segmentation*, etc.).
It is done according to the backends concept.
Generic infrastructure classes (OTXDataModule → DataModule, OTXTaskType → TaskType, …) drop the prefix entirely since the package namespace already provides the context.

AlbertvanHouten: ah sorry, I didn't see this was in the app. Makes sense then

AlbertvanHouten:
There are a couple files still mentioning otx which I think should be updated such as: 

.codecov.yaml
CODEOWNERS
CONTRIBUTING.md
documentation referencing otx imports

kprokofi:
> There are a couple files still mentioning otx which I think should be updated such as:
> 
> .codecov.yaml CODEOWNERS CONTRIBUTING.md documentation referencing otx imports

Will review:
.codecov.yaml
CODEOWNERS
CONTRIBUTING.md

Current documentation is subject to be deleted. It is outdated and new new one will be prepared

kprokofi:
@AlbertvanHouten done

AlbertvanHouten (dismissed):

leoll2 (dismissed):

leoll2:
Great work, ty

leoll2 (approved):

PR #6137, opened 2026-04-16
fix: Display augmentation parameters when dependent parameter does not exist

### LINKED ISSUE #6132: Missing augmentation options
**Steps to reproduce**

1. Open the demo detection project
2. Click 'Train model'
3. Click 'Advanced settings'
4. Scroll down to the augmentation parameters

**Expected**

The augmentation parameters are correctly displayed and the user can configure them.

**Actual**

The augmentation section is unexpectedly empty.

Image

dwesolow:
### Summary

The issue was that implementation assumed that we always have dependent parameter. That was wrong assumption and this PR fixes that.

Fix #6132 

### Checklist

- [X] All related issues are linked to this PR (if applicable)

### DIFF
--- a/application/ui/src/features/models/train-model/advanced-settings/utils.ts
+++ b/application/ui/src/features/models/train-model/advanced-settings/utils.ts
@@ -101,41 +101,39 @@ export const isBoolEnableParameterGroup = (
 export const filterDependentParameters = (
     parameters: TrainingConfigurationParameter[]
 ): TrainingConfigurationParameter[] => {
-    return parameters
-        .reduce<TrainingConfigurationParameter[][]>((acc, curr, _idx, parametersOnTheDepth) => {
-            if (curr.depends_on != null) {
-                const dependentParameter = parametersOnTheDepth.find(
-                    (parameter) => curr.depends_on != null && curr.depends_on[parameter.key] != null
-                );
+    return parameters.reduce<TrainingConfigurationParameter[]>((acc, curr, _idx, parametersOnTheDepth) => {
+        if (curr.depends_on != null) {
+            const dependentParameter = parametersOnTheDepth.find(
+                (parameter) => curr.depends_on != null && curr.depends_on[parameter.key] != null
+            );
 
-                if (dependentParameter === undefined || !isParameter(dependentParameter)) {
-                    acc.push([curr]);
-                    return acc;
-                }
-
-                const dependsOnValue = curr.depends_on[dependentParameter.key];
-                const shouldBeVisible = Array.isArray(dependsOnValue)
-                    ? dependsOnValue.includes(dependentParameter.value)
-                    : dependsOnValue === dependentParameter.value;
-
-                if (shouldBeVisible) {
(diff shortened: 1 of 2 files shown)


application/ui/src/features/models/train-model/advanced-settings/utils.ts
```diff
@@ -101,40 +101,39 @@ export const isBoolEnableParameterGroup = (
...
-
-                    if (Array.isArray(dependsOnValue)) {
-                        return dependsOnValue.includes(curr.value);
-                    }
-
-                    return dependsOnValue === curr.value;
-                });
-
-                acc.push([curr, ...parametersDependingOnCurr]);
+    return parameters.reduce<TrainingConfigurationParameter[]>((acc, curr, _idx, parametersOnTheSameDepth) => {
```
jpggvilaca, line 121: Can we add a comment on this routine? I cannot comprehend what is going on here but i trust your implementation since you also add tests. Let's just add a comment so the next engineer understands what is being done here. Also, np: `!=` -> `!==` : )

dwesolow: If I replace to `!==` I would need to do `parameter !== undefined && parameter !== null`, shorthand is `parameter != null`.

ActiveChooN: `!isNil` from lodash?

application/ui/src/features/models/train-model/advanced-settings/utils.ts
```diff
@@ -101,40 +101,39 @@ export const isBoolEnableParameterGroup = (
...
+        if (curr.depends_on != null) {
+            const dependentParameter = parametersOnTheSameDepth.find(
+                (parameter) => curr.depends_on != null && curr.depends_on[parameter.key] != null
```
ActiveChooN, line 124: Why is an extra `curr.depends_on != null` needed here if it was already checked in the previous condition?

application/ui/src/features/models/train-model/advanced-settings/utils.ts
dwesolow: why to use lodash for such simple thing?

application/ui/src/features/models/train-model/advanced-settings/utils.ts
dwesolow: TypeScript's type narrowing doesn't carry into callback functions passed to methods like .find(). Inside the callback, TypeScript can no longer guarantee that curr.depends_on hasn't changed since the null check on line 105, so it widens the type back to include null | undefined.

application/ui/src/features/models/train-model/advanced-settings/utils.ts
dwesolow: Updated

jpggvilaca (dismissed):

ActiveChooN (dismissed):

jpggvilaca (approved):

PR #6161, opened 2026-04-17
fix: Tool should not reset annotator mode is changed

dwesolow:
### Checklist


jpggvilaca (dismissed):
Did you test the scenario that u have for instance polygon tool selected and then change to prediction (Read only) and you try to draw?

dwesolow:
> Did you test the scenario that u have for instance polygon tool selected and then change to prediction (Read only) and you try to draw?

Yeah, we don't have any tools in the prediction mode.

jpggvilaca (approved):
I remember we wanted the opposite behavior. Did someone request this change? In any case, it makes sense to me

dwesolow:
> I remember we wanted the opposite behavior. Did someone request this change? In any case, it makes sense to me

No one requested, just was testing. It was decided to have opposite behavior because we didn't want to display tools. But it was not a root fix, the root fix was added later. Now this restriction does not make sense.

PR #6166, opened 2026-04-18
Add configurable parameters to control the intensity range of training input data

### LINKED ISSUE #6077: Detect intensity range of training dataset
The training job, when initializing the training configuration, needs to determine the intensity range of the training dataset, so that the training backend can use this information to appropriately export the model with the right pre-processing metadata.

Proposed approach:
- in the training job, after creating the training dataset, inspect the first item to infer the data type (e.g. `np.uint8` -> `"uint8"`)
- when initializing `SubsetConfig`, ensure that its [`intensity`](https://github.com/open-edge-platform/training_extensions/blob/18687db8d352d9de3374d31c0458adfe77ce858c/library/src/otx/config/data.py#L123) parameter is configured accordingly (`mode="scale_to_unit"`,  `max_value` depending on the data type (255, 65535, ...). If necessary, update the GetiConverter.

leoll2:
### Summary

This PR introduce `intensity_mapping` parameters in the training configuration, for all models. These parameters specify how to transform the pixel intensity before training, e.g. rescaling by a fixed factor and/or clipping around a given range. This is particularly important for datasets that use non-standard bit depths, such as 16-bit images. Converters and tests are updated accordingly.

Resolves #6077

### How to test

Train a model and manually inspect the configuration provided to the training backend (getitune): the intensity parameters are correctly serialized.

image

Also verify that the training configuration endpoint returns these parameters:

image

### Checklist


application/backend/app/execution/common/geti_config_converter.py
```diff
@@ -783,6 +790,37 @@ def _disable_deim_framework(config: dict) -> None:
...
+        IntensityConfig format and sets them on train/val/test subsets.
+
+        Args:
+            config: The full OTX config dictionary.
+            intensity_mapping: Dict with keys: mode, max_value, min_value,
+                window_center, window_width, scale_factor.
+        """
+        mode = intensity_mapping.get("mode", "scale_to_unit")
+        intensity_config: dict[str, Any] = {"mode": mode}
+
```
kprokofi, line 807: storage_dtype is never set in the converter. The converter sets no storage_dtype on the IntensityConfig. In getitune, IntensityConfig.storage_dtype defaults to "uint8", which is wrong for 16-bit images.

However, I implemented auto detection on the library, we discussed this previously. My concern: The exporter reads IntensityConfig.storage_dtype to write [("model_info", "input_dtype")] into the IR. Since the app converter never sets storage_dtype and IntensityConfig defaults to "uint8", the exported IR will say input_dtype=u8 even for 16-bit models.

We can set a new parameter in application for storage_dtype or we can rely on the getitune and I will add a propagation to the config on the training side. What do you think?
IMO, adding a propagation step to the config still valid even if we decide to expose dtype parameter in UI explicitly.

application/backend/app/execution/common/geti_config_converter.py
```diff
@@ -783,6 +790,37 @@ def _disable_deim_framework(config: dict) -> None:
...
 
+    @staticmethod
+    def _update_intensity_mapping(config: dict, intensity_mapping: dict) -> None:
+        """Apply intensity mapping parameters to the data subset configs.
+
+        Maps the Geti application intensity_mapping parameters to the library's
+        IntensityConfig format and sets them on train/val/test subsets.
+
+        Args:
+            config: The full OTX config dictionary.
```
kprokofi: please, avoid OTX naming

application/backend/app/api/schemas/training_configuration.py
```diff
@@ -647,6 +647,116 @@ def from_training_configuration(
...
+                                        "type": "parameter",
+                                        "key": "mode",
+                                        "name": "Intensity mapping mode",
+                                        "description": (
+                                            "Strategy used to transform pixel intensities. "
+                                            "'Unit interval scaling' divides by max_value, thus mapping the range "
+                                            "[0, max_value] to [0, 1]. 'Windowing' isolates a specific intensity "
+                                            "range, mapping a specific window (specified with center and "
+                                            "width) to [0, 1] and clipping values outside the window. "
+                                            "'Clipped scaling' multiplies pixel values by a scale factor and clips "
```
kprokofi: getitune's RangeScale does an additional normalization step: (clamped - min) / (max - min) → output in [0, 1]. The description omits this normalization, which may confuse users.

application/backend/app/api/schemas/training_configuration.py
```diff
@@ -647,6 +647,116 @@ def from_training_configuration(
...
+                                            "the result to a specified range (min_value, max_value)."
+                                        ),
+                                        "value": "Unit interval scaling",
+                                        "default_value": "Unit interval scaling",
+                                        "value_type": "str",
+                                        "allowed_values": ["Unit interval scaling", "Windowing", "Clipped scaling"],
+                                    },
+                                    {
+                                        "type": "parameter",
+                                        "key": "max_value",
```
kprokofi, line 681: Is max_value shared across modes?
max_value means "denominator" in scale_to_unit but "upper clip bound" in range_scale. Both modes seems to use the same max_value field. If a user switches modes in the UI, the stale max_value=255 from scale_to_unit will be used as the clip bound in range_scale, which is probably wrong. It is minor one, but should be considered

application/backend/app/models/training_configuration/dataset_preparation.py
```diff
@@ -1,11 +1,108 @@
...
+    """
+    Mode for mapping image intensity values before training.
+
+    Important: the enum keys must match the lowercase strings expected by the training library (getitune),
+    whereas the values correspond to the user-friendly names shown in the UI.
+    """
+
+    SCALE_TO_UNIT = "Unit interval scaling"  # (x / max_value)
+    WINDOW = "Windowing"  # clamp((x - low) / (high - low), 0, 1)
+    RANGE_SCALE = "Clipped scaling"  # clamp(x * factor, min_value, max_value)
```
kprokofi: maybe "Range mapping"? 
Clipped scaling doesn't convey what it's for

application/backend/app/api/schemas/training_configuration.py
```diff
@@ -647,6 +647,116 @@ def from_training_configuration(
                                     },
                                 ],
                             },
+                            {
+                                "type": "parameter_group",
+                                "key": "intensity_mapping",
+                                "name": "Intensity mapping",
+                                "description": (
+                                    "Intensity mapping parameters control how raw pixel values are normalised "
```
kprokofi:
```suggestion
                                    "Intensity mapping parameters control how raw pixel values are normalised to [0, 1] range "
```

application/backend/app/models/training_configuration/dataset_preparation.py
leoll2: In my opinion, "range mapping" or "range scaling" don't make it clear enough that the information outside of the selected range is discarded (clipped). Maybe "Range scaling with clipping"? What do you think?

application/backend/app/api/schemas/training_configuration.py
```diff
@@ -647,6 +647,116 @@ def from_training_configuration(
...
+                                        "type": "parameter",
+                                        "key": "max_value",
+                                        "name": "Maximum pixel value",
+                                        "description": (
+                                            "Maximum possible pixel value in the raw image. "
+                                            "For 8-bit images use 255, for 16-bit images use 65535."
+                                        ),
+                                        "value": 255.0,
+                                        "default_value": 255.0,
+                                        "value_type": "float",
```
A-Artemis, line 689: Aren't these integer values? Can the user enter 204.3 but it will just become 204?

application/backend/app/execution/common/geti_config_converter.py
```diff
@@ -783,6 +790,37 @@ def _disable_deim_framework(config: dict) -> None:
...
+            intensity_mapping: Dict with keys: mode, max_value, min_value,
+                window_center, window_width, scale_factor.
+        """
+        mode = intensity_mapping.get("mode", "scale_to_unit")
+        intensity_config: dict[str, Any] = {"mode": mode}
+
+        if mode == "scale_to_unit":
```
A-Artemis, line 808: Is it possible to reuse the mapping modes you introduced in `class IntensityMappingMode(StrEnum):` ?

application/backend/app/models/training_configuration/dataset_preparation.py
```diff
@@ -1,11 +1,108 @@
...
+        default=1.0,
+        gt=0.0,
+        title="Scale factor",
+        description=(
+            "Multiplicative factor applied to pixel values, before clipping the result to [min_value, max_value]."
+        ),
+        json_schema_extra={"depends_on": {"mode": "Clipped scaling"}},
+    )
+
+    @field_validator("mode", mode="before")
```
A-Artemis, line 103: Where does the validation take place to check that a user doesn't input: `window_center = 0, window_width = 255` ? Is it handled on the dataset preparation side?

application/backend/tests/unit/api/schemas/test_training_configuration_view.py
```diff
@@ -52,6 +54,14 @@ def fxt_training_configuration() -> TrainingConfiguration:
                     min_annotation_objects=MinAnnotationObjects(enable=False, value=1),
                     max_annotation_objects=MaxAnnotationObjects(enable=False, value=50),
                 ),
+                intensity_mapping=IntensityMapping(
+                    mode=IntensityMappingMode.WINDOW,
+                    max_value=255.0,
+                    min_value=0.0,
+                    window_center=200.0,
+                    window_width=400.0,
+                    scale_factor=1.0,
```
A-Artemis, line 64: Can a non-1 scale factor be passed when the mode is not `IntensityMappingMode.RANGE_SCALE` ?

Would it apply two mappings in serial?

application/backend/app/api/schemas/training_configuration.py
leoll2: Good catch, I have updated the description

application/backend/app/api/schemas/training_configuration.py
leoll2: I've just updated the code splitting the parameter in two: `intensity_max_value` (for scale_to_unit) and `clip_max_value` (for range_scale); both are mapped to `max_value` in the getitune configuration, while in the UI they will look like two completely independent parameters and with different descriptions.

application/backend/app/execution/common/geti_config_converter.py
leoll2: I would not expose another parameter in the UI and let `getitune` infer it

application/backend/app/api/schemas/training_configuration.py
leoll2: Using type float is more future proof, in case we want to support images with floating-point pixels (mode `F` in PIL). A value like 204.3 will not be rounded, it will be used as denominator.

application/backend/app/execution/common/geti_config_converter.py
leoll2: I prefer not, to avoid introducing a dependency from the converter code to the app models.

application/backend/app/models/training_configuration/dataset_preparation.py
leoll2: There is no such validation at the moment. `window_center = 0` and `window_width = 255` is unusual but technical valid configuration, it means that [-255, 255] is mapped to [0,1]; in practice, assuming the pixels are positive integers, they will end up in range [0.5, 1].

application/backend/tests/unit/api/schemas/test_training_configuration_view.py
leoll2: When the mode is not `IntensityMappingMode.RANGE_SCALE`, the parameter `scale_factor` has no effect (and is not even shown to the user).

application/backend/app/execution/common/geti_config_converter.py
kprokofi: OK. I will update getitune accordingly

leoll2:
@copilot Rename "Clipped scaling" to "Range scaling with clipping", then adjust the tests accordingly.

application/backend/app/execution/common/geti_config_converter.py
kprokofi: will be propagated in https://github.com/open-edge-platform/training_extensions/pull/6319

kprokofi (approved):

A-Artemis (approved):

PR #6167, opened 2026-04-20
Fix Tauri issue when downloading files; Add new Tauri arch

### LINKED ISSUE #6151: [Windows] Unable to download Media, Datasets, or Models on Tauti Desktop App
In Dataset view, select image/video from dataset -> click Download -> nothing happened

Image

the same when export Dataset:

Image

jpggvilaca:
### Summary

* Add README for tauri installation instructions for all OS's
* [IMPORTANT] Introduce new architecture for handling Tauri logic on our app by leveraging https://rspack.rs/config/resolve#resolveextensions . All details on README [here](https://github.com/open-edge-platform/training_extensions/compare/jvilaca/tauri-issue-downloading-files?expand=1#diff-fc6aac037b70d0b0243be560a8a8d27d7dd4d48e554801fad27f910c1fdb4ef0R14)
* [IMPORTANT] To keep the experience consistent between web app and tauri app, we WILL NOT show the dialog to pick a folder on Tauri, since on the browser YOU CANNOT do this. That means that all downloads on tauri app will automatically start (same as web app). What we need to do moving forward, perhaps trigger a notification letting the user know that the download started (on the browser you see an animation on the top right corner, but on tauri nothing happens). As a bonus we could also add an error notification if the download does not start for some reason.
* Updated rsbuild & added new eslint rule
* Add message to loading screen for increased affordance
* Closes https://github.com/open-edge-platform/training_extensions/issues/6151

Manually tested dowloading datasets/models/images on both webapp and tauri app (MAC OS)

### Checklist


application/ui/package.json
```diff
@@ -36,6 +38,9 @@
         "@geti/smart-tools": "*",
         "@geti/ui": "*",
         "@tanstack/react-query": "~5.80.10",
+        "@tauri-apps/api": "~2.6.0",
```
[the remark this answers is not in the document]
jpggvilaca: Obsolete, I use 2.10.1 now

application/ui/src/platform/download-file.tauri.ts
```diff
@@ -0,0 +1,36 @@
...
+
+const runDownload = async (url: string, name?: string): Promise<void> => {
+    const [{ save }, { download }] = await Promise.all([
+        import('@tauri-apps/plugin-dialog'),
+        import('@tauri-apps/plugin-upload'),
+    ]);
+
+    const path = await save({ defaultPath: name });
+    if (path === null) {
+        return;
```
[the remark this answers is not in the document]
jpggvilaca: Updated

application/ui/src-tauri/README.md
```diff
@@ -0,0 +1,247 @@
...
+
+1. **Plain file is the default.** It runs in both web and Tauri _unless_
+   shadowed by a `.tauri.*` twin.
+2. **`.tauri.*` files may import `@tauri-apps/*`.** Other source files may
+   not — this is enforced by the `no-restricted-imports` rule in
+   [`../eslint.config.js`](../eslint.config.js).
+3. **Tauri-only features:** ship a no-op/null-returning module as the default
+   and the real implementation in `.tauri.tsx`. Consumers render/call
+   unconditionally; the web build tree-shakes the no-op away.
+4. **Tauri-only styles:** same trick with `.scss` / `.tauri.scss` (set up the
```
dwesolow: TBH I would add this to module resolution now so we won't skip that when we really need `tauri.scss`

application/ui/src/platform/download-file.tauri.ts
```diff
@@ -0,0 +1,36 @@
...
+// neither restriction applies. Blob URLs are same-origin to the webview, so
+// the standard anchor flow still works for them.
+
+import { downloadFile as downloadBlobViaAnchor } from './download-file';
+
+export const downloadFile = (url: string, name?: string): void => {
+    if (url.startsWith('blob:')) {
+        downloadBlobViaAnchor(url, name);
+
+        return;
```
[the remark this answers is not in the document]
jpggvilaca: Updated

application/ui/package.json
```diff
@@ -9,12 +9,14 @@
...
     "scripts": {
         "build": "rsbuild build",
+        "build:tauri": "BUILD_TARGET=tauri rsbuild build",
         "build:api": "npx openapi-typescript ./src/api/openapi-spec.json -o ./src/api/openapi-spec.d.ts  --root-types && prettier --write src/api/openapi-spec.d.ts",
         "build:api:download": "curl -o ./src/api/openapi-spec.json http://localhost:7860/api/openapi.json && prettier --write src/api/openapi-spec.json",
         "update-spec": "npm run build:api:download && npm run build:api",
         "clone-geti-ui-packages": "npx tiged --force --mode=git https://github.com/open-edge-platform/geti.git/web_ui/packages/config#ecd4b2a3049eba7c28c53ed81372c92342c7734a packages/config && npx tiged --force --mode=git https://github.com/open-edge-platform/geti.git/web_ui/packages/ui#ecd4b2a3049eba7c28c53ed81372c92342c7734a packages/ui && npx tiged --force --mode=git https://github.com/open-edge-platform/geti.git/web_ui/packages/smart-tools#ecd4b2a3049eba7c28c53ed81372c92342c7734a packages/smart-tools",
         "cyclic-deps-check": "npx madge --circular src/**/*.ts*",
         "start": "rsbuild dev",
+        "start:tauri": "BUILD_TARGET=tauri rsbuild dev",
```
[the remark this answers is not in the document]
jpggvilaca: Updated

application/ui/eslint.config.js
```diff
@@ -73,6 +73,54 @@ export default [
...
     },
+    {
+        // Contain Tauri-specific APIs inside `*.tauri.{ts,tsx}` files under
+        // `src/platform/`. The bundler picks those files for the Tauri build
+        // via `resolve.extensions`, so only they should ever import
+        // `@tauri-apps/*`. Everywhere else, consume the capability module
+        // (e.g. `import { downloadFile } from '../platform/download-file'`)
+        // and let the bundler swap the implementation per target.
+        files: ['src/**/*.{ts,tsx}'],
+        ignores: ['src/**/*.tauri.{ts,tsx}'],
```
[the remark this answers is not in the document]
jpggvilaca: Updated

application/ui/src-tauri/README.md
jpggvilaca: Not sure if overkill but added

application/ui/src-tauri/capabilities/default.json
```diff
@@ -3,5 +3,9 @@
     "identifier": "default",
     "description": "enables the default permissions",
     "windows": ["main"],
-    "permissions": ["core:default", "core:window:allow-on-close-requested"]
```
jpggvilaca, line 6: "core:window:allow-on-close-requested" is included on "core:default"

application/ui/src/platform/download-file-anchor.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright (C) 2026 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+export const downloadViaAnchor = (url: string, name?: string): void => {
```
jpggvilaca: we need this file for this case specifically because both tauri and web leverage this and since we're using module resolution, we cannot have cyclic imports. Same as what happened to us before with types/interfaces

application/ui/src/platform/download-file.tauri.ts
```diff
@@ -27,10 +27,23 @@ const runDownload = async (url: string, name?: string): Promise<void> => {
...
 
     await download(url, path);
 };
+
+// Strip path separators and characters that are invalid in filenames on
+// Windows (and noisy on POSIX), so a user-controlled `name` can't produce
+// an invalid or confusing default path in the save dialog.
+const sanitizeFilename = (name: string): string => {
+    const cleaned = name
+        .replace(/[<>:"/\\|?*\x00-\x1F]/g, '_')
```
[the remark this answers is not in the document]
jpggvilaca: Not much i can do here ^^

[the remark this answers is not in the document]
dwesolow: I'd decide to stick to strict name then or explain the regexp in the comment there

application/ui/src/platform/storage-cleanup.ts
```diff
@@ -0,0 +1,7 @@
+// Copyright (C) 2026 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+// Default (web) implementation: nothing to clean up on close because
+// browsers tear down the tab on their own. The Tauri build replaces this
+// file with `storage-cleanup.tauri.ts` via `resolve.extensions`.
+export const setupStorageCleanup = (): void => {};
```
dwesolow, line 7: IMO this should include if we want to follow the same behavior like desktop app.
```
const cleanDatasetStorage = (): void => {
    Object.keys(localStorage)
        .filter(isDatasetStorageKey)
        .forEach((key) => localStorage.removeItem(key));
};
```

application/ui/src/platform/download-file-anchor.ts
```diff
@@ -0,0 +1,17 @@
...
+    const link = document.createElement('a');
+
+    link.href = url;
+    if (name !== undefined) {
+        link.download = name;
+    }
+    link.hidden = true;
+    link.click();
+
+    if (url.startsWith('blob:')) {
```
camiloHimura: 👍🏼

application/ui/src/platform/download-file.tauri.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export const downloadFile = (url: string, name?: string): void => {
+    if (url.startsWith('blob:')) {
+        downloadViaAnchor(url, name);
+
+        return;
+    }
+
+    void runDownload(url, name).catch((error: unknown) => {
+        console.error('[tauri downloadFile] failed', error);
```
camiloHimura, line 19: Should we show a toast error? The user might think nothing happened.

application/ui/src/platform/download-file.tauri.ts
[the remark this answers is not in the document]
camiloHimura: Let’s use unit tests as documentation so we can understand what it does through a few examples.

application/ui/eslint.config.js
```diff
@@ -73,6 +73,54 @@ export default [
...
+                        {
+                            group: ['@spectrum-icons'],
+                            message: 'Use icons from the @geti/ui/icons folder instead.',
+                        },
+                        {
+                            group: ['src/*'],
+                            message: 'Use relative imports instead of absolute "src/" imports.',
+                        },
+                        {
+                            group: ['@tauri-apps/*'],
```
ActiveChooN: Neat, but does it mean that other rules will be disabled for `*.tauri.ts(x)` files. Should be this part be extracted as a separate rule?
```js
{
    files: ['src/**/*.{ts,tsx}'],
    ignores: ['src/platform/**/*.tauri.{ts,tsx}'],
    ...
                            group: ['@tauri-apps/*'],
                            message: "..."
}
```

application/ui/src/platform/download-file.tauri.ts
jpggvilaca: Agree, added a comment on the PR description

application/ui/src-tauri/README.md
```diff
@@ -0,0 +1,250 @@
...
+   [`../eslint.config.js`](../eslint.config.js).
+3. **Tauri-only features:** ship a no-op/null-returning module as the default
+   and the real implementation in `.tauri.tsx`. Consumers render/call
+   unconditionally; the web build tree-shakes the no-op away.
+4. **Tauri-only styles:** same trick with `.scss` / `.tauri.scss`. The
+   `.tauri.scss` and `.scss` extensions are already in `resolve.extensions`,
+   so an extensionless import (`import './foo'`) resolves to `foo.tauri.scss`
+   on the desktop build and `foo.scss` on the web build. Drop the extension
+   on the import site to opt in to the override.
+5. **No `isTauri()` runtime checks anywhere.** If you find yourself reaching
```
ActiveChooN: Slick. I think it should be enforced on the linter level as well

application/ui/src-tauri/README.md
```diff
@@ -0,0 +1,250 @@
...
+### Conventions
+
+```
+src/platform/
+  download-file.ts         ← web (default) implementation
+  download-file.tauri.ts   ← tauri override (fetches via webview and triggers a blob anchor download to Downloads/)
+  …                        ← future capabilities follow the same pair
+```
+
+Rules of thumb when adding a platform-specific behaviour:
```
ActiveChooN, line 59: Do we have any way to check if exported elements of `*.tauri.ts(x)` twin files are type-compatible with main ones

application/ui/src/platform/storage-cleanup.ts
jpggvilaca: We dont need any storage cleanup for webapps, it happens automatically

application/ui/src-tauri/README.md
```diff
@@ -0,0 +1,250 @@
...
+
+The unselected file is **not parsed and never enters the module graph**, so
+`@tauri-apps/*` imports cannot leak into the web bundle and the web fallbacks
+cannot bloat the desktop bundle.
+
+### Conventions
+
+```
+src/platform/
+  download-file.ts         ← web (default) implementation
```
ActiveChooN: What if I wanted to add platform-specific behaviour to a pre-existing component from the '@geti-ui' package, for example? Would we agree to create a file containing plain re-exports in the `src/platform` directory, alongside a Tauri-specific twin?

application/ui/src-tauri/README.md
jpggvilaca: Nope. But we could in the future improve by adding some sort of assertion. For now i'd rather keep it simple as the PR is big enough, but good idea!

application/ui/src-tauri/README.md
jpggvilaca: I think in this case duplication triumphs over flexibility. Let's keep it simple and not mix concerns at all

application/ui/src-tauri/README.md
jpggvilaca: Amazing idea, added!

application/ui/eslint.config.js
jpggvilaca: Great catch! you're right, moved to shared rules

application/ui/src/platform/download-file.tauri.ts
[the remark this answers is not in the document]
jpggvilaca: Obsolete, removed

application/ui/src-tauri/README.md
```diff
@@ -0,0 +1,252 @@
...
+| Tauri         | `.tauri.tsx`, `.tauri.ts`, `.ts` | `download-file.tauri.ts` (wins) |
+
+The unselected file is **not parsed and never enters the module graph**, so
+`@tauri-apps/*` imports cannot leak into the web bundle and the web fallbacks
+cannot bloat the desktop bundle.
+
+### Conventions
+
+```
+src/platform/
```
dwesolow: I'm not sure if we should move every piece of that app that is platform dependent to `src/platform`. As Dima mentioned below, what if anything related to component that is deeply nested in the directory tree should be specific to tauri?
What if we keep them in the place where the are currently but we would create `tauri.ts{x}`?

IMO it should never happen to dumb components (they only do visualisation) and only utils would require doing platform twins.
Thoughts?

application/ui/src/platform/download-file.tauri.ts
```diff
@@ -0,0 +1,51 @@
...
+// path segment) is used as the filename.
+
+export const downloadFile = (url: string, name?: string): void => {
+    if (url.startsWith('blob:')) {
+        downloadViaAnchor(url, name);
+
+        return;
+    }
+
+    void autoDownload(url, name).catch((error: unknown) => {
```
dwesolow, line 18: When `autoDownload` is called?

application/ui/src/platform/download-file.tauri.ts
```diff
@@ -0,0 +1,51 @@
...
+
+const autoDownload = async (url: string, name?: string): Promise<void> => {
+    const response = await fetch(url);
+    if (!response.ok) {
+        throw new Error(`download failed: ${response.status} ${response.statusText}`);
+    }
+
+    // The urls for downloading assets end with /binary, so we take the preceding segment as the filename.
+    // If that segment is missing for some reason, we fall back to a generic name.
+    const segments = new URL(url).pathname.split('/').filter(Boolean);
```
dwesolow, line 31: what about content disposition header?

application/ui/src/platform/download-file.tauri.ts
```diff
@@ -0,0 +1,51 @@
...
+    const filename = name ?? segments.at(-2) ?? 'download';
+    const blobUrl = URL.createObjectURL(await response.blob());
+
+    downloadViaAnchor(blobUrl, filename);
+};
+
+const downloadViaAnchor = (url: string, name?: string): void => {
```
dwesolow, line 38: it's exactly `downloadFile`, what about creating shared download file for both platforms, e.g.
`download-file-via-anchor.ts`
`download-file.tauri.ts`
`download-file.ts`.

application/ui/rsbuild.config.ts
```diff
@@ -9,6 +9,18 @@ import { pluginSvgr } from '@rsbuild/plugin-svgr';
...
+// prepend `.tauri.*` extensions so the bundler resolves platform-specific
+// overrides (e.g. `foo.tauri.ts` wins over `foo.ts`). Files not shadowed by
+// a `.tauri.*` twin resolve as usual. This keeps Tauri-specific code out of
+// the web graph entirely, and removes the need for runtime `isTauri` checks.
+const isTauriBuild = process.env.BUILD_TARGET === 'tauri';
+const platformExtensions = isTauriBuild ? ['.tauri.tsx', '.tauri.ts', '.tauri.jsx', '.tauri.js', '.tauri.scss'] : [];
+// `.scss` is appended unconditionally so extensionless SCSS imports (used
+// to opt in to the platform-override mechanism, e.g. `import './foo'`)
+// still resolve to `foo.scss` on the web build.
+const styleExtensions = ['.scss'];
```
dwesolow, line 22: What about `.css`? We have `index.css` file in the root.

application/ui/src/platform/download-file.tauri.ts
jpggvilaca: this takes from the url. The url will be something like `../model_id/binary`

application/ui/src/platform/download-file.tauri.ts
jpggvilaca: Reverted **[here](https://github.com/open-edge-platform/training_extensions/pull/6167/changes/d87a9ebe0672212a0df298be990a6aa5026616c4)** . It separates more but then we might end up having 3 files for each function instead of 2. This is a specific case but still, readability > #loc imho

application/ui/rsbuild.config.ts
jpggvilaca: We dont use `.css` in the components, only here, for some reason...

application/ui/src/platform/download-file.tauri.ts
jpggvilaca: Exactly here. `void` is fire-and-forget

application/ui/src-tauri/README.md
jpggvilaca: I agree with you both. I will make the change

jpggvilaca: Updated

application/ui/eslint.config.js
```diff
@@ -73,6 +95,43 @@ export default [
             'max-len': ['off'],
         },
     },
+    {
+        // Relax the `@tauri-apps/*` import restriction for Tauri twin files.
+        // The `.tauri.{ts,tsx}` extension is the contract — these files may
+        // live anywhere under `src/` next to their consumers.
+        files: ['src/**/*.tauri.{ts,tsx}'],
```
dwesolow: IMO we should not copy all of these rules that are required globally.
We should do:
1. All files must follow spectrum paths etc (the current restriction).
2. Files that have `{ts,tsx}` and no `tauri` must follow:
```
{
          group: ['@tauri-apps/*'],
          message:
                                'Import Tauri plugins only from `*.tauri.{ts,tsx}` files. Consumers should ' +
                                'import the capability module (e.g. ./download-file) so the bundler can ' +
                                'swap implementations per build target.',
},
```

jpggvilaca: Totally agree. Updated

warrkan:
The documentation part looks nice, thanks @jpggvilaca !

camiloHimura (approved):

application/ui/src-tauri/README.md
```diff
@@ -0,0 +1,253 @@
...
+rebuilt whenever the backend Python source changes.
+
+### 1. Build the backend
+
+From `application/backend`:
+
+```sh
+just pyinstaller            # CPU build
+# or, for accelerated builds:
+just pyinstaller -a xpu     # Intel GPU
```
ActiveChooN, line 114: This part doesn't work for me out of the box. I had to write Git/bin and GIt/usr/bin to the PATH var

application/ui/src-tauri/README.md
```diff
@@ -0,0 +1,253 @@
...
+
+1. Install [Microsoft Visual Studio C++ Build Tools](https://visualstudio.microsoft.com/visual-cpp-build-tools/)
+   with the **Desktop development with C++** workload (provides the MSVC
+   linker required by Rust).
+2. Install [WebView2 runtime](https://developer.microsoft.com/microsoft-edge/webview2/)
+   (already present on Windows 11 and most up-to-date Windows 10 installs).
+3. Install Rust via [rustup](https://rustup.rs).
+4. From `application/ui` (PowerShell or `cmd`), run the desktop dev shell:
+    ```powershell
+    npm run start:desktop
```
ActiveChooN, line 223: Build works fine, however when I try to run tauri in with this command I'm getting weird CORS errors

ActiveChooN (approved):
I've checked this doc for Windows. Everything works great, except for two minor issues.

application/ui/src-tauri/README.md
jpggvilaca: But does the app run or it crashes?

application/ui/src-tauri/README.md
jpggvilaca: What was the issue you had? and is it windows?

ActiveChooN: I've only tested the build on Windows. It says cygpath.exe and something else are missing, and without Git binaries in the path, it was not able to run just recipes with Bash scripts.

application/ui/src-tauri/README.md
ActiveChooN: It runs, but it fails to load anything because every request fails with CORS error

application/ui/src/platform/storage-cleanup.ts
dwesolow: local storage is not cleaned automatically, session storage is, no?

application/ui/src/platform/download-file.tauri.ts
dwesolow: I meant when the first `if` won't be truthy and this fallback `autoDownload` will be triggerred (the real case scenario)

application/ui/src/platform/download-file.tauri.ts
dwesolow: Yeah but why from the URL and not from the content disposition?

application/ui/src/platform/download-file.tauri.ts
dwesolow: It's not about loc but about feature implementation

application/ui/src-tauri/README.md
jpggvilaca: Thank you, update readme

application/ui/src/platform/download-file.tauri.ts
jpggvilaca: wdym?

dwesolow: I mean to have one feature implementation (in this case downloading a file from the server) in one place. And other places can be something like adapters, i.e.
1. `download-file.feature.ts`
2. `download-file.tauri.ts` uses 1.
3. `download-file.ts` uses 1.

application/ui/src/platform/download-file.tauri.ts
jpggvilaca: Because ``<a download>` is ignored by WKWebView/Chromium for cross-origin URLs` (comment lines 4-9)

application/ui/src/platform/download-file.tauri.ts
jpggvilaca: I had that in the beginning, but then i thought this would backfire and instead of 2 files per util, we would have 3, that is why i preferred repetition. But on my next PR i'll add 3 and see if this is a special case. If i see that on the next util we also need 3, then we revert and use 2 + repetition

